PR #5, opened 2018-04-25
Testing with more environments

Saibamen:


mproch:
Thanks for PR. However, at the moment I think most of environments just won't work - I'm pretty sure we need at least scala 2.11, and flink currently does not support scala 2.12. What's more we have scala version specified in sbt - we are not cross-building nussknacker. Do you have any particular reason you'd like to see build with different scala versions?

Saibamen:
To be more stable and support more environments? For now only 2 envs runs correctly: 
- Scala: 2.11.8, JDK: oraclejdk8
- Scala: 2.11.8, JDK: openjdk8

> I'm pretty sure we need at least scala 2.11

So please make it *really* sure and update your readme.md

mproch:
https://github.com/TouK/nussknacker/blob/master/README.md - I updated readme, after longest weekend in year we'll correct your PR, thanks

Saibamen:
Status?

arkadius:
Closing because of inactivity in this PR.

PR #15, opened 2018-07-01
Decrease memory requirements of demo and remove unnecessary Docker im…

Crack:
Memory allocation changes (MB):
* Kafka: 1024 -> <128, 512>
* Elasticsearch: 2048 -> <128, 512>
* Flink JobManager: 500 -> 256
* Flink JobManager: 10000 -> 1000

I also removed custom images in places when they aren't needed.


mproch:
Thanks @Crack, now it should run with much fewer resources :)

PR #24, opened 2019-03-25
Feature MapVariable

lciolecki:
*Features*:
- Possibility to add / edit Map Variable node type 

*Updates*:
- Some refactor of Node Modal 
- Separate views (stateless Components) for NodeModal

*Fixes*:
- AceEditor: `Automatically scrolling cursor into view after selection change this will be disabled in the next version set editor.$blockScrolling = Infinity to disable this message`
- Ladda Button: `Unknown prop 'loading' on button tag`


oytuntez:
Great! Is there an example on how to use Variable Objects in custom processes?

lciolecki:
> Great! Is there an example on how to use Variable Objects in custom processes?

lciolecki:
> Great! Is there an example on how to use Variable Objects in custom processes?

Actually this node type was always available.. But only by import. I have added option to create and edit this type. I will add some screen and documentation for this.

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -13,24 +13,27 @@ import * as JsonUtils from '../../common/JsonUtils';
...
       codeCompletionEnabled: true,
-      testResultsToHide: new Set()
-    }
-    this.state = {
-      ...this.state,
-      ...TestRenderUtils.stateForSelectTestResults(null, this.props.testResults)
+      testResultsToHide: new Set(),
+    };
+
+    if (_.has(this.state.editedNode, "fields")) {
```
mproch: why?

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -46,12 +49,27 @@ export class NodeDetailsContent extends React.Component {
...
+
+  addElement = (property, element) => {
+    if (_.has(this.state.editedNode, property)) {
+      _.get(this.state.editedNode, property).push(element);
+      this.setState({editedNode: this.state.editedNode})
+      this.props.onChange(this.state.editedNode)
+    }
+  };
 
+  customNode = () => {
```
mproch: maybe let's leave TODO to extract other types to components?

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -428,18 +430,18 @@ export class NodeDetailsContent extends React.Component {
...
     return (
       <div className="node-label" title={label}>{label}:
         {parameter ? <div className="labelFooter">{ProcessUtils.humanReadableType(parameter.typ.refClazzName)}</div> : null}
     </div>)
-  }
+  };
 
   render() {
-    var nodeClass = classNames('node-table', {'node-editable': this.props.isEditMode})
+    let nodeClass = classNames('node-table', {'node-editable': this.props.isEditMode})
```
mproch: we can event have const here ;)

docs/DesigningProcesses.md
```diff
@@ -126,7 +126,16 @@ Variable once defined cannot be overwritten.
 
 ![split graph](img/variable_window.png)
   
-  
+## Variable Object
```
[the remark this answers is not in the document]
lciolecki: Okey - I will change this.

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -108,7 +126,7 @@ export class NodeDetailsContent extends React.Component {
             {this.createReadonlyField("input", "Service Id", "service.id")}
             {this.state.editedNode.service.parameters.map((param, index) => {
               return (
-                <div className="node-block" key={this.props.node.id + param.name + index}>
+                <div className="node-block"key={this.props.node.id + "-service-parameters" + uuid4()}>
```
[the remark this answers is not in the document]
lciolecki: Documentation says (https://reactjs.org/docs/lists-and-keys.html#keys): 
` Keys should be given to the elements inside the array to give the elements a stable identity`
In this case when you change param.name element is rerender - key will alsow change and you lose context of input.

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -257,7 +259,7 @@ export class NodeDetailsContent extends React.Component {
 
       return <ExpressionWithFixedValues
         fieldLabel={fieldConfig.label}
-        onChange={(newValue) => this.setNodeDataAt(`additionalFields.properties.${fieldName}`, newValue)}
+        onChange={(newValue) => this.changeElement(`additionalFields.properties.${fieldName}`, newValue)}
```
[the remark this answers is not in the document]
lciolecki: Right, my bad!

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -13,24 +13,34 @@ import * as JsonUtils from '../../common/JsonUtils';
...
-    this.state = {
-      ...this.state,
-      ...TestRenderUtils.stateForSelectTestResults(null, this.props.testResults)
-    }
-    this.processNode = _.cloneDeep(props.node); //FIXME: props are modified without cloning @{ParameterList} breakks
-    this.nodeObjectDetails = ProcessUtils.findNodeObjectTypeDefinition(this.props.node, this.props.processDefinitionData.processDefinition)
+      testResultsToHide: new Set(),
+    };
+
+    this.generateUUID("fields", "service.parameters", "ref.parameters", "exceptionHandler.parameters");
```
arkadius: For `service.parameters`, `ref.parameters` and `exceptionHandler.parameters` it is redundant because in these cases names are unique - they will be changed if components would be changed in extension library

ui/client/components/graph/NodeModal/MapVariable.js
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    return (
+        <div className="node-table-body node-variable-builder-body">
+            <Input label="Id" value={node.id} path="id" onChange={onChange} isMarked={isMarked("id")} readOnly={readOnly} />
+            <Input label="Variable Name" value={node.varName} path="varName" onChange={onChange} isMarked={isMarked("varName")} readOnly={readOnly} />
+            <div className="node-row">
+                <div className="node-label" title="Fields">Fields:</div>
+                <div className="node-value">
+                    <div className="fieldsControl">
```
arkadius: What do you think about extracting component for this table?

ui/server/src/test/scala/pl/touk/nussknacker/ui/definition/DefinitionPreparerSpec.scala
```diff
@@ -59,7 +59,7 @@ class DefinitionPreparerSpec extends FunSuite with Matchers with TestPermissions
 
     val baseNodes = baseNodeGroups.flatMap(_.possibleNodes)
     // 5 nodes from base + 3 custom nodes
-    baseNodes should have size (5 + 3)
+    baseNodes should have size (6 + 3)
```
arkadius: Please fix the comment above or maybe introduce  constants `baseNodesCount`, `customNodesCount`?

ui/server/src/test/scala/pl/touk/nussknacker/ui/definition/DefinitionPreparerSpec.scala
```diff
@@ -80,7 +80,7 @@ class DefinitionPreparerSpec extends FunSuite with Matchers with TestPermissions
...
```
arkadius: Please fix the comment above or maybe introduce  constants `baseNodesCount`, `customNodesCount`?

docs/DesigningProcesses.md
```diff
@@ -126,7 +126,16 @@ Variable once defined cannot be overwritten.
 
 ![split graph](img/variable_window.png)
   
-  
+## Map Variable 
+
+Evaluates `Expression` and assigns to `MapVariable.FieldName`. Allows
```
arkadius: Maybe better would be something like this: Defines new variable with one or more fields. Each field value is evaluated using `Expression`.

docs/DesigningProcesses.md
```diff
@@ -126,7 +126,16 @@ Variable once defined cannot be overwritten.
...
+to assign many of `Expression`.
+
+![split graph](img/variable_object_window.png)
+
+Doesn't change records flow. 
+Have to have exactly one output.
+MapVariable once defined cannot be overwritten.
```
arkadius: What do you mean by this? Do you mean that defined variable cannot be overwritten by subsequent operations? What will happen if smb try to do this?

lciolecki: When I was writing this then I suggested by Variable documentation - I thought that works in the same way ;) I mean that when you once declare variable then you can't overwrite these var by another var - maybe I'm wrong?

docs/DesigningProcesses.md
lciolecki: Changed

docs/DesigningProcesses.md
arkadius: I don't know, probable it works the same way.

PR #25, opened 2019-04-04
Fix: OutputVar

lciolecki:
*Fixes*:
- Block put outputVar at CustomTransformer when response is VOID
- Back compability: possibility to remove value when it was set before


ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -14,6 +14,8 @@ import Fields from "../Fields";
 import ParameterList from "./ParameterList";
 import ExpressionWithFixedValues from "./ExpressionWithFixedValues";
 
+const VOID_TYPE = "java.lang.Void";
```
arkadius: Instead of this, please modify logic of preparing `returnType` field in schema. Now it is very technical information - it leaks that we have Void/Unit/... markers for situation when no new variable is added to context. I suggest to change this logic to returning null in this case.

ui/client/common/ProcessUtils.js
```diff
@@ -63,7 +63,14 @@ class ProcessUtils {
 
   //FIXME: handle source/sink/exceptionHandler properly here - we don't want to use #input etc here!
   _findVariablesBasedOnGraph = (nodeId, process, processDefinition) => {
-    const globalVariables = _.mapValues(processDefinition.globalVariables, (v) => {return v.returnType})
+    const filteredGlobalVariables = _.reduce(processDefinition.globalVariables, function(result, variable, key) {
```
arkadius: please replace with pick

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjects.scala
```diff
@@ -97,38 +98,57 @@ case class UIProcessObjects(nodesToAdd: List[NodeGroup],
...
+                              returnType: Option[TypingResult],
+                              categories: List[String],
+                              nodeConfig: SingleNodeConfig)
+
+
+object UIObjectDefinition {
+  def apply(objectDefinition: ObjectDefinition): UIObjectDefinition = {
+    UIObjectDefinition(
+      parameters = objectDefinition.parameters,
+      returnType = if (objectDefinition.hasNoReturn) None else Some(objectDefinition.returnType),
```
arkadius: Please add test for this case

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -25,7 +25,7 @@ class DefinitionResources(modelData: Map[ProcessingType, ModelData],
     //TODO maybe always return data for all subprocesses versions instead of fetching just one-by-one?
     path("processDefinitionData" / Segment) { (processingType) =>
       parameter('isSubprocess.as[Boolean]) { (isSubprocess) =>
-        post {
+        post { // POST - because there is complicated logic of retrieving data
```
arkadius: Logic of retrieving is not complicated. Only query parameter (subprocessVersions) is complex and it is the reason why POST was used

lciolecki: Changed.

PR #30, opened 2019-04-17
Unmarshall process additional fields to raw json

jedrz:
This PR adds possibility to access additional fields from process meta data inside services. For now the fields are not deserialized and are None in the result. The downside is the fields are exposed as raw json that needs to be manually parsed but it is actually an instance of `ProcessAdditionalFields` from `restmodel` module.

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
@@ -1,17 +1,19 @@
 package pl.touk.nussknacker.engine.marshall
 
+import argonaut.Argonaut._
+import argonaut.ArgonautShapeless._
 import argonaut._
-import Argonaut._
-import ArgonautShapeless._
 import argonaut.derive._
 import cats.data.Validated
-import pl.touk.nussknacker.engine.api.{MetaData, ProcessVersion, TypeSpecificData, UserDefinedProcessAdditionalFields}
+import pl.touk.nussknacker.engine.api._
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
 import pl.touk.nussknacker.engine.canonicalgraph.canonicalnode._
 import pl.touk.nussknacker.engine.canonize.ProcessCanonizer
-import pl.touk.nussknacker.engine.graph.{EspProcess, node}
 import pl.touk.nussknacker.engine.graph.node.{Case => _, FilterNode => _, SplitNode => _, SwitchNode => _, _}
+import pl.touk.nussknacker.engine.graph.{EspProcess, node}
 import pl.touk.nussknacker.engine.marshall.ProcessUnmarshallError._
+
+import scala.reflect.ClassTag
 
 class ProcessMarshaller(implicit
                         additionalNodeDataFieldsCodec: CodecJson[Option[node.UserDefinedAdditionalNodeFields]] = ProcessMarshaller.additionalNodeDataFieldsCodec,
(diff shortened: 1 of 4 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
```diff
@@ -125,5 +125,8 @@ object ProcessMarshaller {
 
   val additionalNodeDataFieldsCodec: CodecJson[Option[node.UserDefinedAdditionalNodeFields]] = emptyOptionCodec
 
-  val additionalProcessFieldsCodec: CodecJson[Option[UserDefinedProcessAdditionalFields]] = emptyOptionCodec
+  val additionalProcessFieldsCodec: CodecJson[Option[UserDefinedProcessAdditionalFields]] = CodecJson.apply(
+    _ => jEmptyObject,
```
bottaio: Is encoder even needed in this case? It seems superficial right now.

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshallerSpec.scala
```diff
@@ -39,10 +72,7 @@ class ProcessMarshallerSpec extends FlatSpec with Matchers with OptionValues wit
         .enricher("d", "barVar", "dService", "p1" -> "expr3")
         .switch("f", "expr4", "eVar", nestedGraph("e"), Case("e1", GraphBuilder.emptySink("endE1", "")))
 
-    val marshalled = ProcessMarshaller.toJson(process, PrettyParams.spaces2)
-
-    val unmarshalled = ProcessMarshaller.fromJson(marshalled).toOption
-    val result = ProcessCanonizer.uncanonize(unmarshalled.value).toOption
+    val result = marshallAndUnmarshall(process)
```
bottaio: Use `marshallAndUnmarshall` helper in other tests too.

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshallerSpec.scala
```diff
@@ -59,50 +89,41 @@ class ProcessMarshallerSpec extends FlatSpec with Matchers with OptionValues wit
...
+    val process = EspProcessBuilder
+      .id("process1")
+      .additionalFields(TestAdditionalFields(field1 = "additional field 1", field2 = 2))
+      .exceptionHandler()
+      .source("a", "")
+      .processorEnd("d", "dService", "p1" -> "expr3")
+
+    val result = marshallAndUnmarshall(process)
 
+    result.value.metaData.additionalFields shouldBe 'empty
```
bottaio: It doesn't seem right at first glance. Add a comment if this is intended behaviour.

jedrz:
I think better idea can be to duplicate `ProcessAdditionalFields` and `NodeAdditionalFields` in `interpreter` module. Later we can delete `UserDefinedAdditionalNodeFields` trait and classes in `restmodel` module.

bottaio:
> I think better idea can be to duplicate `ProcessAdditionalFields` and `NodeAdditionalFields` in `interpreter` module. Later we can delete `UserDefinedAdditionalNodeFields` trait and classes in `restmodel` module.

Sounds good to me, waiting for the follow up!

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshallerSpec.scala
jedrz: We decided to rework the solution.

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshallerSpec.scala
jedrz: Done.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
jedrz: Reworked the codecs.

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshallerSpec.scala
```diff
@@ -91,18 +105,62 @@ class ProcessMarshallerSpec extends FlatSpec with Matchers with OptionValues wit
...
+        |    ]
+        |}
+      """.stripMargin
+
+    inside(ProcessMarshaller.fromJson(withAdditionalFields)) { case Valid(process) =>
       process.metaData.id shouldBe "custom"
       process.metaData.additionalFields shouldBe None
-      process.nodes.head.data.additionalFields shouldBe None
+      process.nodes should have size 1
+      process.nodes.head.data.additionalFields shouldBe Some(NodeAdditionalFields(description = None))
```
bottaio: would expect it to be None; nonetheless, it's ok.

jedrz: Yeah, it looks strange. I need to handle empty fields manually so it's the same for process additional fields.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
```diff
@@ -121,9 +123,41 @@ class ProcessMarshaller(implicit
...
+
+  val additionalProcessFieldsCodec: CodecJson[Option[UserDefinedProcessAdditionalFields]] = {
+    val derivedCodec = derivedTypeOrNoneCodec[UserDefinedProcessAdditionalFields, ProcessAdditionalFields]
+    CodecJson.derived(
+      derivedCodec.Encoder,
+      DecodeJson(c => {
+        val withEmptyWhenMissing = (c --\ "metaData" --\ "additionalFields")
+          .reattempt
+          .withFocus(_.withObject(additionalFields => {
+            val withGroups = additionalFields("groups")
```
bottaio: It is a little bit confusing to read at the moment. Consider something similar to `additionalFields + ("groups" -> additionalFields("groups").getOrElse(jEmptyArray))` to make it more readable?

Also, I don't like the idea of accessing fields by their names but I don't know if you can easily get around it.

jedrz: I decided to decode each field explicitly. In case of empty option, field is decode to empty set/empty map.

bottaio: Alright, this looks good to me. Unfortunatelly I don't have the power to approve it. @mproch @nadberezny?

PR #35, opened 2019-05-06
Upload jar after job manager restart. Show deployment errors on FE

bottaio:
After job manager restart all uploaded jars cannot be longer used and we have to send them once again. This PR also fixes deployment error appearance on FE - currently deployment always seems to be successful which is not always a case.

### DIFF
--- a/engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
+++ b/engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
@@ -28,7 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 private[management] trait HttpSender {
   def send[T](pair: (Request, FunctionHandler[T]))
-                  (implicit executor: ExecutionContext): Future[T]
+             (implicit executor: ExecutionContext): Future[T]
 }
 
 private[management] object DefaultHttpSender extends HttpSender {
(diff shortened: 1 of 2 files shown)


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -58,17 +60,48 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, sender: HttpSe
 
   private val flinkUrl = dispatch.url(config.restUrl)
 
-  private lazy val uploadedJarId : Future[String] = uploadCurrentJar()
+  // after job manager restart old resources are not available anymore and we have to upload jar once again
+  private var jarUploadedBeforeLastRestart: Option[Future[String]] = None
```
mproch: I wonder if some synchronization would be needed here? Or we rely on ManagementActor which prevents concurrent deploys? Then maybe a comment why we don't do it here?

bottaio: Yeah, everything is synchronous here (at least for now). But I see your point, I've introduced a synchronized block - we'd better be safe than sorry.

mproch: I think for now better add just comment - synchronized is no longer enough, since we do the check asynchronously?

bottaio: Right, totally missed the fact that updates are now done within a Future.

PR #38, opened 2019-05-09
Feature: ci automation, better docker support

lciolecki:
Feature:
- Building docker image for each commit
- Building jar's package for each commit (sonatype)
- Add automate deploy on branch demo
- Add coveralls reports


demo/docker/README.md
```diff
@@ -1,15 +1,19 @@
 This is part of Nussknacker's [quickstart](https://touk.github.io/nussknacker/Quickstart.html).
 
-Building images
+Images
 =====
-* To build app image you can either:
-  * put nussknacker-ui-assembly.jar in app/build folder
-  * use released version (you can change it in docker-compose.yml file) 
+Docker images is build on each commit. Available nusskancker images you can find at: https://hub.docker.com/r/touk/nussknacker/tags.
```
arkadius: Should be "Docker images are published ..."

docs/Quickstart.md
```diff
@@ -96,11 +95,11 @@ The quickstart starts several Docker containers. Let's look at them in detail:
...
-* download `assemblu-code.jar` using `./downloadSampleAssembly.sh`
+* if you want change model download `assemblu-code.jar` using `./downloadSampleAssembly.sh`
 * rebuild docker image by `docker-compose build --no-cache app`
 
-##  Using own version
-If you have modified Nussknacker sources you have to rebuild docker image 
-by `./buildDockerImageFromSources.sh`. 
-If you have modified example module you have to regenerate `code-assembly.jar` 
-using `./buildSampleAssembly.sh`.
+<!--##  Using own version-->
```
arkadius: Why it is commented and why it is commented using xml-style comments?

.travis.yml
```diff
@@ -1,28 +1,65 @@
...
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/
 
-services:
-  - docker
+stages:
+  - tests
+  - name: releases
+    if: fork = false AND type != cron
```
arkadius: Why this fork and type other than cron checks exists? could you write down some comment about it here?

.travis.yml
```diff
@@ -1,28 +1,65 @@
...
+      script: sbt ++$TRAVIS_SCALA_VERSION clean test
+    - name: "Management tests"
+      script: sbt ++$TRAVIS_SCALA_VERSION clean management/it:test
+    - name: "EngineStandalone tests"
+      script: sbt ++$TRAVIS_SCALA_VERSION clean engineStandalone/it:test
+    - name: "UI slow tests"
+      script: sbt ++$TRAVIS_SCALA_VERSION clean assemblySamples ui/slow:test
+    - name: "Tests docker demo"
+      script: cd ./demo/docker && ./testQuickstart.sh
+    - stage: releases
```
arkadius: Shouldn't it be made parallel with tests? It has no dependency to previous stage. Maybe first stage should be "build" which will do parallel tests and real build.

.travis.yml
```diff
@@ -1,28 +1,65 @@
...
+      name: "Release sonatype packages"
+      script: ./ciBuild.sh "${BUILD_VERSION}-SNAPSHOT" ${NEXUS_PASSWORD} ${NEXUS_URL} ${NEXUS_USER} true
+    - name: "Push Docker Image"
```
arkadius: Shouldn't it use artifact from previous stage (build) instead of running parallel?

nussknacker-dist/src/universal/bin/wait-for-it.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
+#   Use this script to test if a given TCP host/port are available
```
arkadius: Can you give here a link to original file with some notice?

demo/docker/docker-compose-lite.yml
```diff
@@ -1,28 +1,24 @@
...
+    environment:
+      - NUSSKNACKER_CONFIG_FILE=/opt/nussknacker/conf/docker-compose-application.conf
     volumes:
-      - ./app/conf:/conf
-      - ${CODE_LOCATION}:/code-assembly.jar
       - sharedStorage:/opt/flinkData
+
   jobmanager:
-    build: ./flink
+    image: touk/flink:2.11-1.7.2
```
mproch: let's not hardcode versions here (and in other dockerfiles)

nussknacker-dist/src/universal/conf/docker-application.conf
```diff
@@ -0,0 +1,75 @@
+db {
```
mproch: I don't think this file should be in ourt dist package

nussknacker-dist/src/universal/bin/wait-for-it.sh
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env bash
```
mproch: I don't think this file should be in our dist

nussknacker-dist/src/universal/bin/docker-run.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
mproch: I don't think this file should be in our dist

build.sbt
```diff
@@ -127,21 +136,51 @@ val postgresV = "42.2.5"
...
 val confluentV = "4.1.2"
 
+lazy val dockerSettings = {
+  val workingDir = "/opt/nussknacker"
+
+  Seq(
+    dockerEntrypoint := Seq(s"$workingDir/bin/docker-run.sh", dockerPort.toString),
+    dockerExposedPorts := Seq(dockerPort),
+    dockerExposedVolumes := Seq(
+      "/opt/flinkData/savepoints",
```
mproch: this is good for demo, but not for production deployment...

build.sbt
```diff
@@ -127,21 +136,51 @@ val postgresV = "42.2.5"
...
+      s"$workingDir/logs",
+      s"$workingDir/db"
+    ),
+    defaultLinuxInstallLocation in Docker := workingDir,
+    dockerBaseImage := "openjdk:8-jdk",
+    dockerUsername := dockerUserName,
+    packageName := dockerPackageName,
+    dockerUpdateLatest := dockerUpLatest,
+    dockerLabels := Map(
```
mproch: scala & flink versions also should be in tags?

build.sbt
```diff
@@ -14,6 +16,13 @@ val nexusUrl = Option(System.getProperty("nexusUrl"))
 //TODO: this is pretty clunky, but works so far for our case...
 val nexusHost = nexusUrl.map(_.replaceAll("http[s]?://", "").replaceAll("[:/].*", ""))
 
+//Docker release configuration
```
mproch: all these should be really configurable?

demo/docker/docker-compose.yml
```diff
@@ -1,28 +1,24 @@
...
-    build:
-      context: ./app
-      args:
-        #TODO: consider sth like:
-        #curl -s https://api.github.com/repos/touk/nussknacker/releases/latest | grep tag_name | awk '{ print $2 }' | sed -e 's/"//g' -e 's/,//g'
-        #to retrieve latest version
-        version: ${NUSSKNACKER_VERSION}
+    image: ${NUSSKNACKER_IMAGE}:${NUSSKNACKER_VERSION}
+    entrypoint: ["/opt/nussknacker/bin/wait-for-it.sh"]
+    command: ["-h", "taskmanager", "-p", "6126", "-t", "60", "--", "/opt/nussknacker/bin/docker-run.sh"]
```
mproch: let's fix NK if it fails on waiting for taskmanager, but we should rely on tm being available...
besides - we should be waiting for jobmanager, not task manager?

nussknacker-dist/src/universal/conf/docker-compose-application.conf
```diff
@@ -1,5 +1,5 @@
 db {
```
mproch: this file certainly shouldn't be in our dist package

.travis.yml
```diff
@@ -1,28 +1,65 @@
...
-sudo: required
+# These directories are cached to S3 at the end of the build
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/
 
-services:
-  - docker
+stages:
```
arkadius: Please add here a TODO comment that we should use artifacts deployed on S3 between steps (https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs) or we should move to circle-ci.

lciolecki: Ok.

.travis.yml
lciolecki: This is a secure case. We don't want releases from third parties pull requests . I disabled alswo cron type, because I think this shouldbe use only for test - not deploy.

build.sbt
lciolecki: :) Version of application can be different than docker tag. I mean that docker-latest docker tag can store many of versions.

build.sbt
lciolecki: Yep, it's easier to run script with docker tag and version of application (automate configurable by travis) than change files with these information.

.travis.yml
arkadius: Regards to first part: what if someone will remove this line in their fork?
Regards to second part: how cron type is related to deploy - I don't get it. Maybe some comment would make this dependency more explicit?

nussknacker-dist/src/universal/conf/docker-application.conf
lciolecki: This is configuration for standalone version of docker.

nussknacker-dist/src/universal/bin/docker-run.sh
lciolecki: Now this is nussknacker-entrypoint and this file must be in our repo - it's an antry for docker.

build.sbt
lciolecki: I'v added a scala version, we can think about flink version.

demo/docker/docker-compose.yml
lciolecki: When we set `queryableStateProxyUrl: "taskmanager:9069"` and it's not avaiable then application throws exception and doesn't work.

nussknacker-dist/src/universal/conf/docker-compose-application.conf
lciolecki: I'v moved it to demo/docker/app

PR #40, opened 2019-05-22
Fix disabled subprocess compilation

bottaio:
Revert previous behavior of subprocess compilation when node is disabled as we don't have type information in this case.


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
```diff
@@ -148,7 +148,21 @@ class PartSubGraphCompiler(protected val classLoader: ClassLoader,
         fa = compile(next, ctx))(
         f = compiledNext => compiledgraph.node.CustomNode(id, compiledNext))
 
-    case subprocessInput@SubprocessInput(id, ref, _, _, subParams) =>
+    // subprocessParams is None for disabled node
+    case SubprocessInput(id, ref, _, Some(true), _) =>
```
mproch: Do we want to refactor it a bit in the future? If so, maybe let's add some comment here?

PR #41, opened 2019-05-22
Feature: frontend libraries update

lciolecki:
TODO:
- react-bootstrap
- react-redux
- react-router
- jointjs


ui/client/package.json
```diff
@@ -11,78 +11,81 @@
     "test": "NODE_ENV=production ./node_modules/karma/bin/karma start --single-run"
   },
   "dependencies": {
-    "bootstrap": "3.3.6",
-    "classnames": "2.2.5",
-    "file-saver": "1.3.3",
-    "flattenizer": "1.0.4",
+    "bootstrap": "^3.4.1",
```
fijolekProjects: Please use explicit versioning, i.e "3.4.1" instead of "^3.4.1" (which is non-deterministic over time)(https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json)

lciolecki: I can agree with that we shouldn't use ^ but ~ it's okey. I will replace ^ to ~.

mproch:
Great job!
some minor things that don't work
- counts fail with bad date format
- ladda button display is awkward (animation is moved)
- headers in main process table does not display process captions, column width is not correct

ui/client/containers/Processes.js
```diff
@@ -102,67 +102,68 @@ class Processes extends PeriodicallyReloadingComponent {
...
-            <Th column="status" className="status-column">Status</Th>
-            <Th column="edit" className="edit-column">Edit</Th>
-            <Th column="metrics" className="metrics-column">Metrics</Th>
-          </Thead>
-
           {this.state.processes.map((process, index) => {
             return (
               <Tr className="row-hover" key={index}>
-                <Td column="name" value={process.name}>
+                <Td column="Process name" value={process.name}>
```
mproch: Let's pass column labels as 
columns =       {key: 'name', label: 'Name'},
      {key: 'category', label: 'Action'}, ...
    ]
and then <Table columns={columns}>

This still works, and we avoid passing labels as keys, which is kind of weird ;)

lciolecki: I was looking for this option ... :) Thanks!

PR #42, opened 2019-05-23
Do not validate disabled nodes

nadberezny:


jedrz:
LGTM though someone else should approve too.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessRewriter.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  def removeAndBypassNode(process: DisplayableProcess, node: NodeData): DisplayableProcess =
+    new RemoveAndBypassNode(process, node).apply
+
+  def removeBranch(process: DisplayableProcess, branchRoot: NodeData): DisplayableProcess =
+    new RemoveBranch(process, branchRoot).apply
+
+  def removeDisabledNodes(process: DisplayableProcess): DisplayableProcess =
+    process.nodes.collectFirst { case d: Disableable => d } match {
+      case Some(node) if node.isDisabled.getOrElse(false) =>
```
arkadius: 1. Why not to merge both checks: `collectFirst { case d: Disableable if d.isDisabled ...`
2. Can you replace this with `contains(true)`? It is IMO less brain-damaging.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveAndBypassNode.scala
```diff
@@ -0,0 +1,68 @@
...
+import pl.touk.nussknacker.restmodel.displayedgraph.{DisplayableProcess, ProcessRewriter}
+import pl.touk.nussknacker.restmodel.displayedgraph.displayablenode.Edge
+import pl.touk.nussknacker.restmodel.displayedgraph.displayablenode.EdgeType.FilterFalse
+
+
+object RemoveAndBypassNode {
+  private case class NodeEdges(incoming: List[Edge], outgoing: List[Edge])
+}
+
+class RemoveAndBypassNode(process: DisplayableProcess, node: NodeData) {
```
arkadius: I think that moving `process` to arguments of `apply` will better show how it is processed (where it is used). I see that `node` is really used "globally" by this class so it is in correct place. Also we can think about renaming `apply` to more expressive name e.g. `RemoveAndBypassNode(node).in(process)`. WDYT?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveAndBypassNode.scala
```diff
@@ -0,0 +1,68 @@
...
+          }
+
+        process.copy(
+          nodes = process.nodes.filterNot(_ == node),
+          edges = newEdges)
+    }
+  }
+
+  private def getBranchRootToRemove(nodeEdges: NodeEdges): Option[NodeData] =
+    nodeEdges.outgoing.collectFirst {
```
arkadius: I'm not getting it. Why we checking only `FilterFalse`?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveAndBypassNode.scala
```diff
@@ -0,0 +1,68 @@
...
+            (nodeEdges.incoming ++ nodeEdges.outgoing).contains(disabledEdge)
+          }
+
+        process.copy(
+          nodes = process.nodes.filterNot(_ == node),
+          edges = newEdges)
+    }
+  }
+
+  private def getBranchRootToRemove(nodeEdges: NodeEdges): Option[NodeData] =
```
arkadius: What do you think about skipping "get" prefixes in methods in this class? We rather sticking with scala-way approach - without those prefixes... Or replace with `find` prefix.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveBranch.scala
```diff
@@ -0,0 +1,39 @@
...
+    val nodeIdsToRemove =
+      branchRoot.id :: edgesToRemove.map(_.to)
+
+    process.copy(
+      edges = process.edges.filterNot(edgesToRemove.contains(_)),
+      nodes = process.nodes.filterNot { n => nodeIdsToRemove.contains(n.id) }
+    )
+  }
+
+  private def incomingEdges: List[Edge] = process.edges.filter(_.to == branchRoot.id)
```
arkadius: Shouldn't be lazy val?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveBranch.scala
```diff
@@ -0,0 +1,39 @@
+package pl.touk.nussknacker.restmodel.displayedgraph.rewriter
+
+import pl.touk.nussknacker.engine.graph.node.NodeData
+import pl.touk.nussknacker.restmodel.displayedgraph.DisplayableProcess
+import pl.touk.nussknacker.restmodel.displayedgraph.displayablenode.Edge
+
+class RemoveBranch(process: DisplayableProcess, branchRoot: NodeData) {
+
+  def apply: DisplayableProcess = {
+    val edgesToRemove = outgoingEdges
```
arkadius: why not just `incomingEdges ++ outgoingEdges.flatMap(subsequentEdges)`?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/rewriter/RemoveBranch.scala
```diff
@@ -0,0 +1,39 @@
+package pl.touk.nussknacker.restmodel.displayedgraph.rewriter
+
+import pl.touk.nussknacker.engine.graph.node.NodeData
+import pl.touk.nussknacker.restmodel.displayedgraph.DisplayableProcess
+import pl.touk.nussknacker.restmodel.displayedgraph.displayablenode.Edge
+
+class RemoveBranch(process: DisplayableProcess, branchRoot: NodeData) {
```
arkadius: The same comment to process field as in `RemoveAndBypassNode`. Without this it is hard to say how many times you are iterating through the edges/nodes of process.

ui/server/src/test/scala/pl/touk/nussknacker/ui/validation/ProcessValidationSpec.scala
```diff
@@ -248,6 +249,28 @@ class ProcessValidationSpec extends FunSuite with Matchers {
     validationResult.saveAllowed shouldBe true
   }
 
+  test("invalid disabled nodes don't invalidate process") {
```
arkadius: Can you provide more tests for scenarios with more complex connections like `Switch`, `Split` and `Joins` to prove that this recursion really works?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessRewriter.scala
```diff
@@ -0,0 +1,21 @@
+package pl.touk.nussknacker.restmodel.displayedgraph
+
+import pl.touk.nussknacker.engine.graph.node.{Disableable, NodeData}
+import pl.touk.nussknacker.restmodel.displayedgraph.rewriter.{RemoveAndBypassNode, RemoveBranch}
+
+object ProcessRewriter {
+
+  def removeAndBypassNode(process: DisplayableProcess, node: NodeData): DisplayableProcess =
```
arkadius: Why `removeAndBypassNode` and `removeBranch` are here? They are used someway? `RemoveAndBypassNode` and `RemoveBranch` are rather implementation details so IMHO they should be package private protected and here should be only one method.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessRewriter.scala
```diff
@@ -0,0 +1,21 @@
+package pl.touk.nussknacker.restmodel.displayedgraph
```
arkadius: I think that all three classes could be in in separate `rewrite` package witch one public class and method. WDYT?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessRewriter.scala
```diff
@@ -0,0 +1,21 @@
...
+
+object ProcessRewriter {
+
+  def removeAndBypassNode(process: DisplayableProcess, node: NodeData): DisplayableProcess =
+    new RemoveAndBypassNode(process, node).apply
+
+  def removeBranch(process: DisplayableProcess, branchRoot: NodeData): DisplayableProcess =
+    new RemoveBranch(process, branchRoot).apply
+
+  def removeDisabledNodes(process: DisplayableProcess): DisplayableProcess =
```
arkadius: How it is related with checking of `isDisabled` in `FlinkProcessRegistrar`, `PartSubGraphCompiler` and `ProcessValidation`? Is this flag in production code will be used only on frontend side (`DisplayableProcess`) now?

nadberezny:
@arkadius 
Thanks fot the great review :1st_place_medal: 
As you've suggested in our talk, it would be better to rewrite CanonicalProcess instead of DisplayableProcess. I'll prepare another PR.

PR #44, opened 2019-05-23
Configuring queryable state from Flink configuration only

bottaio:
There is a misleading configuration option `flinkConfig.queryableStateProxyUrl` that is now being used for determining cluster location for querying jobs' state. It is also invalid to use it as different processing types might refer to different cluster locations. This change removes this legacy configuration and uses already defined, but not used `processTypes.[...].queryableStateProxyUrl`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -584,7 +584,7 @@ lazy val queryableState = (project in engine("queryableState")).
         "com.typesafe" % "config" % configV
       )
     }
-  ).dependsOn(api)
+  ).dependsOn(api, interpreter)
 
 
 
--- a/engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/management/StandaloneProcessManager.scala
+++ b/engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/management/StandaloneProcessManager.scala
@@ -22,6 +22,7 @@ import pl.touk.nussknacker.engine.definition._
 import pl.touk.nussknacker.engine.graph.EspProcess
 import pl.touk.nussknacker.engine.graph.node.Source
 import pl.touk.nussknacker.engine.marshall.ProcessMarshaller
+import pl.touk.nussknacker.engine.queryablestate.QueryableClient
 import pl.touk.nussknacker.engine.standalone.StandaloneProcessInterpreter
 import pl.touk.nussknacker.engine.standalone.api.DeploymentData
 import pl.touk.nussknacker.engine.standalone.api.types._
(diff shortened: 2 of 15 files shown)


build.sbt
```diff
@@ -328,7 +328,7 @@ lazy val interpreter = (project in engine("interpreter")).
     buildInfoPackage := "pl.touk.nussknacker.engine.version",
     buildInfoOptions ++= Seq(BuildInfoOption.ToMap)
   ).
-  dependsOn(util)
+  dependsOn(util, queryableState)
```
mproch: this adds Flink to interpreter dependencies - let's try to avoid this. Maybe extract trait QueryableStateClient to interpreter (or somewhere...) or detec appropriate ProcessManagerProviders by checking if they implement some additional trait?

engine/queryableState/src/main/scala/pl/touk/nussknacker/engine/flink/queryablestate/QueryableClient.scala
```diff
@@ -0,0 +1,12 @@
+package pl.touk.nussknacker.engine.flink.queryablestate
+
+import org.apache.flink.api.common.typeinfo.TypeInformation
```
mproch: let's make this trait Flink free - only with fetchJson state?

demo/docker/app/conf/application.conf
```diff
@@ -27,7 +27,6 @@ customProcesses {}
 
 flinkConfig {
   restUrl: "http://jobmanager:8081"
-  queryableStateProxyUrl: "taskmanager:9069"
```
mproch: Let's leave it here. This config uses older configuration format - we'll change it hopefully in near future...

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
```diff
@@ -24,13 +26,19 @@ class SettingsResources(config: FeatureTogglesConfig)(implicit ec: ExecutionCont
...
-            signals = config.queryableStateProxyUrl.isDefined,
+            signals = signalsSupported,
             attachments = config.attachments.isDefined
           )
           UISettings(toggleOptions)
         }
       }
     }
+
+  private val signalsSupported: Boolean = {
```
mproch: is it really the same? currently it is...

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/management/StandaloneProcessManager.scala
```diff
@@ -200,6 +201,8 @@ class StandaloneProcessManagerProvider extends ProcessManagerProvider {
   override def createProcessManager(modelData: ModelData, config: Config): ProcessManager
     = StandaloneProcessManager(modelData, config)
 
+  override def createQueryableClient(config: Config): Option[QueryableClient] = None
```
mproch: maybe we should also drop unused flags below? supportsSignals/QueryableState? Or let's use them

PR #53, opened 2019-06-04
Fixes bugs and add upates part I 

lciolecki:
Fix bugs: 
- search subprocesses, 
- run service
- missing arrows. 

Updates:
- Add docker-compose tests for flink, kibana, grafana

### DIFF
--- a/ui/client/common/JsonUtils.js
+++ b/ui/client/common/JsonUtils.js
@@ -5,9 +5,7 @@ import Flattenizer from "flattenizer";
 // They should be removed as soon as friendly UI components for complex input objects are ready
 
 export function tryStringify(input) {
-  return _.isString(input) ? input :
-    _.isObject(input) ? JSON.stringify(input) :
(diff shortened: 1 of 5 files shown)


ui/client/common/JsonUtils.js
```diff
@@ -5,9 +5,7 @@ import Flattenizer from "flattenizer";
 // They should be removed as soon as friendly UI components for complex input objects are ready
 
 export function tryStringify(input) {
-  return _.isString(input) ? input :
-    _.isObject(input) ? JSON.stringify(input) :
-      input;
+  return _.isString(input) ? input : _.isObject(input) ? JSON.stringify(input) : input;
```
mproch: if we change this, maybe let's add some parentheses... I'm still not sure if I can parse it correctly :)

lciolecki: this is only text formatter.. :)

PR #54, opened 2019-06-06
FIX: mixed eager/lazy params list now work correctly

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ProcessObjectFactory.scala
```diff
@@ -32,7 +32,7 @@ class ProcessObjectFactory(expressionEvaluator: ExpressionEvaluator) extends Laz
     val withDefs = params.sortBy(_.name).zip(objectWithMethodDef.parameters.sortBy(_.name))
 
     // TODO JOIN: Handle not lazy evaluated branch params
-    val (lazyInterpreterParameters, paramsToEvaluate) = withDefs.span(p => p._2.isLazyParameter || p._2.branchParam)
+    val (lazyInterpreterParameters, paramsToEvaluate) = withDefs.partition(p => p._2.isLazyParameter || p._2.branchParam)
```
bottaio: ouch :)

arkadius: ouch++ ;-)

PR #60, opened 2019-06-18
Properties variable

bottaio:
This PR introduces properties variable - allows business users to access process properties in diagram.

Additionally, I've decided to remove UserDefinedProcessAdditionalFields trait as it is source of many argonaut-related problems.

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/build/EspProcessBuilder.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/build/EspProcessBuilder.scala
@@ -2,7 +2,7 @@ package pl.touk.nussknacker.engine.build
 
 import cats.data.NonEmptyList
 import pl.touk.nussknacker.engine.api.process.ProcessName
-import pl.touk.nussknacker.engine.api.{MetaData, StandaloneMetaData, StreamMetaData, UserDefinedProcessAdditionalFields}
+import pl.touk.nussknacker.engine.api.{Group, MetaData, ProcessAdditionalFields, StandaloneMetaData, StreamMetaData}
 import pl.touk.nussknacker.engine.build.GraphBuilder.Creator
 import pl.touk.nussknacker.engine.graph.exceptionhandler.ExceptionHandlerRef
 import pl.touk.nussknacker.engine.graph.expression.Expression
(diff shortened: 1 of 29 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -96,8 +97,13 @@ protected trait ProcessCompilerBase {
 
   private val expressionCompiler = ExpressionCompiler.withoutOptimization(classLoader, expressionConfig)
 
-  private def contextWithOnlyGlobalVariables = ValidationContext(Map.empty,
-    expressionConfig.globalVariables.mapValuesNow(_.returnType) + (Interpreter.MetaParamName -> Typed[MetaVariables]))
+  private def contextWithOnlyGlobalVariables(implicit metaData: MetaData): ValidationContext = {
+    val globalTypes = expressionConfig.globalVariables.mapValuesNow(_.returnType)
+    val typesWithMeta = MetaVariables.withType(globalTypes)
+    val typesWithProperties = PropertiesVariable.withType(typesWithMeta)
```
arkadius: IMO would be better to have only one variable meta with nested properties (maybe extracted?). It is simple to say to user, that he/she has in context: global variables, variables created in process and "meta" property and he/she will check what is available in the variable by typing. If you put more than one additional variable to context, than you will need to maintain documentation for all available variables in context.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/build/EspProcessBuilder.scala
```diff
@@ -21,8 +21,8 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
   def subprocessVersions(subprocessVersions: Map[String, Long]) =
     new ProcessMetaDataBuilder(metaData.copy(subprocessVersions = subprocessVersions))
 
-  def additionalFields(userDefinedProcessAdditionalFields: UserDefinedProcessAdditionalFields) =
-    new ProcessMetaDataBuilder(metaData.copy(additionalFields = Some(userDefinedProcessAdditionalFields)))
+  def additionalFields(processAdditionalFields: ProcessAdditionalFields) =
```
arkadius: I know that it already has been done this way, but maybe better approach would be pass here all fields of `ProcessAdditionalFields` instead of `ProcessAdditionalFields` object. Now it doesn't look too much DSL-y. What do you think about that?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
bottaio: good point, I've nested `properties` under `meta`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/build/EspProcessBuilder.scala
bottaio: with default parameter values it seems reasonable to me

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/marshall/UiProcessMarshallerSpec.scala
```diff
@@ -12,7 +12,7 @@ class UiProcessMarshallerSpec extends FlatSpec with Matchers {
   val processWithAdditionalFields =
     s"""
        |{
-       |    "metaData" : { "id": "custom", "typeSpecificData": {"type": "StreamMetaData", "parallelism" : 2}, "additionalFields": { "description": "$someProcessDescription"} },
+       |    "metaData" : { "id": "custom", "typeSpecificData": {"type": "StreamMetaData", "parallelism" : 2}, "additionalFields": {"description": "$someProcessDescription", "groups": [], "properties": {}}},
```
jedrz: I think it should be left as it was before. Am I right?

bottaio: You are right. One test depends on previous structure. But at the same time, this is the marshalling logic I would expect. Thus, I've changed data for `marshall and unmarshall process` test.

jedrz:
LGTM.

PR #66, opened 2019-06-26
Can pass global buildInfo from config in app/buildInfo

mproch:
Can be useful e.g. for some healthcheck requirements


mproch:
@arkadius I added processingType field

PR #70, opened 2019-07-05
Add healthcheck endpoint in standalone app

jedrz:
Simple healthcheck endpoint to ensure standalone application is running.


engine/standalone/app/src/test/scala/pl/touk/nussknacker/engine/standalone/http/StandaloneHttpAppSpec.scala
```diff
@@ -271,6 +271,13 @@ class StandaloneHttpAppSpec extends FlatSpec with Matchers with ScalatestRouteTe
     }
   }
 
+  it should "return health" in {
+    Get("/healthCheck") ~> managementRoute ~> check {
+      println(status)
```
mproch: let's remove println :)

jedrz: Whoops, removed. :flushed:

PR #74, opened 2019-07-09
Fetch processes details by process names batch

bottaio:
Ultimately, I'd like to get rid of processesDetails endpoints altogether. This change mitigates problem with exposing all processes' jsons in one request.

Additionally, it adds batching support for migration testing

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -90,6 +90,13 @@ class ProcessesResources(val processRepository: FetchingProcessRepository,
             }
           }
         } ~ path("processesDetails") {
+          post {
+            entity(as[List[String]]) { namesToFetch =>
+              complete {
+                validateAll(processRepository.fetchProcessesDetails(namesToFetch.map(ProcessName(_))))
+              }
+            }
+          } ~
           get {
             complete {
               validateAll(processRepository.fetchProcessesDetails())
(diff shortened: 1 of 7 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -182,21 +178,27 @@ trait StandardRemoteEnvironment extends Argonaut62Support with RemoteEnvironment
...
+        HttpMethods.POST,
+        "processesDetails" :: Nil,
+        requestEntity = HttpEntity(ContentTypes.`application/json`, processes.map(_.name).asJson.nospaces))
+    }
+
+    EitherT(Future.successful(
+      basicProcesses.grouped(config.batchSize)
+        .foldLeft(List.empty[ValidatedProcessDetails].asRight[EspError]) { case (acc, batch) => for {
+            current <- acc
+            fetched <- Await.result(getMany(batch), config.batchTimeout)
```
mproch: can we do it without await?
If we don't want to sequence because of thread pool limitations, maybe we can do flatmap instead?
Then timeout wouldn't be needed?

bottaio: good point, fixed that

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -90,6 +90,13 @@ class ProcessesResources(val processRepository: FetchingProcessRepository,
             }
           }
         } ~ path("processesDetails") {
+          post {
+            entity(as[List[String]]) { namesToFetch =>
```
mproch: I wonder if it would be more "RESTful" if we did it with GET and passed ids as query string?

PR #76, opened 2019-07-10
[FIX] - center text in counts vertically

mproch:


ui/client/components/graph/EspNode.js
```diff
@@ -232,7 +232,9 @@ function getTestResultsSummaryAttr(processCounts, width) {
     text: countsContent,
     style: { 'font-size': summaryCountFontSize },
     fill: hasErrors ? 'red' : 'white',
-    refX: width + rectHeight/2
+    refX: width + rectHeight/2,
```
lciolecki: This realy resloves problem..?

PR #79, opened 2019-07-12
processesDetails GET endpoint

bottaio:
Making processesDetails more RESTful by changing POST endpoint into GET.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -19,7 +19,6 @@ import pl.touk.nussknacker.ui.process.repository.ProcessRepository._
 import pl.touk.nussknacker.ui.util._
 import pl.touk.nussknacker.ui._
 import EspErrorToHttp._
-import akka.http.scaladsl.unmarshalling.PredefinedFromStringUnmarshallers.CsvSeq
 import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.ui.codec.UiCodecs
 import pl.touk.nussknacker.ui.validation.{FatalValidationError, ProcessValidation}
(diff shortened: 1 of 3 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -90,14 +92,12 @@ class ProcessesResources(val processRepository: FetchingProcessRepository,
             }
           }
         } ~ path("processesDetails") {
```
lciolecki: Mayby better way is do GET /processes?names=[comma separeted names]? I think we should use only one endpoint with filter queries to fetching list of processes,

bottaio: /process endpoint serves different data than /processesDetails which returns more verbose representation of processes

lciolecki: Okey! First: better name for this endpoint is /processes-datails/. Second: where are tests? :)

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/ProcessesResourcesSpec.scala
```diff
@@ -42,21 +42,22 @@ class ProcessesResourcesSpec extends FunSuite with ScalatestRouteTest with Match
...
-  val routWithAdminPermission = withPermissions(processesRoute, testPermissionAdmin)
+  val routeWithAllPermissions = withAllPermissions(processesRoute)
+  val routeWithAdminPermission = withPermissions(processesRoute, testPermissionAdmin)
   val processActivityRouteWithAllPermission = withAllPermissions(processActivityRoute)
   implicit val loggedUser = LoggedUser("lu",  testCategory)
 
   private val processName = ProcessName(SampleProcess.process.id)
 
   test("return list of process") {
     saveProcess(processName, ProcessTestData.validProcess) {
```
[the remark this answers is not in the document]
bottaio: I'll fix these tests in another PR as for now it blocks further development and I'm short on time

PR #80, opened 2019-07-12
Copy and paste arbitrary nodes between processes

jedrz:
This PR allow to copy and paste single node between processes with <kbd>Ctrl+C</kbd> and <kbd>Ctrl+V</kbd>. For now groups cannot be copied and pasted but it seems doable.

At the beginning I wanted to use [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) and add buttons to copy and paste in the right panel but the API is not widely supported, especially by Firefox 67.

Below is a simple demo of this feature:
copy-paste-single-node

### DIFF
--- a/ui/client/common/JsonUtils.js
+++ b/ui/client/common/JsonUtils.js
@@ -16,6 +16,14 @@ export function tryParse(input){
   }
 }
 
+export function tryParseOrNull(input) {
+  try {
+    return JSON.parse(input)
+  } catch (e) {
+    return null
+  }
+}
+
 //Flattenizer is nice, but unfortunately its array notation is different than Lodash's,
 // so we use regex to make it consistent with Lodash'a array notation here
 export function flattenObj(obj) {
--- a/ui/client/components/graph/Graph.js
+++ b/ui/client/components/graph/Graph.js
@@ -81,8 +81,16 @@ class Graph extends React.Component {
 		this.processGraphPaper.setDimensions(area.offsetWidth, area.offsetHeight)
 	}
 
+	canCopyNode() {
+		return !NodeUtils.isNotPlainNode(this.props.nodeToDisplay) && this.props.allModalsClosed
+	}
+
+	canCutNode() {
+		return this.canCopyNode() && this.props.capabilities.write
+	}
+
 	copyNode() {
-		if (!NodeUtils.isNotPlainNode(this.props.nodeToDisplay) && this.props.allModalsClosed) {
+		if (this.canCopyNode()) {
 			navigator.clipboard.writeText(JSON.stringify(this.props.nodeToDisplay))
 		}
 	}
@@ -98,7 +106,7 @@ class Graph extends React.Component {
 	}
 
 	cutNode() {
-		if (!NodeUtils.isNotPlainNode(this.props.nodeToDisplay)) {
+		if (this.canCutNode() ) {
 			this.copyNode()
 			this.props.actions.deleteNode(this.props.nodeToDisplay.id)
 		}
--- a/ui/client/components/graph/Graph.js
+++ b/ui/client/components/graph/Graph.js
@@ -18,6 +18,7 @@ import cssVariables from "../../stylesheets/_variables.styl"
 import * as GraphUtils from "./GraphUtils";
 import * as JointJsGraphUtils from "./JointJsGraphUtils";
 import PropTypes from 'prop-types';
+import * as JsonUtils from "../../common/JsonUtils";
 
 class Graph extends React.Component {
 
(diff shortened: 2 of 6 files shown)


ui/client/common/JsonUtils.js
```diff
@@ -16,6 +16,14 @@ export function tryParse(input){
   }
 }
 
+export function tryParseOrNull(input) {
```
lciolecki: Mayby better will be parseOrNull?

ui/client/components/graph/NodeUtils.js
```diff
@@ -13,6 +17,10 @@ class NodeUtils {
     return type === "Properties";
   }
 
+  noChosenNode = (node) => {
```
lciolecki: Good idea! :+1:

ui/client/components/graph/NodeUtils.js
```diff
@@ -4,6 +4,10 @@ import ProcessUtils from '../../common/ProcessUtils.js'
 
 class NodeUtils {
 
+  isNode = (obj) => {
+    return !_.isEmpty(obj) && 'id' in obj && 'type' in obj
```
lciolecki: You can use there _.has from lodash.

ui/client/containers/Visualization.js
```diff
@@ -167,7 +167,8 @@ class Visualization extends React.Component {
         />
 
         <SpinnerWrapper isReady={!graphNotReady}>
-          <Graph ref={this.graphRef} />
+          <Graph ref={this.graphRef}
```
lciolecki: Write this in oneline please :)

ui/client/components/graph/Graph.js
```diff
@@ -72,8 +78,30 @@ class Graph extends React.Component {
 		this.processGraphPaper.setDimensions(area.offsetWidth, area.offsetHeight)
 	}
 
+	copyNode(event) {
+		if (NodeUtils.noChosenNode(this.props.nodeToDisplay) || !this.props.allModalsClosed) {
+			return
```
lciolecki: Hmm, I don't like empty return.. Mayby we can do this in better way?

ui/client/common/JsonUtils.js
jedrz: It is consistent with the above function `tryParse`.

ui/client/components/graph/Graph.js
jedrz: I like short returns but I can put writing to clipboard under the if statement since nothing else happens in this function.

ui/client/components/graph/NodeUtils.js
jedrz: Yup, I know. Can we leave it as it is or should I switch to lodash helper?

ui/client/containers/Visualization.js
jedrz: Done.

ui/client/common/JsonUtils.js
lciolecki: But in your function you parse or return null - not trying :)

ui/client/components/graph/NodeUtils.js
lciolecki: I think it will be looks better with lodash.

ui/client/components/graph/Graph.js
jedrz: Done.

ui/client/components/graph/NodeUtils.js
jedrz: Done.

ui/client/components/graph/NodeUtils.js
mproch: isNotPlainNode?

ui/client/components/graph/Graph.js
```diff
@@ -72,8 +78,29 @@ class Graph extends React.Component {
...
+		}
+	}
+
+	pasteNode(event) {
+		if (!this.props.allModalsClosed) {
+			return
+		}
+		const clipboardText = (event.clipboardData || window.clipboardData).getData('text');
+		const node = JsonUtils.tryParseOrNull(clipboardText)
+		if (!this.props.capabilities.write || !NodeUtils.isNode(node) || NodeUtils.nodeIsGroup(node)) {
```
mproch: maybe move this.addNode into if - so there are fewer returns?
I also wonder if it's possible/easy to generate new id if the current one is already used - e.g. when copy/pasting to same diagram?

jedrz: Done. You're right, for now new id is always generated even if it's unique.

ui/client/components/graph/NodeUtils.js
jedrz: Done.

PR #82, opened 2019-07-15
Workaround copying to clipboard for content delivered via HTTP

jedrz:
`navigator.clipboard` is not accessible via HTTP. As a workaround in order to copy node's JSON, a hidden element is created, then text content is written to it and finally, second copy is executed. Because of copying twice, two events get generated so we need to discard artificial one.

### DIFF
--- a/ui/client/common/ClipboardUtils.js
+++ b/ui/client/common/ClipboardUtils.js
@@ -1,0 +1,18 @@
+const readText = event => (event.clipboardData || window.clipboardData).getData('text')
+
+const writeText = (text, elementId) => {
+  const el = document.createElement('textarea');
+  el.value = text;
+  el.setAttribute("id", elementId)
+  el.setAttribute('readonly', '');
+  el.style = {position: 'absolute', left: '-9999px'};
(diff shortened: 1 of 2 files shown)


lciolecki:
> `navigator.clipboard` is not accessible via HTTP. As a workaround in order to copy node's JSON, a hidden element is created, then text content is written to it and finally, second copy is executed. Because of copying twice, two events get generated so we need to discard artificial one.

lciolecki:
Please improve formatting ;)

jedrz:
@lciolecki You are right but I don't want to untabify the file because all the history would be lost.

ui/client/common/ClipboardUtils.js
```diff
@@ -0,0 +1,18 @@
+const readText = event => (event.clipboardData || window.clipboardData).getData('text')
+
+const writeText = (text, elementId) => {
```
mproch: maybe some comment why this is needed and reference where does it come from?

jedrz: Done.

PR #85, opened 2019-07-17
Select and operate on multiple nodes

jedrz:
* Select nodes with <kbd>Ctrl + click</kbd>.
* Copy, cut, paste, delete, duplicate selected nodes.
* Undo/redo paste in single step.
* I have discovered when deleting a group but I think it's a deprecated feature.

In next PR I am going to implement simple validation to ensure pasted nodes are available to choose from the left toolbox.

Short demo:
multiple-nodes

### DIFF
--- a/ui/client/components/graph/Graph.js
+++ b/ui/client/components/graph/Graph.js
@@ -23,623 +23,623 @@ import ClipboardUtils from "../../common/ClipboardUtils";
 
 class Graph extends React.Component {
 
-	redrawing = false
+  redrawing = false
 
-	static propTypes = {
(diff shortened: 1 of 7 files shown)


ui/client/components/graph/Graph.js
```diff
@@ -43,18 +43,19 @@ class Graph extends React.Component {
 		})
 
 		this.espGraphRef = React.createRef()
+
+		this.windowListeners = {
```
lciolecki: Code formatting ;)

jedrz: Done. Are you sure we should leave new commit? :slightly_smiling_face:

PR #89, opened 2019-07-19
Add only nodes available from toolbox

jedrz:
Unknown nodes or available in other categories cannot be added to process.

### DIFF
--- a/ui/client/common/ProcessDefinitionUtils.js
+++ b/ui/client/common/ProcessDefinitionUtils.js
@@ -1,0 +1,23 @@
+import _ from 'lodash';
+
+const getPossibleNodesInCategory = (possibleNodes, category) =>
+  possibleNodes.filter(node => node.categories.includes(category))
+
+const getNodesToAddInCategory = (processDefinitionData, category) => {
+  return (processDefinitionData.nodesToAdd || []).map(group => {
+    return {
+      ...group,
+      possibleNodes: getPossibleNodesInCategory(group.possibleNodes, category)
+    }
+  })
+}
+
+const getFlatNodesToAddInCategory = (processDefinitionData, category) => {
+  const nodesToAdd = getNodesToAddInCategory(processDefinitionData, category)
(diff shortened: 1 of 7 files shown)


ui/client/common/ProcessDefinitionUtils.js
```diff
@@ -0,0 +1,23 @@
...
+    return {
+      ...group,
+      possibleNodes: getPossibleNodesInCategory(group.possibleNodes, category)
+    }
+  })
+}
+
+const getFlatNodesToAddInCategory = (processDefinitionData, category) => {
+  const nodesToAdd = getNodesToAddInCategory(processDefinitionData, category)
+  return _.flatMap(nodesToAdd, group => getPossibleNodesInCategory(group.possibleNodes, category))
```
mproch: you don't need getPossibleNodesInCategory here? They are already filtered?

ui/client/test/ProcessDefinitionUtils-test.js
```diff
@@ -0,0 +1,145 @@
...
+  it("should return all nodes available in category", () => {
+    const nodesToAdd = ProcessDefinitionUtils.getNodesToAddInCategory(processDefinition, "Category1")
+
+    expect(nodesToAdd).toEqual(processDefinition.nodesToAdd)
+  })
+
+  it("should filter out unavailable nodes in category", () => {
+    const nodesToAdd = ProcessDefinitionUtils.getNodesToAddInCategory(processDefinition, "Category2")
+
+    expect(nodesToAdd).toEqual([
```
mproch: can we make this assertion a bit more understandable? ;)

ui/client/common/ProcessDefinitionUtils.js
jedrz: Yup, you're right.

ui/client/test/ProcessDefinitionUtils-test.js
jedrz: Done.

PR #90, opened 2019-07-22
[FIX] Deployment of process with disabled subprocess.

nadberezny:


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -175,7 +176,9 @@ class ManagementActor(environment: String, managers: Map[ProcessingType, Process
   }
 
   private def resolveGraph(canonicalJson: String): Future[String] = {
-    val validatedGraph = UiProcessMarshaller.fromJson(canonicalJson).toValidatedNel
+    val validatedGraph = UiProcessMarshaller.fromJson(canonicalJson)
+      .andThen(canonical => Valid(canonical.withoutDisabledNodes))
```
mproch: andThen(Valid) == map?

nadberezny: :+1: 
right, flatMap is unnecessary here.

PR #91, opened 2019-07-22
Custom Json marshaller, Jackson implementation

mproch:


engine/standalone/app/src/test/scala/pl/touk/nussknacker/engine/standalone/http/StandaloneHttpAppSpec.scala
```diff
@@ -31,6 +32,8 @@ class StandaloneHttpAppSpec extends FlatSpec with Matchers with ScalatestRouteTe
   import argonaut.ArgonautShapeless._
   import spel.Implicits._
 
+  private implicit val jsonMarshaller: JacksonJsonMarshaller.type = JacksonJsonMarshaller
```
arkadius: Please change type to `JsonMarshaller`. 
Also can u extract method e.g. processJsonMarshaller in `JsonMarshaller`, add there comment that it is temporary fix, we are using this marshalling strategy only for process json's marshalling and TODO that after switching to circe it will be probable removed (if it won't give performance benefits) and use it in all places where object was used?

PR #93, opened 2019-07-22
Feature: intervals improvement

lciolecki:
* Fix interval reloading data
* Retrive interval configuration from back-end

### DIFF
--- a/ui/client/components/HealthCheck.js
+++ b/ui/client/components/HealthCheck.js
@@ -1,30 +1,34 @@
-import React from "react";
-import {render} from "react-dom";
-import {connect} from "react-redux";
-import HttpService from "../http/HttpService";
+import React from "react"
+import {connect} from "react-redux"
+import HttpService from "../http/HttpService"
 import InlinedSvgs from '../assets/icons/InlinedSvgs'
 import PeriodicallyReloadingComponent from './PeriodicallyReloadingComponent'
+import ActionsUtils from "../actions/ActionsUtils";
 
 class HealthCheck extends PeriodicallyReloadingComponent {
 
   constructor(props) {
-    super(props);
+    super(props)
 
     this.state = {
       healthCheck: undefined,
-    };
+    }
+  }
+
+  getIntervalTime() {
+    let baseIntervalTime = _.get(this.props, "featuresSettings.intervalSettings.base", this.baseIntervalTime)
+    let processesIntervalTime = _.get(this.props, "featuresSettings.intervalSettings.healthCheck", this.intervalTime)
+    return processesIntervalTime || baseIntervalTime
   }
 
   reload() {
-    HttpService.fetchHealthCheck().then(
-      (check) => this.setState({ healthCheck: check })
-    );
+    HttpService.fetchHealthCheck().then((check) => this.setState({ healthCheck: check }))
   }
 
   render() {
-    const { healthCheck } = this.state;
+    const { healthCheck } = this.state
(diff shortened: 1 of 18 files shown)


ui/client/components/HealthCheck.js
```diff
@@ -1,40 +1,48 @@
...
-    );
-
+    )
   }
 }
 
-export default HealthCheck;
+
+const mapState = state => ({
+  featuresSettings: state.settings.featuresSettings,
```
mproch: maybe let's add map only intervalSettings here? 
Here and in all places below?

ui/client/components/PeriodicallyReloadingComponent.js
```diff
@@ -1,14 +1,19 @@
-import React from "react";
-
+import React from "react"
 
 class PeriodicallyReloadingComponent extends React.Component {
+  baseIntervalTime = 10000
```
mproch: we have two defaults now - here and in backend? Let's leave only one?

ui/client/containers/BaseProcesses.js
```diff
@@ -8,6 +8,8 @@ import HttpService from "../http/HttpService"
 
 class BaseProcesses extends PeriodicallyReloadingComponent {
   searchItems = ['categories']
+  hasReloadStatuses = false
```
mproch: shouldReloadStatuses?

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
```diff
@@ -47,6 +48,15 @@ case class RemoteEnvironmentConfig(targetEnvironmentId: String)
 case class EnvironmentAlert(content: String, cssClass: String)
 case class CommentSettings(matchExpression: String, link: String)
 case class DeploySettings(requireComment: Boolean)
+case class IntervalSettings(base: Int, processes: Int, healthCheck: Int)
```
mproch: Maybe it's better to put Options here? So that we can override only some settings?

And maybe name like ProcessListRefreshPeriodSettings? Interval can mean a lot of different things ;)

ui/client/components/PeriodicallyReloadingComponent.js
lciolecki: I'v changed it. Now default is only at frontend.

ui/client/containers/BaseProcesses.js
lciolecki: Better ;)

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
lciolecki: I'v changed it to IntervalTimeSettings - now it's okey.

ui/client/components/HealthCheck.js
lciolecki: I think, this is okey. There is only reference to the object.

mproch: yes, but it's a bit messy now as we pass all our configuration around...

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
mproch: kind of better, but still doesn't look like frontend refresh period...

PR #95, opened 2019-07-22
[Performance improvement] Less process json fetching

arkadius:


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/processdetails.scala
```diff
@@ -62,6 +61,14 @@ object processdetails {
     def idWithName: ProcessIdWithName = ProcessIdWithName(ProcessId(id.toLong), ProcessName(name))
   }
 
+  sealed trait ProcessShapeFetchStrategy[ProcessShape]
+
+  object ProcessShapeFetchStrategy {
```
mproch: We should think about unifying BasicProcess and BaseProcessDetails[Unit] - for now, let's add some comment/TODO for that?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -127,26 +127,27 @@ abstract class DBFetchingProcessRepository[F[_]](val dbConfig: DbConfig) extends
           businessView = false,
           subprocessesVersions = subprocessesVersions
         )
+        processDetails.asInstanceOf[BaseProcessDetails[PS]]
```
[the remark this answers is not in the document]
arkadius: I've moved it to one place and added comment

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/processdetails.scala
arkadius: Added

PR #99, opened 2019-07-23
Scala 2.12 build

mproch:


engine/flink/management/src/it/resources/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM flink:1.7.2-scala_2.11
+FROM flink:1.7.2-scala_${scala.binary.version}
```
lciolecki: Maybe we should also support different flink version? Because we will support alsow flink 1.9: https://github.com/TouK/nussknacker/pull/247?

mproch: In theory yes it would be great, but it's kind of tricky, because we'd have to release versions like:
nussknacker-api_1.9_2.11  :)
And keeping code base working with two versions of flink is more tricky than with two scala versions...
I think we'll have to rather have branch like flink_{older}_backports and release it when needed...

.travis.yml
```diff
@@ -60,23 +57,32 @@ stages:
...
-    - stage: tests
-      name: "Backend tests"
-      script: ./ciBackendTest.sh $TRAVIS_SCALA_VERSION
-      after_success: if [ "$COVERAGE" = true ]; then sbt ++$TRAVIS_SCALA_VERSION coveralls; else echo "Skipping coveralls"; fi
+    - stage: test
+      name: "Backend tests 2.11"
+      script: ./ciBackendTest.sh 2.11.12
+      after_success: if [ "$COVERAGE" = true ]; then sbt ++2.11.12 coveralls; else echo "Skipping coveralls"; fi
+    - name: "UI slow tests 2.11"
+      script: sbt ++2.11.12 clean ui/slow:test
```
arkadius: Why not to just one crossbuild task with single plus?

README.md
```diff
@@ -16,10 +16,3 @@ Talk to us on our [mailing list](https://groups.google.com/forum/#!forum/nusskna
 ## License
 
 **Nussknacker** is published under [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0).
-
-
-## Scala compatibility
```
arkadius: What about leaving this with notice that we support both 2.11.x and 2.12.x? Because there will be still some not supported versions...

engine/flink/management/src/it/resources/docker/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN chmod +x /docker-entrypoint.sh
 
 USER flink
 RUN mkdir -p /tmp/storage
-RUN mkdir -p /tmp/storage/1.4/blob/cache
+RUN mkdir -p /tmp/storage/1.7/blob/cache
```
arkadius: Wow it we have here old flink version? What does it cause? Please name this magic number so we will prevent future similar problems with it.

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingProcessManagerSpec.scala
```diff
@@ -191,22 +194,14 @@ class FlinkStreamingProcessManagerSpec extends FunSuite with Matchers with Scala
...
-
-      val readSignals = consumer.consume(signalsTopic).take(1).map(m => new String(m.message(), StandardCharsets.UTF_8)).toList
-      val signalJson = CirceUtil.decodeJsonUnsafe[Json](readSignals.head, "invalid signals").hcursor
-      signalJson.downField("processId").focus shouldBe Some(Json.fromString("test-process"))
-      signalJson.downField("action").downField("type").focus shouldBe Some(Json.fromString("RemoveLock"))
-      signalJson.downField("action").downField("lockId").focus shouldBe Some(Json.fromString("test-lockId"))
-    } finally {
-      kafkaClient.shutdown()
-    }
+    val consumer = kafkaClient.createConsumer()
```
arkadius: Why we needn't `kafkaClient.shutdown()` anymore?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/sql/HsqlSqlQueryableDataBase.scala
```diff
@@ -20,8 +22,14 @@ class HsqlSqlQueryableDataBase(query: String, tables: Map[String, ColumnModel])
 
   import HsqlSqlQueryableDataBase._
 
-  private val connection: Connection =
-    DriverManager.getConnection(s"jdbc:hsqldb:mem:${UUID.randomUUID()};shutdown=true;allow_empty_batch=true", "SA", "")
+  private val connection: Connection = {
+    val url = s"jdbc:hsqldb:mem:${UUID.randomUUID()};shutdown=true;allow_empty_batch=true"
+    //TODO: should not really happen, but sometimes happens in tests..
```
arkadius: Why it happens? Is there some kind of race? Can u be more specific here? Also can you catch more concrete exceptions e.g. using `catching[...] opt` syntax?

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/types/EspTypeUtilsSpec.scala
```diff
@@ -187,7 +186,8 @@ class EspTypeUtilsSpec extends FunSuite with Matchers {
     final val quxDocs = "This is sample documentation for qux method"
   }
 
-  test("should extract description and params from method") {
+  //FIXME: scala 2.11, 2.12 have different behaviour...
```
arkadius: How we will live with it? Is scala 2.12 build is still usable without this fix?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaBinaryConfig.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.engine.util.config
+
+import java.util.Collections
+
+import com.typesafe.config.{Config, ConfigFactory}
+
+object ScalaBinaryConfig {
+
+  val scalaBinaryVersion: String = util.Properties.versionNumberString.replaceAll("(\\d+\\.\\d+)\\..*$", "$1")
```
arkadius: In `Properties.versionNumberString` we have nk version? If so, can you change name? Because we now we interested in at least four versions: nk ui version, nk runtime extension version, scala version, flink version. If u can't at least extract local variable `nkVersion`. Also can we have in properties scala version so it regex won't be needed?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessTypesForCategories.scala
```diff
@@ -9,7 +9,7 @@ import scala.collection.JavaConverters._
 import net.ceedubs.ficus.Ficus._
 
 
-class ProcessTypesForCategories(config: Config) {
+case class ProcessTypesForCategories(config: Config) {
```
arkadius: Why it is case class? It is now a bean that provide publicly `config` and will be somewhere serialized with underlying `config`?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaBinaryConfig.scala
mproch: no, in Properties.versionNumberString there is exact scala library version.

README.md
mproch: true, I'll add it back

.travis.yml
mproch: Well, in theory it gives us better parallelism... I can try + version

engine/flink/management/src/it/resources/docker/Dockerfile
mproch: I think it was not needed for a long time ;) I'll try to remove this line

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessTypesForCategories.scala
mproch: true that, we don't need case class here

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/sql/HsqlSqlQueryableDataBase.scala
mproch: I'll extract it to another PR as it's not scala 2.12 related...

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingProcessManagerSpec.scala
mproch: kafkaClient is created in StreamingDockerTest, and we shut it down in afterAll method

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/types/EspTypeUtilsSpec.scala
mproch: I added more accurate description, and ignore only two problematic cases, we can try to fix it in next PR

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaBinaryConfig.scala
arkadius: Ok, it is `scala/util/Properties`... What do u thing about renaming this method to scalaMajorVersion? I was confused because I thought that you here extract scala version from some other version... because eg. 2.12.3 it is version of scala binary.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/sql/HsqlSqlQueryableDataBase.scala
arkadius: Ok, so please only add notice that it is because some race condition or what u find out during work on this PR...

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaBinaryConfig.scala
mproch: oook, by scalaBinary I meant binary compatilibility - I'll change to major :+1:

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessTypesForCategories.scala
mproch: done

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaBinaryConfig.scala
mproch: done, changed to major everywhere

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ScalaMajorVersionConfig.scala
```diff
@@ -0,0 +1,17 @@
...
+import java.util.Collections
+
+import com.typesafe.config.{Config, ConfigFactory}
+
+object ScalaMajorVersionConfig {
+
+  val scalaMajorVersion: String = util.Properties.versionNumberString.replaceAll("(\\d+\\.\\d+)\\..*$", "$1")
+
+  def configWithScalaMajorVersion(config: Config = ConfigFactory.defaultApplication()): Config = {
+    val withBinary = ConfigFactory.parseMap(Collections.singletonMap("scala.major.version", scalaMajorVersion))
```
arkadius: withMajor

mproch: fixed...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/sql/HsqlSqlQueryableDataBase.scala
arkadius: I would like to leave it with `Try` and description as it was instead of just removing it :-D But your choice

PR #103, opened 2019-07-24
Query archived processes in archive tab

jedrz:


ui/client/containers/Archive.js
```diff
@@ -15,7 +15,7 @@ import ProcessUtils from "../common/ProcessUtils"
 
 class Archive extends BaseProcesses {
   queries = {
-    isArchived: false
+    isArchived: true
```
arkadius: woops :)

lciolecki: Thanks guys! :)

PR #134, opened 2019-09-19
Feature: when checking deployment status check if versions are the same. Also: better error messages on process list.

mproch:


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -119,21 +120,39 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, sender: HttpSe
...
   }
 
+  //TODO: cache by jobId?
+  private def checkVersion(jobId: String, name: ProcessName): Future[Option[ProcessVersion]] = {
+    sender.send {
+      (flinkUrl / "jobs" / jobId / "config").GET OK utils.asJson[JobConfig]
+    }.map { config =>
+      val userConfig = config.`execution-config`.`user-config`
+      for {
+        version <- userConfig.get("versionId").flatMap(_.string).map(_.toLong)
```
arkadius: Why you flatMap on this? Does all this things are normal business case or rather they are kind of really unexpected behaviour? If the second one, maybe will be better to throw exception? Looking on this I think about how many strange things could happen (missing versionId, versionId which is not string, versionId which is not string holding long...) and wondering if developer could see here only real business scenarios like: there is correct process config in user config or there isn't.

mproch: good catch :) it's mainly for backward compatibility of deployed jobs, I added TODO and explanation

PR #140, opened 2019-09-30
Typed global variables

arkadius:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -160,6 +160,18 @@ object typing {
       }
     }
 
+    def apply(obj: TypedMap): TypingResult = {
```
mproch: maybe we could have apply(obj: AnyRef): TypingResult? 
In other cases (currently) it would just do Typed(obj.getClass)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ProcessDefinitionExtractor.scala
```diff
@@ -48,9 +49,13 @@ object ProcessDefinitionExtractor {
 
     //TODO: this is not so nice...
     val globalVariablesDefs = expressionConfig.globalProcessVariables.map { case (varName, globalVar) =>
-      val klass = Typed(globalVar.value.getClass)
-      (varName, ObjectWithMethodDef(globalVar.value, MethodDefinition(varName, (_, _) => globalVar, new OrderedDependencies(List()), klass,  klass, List()),
-        ObjectDefinition(List(), klass, globalVar.categories, SingleNodeConfig.zero)))
+      val runtimeClass = Typed(globalVar.value.getClass)
+      val typingResult = globalVar.value match {
```
mproch: I would move this pattern matching to Typed.apply

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
arkadius: I thought about it, but have some doubts that this approach will make `Typed` api to be unclear, because `Class | ClassRef <: AnyRef`. I know that we have priorities in method overloading but still I think that it is not the best api design choice. Maybe we can name this factory method? It should be done carefully because we already have: `apply`, `detailed`. Maybe we should name this `fromInstance`? Than for consistence `detailed` will be renamed to `fromDetailedType` or sth similar. WDYT?

mproch: yes, can be some other method, or even other class - but let's keep it in one place

arkadius: Ok, added to Typed as a named factory method. Also renamed detailed -> fromDetailedType for consistency.

PR #146, opened 2019-10-02
Fix bug with permissions

### LINKED ISSUE #121: Process list shows processes within Category that shouldn't be visible by user.
**Describe the bug**
For user, on the list of processes are visible all processes - even if he/she hasn't got configured Read permission for some category.

**To Reproduce**
1. Having Flink and Nussknacker ran e.g. in docker with configured two categories: one visible for user A and the second not visible.
2. Open browser, login as a user A and go to 'main page'

**Expected behavior**
User A should see processes in categories that he/she has Read permission,
, but:
User A see all processes.

lciolecki:
This PR closes #121 
 
- Remove Permission.Admin
- Add isAdmin property at user config file
- Add security for changing category, now this can do only admin
- Add some todos
- Some tests refactor

### DIFF
--- a/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -1,15 +1,17 @@
 package pl.touk.nussknacker.ui.security.api
 
-import Permission.Permission
+import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
-case class LoggedUser(id: String,
-                      categoryPermissions: Map[String, Set[Permission]]=Map.empty) {
+case class LoggedUser(
+  id: String,
+  categoryPermissions: Map[String, Set[Permission]] = Map.empty,
+  isAdmin: Option[Boolean] = Option.apply(false)
+) {
   private val permissions = categoryPermissions.values.flatten.toSet
+
   def hasPermission(permission: Permission): Boolean = {
-    permissions.contains(permission) || isAdmin
+    hasAdminPermission || permissions.contains(permission)
   }
 
-  //TODO: remove
-  def isAdmin: Boolean = permissions.contains(Permission.Admin)
-
+  def hasAdminPermission: Boolean = isAdmin.getOrElse(false)
 }
--- a/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -5,13 +5,11 @@ import pl.touk.nussknacker.ui.security.api.Permission.Permission
 case class LoggedUser(
   id: String,
   categoryPermissions: Map[String, Set[Permission]] = Map.empty,
-  isAdmin: Option[Boolean] = Option.apply(false)
+  isAdmin: Boolean = false
 ) {
   private val permissions = categoryPermissions.values.flatten.toSet
 
   def hasPermission(permission: Permission): Boolean = {
-    hasAdminPermission || permissions.contains(permission)
+    isAdmin || permissions.contains(permission)
   }
-
-  def hasAdminPermission: Boolean = isAdmin.getOrElse(false)
 }
--- a/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntax.scala
+++ b/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntax.scala
@@ -5,12 +5,11 @@ object PermissionSyntax {
   implicit class UserPermissionsQuery(user: LoggedUser) {
     def can(permission: Permission.Permission): Set[String] =
       user.categoryPermissions.collect{
-        case (c, p) if user.hasAdminPermission || p.exists(containsPermission(permission)) =>
-          c
+        case (c, p) if user.isAdmin || p.exists(containsPermission(permission)) => c
       }.toSet
 
     def can(category:String, permission: Permission.Permission): Boolean =
-      user.hasAdminPermission || user.categoryPermissions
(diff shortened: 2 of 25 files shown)


engine/security-api/src/test/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntaxTest.scala
```diff
@@ -21,14 +20,26 @@ class PermissionSyntaxTest extends FunSuite with Matchers {
       u.can(p) shouldEqual c
     }
   }
+
   test("Admin permission grands other permissions") {
-    PermissionSyntax.containsPermissionOrAdmin(Permission.Read)(Permission.Admin) shouldBe true
+    def admin(cp: Map[String, Set[Permission]]) = LoggedUser("admin", cp, Option.apply(true))
```
mproch: Option(true), and later Option(false)?

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -1,15 +1,17 @@
...
 
-import Permission.Permission
+import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
-case class LoggedUser(id: String,
-                      categoryPermissions: Map[String, Set[Permission]]=Map.empty) {
+case class LoggedUser(
+  id: String,
+  categoryPermissions: Map[String, Set[Permission]] = Map.empty,
+  isAdmin: Option[Boolean] = Option.apply(false)
```
mproch: if we have default value does it make sense to have option here?

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -1,15 +1,17 @@
...
+
   def hasPermission(permission: Permission): Boolean = {
-    permissions.contains(permission) || isAdmin
+    hasAdminPermission || permissions.contains(permission)
   }
 
-  //TODO: remove
-  def isAdmin: Boolean = permissions.contains(Permission.Admin)
-
+  def hasAdminPermission: Boolean = isAdmin.getOrElse(false)
```
mproch: maybe just isAdmin?

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntax.scala
```diff
@@ -5,17 +5,16 @@ object PermissionSyntax {
   implicit class UserPermissionsQuery(user: LoggedUser) {
     def can(permission: Permission.Permission): Set[String] =
```
mproch: If we make some cleaning maybe let's remove/change this method? It's a bit misleading in admin context - in fact for admin we want *all* categories and not only those to which admin has explicit permissions? See usages

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -17,8 +17,12 @@ trait ProcessRepository[F[_]] extends Repository[F] with EspTables {
   import api._
   import pl.touk.nussknacker.ui.security.api.PermissionSyntax._
   protected def processTableFilteredByUser(implicit loggedUser: LoggedUser): Query[ProcessEntityFactory#ProcessEntity, ProcessEntityData, Seq] = {
+    if (loggedUser.hasAdminPermission) {
```
mproch: let's not use explicit return and many exit points...

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
lciolecki: Good idea :)

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
lciolecki: Changed.

engine/security-api/src/test/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntaxTest.scala
lciolecki: Now is just LoggedUser("admin", cp, true).

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
lciolecki: Changed.

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntax.scala
lciolecki: I agree.. But if we want do this good, than we should create separated config which has all available categories.. It shoud be more complex solutions - now I want only fix it ;)

ui/server/develConf/sample/users.conf
```diff
@@ -2,13 +2,8 @@ users: [
...
     id: "admin"
     password: "admin"
-    categoryPermissions: {
-      "Category1": ["Read", "Write", "Deploy", "Admin"]
-      "Category2": ["Read", "Write", "Deploy", "Admin"]
-      "Technical": ["Read", "Write", "Deploy", "Admin"]
-      "Default": ["Read", "Write", "Deploy", "Admin"]
-      "StandaloneCategory1": ["Read", "Write", "Deploy", "Admin"]
-    }
+    isAdmin: true
```
mproch: I think we should fix it also in different places - e.g. docker-users.conf, etc?

lciolecki: Ofc you are right :)

PR #148, opened 2019-10-03
Convert isDisabled parameter in split to option

jedrz:
Now it's consistent with other nodes like `Sink`, `Processor`, etc. What is more processes details could not be deserialized by circe because for some reason isDisabled field was omitted by argonaut - probably there is a different optional parameters behaviour in these libraries.


jedrz:
@lciolecki Unfortunately I don't see any way to provide useful tests. I've spotted difference in how split node is serialized by argonaut in older version of nussknacker, simutaneoulsy it cannot be deserialized by circe in current version.

jedrz:
Finally I decided to remove isDisabled parameter. Split cannot be disabled easily and it's not possible to disable it in frontend.

PR #199, opened 2019-10-09
Remove argonaut from examples, minor cleaning

mproch:


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/sources.scala
```diff
@@ -35,26 +35,20 @@ object sources {
...
-  private def jsonToMap(jo: JsonObject): Map[String, _] = {
-    jo.toMap.mapValuesNow { jsonField =>
+  private def jsonToMap(jo: Json): Map[String, _] = {
+    jo.asObject.getOrElse(JsonObject()).toMap.mapValuesNow { jsonField =>
       jsonField.fold(
         jsonNull = null,
-        jsonBool = identity,
-        jsonNumber = _.truncateToLong,
+        jsonBoolean = identity,
+        //FIXME:
```
jedrz: What about this? Is there anything to do?

PR #200, opened 2019-10-10
Include deployment comment in /api/processes/:name/deployments endpoint

bottaio:
Returning merely commentId from this endpoint brings close to no value to the user. In scenario when user wants to read this comment's content they have to fetch all comments for all versions and filter them accordingly. Having explicit endpoint just to fetch comment seems redundant thus this change.


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -161,17 +161,19 @@ abstract class DBFetchingProcessRepository[F[_]](val dbConfig: DbConfig) extends
 
   def fetchDeploymentHistory(processId: ProcessId)(implicit ec: ExecutionContext): F[List[DeploymentHistoryEntry]] =
     run(deployedProcessesTable.filter(_.processId === processId.value)
-      .sortBy(_.deployedAt.desc)
-      .result.map(_.map(de =>
-      DeploymentHistoryEntry(
+      .joinLeft(commentsTable.filter(_.processId === processId.value))
```
mproch: is this filter necessary?

bottaio: nope, i'll delete it

PR #203, opened 2019-10-11
Remove argonaut from clients, move to sttp

mproch:


engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/management/StandaloneProcessClient.scala
```diff
@@ -68,45 +70,34 @@ class MultiInstanceStandaloneProcessClient(clients: List[StandaloneProcessClient
...
-        Some(asJson[ProcessState](r))
-      else
-        throw StatusCode(r.getStatusCode)
-    }
-
-    val jobStatusUrl = dispatch.url(managementUrl) / "checkStatus" / name.value
-    dispatchClient {
-      jobStatusUrl > notFoundHandler _
-    }
+    sttp
```
lciolecki: There is something like: isSuccess: https://sttp.readthedocs.io/en/latest/responses/basics.html#response-code. I think this is nicer soluton :)

PR #210, opened 2019-10-15
Add demo backend mode for FE development

ppd-touk:


ui/client/config.js
```diff
@@ -15,7 +15,7 @@ let API_URL = `${nkPath}/api`
 const dateFormat = "YYYY-MM-DD HH:mm:ss"
 
 if (__DEV__) {
-  API_URL = 'http://localhost:8081/api'
+  API_URL = 'http://localhost:3000/api'
```
mproch: can't we remove this "if" now - we'll be able to use relative path just like in production?

ppd-touk: Done

ui/client/package.json
```diff
@@ -6,7 +6,8 @@
   "license": "UNLICENSED",
   "private": true,
   "scripts": {
-    "start": "webpack-dev-server",
+    "start": "PROXY_API_DOMAIN=http://localhost:8081/api webpack-dev-server",
```
lciolecki: There should be PROXY_API_URI not PROXY_API_DOMAIN.

PR #214, opened 2019-10-15
Use "cancel" instead of "stop" in reference to a process Closes #209

### LINKED ISSUE #209: Deployment "stop"
"stop" in fact undeploys the process (cancels flink job). You cannot play or resume the same process again. You have to deploy the current version. I would rename it to "cancel" or "remove".

ppd-touk:


ui/client/components/right-panel/UserRightPanel.js
```diff
@@ -119,7 +119,7 @@ class UserRightPanel extends Component {
         panelName: "Deployment",
         buttons:[
           {name: "deploy", visible: this.props.capabilities.deploy, disabled: !conf.deployPossible, icon: InlinedSvgs.buttonDeploy, btnTitle: conf.deployToolTip, onClick: this.deploy, onMouseOver: conf.deployMouseOver, onMouseOut: conf.deployMouseOut},
-          {name: "stop", visible: this.props.capabilities.deploy, disabled: !this.isRunning(), onClick: this.stop, icon: InlinedSvgs.buttonStop},
+          {name: "cancel", visible: this.props.capabilities.deploy, disabled: !this.isRunning(), onClick: this.stop, icon: InlinedSvgs.buttonStop},
```
mproch: maybe let's change also method and button name to avoid confustion?

ppd-touk: You're right, done

PR #218, opened 2019-10-16
Create subprocess when migrating one

jedrz:
Currently subprocess is migrated as a process.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
@@ -6,7 +6,7 @@ import java.nio.charset.StandardCharsets
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshalling.Marshal
-import akka.http.scaladsl.model.Uri.Path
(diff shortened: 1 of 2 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -163,7 +162,7 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
     EitherT {
       invokeStatus(HttpMethods.GET, List("processes", localProcess.id)).flatMap { status =>
         if (status == StatusCodes.NotFound)
-          invokeForSuccess(HttpMethods.POST, List("processes", localProcess.id, category))
+          invokeForSuccess(HttpMethods.POST, pathParts = List("processes", localProcess.id, category), queryString = Some(s"isSubprocess=${localProcess.metaData.isSubprocess}"))
```
mproch: can we do it in more "typesafe" way? not glueing strings like that?

jedrz: Done. For no query string, single `?` character is not appended anymore. I extracted some methods that invoke API - please have a look at last commit.

PR #224, opened 2019-10-17
Plugin API

bottaio:
This PR adds support for pluggable components that essentially are implemented as event handlers. Nussknacker exposes set of events as `ChangeEvent` trait. This makes it easier for users to implement custom logic based on processes' changes.


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/db/entity/ProcessVersionEntityData.scala
```diff
@@ -0,0 +1,28 @@
+package pl.touk.nussknacker.restmodel.db.entity
```
mproch: We should rethink what should be in this package. db.entity in REST model doesn't look too good...
Maybe we need another layer?

mproch:
Let's split it to smaller parts

PR #226, opened 2019-10-17
LoggedUser refactoring

### LINKED ISSUE #206: empty "creator panel"
**Describe the bug**
There are no items In the "creator panel". Categories are shown properly.

**To Reproduce**
1. https://demo.nussknacker.io/
2. edit any process

**Screenshots**

image

bottaio:
New permissions system introduced bug where admin user sees no elements in control panel. I've decided to redesign LoggedUser class to explicitly handle admin case on type level. The major problem currently is the need for categoryPermissions map for non-admin users. When admin is concerned, he shall always be permitted to do every action he needs and thus returning list of categories for which the permission is granted is redundant and has been removed.

Fixes #206

### DIFF
--- a/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -4,7 +4,7 @@ import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
 sealed trait LoggedUser {
   val id: String
-  def hasPermission(category: String, permission: Permission): Boolean
+  def can(category: String, permission: Permission): Boolean
   val isAdmin: Boolean
 }
 
@@ -25,7 +25,7 @@ case class CommonUser(id: String,
     case (category, permissions) if permissions contains permission => category
   }.toSet
 
-  override def hasPermission(category: String, permission: Permission): Boolean = {
+  override def can(category: String, permission: Permission): Boolean = {
     categoryPermissions.get(category).exists(_ contains permission)
   }
 
@@ -33,6 +33,6 @@ case class CommonUser(id: String,
 }
 
 case class AdminUser(id: String) extends LoggedUser {
-  override def hasPermission(category: String, permission: Permission): Boolean = true
+  override def can(category: String, permission: Permission): Boolean = true
   override val isAdmin: Boolean = true
 }
--- a/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -2,14 +2,37 @@ package pl.touk.nussknacker.ui.security.api
 
 import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
-case class LoggedUser(
-  id: String,
-  categoryPermissions: Map[String, Set[Permission]] = Map.empty,
-  isAdmin: Boolean = false
-) {
-  private val permissions = categoryPermissions.values.flatten.toSet
+sealed trait LoggedUser {
+  val id: String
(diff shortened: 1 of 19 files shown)


engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -2,14 +2,37 @@ package pl.touk.nussknacker.ui.security.api
...
-  categoryPermissions: Map[String, Set[Permission]] = Map.empty,
-  isAdmin: Boolean = false
-) {
-  private val permissions = categoryPermissions.values.flatten.toSet
-
-  def hasPermission(permission: Permission): Boolean = {
-    isAdmin || permissions.contains(permission)
+sealed trait LoggedUser {
+  val id: String
+  def hasPermission(category: String, permission: Permission): Boolean
```
witekw: can we "hide" it somehow, so the only api for checking permission is u.can(c, p) ?

bottaio: I was actually thinking about removing `PermissionSyntax` and renaming `hasPermission` -> `can`. I'll do it that way.

witekw: either way is good

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/UserResources.scala
```diff
@@ -13,14 +13,29 @@ class UserResources(typesForCategories: ProcessTypesForCategories)(implicit ec:
...
\ No newline at end of file
+@JsonCodec case class DisplayableUser private(id: String, isAdmin: Boolean, categories: List[String], categoryPermissions: Map[String, Set[String]])
+
+object DisplayableUser {
+  import pl.touk.nussknacker.engine.util.Implicits._
+
+  def apply(user: LoggedUser, allCategories: List[String]): DisplayableUser = user match {
+    case CommonUser(id, categoryPermissions) => new DisplayableUser(
+      id = id,
+      isAdmin = false,
```
lciolecki: false is default, you don't need set it.

bottaio:
> @bottaio so admin users shoud alsow have categoryPermissions, because at backend api we filtering elements from control panel?

yeah, in current version we rely on categoryPermissions when filtering elements -- we grant each user permission to use an element if he has read permission for processes category; we have handled it previously on database level -- by getting rid of filtering statement but this case was not handled properly

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/UserResources.scala
bottaio: it's not and btw, I'd prefer to make it explicit

engine/security-api/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
bottaio: done

engine/security-api/src/test/scala/pl/touk/nussknacker/ui/security/api/PermissionSyntaxTest.scala
```diff
@@ -8,19 +8,6 @@ import org.scalatest.prop.{TableFor3, TableFor4}
...
-      ("categoryPermissions", "permission", "categories"),
-      (u(Map("c1"->Set(Read))), Read, Set("c1")),
-      (u(Map("c2"->Set(Write))), Read, Set.empty)
-    )
-    forAll(perms) { (u: LoggedUser, p: Permission, c: Set[String]) =>
-      u.can(p) shouldEqual c
-    }
-  }
-
   test("Admin permission grands other permissions") {
```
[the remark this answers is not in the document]
bottaio: done

Issue #192, opened 2019-10-07
readonly fields are indistinguishable from editable ones


witekw:
![image](https://user-images.githubusercontent.com/632221/66350682-437c6680-e95c-11e9-872f-2da83736d02c.png)

"Ref Type" in the above case

arkadius:
This problem still exists in many places:
- in compare modal
- in grouped nodes details modal

Some fields are distinguishable like id, node type in "compare modal" but other are not like: expression editor or area input (used in description)

arkadius:
There are still problems:

- checkboxes not change background
- simple/raw editor switcher not changes style
- some other fields not changes style: description, some other fields in compare

PR #227, opened 2019-10-17
Change comment section design, focus on first item in newly opened dialog, distinguish between read-only and other fields

### LINKED ISSUE #220: process comments redesign
image

The most important information - the comment is:
- hiding in less relevant surrounding information e.g. redundant process name,
- is badly aligned vertically, you cannot distinguish items on the comments list

I would:
1. move label with the user name to the right
2. remove process name
3. move version to the "header"
2019-10-16 | 12:34 | v1 (admin)

"Write a comment ..." box accepts multiline input, but shows it later without new lines. Maybe change textarea to simple input?

ppd-touk:
Closes #220, #197, #192

### DIFF
--- a/ui/client/app.styl
+++ b/ui/client/app.styl
@@ -83,6 +83,8 @@ processActivityList()
       &.date
         color panelTitleTextColor
         font-style italic
+      &.comment-header
+        color #808080
     p
       font-style italic
       color panelTitleTextColor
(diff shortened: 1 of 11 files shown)


lciolecki:
In comments maybe you can try add underlined line ? :)

ppd-touk:
> In comments sections under each comment you can try add underlined line ? :)

I've added lines between comments. Please look at this photo to check how it looks like.
comments

lciolecki:
IMO now is definitely better! Good job!

PR #228, opened 2019-10-17
Fixing /api/processes/import/:processName route

bottaio:
Instead of importing process currently route` /api/processes/:name/:category` is matched instead.
Just reorganizing routes so that `/api/processes/import/:name` gets matched first.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -145,6 +145,32 @@ class ProcessesResources(val processRepository: FetchingProcessRepository,
               } yield statuses
             }
           }
+        } ~ path("processes" / "import" / Segment) { processName =>
+          processId(processName) { processId =>
+            (canWrite(processId) & post) {
+              fileUpload("process") { case (metadata, byteSource) =>
+                complete {
+                  MultipartUtils.readFile(byteSource).map[ToResponseMarshallable] { json =>
+                    (ProcessMarshaller.fromJson(json) match {
+                      case Valid(process) if process.metaData.id != processId.name.value => Invalid(WrongProcessId(processId.name.value, process.metaData.id))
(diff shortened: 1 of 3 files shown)


mproch:
Can you rebase it to master, without logged user refactor? 
Also: do you know why tests don't detect the problem? We have some for importing and they seem to pass :/

bottaio:
> Can you rebase it to master, without logged user refactor?
> Also: do you know why tests don't detect the problem? We have some for importing and they seem to pass :/

Yeah, it is easier to see a problem when you are logged in as admin. Only then routes always get mismatched. 

Look at `authorize(user.can(category, Permission.Write))` for `api/processes/:name/:category`. In tests we are using common user with permissions for all explicitly specified categories (`TESTCAT`). Here, router interpreted process name (`fooProcess`) as category, thus Akka rejected this route as user didn't have permissions to access it. On the other hand admin has permission to access whatever he wishes -- including nonexisting categories :(

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -145,6 +145,32 @@ class ProcessesResources(val processRepository: FetchingProcessRepository,
...
+                    }) match {
+                      case Valid(process) =>
+                        processRepository.fetchLatestProcessDetailsForProcessIdEither[Unit](processId.id).map { detailsXor =>
+                          val validatedProcess = detailsXor
+                            .map(details => ProcessConverter.toDisplayable(process, details.processingType))
+                            .map(processValidation.toValidated)
+                          toResponseXor(validatedProcess)
+                        }
+
+                      case Invalid(error) => EspErrorToHttp.espErrorToHttp(error)
```
lciolecki: Maybe you can move it to some service or dedicated method?

bottaio: good point

PR #234, opened 2019-10-18
Fix for typing error: #input was available in parameter's expressions + some refactoring

arkadius:
Refactoring:

- moved "factory" logic from ProcessCompiler to ProcessCompiler.compile to see which ExpressionCompiler's are used
- moved "factory" logic from SpelExpressionParser to SpelExpressionParser.default
- moved  typing logic from SpelExpressionValidator to Typer
- moved wrapping in dto logic from ProcessValidation to Resources to have one point to every validations in gui
- cleanup: passes ValidationContext to compileValidatedObjectParameters instead of Option[ValidationContext]
- Validated.fold -> Validated.map().valueOr()
- Validated[NonEmptyList[E],A] => ValidatedNel[E,A]

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompilationResult.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompilationResult.scala
@@ -2,7 +2,10 @@ package pl.touk.nussknacker.engine.compile
 
 import cats.data.Validated.{Invalid, Valid}
 import cats.data.{NonEmptyList, ValidatedNel}
+import cats.kernel.Semigroup
+import cats.instances.map._
 import cats.{Applicative, Traverse}
+import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.api.context.{ProcessCompilationError, ProcessUncanonizationError, ValidationContext}
 import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 import pl.touk.nussknacker.engine.canonize.{MaybeArtificial, MaybeArtificialExtractor}
@@ -11,6 +14,16 @@ import scala.language.{higherKinds, reflectiveCalls}
 
 case class CompilationResult[+Result](private [compile] val typing: Map[String, ValidationContext],
                                      result: ValidatedNel[ProcessCompilationError, Result]) {
+
+  import CompilationResult._
+
+  def andThen[B](f: Result => CompilationResult[B]): CompilationResult[B] =
+    result match {
+      case Valid(a)       =>
+        val newResult = f(a)
+        newResult.copy(typing = Semigroup.combine(typing, newResult.typing))
+      case i @ Invalid(_) => CompilationResult(typing, i)
+    }
 
   def map[T](action: Result => T) : CompilationResult[T] = copy(result = result.map(action))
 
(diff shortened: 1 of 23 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompilationResult.scala
```diff
@@ -45,4 +58,17 @@ object CompilationResult extends Applicative[CompilationResult] {
...
-}
+
+  implicit def takingLastNodeTypingInfoSemigroup: Semigroup[ValidationContext] = new Semigroup[ValidationContext] with LazyLogging {
+    override def combine(x: ValidationContext, y: ValidationContext): ValidationContext = {
+      logger.whenWarnEnabled {
+        if (x != y) {
+          logger.warn(s"Merging different ValidationContext for the same nodes: $x != $y. This can be a bug in code or duplicated node ids with different node typing info")
+        }
+      }
+      // we should be lax here because we won't to detect duplicate nodes and context can be different than
```
mproch: I think some words are missing from this comment :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -140,43 +139,25 @@ class SpelExpression(parsed: ParsedSpelExpression,
   }
 }
 
-class SpelExpressionParser(expressionFunctions: Map[String, Method],
+class SpelExpressionParser(// for SpelExpression creation
```
mproch: maybe we can extract class here, so we don't have to write these comments? :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -140,43 +139,25 @@ class SpelExpression(parsed: ParsedSpelExpression,
...
+                           expressionFunctions: Map[String, Method],
                            expressionImports: List[String],
                            classLoader: ClassLoader,
+                           propertyAccessors: Seq[PropertyAccessor],
```
mproch: why do we extract it to constructor? Will we need it in the future?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompilationResult.scala
arkadius: It should be "then" instead "than". Have it sense?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
arkadius: It is used in `SpelExpression` in creation of `EvaluationContext`. My main point in this change was to move factory logic from `SpelExpressionParser` to factory method (`default`), because then we can easily modify `SpelExpressionParser` (by some with... method) without recreating it from scratch. We could think about also moving factory logic from `SpelExpression.evaluate` and then we will replace these few properties with one `createEvaluationContext: () => EvaluationContext`. What do you think about it?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
arkadius: What do you think about replacing it with one `createEvaluationContext: () => EvaluationContext`?

PR #235, opened 2019-10-18
Typing ast

arkadius:
This change adds typing info for each expression in each node in process. Typing info depends on used expression parser. Currently is implemented only typing info for SpEL. It contains computed type for each SpEL AST node.

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
@@ -35,6 +35,8 @@ case class ValueWithLazyContext[T](value: T, lazyContext: LazyContext)
 
 sealed trait TypedValue
 
-case class TypedExpression(expression: Expression, returnType: TypingResult) extends TypedValue
+case class TypedExpression(expression: Expression, returnType: TypingResult, typingInfo: ExpressionTypingInfo) extends TypedValue
 
 case class TypedExpressionMap(valueByKey: Map[String, TypedExpression]) extends TypedValue
+
+trait ExpressionTypingInfo
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
@@ -39,4 +39,8 @@ case class TypedExpression(expression: Expression, returnType: TypingResult, typ
 
 case class TypedExpressionMap(valueByKey: Map[String, TypedExpression]) extends TypedValue
 
+/**
+  * It contains information about intermediate result of typing of expression. Can be used for further processing of expression
+  * like some substitutions base on type...
+  */
 trait ExpressionTypingInfo
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
@@ -1,11 +1,9 @@
 package pl.touk.nussknacker.engine.compile
 
-import cats.Applicative
 import cats.data.Validated._
 import cats.data.{NonEmptyList, ValidatedNel}
 import cats.instances.list._
 import cats.instances.option._
-import cats.kernel.Semigroup
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError._
 import pl.touk.nussknacker.engine.api.context.{PartSubGraphCompilationError, ProcessCompilationError, ValidationContext}
 import pl.touk.nussknacker.engine.api.definition.Parameter
@@ -27,7 +25,7 @@ import pl.touk.nussknacker.engine.util.validated.ValidatedSyntax
 import pl.touk.nussknacker.engine.{api, compiledgraph, _}
 
 import scala.util.{Failure, Success, Try}
-
+import PartSubGraphCompiler._
 
 class PartSubGraphCompiler(protected val classLoader: ClassLoader,
                            protected val expressionCompiler: ExpressionCompiler,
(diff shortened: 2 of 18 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
```diff
@@ -35,6 +35,8 @@ case class ValueWithLazyContext[T](value: T, lazyContext: LazyContext)
 
 sealed trait TypedValue
 
-case class TypedExpression(expression: Expression, returnType: TypingResult) extends TypedValue
+case class TypedExpression(expression: Expression, returnType: TypingResult, typingInfo: ExpressionTypingInfo) extends TypedValue
 
 case class TypedExpressionMap(valueByKey: Map[String, TypedExpression]) extends TypedValue
+
+trait ExpressionTypingInfo
```
mproch: maybe some comment how it can be used?

mproch: do we want to assume it can be displayed as json?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
```diff
@@ -51,24 +53,23 @@ class PartSubGraphCompiler(protected val classLoader: ClassLoader,
...
   protected def createServiceInvoker(obj: ObjectWithMethodDef) = ServiceInvoker(obj)
 
-  private val globalVariableTypes = expressionConfig.globalVariables.mapValues(_.returnType)
-
   def compile(n: SplittedNode[_], ctx: ValidationContext) : CompilationResult[compiledgraph.node.Node] = {
     implicit val nodeId: NodeId = NodeId(n.id)
 
-    def toCompilationResult[T](validated: ValidatedNel[ProcessCompilationError, T]) =
-      CompilationResult(Map(n.id -> ctx), validated)
+    def toCompilationResult[T](validated: ValidatedNel[ProcessCompilationError, T], expressionsTypingInfo: Map[String, ExpressionTypingInfo]) =
```
mproch: would Writer help here or it's not worth it?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
```diff
@@ -273,3 +297,21 @@ class PartSubGraphCompiler(protected val classLoader: ClassLoader,
     }
   }
 }
+
+case class ExpressionTypingResult(typingResult: TypingResult, typingInfo: Option[ExpressionTypingInfo]) {
```
mproch: can they be private?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -331,12 +349,23 @@ protected trait ProcessCompilerBase {
...
                                      (implicit nodeId: NodeId,
-                                      metaData: MetaData): ValidatedNel[ProcessCompilationError, T] = {
-    objectParametersExpressionCompiler.compileObjectParameters(nodeDefinition.parameters,
+                                      metaData: MetaData): (Map[String, ExpressionTypingInfo], ValidatedNel[ProcessCompilationError, T]) = {
+    val compiledObjectWithTypingInfo = objectParametersExpressionCompiler.compileObjectParameters(nodeDefinition.parameters,
       parameters,
       branchParameters, ctx.clearVariables, ctx).andThen { compiledParameters =>
       try {
-        Valid(factory.create[T](nodeDefinition, compiledParameters, outputVariableNameOpt))
+        val obj = factory.create[T](nodeDefinition, compiledParameters, outputVariableNameOpt)
```
mproch: can we extract it somewhere?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/ast/SpelNodePrettyPrinter.scala
```diff
@@ -0,0 +1,26 @@
+package pl.touk.nussknacker.engine.spel.ast
```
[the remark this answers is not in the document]
arkadius: In `Typer` for debugging purpose.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/expression/Expression.scala
arkadius: Some comment will be good. Regards to proposed assumption - I don't think so. I want to use it on backend side. It could be usefull if we could compute partial Typing for not completed expressions. Then we could move completion logic from FE to BE and do much more stuff. But for now it will be only helpful if whole expression is correct.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
arkadius: I thought about using Writer here and in `Typer` but IMO disadvantages outweigh the advantages. To use it I should use some kind of MonadTransformer to compose effects of Validated and Writer. IMO It would be too complicated for scala dev with medium knowledge. I think that adding such an abstraction give the best benefits if this pattern is shared between many classes. In this case it is very local thing.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
arkadius: Yep, I'll add it to companion object with private access

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
mproch: probably you're right, although this method is already far too complicated for mid scala developer :P

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
arkadius: You mean `factory.create` + surrounding try/catch? Good idea, will do that.

arkadius:
> not done yet I think :)
You mean that there are some enhancement that could be done, or you see functional things that are missed (not implemented cases)?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
arkadius: Yep, but I see no perfect solution for that. Also I'm not pretty sure if further polishing is worth doing it, because maybe in some near future we can do some general simplifications that will throw this code away like:
- Replacing all specified nodes like source, sink, filter and so on with simple "Node" with some `ContextTransformation`
- Removing `CanonicalProcess` layer
- Replacing PartSubGraphCompiler by some optional Optimizer that can be turned on or not.

PR #236, opened 2019-10-18
Global permissions

bottaio:
This PR introduces additional permissions configuration for use cases where one want to control access to certain functionalities but it is impossible or makes no sense to assign it to some category. 

Motivating example: admin tab in the UI. At the moment just admin user can access it but it also has pretty useful components for the end user - namely search / unused components and services. 

I was thinking about moving `custom processes` to a different place and eventually renaming `Admin` tab to something different. To make such change I'd like to discuss it with you guys first.


engine/security-api/src/test/scala/pl/touk/nussknacker/ui/security/api/LoggedUserTest.scala
```diff
@@ -8,7 +8,7 @@ import org.scalatest.prop.{TableFor3, TableFor4}
 class LoggedUserTest extends FunSuite with Matchers {
 
   test("Admin permission grants other permissions") {
```
jedrz: Can you provide tests for admin and common user?

PR #248, opened 2019-10-23
Add movable modal dialogs

### LINKED ISSUE #174: "business view" destroys graph
- create new process
- do not save it
- check "business view"
- uncheck "business view"
- process disappears

Maybe "Business view" should be unavailable before saving the process first.

### LINKED ISSUE #176: movable modal dialogs (UX)
Some windows are quite large. When they cover diagram, you lose some context e.g. can not quick check where in the diagram they are. 

(react-draggable?)

### LINKED ISSUE #232: new (invalid) node on diagram dissapears
1. open diagram
2. create node with an id X
3. try to add a node with the same id X, save
4. notification about duplicate id is displayed (fatal validation error)
5. node being added disappears

expected behaviour

You cannot save the node, you have to 1) change id and save, 2) cancel.
or
Node stays, diagram is in invalid state and not saved.

PPiedel:
This resolves #176 
This resolves #174 
This resolves #232

### DIFF
--- a/ui/client/components/AddProcessDialog.js
+++ b/ui/client/components/AddProcessDialog.js
@@ -9,7 +9,8 @@ import ActionsUtils from "../actions/ActionsUtils";
 import EspModalStyles from "../common/EspModalStyles";
 import "../stylesheets/visualization.styl";
 import HttpService from "../http/HttpService";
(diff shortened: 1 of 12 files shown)


PPiedel:
captured (4)

PR #255, opened 2019-10-25
wip: fix nodes naming inconsistencies

### LINKED ISSUE #242: creating subprocess with duplicate name
1. create subprocess with same name as an existing process (or subprocess)
2. error is 

image

I would expect that process / subprocess names are separate sets.
It would be nice not to be redireced to process edition at all when names is not available (you are redirected to invalid state).
When you start editing subprocess and save it, existing process is overwritten.

### LINKED ISSUE #243: base nodes naming inconsistencies
AS IS
  map variable
    Id
    Variable Name
    Fields
      possibility to add many fields
      one field already added, with default values: varName, 'value'
      when we click on "+" (add fields)
      placeholder "Field name", empty input box to pass an expression

  spel variable
    Id
    Variable Name
    Expression, default value: 'value'

  sql variable
    Id
    Variable Name
    Expression, default value: SELECT * FROM input

  switch
    Id
    Expression, default value: true
    exprValue, default value: output

TO BE

  less inital values, rather placeholders
  unified naming

  map variable
    no initial values in Fields, just placeholders: key, 'value'

  spel variable
    rename node to "variable" (we will allow different expression libraries?)
    Variable Name placeholder: name
    Expression placeholder: 'value'

  sql variable
    Variable Name placeholder: name
    Expression placeholder: SELECT * FROM input

  switch
    exprVal -> Variable Name (as variable is available later on in the flow, not only in scope of switch and outgoing edges)
    change placeholder value of Variable Name to name

  updated docs

PPiedel:
This resolves #242
This resolves #243 
This fixes some movable modal input bugs
This fixes "properties" modal "save" availability

### DIFF
--- a/ui/client/components/AddProcessDialog.js
+++ b/ui/client/components/AddProcessDialog.js
@@ -11,6 +11,7 @@ import "../stylesheets/visualization.styl";
 import HttpService from "../http/HttpService";
 import * as VisualizationUrl from '../common/VisualizationUrl';
 import Draggable from 'react-draggable';
+import {preventFromDraggableMoveSelectors} from "./modals/GenericModalDialog";
 
 //TODO: Consider integrating with GenericModalDialog 
 class AddProcessDialog extends React.Component {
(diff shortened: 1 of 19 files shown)


mproch:
Let's talk about this one. It introduces inconsistent behaviour - sometimes we have placeholders, sometimes default values. Currently it is possible to have default values in node configuration - and we *want* it to stay. But for filters etc we will now have placeholders - what's more they are hardcoded in FE, this is not so nice

mproch:
We'll tackle that in another PR

Issue #233, opened 2019-10-18
show params validation error in the context of params


witekw:
![image](https://user-images.githubusercontent.com/632221/67080244-1702ee80-f195-11e9-8776-953876285599.png)

Instead of showing "Field param2 (Required param - text) cannot be empty (field: param2)" at the top:

1. emphasize "Required param - text:" input field with orange border
2. display message or description below the input field

witekw:
After completing this issue please consider calling "/api/processValidation" during properties edit, so users get feedback about error earlier.

arkadius:
We can consider if we should use (ace/brace) errors indicator visible here: https://thlorenz.com/brace/
It will be nice if calling of api would be debounced during typing. Also we car return position range of error in api so we can underscore exact place of error

arkadius:
It is connected with: https://github.com/TouK/nussknacker/issues/249 ... During typing, type of expression can change, also errors can block inferring this type

PPiedel:
We also need to return field name for each validation error - now not all responses contain such a name.
Should we create separate issue for that?

witekw:
Can we make this change gradually?

- show errors/warnings near parameters when possible (GUI work only)
- improve on error reporting in API returns (#179)
- introduce dynamic validation (e.g. before node properties save on focus change)

witekw:
Consistent style to indicate errors (new issue?)

PR #267, opened 2019-10-29
Validate empty properties and duplicate (sub)process name

### LINKED ISSUE #242: creating subprocess with duplicate name

### LINKED ISSUE #263: Validate that node id and expression is not empty

PPiedel:
This resolves #242 
This resolves #263


ui/client/common/StringUtils.js
```diff
@@ -0,0 +1,3 @@
+export function capitalizeFirstLetter(phrase) {
+  return phrase.charAt(0).toUpperCase() + phrase.slice(1)
+}
```
witekw: css ::first-letter text-transform: capitalize ?

ui/client/containers/Processes.js
```diff
@@ -138,6 +140,8 @@ class Processes extends BaseProcesses {
           isSubprocess={false}
           visualizationPath={Processes.path}
           message="Create new process"
+          processes={this.state.processes}
```
witekw: usedProcessNames, takenProcessNames?

ui/client/stylesheets/processes.styl
```diff
@@ -206,4 +206,7 @@
   margin auto
 
 .transparent
-  background-color: rgba(0, 0, 0, 0)
\ No newline at end of file
+  background-color: rgba(0, 0, 0, 0)
+
+.validation-label
+  color red
```
witekw: why not use variable we have for color of warnings?

ui/client/common/StringUtils.js
PPiedel: Nice! Done.

ui/client/stylesheets/processes.styl
PPiedel: Good point, done

ui/client/containers/Processes.js
PPiedel: It contains info about processes, not only names. I think that processes is the shortest acceptable variable name.

ui/client/components/graph/node-modal/Input.js
```diff
@@ -15,6 +15,10 @@ const Input = (props) => {
                     onChange={(event) => onChange(path, event.target.value)}
                     readOnly={readOnly}
                 />
+                {
+                    isValid(value) ?
+                        <label className='node-details-validation-label'>{label + " can not be empty"}</label> : null
```
witekw: Message from resources?

ui/client/components/AddProcessDialog.js
```diff
@@ -105,5 +122,21 @@ function mapState(state) {
...
+
+const validations = [
+  {
+    isValid: (processes, subprocesses, name) => !_.isEmpty(name),
+    message: "Id can not be empty"
+  },
+  {
+    isValid: (processes, subprocesses, name) => !nameAlreadyExists(processes, subprocesses, name),
+    message: "Process or subprocess with given id already exists"
+  }
```
witekw: "Id already taken" ?

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -135,6 +135,10 @@ export class NodeDetailsContent extends React.Component {
     });
   };
 
+  validate(fieldValue, isRequired) {
```
witekw: It does not validate "in general" it merely checks if the value is not empty.

ui/client/components/graph/NodeDetailsContent.js
```diff
@@ -173,15 +179,15 @@ export class NodeDetailsContent extends React.Component {
       case 'SubprocessOutputDefinition':
         return (
           <div className="node-table-body">
-            {this.createField("input", "Id", "id", true)}
-            {this.createField("input", "Output name", "outputName")}
+            {this.createField("input", "Id", "id", true, true)}
```
witekw: Shouldn't this be more general solution? We could pass list of validators, in most cases predefined 1-element list with "notEmptyValidator", later we can put there a validator extracting value from processValidation response (#233). I think there should be single way od putting error information below fields. Later we have to consider if several messages can be presented below a field (I think it is a case, processValidation allows it). And how to distinguish between validations that prevent saving the modal from ones that "warn" only.

ui/client/components/AddProcessDialog.js
```diff
@@ -105,5 +122,21 @@ function mapState(state) {
...
 
+const nameAlreadyExists = (processes, subprocesses, name) => {
+  return processes.map(process => process.name).includes(name)
+    || subprocesses.map(subProcess => subProcess.name).includes(name)
+}
+
+const validations = [
+  {
+    isValid: (processes, subprocesses, name) => !_.isEmpty(name),
+    message: "Id can not be empty"
```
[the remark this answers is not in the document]
PPiedel: Done

ui/client/components/graph/node-modal/Input.js
PPiedel: Done

ui/client/components/AddProcessDialog.js
PPiedel: Done

ui/client/components/graph/NodeDetailsContent.js
PPiedel: Done

ui/client/components/graph/NodeDetailsContent.js
PPiedel: Done

PPiedel:
> You always repeat:
> 
> ```
> validators.map(validator => validator.isValid(fieldValue) ? null : <label key={fieldLabel + uuid4()} className='validation-label'>{validator.message}</label>
> ```
> 
> Create some helper method for this and reuse it.

I've created separate component "ValidationLabels"

ui/client/components/AddProcessDialog.js
```diff
@@ -105,5 +118,21 @@ function mapState(state) {
...
+}
+
+const validators = [
+  {
+    isValid: (processes, subprocesses, name) => notEmptyValidator.isValid(name),
+    message: notEmptyValidator.message
+  },
+  {
+    isValid: (processes, subprocesses, name) => !nameAlreadyExists(processes, subprocesses, name),
+    message: duplicate("Id")
```
witekw: Validations are always shown in the context of a field. If a field is the "Id" we do not need to repeat this information.

ui/client/components/ValidationLabels.js
```diff
@@ -0,0 +1,11 @@
...
+import {v4 as uuid4} from "uuid";
+
+export default class ValidationLabels extends React.Component {
+
+  render() {
+    const {validators, value} = this.props
+    return validators.map(validator => validator.isValid(value) ? null :
+      <label key={uuid4()} className='validation-label'>{validator.message}</label>)
+  }
+}
```
witekw: This component is to be used by node modal only, please move it to some specific subdirecotry in the hierarchy.

PPiedel: Actually it can also be reused by other components, but for now I've moved it to modals subdirectory

ui/client/components/AddProcessDialog.js
PPiedel: I've changed it to "This value is already taken"

ui/client/components/ValidationLabels.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import {v4 as uuid4} from "uuid";
+
+export default class ValidationLabels extends React.Component {
+
+  render() {
+    const {validators, value} = this.props
+    return validators.map(validator => validator.isValid(value) ? null :
+      <label key={uuid4()} className='validation-label'>{validator.message}</label>)
```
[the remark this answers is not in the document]
PPiedel: Ok, changed to span

PR #269, opened 2019-10-30
Copy global expressions after custom node with join transformation

gadomsky:
After custom node with join transformation global variables disappear. This is ugly fix - in future global variables should be moved out from ValidationContext

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
@@ -310,13 +310,15 @@ protected trait ProcessCompilerBase {
                                       (implicit nodeId: NodeId): ValidatedNel[ProcessCompilationError, ValidationContext] = {
       contextTransformationDefOpt match {
         case Some(transformation: ContextTransformationDef) =>
-          transformation.transform(validationContext)
+          // copying global variables because custom transformation may override them -> todo in ValidationContext
+          transformation.transform(validationContext).map(_.copy(globalVariables = validationContext.globalVariables))
         case Some(transformation: JoinContextTransformationDef) =>
           // TODO JOIN: better error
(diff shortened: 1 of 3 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -316,7 +316,7 @@ protected trait ProcessCompilerBase {
           val joinNode = node.cast[Join].getOrElse(throw new IllegalArgumentException(s"Should be used join element in node ${nodeId.id}"))
           // TODO JOIN: use correct contexts for branches
           val contexts = joinNode.branchParameters.groupBy(_.branchId).mapValuesNow(_ => validationContext)
-          transformation.transform(contexts)
+          transformation.transform(contexts).map(_.copy(globalVariables = validationContext.globalVariables))
```
arkadius: Can you put here note that it is because of TODO in `ValidationContext`?

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/CustomNodeValidationSpec.scala
```diff
@@ -226,7 +226,7 @@ class CustomNodeValidationSpec extends FunSuite with Matchers with OptionValues
     }
   }
 
-  test("valid process using context transformation api - union") {
+  test("valid process using context transformation api with global variables - union") {
```
arkadius: The rule of thumb should be: new bug - new test but ... we I can close eye to this ;-)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
gadomsky: done

PR #271, opened 2019-10-30
Simple Dicts, enums as dicts

arkadius:


ui/server/src/main/scala/pl/touk/nussknacker/ui/uiresolving/UIProcessResolving.scala
```diff
@@ -0,0 +1,42 @@
...
+import pl.touk.nussknacker.restmodel.validation.ValidationResults.ValidationResult
+import pl.touk.nussknacker.ui.process.marshall.ProcessConverter
+import pl.touk.nussknacker.ui.validation.ProcessValidation
+
+/**
+  * This class handles resolving of expression (e.g. dict labels resolved to keys) which should be done before
+  * validation of process created in UI and before process save.
+  * Also it handles "reverse" resolving process done before returning process to UI
+  */
+class UIProcessResolving(validation: ProcessValidation, dictSubstitutor: ProcessDictSubstitutor) {
```
mproch: I'd like to think a bit how this can be changed a bit...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/SpelDictTyper.scala
```diff
@@ -0,0 +1,64 @@
...
+import cats.data.ValidatedNel
+import org.springframework.expression.spel.SpelNode
+import org.springframework.expression.spel.ast.{Indexer, PropertyOrFieldReference, StringLiteral}
+import pl.touk.nussknacker.engine.api.expression.ExpressionParseError
+import pl.touk.nussknacker.engine.api.typed.typing.{DynamicTypedDict, StaticTypedDict, Typed, TypedDict, TypingResult}
+import pl.touk.nussknacker.engine.spel.ast
+
+trait DictTyper {
+
+  def typeDictValue(dict: TypedDict, node: SpelNode): ValidatedNel[ExpressionParseError, TypingResult]
```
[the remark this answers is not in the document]
arkadius: It is impossible, but dict.valueType has general type (TypingResult) - for enums it is TypedClass, for simple string dicts, Tagged, for user-defined dicts, anything. So looks like we can't be here more strict... any thoughts?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/ProcessDictSubstitutor.scala
```diff
@@ -0,0 +1,114 @@
+package pl.touk.nussknacker.engine.dict
```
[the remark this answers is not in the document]
arkadius: The idea was to be open for handling dicts in any expression type. It is why in this class is dispatching by expression language.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/processdetails.scala
```diff
@@ -77,7 +78,10 @@ object processdetails extends JavaTimeEncoders with JavaTimeDecoders {
   sealed trait ProcessShapeFetchStrategy[ProcessShape]
 
   object ProcessShapeFetchStrategy {
-    implicit case object Fetch extends ProcessShapeFetchStrategy[DisplayableProcess]
+    // TODO: Find places where FetchDisplayable is used and think about replacing it with FetchCanonical. We generally should use displayable representation
```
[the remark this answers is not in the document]
arkadius: This idea is still up to date. I would like to have kind of generic graph (like Displayable) in: database and import/export, but canonization is helpful in rewriting process, and it is why it is used in disabling of nodes or very similar task (dict value substitution). IMO as far as we have canonical process in db, we should use it in rewriting. When we switch to Displayable, we should add some general purpose rewriter (which maybe will partially do canonization) and which should handle both rewriting tasks (disabling and substitution). I'll will put some TODO in rewriter when it will be done... WDYT about this idea?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/SpelDictTyper.scala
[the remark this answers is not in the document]
arkadius: Ok, I get your point now. Added `type ValueType` to `TypedDict`

ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/PrettyValidationErrors.scala
```diff
@@ -24,7 +25,7 @@ object PrettyValidationErrors {
       case InvalidRootNode(_) => node("Invalid root node", "Process can start only from source node")
       case InvalidTailOfBranch(_) => node("Invalid end of process", "Process branch can only end with sink or processor")
 
-      case MissingParameters(params, ProcessCompilationError.ProcessNodeId) =>
+      case MissingParameters(params, NodeTypingInfo.ExceptionHandlerNodeId) =>
```
mproch: Hmm... I wonder if these are only ExceptionHandler params...

ui/server/src/main/scala/pl/touk/nussknacker/ui/uiresolving/UIProcessResolving.scala
```diff
@@ -0,0 +1,42 @@
...
```
mproch: let's add some TODO to refactor that :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -269,23 +272,32 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
     case Unknown => Valid(Unknown)
     case s: SingleTypingResult =>
       extractSingleProperty(e)(s)
-        .map(Valid(_)).getOrElse(invalid(s"There is no property '${e.getName}' in type: ${s.display}"))
     case TypedUnion(possible) =>
-      val l = possible.toList.flatMap(single => extractSingleProperty(e)(single))
+      val l = possible.toList.flatMap(single => extractSingleProperty(e)(single).toOption)
```
mproch: can we use some validation method and not convert toOption and vice versa?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/SpelDictTyper.scala
```diff
@@ -0,0 +1,67 @@
...
+            findKey(dict, key)
+          case _ =>
+            Invalid(ExpressionParseError(s"Illegal spel construction: ${node.toStringAST}. Dict should be indexed by a single key")).toValidatedNel
+        }
+      case pf: PropertyOrFieldReference  =>
+        findKey(dict, pf.getName)
+    }
+  }
+
+  private def findKey(dict: TypedDict, key: String) = {
```
mproch: maybe we should consider moving this logic to substitution? And let Typer focus solely on Typing?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/ProcessDictSubstitutor.scala
```diff
@@ -0,0 +1,94 @@
...
+    protected def findStaticDictReplacement(static: StaticTypedDict, indexerKeyValue: String): Option[String]
+
+  }
+
+  object LabelToKeyReplacingStrategy extends BaseReplacingStrategy {
+
+    override protected def findStaticDictReplacement(static: StaticTypedDict, indexerKeyValue: String): Option[String] = {
+      val replacement = static.keysByLabel.get(indexerKeyValue)
+      if (replacement.isEmpty) {
+        logger.warn(s"Can't find key for label: $indexerKeyValue in dict: $static. Probable change in dict definition. Will be used label in this place.")
```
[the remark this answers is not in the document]
arkadius: Is it quite ok to do this. We probably should add some warning that will be displayed in process, but it shouldn't be blocking error because it could be the situation that some key->label pair was missing for a while. With current implementation it will be fine, because key wouldn't be replaced by label during "reverse resolving" process and then also wouldn't be replaced by key during "resolving" process. At the end we will have the same "key" in saved process and it will be fine after key->label pair come back. 
All this logic will be more clear if we have something like ResolvedExpression instead of Expression in our model and there will be "original" (unresolved) expression and list of substitutions, but I don't want to change model on this phase of incubation of this feature...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/SpelDictTyper.scala
arkadius: Yeah I will do kind of such stuff when I will resolve TODO in TypedDict. It is connected.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
arkadius: I'm not sure if I understand your comment... As a user I don't want to see such detailed (accumulated) error that typer can't find property in type A because ..., type B because ..., etc. I just want to have simple message. It is why I skip this error part. It is not a walk-around, it is a part of business logic...

ui/server/src/main/scala/pl/touk/nussknacker/ui/uiresolving/UIProcessResolving.scala
arkadius: But where? I think that real TODO should be in each `Resource`, and in `CanonicalProcess` (it is there). Here we have only result of missing refactoring in other places ;-)

ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/PrettyValidationErrors.scala
arkadius: I also was wonder but than I found usage of `ProcessNodeId` and it was only in one place: `ProcessCompiler.compile(ref: ExceptionHandlerRef)`. I know that will be better to have `MissingParametersInExceptionHandler`, but during this feature there was a lot of refactors... Maybe we will do some in other one? ;-)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
mproch: yes, yes I understand the idea, but still - I'd expect some fancy method that'll handle that, like .sequence.leftMap(_.head) ;) 
but we can leave it as it is

PR #276, opened 2019-11-03
Fix default value decoding error while getting counts report

gskrobisz:


engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/SimpleInfluxClient.scala
```diff
@@ -44,12 +42,26 @@ class SimpleInfluxClient(config: InfluxConfig)(implicit backend: SttpBackend[Fut
   }
 }
 
-@JsonCodec(decodeOnly = true) case class InfluxResponse(results: List[InfluxResult] = List())
+case class InfluxResponse(results: List[InfluxResult] = List())
```
lciolecki: = List.empty or simple Nil

PR #285, opened 2019-11-06
Fix for regression bug after strict type checking: not acceptable single operand minus and accepting Number in place of BigDecimal

arkadius:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -166,7 +166,13 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
       case e: OpInc => checkSingleOperandArithmeticOperation(validationContext, e, current)
 
       case e: OpDivide => checkTwoOperandsArithmeticOperation(validationContext, e, current)(NumberTypesPromotionStrategy.ToCommonWidestType)
-      case e: OpMinus => checkTwoOperandsArithmeticOperation(validationContext, e, current)(NumberTypesPromotionStrategy.ToCommonWidestType)
+      case e: OpMinus => withTypedChildren {
+        case left :: right :: Nil if left.canBeSubclassOf(Typed[Number]) || right.canBeSubclassOf(Typed[Number]) => Valid(commonSupertypeFinder.commonSupertype(left, right)(NumberTypesPromotionStrategy.ToCommonWidestType))
+        case left :: right :: Nil => invalid(s"Operator '${e.getOperatorName}' used with mismatch types: ${left.display} and ${right.display}")
+        case left :: Nil if left.canBeSubclassOf(Typed[Number]) => Valid(left)
```
mproch: maybe we can take one child as special case and then move to checkTwoOperandsArithmeticOperation?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -49,7 +49,17 @@ private[typed] object CanBeSubclassDeterminer {
     //throw validation errors in this case. It's better to accept to much than too little
       superclassCandidate.params.zip(givenClass.params).forall(t => canBeSubclassOf(t._1, t._2) || canBeSubclassOf(t._2, t._1))
 
-    givenClass == superclassCandidate || ClassUtils.isAssignable(givenClass.klass, superclassCandidate.klass) && hasSameTypeParams
+    val canBeSubclass = givenClass == superclassCandidate || ClassUtils.isAssignable(givenClass.klass, superclassCandidate.klass) && hasSameTypeParams
+    canBeSubclass || canBeConvertedTo(givenClass, superclassCandidate)
+  }
+
+  private def canBeConvertedTo(givenClass: TypedClass, superclassCandidate: TypedClass): Boolean = {
```
mproch: maybe some better name? Also wonder what happens when we use non-spel expressions that don't have such type conversions?? let's add at least some comment about it?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
arkadius: I was afraid of using orElse but after deeper consideration it shouldn't break anything.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
arkadius: Can't find better name. This method check if some (safe) conversion is possible. In evaluation it is also named "conversion". What is your proposal for name? I've added some comment, can you check if it is all right?

PR #286, opened 2019-11-06
Show params validation error in the context of params when possible

### LINKED ISSUE #233: show params validation error in the context of params

PPiedel:
This resolves #233

### DIFF
--- a/ui/client/components/graph/node-modal/Input.js
+++ b/ui/client/components/graph/node-modal/Input.js
@@ -1,6 +1,7 @@
 import PropTypes from "prop-types";
 import React from "react";
 import ValidationLabels from "../../modals/ValidationLabels";
+import {allValid} from "../../../common/Validators";
 
 const Input = (props) => {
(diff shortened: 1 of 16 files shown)


ui/client/components/graph/node-modal/Input.js
```diff
@@ -13,12 +12,13 @@ const Input = (props) => {
                 <input
                     key={label}
                     type="text"
-                    className="node-input"
+                    className={validators.some(validator => validator.isValid(value) === false) ?
```
lciolecki: Do you can create some method at validators kind of isValid / hasErrors with checks each validator.isValid?

PPiedel: Done

PR #317, opened 2019-11-07
Ui improvements

### LINKED ISSUE #287: in creator panel click on category field should open it, now it works only if you click the text

### LINKED ISSUE #288: in drop/choose file field hovering over text hovers the text only. It should still hover whole field

### LINKED ISSUE #289: in creator panel disabled categories should not have cursor:pointer on them

### LINKED ISSUE #290: Selected menu item is not emphasised enough.
change alpha value in rgba to 0.3 instead of 0.1

### LINKED ISSUE #291: unchecked checkboxes appear weirdly invisible in node edit modal
checkbox needs to have same background color when selected and not selected

### LINKED ISSUE #292: remove focus outline in right panel icons
on right panel icons have focus outline which seems useless and also it covers error in properties icon when there is one

### LINKED ISSUE #293: Attachments "drop or choose a file" should be darker like all data input elements.
.espButton should have #333 instead of #4a4a4a as background color

PPiedel:
This fixes #291
This fixes #287 
This fixes #289
This fixes #290
This fixes #292
This fixes #293
This fixes #288

### DIFF
--- a/ui/client/app.styl
+++ b/ui/client/app.styl
@@ -127,6 +127,7 @@ processActivityList()
 .process-attachments
   cursor: default
   margin-bottom 10px
+  display grid
   .process-attachments-list
     processActivityList()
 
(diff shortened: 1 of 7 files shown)


witekw:
I think @fi- should look at this running in Chrome / Firefox and approve if he likes what he sees.

witekw:
#292 - in general there should be focus indication when using keyboard to move between buttons / elements

fi-:
2019-11-07 16_38_32-Nussknacker
this bug i caused by:
`.group-label
  display inline-block
  width 95% `

fi-:
#292 - we need to go back a step in this one and make the focus visible again, because we need to allow users to navigate using TAB.
We will fix not-visible properties icon some other way.

fi-:
#291 - this looks good on chrome but on firefox it looks like this:
2019-11-07 16_49_51-Nussknacker

PPiedel:
> #292 - in general there should be focus indication when using keyboard to move between buttons / elements
> 
> All the rest looks good on Firefox, I do not like spacing here:
> 
> image

Done

PPiedel:
> #291 - this looks good on chrome but on firefox it looks like this:
> 2019-11-07 16_49_51-Nussknacker

Done

PPiedel:
> #292 - we need to go back a step in this one and make the focus visible again, because we need to allow users to navigate using TAB.
> We will fix not-visible properties icon some other way.

Done

PPiedel:
> 2019-11-07 16_38_32-Nussknacker
> this bug i caused by:
> 
> ```
> .group-label
>   display inline-block
>   width 95%
> ```

Done

lciolecki:
There are conflicts

fi-:
I reviewed the changes in both chrome and firefox. All looks good!

PR #322, opened 2019-11-07
Feature: Move all security data to enigne security-api

lciolecki:


build.sbt
```diff
@@ -542,20 +543,23 @@ lazy val api = (project in engine("api")).
...
   settings(commonSettings).
   settings(
-    name := "nussknacker-security-api",
+    name := "nussknacker-security",
     libraryDependencies ++= {
       Seq(
         "org.scalatest" %% "scalatest" % scalaTestV % "test",
+        "org.mindrot" % "jbcrypt" % jbcryptV,
         "com.typesafe.akka" %% "akka-http" % akkaHttpV force(),
+        "com.typesafe.akka" %% "akka-stream" % akkaV force(),
```
arkadius: I wonder if we can have all stuff provided in our ui-server in scope "provided" and if this scope will be respected in library using this module... Thanks to this we will be sure that during assembling it won't go to the jar again...

build.sbt
```diff
@@ -542,20 +543,23 @@ lazy val api = (project in engine("api")).
...
+        "de.heikoseeberger" %% "akka-http-circe" % akkaHttpCirceV,
         "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpV % "test" force(),
         "com.typesafe" % "config" % configV
       )
     }
   )
-  .dependsOn(util)
+  .dependsOn(util, httpUtils)
```
arkadius: The same notice about "provided" scope here.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/LoggedUser.scala
```diff
@@ -1,7 +1,7 @@
-package pl.touk.nussknacker.ui.security.api
```
arkadius: What do you thing about having all this api stuff in api package and other implementation have in security/module e.g. security/oauth?

PR #324, opened 2019-11-07
Feature: security module

lciolecki:


build.sbt
```diff
@@ -561,20 +562,23 @@ lazy val api = (project in engine("api")).
...
   settings(commonSettings).
   settings(
-    name := "nussknacker-security-api",
+    name := "nussknacker-security",
     libraryDependencies ++= {
       Seq(
         "org.scalatest" %% "scalatest" % scalaTestV % "test",
+        "org.mindrot" % "jbcrypt" % jbcryptV,
         "com.typesafe.akka" %% "akka-http" % akkaHttpV force(),
+        "com.typesafe.akka" %% "akka-stream" % akkaV force(),
```
arkadius: Can all this stuff provided to be sure that in final security extension pack, wouldn't be those libs? Will it work with transitive dependencies?

build.sbt
```diff
@@ -561,20 +562,23 @@ lazy val api = (project in engine("api")).
...
+        "de.heikoseeberger" %% "akka-http-circe" % akkaHttpCirceV,
         "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpV % "test" force(),
         "com.typesafe" % "config" % configV
       )
     }
   )
-  .dependsOn(util)
+  .dependsOn(util, httpUtils)
```
arkadius: The same as above

lciolecki: Changed.

build.sbt
lciolecki: Good idea! Changed.

PR #336, opened 2019-11-08
Improvements: Normalize users config

lciolecki:


docs/Security.md
```diff
@@ -31,20 +31,33 @@ authentication: {
...
 users: [
   {
-    id: "user1"
-    password: "pass"
-    categoryPermissions: {
-      "Category1": ["Read", "Deploy"]
-      "Category2": ["Read", "Write"]
-    }
-    globalPermissions: ["AdminTab"]
+    identity: "user1"
```
arkadius: I think that we need to start updating `BreakingChanges.md`...

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/RulesSet.scala
```diff
@@ -35,8 +34,10 @@ class RulesSet(rules: List[OAuth2ConfigRule], allCategories: List[String]) {
 }
 
 object RulesSet {
-  def getOnlyMatchingRules(roles: List[String], rules: List[OAuth2ConfigRule], allCategories: List[String]): RulesSet = {
+  def getOnlyMatchingRules(roles: List[String], rules: List[ConfigRule], allCategories: List[String]): RulesSet = {
     val filtered = rules.filter(rule => roles.map(_.toLowerCase).contains(rule.role.toLowerCase))
     new RulesSet(filtered, allCategories)
   }
+
+  def apply(rules: List[ConfigRule], allCategories: List[String]): RulesSet = new RulesSet(rules, allCategories)
```
arkadius: Is it needed? `getOnlyMatchingRules` is not the only usage? There u use constructor.

docs/Security.md
lciolecki: In future.. :)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/RulesSet.scala
lciolecki: You'r right - removed.

PR #340, opened 2019-11-08
Handle management sync failure

arkadius:


ui/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerApp.scala
```diff
@@ -116,7 +116,8 @@ object NussknackerApp extends App with Directives with LazyLogging {
 
     Initialization.init(modelData.mapValues(_.migrations), db, environment, config.getAs[Map[String, String]]("customProcesses"))
 
-    val managementActor = ManagementActor(environment, managers, processRepository, deploymentProcessRepository, subprocessResolver)
+    val managementActor = system.actorOf(
```
mproch: is it needed now?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -1,40 +1,39 @@
...
 import pl.touk.nussknacker.ui.security.api.{LoggedUser, NussknackerInternalUser}
-import pl.touk.nussknacker.ui.util.CatsSyntax
+import pl.touk.nussknacker.ui.util.{CatsSyntax, FailurePropagatingActor}
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
 object ManagementActor {
-  def apply(environment: String,
+  def props(environment: String,
```
mproch: is it needed now? Or we can just extend trait?

ui/server/src/main/scala/pl/touk/nussknacker/ui/util/FailurePropagatingActor.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.ui.util
+
+import akka.actor.{Actor, Status}
+
+// see: https://medium.com/@linda0511ny/error-handling-in-akka-actor-with-future-ded3da0579dd
+trait FailurePropagatingActor extends Actor {
```
mproch: :+1:

ui/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerApp.scala
arkadius: It is a good practise to have a `Props` and create actors in one place in application.

PR #344, opened 2019-11-13
Change and move zoom in-out icons as well as arrow icons

### LINKED ISSUE #188: show process name on the diagram editing view
Consider if process name edit should be in diagram edit view as well (permission issues?).

### LINKED ISSUE #296: remove zoom in/zoom-out icons or move them to new "view" category
zoom in and zoom out icons shouldn't be in "process" group.
Since we are considering putting some tools' icons in the node area (most of them appearing as context actions) we will start by moving zoom in and out icons first.
It should look like this:
ui-improvements1
Currently there are arrow icons that hide left and right panels that are in conflict with zoomin/out icons. We plan to remove them later as soon as we put right panel on grid and make it scalable. For now let's move these arrows to the bottom of the screen. 
Here are the zoom icons:
[zoomicons.zip](https://github.com/TouK/nussknacker/files/3836130/zoomicons.zip)

### LINKED ISSUE #307: node selection on the graph hides warning indication

PPiedel:
This resolves #296
This resolves #307
This resolves #188

### DIFF
--- a/ui/client/assets/img/arrows/arrow-left.svg
+++ b/ui/client/assets/img/arrows/arrow-left.svg
@@ -1,9 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- Generator: Adobe Illustrator 21.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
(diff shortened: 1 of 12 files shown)


PPiedel:
Screenshot from 2019-11-13 11-14-48

witekw:
Zoom icons - cannot get used to it, icons are ligter than ones we have, but its fine
Node selecion hides error - ok
Process name - mixed feelings, we do not handle overlapping with diagram in any way, we introduce new color, has Filip seen it?

I've notice panels "handle bars", it is better, though arrows should switch direction at the end of movement.

PR #345, opened 2019-11-13
Enhanced aggregates - more generic interface, sample aggregations

mproch:


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
```diff
@@ -0,0 +1,127 @@
...
+
+import scala.collection.immutable.TreeMap
+import scala.concurrent.duration.Duration
+
+object AggregateTransformer extends CustomStreamTransformer {
+
+  @MethodToInvoke(returnType = classOf[AnyRef])
+  def execute(@ParamName("keyBy") keyBy: LazyParameter[String],
+              @ParamName("length") length: String,
+              @ParamName("aggregator") aggregator: Aggregator,
```
arkadius: How do you want to pass `aggregator`? Some map of all possible values in global variables? WDYT about replacing it with java Enum and make it more standardized and ready for further better enum handling?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
```diff
@@ -0,0 +1,127 @@
...
+import pl.touk.nussknacker.engine.api.context.ContextTransformation
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.NodeId
+import pl.touk.nussknacker.engine.api.{Context => NkContext, _}
+import pl.touk.nussknacker.engine.flink.api.process._
+import pl.touk.nussknacker.engine.flink.api.state.LatelyEvictableStateFunction
+
+import scala.collection.immutable.TreeMap
+import scala.concurrent.duration.Duration
+
+object AggregateTransformer extends CustomStreamTransformer {
```
arkadius: I wonder if we can think about making those all generic transformers unified and think about them as one is the special case for another. Proposed `AggregateTransformer` looks like most generic one. 
1. One thing that is missing for me is filtering (`transformWhen` in `TransformStateTransformer`). How do you handle the scenario: You want to handle stream of events, based on part of them you want to have aggregation and for other part you want to just read value of state? 
2. If we decide to go this way, we should have the same naming for window time, key be and so on.
3. Mayby those special cases like `TransformStateTransformer`, `PreviousValueTransformer` won't be needed then?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
```diff
@@ -0,0 +1,127 @@
...
+    ValueWithContext((keyInterpreter(value), valueInterpreter(value)), value)
+  }
+}
+
+
+//TODO: figure out if it's more convenient/faster to just use SlidingWindow with 60s slide and appropriate trigger?
+class AggregatorFunction[T <: AnyRef](aggregator: Aggregator, lengthInMillis: Long, nodeId: String)
+  extends LatelyEvictableStateFunction[ValueWithContext[(String, AnyRef)], ValueWithContext[Any], TreeMap[Long, AnyRef]] {
+
+  private val minimalResolutionMs = 60000L
```
arkadius: Why you need this?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -0,0 +1,115 @@
...
+    override type Element = Number
+
+    override def zero: Number = Double.MaxValue
+
+    override def add: (Number, Number) => Number = (n1, n2) => Math.min(n1.doubleValue(), n2.doubleValue())
+
+    override def zeroType: TypingResult = Typed[Number]
+  }
+
+  object ListAggregator extends Aggregator {
```
arkadius: What about adding First, Last? that will contain only one element?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/Aggregator.scala
```diff
@@ -0,0 +1,44 @@
...
+ */
+trait Aggregator extends AggregateFunction[Any, Any, Any] {
+
+  type Aggregate <: AnyRef
+
+  type Element <: AnyRef
+
+  def zero: Aggregate
+
+  def add: (Element, Aggregate) => Aggregate
```
arkadius: Why this method, `merge` and `result` return function instead of simple handle `Element` and `Aggregate` as a method's parameters?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -0,0 +1,115 @@
...
+    override def merge: (Aggregate, Aggregate) => Aggregate = _ ++ _
+
+    override def result: Aggregate => Any = _.asJava
+
+    override def computeOutputType(input: TypingResult): Validated[String, TypingResult]
+      = Valid(TypedClass(classOf[java.util.List[_]], List(input)))
+
+  }
+
+  class MapAggregator(fields: java.util.Map[String, Aggregator]) extends Aggregator {
```
arkadius: Can you give some example how it works? You can have max on some field and min on some other? How it handle null in values? Will omit aggregations?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
mproch: to reduce the number of values we are storing. It's kind of window slide.

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
mproch: I added it

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/Aggregator.scala
mproch: true, I changed it

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
mproch: 1. I think I can add it to this function
2. I'll look at that
3. PreviousValueTransformer is kind of different as it has no notion of time and can be more efficient I think than those based on windows

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
```diff
@@ -0,0 +1,127 @@
...
+import pl.touk.nussknacker.engine.flink.api.state.LatelyEvictableStateFunction
+
+import scala.collection.immutable.TreeMap
+import scala.concurrent.duration.Duration
+
+object AggregateTransformer extends CustomStreamTransformer {
+
+  @MethodToInvoke(returnType = classOf[AnyRef])
+  def execute(@ParamName("keyBy") keyBy: LazyParameter[String],
+              @ParamName("length") length: String,
```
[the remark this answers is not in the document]
mproch: I refactored it a bit

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregateTransformer.scala
mproch: I refactored it a bit, to be able to separate
- aggregators
- functions/transformers 
- UI/API - that is, how do we present it to user, what is fixed and what can be changed

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
mproch: I'll add some samples and tests

mproch: comments/test added, I added TODO for null handling, I think some more general consideration would be needed - e.g. can we implement #input.complexField.?field1? :D

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/sampleTransformers.scala
```diff
@@ -0,0 +1,46 @@
...
+  object SimpleSlidingAggregateTransformer extends CustomStreamTransformer {
+
+    @MethodToInvoke(returnType = classOf[AnyRef])
+    def execute(@ParamName("keyBy") keyBy: LazyParameter[String],
+                @PossibleValues(Array("Max", "Min", "ApproximateSetCardinality", "Set")) @ParamName("aggregator") aggregatorType: String,
+                @ParamName("aggregateBy") aggregateBy: LazyParameter[AnyRef],
+                @ParamName("windowLengthInSeconds") length: Long,
+                @OutputVariableName variableName: String)(implicit nodeId: NodeId): ContextTransformation = {
+      val windowDuration = Duration(length, TimeUnit.SECONDS)
+      transformers.tumblingTransformer(keyBy, aggregateBy, toAggregator(aggregatorType), windowDuration, variableName)
```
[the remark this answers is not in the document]
mproch: yes, thanks :)

PR #350, opened 2019-11-14
Improvements: Move buildInfo to public resources

lciolecki:


ui/server/src/test/scala/pl/touk/nussknacker/ui/api/AppResourcesSpec.scala
```diff
@@ -98,25 +90,6 @@ class AppResourcesSpec extends FunSuite with ScalatestRouteTest
     }
   }
 
-  test("it should return global config") {
```
arkadius: Why u removed that?

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/PublicResourcesSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+    override def buildInfo(): Map[String, String] = Map("fromModel" -> "value1")
+  }
+
+  val modelData: LocalModelData = LocalModelData(ConfigFactory.empty(), creatorWithBuildInfo)
+  val globalConfig: Map[String, String] = Map("testConfig" -> "testValue", "otherConfig" -> "otherValue")
+  val resources: PublicResources = new PublicResources(ConfigFactory.parseMap(Collections.singletonMap("globalBuildInfo", globalConfig.asJava)), Map("test1" -> modelData))
+
+  def routeResult(url: String): RouteTestResult = Get(url) ~> withoutPermissions(resources)
+
+  test("it should return buildInfo") {
```
arkadius: I know that you named class "Public..." but IMO would be better if add more descriptive info in test e.g. "It should return build info without authentication"

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/PublicResources.scala
```diff
@@ -0,0 +1,34 @@
...
+import scala.concurrent.ExecutionContext
+
+class PublicResources(config: Config, modelData: Map[ProcessingType, ModelData])(implicit ec: ExecutionContext)
+  extends Directives with FailFastCirceSupport with LazyLogging with RouteWithoutUser with SecurityDirectives {
+
+  import io.circe.syntax._
+  import net.ceedubs.ficus.Ficus._
+
+  def route(): Route = {
+    pathPrefix("app") {
```
arkadius: Now we have two resources prefixed by "app"? How it supposed to work? IMO we should create one AppResources and create it from public and non-public resource paths... or create new resource prefix `/public`. The first option I think is nicer...

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/AppResourcesSpec.scala
lciolecki: It was moved to  test("it should return buildInfo"). And this test doesn't check global config.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/PublicResourcesSpec.scala
lciolecki: Changed.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/PublicResourcesSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+  import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+  import scala.collection.JavaConverters._
+  import io.circe.syntax._
+
+  lazy val creatorWithBuildInfo: EmptyProcessConfigCreator = new EmptyProcessConfigCreator {
+    override def buildInfo(): Map[String, String] = Map("fromModel" -> "value1")
+  }
+
+  val modelData: LocalModelData = LocalModelData(ConfigFactory.empty(), creatorWithBuildInfo)
+  val globalConfig: Map[String, String] = Map("testConfig" -> "testValue", "otherConfig" -> "otherValue")
```
arkadius: I think that it should be a part of test below instead on class level. It is Given part in Then you check if result is equals to this Given config... regards to BDD

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/PublicResources.scala
lciolecki: There is an joining paths pattern. Hmm I can extends from RouteWithUser and RouteWithoutUser, and implements two kind of method rout- one with user second without.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/PublicResourcesSpec.scala
lciolecki: I will rollback tests. And change it at original AppResourcesSpec.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -26,22 +26,27 @@ class AppResources(config: Config,
                    processRepository: FetchingProcessRepository[Future],
                    processValidation: ProcessValidation,
                    jobStatusService: JobStatusService)(implicit ec: ExecutionContext)
-  extends Directives with FailFastCirceSupport with LazyLogging with RouteWithUser with SecurityDirectives {
+  extends Directives with FailFastCirceSupport with LazyLogging with RouteWithUser with RouteWithoutUser with SecurityDirectives {
+
+  //Public app resources
+  def route(): Route = pathPrefix("app") {
```
arkadius: maybe let's name this routes: publicRoute, securedRoute instead of giving comment. I know that there is implicit in second version but in invocation it looks the same... Also still both routes have the same prefix. Hot it supposed to work?

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/PublicResources.scala
arkadius: I'm not convinced that we want this joining pattern to use... Now you have two paths prefix that someone might change separately introducing regression bug. Also path hierarchy is not visible. IMO would be better to have one route() but inside of this have two private route parts...

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
lciolecki: Akka http has join paths - at end everything is works fine.

lciolecki: Changed. Now we have publicPublic and securedRoute.

arkadius: Now comment is not needed, but we can leave it just in case ;-)

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/PublicResources.scala
lciolecki: I'v added some comments.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -28,7 +28,7 @@ class AppResources(config: Config,
                    jobStatusService: JobStatusService)(implicit ec: ExecutionContext)
   extends Directives with FailFastCirceSupport with LazyLogging with RouteWithUser with RouteWithoutUser with SecurityDirectives {
 
-  //Public app resources
+  //We use duplicate pathPrefix("app") code - look at comment in NussknackerApp where routes are creating
```
arkadius: Can be: "we duplicate ... code" or "we use dulicate*d* ... code" The first one is IMO better .
Also should be: "where routes are creat*ed*"

arkadius:
There are still some little grammatical fixes in comments, but other things LGTM.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
lciolecki: Removed ;)

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
lciolecki: Ups.. Changed.

PR #364, opened 2019-11-14
[Issue 268] UI validation - add field name to node's expression validation

nadberezny:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSubGraphCompiler.scala
```diff
@@ -76,7 +76,7 @@ class PartSubGraphCompiler(protected val classLoader: ClassLoader,
             isDisabled = f.isDisabled.contains(true)))
 
       case splittednode.SwitchNode(graph.node.Switch(id, expression, exprVal, _), nexts, defaultNext) =>
-        val (expressionTyping, validatedExpression) = compile(expression, None, ctx, Unknown)
+        val (expressionTyping, validatedExpression) = compile(expression, Some("Expression"), ctx, Unknown)
```
arkadius: 1. What about my notice on rocket about using `NodeTypingInfo.DefaultExpressionId`? IMO we need to have one way to identify expressions and prefixing with special character prevent us from mixing expression with parameter with name `Expression` - we can have it in e.g. sinks.
2. Is Option is still needed after those changes? If so, can we split logic in some point and have method with explicit name handling missing identifier? Now someone can not intentionally put None in new invocation.
3. If we don't have Option, what do you think about using `NodeExpressionId` and remove it from this parameters lists. IMO we have too many params in those methods and it is easy to put somewhere Uknown, None, ctx.empty, just to start it working...

PR #368, opened 2019-11-14
ExpressionSuggester async

arkadius:


ui/client/components/graph/ExpressionSuggester.js
```diff
@@ -35,20 +35,21 @@ export default class ExpressionSuggester {
...
       const allVariablesWithClazzRefs = _.map(variables, (val, key) => {
         return {'methodName': key, 'refClazz': val}
       })
-      return this._filterSuggestionsForInput(allVariablesWithClazzRefs, value)
-    }
-    else {
-      return []
+      result = this._filterSuggestionsForInput(allVariablesWithClazzRefs, value)
+    } else {
+      result = []
```
lciolecki: You can do  at start ` let result = []` and remove this else.  Or better in each if do simply return and at end return [];

arkadius: I can't do return in each if/else because need to wrap return in `Promise` but I can init variable with empty array :+1:

PR #370, opened 2019-11-14
Dict suggestions - backend

arkadius:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompiledProcess.scala
```diff
@@ -24,9 +25,12 @@ object CompiledProcess {
              ): ValidatedNel[ProcessCompilationError, CompiledProcess] = {
     val servicesDefs = definitions.services
 
-    val expressionCompiler = ExpressionCompiler.withOptimization(userCodeClassLoader, definitions.expressionConfig)
+    val dictRegistryFactory = DictServicesFactoryLoader.justOne(userCodeClassLoader)
```
mproch: ugh... I *don't* think this should be here... We shouldn't mess with classloaders here...

arkadius: Why? Which level will be better? Near creating of `ProcessConfigCreator`?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
```diff
@@ -0,0 +1,41 @@
...
+  def value(key: String): Any
+
+}
+
+/**
+ * If this instance wil be used in global variables, will be typed to TypedDict.
+ */
+case class DictInstance(dictId: String, definition: DictDefinition) {
+
+  def valueType: SingleTypingResult = definition.valueType(dictId)
```
witekw: Does DictInstance "contain" exactly one dictionary (map)? If so, it should not require dictId to return its type.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictDefinition.scala
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * It is DictDefinition which contains embedded label <> key transformation.
+ */
+trait StaticDictDefinition extends DictDefinition {
+
+  def labelByKey: Map[String, String]
+
+  lazy val keyByLabel: Map[String, String] = labelByKey.map(_.swap)
+
```
witekw: Do we check somewhere if the map is invertible?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
arkadius: `DictInstance` not need `dictId` to return its value type. Maybe you mean `DictDefinition`? It maybe confusing but our "instances" are objects that we put to global variables map. `DictIntance` refer to `DictDefinition` and `DictDefinition` is kind of abstract type. Concrete class is `DictDefinition` + dictId. Take a look at `TestProcessConfigCreator` to see how it used.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictDefinition.scala
arkadius: No, we don't. I can put some assertion for example that `labels` are unique. WDYT?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompiledProcess.scala
mproch: ok, for now let's just exctract creating registry to a method, to make it more visible, ok?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictDefinition.scala
```diff
@@ -0,0 +1,78 @@
...
+import pl.touk.nussknacker.engine.api.dict.{DictDefinition, ReturningKeyWithoutTransformation}
+import pl.touk.nussknacker.engine.api.typed.ClazzRef
+import pl.touk.nussknacker.engine.api.typed.typing.{SingleTypingResult, TypedClass}
+
+import scala.reflect.ClassTag
+
+/**
+ * It is DictDefinition which contains embedded label <> key transformation.
+ */
+trait StaticDictDefinition extends DictDefinition {
```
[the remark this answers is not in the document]
mproch: ok, now I understand... I think it makes the whole thing even more complex, but there is some convenience in it...

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
```diff
@@ -0,0 +1,44 @@
...
+  def valueType: SingleTypingResult = definition.valueType
+
+  def value(key: String): Any = definition.value(key)
+
+}
+
+/**
+ * It is helper mixin when value is exact as key (without additional transformation)
+ */
+trait ReturningKeyWithoutTransformation { self: DictDefinition =>
```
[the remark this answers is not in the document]
mproch: oook... I kind of understand what you want to achieve now...
I think we should maybe add some helpers or checks (in ProcessConfigCreation) to make sure that dictIds are the same as global var names? Because now it can be *very* confusing I think ;)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+import pl.touk.nussknacker.engine.api.dict.{DictDefinition, DictRegistry}
+import pl.touk.nussknacker.engine.api.dict.DictRegistry.{DictEntryWithKeyNotExists, DictEntryWithLabelNotExists, DictNotDeclared}
+
+trait StaticDictRegistry extends DictRegistry {
+
+  protected def declarations: Map[String, DictDefinition]
+
+  override def keyByLabel(dictId: String, label: String): Validated[DictRegistry.DictLookupError, String] = {
+    declarations.get(dictId).map(Valid(_)).getOrElse(Invalid(DictNotDeclared(dictId))).andThen {
+      case static: StaticDictDefinition =>
```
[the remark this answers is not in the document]
mproch: yes, I kind of understand this, I more meant:
        static.keyByLabel.get(label).map(key => Valid(key))
          .getOrElse(Invalid(DictEntryWithLabelNotExists(dictId, label, Some(static.keyByLabel.keys.toList))))
==> why this is not part of static?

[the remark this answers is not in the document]
arkadius: It can be part of `StaticDictDefinition` but then user's implementation which will check the same but in some database/service will do it in `DictRegistry` or pass fat, managed object like connection pool as a method's parameter to his own `DictDefinition`. For first option we will have inconsistency (part of logic in `DictRegistry`, part of logic in  `StaticDictDefinition`. For the second one we will pass managed object as a parameter that is a bad design. Or maybe I still misunderstand your point?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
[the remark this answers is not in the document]
arkadius: There are three things:
1. duplication of dictId between DictInstance and SimpleDictDefinition -> I've fixed it by adding dictId to valueType as a parameter
2. duplication of dictId between DictInstance and `ExpressionConfig.dictionaries` (definitions are in map like sources etc and we have there categories so it will be hard to change that). I can add TODO that after category refactor we can move dictId to `DictDefinition` and then this duplication will disappear.
3. DictInstance's dictId vs global variable name - IMO it's not a duplication, because developer can organize dict instance's in many ways e.g. he or she can put it to nested variable "dicts" as a properties.

Regards to 3. how to assert 2.? One idea that I have is to add factory method to `DictInstance` which will get `Map[String, WithCategories[DictDefinition]]` and `dictId` and create `DictInstance`. It will be ugly IMO :-( Any thoughts?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictDefinition.scala
witekw: +1

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
witekw: ok, i think I get it

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/CompiledProcess.scala
arkadius: extracted

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/static/StaticDictDefinition.scala
arkadius: added

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/dict/DictDefinition.scala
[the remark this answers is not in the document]
mproch: let's add TODO...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
```diff
@@ -62,7 +64,11 @@ trait ModelData extends ConfigCreatorSignalDispatcher {
 
   lazy val processDefinition: ProcessDefinition[ObjectDefinition] = ProcessDefinitionExtractor.toObjectDefinition(processWithObjectsDefinition)
 
-  lazy val validator: ProcessValidator = ProcessValidator.default(processWithObjectsDefinition, modelClassLoader.classLoader)
+  // We can create restrictive services because I assume that ModelData is created once on UI side
```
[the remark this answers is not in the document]
mproch: let's leave it for further refactor...

PR #376, opened 2019-11-15
More deterministic tests (PatientScalaFutures)

arkadius:
- (Very)PatientScalaFutures
- global execution context with more threads
Refactoring:
- extracted test-utils module with logging and scalatest detault configuration


engine/test-util/src/main/resources/reference.conf
```diff
@@ -0,0 +1,5 @@
+scala.concurrent.context {
```
mproch: :+1:

engine/test-util/src/main/resources/logback-test.xml
```diff
@@ -0,0 +1,29 @@
...
+        <encoder>
+            <Pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</Pattern>
+        </encoder>
+    </appender>
+
+    <root level="WARN">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+    <!-- let's keep here INFO - if it will produce too many logs, we should rather think if our INFO logs are too much verbose... -->
```
mproch: we can try, but OTOH in tests deployment/normal work ratio is MUCH different than in prod...

engine/test-util/src/main/resources/logback-test.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mproch: I wonder if it would be a good idea to name it logback.xml,
so that we can override this file in some modules without warnings (e.g. when we test deployments we WILL have more INFO logs and we may not want them in this module?

arkadius: It will work with logback-test.xml. Configuration is taken based on order in classpath and order in classpath (I assume) is linearized based on sbt dependencies. I've tested it and it works.

PR #382, opened 2019-11-17
[Issue 268] UI - better node expression validation result

nadberezny:
Pass expression in field name in validation result, and display expression errors in scope of expression field

### DIFF
--- a/engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
+++ b/engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
@@ -250,7 +250,7 @@ class CustomNodeProcessSpec extends FunSuite with Matchers {
 
     val thrown = the [IllegalArgumentException] thrownBy processInvoker.invoke(process, List.empty)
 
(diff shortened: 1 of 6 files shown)


nadberezny:
@arkadius 
1. What about my notice on rocket about using NodeTypingInfo.DefaultExpressionId? IMO we need to have one way to identify expressions and prefixing with special character prevent us from mixing expression with parameter with name Expression - we can have it in e.g. sinks.
2. Is Option is still needed after those changes? If so, can we split logic in some point and have method with explicit name handling missing identifier? Now someone can not intentionally put None in new invocation.

Ad. 1
Done

Ad. 2
Yes, we still need Option to handle validation of an empty graph (without nodes)
As far as splitting validation logic is concerned, can't agree more, but I think, it would be good to address this issue in a seprarate PR.

arkadius:
> Ad. 1
> Done
:+1: 
 
> Ad. 2
> Yes, we still need Option to handle validation of an empty graph (without nodes)
> As far as splitting validation logic is concerned, can't agree more, but I think, it would be good to address this issue in a seprarate PR.

Can you put TODO in `compile` method that we now pass None very rare and it should be splitted to two methods with explicit name like `compileExpressionWithoutId`?

3. I think that we still incorrectly handle branch parameters, am I right? Or this `branchParameterExpressionId` logic is correctly duplicated on UI side?

nadberezny:
@arkadius thanks for the review!
TODO added

PR #383, opened 2019-11-17
JS static typing with Flow:

nadberezny:
1. Add init config
2. Type Validators.js

### DIFF
--- a/ui/client/package.json
+++ b/ui/client/package.json
@@ -6,9 +6,10 @@
   "license": "UNLICENSED",
   "private": true,
   "scripts": {
+    "build": "NODE_ENV=production webpack -p --progress && cp -R assets/icons/license dist",
+    "flow": "flow",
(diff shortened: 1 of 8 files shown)


ui/client/package.json
```diff
@@ -6,9 +6,10 @@
   "license": "UNLICENSED",
   "private": true,
   "scripts": {
+    "build": "NODE_ENV=production webpack -p --progress && cp -R assets/icons/license dist",
+    "flow": "flow",
     "start": "PROXY_API_DOMAIN=http://localhost:8081/api webpack-dev-server",
     "start-backend-demo": "PROXY_API_DOMAIN=https://staging.nussknacker.io/api webpack-dev-server",
-    "build": "NODE_ENV=production webpack -p --progress && cp -R assets/icons/license dist",
     "test": "./node_modules/karma/bin/karma start --single-run"
```
lciolecki: Does flow run with tests? It will be good to tests flow syntax at test or we can do it after tests.

PR #390, opened 2019-11-18
Listener api

gadomsky:


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/process/deployment/package.scala
```diff
@@ -0,0 +1,12 @@
+package pl.touk.nussknacker.restmodel.process
+
+package object deployment {
```
mproch: why is it in restmodel?

gadomsky:
> So the idea is that currently listeners do not have direct access to repositories?

Yes, but in the next PR I will introduce separete module with interfaces with such an access

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/process/deployment/package.scala
gadomsky: It's not any more :)

build.sbt
```diff
@@ -662,7 +662,14 @@ lazy val restmodel = (project in file("ui/restmodel"))
       "io.circe" %% "circe-java8" % circeV
     )
   )
-  .dependsOn(interpreter, testUtil % "test")
+  .dependsOn(api, interpreter, security, testUtil % "test")
+
+lazy val listenerApi = (project in file("ui/listner-api"))
```
arkadius: typo: ui/listEner-api

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/Implicits.scala
```diff
@@ -22,4 +23,11 @@ object Implicits {
       else Some(s)
     }
   }
+
+  implicit class RichFuture[A](future: Future[A]) {
+    def effect(f: () => Unit)(implicit ec: ExecutionContext): Future[A] = for {
```
arkadius: Maybe `withSideEffect` or at least `withEffect` - to be visible that it doesn't only add effect but also returns Future

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.listner
```
arkadius: What about moving it to `pl.touk.nussknacker.ui.listener`? Because I assume that that are ui events... Also typo: listner => listEner

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import pl.touk.nussknacker.engine.api.process.ProcessName
+
+sealed trait ChangeEvent
```
arkadius: maybe `ProcessChangeEvent` to make it visible that all events are in context of process? Also shouldn't it be processId everywhere? I would like to write listener implementation that will partition all events by some identifier to ensure that changes will have correct order for this identifier.

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
```diff
@@ -0,0 +1,17 @@
...
+
+object ChangeEvent {
+  final case class OnArchived(processName: ProcessName) extends ChangeEvent
+  final case class OnCategoryChanged(processName: ProcessName, category: String) extends ChangeEvent
```
arkadius: Can't we have also old category?

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
```diff
@@ -0,0 +1,17 @@
...
+  final case class OnDeleted(processName: ProcessName) extends ChangeEvent
+  final case class OnDeploySuccess(processName: ProcessName) extends ChangeEvent
+  final case class OnDeployCancelled(processName: ProcessName) extends ChangeEvent
+  final case class OnDeployFailed(processName: ProcessName) extends ChangeEvent
```
arkadius: Can't we have also some failure info?

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
```diff
@@ -0,0 +1,17 @@
...
+  final case class OnRenamed(oldName: ProcessName, newName: ProcessName) extends ChangeEvent
+  final case class OnSaved(processName: ProcessName) extends ChangeEvent
```
arkadius: I think that version will be necessary here... How someone will now which version should pull after this event? Also it will be necessary in other events like `OnDeploy*`

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ListenerManagementFactory.scala
```diff
@@ -0,0 +1,29 @@
...
+  def serviceLoader(classLoader: ClassLoader): ListenerManagementFactory = {
+    aggregate(ScalaServiceLoader.load[ListenerManagementFactory](classLoader): _*)
+  }
+
+  def aggregate(factories: ListenerManagementFactory*): ListenerManagementFactory = new ListenerManagementFactory {
+    override def create(config: Config): ListenerManagement = {
+      val listeners = factories.map(_.create(config))
+      new ListenerManagement {
+        override def handler(event: ChangeEvent)(implicit ec: ExecutionContext, user: LoggedUser): Unit = {
+          listeners.foreach(_.handler(event))
```
arkadius: What about error handling here or rather on `ListenerManagement` level? Failures in some listener handling shouldn't have influence on other listeners... Also we need to have some logging of failures on top level.

build.sbt
gadomsky: fixed

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/Implicits.scala
gadomsky: `withSideEffect`sounds great

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
gadomsky: You're totally right

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
gadomsky: I like the idea of `ProcessId` everywhere so I've added that to this trait. I have also removed `ProcessName` from some events because its unnecessary now.
Ofc youre write about name - also changed.

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
gadomsky: category included

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
gadomsky: :+1:

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ChangeEvent.scala
gadomsky: :+1: 
included where needed

ui/listner-api/src/main/scala/pl/touk/nussknacker/listner/ListenerManagementFactory.scala
gadomsky: now events are handled in a safe way.

arkadius: For aggregates it's ok now, but I think that we need the same in single operation `ListenerManagement` or even upper in `withSideEffect`. Because IMO we don't want to some listener would break our application by throwing Exception. What about changing implementation of `withSideEffect` to:
```
f.onSuccess(f)
f
```
?

gadomsky: good point, fixed

PR #397, opened 2019-11-19
Less verbose logging, slightly better exception on null pointer

mproch:


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/DefaultEspExceptionHandler.scala
```diff
@@ -62,11 +64,22 @@ case class VerboselyLoggingRestartingExceptionHandler(processMetaData: MetaData,
...
+  override def consume(e: EspExceptionInfo[NonTransientException]): Unit = {
     logger.error(s"${processMetaData.id}: Exception during processing job, params: $params, context: ${e.context}", e.throwable)
   }
 }
 
+case class BrieflyLoggingExceptionConsumer(processMetaData: MetaData, params: Map[String, String] = Map.empty)
+  extends FlinkEspExceptionConsumer
+    with LazyLogging {
+  override def consume(e: EspExceptionInfo[NonTransientException]): Unit = {
+    logger.whenDebugEnabled {
```
arkadius: What do u think about:
```
if (logger.underlying.isDebugEnabled) {
  logger.debug(...)
} else {
  logger.warn(...)
}
```
to avoid double logging? Here and in other places in this change...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -404,10 +405,12 @@ object SpelExpressionParser extends LazyLogging {
 
     protected def findMethod(name: String, target: Any): Option[Method] = {
       val targetClass = extractClassFromTarget(target)
+        //TODO: better exception message
```
arkadius: Logic of Caching PropertyAccessors is complicated. In concrete class, in `canRead` we invoke `findMethod` from (two level up) `CachingBase` which invokes `extractClassFromTarget` (from one level down) `Caching` which check some nulls. Maybe this is the moment to simplify it somehow? I wonder if we can have only one level of inheritance and null check at the beginning (`canRead`) to avoid checking it in nested invocations? Maybe some separate `PropertyAccessor` for nulls? Or maybe some composition will help (like extracting new `ClassMethodsMemo` class)? For now I'm not sure why you changed it that way and what exactly is fixed...

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/DefaultEspExceptionHandler.scala
```diff
@@ -62,11 +64,22 @@ case class VerboselyLoggingRestartingExceptionHandler(processMetaData: MetaData,
 case class VerboselyLoggingExceptionConsumer(processMetaData: MetaData, params: Map[String, String] = Map.empty)
   extends FlinkEspExceptionConsumer
     with LazyLogging {
-  override def consume(e: EspExceptionInfo[NonTransientException]) = {
+  override def consume(e: EspExceptionInfo[NonTransientException]): Unit = {
     logger.error(s"${processMetaData.id}: Exception during processing job, params: $params, context: ${e.context}", e.throwable)
   }
 }
 
+case class BrieflyLoggingExceptionConsumer(processMetaData: MetaData, params: Map[String, String] = Map.empty)
```
arkadius: I can't see usage of `params`

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/DefaultEspExceptionHandler.scala
mproch: Hmm... I also thought about it, but when I look at logs I often filter out only WARN/ERROR - and that would get lost if we log only DEBUG here?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
mproch: I don't fix too much here :) 
I just don't want to throw "raw" NullpointerException, as it is later reported as "Unknown Exception", whereas I think it's kind of "normal" that user created diagram where she invokes method on null - it's an exception, but nothing unexpected... 

In https://travis-ci.org/TouK/nussknacker/jobs/613989223
org.springframework.expression.spel.SpelEvaluationException: EL1021E: A problem occurred whilst attempting to access the property 'name': 'null'
...
Caused by: java.lang.NullPointerException: null
	at pl.[secure].nussknacker.engine.spel.SpelExpressionParser$ScalaLazyPropertyAccessor.reallyFindMethod(SpelExpression.scala:307) ~[classes/:na]

And in https://travis-ci.org/TouK/nussknacker/jobs/614155476:

18:56:58.476 [pool-1-thread-1-ScalaTest-running-FlinkTestMainSpec] INFO  p.t.n.engine.spel.SpelExpression - Expression evaluation failed. Original #input.value1.toString(), ctxId: , message: EL1021E: A problem occurred whilst attempting to access the property 'value1': 'Cannot invoke method/property value1 on null object'

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/DefaultEspExceptionHandler.scala
mproch: true that, I didn't copy/paste with enough care ;)

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/DefaultEspExceptionHandler.scala
arkadius: Good point, it is one of scenario. The second one is that developer just look at logs stream line-by-line. But this grepping scenario looks like more critical. WDYT about extracting some LoggingExt utililty method for that pattern? It will prevent us from incidental refactor of this (by developer who've just perform line-by-line scenario)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
arkadius: Ok, good logging cleanup. WDYT about extracting NullPropertyAccessor and in other PropertyAccessors we won't worry about it?

PR #411, opened 2019-11-22
add scala 2.11 & 2.12 compile job in CI

gadomsky:


.travis.yml
```diff
@@ -68,6 +68,8 @@ jobs:
       script: ./ciRunSbt.sh clean ui/slow:test
     - name: "Management tests"
       script: ./ciRunSbt.sh clean management/it:test
+    - name: "Scala 2.11-2.12 compile"
+      script: CROSS_BUILD=true ./ciRunSbt.sh clean compile
```
mproch: test:compile?
Can we make it run only if "original" CROSS_BUILD==false?

arkadius: I think that @mproch want to have both `compile:compile` and `test:compile`. Am I right?

.travis.yml
```diff
@@ -68,6 +68,9 @@ jobs:
       script: ./ciRunSbt.sh clean ui/slow:test
     - name: "Management tests"
       script: ./ciRunSbt.sh clean management/it:test
+    - name: "Scala 2.11-2.12 compile"
```
arkadius: WDYT about changing name of this task to `Cross scala compile`. Let's keep information about all supported versions in one place, in sbt...

mproch: :+1:

.travis.yml
gadomsky: fixed

.travis.yml
gadomsky: fixed

PR #413, opened 2019-11-24
Use proper validation context for branch parameters compilation

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSort.scala
```diff
@@ -0,0 +1,39 @@
...
+import pl.touk.nussknacker.engine.splittedgraph.SplittedNodesCollector
+import pl.touk.nussknacker.engine.splittedgraph.part.SourcePart
+import pl.touk.nussknacker.engine.splittedgraph.splittednode.EndingNode
+
+import scala.annotation.tailrec
+
+object PartSort {
+
+  @tailrec
+  //this is primitive topological sort, we want to sort parts,
```
arkadius: Can we move it on the class level? And also write some details why this sorting is need? Or at least notice in the place of invocation...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -126,12 +129,23 @@ protected trait ProcessCompilerBase {
...
-      splittedProcess.sources.map(compile).sequence
+      compileSources(splittedProcess.sources)
     ) { (_, exceptionHandler, sources) =>
       CompiledProcessParts(splittedProcess.metaData, exceptionHandler, sources)
     }
   }
 
+  private def compileSources(sources: NonEmptyList[SourcePart])(implicit meta: MetaData): CompilationResult[NonEmptyList[PotentiallyStartPart]] = {
+    val zeroAcc = (CompilationResult(Valid(List[PotentiallyStartPart]())), BranchEndContexts(Map(), None))
+    val (result, _) = PartSort.sort(sources.toList).foldLeft(zeroAcc) { case ((resultSoFar, branchContexts), nextPart) =>
```
arkadius: Can we put here some short notice why this sorting is need and reference to longer notice in PartSort? + maybe some TODO if it isn't final solution?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -126,12 +129,23 @@ protected trait ProcessCompilerBase {
...
+        val compiledPart = compile(nextPart, branchContexts)
+        //we don't use andThen on CompilationResult, since we don't want to stop if there are errors in part
+        val nextResult = CompilationResult.map2(resultSoFar, compiledPart)(_ :+ _)
+        (nextResult, branchContexts.addPart(nextPart.node, compiledPart))
```
arkadius: I don't get it ... we add part also for non-join parts? Do we really need to do it in folding (next step depends on previous) or we can do it without such dependency: first some map and then sequence

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/CustomNodeValidationSpec.scala
```diff
@@ -257,6 +258,28 @@ class CustomNodeValidationSpec extends FunSuite with Matchers with OptionValues
...
+          .branch("join1", "unionTransformer", Some("outPutVar"),
+            List(
+              "branch1" -> List("key" -> "'key1'", "value" -> "#input")
+            )
+          )
+          .processorEnd("stringService", "stringService" , "stringParam" -> "''")
+      ))
+    val validationResult = validator.validate(process)
+
+    validationResult.variablesInNodes("stringService")("outPutVar") shouldBe TypedObjectTypingResult(Map("branch1" -> Typed[String]))
```
arkadius: Can you add two branches with different value's types?

mproch: added assertion in other test, this one is for different case

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
mproch: I added comment in code

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
mproch: Added comment

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/PartSort.scala
mproch: Added comment

PR #417, opened 2019-11-25
add read only access to process details for listeners

gadomsky:
* I've extracted additional trait for accessing process details in listeners module so we are in more controll what operations are avaialble in listener implementations.
* maybe we should also rewrite data to special DTOs instead of providing restmodel classes? (now, UI changes will have impact on listeners implementations)

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -52,3 +52,14 @@ case class AdminUser(id: String) extends LoggedUser {
   override def can(category: String, permission: Permission): Boolean = true
   override val isAdmin: Boolean = true
 }
+
+case object ReadOnlyTechUser extends LoggedUser {
+  override val id: String = "tech_user"
+
+  override def can(category: String, permission: Permission): Boolean = permission match {
+    case Permission.Read => true
+    case _ => false
+  }
+
+  override val isAdmin: Boolean = false
+}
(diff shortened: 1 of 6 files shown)


gadomsky:
> maybe we should also rewrite data to special DTOs instead of providing restmodel classes? (now, UI changes will have impact on listeners implementations)

@mproch @arkadius And WDYT about this?

ui/listener-api/src/main/scala/pl/touk/nussknacker/ui/listener/services/PullProcessRepository.scala
```diff
@@ -0,0 +1,15 @@
+package pl.touk.nussknacker.ui.listener.services
+
+import pl.touk.nussknacker.restmodel.process.ProcessId
+import pl.touk.nussknacker.restmodel.processdetails.BaseProcessDetails
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+
+import scala.concurrent.ExecutionContext
+
+trait PullProcessRepository[F[_]] {
+  def fetchLatestProcessDetailsForProcessId(id: ProcessId)
```
mproch: I wonder if it makes sense to abstract over F? Will we have other implementation in the future? 
And also if we should pass LoggedUser from plugin? Maybe it'd be better to pass some internal user in NK itself?

gadomsky: You're right - it makes no sense.
I like the idea of deidcated user for these kind of actions. Check out the code now :)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -51,4 +51,15 @@ case class CommonUser(id: String,
 case class AdminUser(id: String) extends LoggedUser {
   override def can(category: String, permission: Permission): Boolean = true
   override val isAdmin: Boolean = true
+}
+
+case object ReadOnlyTechUser extends LoggedUser {
```
mproch: NussknackerInternalUser -> we have one internal user here, let's keep them consistent? This one will be read-only,
but let's put them in one place and with id like NussknackerAdmin, NussknackerReadOnly?

PR #422, opened 2019-11-25
Less restrictive supertype for two TypedObjectTypingResult

arkadius:
using union in computing value's type instead of looking for common supertype.


arkadius:
We need to do this in other way...

PR #423, opened 2019-11-25
Fix: docker release

lciolecki:


nussknacker-dist/src/universal/conf/docker-application.conf
```diff
@@ -81,7 +81,8 @@ processConfig {
   }
   restartInterval: "10s"
   kafka = {
-    kafkaAddress = "kafka:9092"
+    zkAddress: "zookeeper:2181"
```
mproch: do we need it?

nussknacker-dist/src/universal/conf/docker-application.conf
```diff
@@ -20,15 +20,15 @@ authentication: {
...
-  headers {
-    Accept: "application/json"
+    grant_type: ${?OAUTH2_GRANT_TYPE]
   }
   authorizeParams {
-    response_type: "code"
+    response_type: ${?OAUTH2_RESPONSE_TYPE}
+    scope: ${?OAUTH2_SCOPE}
+  }
+  headers {
```
mproch: maybe we should have defaults here? i Accept, response_type etc.?

lciolecki: Hmm, you are right. I will add default at docker.

nussknacker-dist/src/universal/conf/docker-application.conf
lciolecki: Removed.

PR #444, opened 2019-11-29
Modal actions naming

### LINKED ISSUE #420: Cancel & Apply instead of Close & Save
in node edit modal change Close to Cancel and Save to Apply

sledzmateusz:
* Resolved #420 
* Fixed error on `EdgeDetailsContent` component caused by undefined property which is not used.

### DIFF
--- a/ui/client/components/graph/EdgeDetailsModal.js
+++ b/ui/client/components/graph/EdgeDetailsModal.js
@@ -16,8 +16,7 @@ import {preventFromMoveSelectors} from "../modals/GenericModalDialog";
 class EdgeDetailsModal extends React.Component {
 
   static propTypes = {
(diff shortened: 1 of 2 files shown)


ui/client/components/graph/EdgeDetailsModal.js
```diff
@@ -16,8 +16,7 @@ import {preventFromMoveSelectors} from "../modals/GenericModalDialog";
 class EdgeDetailsModal extends React.Component {
 
   static propTypes = {
-    edgeToDisplay: PropTypes.object.isRequired,
-    showValidation: PropTypes.bool.isRequired
```
lciolecki: Are you sure, that this should be removed?

PR #461, opened 2019-12-03
add marker interface for hiding toString method in code completion

gadomsky:


arkadius:
What about adding this mixin to currently existing utils in our models (dev and demo)?

PR #467, opened 2019-12-04
Fix: OAuth2 expired token

lciolecki:
At start page when token is expired then for a while we can see error page and after that we have redirect to sign in page. This PR fix this problem - we don't show error page when user api data returns 401 and backen is OAuth2.

### DIFF
--- a/ui/client/containers/NussknackerInitializer.js
+++ b/ui/client/containers/NussknackerInitializer.js
@@ -54,26 +54,31 @@ class NussknackerInitializer extends React.Component {
   componentDidMount() {
     //It looks like callback hell.. Can we do it better?
     HttpService.fetchSettings().then(settingsResponse => {
-      this.props.actions.assignSettings(settingsResponse.data)
-      this.authenticationStrategy(settingsResponse.data.authentication).then(response => {
+      const settings = settingsResponse.data
+      this.props.actions.assignSettings(settings)
+
+      this.authenticationStrategy(settings.authentication).then(response => {
         HttpService.fetchLoggedUser().then(userResponse => {
           this.props.actions.assignUser(userResponse.data)
           this.setState({initialized: true})
-        }).catch(this.httpErrorHandler)
+        }).catch(this.httpErrorHandler, _.isEqual(settings.backend, OAUTH2_BACKEND))
       })
     }).catch(this.httpErrorHandler)
   }
 
-  httpErrorHandler = (error) => {
+  httpErrorHandler = (error, redirect) => {
+    const code = _.get(error, 'response.status')
+    const showError = _.isEqual(redirect, true) === false || _.eq(code, HTTP_UNAUTHORIZED_CODE) === false
+
     this.setState({
-      error: _.get(this.errors, _.get(error, 'response.status'), this.errors[HTTP_APPLICATION_CODE])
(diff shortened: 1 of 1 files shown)


ui/client/containers/NussknackerInitializer.js
```diff
@@ -54,26 +54,31 @@ class NussknackerInitializer extends React.Component {
...
-      this.authenticationStrategy(settingsResponse.data.authentication).then(response => {
+      const settings = settingsResponse.data
+      this.props.actions.assignSettings(settings)
+
+      this.authenticationStrategy(settings.authentication).then(response => {
         HttpService.fetchLoggedUser().then(userResponse => {
           this.props.actions.assignUser(userResponse.data)
           this.setState({initialized: true})
-        }).catch(this.httpErrorHandler)
+        }).catch(this.httpErrorHandler, _.isEqual(settings.backend, OAUTH2_BACKEND))
```
arkadius: shouldn't it behave differently for non oauth backend? after your change I assume that we won't handle it at all...

ui/client/containers/NussknackerInitializer.js
```diff
@@ -54,26 +54,31 @@ class NussknackerInitializer extends React.Component {
...
-      error: _.get(this.errors, _.get(error, 'response.status'), this.errors[HTTP_APPLICATION_CODE])
+      error: showError ? _.get(this.errors, code, this.errors[HTTP_APPLICATION_CODE]) : null
     })
   }
 
   authenticationStrategy = (settings) => {
     // Automatically redirect user when he is not authenticated and backend is OAUTH2
     api.interceptors.response.use(response => response, (error) => {
-      if (error.response.status === HTTP_UNAUTHORIZED_CODE && settings.backend === OAUTH2_BACKEND) {
+      if (_.eq(error.response.status, HTTP_UNAUTHORIZED_CODE) && _.isEqual(settings.backend, OAUTH2_BACKEND)) {
```
arkadius: why triple equals is not ok here? we not stick to more standarized way (es6)?

ui/client/containers/NussknackerInitializer.js
lciolecki: For BasicAuth it will behave in different way - will show page with information about non auth user (for BasicAuth there is no redirect).

ui/client/containers/NussknackerInitializer.js
lciolecki: For `===` IDE shows some error: it can't resolve property `.respnse.status` from error object.

ui/client/containers/NussknackerInitializer.js
```diff
@@ -54,19 +54,24 @@ class NussknackerInitializer extends React.Component {
...
-        }).catch(this.httpErrorHandler)
+        }).catch(err => this.httpErrorHandler(err, settings.authentication.backend === OAUTH2_BACKEND))
       })
     }).catch(this.httpErrorHandler)
   }
 
-  httpErrorHandler = (error) => {
+  httpErrorHandler = (error, redirect) => {
+    const code = _.get(error, 'response.status')
+    const showError = (redirect !== true && code === HTTP_UNAUTHORIZED_CODE) || code !== HTTP_UNAUTHORIZED_CODE
```
arkadius: it is just: `code !== HTTP_UNAUTHORIZED_CODE || !redirect`

PR #493, opened 2019-12-11
Update imports with aliases

### LINKED ISSUE #478: GUI - cleanup relative path hell
Imports like this one https://github.com/TouK/nussknacker/blob/9e6c2041249ddbdb79b9836c12698440e8e33be1/ui/client/components/graph/node-modal/editors/map/Map.js#L3 look bad.

Please see https://webpack.js.org/configuration/resolve/, we could start with resolve.alias for "common" directory.

PPiedel:
This resolves #478

### DIFF
--- a/ui/client/actions/actions.js
+++ b/ui/client/actions/actions.js
@@ -1,13 +1,13 @@
-import HttpService from "../http/HttpService";
-import * as GraphUtils from "../components/graph/GraphUtils";
-import NodeUtils from "../components/graph/NodeUtils";
-import * as SubprocessSchemaAligner from "../components/graph/SubprocessSchemaAligner";
-import _ from "lodash";
-import * as UndoRedoActions from "./undoRedoActions";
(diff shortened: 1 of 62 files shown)


ui/client/actions/actions.js
```diff
@@ -1,13 +1,13 @@
-import HttpService from "../http/HttpService";
-import * as GraphUtils from "../components/graph/GraphUtils";
-import NodeUtils from "../components/graph/NodeUtils";
-import * as SubprocessSchemaAligner from "../components/graph/SubprocessSchemaAligner";
-import _ from "lodash";
-import * as UndoRedoActions from "./undoRedoActions";
-import * as VisualizationUrl from '../common/VisualizationUrl';
-import {dateFormat} from "../config";
+import HttpService from "../http/HttpService"
```
lciolecki: Http? :)

lciolecki:
I think we should change all references.. And where we have small letter package like: node-modal change it to NodeModal. WDYT?

PR #501, opened 2019-12-11
Liveautocompletion ehnancements

arkadius:
Bump react-ace 7.0.4 -> 8.0.0 and liveautocompletion ehnancements:
- it works after hash now
- it works after dot now
- suggester is invoked even if some suggestions are computed

### DIFF
--- a/ui/client/components/graph/node-modal/ExpressionSuggest.js
+++ b/ui/client/components/graph/node-modal/ExpressionSuggest.js
@@ -8,10 +8,9 @@ import ProcessUtils from '../../../common/ProcessUtils';
 import ExpressionSuggester from './ExpressionSuggester'
 
 import AceEditor from 'react-ace';
-import 'brace/mode/jsx';
-
-import 'brace/ext/language_tools'
-import 'brace/ext/searchbox';
+import 'ace-builds/src-noconflict/mode-jsx';
+import 'ace-builds/src-noconflict/ext-language_tools'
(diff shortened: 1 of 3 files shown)


ui/client/components/graph/node-modal/ExpressionSuggest.js
```diff
@@ -8,10 +8,9 @@ import ProcessUtils from '../../../common/ProcessUtils';
 import ExpressionSuggester from './ExpressionSuggester'
```
lciolecki: Please remove all semicolons.

ui/client/components/graph/node-modal/ExpressionSuggest.js
```diff
@@ -35,26 +34,38 @@ class ExpressionSuggest extends React.Component {
     processingType: PropTypes.string
   }
 
+  static identifierRegexpsWithoutDot = [/[#a-z0-9-_]/]
```
lciolecki: I think better solution will bemove these props as const before class definition..

arkadius: done

ui/client/components/graph/node-modal/ExpressionSuggest.js
arkadius: done - but removed only in imports, I don't want to touch whole file because I want to have option to quick revert this change in case of some problems with new completion

PR #506, opened 2019-12-12
Showcase for types of parameters 

arkadius:
+ naming refactor: now we have only one DemoProcessConfigCreator and TestProcessConfigCreator was renamed to DevProcessConfigCreator


engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -227,6 +240,8 @@ class TestProcessConfigCreator extends ProcessConfigCreator {
       dictionaries = Map(dictId -> WithCategories(dictDef, "Category1", "Category2")))
   }
 
+  private def all[T](value: T) = WithCategories(value, "Category1", "Category2")
```
mproch: can you change all occurrences?

arkadius: yep, changed

PR #508, opened 2019-12-12
Fix for: Fields types registry on UI, breaks productions build

arkadius:


ui/client/components/graph/node-modal/NodeDetailsContent.js
```diff
@@ -500,7 +500,9 @@ export class NodeDetailsContent extends React.Component {
                   showValidation={showValidation}
                   validators={validators}
                   onChange={handleChange}
-                  className={(!showValidation || allValid(validators, [fieldValue])) ? "node-input" : "node-input node-input-with-error"}/>
+                  className={(!showValidation || allValid(validators, [fieldValue])) ? "node-input" : "node-input node-input-with-error"}
+                  placeholder={""} // FIXME: what should be here?
```
lciolecki: Do we need placeholder?

arkadius: don't know - it is used in lower level in concrete fields

ui/client/components/graph/node-modal/editors/field/Field.js
```diff
@@ -5,33 +5,20 @@ import Checkbox from "./Checkbox"
...
-const resolveComponentType = type => {
-  switch (type) {
+export const Field = (props) => {
+  const {fieldType} = props
+  switch (fieldType) {
     case 'input':
-      return LabeledInput.name
+      return (<LabeledInput {...props}/>)
     case 'unlabeled-input':
-      return Input.name
```
PPiedel: Did you tested it with just using "LabeledInput", "Input", "Checkbox" and so on? Maybe we still can use "Capitalized Reference Technique" (https://medium.com/@j5bot/react-dynamically-rendering-different-components-without-switch-the-capitalized-reference-e668d89e460b) ?

PR #519, opened 2019-12-14
Improvements: process additional fields

lciolecki:
- Add createdAt and createdBy fields at process
- Add hsql and psql migrations


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -5,28 +5,34 @@ import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
 sealed trait LoggedUser {
   val id: String
-  def can(category: String, permission: Permission): Boolean
+  val username: Option[String]
   val isAdmin: Boolean
+  def can(category: String, permission: Permission): Boolean
+  def getName: String = username.getOrElse(id)
```
witekw: It's a trick to save something in not null db column. I would move it closer to the database not to be reused.

arkadius: Indeed @lciolecki can we have `username` as a field as it was before and fill it with the same value in situation when username is the same as id?

ui/server/src/main/resources/db/migration/hsql/V1_026__ProcessCreatedAtByFields.sql
```diff
@@ -0,0 +1,12 @@
+-- create fields
+ALTER TABLE "processes" ADD COLUMN "created_at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
arkadius: Good job :+1:

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
lciolecki: Ofc I can :)

PR #520, opened 2019-12-16
Change search-label appearance in process list (#434)

### LINKED ISSUE #434: search button/label in process list is useless and should be just be a small icon
Since this is not a button you actually click I suggest changing the field to look like this:
Asset 4@2x
the icon should turn blue when filtering is applied

mateuszkp96:
This closes #434 
image
image


ui/client/components/table/SearchFilter.js
```diff
@@ -1,22 +1,23 @@
...
-          onChange={onChange}/>
-        <span className="input-group-addon" id="basic-addon1">
-              <img id="search-icon" src={filterIcon}/>
-        </span>
+        <div className="search-container">
+          <input
+            type="text"
+            placeholder="Filter by text.."
+            className="form-control"
+            aria-describedby="basic-addon1"
```
PPiedel: Are you sure?

ui/client/components/table/SearchFilter.js
```diff
@@ -1,22 +1,23 @@
...
-          type="text"
-          placeholder="Filter by text.."
-          className="form-control"
-          aria-describedby="basic-addon1"
-          value={value}
-          onChange={onChange}/>
-        <span className="input-group-addon" id="basic-addon1">
-              <img id="search-icon" src={filterIcon}/>
-        </span>
+        <div className="search-container">
```
PPiedel: Maybe it could be replaces with Input.js component?

ui/client/components/table/SearchFilter.js
mateuszkp96: Removed

PR #521, opened 2019-12-16
[Refactor] test setup cleanup in process module

mproch:


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/sources.scala
```diff
@@ -58,6 +59,9 @@ object sources {
 
   object JsonMapDeserialization extends EspDeserializationSchema[java.util.Map[_, _]](m => deserializeToMap(m).asJava)
 
-  object JsonTypedMapDeserializaion extends EspDeserializationSchema[TypedMap](m => TypedMap(deserializeToMap(m)))
+  object JsonTypedMapDeserialization extends EspDeserializationSchema[TypedMap](m => TypedMap(deserializeToMap(m)))
+
+  //TOOD: better error handling?
+  class JsonDecoderDeserialization[T:Decoder:TypeInformation] extends EspDeserializationSchema[T](ba => CirceUtil.decodeJsonUnsafe(ba))
```
jedrz: Is it used anywhere? Anyway it should be useful in other projects using nk :)

PR #522, opened 2019-12-16
Humanize naming of nodes (#495)

### LINKED ISSUE #495: Humanizing naming on UI
Proposition of naming changes:
- Create new process, Id -> Name
- Edition of node: Id -> Name
- Creating new node: node1, node2, ... -> Id (name) of node + number if collision

mateuszkp96:
This closes #495 

Adding nodes from creator panel:
image
Copy of node:
image
Copy of copy of node:
image
Id changed to name:
image

### DIFF
--- a/ui/client/components/AddProcessDialog.js
+++ b/ui/client/components/AddProcessDialog.js
@@ -69,7 +69,7 @@ class AddProcessDialog extends React.Component {
                 <div className="node-table">
                   <div className="node-table-body">
                     <div className="node-row">
-                      <div className="node-label">Id</div>
+                      <div className="node-label">Name</div>
                       <div className="node-value">
                         <input autoFocus={true} type="text" id="newProcessId" className="node-input"
                                value={this.state.processId}
--- a/ui/client/components/Tool.js
+++ b/ui/client/components/Tool.js
@@ -27,8 +27,12 @@ class Tool extends React.Component {
 }
 
 var spec = {
-  beginDrag: (props, monitor, component) => props.nodeModel
-};
+  beginDrag: (props, monitor, component) => {
+    const nodeModel = _.cloneDeep(props.nodeModel)
+    _.set(nodeModel, 'id', props.label)
+    return nodeModel
+  }
+}
 
 export default DragSource("element", spec, (connect, monitor) => ({
   connectDragSource: connect.dragSource()
--- a/ui/client/reducers/graph.js
+++ b/ui/client/reducers/graph.js
@@ -173,11 +173,11 @@ export function reducer(state, action) {
         prepareNewNodesWithLayout(state,[{
           node: action.node,
           position: action.position
-        }])
+        }], false)
       )
     }
     case "NODES_WITH_EDGES_ADDED": {
-      const {nodes, layout, uniqueIds} = prepareNewNodesWithLayout(state, action.nodesWithPositions)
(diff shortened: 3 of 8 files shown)


ui/client/components/AddProcessDialog.js
```diff
@@ -69,7 +69,7 @@ class AddProcessDialog extends React.Component {
                 <div className="node-table">
                   <div className="node-table-body">
                     <div className="node-row">
-                      <div className="node-label">Id</div>
+                      <div className="node-label">{"Name"}</div>
```
arkadius: Why in braces?

ui/client/reducers/graph.js
```diff
@@ -339,15 +339,15 @@ function updateAfterNodeDelete(state, idToDelete) {
...
     ? initialId
-    : generateUniqueNodeId(usedIds, nodeCounter)
+    : generateUniqueNodeId(initialId, usedIds, 1, isCopy)
 }
 
-function generateUniqueNodeId(usedIds, nodeCounter) {
-  const newId = `node${nodeCounter}`;
-  return _.includes(usedIds, newId) ? generateUniqueNodeId(usedIds, nodeCounter + 1) : newId
+function generateUniqueNodeId(initialId, usedIds, nodeCounter, isCopy) {
+  const newId = isCopy ? `${initialId}(${nodeCounter} copy)` : `${initialId}${nodeCounter}`
```
arkadius: can you put a space: `nodeName (1 copy)`? also maybe will be better `nodeName (copy 1)` because if we write `1 copy` then we should rather write `1st copy` otherwise it won't be grammatically correct. `copy 1` looks more technical and then gramatics are not so important

arkadius: ... also IMO `Node name 1` will looks better than `Node name1`, because we want to move into rather human names than ids in this presentation...

mateuszkp96: Now looks like:
image
image

ui/client/components/AddProcessDialog.js
mateuszkp96: Removed

ui/client/reducers/graph.js
lciolecki: But copy of copy? :)

ui/client/components/Tool.js
```diff
@@ -27,8 +27,12 @@ class Tool extends React.Component {
 }
 
 var spec = {
-  beginDrag: (props, monitor, component) => props.nodeModel
-};
+  beginDrag: (props, monitor, component) => {
+    const nodeModel = _.cloneDeep(props.nodeModel)
+    _.set(nodeModel, 'id', props.label)
+    return nodeModel;
```
lciolecki: remove ;

ui/client/reducers/graph.js
```diff
@@ -339,15 +339,15 @@ function updateAfterNodeDelete(state, idToDelete) {
   }
 }
 
-function createUniqueNodeId(initialId, usedIds, nodeCounter) {
+function createUniqueNodeId(initialId, usedIds, isCopy) {
   return initialId && !_.includes(usedIds, initialId)
     ? initialId
-    : generateUniqueNodeId(usedIds, nodeCounter)
+    : generateUniqueNodeId(initialId, usedIds, 1, isCopy)
```
PPiedel: Always 1? It looks like a magic number ;)

PR #523, opened 2019-12-16
Fixes: custom stream transformers aren't grouped in categories

arkadius:


ui/server/src/test/resources/ui.conf
```diff
@@ -90,8 +90,9 @@ processConfig {
   }
 
   nodeCategoryMapping: {
```
lciolecki: Hmm maybe nodeGroup is better name? We already use category in another meaning..

arkadius: I don't want to refactor in this change... also you have categories of sports, categories of movies, IMO it is not a bad name, but always should exist with name of thing that you want to categorize.

lciolecki: It's only suggestion ;) I've already added approve :)

PR #528, opened 2019-12-17
Fix side node details

### LINKED ISSUE #514: Details of process with grouped nodes are badly presented
Steps to reproduce:
- group some nodes on process
- click on empty space on process

Expected behaviour: should be visible nice nodes details lists
Current behaviour: are presented: each letter of nodes name in separate line (looks like some iterating over string instead of list of nodes)

Screenshot from 2019-12-13 10-23-47

PPiedel:
This resolves #514

### DIFF
--- a/ui/client/components/right-panel/SideNodeDetails.js
+++ b/ui/client/components/right-panel/SideNodeDetails.js
@@ -1,8 +1,7 @@
-import React, {Component} from "react";
-import {render} from "react-dom";
-import {connect} from "react-redux";
-import _ from "lodash";
-import ActionsUtils from "../../actions/ActionsUtils";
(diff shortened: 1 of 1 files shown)


ui/client/components/right-panel/SideNodeDetails.js
```diff
@@ -11,23 +10,25 @@ class SideNodeDetails extends Component {
       if (_.isArray(val)) {
```
lciolecki: What a beautiful ifhell :)

lciolecki: I think need there some factory to creating FlatObjectEntry.  Alsow in render method you don't need else statement.

PR #530, opened 2019-12-17
Hiding meta variable

### LINKED ISSUE #499: Hiding meta variable
I would like to hide "meta" global variable. Can be done by some feature flag or by transformation of global variables in ConfigCreator instead of defining only variables to add.

arkadius:
Fixes #499


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TestInfoProvider.scala
```diff
@@ -63,9 +65,8 @@ class ModelDataTestInfoProvider(modelData: ModelData) extends TestInfoProvider {
   }
 
   private def contextWithOnlyGlobalVariables(implicit metaData: MetaData): ValidationContext = {
-    val globalTypes = modelData.processDefinition.expressionConfig.globalVariables.mapValues(_.returnType)
-    val typesWithMeta = MetaVariables.withType(globalTypes)
-    ValidationContext(Map.empty, typesWithMeta)
+    val globalTypes = globalVariablesPreparer.prepareGlobalVariables(metaData).mapValues(_.typ)
```
lciolecki: Maybe you can create something like `prepareGlobalTypes(metadata)`? That will hide mapValues inside preparer.

PR #533, opened 2019-12-18
Notify user that selected content can not be copied

### LINKED ISSUE #516: Show error that selected element cannot be copied
Currently AFAIK we support copying only nodes.
If user select something that cannot be copied, and press ctrl+c or click copy button, we should display error similar to this presented during pasting not supported object.

Scenarios:
- selection of empty space on process, then ctrl+c
- selection of grouped node, then ctrl+c

But this should work in general if we not support copying of given element...

PPiedel:
This resolves #516

### DIFF
--- a/ui/client/containers/Visualization.js
+++ b/ui/client/containers/Visualization.js
@@ -181,7 +181,7 @@ class Visualization extends React.Component {
     const isNotThisCopyEvent = this.isNotThisCopyEvent(event, copyNodeElementId)
 
     isNotThisCopyEvent && this.canCopySelection() ? this.copyToClipboard(shouldCreateNotification) :
-      this.props.notificationActions.error("Can not copy selected content. It should contains only plain nodes without groups")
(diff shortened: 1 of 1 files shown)


ui/client/containers/Visualization.js
```diff
@@ -179,19 +179,23 @@ class Visualization extends React.Component {
...
-        nodes: selectedNodes,
-        edges: edgesForNodes
-      }
-      ClipboardUtils.writeText(JSON.stringify(selection), copyNodeElementId);
-      if (shouldCreateNotification) {
-        this.props.notificationActions.success(this.successMessage('Copied', selectedNodes))
-      }
+
+    isNotThisCopyEvent && this.canCopySelection() ? this.copyToClipboard(shouldCreateNotification) :
+      this.props.notificationActions.error("Can not copy selected content. It should contains only plain nodes without groups")
```
witekw: ... contain ...

PPiedel: Done

PR #535, opened 2019-12-19
Add loose nodes tip

### LINKED ISSUE #496: Invalid end error is presented as "Errors IN ..." on the left panel
The problem is in word "IN" the problem is not inside node, but the node is a problem. One solution for that could be excluding this type of error from grouping in "Errors in ..."

PPiedel:
This resolves #496


ui/client/components/tips/ErrorhHeader.js
```diff
@@ -0,0 +1,8 @@
+import React from "react"
+
+export function ErrorHeader(props) {
```
lciolecki: Do we need external file for this and for Node ErrorLink? If you use it only in Errors, maybe better way is creating two functional components in Errors?

PPiedel: Yes, you're right, done

Issue #475, opened 2019-12-06
with grouping mode enabled double clicking on a node opens process properties


witekw:
We should not open anything.

PR #545, opened 2019-12-23
Fix: with grouping mode enabled double clicking on a node opens process properties #475

JulianWielga:


witekw:
This issue might be outdated. This error does not occure on staging branch.

PR #547, opened 2019-12-27
Improvements: More precise process state presentation

### LINKED ISSUE #449: More precise process state presentation
We need to show more details about current process state. Consider scenario:
- Developer implement extension that have multiple stages of deployment e.g.: preparering, depoyment in progress, deployed
- As a user I would like to see exact state of deployment in two places: on 1) list of processes and 2) process details

lciolecki:
This resolves #449.

Before state loaded:
loading-sate-process-icons

After state loaded:
loaded-state-icons


ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
+    DURING_DEPLOY: "DURING_DEPLOY",
+    RUNNING: "RUNNING",
+    CANCELED: "CANCELED",
+    RESTARTING: "RESTARTING",
+    FAILED: "FAILED",
+    FINISHED: "FINISHED",
+  },
+
+  TOOLTIPS: {
+    UNKNOWN: "Unknown state of process.",
```
witekw: Unknown state of the process.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
+    NOT_DEPLOYED: "Process has been never deployed.",
```
witekw: The process has never been deployed.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
+    DURING_DEPLOY: "Process has been already started and currently is during deploying.",
```
witekw: The process has been already started and currently is being deployed.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
+    RUNNING: "Process successful deployed and currently is running.",
+    CANCELED: "Process is successfully canceled.",
```
witekw: The process is successfully cancelled.
The process is cancelled.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
+    RESTARTING: "RESTARTING",
+    FAILED: "FAILED",
+    FINISHED: "FINISHED",
+  },
+
+  TOOLTIPS: {
+    UNKNOWN: "Unknown state of process.",
+    NOT_DEPLOYED: "Process has been never deployed.",
+    DURING_DEPLOY: "Process has been already started and currently is during deploying.",
+    RUNNING: "Process successful deployed and currently is running.",
```
witekw: The process is running.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,116 @@
...
 
-  getStatusMessage(processState, shouldRun, loaded) {
-    const isRunning = _.get(processState, 'isRunning')
-    if (isRunning) {
-      return "Running"
-    } else if (shouldRun) {
-      const message = (processState || {}).errorMessage || "Not found in engine"
-      return loaded ? `Not running: ${message}` : "Unknown state"
+  getProcessTooltip(process, isStateLoaded) {
+    const pendingTitle = (title) => title + (isStateLoaded === false ? " Loading actually state of process..." : "")
```
witekw: Loading current state of the process

https://english.stackexchange.com/questions/29654/using-actual-to-signify-current#29655

ui/client/components/Process/ListState.js
```diff
@@ -0,0 +1,90 @@
+import React from "react"
+import PropTypes from 'prop-types'
+import ProcessStateUtils from '../../common/ProcessStateUtils'
+import {connect} from "react-redux"
+import ActionsUtils from "../../actions/ActionsUtils"
+
+export class ListStateComponent extends React.Component {
+  iconStyles = {
```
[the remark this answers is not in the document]
lciolecki: I think that in this case we can leave it :)

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,124 @@
...
+    DEPLOYED: "DEPLOYED",
+    RUNNING: "RUNNING",
+    CANCELED: "CANCELED",
+    RESTARTING: "RESTARTING",
+    FAILED: "FAILED",
+    ERROR: "ERROR",
+    FINISHED: "FINISHED",
+  },
+
+  STATUS_TOOLTIPS: {
```
arkadius: As I understand, these are tooltips that will be used if api won't return own tooltip? If it's true, maybe we can explicitly name? Like `DEFAULT_STATUS_TOOLTIPS`? Don't you think that it will be confusing that part tooltips preparing will be on BE side, and other part on FE side?

engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultStateStatus.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.engine.defaults.deployment
+
+import pl.touk.nussknacker.engine.api.deployment.StatusState
+
+object DefaultStateStatus extends Enumeration {
```
arkadius: Is it really global defaults? Maybe better would be FlinkStateStatus or at least StreamingStateStatus?

ui/client/common/ProcessStateUtils.js
lciolecki: I agree with taht we can named it DEFAULT_*. It can be confusing, but I have done it in this way because of optimization case - all tooltips and icons now are cached by http server.

engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultStateStatus.scala
lciolecki: Hmm.. I can do StreamingStateStatus and StandaloneStateStatus.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -1,21 +1,64 @@
 package pl.touk.nussknacker.engine.api.deployment
 
-import io.circe.{Decoder, Encoder}
 import io.circe.generic.JsonCodec
+import io.circe.{Decoder, Encoder, Json}
 import pl.touk.nussknacker.engine.api.ProcessVersion
+import pl.touk.nussknacker.engine.api.deployment.StateAction.StateAction
+import pl.touk.nussknacker.engine.api.deployment.StatusState.StateStatus
```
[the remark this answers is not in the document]
lciolecki: I've changed it to:

```
object EnumStateStatus extends Enumeration {

  implicit val typeEncoder: Encoder[StateStatus] = Encoder.encodeString.contramap(_.toString())
  implicit val typeDecoder: Decoder[StateStatus] = Decoder.decodeNone.map(_ => null)

  class StateStatus(name: String) extends Val(name)
}
```

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -1,21 +1,64 @@
...
+import pl.touk.nussknacker.engine.defaults.deployment.DefaultProcessStateConfigurator
 
-@JsonCodec case class ProcessState(id: DeploymentId,
-                                   runningState: RunningState.Value,
+trait ProcessStateConfigurator {
```
arkadius: Please write comment here about usage - why maps and so on. WDYT about rename it to something with ProcessStateDefinitition in name? Like ProcessStateDefinitionProvider or ProcessStateDefinitionManager?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -1,21 +1,64 @@
...
+  def processStateStatus: Enumeration
+  def statusTooltips: Map[StateStatus, String]
+  def statusIcons: Map[StateStatus, String]
+  def isRunning(statusState: String): Boolean
```
arkadius: Please move it to traits that will be mixed to StateStatus. Lets this class be as simple as it can. Also IMO we can leave TODO near this traits that we should find usage of them and think if it is really needed to use it in custom implementations. Maybe we can replace them with some more explicit methods like `prepareAnnouncementMessage` and in simple implementation it will check which processes are running or any other state... but in custom one, it can return whatever - also empty string.

engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultProcessStateConfigurator.scala
```diff
@@ -0,0 +1,38 @@
+package pl.touk.nussknacker.engine.defaults.deployment
```
arkadius: 1. please replace order of package - should be `deployment.defaults` - take a look at api.dict.embedded
2. we should use singular package names instead of plural
3. what about replacing defaults with something like "simple"

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -114,10 +119,15 @@ class AppResources(config: Config,
...
-  private def statusList(processes: Seq[BaseProcessDetails[_]])(implicit user: LoggedUser) : Seq[Future[(String, Option[ProcessStatus])]] = {
+  private def statusList(processes: Seq[BaseProcessDetails[_]])(implicit user: LoggedUser) : Seq[Future[(BaseProcessDetails[_], Option[ProcessStatus])]] = {
     processes
       .filterNot(_.deployment.isEmpty)
-      .map(process => findJobStatus(process.idWithName, process.processingType).map((process.name, _)))
+      .map(process => findJobStatus(process.idWithName, process.processingType).map((process, _)))
+  }
+
+  private def isStatusOkForDeployed(status: ProcessStatus, process: BaseProcessDetails[_]): Boolean = {
+    val processStateConfigurator = typeToConfig(process.processingType).processManager.processStateConfigurator
```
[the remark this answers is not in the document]
lciolecki: I've moved it to StateStatus definition.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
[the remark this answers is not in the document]
lciolecki: I've changed it:
```
object StateStatus {
  class StateStatus(val name: String) {
    def isOkForDeployed: Boolean = isDuringDeploy || isRunning
    def isDuringDeploy: Boolean = false
    def isFinished: Boolean = false
    def isRunning: Boolean = false
  }

  final class DuringDeployStateStatus(name: String) extends StateStatus(name) {
    override def isDuringDeploy: Boolean = true
  }

  final class FinishedStateStatus(name: String) extends StateStatus(name) {
    override def isFinished: Boolean = true
  }

  final class RunningStateStatus(name: String) extends StateStatus(name) {
    override def isRunning: Boolean = true
  }
}```

File: engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala (hunk 11 above)
Comment (lciolecki) [reply engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala]: Changed ProcessStateDefinitionManager.

File: engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultProcessStateConfigurator.scala (hunk 13 above)
Comment (lciolecki) [reply engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultProcessStateConfigurator.scala]: I've moved it to simple in development.

File: engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala (hunk 12 above)
Comment (lciolecki) [reply engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala]: I've moved it to StateStatus:
'''
object StateStatus {
  class StateStatus(val name: String) {
    def isOkForDeployed: Boolean = isDuringDeploy || isRunning
    def isDuringDeploy: Boolean = false
    def isFinished: Boolean = false
    def isRunning: Boolean = false
  }

  final class DuringDeployStateStatus(name: String) extends StateStatus(name) {
    override def isDuringDeploy: Boolean = true
  }

  final class FinishedStateStatus(name: String) extends StateStatus(name) {
    override def isFinished: Boolean = true
  }

  final class RunningStateStatus(name: String) extends StateStatus(name) {
    override def isRunning: Boolean = true
  }
}
'''

File: engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala (hunk 11 above)
Comment (lciolecki) [reply engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala]: Added.

File: engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultProcessStateConfigurator.scala (hunk 13 above)
Comment (lciolecki) [reply engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultProcessStateConfigurator.scala]: Moved

File: ui/client/common/ProcessStateUtils.js [hunk 15]
```diff
@@ -1,39 +1,124 @@
...
+
+  getProcessAction(process) {
+    return _.get(process, 'deployment.action', null)
+  },
+
+  getProcessTooltip(process) {
+    return this.getStatusTooltip(this.getProcessStatus(process))
+  },
+
+  getProcessStatus(process) {
```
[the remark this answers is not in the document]
lciolecki: I map deployment information from process to status, because I in next step I map status to icon. I want normalize getting icon / tooltip to getting it by status.

engine/api/src/main/scala/pl/touk/nussknacker/engine/defaults/deployment/DefaultStateStatus.scala
lciolecki: I've changed it to SimpleStateStatus. And I've crated FlinkStateStatus.

ui/client/common/ProcessStateUtils.js
```diff
@@ -1,39 +1,134 @@
...
+
+  getStateStatus(state) {
+    const status = _.get(state, "status", null)
+
+    if (_.isUndefined(state) || (state && status === null)) {
+      return this.STATUSES.UNKNOWN
     }
-    return null
+
+    //TODO: In future it will be nice when API return state with status CANCELED
```
arkadius: Please check if it can be on backend site - let's keep this walkaround as close to source of problem as it possible. Or at least write more verbose comment, that in situation when process was ran and then was stopped api returns null - it is the only possible situation when there is null in response so we assume that null mean CANCELED

.gitignore
```diff
@@ -14,7 +14,6 @@ dist/
 **/db.tmp
 dbFiles/*
 *.jar
-ui/server/src/main/resources/web
```
arkadius: Please check if it isn't used in dist preparing. If so, split web to web (sources) and web-generated (generated sources - ignored in git).

ui/client/assets/img/states/canceled.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"> <defs><style>.jfjb-1,.jfjb-2{fill:none}.jfjb-1{opacity:0.35}.jfjb-3{fill:#4d4d4d}.jfjb-4{fill:#64d864}</style></defs> <g id="Layer_2"> <g id="Layer_1-2"> <rect class="jfjb-1" width="24" height="24"/> <circle class="jfjb-2" cx="18" cy="18" r="6"/> <path class="jfjb-3" d="M17,12.08a6,6,0,0,1,3.32.39A.49.49,0,0,0,21,12V3.5a.5.5,0,0,0-.5-.5H3.5a.5.5,0,0,0-.5.5v17a.5.5,0,0,0,.5.5H12a.49.49,0,0,0,.46-.68A6,6,0,0,1,17,12.08Z"/> <path class="jfjb-4" d="M18,13a5,5,0,1,0,5,5A5,5,0,0,0,18,13Zm-.53,7.34L16.33,19.2,15,17.87l1.14-1.14,1.33,1.33,2.39-2.39L21,16.8Z"/> </g> </g> </svg>
```
arkadius: Please reformat this files

build.sbt
```diff
@@ -614,6 +613,7 @@ lazy val api = (project in engine("api")).
         "io.circe" %% "circe-parser" % circeV,
         "io.circe" %% "circe-generic" % circeV,
         "io.circe" %% "circe-generic-extras" % circeV,
+        "com.iheart" %% "ficus" % ficusV,
```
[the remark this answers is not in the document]
lciolecki: Yep. I've created URI encoder

ui/client/common/ProcessStateUtils.js
lciolecki: I've added support for CANCELED / DURING CANCEL status.

.gitignore
lciolecki: Frontend stuff is copied by webpack:

```
  output: {
    //by default we use default webpack value, but we want to be able to override it for building frontend via sbt
    path: process.env.OUTPUT_PATH ? path.join(process.env.OUTPUT_PATH, "classes", "web", "static") : path.join(process.cwd(), "dist"),
    filename: "[name].js",
    //see config.js
    publicPath: isProd ? "__publicPath__/static/" : "/static/",
  },
```
It is copied to target.

PR #550, opened 2019-12-30
Add String editor

### LINKED ISSUE #505: Editor for string type
We need to have editor for string type that will not need to add quotes around value - user will fill `foo bar` instead of `'foo bar'`
IMO:
- should be possible to switch between human-friendly editor and expressions (like in grafana)
- editor should be accessible for all fields with type: java.lang.String
- editor should be by default in "expression" mode
- if expression saved in process have value matching `/'.*'/` pattern, we should render it using string editor. Otherwise (e.g. #input.foo) we should use "expression" mode
- default editor can be changed on backend side (will be done in other task)

UI challenge that I can see: user can by mistake write `'foo bar'` in string editor. We should give clear experience about which mode is enabled.

PPiedel:
This resolves #505

### DIFF
--- a/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
+++ b/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
@@ -5,8 +5,8 @@ import PropTypes from "prop-types"
 export default function BoolEditor(props) {
 
   const {
-    renderFieldLabel, fieldLabel, fieldName, expressionObj, isMarked, readOnly, onValueChange, switchableTo, toggleEditor,
-    shouldShowSwitch, rowClassName, valueClassName, displayRawEditor, switchableToHint, notSwitchableToHint
+    expressionObj, readOnly, onValueChange,
+    shouldShowSwitch
   } = props
 
   const trueValue = {expression: "true", label: "true"}
@@ -19,34 +19,20 @@ export default function BoolEditor(props) {
         falseValue
       ]}
       defaultValue={trueValue}
-      renderFieldLabel={renderFieldLabel}
-      fieldLabel={fieldLabel}
       expressionObj={expressionObj}
       onValueChange={onValueChange}
       readOnly={readOnly}
-      switchableTo={switchableTo}
-      switchableToHint={switchableToHint}
-      notSwitchableToHint={notSwitchableToHint}
-      toggleEditor={toggleEditor}
       shouldShowSwitch={shouldShowSwitch}
-      rowClassName={rowClassName}
-      valueClassName={valueClassName}
-      displayRawEditor={displayRawEditor}
     />
   )
 }
 
 BoolEditor.propTypes = {
-  renderFieldLabel: PropTypes.func,
-  fieldLabel: PropTypes.string,
   expressionObj: PropTypes.object,
   onValueChange: PropTypes.func,
   readOnly: PropTypes.bool,
-  switchable: PropTypes.bool,
-  toggleEditor: PropTypes.func,
   shouldShowSwitch: PropTypes.bool,
   rowClassName: PropTypes.string,
-  valueClassName: PropTypes.string
 }
 
 const parseable = (expressionObj) => {
--- a/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
+++ b/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
@@ -5,8 +5,8 @@ import PropTypes from "prop-types"
 export default function BoolEditor(props) {
 
   const {
-    renderFieldLabel, fieldLabel, fieldName, expressionObj, isMarked, readOnly, onValueChange, switchable, toggleEditor,
-    shouldShowSwitch, rowClassName, valueClassName, displayRawEditor, switchableHint
+    renderFieldLabel, fieldLabel, fieldName, expressionObj, isMarked, readOnly, onValueChange, switchableTo, toggleEditor,
+    shouldShowSwitch, rowClassName, valueClassName, displayRawEditor, switchableToHint, notSwitchableToHint
   } = props
 
   const trueValue = {expression: "true", label: "true"}
@@ -24,8 +24,9 @@ export default function BoolEditor(props) {
       expressionObj={expressionObj}
       onValueChange={onValueChange}
       readOnly={readOnly}
-      switchable={switchable}
-      hint={switchableHint()}
+      switchableTo={switchableTo}
+      switchableToHint={switchableToHint}
+      notSwitchableToHint={notSwitchableToHint}
       toggleEditor={toggleEditor}
       shouldShowSwitch={shouldShowSwitch}
       rowClassName={rowClassName}
@@ -55,9 +56,12 @@ const parseable = (expressionObj) => {
 }
 
 const supportedLanguage = "spel"
+const supportedFieldType = "Boolean"
 
-BoolEditor.switchableOnto = (expressionObj) => parseable(expressionObj) || _.isEmpty(expressionObj.expression)
-BoolEditor.nonSwitchableOntoHint = "Expression must be equal to true or false to switch to basic mode"
+BoolEditor.switchableTo = (expressionObj) => parseable(expressionObj) || _.isEmpty(expressionObj.expression)
 
-BoolEditor.switchableFrom = (_) => true
-BoolEditor.switchableFromHint = "Switch to expression mode"
+BoolEditor.switchableToHint = "Switch to basic mode"
+
+BoolEditor.notSwitchableToHint = "Expression must be equal to true or false to switch to basic mode"
+
+BoolEditor.isSupported = (fieldType) => fieldType === supportedFieldType
--- a/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
+++ b/ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
@@ -51,8 +51,10 @@ BoolEditor.propTypes = {
 const parseable = (expressionObj) => {
   const expression = expressionObj.expression
   const language = expressionObj.language
-  return (expression === "true" || expression === "false") && language === "spel"
+  return (expression === "true" || expression === "false") && language === supportedLanguage
 }
+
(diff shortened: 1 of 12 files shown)


ui/client/components/graph/node-modal/editors/expression/SwitchIcon.js
```diff
@@ -1,18 +1,25 @@
...
+
+    if (!displayRawEditor)
+      return "Switch to expression mode"
+
+    if (switchable) {
+      return "Switch to basic mode"
+    } else if (fieldType === Types.BOOLEAN || fieldType === Types.EXPRESSION) {
+      return "Expression must be equal to true or false to switch to basic mode"
+    } else if (fieldType === Types.STRING) {
+      return "Expression must be surrounded by single quotation marks or be empty to switch to basic mode"
```
witekw: This is not true, "expression must be a simple string literal i.e. text surrounded by single quotation marks".

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
```diff
@@ -0,0 +1,20 @@
...
+  return <RawEditor {...props} trimValue={trim} formatValue={format}/>
+}
+
+export const parseableString = (expressionObj) => {
+  const expression = expressionObj.expression
+  const language = expressionObj.language
+  return stringPattern.test(expression) && language === "spel"
+}
+
+const stringPattern = /^'.*'$/
```
mproch: What about expression:   'first_part' + #input.myVar + 'second_part'? And what about "?

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
```diff
@@ -0,0 +1,20 @@
+import React from "react"
```
mproch: I don't think we should have suggestions in "simple" mode? they won't work anyway?

PPiedel: What about completion suggestions like: #meta.processName ? They still work.

ui/client/components/graph/node-modal/editors/expression/SwitchIcon.js
PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
PPiedel: I see now that is a bit more complicated than we thought. 
Two questions:
1. Any phrase which is not in '...' should start with # ? 
2. Are "..." valid replacement for '...'?

ui/client/components/graph/node-modal/editors/expression/SwitchIcon.js
```diff
@@ -1,18 +1,25 @@
...
 
 export default function SwitchIcon(props) {
 
-  const {switchable, onClick, shouldShowSwitch, displayRawEditor, title, readOnly} = props
+  const {switchable, onClick, shouldShowSwitch, displayRawEditor, title, readOnly, fieldType} = props
 
-  function hint(switchable, displayRawEditor, readOnly) {
-    if (readOnly) {
+  function hint() {
+    if (readOnly)
```
lciolecki: brackets

ui/client/components/graph/node-modal/editors/expression/SwitchIcon.js
```diff
@@ -1,18 +1,25 @@
...
       return "Switching to basic mode is disabled. You are in read-only mode"
-    } else {
-      return !displayRawEditor ? "Switch to expression mode" :
-        (switchable ? "Switch to basic mode" :
-          "Expression must be equal to true, false or be empty to switch to basic mode")
+
+    if (!displayRawEditor)
```
lciolecki: brackets

PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/SwitchIcon.js
PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
PPiedel: Change to standard input, without suggestions

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
PPiedel: "" is supported now

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
```diff
@@ -0,0 +1,39 @@
...
+
+}
+
+export const parseableString = (expressionObj) => {
+  const expression = expressionObj.expression
+  const language = expressionObj.language
+  return stringPattern.test(expression) && language === "spel"
+}
+
+const stringPattern = /(^['].*'$)|(^".*"$)/
```
mproch: the second part should also contain ^["]? Let's also add (at least) comment that we don't handle expressions with escaped '/"...

ui/client/components/graph/node-modal/editors/expression/EditorType.js
```diff
@@ -13,32 +13,54 @@ class EditorType {
...
+      case Types.STRING:
+        return !displayRawEditor && this.switchableToStringEditor(expressionObj) ? Types.STRING_EDITOR : Types.RAW_EDITOR
       case Types.EXPRESSION_WITH_FIXED_VALUES || Types.RAW_EDITOR:
         return fieldType
       default:
         return Types.RAW_EDITOR
     }
   }
 
+  switchable = (editorName, expressionObj, fieldType) => {
```
mproch: wouldn't it be more clear if custom editors were classes with switchable method, checking both type and custom conditions? And also some nonSwitchableHint field/method? Now the logic is pretty scattered and it's not so clear how to add new editors without adding many case expressions...

ui/client/components/graph/node-modal/editors/expression/StringEditor.js
PPiedel: [] brackets are actually unnecessary, I've added also comment about ecaped characters

ui/client/components/graph/node-modal/editors/expression/EditorType.js
PPiedel: I've moved some logic to editors

ui/client/components/graph/node-modal/editors/expression/BoolEditor.js
```diff
@@ -45,4 +46,16 @@ BoolEditor.propTypes = {
...
   rowClassName: PropTypes.string,
   valueClassName: PropTypes.string
-}
\ No newline at end of file
+}
+
+const parseable = (expressionObj) => {
+  const expression = expressionObj.expression
+  const language = expressionObj.language
+  return (expression === "true" || expression === "false") && language === "spel"
```
lciolecki: "spel" should be defined as const

ui/client/components/graph/node-modal/editors/expression/EditorType.js
```diff
@@ -1,41 +1,48 @@
...
+  [Types.STRING_EDITOR]: {
+    editor: StringEditor,
+    switchable: BoolEditor.switchableFrom(),
+    switchableHint: (_) => StringEditor.switchableFromHint,
+    supportedFieldType: Types.STRING
+  },
+  [Types.EXPRESSION_WITH_FIXED_VALUES]: {
+    editor: ExpressionWithFixedValues,
+    switchable: false,
+    switchableHint: null
```
lciolecki: Do we need set there null?

ui/client/components/graph/node-modal/editors/expression/EditorType.js
```diff
@@ -1,41 +1,47 @@
...
+  },
+  [Types.BOOL_EDITOR]: {
+    editor: BoolEditor,
+    switchable: BoolEditor.switchableFrom(),
+    switchableHint: (_) => BoolEditor.switchableFromHint,
+  },
+  [Types.STRING_EDITOR]: {
+    editor: StringEditor,
+    switchable: BoolEditor.switchableFrom(),
+    switchableHint: (_) => StringEditor.switchableFromHint,
```
mproch: but it should be other way round:
- switchableTo 
- switchableToHint
- notSwitchableToHint

ui/client/components/graph/node-modal/editors/expression/EditableExpression.js
```diff
@@ -28,20 +33,22 @@ export default class EditableExpression extends React.Component {
...
-    displayRawEditor: !this.state.displayRawEditor
-  })
-
-  switchable = (editorName, expressionObj) => {
-    switch (editorName) {
-      case Types.BOOL_EDITOR:
-        return true
-      case Types.RAW_EDITOR:
-        return editorType.switchableToBooleanEditor(expressionObj)
+  editorName = (fieldType, expressionObj, displayRawEditor) => {
```
mproch: Let's just iterate on editors instead of switch? Only with some RawEditor special handling?

ui/client/components/graph/node-modal/editors/expression/EditorType.js
PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/EditableExpression.js
PPiedel: Done

ui/client/components/graph/node-modal/editors/map/MapValue.js
```diff
@@ -16,8 +16,8 @@ export default function MapValue(props) {
         showSwitch={showSwitch}
         onValueChange={(value) => onChange(`${paths}.expression.expression`, value)}
         expressionObj={value}
-        rowClassName={null}
-        valueClassName={null}
+        rowClassName={" "}
```
lciolecki: Why empty?

PR #563, opened 2020-01-03
Improvements: Remove handlerErrors example params from models

lciolecki:
We don't need example params at errorHandler:
error-handler-params
After changes:
error-handler-no-params


engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -225,7 +225,8 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
     )
   }
 
-  override def exceptionHandlerFactory(config: Config) = ParamExceptionHandler
```
mproch: I would leave it in dev. Sometimes we want to test  e.g. parameter validation or sth?

PR #573, opened 2020-01-08
Moving modal on grabbing its name label (#482)

### LINKED ISSUE #482: draggable modal fix
1) moving modal should be possible only by grabbing its name label (not subtype). Hovering over it should change cursor to open hand
2) while moving modal cursor should change to hand grabbing
3) we should eliminate stuttering if possible

mateuszkp96:
This closes #482 

Modal is now draggable with hover on modal header title. Modals without header, ConfirmDialog and PreventExitDialog are no longer draggable. 

On hover outside header:
image

On hover header:
image

On drag:
image

EdgeDetailsModal:
image


lciolecki:
Maybe better will be than ege bar looks like others? I mean will be smaller?

mateuszkp96:
Now EdgeDetailsModal bar looks like others
image

PR #574, opened 2020-01-08
Fix React warnings in properties modal

### LINKED ISSUE #571: Warnings in react console
https://staging.nussknacker.io/visualization/Test123?businessView=false,  in Properties modal:

- VM51 react.development.js:188 Warning: Each child in a list should have a unique "key" prop.
- Warning: Failed prop type: Invalid prop `value` of type `number` supplied to `LabeledInput`, expected `string`.
- Warning: Failed prop type: Invalid prop `value` of type `number` supplied to `Input`, expected `string`.

PPiedel:
This resolves #571

### DIFF
--- a/ui/client/components/graph/node-modal/editors/field/Input.js
+++ b/ui/client/components/graph/node-modal/editors/field/Input.js
@@ -10,11 +10,10 @@ export const Input = (props) => {
     <div className={className}>
       <div className={isMarked ? " marked" : ""}>
         {
-          readOnly ?
-            <div className="read-only" title={value}>{value}</div> :
             <input
               autoFocus={autoFocus}
               type="text"
+              readOnly={readOnly}
               placeholder={placeholder}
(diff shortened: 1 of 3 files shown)


ui/client/components/graph/node-modal/editors/field/Input.js
```diff
@@ -29,7 +28,10 @@ export const Input = (props) => {
 
 Input.propTypes = {
   isMarked: PropTypes.bool,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([
```
mproch: I wonder if it's enough? E.g. when we happen to use boolean or sth?

PPiedel: There is a separate component for boolean (Checkbox)

mproch: true

PR #588, opened 2020-01-10
GenericFlinkSource - source can define more complex behaviour

mproch:


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
```diff
@@ -1,17 +1,28 @@
...
+import pl.touk.nussknacker.engine.api.{MetaData, MethodToInvoke}
 import pl.touk.nussknacker.engine.api.process.{Source, SourceFactory}
-import pl.touk.nussknacker.engine.api.test.TestDataParser
 
 /**
   * Source with methods specific for Flink
   * @tparam T - type of event that is generated by this source. This is needed to handle e.g. syntax suggestions in UI
   */
-trait FlinkSource[T] extends Source[T] {
+trait BaseFlinkSource[T] extends Source[T] {
```
arkadius: Please swap names: FlinkSource <> BaseFlinkSource and add the same abstraction layers for tests

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
```diff
@@ -20,8 +31,23 @@ trait FlinkSource[T] extends Source[T] {
 
   def timestampAssigner : Option[TimestampAssigner[T]]
 
-}
+  override def startStream(env: StreamExecutionEnvironment, metaData: MetaData): DataStream[T] = {
```
arkadius: WDYT about rename it to `sourceStream`? WDYT about extracting metaData to implicit parameters - like it is in other places in our API?

mproch: done :+1:

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
mproch: :+1:  I swapped names, tests are a bit clumsy for now but I don't have really elegant idea for now..

arkadius: Looks good. If you already have changed this api, can you also rename `toFlinkSource`->`flinkSource` to clarify that objects of this class are not converted to flink source but flink source is used for preparing DataStream?

PR #594, opened 2020-01-13
Hotfix: Rename process

lciolecki:


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/processdetails.scala
```diff
@@ -75,6 +75,7 @@ object processdetails extends JavaTimeEncoders with JavaTimeDecoders {
 
     def isDeployed: Boolean = deployment.exists(_.isDeployed)
     def isCanceled: Boolean = deployment.exists(_.isCanceled)
+    def isNotWorking: Boolean = deployment.isEmpty || isCanceled
```
mproch: but it's just !isDeployed?

lciolecki: You are right ;)

PR #600, opened 2020-01-14
Make snapshot and cancel process in single action

jedrz:
During Flink upgrade it might be wise to snapshot job's state and cancel it in single action to avoid processing duplicates when restoring the job on new version of cluster. It is currently possible in two steps - making a snapshot and then canceling process so some duplicates are possible.

This PR also fixes making a snapshot with custom directory - it is ignored currently.

In next change redeploying a process could be improved by making a snapshot and canceling the job simultaneously.

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
@@ -16,7 +16,9 @@ trait ProcessManager {
   def findJobStatus(name: ProcessName) : Future[Option[ProcessState]]
 
   //TODO: this is very flink specific, how can we handle that differently?
-  def savepoint(name: ProcessName, savepointDir: Option[String], cancelProcess: Boolean): Future[String]
+  def savepoint(name: ProcessName, savepointDir: Option[String]): Future[SavepointResult]
+
+  def stop(name: ProcessName, savepointDir: Option[String]): Future[SavepointResult]
 
   def cancel(name: ProcessName) : Future[Unit]
 
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
@@ -16,7 +16,7 @@ trait ProcessManager {
   def findJobStatus(name: ProcessName) : Future[Option[ProcessState]]
 
   //TODO: this is very flink specific, how can we handle that differently?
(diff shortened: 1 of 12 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
```diff
@@ -16,7 +16,7 @@ trait ProcessManager {
   def findJobStatus(name: ProcessName) : Future[Option[ProcessState]]
 
   //TODO: this is very flink specific, how can we handle that differently?
-  def savepoint(name: ProcessName, savepointDir: String): Future[String]
+  def savepoint(name: ProcessName, savepointDir: Option[String], cancelProcess: Boolean): Future[String]
```
arkadius: What do you think about always making savepoint during cancel, and rename it to stop? Flink introduced second operation, stop: https://ci.apache.org/projects/flink/flink-docs-stable/ops/cli.html This concept will be close to their approach.

Proposed change will make this api more complex (three operations and two of them hidden in one method with flag).

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
```diff
@@ -117,11 +116,11 @@ class ManagementResources(processCounter: ProcessCounter,
     }
 
   def securedRoute(implicit user: LoggedUser): Route = {
-    path("adminProcessManagement" / "snapshot" / Segment / Segment) { (processName, savepointDir) =>
-      (post & processId(processName)) { processId =>
+    path("adminProcessManagement" / "snapshot" / Segment) { processName =>
+      (post & processId(processName) & parameters('savepointDir.?, 'cancelProcess.as[Boolean] ? false)) { (processId, savepointDir, cancelProcess) =>
```
lciolecki: I think this should be send in body not as query params. Wht do you think?

jedrz: I left savepointDir as query param since it's an optional parameters and being a query param is it much simpler to omit it when making the request. Sending empty body is not necessary when you would like to stick to the default value.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
jedrz: Done. I left savepointDir parameter because it can be useful when performing Flink upgrade to save savepoints temporarily in other directory.

arkadius: Indeed, can be useful :+1:

PR #609, opened 2020-01-15
Add possibility to configure editor for parameter

PPiedel:
Configuration and FE support will be added as separate PRs

### DIFF
--- a/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
+++ b/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
@@ -1,0 +1,15 @@
+package pl.touk.nussknacker.engine.api.editor;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SimpleEditor {
+
+    SimpleEditorType type();
+
+    LabeledExpression[] possibleValues() default {};
+}
--- a/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
+++ b/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
@@ -11,5 +11,8 @@ public @interface SimpleEditor {
 
     SimpleEditorType type();
 
+    /**
+     * This field should be used only with {@link SimpleEditorType#FIXED_VALUES_EDITOR}
+     */
     LabeledExpression[] possibleValues() default {};
 }
--- a/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditorType.java
+++ b/engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditorType.java
@@ -1,0 +1,5 @@
+package pl.touk.nussknacker.engine.api.editor;
+
+public enum SimpleEditorType {
+    BOOL_EDITOR, STRING_EDITOR, FIXED_VALUES_EDITOR
+}
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
@@ -38,7 +38,7 @@ case object RawParameterEditor extends ParameterEditor
 @JsonCodec(encodeOnly = true) case class SimpleParameterEditor(simpleEditorType: SimpleEditorType, possibleValues: List[FixedExpressionValue]) extends ParameterEditor
 
 object SimpleParameterEditor {
-  implicit val simpleEditorTypeEncoder: Encoder[SimpleEditorType] = simpleEditorType => Encoder.encodeString(simpleEditorType.name())
+  implicit val simpleEditorTypeEncoder: Encoder[SimpleEditorType] = (simpleEditorType : SimpleEditorType) => Encoder.encodeString(simpleEditorType.name())
 }
 
 @JsonCodec(encodeOnly = true) case class DualParameterEditor(simpleEditor: SimpleParameterEditor, defaultMode: DualEditorMode) extends ParameterEditor
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
@@ -1,8 +1,10 @@
 package pl.touk.nussknacker.engine.api.definition
 
+import io.circe.Encoder
 import io.circe.generic.JsonCodec
 import io.circe.generic.extras.ConfiguredJsonCodec
 import pl.touk.nussknacker.engine.api.LazyParameter
+import pl.touk.nussknacker.engine.api.editor.{DualEditorMode, SimpleEditorType}
 import pl.touk.nussknacker.engine.api.typed.ClazzRef
 import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypingResult}
 import pl.touk.nussknacker.engine.api.CirceUtil._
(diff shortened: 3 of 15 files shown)


engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/DefaultEditorMode.java
```diff
@@ -0,0 +1,5 @@
+package pl.touk.nussknacker.engine.api.editor;
+
+public enum DefaultEditorMode {
```
arkadius: I think that it should be named `DualEditorMode` which is used to determine `defaultMode` for `DualEditor`. WDYT?

engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditorType.java
```diff
@@ -0,0 +1,5 @@
+package pl.touk.nussknacker.engine.api.editor;
+
+public enum SimpleEditorType {
```
arkadius: Can't it be a `SimpleEditor` interface with two singleton implementations? We will be ready for `EnumEditor` extending `SimpleEitor` having `possibleValues` field. How you will do it with current implementation? Mayby a good thing will be to add this `EnumEditor` now?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -21,13 +23,30 @@ case class Parameter(name: String,
...
                      additionalVariables: Map[String, TypingResult] = Map.empty,
                      branchParam: Boolean = false) extends NodeDependency {
 
   def isLazyParameter: Boolean = classOf[LazyParameter[_]].isAssignableFrom(runtimeClass)
 
 }
 
+@ConfiguredJsonCodec(encodeOnly = true) sealed trait ParameterEditor
+
+@JsonCodec(encodeOnly = true) case class RawParameterEditor() extends ParameterEditor
```
arkadius: You shouldn't use `case class` with empty list of parameters - you have a `case object` for this.

ui/server/src/test/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsSpec.scala
```diff
@@ -56,17 +72,39 @@ class UIProcessObjectsSpec extends FunSuite with Matchers {
...
       .map(_.node.asInstanceOf[Enricher].service.parameters) shouldBe Some(List(Parameter("param", Expression("spel", "'default value'")),
-      Parameter("param2", Expression("spel", "'a'"))))
+      Parameter("param2", Expression("spel", "'a'")),
+      Parameter("param3", Expression("spel", "''")),
+      Parameter("param4", Expression("spel", "''"))
+    ))
 
   }
 
+  test("should read editor from config") {
```
arkadius: 1. Can you also add some test in `*ResourcesSpec` with returned json checking? It will show how api looks like and will test feature end-to-end (also json encoding).
2. Can you add some editor to our sample model that we can test it manually? Take a look at `DevProcessConfigCreator.SimpleTypesService`

engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditorType.java
mproch: Why do we need to specify editor here? I mean, it doesn't make sense to have STRING_EDITOR for boolean type etc. anyway? We can detect it via type - either in BE or FE. But currently I don't see where I will be able to really **decide** which simple editor to use?

arkadius: e.g. simple editor for strings - one option: current string editor, second option: editor with list of options (currently implemented as restrictions). WDYT about this case? We are planning to replace restrictions api with current more elastic, low level api. I agree that developer can introduce but using bad editor but ... this is developer, not an end user, it is minor problem, major problem is to give him flexibility...

arkadius: ... another option: editor for booleans: with true/false (for flag parameters), second option: kind of query builder.

mproch: Real flexibility will come only with ability to add own editors :) But I get your point... Maybe let's just add TODO for some validation?

arkadius: I agree with validations. We with @PPiedel considering being even more flexible - by replacing type enum with String and adding kind of key-value map for further extensions of editor created in "user space" but we come to conclusion that it could be done when such an editors appear. WDYT?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/EditorExtractor.scala
```diff
@@ -0,0 +1,24 @@
...
+import pl.touk.nussknacker.engine.api.definition._
+import pl.touk.nussknacker.engine.api.editor.{DualEditor, RawEditor, SimpleEditor}
+
+object EditorExtractor {
+
+  def extract(param: Parameter): Option[ParameterEditor] = {
+    val dualEditorAnnotation = param.getAnnotation(classOf[DualEditor])
+    val simpleEditorAnnotation = param.getAnnotation(classOf[SimpleEditor])
+    val rawEditorAnnotation = param.getAnnotation(classOf[RawEditor])
+    if (dualEditorAnnotation != null)
```
lciolecki: You can use there match pattern it will look better :)

engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/DefaultEditorMode.java
PPiedel: Done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala

ui/server/src/test/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsSpec.scala
PPiedel: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/EditorExtractor.scala
PPiedel: Done

engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SimpleEditor {
+
+    SimpleEditorType type();
+
+    LabeledExpression[] possibleValues() default {};
```
arkadius: Can you put here a note that this field is for `FIXED_VALUES_EDITOR`?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -21,13 +23,38 @@ case class Parameter(name: String,
...
 
   def isLazyParameter: Boolean = classOf[LazyParameter[_]].isAssignableFrom(runtimeClass)
 
 }
 
+@ConfiguredJsonCodec(encodeOnly = true) sealed trait ParameterEditor
+
+case object RawParameterEditor extends ParameterEditor
+
+@JsonCodec(encodeOnly = true) case class SimpleParameterEditor(simpleEditorType: SimpleEditorType, possibleValues: List[FixedExpressionValue]) extends ParameterEditor
```
arkadius: Can you rename `simpleEditorType` to `type` like it is in annotation API?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -21,13 +23,38 @@ case class Parameter(name: String,
...
+        new Encoder[SimpleEditorType] {
+          override def apply(editorType: SimpleEditorType): Json =  Encoder.encodeString(editorType.name())
+        }
+  }
+}
+
+@JsonCodec(encodeOnly = true) case class DualParameterEditor(simpleEditor: SimpleParameterEditor, defaultMode: DualEditorMode) extends ParameterEditor
+
+object DualParameterEditor {
+  implicit val defaultEditorModeEncoder: Encoder[DualEditorMode] = {
```
arkadius: dualEditorModeEncoder

ui/server/src/test/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsSpec.scala
arkadius: nice :+1:

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -491,9 +492,20 @@ class SimpleTypesCustomStreamTransformer extends CustomStreamTransformer with Se
 // In services all parameters are lazy evaluated
 class SimpleTypesService extends Service with Serializable {
   @MethodToInvoke
-  def invoke(@ParamName("booleanParam") booleanParam: Boolean,
-             @ParamName("stringParam") string: String,
-             @ParamName("intParam") intParam: Int,
+  def invoke(@ParamName("booleanParam")
```
arkadius: Can you also add parameter with fixed values? So in FE change won't be needed to change this definition.

engine/api/src/main/java/pl/touk/nussknacker/engine/api/editor/SimpleEditor.java
PPiedel: Done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
PPiedel: Done

PR #611, opened 2020-01-15
Fix: canceled state

lciolecki:
- Fix problem when process was canceled and manager doesn't return status for it
- Fix bug with comments at edit process
- Fix bug with comparing version when process is not deployed

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
@@ -5,7 +5,6 @@ import java.time.LocalDateTime
 import akka.actor.{ActorRefFactory, Props, Status}
 import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.ProcessingTypeData.ProcessingType
-import pl.touk.nussknacker.engine.api.ProcessVersion
 import pl.touk.nussknacker.engine.api.deployment.TestProcess.TestData
 import pl.touk.nussknacker.engine.api.deployment._
 import pl.touk.nussknacker.engine.api.deployment.simple.SimpleStateStatus
@@ -14,7 +13,7 @@ import pl.touk.nussknacker.engine.marshall.ProcessMarshaller
 import pl.touk.nussknacker.ui.listener.ProcessChangeEvent.{OnDeployActionFailed, OnDeployActionSuccess}
 import pl.touk.nussknacker.restmodel.displayedgraph.{DisplayableProcess, ProcessStatus}
 import pl.touk.nussknacker.restmodel.process.{ProcessId, ProcessIdWithName}
-import pl.touk.nussknacker.restmodel.processdetails.{DeploymentAction, DeploymentHistoryEntry}
+import pl.touk.nussknacker.restmodel.processdetails.{ProcessDeploymentAction}
 import pl.touk.nussknacker.ui.EspError
 import pl.touk.nussknacker.ui.db.entity.{DeployedProcessInfoEntityData, ProcessVersionEntityData}
 import pl.touk.nussknacker.ui.listener.ProcessChangeListener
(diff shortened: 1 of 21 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -116,9 +117,16 @@ class ManagementActor(environment: String,
       reply(Future.successful(DeploymentStatusResponse(beingDeployed)))
   }
 
+  private def handleCheckStatus(processState: Option[ProcessState], deployedVersion: Option[Long], lastDeployHistory: Option[DeploymentHistoryEntry]): Option[ProcessStatus] =
```
arkadius: What does mean this method name? You handle some checking? What checking? Maybe `handleObsoleteManagerState`? This code is very Flink-hermetic, we should add very verbose comment here e.g.:
```
Some ProcessManagers (e.g. Flink's) lost information about canceled state after some duration. In case of this we use state from our cached history of states. TODO: we should move this ManagementActor behind ProcessManager API and do this logic only in Flink's case.
```

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -187,7 +187,7 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](val dbConfig: DbConfig)
       isLatestVersion = isLatestVersion,
       tags = tags,
       history = processVersions.map(pvs => ProcessRepository.toProcessHistoryEntry(
-        process, pvs, deployments.filterNot(_.isCanceled).toList //We want't canceled deployments here
+        process, pvs, deployments.toList
```
arkadius: This change is really safe?

lciolecki: Yep, we want to have all version actions. And this doesn't change anything at FE.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
lciolecki: Done.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -94,13 +94,15 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](val dbConfig: DbConfig)
   private def fetchProcessDetailsByQueryAction[PS: ProcessShapeFetchStrategy](query: ProcessEntityFactory#ProcessEntity => Rep[Boolean],
                                                                               isDeployed: Option[Boolean])(implicit loggedUser: LoggedUser, ec: ExecutionContext): DBIOAction[List[BaseProcessDetails[PS]], NoStream, Effect.All with Effect.Read] = {
     (for {
-      deployments <- fetchLastDeploymentActionPerProcess.result
+      deployments <- fetchLastActionPerProcess.result
```
arkadius: It is not `deployments` anymore, but `lastAction`, you can also change `latestDeployed` to `lastDeployedAction` below.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -30,7 +30,11 @@ trait ProcessRepository[F[_]] extends Repository[F] with EspTables {
       .filter(_.processId === processId.value)
       .sortBy(_.createDate.desc)
 
-  protected def fetchLastDeploymentActionPerProcess: Query[(Rep[Long], ProcessDeploymentInfoEntityFactory#ProcessDeploymentInfoEntity), (Long, DeployedProcessInfoEntityData), Seq] =
+
+  protected def fetchLastDeployedActionPerProcess: Query[(api.Rep[Long], ProcessDeploymentInfoEntityFactory#ProcessDeploymentInfoEntity), (Long, DeployedProcessInfoEntityData), Seq] =
```
arkadius: Can you change names of methods that return `Query` to be suffixed by `Query` e.g. `fetchLastDeployedActionPerProcessQuery`. I know that it was that way before, but now it looks like some fetching is done in this method but it is only building a query...

lciolecki: It's good idea. I will change it.

PR #616, opened 2020-01-16
Custom (pluginable) icons for nodes 

arkadius:
For now on the left panel and inside displayed graph, in pdfs still default once.


lciolecki:
Please change all touched js file to tsx :)

PR #622, opened 2020-01-19
Improvements: Cleanup states

lciolecki:
*Business Logic*
- From now we use ProcessAction instead ProcessDeploy
- From now we use ProcessVersion instead ProcessHistory*

*DB Changes*
- rename table process_deployment_info to process_actions
- rename column deployment_action to action
- rename column deploy_at to performed_at
- drop column environment // we don't support it now
- TODO: remove environments table and all depenedencies

*Rest Model Changes*
-  Now we support two models: ProcessVersion and ProcessAction
- BaseProcessDetails.id temporary keeps process name, because GUI uses it 
- BaseProcessDetails.processId temporary keeps process id - some BE stuff use it
- TODO: add support id as ProcessId and remove processId when we will support ProcessId
- TODO: Add Comment Attachment and DTO's
- TODO: ProcessVersion and ProcessAction should support processId as ProcessId

*Queries changes*
- separation queries from fetch methods
- normalization ProcessActionEntity queries, now we allways return action with comment


ui/server/src/main/resources/db/migration/postgres/V1_027__ProcessActionCleanUp.sql
```diff
@@ -0,0 +1,17 @@
+-- DROP CONSTRAINTS
```
mproch: Are they the same? Maybe we can move it to common?

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessActionEntityFactory.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  implicit def deploymentMapper: JdbcType[ActionType] with BaseTypedType[ActionType] =
+    MappedColumnType.base[ActionType, String](_.toString, ActionType.withName)
+
+  val processActionsTable: LTableQuery[ProcessActionEntityFactory#ProcessActionEntity] =
+    LTableQuery(new ProcessActionEntity(_))
+
+  val processVersionsTable: LTableQuery[ProcessVersionEntityFactory#ProcessVersionEntity]
+  val commentsTable: LTableQuery[CommentEntityFactory#CommentEntity]
+  val environmentsTable: LTableQuery[EnvironmentsEntityFactory#EnvironmentsEntity]
```
mproch: is it currently used? maybe TODO for remove?

lciolecki: Removed.

PR #628, opened 2020-01-20
Handle editor configuration on frontend

PPiedel:


PPiedel:
Due to multiple errors with external libraries like lodash / moment I decided not to change files to tsx right now

PR #629, opened 2020-01-20
Fix: FE bugs

### LINKED ISSUE #619: Cancel button in custom processes is badly positioned
Screenshot from 2020-01-17 11-07-25

lciolecki:
- Resolves #619 
- OAuth2 infinite redirects


ui/client/common/SystemUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public getUserId = (): string => {
+    if (_.isEmpty(localStorage.getItem(userId))) {
+      localStorage.setItem(userId, uuid4())
+    }
+    return localStorage.getItem(userId)
+  }
+
+  public getAuthorizationHeader = (): string => AUTHORIZATION_HEADER_NAMESPACE
```
arkadius: Why this is a method instead of some kind of static constant?

lciolecki: Done.

PR #630, opened 2020-01-21
Metrics refactor - first stage, with legacy modes

mproch:
This is first step of cleaning/refactor of metrics. We add explicit tags, more documentation and some utility functions.
- This version *is* compatible with old metrics for Flink mode - new metrics are opt-in
- This version *is not* compatible with old metrics for standalone mode

Please see Metrics.md for docs and MetricUtils for Flink metrics compatibility


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/metrics/MetricUtils.scala
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ServiceLoader
+
+import cats.data.NonEmptyList
+import org.apache.flink.api.common.functions.RuntimeContext
+import org.apache.flink.metrics.{Counter, Gauge, Histogram, Meter, MetricGroup}
+import pl.touk.nussknacker.engine.flink.api.RuntimeContextLifecycle
+
+//Configure this as Service to use new reporting for Flink
+//this is a bit weird way, but it's temporary and passing configuration would be cumbersome
+trait UseNewMetrics
```
arkadius: Nice trick for now :+1: I wonder if in a near feature we can change deployment to use new flink rest api handling complex arguments. Then we can pass some environment vars and use it as a low level properties similar to `System.getProperty`

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/metrics/MetricUtils.scala
```diff
@@ -0,0 +1,110 @@
...
+
+/*
+  IMPORTANT: PLEASE keep Metrics.md up to date
+
+  Handling Flink metrics is a bit tricky. For long time we parsed tags directly in Influx, via graphite plugin
+  This is complex and error prone, so we'd like to switch to passing flink metric variables as tags via native influx API
+  Unfortunately, current Flink Influx report doesn't allow for elastic configuration, so for now
+  we translate by default to `old` way in groupsWithNameForLegacyMode
+ */
+class MetricUtils(runtimeContext: RuntimeContext) {
```
arkadius: I don't like generic *Utils - this name tempting to put here whatever not matches anywere. Maybe better name would be `MetricsCompat`? It is not critical notice, we can merge it with this name also

docs/Metrics.md
```diff
@@ -11,3 +11,69 @@ For each process following metrics are collected:
 * http services' invocation times, errors and throughput
 * event processing delays
 
+
+Metrics technical details
```
arkadius: Nice docs, appreciate this :+1:

PR #635, opened 2020-01-23
Refactor: Map Process Action to State

lciolecki:
*Technical changes*
- Add field state in Process DTO's 
- Add mapActionToStatus method in ProcessStateDefinitionManager, for now we use it to mock ProcessStatus based on ProcessAction
- At list of process we fill state field by mapping action to status
- Some tests improvements
- Changes at FE ListState - now we assume that all data is from process.state or loaded processState.

*Refactor changes*
- Rename ProcessAction to ProcessActionType in from engine/api/deployment 
- Remove ProcessAction enum from restmodel and replace it by ProcessActionType, nowe we have only one implementation of ProcessAction.
- FE: TypeScript adaptation


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
```diff
@@ -54,12 +49,17 @@ object ProcessStatus {
       errorMessage
     )
 
-  def create(processState: ProcessState, expectedDeploymentVersion: Option[Long]): ProcessStatus = {
-    val versionMatchMessage = (processState.version, expectedDeploymentVersion) match {
-      case (Some(stateVersion), Some(expectedVersion)) if stateVersion.versionId == expectedVersion => None
-      case (Some(stateVersion), Some(expectedVersion)) if processState.isDeployed => Some(s"Process deployed in version ${stateVersion.versionId} (by ${stateVersion.user}), expected version $expectedVersion")
+  def isStateAfterDeploy(processState: ProcessState): Boolean =
```
arkadius: What does it mean state is after deploy? Maybe some better verb will help like `isFollowingDeployAction`. Please move it to `ProcessState`. It is one more method describing lifecycle that we should remember about if we want to make this lifecycle more elastic.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
```diff
@@ -54,12 +49,17 @@ object ProcessStatus {
...
+
+  def create(processState: ProcessState, lastAction: Option[ProcessDeploymentAction]): ProcessStatus = {
+    //TODO: Move this logic to another place.. This should be moved together with ManagementActor.handleObsoleteStatus
+    val versionMatchMessage = (processState.processVersionId, lastAction) match {
+      case (Some(stateVersion), Some(action)) if stateVersion.versionId == action.processVersionId => None
+      case (Some(stateVersion), Some(action)) if action.isDeployed && !isStateAfterDeploy(processState) => Some(s"Process deployed in version ${stateVersion.versionId} (by ${stateVersion.user}), but currently not working")
+      case (Some(stateVersion), Some(action)) if action.isDeployed && stateVersion.versionId != action.processVersionId => Some(s"Process deployed in version ${stateVersion.versionId} (by ${stateVersion.user}), expected version ${action.processVersionId}")
       case (Some(stateVersion), None) if processState.isDeployed => Some(s"Process deployed in version ${stateVersion.versionId} (by ${stateVersion.user}), should not be deployed")
-      case _  => None
+      case _ => None
```
arkadius: What about the `case (None, Some(action))`? Why we not need to handle cases like `case (None, Some(action)) if !action.isDeployed`?

If this message is only for deployed case, WDYT about extracting it to method like `deployedVersionMismatchMessage`? and change local variable name in the same manner.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -382,6 +384,17 @@ class ProcessesResources(val processRepository: FetchingProcessRepository[Future
     GraphProcess(ProcessMarshaller.toJson(emptyCanonical).noSpaces)
   }
 
+  //This is temporary function to enriching process status data
+  //TODO: Remove it when we will support cache for state
+  private def enrichProcess(process: BaseProcessDetails[_]): BaseProcessDetails[_] =
```
arkadius: `enrichDetailsWithProcessState`?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
lciolecki: Changed to isFollowingDeployAction.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
lciolecki: Changed.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
lciolecki: Changed.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
arkadius: What about 2nd part:

> Please move it to ProcessState. It is one more method describing lifecycle that we should remember about if we want to make this lifecycle more elastic.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/ProcessesResourcesSpec.scala
```diff
@@ -233,27 +237,36 @@ class ProcessesResourcesSpec extends FunSuite with ScalatestRouteTest with Match
...
     Get(s"/processes?isDeployed=false") ~> routeWithAllPermissions ~> check {
       status shouldEqual StatusCodes.OK
       responseAs[List[Json]].size shouldBe 2
+
       val process = findJsonProcess(responseAs[String], thirdProcessor.value)
-      process.flatMap(_.name) shouldBe Option.empty
+      process.map(_.name) shouldBe Option.empty
+
+      val canceledProcess = findJsonProcess(responseAs[String], secondProcessor.value)
+      canceledProcess.flatMap(_.stateStatus) shouldBe Some(SimpleStateStatus.Canceled.name)
```
arkadius: can you use: `canceledProcess.value.stateStatus.value...`? or even split it to two lines:
```
val canceledProcess = findJsonProcess(responseAs[String], secondProcessor.value).value
canceledProcess.stateStatus.value...
```
(In all places where you used flatMap in checks)
Using current syntax after fail you can't recognize what was wrong (two checks are flatten to one)

lciolecki: Great thinks - done.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
lciolecki: You'r right :) Done.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -18,8 +18,15 @@ trait ProcessStateDefinitionManager {
 object ProcessState {
   import io.circe.syntax._
 
-  implicit val statusEncoder: Encoder[StateStatus] = Encoder.encodeJson.contramap(st => StateStatusCodec(st.getClass.getName, st.name).asJson)
-  implicit val statusDecoder: Decoder[StateStatus] = Decoder.decodeNone.map(st => SimpleStateStatus.Unknown) //TODO: Add decode implementation by clazz and value. At now we don't need it.
+  implicit val statusEncoder: Encoder[StateStatus] = Encoder.encodeJson.contramap(st => StateStatusCodec(st.getClass.getSimpleName, st.name).asJson)
+  implicit val statusDecoder: Decoder[StateStatus] = Decoder.decodeJson
```
arkadius: I don't think that you should invoke `json.as` inside Decoder. It is rather end-user api. You should rather do `Decoder[StateStatusCodec].map(StateStatus.codecToStateStatus)`. In your approach you will probably end with some strange exception during decoding in case of `Left(exception)`...

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/SimpleProcessStateSpec.scala
```diff
@@ -26,4 +29,38 @@ class SimpleProcessStateSpec extends FunSpec with Matchers with Inside {
...
+    val json = s"""
+      {
+        "clazz": "${SimpleStateStatus.Canceled.getClass.getSimpleName}",
+        "value": "${SimpleStateStatus.Canceled.name}"
+      }
+    """
+
+    val decodedStatus = parser.decode[StateStatus](json) match {
+      case Right(status) => status
+      case Left(error) => throw error
```
arkadius: You shouldn't throw an exception in tests. It will cause test abortion instead of failure. You should use `EitherValues` with `.right.value`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -75,6 +85,30 @@ sealed trait StateStatus {
...
+    classOf[RunningStateStatus]
+  )
+    .map(clz => clz.getSimpleName -> clz)
+    .toMap[String, Class[_ <: StateStatus]]
+  
+  def codecToStateStatus(codec: StateStatusCodec): StateStatus =
+    availableStatusClasses
+      .get(codec.clazz)
+      .map(_.getConstructor(classOf[String]).newInstance(codec.value))
+      .getOrElse(SimpleStateStatus.Unknown)
```
arkadius: I see two problems here:
1. I don't think that `getOrElse(SimpleStateStatus.Unknown)` is a good solution for validation handling. why not to return Either with some decoding error in this place? How consumer of this api will now that he/she pass invalid string?
2. During `getConstructor(classOf[String]).newInstance` can be thrown lot of strange errors. Maybe would be better to catch them and throw something more meaningful?

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/SimpleProcessStateSpec.scala
lciolecki: Done.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
lciolecki: Hmm, you'a right. Changed.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -75,6 +76,39 @@ sealed trait StateStatus {
...
+      .map(clazz =>
+        Try(clazz.getConstructor(classOf[String]).newInstance(codec.value)) match {
+          case Failure(exception) => Left(s"Failed to decode StateStatus. Error: ${exception.getMessage}.")
+          case Success(stateStatus) => Right(stateStatus)
+        }
+      )
+      .getOrElse(Left(s"Failed to decode StateStatus. Error: class ${codec.clazz} doesn't exist."))
+  )
+
+  //This field keeps all available statuses. Remember!! If you want add new status class you have to add it also here!
```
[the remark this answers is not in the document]
arkadius: It is not so easy. It is abstract factory pattern. We need to somehow connect factory (now implemented as reflection call) with class representing created value (ProcessState).

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -75,6 +78,11 @@ sealed trait StateStatus {
   def name: String
 }
 
+trait StateStatusFollowingDeployAction {
```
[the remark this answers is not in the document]
arkadius: Yep this another layer of abstraction doesn't help

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/SimpleProcessStateSpec.scala
```diff
@@ -1,29 +1,80 @@
...
   }
 
   it ("process state should be finished") {
     val state = createProcessState(SimpleStateStatus.Finished)
     state.status.isFinished shouldBe true
-    state.allowedActions shouldBe List(StateAction.Deploy)
+    state.allowedActions shouldBe List(ProcessActionType.Deploy)
+  }
+
+  it ("should properly encode StateStatus") {
```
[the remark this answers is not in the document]
arkadius: Yep, `toString` is risky, pls try to use explicit functions like `asString.value`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -1,26 +1,27 @@
...
+import pl.touk.nussknacker.engine.api.deployment.StateStatus.availableStatusClasses
+
+import scala.util.{Failure, Success, Try}
 
 trait ProcessStateDefinitionManager {
-  def statusActions(stateStatus: StateStatus): List[StateAction]
+  def statusActions(stateStatus: StateStatus): List[ProcessActionType]
   def statusTooltip(stateStatus: StateStatus): Option[String]
   def statusIcon(stateStatus: StateStatus): Option[URI]
+  def mapActionToStatus(stateAction: Option[ProcessActionType]): StateStatus
```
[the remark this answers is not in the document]
lciolecki: Done.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -47,8 +50,8 @@ object ProcessState {
 
 @JsonCodec case class ProcessState(deploymentId: DeploymentId,
                                    status: StateStatus,
-                                   version: Option[ProcessVersion] = Option.empty,
-                                   allowedActions: List[StateAction] = List.empty,
+                                   processVersionId: Option[ProcessVersion] = Option.empty,
```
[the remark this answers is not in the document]
lciolecki: You,r right. Changed.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -1,26 +1,29 @@
...
   def statusIcon(stateStatus: StateStatus): Option[URI]
+  def mapActionToStatus(stateAction: Option[ProcessActionType]): StateStatus
 }
 
 object ProcessState {
   import io.circe.syntax._
 
   implicit val statusEncoder: Encoder[StateStatus] = Encoder.encodeJson.contramap(st => StateStatusCodec(st.getClass.getName, st.name).asJson)
-  implicit val statusDecoder: Decoder[StateStatus] = Decoder.decodeNone.map(_ => null) //TODO: Add decode implementation by clazz and value. At now we don't need it.
+  implicit val statusDecoder: Decoder[StateStatus] = Decoder.decodeNone.map(st => SimpleStateStatus.Unknown) //TODO: Add decode implementation by clazz and value. At now we don't need it.
```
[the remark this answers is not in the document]
lciolecki: Thanks for tip! It looks better!

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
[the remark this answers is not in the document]
lciolecki: Changed.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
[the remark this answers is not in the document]
lciolecki: Changed.

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/SimpleProcessStateSpec.scala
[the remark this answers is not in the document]
lciolecki: I've removed these test after I added circe native decode for seald trait.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -68,62 +66,39 @@ object ProcessActionType extends Enumeration {
...
 
-trait StateStatusFollowingDeployAction {
-  def isFollowingDeployAction(stateStatus: StateStatus): Boolean =
-    stateStatus.isDuringDeploy || stateStatus.isRunning || stateStatus.isFinished
+  def isFollowingDeployAction: Boolean =
+    isDuringDeploy || isRunning || isFinished
 }
 
-final class NotEstablishedStateStatus(val name: String) extends StateStatus
+final case class NotEstablishedStateStatus(val name: String) extends StateStatus
```
mproch: val is not necessary anymore?

PR #647, opened 2020-02-03
Fix to not deployed process

lciolecki:
*Fixes*
- Show proper StateStatus for not deployed process.

*Features*
- Show process info for isArchived and isSubprocess at process window
archived-process
subprocess


ui/client/components/right-panel/UserRightPanel.js
```diff
@@ -65,9 +66,12 @@ class UserRightPanel extends Component {
         <SpinnerWrapper isReady={isReady}>
 
           <Scrollbars renderThumbVertical={props => <div {...props} className="thumbVertical"/>} hideTracksWhenNotNeeded={true}>
-            
-            <ProcessState process={fetchedProcessDetails} processState={processState} isStateLoaded={isStateLoaded}/>
-            
+
+            { fetchedProcessDetails.isArchived || fetchedProcessDetails.isSubprocess ?
+              <ProcessInfo process={fetchedProcessDetails}/> :
```
mproch: it's kind of misleading when we have ProcessInfo only for subprocess/archive... I think it would look better if we put this logic inside ProcessState/StateIcon - they look more or less the same anyway?

lciolecki: Hmmm... Okey, I will change it to ProcessInfo component.

lciolecki: Done

ui/client/components/Process/ProcessInfo.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  private animationTimeout = {
+    enter: 500,
+    appear: 500,
+    exit: 500,
+  }
+
+  private animationListener = (node, done) => node.addEventListener("transitionend", done, false)
+
+  private getDescription = (process: ProcessType, processState: ProcessStateType, isStateLoaded: boolean): string =>
```
arkadius: Please add some little TODO, that logic from `getDescription` and `getProcessIcon` should be moved to BE

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -120,11 +120,11 @@ class ManagementActor(environment: String,
...
   //TODO: In future we should move this functionality to ProcessManager.
-  private def handleObsoleteStatus(processState: Option[ProcessState], processStateDefinitionManager: ProcessStateDefinitionManager, lastAction: Option[ProcessDeploymentAction]): Option[ProcessStatus] =
+  private def handleObsoleteStatus(processStateDefinitionManager: ProcessStateDefinitionManager, processState: Option[ProcessState], lastAction: Option[ProcessDeploymentAction]): ProcessStatus =
     (processState, lastAction) match {
-      case (Some(state), _)  => Option(ProcessStatus.create(state, lastAction))
-      case (None, Some(action)) if action.isCanceled => Option(ProcessStatus.canceled(processStateDefinitionManager))
-      case _ => Option.empty
+      case (Some(state), _)  => ProcessStatus.create(state, lastAction)
+      case (None, Some(action)) if action.isCanceled => ProcessStatus(SimpleStateStatus.Canceled, processStateDefinitionManager)
+      case _ => ProcessStatus(SimpleStateStatus.NotDeployed, processStateDefinitionManager)
```
arkadius: please explicitly handle `(None, None)`, there are some cases that we are mathing now as a not deployed but it isn't "perfect match" ;-)

lciolecki: You'r right, changed! ;)

PR #649, opened 2020-02-03
Fix: Bug with bussines view

lciolecki:
- Verify business view at start in process window
- Reload data after change business view

### DIFF
--- a/ui/client/http/HttpService.js
+++ b/ui/client/http/HttpService.js
@@ -113,9 +113,8 @@ export default {
   },
 
   fetchProcessDetails(processId, versionId, businessView) {
-    let url = versionId ? `/processes/${processId}/${versionId}` : `/processes/${processId}`
-    const queryParams = this.businessViewQueryParams(businessView)
-    return api.get(url, {params: queryParams})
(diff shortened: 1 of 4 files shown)


ui/client/http/HttpService.js
```diff
@@ -267,10 +266,9 @@ export default {
   },
 
   compareProcesses(processId, thisVersion, otherVersion, businessView, remoteEnv) {
-    const queryParams = this.businessViewQueryParams(businessView)
     const path = remoteEnv ? "remoteEnvironment" : "processes"
 
-    return api.get(`/${path}/${processId}/${thisVersion}/compare/${otherVersion}`, {params: queryParams})
+    return api.get(`/${path}/${processId}/${thisVersion}/compare/${otherVersion}`, {params: {businessView}})
```
mproch: can we get rid of this method altogether (businessViewQueryParams)?

lciolecki: Yes, we can! :)

lciolecki: Done.

PR #650, opened 2020-02-04
Pass user performing action to ProcessManager

mproch:


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -201,7 +201,7 @@ class ManagementActor(environment: String,
       deploymentResolved <- resolvedDeploymentData
       maybeProcessName <- processRepository.fetchProcessName(ProcessId(latestVersion.processId))
       processName = maybeProcessName.getOrElse(throw new IllegalArgumentException(s"Unknown process Id ${latestVersion.processId}"))
-      _ <- processManagerValue.deploy(latestVersion.toProcessVersion(processName), deploymentResolved, savepointPath)
+      _ <- processManagerValue.deploy(latestVersion.toProcessVersion(processName), deploymentResolved, savepointPath, user.username)
```
arkadius: I think that we should rather pass everywhere User Id instead of user name? User name may be not unique.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessManager.scala
```diff
@@ -9,7 +9,11 @@ import scala.concurrent.Future
 trait ProcessManager {
 
   //TODO: savepointPath is very flink specific, how can we handle that differently?
-  def deploy(processId: ProcessVersion, processDeploymentData: ProcessDeploymentData, savepointPath: Option[String]) : Future[Unit]
+  def deploy(processId: ProcessVersion, processDeploymentData: ProcessDeploymentData, savepointPath: Option[String], user: String) : Future[Unit]
```
arkadius: WDYT about using value objects pattern? Naming something "user" without adding name or id is the worst option ... ;-)

mproch: I added User object with id and name :)

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
mproch: I added User object with id and name :)

PR #654, opened 2020-02-04
Fix: Specification state errors

lciolecki:


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -120,11 +120,33 @@ class ManagementActor(managers: Map[ProcessingType, ProcessManager],
...
+      )
+      case (Some(_), Some(act)) if act.isDeployed && !processState.status.isFollowingDeployAction => ProcessStatus.error(
+        SimpleStateStatus.ErrorShouldRunning, processStateDefinitionManager, act.user, act.processVersionId, Option(processState)
+      )
+      case (Some(stateVersion), Some(action)) if stateVersion.versionId == action.processVersionId =>
+        ProcessStatus(processState)
+      case (Some(sv), None) if processState.isDeployed => ProcessStatus.error(
+        SimpleStateStatus.ErrorShouldNotBeDeployed, processStateDefinitionManager, sv.user, sv.versionId, Option(processState)
+      )
+      case _ => ProcessStatus(processState)
```
arkadius: What cases are here? Can you remove this line or compiler return "match is not exhaustive"? If so, maybe we can explicitly write them down and think a bit if they are really normal, correct cases. It they aren't we should return other state or throw exception. IMO e.g. `(None, Some(_))` isn't normal.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
```diff
@@ -33,6 +32,27 @@ object ProcessStatus {
...
+            user: String,
+            versionId: Long,
+            processState: Option[ProcessState] = Option.empty,
+            exceptedVersion: Option[Long] = Option.empty): ProcessStatus = {
+    ProcessStatus(
+      status,
+      processState.map(_.deploymentId.value),
+      allowedActions = processStateDefinitionManager.statusActions(status),
+      icon = processStateDefinitionManager.statusIcon(status),
+      tooltip = processStateDefinitionManager.statusTooltip(status).map(_.format(
```
arkadius: What will happen if someone will change the order of placeholders in tooltip? How developer would know what is the correct order? How developer will find out that order of variables was changed in some versions in a future?
You have status classes, you can put there useful info like version, you can use this version during tooltip preparing. You just need to make Simple*Manager more OO friendly - use methods instead of maps...

mproch: I agree. But there is also one more problem: we mix statuses from ProcessManager and ManagementActor. Here we assume that processStateDefinitionManager has to handle (some) statuses from SimpleProcessManager?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
```diff
@@ -34,7 +34,10 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
     SimpleStateStatus.DuringCancel -> "/assets/states/stopping-running-animated.svg",
     SimpleStateStatus.Failed -> "/assets/states/failed.svg",
     SimpleStateStatus.Finished -> "/assets/states/success.svg",
-    SimpleStateStatus.Error -> "/assets/states/error.svg"
+    SimpleStateStatus.Error -> "/assets/states/error.svg",
+    SimpleStateStatus.ErrorShouldRunning -> "/assets/states/error.svg",
+    SimpleStateStatus.ErrorShouldNotBeDeployed -> "/assets/states/error.svg",
+    SimpleStateStatus.ErrorMismatchVersion -> "/assets/states/error.svg"
```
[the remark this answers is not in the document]
mproch: No, it's other situation. It's the situation when we mark version A as deployed (not saved!) in NK DB, but e.g. Flink tells us version B is deployed. This is weird and should not happen - probably some error during deploy we didn't catch or sb messes up with Flink cluster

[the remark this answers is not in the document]
arkadius: Indeed now I see that. Let's change the name to `ErrorMismatchDeployedVersion` and add little comment there?

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
arkadius: We have TODO's that at the end we want to hide `ManagementActor` behind ProcessManager's API. But indeed, during this transition we can handle this situation better. Two ideas that I have:
1. Maybe we should pass to tooltip method some additional "deployment details" which will be used during formatting
2. Maybe we can extract "base statuses" from "simple statuses". In fact this cases are quite specific for situation when you have "remote engine" that can be desynchronized. Let's assume local engine (like in NiFi). In this scenario you won't have such statuses. Maybe we should call those extracted statuses "remote statuses"?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
```diff
@@ -79,6 +77,24 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
       .map(sa => actionStatusMap.getOrElse(sa, SimpleStateStatus.Unknown))
       .getOrElse(SimpleStateStatus.NotDeployed)
 
+  override def statusTooltip(stateStatus: StateStatus): Option[String] =
```
mproch: I wonder is it the best place for this? Maybe in ProcessStatus? Where we create state objects?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -120,11 +120,27 @@ class ManagementActor(managers: Map[ProcessingType, ProcessManager],
...
-      case (None, None) => ProcessStatus(SimpleStateStatus.NotDeployed, processStateDefinitionManager)
+      case (Some(state), _)  => handleMismatchStateVersion(state, lastAction)
+      case (None, Some(act)) if act.isCanceled => ProcessStatus.simple(SimpleStateStatus.Canceled)
+      case (None, Some(act)) if act.isDeployed => ProcessStatus.simpleErrorShouldRunning(act.processVersionId, act.user, Option.empty)
+      case (None, None) => ProcessStatus.simple(SimpleStateStatus.NotDeployed)
     }
 
+  //This method handles some corner cases for mismatch states version
+  //TODO: In future we should move this functionality to ProcessManager.
+  private def handleMismatchStateVersion(processState: ProcessState, lastAction: Option[ProcessAction]): ProcessStatus = {
```
arkadius: I still see here a lot not matched strange situations:
1. `case (None, _)` - AFAIK it is when old process has missing version in state - we should return `ProcessStatus(processState)` + add some comment
2. `case (Some(_), Some(act)) if act.isCanceled && stateVersion.versionId != action.processVersionId` - what will be then? 
Currently will throw `MatchError`. I wonder if we could handle situation that we already know and if unknown, strange situation occurs, return some unknown state and log it as a warning. WDYT?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
```diff
@@ -13,6 +13,8 @@ object SimpleStateStatus {
   val DuringCancel: NotEstablishedStateStatus = NotEstablishedStateStatus("DURING_CANCEL")
   val Canceled: StoppedStateStatus = StoppedStateStatus("CANCELED")
   val Failed: NotEstablishedStateStatus = NotEstablishedStateStatus("FAILED")
-  val Error: NotEstablishedStateStatus = NotEstablishedStateStatus("ERROR")
-  val Warning: NotEstablishedStateStatus = NotEstablishedStateStatus("WARNING")
+  val Error: ErrorStateStatus = ErrorStateStatus("ERROR")
+  val ErrorShouldRunning: ErrorStateStatus = ErrorStateStatus("ERROR_SHOULD_RUNNING")
```
arkadius: Those three is not used anymore?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
mproch: Good catches!
1. think it's not possible anymore - but if so we can probably remove Option from version..
2. I think in this method we can just ignore actions other than deploy (filter) and case Some(version), None should apply
3. And maybe we should return "generic" error in other cases?

arkadius: 1, 3 looks ok for me
Regards to 2. We can do it in this method (then we should also change method name to `handleMismatchDeployedStateVersion`), but on upper level (`handleObsoleteStatus`) we should handle `case (Some(_), Some(act)) if act.isCanceled`. It can be done be generalization existing case to `case (_, Some(act)) if act.isCanceled` or adding another one. Otherwise we will got MatchError there....

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
```diff
@@ -79,6 +77,24 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
...
+    statusTooltipsMap.get(stateStatus)
+
+  def errorShouldRunningTooltip(deployedVersionId: Long, user: String): String =
+    s"Process deployed in version ${deployedVersionId.toString} (by ${user}), but currently is not running!"
+
+  def errorShouldNotBeDeployedTooltip(deployedVersionId: Long, user: String): String =
+    s"Process deployed in version ${deployedVersionId.toString} (by ${user}), should not be deployed!"
+
+  def errorMismatchDeployedVersionTooltip(deployedVersionId: Long, exceptedVersionId: Long, user: String): String =
+    s"Process deployed in version ${deployedVersionId.toString} (by ${user}), expected version ${exceptedVersionId.toString}!"
```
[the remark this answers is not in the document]
lciolecki: Done.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
lciolecki: I will leave it now.. We'll clean it up in the future.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
lciolecki: Yep, good catch - thanks!

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -118,28 +118,52 @@ class ManagementActor(managers: Map[ProcessingType, ProcessManager],
...
-      case (Some(version), None) if processState.isDeployed =>
-        ProcessStatus.simpleErrorShouldNotBeDeployed(version.versionId, version.user, Some(processState))
-      case (Some(stateVersion), Some(action)) if stateVersion.versionId == action.processVersionId =>
-        ProcessStatus(processState)
+  private def handleFollowingDeployState(state: ProcessState, lastAction: Option[ProcessAction]): ProcessStatus =
+    lastAction match {
+      case Some(action) if action.isCanceled =>
+        state.version match {
+          case Some(ver) =>
+            ProcessStatus.simpleErrorShouldNotBeDeployed(ver.versionId, ver.user, Option(state))
```
arkadius: can you add here and in other error places some log.warn with message that will help us distinguish those cases which has the same message but in different case handler?

lciolecki: I've removed this `state.version match`. Now we have only ` case Some(action) if action.isCanceled =>` in fact these old cases did same things.

PR #655, opened 2020-02-05
Move validators configuration to backend

PPiedel:
How to add new validator?
### Backend (configuration)
1. Add new implementation of ValidatorEvaluator (based on some annotation probably)
2. Put new validator inside ValidatorsEvaluatorChain

### Frontend
It's a little bit more tricky.
1.  Add new const/function which returns "Validator" type.
2. Put validator name (should match backend name) into enum validatorType
3. Put new validator into validators: Record<validatorType, Function> with name from step 2.
It should works for params, but you should also check if it works for "expression" and fields, becasue right now validators for them are configured on FE (like fieldLabel, fieldName and so on...) and you will have to probably pass new validator manually. 

Be happy of adding new validator (yet on frontened)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -59,3 +60,7 @@ object DualParameterEditor {
     Decoder.decodeString.emapTry(name => Try(DualEditorMode.fromName(name)))
   }
 }
+
+@ConfiguredJsonCodec sealed trait ParameterValidator
+
+case object NotEmptyValidator extends ParameterValidator
```
mproch: let's add comment/javadoc here describing what and where should be added when I want to add one more validator (e.g. in Validators.ts, but anywhere else?)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ProcessDefinitionExtractor.scala
```diff
@@ -81,10 +81,16 @@ object ProcessDefinitionExtractor {
...
 
-    implicit val nodeConfig: ValueReader[ParameterEditor] = ValueReader.relative(config => {
+    implicit val paramEditorReader: ValueReader[ParameterEditor] = ValueReader.relative(config => {
       val json = config.root().render(ConfigRenderOptions.concise().setJson(true))
-      CirceUtil.decodeJsonUnsafe[ParameterEditor](json, "invalid process config")
+      CirceUtil.decodeJsonUnsafe[ParameterEditor](json, "invalid parameter editor config")
     })
+
+    implicit val paramValidatorReader: ValueReader[ParameterValidator] = ValueReader.relative(config => {
+      val json = config.root().render(ConfigRenderOptions.concise().setJson(true))
```
mproch: let's extract it some util? It's a bit hacky IMHO btw ;)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorEvaluator.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.engine.definition.validator
```
mproch: where is it used?

PPiedel: NotEmptyValidatorEvaluator extends this trait. All new validators should evaluate based on parameter passed

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
PPiedel: I've updated PR description with detailed instruction of how to add new validator

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ProcessDefinitionExtractor.scala
PPiedel: Done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
```diff
@@ -30,10 +30,10 @@ object WithCategories {
 }
 
 object ParameterConfig {
-  val empty: ParameterConfig = ParameterConfig(None, None)
+  val empty: ParameterConfig = ParameterConfig(None, None, None)
 }
 
-@JsonCodec case class ParameterConfig(defaultValue: Option[String], editor: Option[ParameterEditor])
+@JsonCodec case class ParameterConfig(defaultValue: Option[String], editor: Option[ParameterEditor], validators: Option[List[ParameterValidator]])
```
arkadius: Why it is `Option[List[]]` instead of just `List[]`?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
arkadius: I would like to call it rather 'MandatoryValueValidator`  instead of `NonEmpty...`. It should be used in situation when something is not optional but mandatory. On the frontend we will interpret it by checking if field is non empty, but we should think about it in domain categories instead of presentation. This validators at the end should be interpreted on both FE and BE side.
Also we can introduce NonEmptyValidator for non empty strings (`''` in spel) or non empty lists (`{}` in spel) in the future.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
PPiedel: It can be undefined in config file. Then it maps to None.

You could also define [] list in config file. Then in maps to empty list

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
PPiedel: Ok,  done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
arkadius: Ok, now I see that it is used in config file.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -25,6 +25,7 @@ case class Parameter(name: String,
                      typ: TypingResult,
                      runtimeClass: Class[_],
                      editor: Option[ParameterEditor] = None,
+                     validators: Option[List[ParameterValidator]] = None,
```
arkadius: Ok, but here we can have only List :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorEvaluator.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import java.lang.reflect.Parameter
+
+import pl.touk.nussknacker.engine.api.definition.ParameterValidator
+
+trait ValidatorEvaluator {
```
arkadius: Can you name it `ValidatorExtractor`? We have a lot of lazy paramerters, interpreters and you word "evaluator" when we need to evaluate value of expression.  Here you extract some information from method definition. It is very similiar to `TypesInformationExtractor`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorsEvaluatorChain.scala
```diff
@@ -0,0 +1,20 @@
+package pl.touk.nussknacker.engine.definition.validator
+
+import java.lang.reflect.Parameter
+
+import pl.touk.nussknacker.engine.api.definition.ParameterValidator
+
+object ValidatorsEvaluatorChain {
```
arkadius: I don't think that we need chain for that. Chain is need when you need to find final value based on some chain of logic which can be broken on some step. Here you have a list of extractors, and want to ask them to find all possible validators and merge all results (using ++)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotEmptyValidatorEvaluator.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import javax.annotation.Nullable
+import pl.touk.nussknacker.engine.api.definition.{MandatoryValueValidator, ParameterValidator}
+
+object NotEmptyValidatorEvaluator extends ValidatorEvaluator {
+
+  override def evaluate(param: Parameter): Option[ParameterValidator] = {
+    param.getAnnotation(classOf[Nullable]) match {
+      case _: Nullable => None
+      case _ => Some(MandatoryValueValidator)
```
arkadius: IMO we should hande here also `Option[]` and `Optional[]` - let's keep all of them in one place.

PPiedel: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorsEvaluatorChain.scala
PPiedel: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorEvaluator.scala
PPiedel: Done

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -438,7 +438,9 @@ object ComplexObject {
 
 case object MultipleParamsService extends Service {
   @MethodToInvoke
-  def invoke(@ParamName("foo") foo: String,
+  def invoke(@ParamName("foo")
```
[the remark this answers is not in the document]
PPiedel: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorsExtractor.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import pl.touk.nussknacker.engine.api.definition.ParameterValidator
+
+object ValidatorsExtractor {
+  def extract(parameter: Parameter): List[ParameterValidator] = {
+    new ValidatorsExtractor(List(MandatoryValueValidatorExtractor)).extract(parameter)
+  }
+}
+
+class ValidatorsExtractor(validatorExtractors: Iterable[ValidatorExtractor]) {
```
mproch: is this class really needed?

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/validator/ParameterValidatorsExtractorChain.scala
```diff
@@ -0,0 +1,25 @@
...
+      new ParameterConfigValidatorsExtractor(parameterConfig),
+      new ParameterValidatorsExtractor
+    )
+    new ParameterValidatorsExtractorChain(strategies)
+  }
+}
+
+class ParameterValidatorsExtractorChain(strategies: Iterable[ParameterValidatorsExtractorStrategy]) {
+
+  def evaluate(param: Parameter): List[ParameterValidator] = {
```
mproch: why we cannot add some validators via config and leave rest from methods?

PPiedel: I we would merge validators from config with those from annotations then you would not be able to define in config that you want f.e. empty list of validators instead of default ones

arkadius: One option that I see is that we can sometimes want to remove some validator. Currently all non optional, not nullable parameters will have mandatory validator. If will be only possibility to add new one, you won't be able to remove this validator without adding `@Nullable` annotation...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorsExtractor.scala
PPiedel: Inlined in `object ValidatorsExtractor`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/MandatoryValueValidatorExtractor.scala
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Optional
+
+import javax.annotation.Nullable
+import pl.touk.nussknacker.engine.api.definition.{MandatoryValueValidator, ParameterValidator}
+
+object MandatoryValueValidatorExtractor extends ValidatorExtractor {
+
+  override def extract(param: Parameter): Option[ParameterValidator] = {
+    param match {
+      case param if param.getType == classOf[Option[Any]] => None
```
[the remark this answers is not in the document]
PPiedel: Thanks, done!

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -193,8 +193,16 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
       "echoEnumService" -> all(EchoEnumService),
       // types
       "simpleTypesService"  -> all(new SimpleTypesService).withNodeConfig(SingleNodeConfig.zero.copy(category = Some("types"))),
-      "optionalTypesService"  -> all(new OptionalTypesService).withNodeConfig(SingleNodeConfig.zero.copy(category = Some("types"))),
-      "collectionTypesService"  -> all(new CollectionTypesService).withNodeConfig(SingleNodeConfig.zero.copy(category = Some("types"))),
+      "optionalTypesService"  -> all(new OptionalTypesService)
+        .withNodeConfig(SingleNodeConfig.zero.copy(
+          category = Some("types"),
+          params = Some(Map(
+            "scalaOptionParam" -> ParameterConfig(None, None, Some(List.empty)),
```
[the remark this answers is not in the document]
PPiedel: I've added new `overridenByDevConfigParam` and `overridenByFileConfigParam` params

PPiedel:
>  looks pretty good, I added minor comments.
> Could you add add comment describing adding validators as javadoc/comment to ParameterValidator?

Done using ScalaDoc

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/validator/ParameterValidatorsExtractorChain.scala
PPiedel: Are we agree with that?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/ValidatorsExtractor.scala
arkadius: I think that @mproch want to remove this class at all and inline in place of usage, but for me can be like now also

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/validator/ParameterValidatorsExtractorChain.scala
mproch: Yes, I see your point. This indeed enables some cases, but OTOH it's slightly non-intuitive... Maybe at least let's add comment somewhere?

PPiedel: Added above apply method

ui/client/components/AddProcessDialog.js
```diff
@@ -119,12 +120,12 @@ const nameAlreadyExists = (clashedNames, name) => {
...
   {
-    isValid: (clashedNames, name) => notEmptyValidator.isValid(name),
-    message: notEmptyValidator.message,
+    isValid: (clashedNames, name) => mandatoryValueValidator.isValid(name),
+    message: mandatoryValueValidator.message,
   },
   {
     isValid: (clashedNames, name) => !nameAlreadyExists(clashedNames, name),
-    message: duplicateValue,
+    message: i18next.t("validation.duplicateValue", "This value is already taken"),
```
JulianWielga: i18next.t Is called once, it'll not react to lang change or lazy loaded dicts.

ui/client/components/graph/node-modal/editors/expression/EditorType.ts
```diff
@@ -27,6 +28,7 @@ type EditorConfig = {
   switchableTo?: (expressionObj: ExpressionObj, param?: ParamType, values?: ValuesType) => boolean,
   values?: (param: ParamType, values: ValuesType) => $TodoType,
   switchable?: Function,
+  validators?: Function,
```
JulianWielga: if you know this function args and return type, provide type.

ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
```diff
@@ -21,14 +20,16 @@ type Props = {
   toggleTestResult: Function,
   renderFieldLabel: Function,
   fieldType: string,
+  errors: Array<$TodoType>,
```
JulianWielga: do you know error type?

ui/client/components/graph/node-modal/editors/Validators.ts
```diff
@@ -0,0 +1,35 @@
+import _ from "lodash"
+
+export enum validatorType {
+  MANDATORY_VALUE_VALIDATOR = "MandatoryValueValidator",
+  ERROR_VALIDATOR = "ErrorValidator",
+}
+
+export type Validator = {
+  isValid: (...args: any[]) => boolean
```
JulianWielga: are possible args really unknown?

ui/client/components/graph/node-modal/editors/Validators.ts
```diff
@@ -0,0 +1,35 @@
...
+  isValid: value => !_.isEmpty(value),
+  message: "This field cannot be empty",
+  description: "This field cannot be empty",
+}
+
+export function allValid(validators: Array<Validator>, values: Array<any>): boolean {
+  return validators.every(validator => validator.isValid(...values))
+}
+
+export const validators: Record<validatorType, Function> = {
```
JulianWielga: Function - is more specific type known?

ui/client/components/graph/node-modal/editors/Validators.ts
PPiedel: Values which can be passed depends on specific valdiator. I would leave like that right now.

ui/client/components/graph/node-modal/editors/Validators.ts
PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
PPiedel: Done, added Error type

ui/client/components/graph/node-modal/editors/expression/EditorType.ts
PPiedel: Done

ui/client/components/AddProcessDialog.js
PPiedel: I've moved it to DialogMessages for now

PR #658, opened 2020-02-06
PropertyFromGetterExtractionStrategy: getFoo | foo | (getFoo & foo)

arkadius:


engine/api/src/main/java/pl/touk/nussknacker/engine/javaapi/process/ClassExtractionSettings.java
```diff
@@ -1,8 +1,6 @@
 package pl.touk.nussknacker.engine.javaapi.process;
 
-import pl.touk.nussknacker.engine.api.process.ClassExtractionSettings$;
-import pl.touk.nussknacker.engine.api.process.ClassMemberPredicate;
-import pl.touk.nussknacker.engine.api.process.ClassPredicate;
+import pl.touk.nussknacker.engine.api.process.*;
```
alien11689: Imports without wildcards were better.

PR #660, opened 2020-02-06
Remove ClazzRef - use TypingResult/TypedClass instead

mproch:
1. Let's remove ClazzRef class. It's role can be served by TypingResult, specializing to TypedClass in some places
2. We have to be careful with creating TypeClass object - as TypedClass[Any] should be Unknown. Let's not invoke TypedClass constructor directly, but with factory methods which in most cases return safe TypingResult. When we **really** need TypedClass we have to make sure it's not Unknown
3. Let's leave REST API backward compatible ATM. I've intentionally serialize Unknown in TypeEncoders just like before. We may change that in the future, but let's see how this change works out before
4. realReturnType is simplified to Class[_]


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/WithExplicitMethodToInvoke.scala
```diff
@@ -21,7 +21,7 @@ trait WithExplicitMethodToInvoke {
 
   def returnType: TypingResult
 
-  def realReturnType: TypingResult
+  def realReturnType: Class[_]
```
arkadius: Please also rename all `realReturnType` fields to `runtimeClass`. Just like in `Parameter` class.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ServiceInvoker.scala
```diff
@@ -65,11 +65,7 @@ object ServiceInvoker {
   )
 
   def apply(objectWithMethodDef: ObjectWithMethodDef, collector: Option[ServiceInvocationCollector] = None): ServiceInvoker = {
-    val detectedReturnType = (objectWithMethodDef.methodDef.realReturnType match {
-      case tc: SingleTypingResult =>
-        Some(tc.objType.klass)
-      case _ => None
-    }).getOrElse(classOf[Any])
+    val detectedReturnType = objectWithMethodDef.methodDef.realReturnType
```
arkadius: `detectedRuntimeClass`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
```diff
@@ -185,7 +184,7 @@ object node {
 
     object SubprocessClazzRef {
 
-      def apply[T: TypeTag]: SubprocessClazzRef = SubprocessClazzRef(ClazzRef.fromDetailedType[T].refClazzName)
+      def apply[T: ClassTag]: SubprocessClazzRef = SubprocessClazzRef(implicitly[ClassTag[T]].runtimeClass.getName)
```
arkadius: WDYT about leaving comment on `SubprocessClazzRef` level that we should also replace this class with `TypingResult`. I know that for now we don't have options on UI to declare that input will be `TypedObjectTypingResult` or `TypedUnion` but can't see any blocker to have such option on BE side. Thanks to that we will have one class less and will be cleaner (we won't mix up compilation (TypingResult) with runtime (classes)).

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/uiconfig/defaults/ParameterEvaluatorSpec.scala
```diff
@@ -3,34 +3,32 @@ package pl.touk.nussknacker.ui.process.uiconfig.defaults
...
   private val DEFAULT_NODE_NAME = "undefined"
   private val DEFAULT_PARAMETER_VALUE = "defVal"
   private val DEFINED_NODE_NAME = "defined"
 
-  private def dummyParam(nodeName: String,
-                         paramName: String,
-                         classRef: ClazzRef) =
-    NodeDefinition(nodeName, List(Parameter(paramName, classRef)))
+  private def dummyParam[T:ClassTag](nodeName: String,
+                         paramName: String) =
```
arkadius: can you fix formatting here and below?

mproch: :+1: , done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
mproch: In fact we already have this TODO few lines above :D I added clarification

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ServiceInvoker.scala
mproch: :+1: , done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/WithExplicitMethodToInvoke.scala
mproch: :+1: , done

PR #666, opened 2020-02-10
add more basic methods to collection types definitions

gadomsky:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ClassExtractionSettings.scala
```diff
@@ -165,12 +165,15 @@ object ClassExtractionSettings {
       ClassMemberPatternPredicate(
         SuperClassPatternPredicate(Pattern.compile("java\\.util\\.Collection")),
         Pattern.compile(s"contains|containsAll|get|getOrDefault|indexOf|isEmpty|size")),
+      ClassMemberPatternPredicate(
+        SuperClassPatternPredicate(Pattern.compile("java\\.util\\.Map")),
+        Pattern.compile(s"contains|get|getOrDefault|isEmpty|size|values|keySet")),
```
arkadius: there is no `contains` method, please replace it with `containsKey` and `containsValue`.

gadomsky: right.. 
fixed

PR #667, opened 2020-02-10
Allow empty expression

PPiedel:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -120,6 +122,27 @@ class SpelExpression(parsed: ParsedSpelExpression,
   }
 }
 
+object NullExpression {
```
arkadius: Can u extract it to separate class? IMO it is not specially spel related. We can use it with any other expression language...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -120,6 +122,27 @@ class SpelExpression(parsed: ParsedSpelExpression,
...
+  def apply(original: String, expectedReturnType: TypingResult, flavour: Flavour): NullExpression =
+    new NullExpression(original, returnedValue(expectedReturnType), flavour)
+
+  def returnedValue(expectedReturnType: TypingResult): Any =
+    expectedReturnType.asInstanceOf[TypedClass].klass match {
```
arkadius: We can't use this `asInstanceOf[TypedClass]`. If someone declare parameter as `java.lang.Object` u will have `Unknown` class here. Maybe more safe wil be `expectedReturnType match {`. and for all unknown situation we will return null. WDYT?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -138,10 +161,19 @@ class SpelExpressionParser(parser: org.springframework.expression.spel.standard.
...
   override def parse(original: String, ctx: ValidationContext, expectedType: TypingResult): Validated[NonEmptyList[ExpressionParseError], TypedExpression] = {
-    baseParse(original).andThen { parsed =>
-      validator.validate(parsed, ctx, expectedType).map((_, parsed))
-    }.map { case (combinedResult, parsed) =>
-      TypedExpression(expression(ParsedSpelExpression(original, () => baseParse(original), parsed), expectedType), combinedResult.finalResult, combinedResult.typingInfo)
+    if (StringUtils.isBlank(original)) {
+      val typingResult = CollectedTypingResult.withEmptyIntermediateResults(expectedType)
+      Valid(TypedExpression(
+        NullExpression(original, expectedType, flavour),
+        typingResult.finalResult,
```
arkadius: can u use here directly `expectedType` and it 3rd param just `SpelExpressionTypingInfo(Map.empty)`. Currently it is tricky that you create instance just to extract it one line below :)

PPiedel: Ok, done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
PPiedel: Indeed, it failed. Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
PPiedel: Moved to `expression` package

PR #675, opened 2020-02-11
[FIX] - do not display Finished as Error status

mproch:


ui/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActorSpec.scala
```diff
@@ -182,7 +186,7 @@ class ManagementActorSpec extends FunSuite  with Matchers with PatientScalaFutur
     processDetails.isNotDeployed shouldBe true
   }
 
-  private def isOkForDeployed(state: ProcessStatus): Boolean =
+  private def isFollowingDeploy(state: ProcessStatus): Boolean =
```
lciolecki: Now we can use: `state.status.isFollowingDeployAction`. So we can remove this method.

PR #677, opened 2020-02-11
introduce strictMethodsChecking flag for expression typer

gadomsky:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -337,7 +337,7 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
     Invalid(NonEmptyList.of(ExpressionParseError(message)))
 
   def withDictTyper(dictTyper: SpelDictTyper) =
-    new Typer(classLoader, commonSupertypeFinder, dictTyper)
+    new Typer(classLoader, commonSupertypeFinder, dictTyper, true)
```
arkadius: please rewrite this property here

gadomsky: done :)

PR #681, opened 2020-02-11
Fix: fetching lastAction for process

lciolecki:
- Refactor old deployAction namespace to action
- Most important think is change sortBy place at fetchProcessLatestActionsQuery

Example query before:
```
select * from (select * where "process_id" = 20 order by "performed_at" desc) x2 left outer join "process_comments" x10 on x2.x8 = x10."id"
```

Example query after:
```
select * from (select * where "process_id" = 2) x2 left outer join "process_comments" x10 on x2.x8 = x10."id" order by x2.x6 desc
```

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
@@ -32,15 +32,16 @@ trait ProcessRepository[F[_]] extends Repository[F] with EspTables {
       .sortBy(_.createDate.desc)
 
   protected def fetchLastDeployedActionPerProcessQuery: Query[(api.Rep[Long], (ProcessActionEntityFactory#ProcessActionEntity, api.Rep[Option[CommentEntityFactory#CommentEntity]])), (Long, (ProcessActionEntityData, Option[CommentEntityData])), Seq] =
-    fetchLastActionPerProcessQuery.filter(_._2._1.action === ProcessActionType.Deploy)
+    fetchLastActionPerProcessQuery
+      .filter(_._2._1.action === ProcessActionType.Deploy)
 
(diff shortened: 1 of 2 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -32,26 +32,27 @@ trait ProcessRepository[F[_]] extends Repository[F] with EspTables {
...
 
   protected def fetchLastActionPerProcessQuery: Query[(Rep[Long], (ProcessActionEntityFactory#ProcessActionEntity, Rep[Option[CommentEntityFactory#CommentEntity]])), (Long, (ProcessActionEntityData, Option[CommentEntityData])), Seq] =
     processActionsTable
       .groupBy(_.processId)
       .map { case (processId, group) => (processId, group.map(_.performedAt).max) }
       .join(processActionsTable)
-      .on { case ((processId, latestDeployedAt), deployAction) => deployAction.processId === processId && deployAction.performedAt === latestDeployedAt } //We fetch exactly this one  with max deployment
-      .map { case ((processId, _), deployAction) => processId -> deployAction }
+      .on { case ((processId, maxPerformedAt), action) => action.processId === processId && action.performedAt === maxPerformedAt } //We fetch exactly this one  with max deployment
+      .map { case ((processId, _), action) => processId -> action }
```
arkadius: Why in this query we don't need to sort? Also IMO names of those query methods are confusing. Those queries return sequences, not one action. We should either change those queries to return first element or change name that will be more descriptive that we fetch actions (not last) per process in desc order.

lciolecki: This query is something like big cache for last action for each process (there is filtering by max date - latest). We use it at process list - in that way we create only one query for last action for each process instead query per process.

PR #682, opened 2020-02-11
ParameterDefaultValueExtractorStrategy moved to api

arkadius:
+ naming changes:
  - ParameterValueExtractorStrategy -> ParameterDefaultValueDeterminer (because we don't extract anything)
  - evaluateParameterDefaultValue -> determineParameterDefaultValue (because evaluate is reserved to expressions evaluation etc in our project)
  - ParameterEvaluatorExtractor -> EvaluatedParameterPreparer (because this class doesn't extract paramter's "evaluator")
  - ParameterEditorExtractorStrategy -> ParameterEditorDeterminer
  - ParameterValidatorsExtractorStrategy -> ParameterValidatorsDeterminer
+ TypeRelatedParameterValueDeterminer - also handles boxed ints/floats now


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
```diff
@@ -0,0 +1,9 @@
+package pl.touk.nussknacker.engine.api.defaults
+
+import pl.touk.nussknacker.engine.api.definition.Parameter
+
+trait ParameterDefaultValueDeterminer {
```
PPiedel: Why is only this trait moved to api module and the rest stayed in ui module?

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/defaults/ConfigParameterDefaultValueDeterminer.scala
```diff
@@ -0,0 +1,19 @@
+package pl.touk.nussknacker.ui.definition.defaults
+
+import pl.touk.nussknacker.engine.api.defaults.{NodeDefinition, ParameterDefaultValueDeterminer}
+import pl.touk.nussknacker.engine.api.definition.Parameter
+import pl.touk.nussknacker.engine.api.process.ParameterConfig
+
+class ConfigParameterDefaultValueDeterminer(config: ParamDefaultValueConfig) extends ParameterDefaultValueDeterminer {
```
PPiedel: Now we have Determiner and Extractor for very similar classes.

ParameterConfigValidatorsExtractor / ParameterConfigEditorExtractor vs ConfigParameterDefaultValueDeterminer.

Maybe we should equalize all mentioned names?

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/EvaluatedParameterPreparer.scala
```diff
@@ -1,37 +1,37 @@
...
 import pl.touk.nussknacker.engine.api.definition
-import pl.touk.nussknacker.engine.definition.defaults.{NodeDefinition, ParameterDefaultValueExtractorStrategy}
 import pl.touk.nussknacker.engine.graph.evaluatedparam.Parameter
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-class ParameterEvaluatorExtractor(defaultValueEvaluator: ParameterDefaultValueExtractorStrategy) {
-  def evaluateParameters(nodeDefinition: NodeDefinition): List[Parameter] = {
-    val strategy: definition.Parameter => Option[String] = p => defaultValueEvaluator.evaluateParameterDefaultValue(nodeDefinition, p)
+class EvaluatedParameterPreparer(defaultValueEvaluator: ParameterDefaultValueDeterminer) {
+  def prepareEvaluatedParameter(nodeDefinition: NodeDefinition): List[Parameter] = {
```
PPiedel: BTW prepareEvaluatedParameter and prepareEvaluatedBranchParameter could be merged into one method with some sort of Predicate passed as method param and then renamed to just `prepare`

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/defaults/ConfigParameterDefaultValueDeterminer.scala
arkadius: Indeed WDYT about using Determiner in all cases? Extraction is special case of determining when logic is very straightforward. Should I do it in this change or another one?

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/EvaluatedParameterPreparer.scala
arkadius: I suggest not to change it for now. Those branch parameters are currently not in "final" state. It should be done as a normal parameters of connections at the end. Doing such refactoring can make transition to final state even harder.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
arkadius: I thought about it but then I realize that we don't expose in api ability to totally change chain but only first step - custom default values. And it is generally good approach because why someone would like to skip determining based on configuration from file? By expose in api I mean using `ServiceLoader`

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/defaults/ConfigParameterDefaultValueDeterminer.scala
PPiedel: i would go for "now" option :)

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/EvaluatedParameterPreparer.scala
PPiedel: Ok

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
PPiedel: I didn't notice `ServiceLoader`. Then that's ok, I think

ui/server/src/main/scala/pl/touk/nussknacker/ui/definition/defaults/ConfigParameterDefaultValueDeterminer.scala
arkadius: Ok I'll do it

PR #689, opened 2020-02-12
Validate parameters with configured custom validators on backend

PPiedel:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -73,6 +74,14 @@ object DualParameterEditor {
...
+@ConfiguredJsonCodec sealed trait ParameterValidator {
 
-case object MandatoryValueValidator extends ParameterValidator
+  def isValid(expression: String) : Boolean
+
+}
+
+case object MandatoryValueValidator extends ParameterValidator {
+
+  override def isValid(expression: String): Boolean = StringUtils.isNotEmpty(expression)
```
arkadius: should be `isNotBlank`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
```diff
@@ -1,24 +1,67 @@
...
+  private def validateWithCustomValidators[T >: PartSubGraphCompilationError <: ProcessCompilationError](parameterDefinitions: List[Parameter],
+                                                                                                         parameters: List[evaluatedparam.Parameter])
+                                                                                                        (implicit nodeId: NodeId) = {
+    val validators = parameterDefinitions.map(param => (param.name, param.validators)).toMap
+    val validationErrors = parameters
+      .flatMap(param => validators
+        .getOrElse(param.name, List.empty)
+        .map(validator => (validator, validator.isValid(param.expression.expression), param.name)))
+      .filter(!_._2)
+      .map((t: (ParameterValidator, Boolean, String)) => ParameterValidationError(t._1, t._3))
```
arkadius: please use syntax:
```
.map {
  case (validator, isValid, paramName) => ...
}
```
Names are more descriptive then types.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -73,6 +74,14 @@ object DualParameterEditor {
  * to [[pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor]] which should decide whether new validator
  * should appear in configuration for certain parameter
  */
-@ConfiguredJsonCodec sealed trait ParameterValidator
+@ConfiguredJsonCodec sealed trait ParameterValidator {
 
-case object MandatoryValueValidator extends ParameterValidator
+  def isValid(expression: String) : Boolean
```
arkadius: WDYT about returning something like `Validated[ProcessCompilationError, Unit]`. We should minimize places that should be changed after adding new validator. One place that we will win is object `ParameterValidationError` and pattern matching inside factory method.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
```diff
@@ -1,24 +1,67 @@
...
-                                                                                       usedParamNamesSet: Set[String])
+  def validateParameters[T >: PartSubGraphCompilationError <: ProcessCompilationError](parameterDefinitions: List[Parameter],
+                                                                                       parameters: List[evaluatedparam.Parameter])
                                                                                       (implicit nodeId: NodeId): ValidatedNel[T, Unit] = {
+    val definedParamNamesSet = parameterDefinitions.map(_.name).toSet
+    val usedParamNamesSet = parameters.map(_.name).toSet
+
+    val validatedRedundant = validateRedundancy(definedParamNamesSet, usedParamNamesSet)
+    val validatedMissing = validateMissingness(definedParamNamesSet, usedParamNamesSet)
+    val validatedCustom = validateWithCustomValidators(parameterDefinitions, parameters)
```
arkadius: This is not the best place to invoke this validators. The best option will be somewhere after `ExpressionEvaluator.evaluate` and/or `ExpressionEvaluator.evaluateParameters`. and real check in validator should be `if (evaluatedValue == null) then invalid(...)`. Consider validator `GreaterThan(5)` and expression `2 + 4`. In this place you won't be able to validate it. But ... I don't think that it is a blocker for this PR (we can fix it in some next PR). But please think about it a moment - maybe you will find good place to move this code into.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
PPiedel: Ok, will fix it in next PR, cause it's not obvious where to put these validations

arkadius: Can you leave here a TODO that this validations should be in other place (check evaluated result of expression)?

PPiedel: Done :)

PR #692, opened 2020-02-14
Allow for better checkpoint configuration

alien11689:


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
```diff
@@ -261,7 +270,9 @@ object FlinkStreamingProcessRegistrar {
 
     val enableObjectReuse = config.getOrElse[Boolean]("enableObjectReuse", true)
     val eventTimeMetricDuration = config.getOrElse[FiniteDuration]("eventTimeMetricSlideDuration", 10.seconds)
-    val checkpointInterval = config.as[FiniteDuration]("checkpointInterval")
+    // TODO checkpointInterval is deprecated - remove it in future
+    val checkpointInterval = if (config.hasPath("checkpointInterval")) Some(config.as[FiniteDuration](path = "checkpointInterval")) else None
```
jedrz: Maybe `config.getAs` instead of if?

docs/Configuration.md
```diff
@@ -124,15 +129,20 @@ flinkConfig {
...
 Configuration of processes has few common keys:
-*  timeout (e.g. 10s)- for synchronous services
-*  checkpointInterval - e.g. 10m
+*  `timeout` (e.g. 10s)- for synchronous services
+*  `checkpointInterval` - deprecated (use `checkpointConfig.checkpointInterval` instead), e.g. 10m
+*  `checkpointConfig` (more about checkpoint configuration you can find in [Flink Documentation](https://ci.apache.org/projects/flink/flink-docs-release-1.9/api/java/org/apache/flink/streaming/api/environment/CheckpointConfig.html))
+    * `checkpointInterval` - e.g. 10m
+    * `minPauseBetweenCheckpoints` - optional, default to half of `checkpointInterval`, e.g. 5m
+    * `maxConcurrentCheckpoints` - optional, default to 1, e.g. 4
+    * `tolerableCheckpointFailureNumber` - optional, default not set, e.g. 6
```
mproch: what does 'default not set' mean in this setting? Int.Max?

docs/Configuration.md
```diff
@@ -124,15 +129,20 @@ flinkConfig {
...
+* `classpath` - list of files/URLs with jars with model for processes
 
 ##Process  {#model}
 
 Configuration of processes has few common keys:
-*  timeout (e.g. 10s)- for synchronous services
-*  checkpointInterval - e.g. 10m
+*  `timeout` (e.g. 10s)- for synchronous services
+*  `checkpointInterval` - deprecated (use `checkpointConfig.checkpointInterval` instead), e.g. 10m
+*  `checkpointConfig` (more about checkpoint configuration you can find in [Flink Documentation](https://ci.apache.org/projects/flink/flink-docs-release-1.9/api/java/org/apache/flink/streaming/api/environment/CheckpointConfig.html))
```
mproch: let's use "flinkMajorVersion" variable (in book.json) and update it to 1.9 :)

engine/demo/src/test/resources/application.conf
```diff
@@ -1,4 +1,9 @@
-checkpointInterval: 10s
+checkpointConfig {
+    checkpointInterval: 10s
```
mproch: isn't checkpoint in all these properties necessary? they're in checkpointConfig anywa? maybe interval, maxConcurrent etc. ?

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
```diff
@@ -55,8 +55,10 @@ import scala.util.{Failure, Success}
 
 class FlinkStreamingProcessRegistrar(compileProcess: (EspProcess, ProcessVersion) => (ClassLoader) => CompiledProcessWithDeps,
                                      eventTimeMetricDuration: FiniteDuration,
-                                     checkpointInterval: FiniteDuration,
-                                     enableObjectReuse: Boolean, diskStateBackend: Option[StateBackend])
+                                     @Deprecated checkpointInterval: Option[FiniteDuration],
+                                     enableObjectReuse: Boolean,
+                                     diskStateBackend: Option[StateBackend],
+                                     checkpointConfig: Option[CheckpointConfig] = None)
```
mproch: let's use only CheckpointConfig here, and handle legacy method only around line 174 (parsing config)

docs/Configuration.md
alien11689: I set the default to 0 as it is in flink docs

engine/demo/src/test/resources/application.conf
alien11689: I'd prefer to keep long names, because flink has them in such format

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
```diff
@@ -261,18 +267,27 @@ object FlinkStreamingProcessRegistrar {
 
     val enableObjectReuse = config.getOrElse[Boolean]("enableObjectReuse", true)
     val eventTimeMetricDuration = config.getOrElse[FiniteDuration]("eventTimeMetricSlideDuration", 10.seconds)
-    val checkpointInterval = config.as[FiniteDuration]("checkpointInterval")
+
+    // TODO checkpointInterval is deprecated - remove it in future
+    val checkpointInterval = config.getAs[FiniteDuration](path = "checkpointInterval")
+    if(checkpointInterval.isDefined) {
```
arkadius: you missed space after if :P

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
```diff
@@ -240,18 +242,22 @@ class FlinkStreamingProcessRegistrar(compileProcess: (EspProcess, ProcessVersion
...
-    val checkpointIntervalToSetInMillis = streamMetaData.checkpointIntervalDuration.getOrElse(checkpointInterval).toMillis
-    env.enableCheckpointing(checkpointIntervalToSetInMillis)
+    val processSpecificCheckpointIntervalDuration = streamMetaData.checkpointIntervalDuration
+    val checkpointIntervalToSetInMillis = processSpecificCheckpointIntervalDuration.getOrElse(checkpointConfig.checkpointInterval).toMillis
 
-    //TODO: should this be configurable?
-    env.getCheckpointConfig.setMinPauseBetweenCheckpoints(checkpointIntervalToSetInMillis / 2)
-    env.getCheckpointConfig.setMaxConcurrentCheckpoints(1)
+    env.enableCheckpointing(checkpointIntervalToSetInMillis)
+    env.getCheckpointConfig.setMinPauseBetweenCheckpoints(checkpointConfig.minPauseBetweenCheckpoints.map(_.toMillis).getOrElse(checkpointIntervalToSetInMillis / 2))
```
arkadius: Can you replace this logic with:
```
checkpointConfig.minPauseBetweenCheckpoints.map(_.toMillis).foreach(env.getCheckpointConfig.setMinPauseBetweenCheckpoints)
```
?
Thanks to that won't be need this `getOrElse(checkpointIntervalToSetInMillis / 2)` with magic numbers - let's use flink's defaults

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
```diff
@@ -261,18 +267,27 @@ object FlinkStreamingProcessRegistrar {
...
+
+    // TODO checkpointInterval is deprecated - remove it in future
+    val checkpointInterval = config.getAs[FiniteDuration](path = "checkpointInterval")
+    if(checkpointInterval.isDefined) {
+      logger.warn("checkpointInterval config property is deprecated, use checkpointConfig.checkpointInterval instead")
+    }
+
+    val checkpointConfig = config.getAs[CheckpointConfig](path = "checkpointConfig")
+      .orElse(checkpointInterval.map(CheckpointConfig(_)))
+      .getOrElse(throw new IllegalArgumentException(s"At least checkpointInterval should be defined in process configuration"))
```
arkadius: WDYT about make this also optional? Doesn't Flink has some defaults for that? If it doesn't, we can introduce our default in code (10 seconds will be suitable) and not enforce our users to add this in their configs.

docs/Configuration.md
arkadius: Let's write that in case of missing, will be used Flink's default (0)

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
alien11689: I made it optional - without value checkpoint are not enabled for process (unless checkpoint interval is set in process properties)

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrar.scala
alien11689: I will keep `checkpointIntervalToSetInMillis / 2` to make NK compatible with previous versions

PR #696, opened 2020-02-17
Improvements: FE boosts up loading process state

lciolecki:


ui/client/containers/admin/CustomProcesses.js
```diff
@@ -40,13 +40,13 @@ class CustomProcesses extends BaseProcesses {
 
   deploy = (process) => () => {
     this.props.actions.toggleConfirmDialog(true, DialogMessages.deploy(process.name), () => {
-      return HttpService.deploy(process.name).then(() => this.reload())
+      return HttpService.deploy(process.name).finally(() => this.reload(true))
```
arkadius: What is this argument of reload? I see in code only non-argument version of this method. Why you don't do something like `this.props.actions.loadProcessState`?

ui/client/reducers/graph.js
```diff
@@ -16,6 +16,8 @@ const emptyGraphState = {
   processCounts: {},
   testResults: null,
   businessView: false,
+  processState: null,
+  stateLoaded: false,
```
arkadius: please rename to processStateLoaded

lciolecki: Changed

ui/client/containers/admin/CustomProcesses.js
lciolecki: Changed

PR #705, opened 2020-02-18
Test capabilities resource available for users without deploy role

arkadius:


arkadius:
Deployed on demo, @fijolekProjects please check if error is not displayed anymore. I can't check it on my own, because I have other roles and problem didn't appear for me...

fijolekProjects:
Now it works!

PR #708, opened 2020-02-19
Fix default value based on editor default values determining

### LINKED ISSUE #703: Default parameter value for parameter with fixed values editor is badly determined
Take a look at 'enricher' service, 'tariffType` param, available on staging:

Screenshot from 2020-02-18 18-35-41

Should be NORMAL tariff but is empty expression. Looks like somewhere in code validators are not taken into account and:
```
object OptionalityBasedDefaultValueDeterminer extends ParameterDefaultValueDeterminer {

  override def determineParameterDefaultValue(nodeDefinition: NodeDefinition, parameter: Parameter): Option[String] =
    if (parameter.isOptional) Some("") else None

}
```
returns `Some("")` instead of `None`.

Validators on api are returned correctly.

PPiedel:
This fixes #703

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
@@ -1,9 +1,0 @@
-package pl.touk.nussknacker.engine.api.defaults
-
-import pl.touk.nussknacker.engine.api.definition.Parameter
-
-trait ParameterDefaultValueDeterminer {
-  def determineParameterDefaultValue(nodeDefinition: NodeDefinition, parameter: Parameter): Option[String]
-}
-
-final case class NodeDefinition(id: String, parameters: List[Parameter])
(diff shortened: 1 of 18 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/defaults/ParameterDefaultValueDeterminer.scala
```diff
@@ -1,9 +0,0 @@
-package pl.touk.nussknacker.engine.api.defaults
```
arkadius: Moving this class to ui model in general is ok, but in this change we should also remove `ServiceLoader` usage in chain of determiners. Currently it is not usable because if someone want to load own determiner he/she need to add dependency to nk-ui module. It is not a good idea because it is huge module. AFAIK it wasn't used anywhere. Also programmer still has option to override this defaults via `NodeConfig`

PPiedel: I've removed user strategies, done

PR #715, opened 2020-02-21
Allow simple duration editor with empty expression

### LINKED ISSUE #711: Duration editor enhancements
- [ ]  dual editor with default simple for duration and period
- [ ]  better handling of empty string - showing editor but with empty nested fields

PPiedel:
This resolves #711

### DIFF
--- a/ui/client/components/graph/node-modal/editors/expression/duration/PeriodEditor.tsx
+++ b/ui/client/components/graph/node-modal/editors/expression/duration/PeriodEditor.tsx
@@ -4,6 +4,7 @@ import {Validator} from "../../Validators"
 import React from "react"
 import moment from "moment"
 import TimeRangeEditor from "./TimeRangeEditor"
+import _ from "lodash";
 
 export type Period = {
   years: number,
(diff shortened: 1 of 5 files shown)


ui/client/components/graph/node-modal/editors/expression/duration/PeriodEditor.tsx
```diff
@@ -23,23 +24,31 @@ type Props = {
...
+  months: () => undefined,
+  days: () => undefined
+}
 
 export default function PeriodEditor(props: Props) {
 
   const {expressionObj, onValueChange, validators, showValidation, readOnly, isMarked} = props
 
+  function isPeriodUndefined(period: Period): boolean {
+    return period.years === undefined && period.months === undefined && period.days === undefined
```
arkadius: I think that we should assume that if some part of period is defined then other values are implicitly zero and whole period is defined. Also wdyt about changing logic from negative to positive (isPeriodDefined)? The same in duration

ui/client/components/graph/node-modal/editors/expression/duration/PeriodEditor.tsx
```diff
@@ -23,23 +24,31 @@ type Props = {
 const SPEL_PERIOD_DECODE_REGEX = /^T\(java\.time\.Period\)\.parse\('(.*?)'\)$/
 const SPEL_PERIOD_SWITCHABLE_TO_REGEX = /^T\(java\.time\.Period\)\.parse\('P([0-9]{1,}Y)?([0-9]{1,}M)?([0-9]{1,}W)?([0-9]{1,}D)?'\)$/
 const SPEL_FORMATTED_PERIOD = (isoPeriod) => `T(java.time.Period).parse('${isoPeriod}')`
+const UNDEFINED_PERIOD = {
```
lciolecki: there should be null and please use == / !=.

PR #721, opened 2020-02-24
Added ability to create own CustomStateStatus

arkadius:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -44,7 +44,6 @@ object ProcessState {
       errors
     )
 
-  @JsonCodec case class StateStatusCodec(clazz: String, value: String)
```
alien11689: Is this class a part of public API? Should we change it?

arkadius: Nope, it was left by mistake. I know this from "source of primary change" :)

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/StateStatusCodingSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    val decodedStatus = Json.obj(
+      "type" -> Json.fromString("CustomStateStatus"),
+      "name" -> Json.fromString("CUSTOM")
+    ).as[StateStatus].right.value
+    // we don't decode correctly custom statuses be design - their role is to encapsulate business status of process which will be
+    // then presented by ProcessStateDefinitionManager
+    decodedStatus should not equal givenStatus
+  }
+
+  case class MyCustomStateStatus(someField: String) extends CustomStateStatus("CUSTOM") {
```
lciolecki: Why not: `case class MyCustomStateStatus(name: String, someField: String) extends CustomStateStatus(name)`? After that you should test veryfication name  by givenStatus.name and type by string "CustomStateStatus"

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/StateStatusCodingSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+import io.circe.syntax._
+import org.scalatest.{EitherValues, FunSuite, Matchers, OptionValues}
+import pl.touk.nussknacker.engine.api.deployment.simple.SimpleStateStatus
+
+class StateStatusCodingSpec extends FunSuite with Matchers with EitherValues with OptionValues {
+
+  test("simple status coding") {
+    val givenStatus: StateStatus = SimpleStateStatus.Running
+    val statusJson = givenStatus.asJson
+    statusJson.hcursor.get[String]("type").right.value shouldEqual "RunningStateStatus"
```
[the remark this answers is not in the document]
arkadius: I don't want to have any logic in verification. It was design choice. "RunningStateStatus" is part of API. If you change this name, you should know that someone can rely on this string.

engine/api/src/test/scala/pl/touk/nussknacker/engine/api/deployment/StateStatusCodingSpec.scala
arkadius: Because I don't need passing this `name`. `MyCustomStateStatus` has unchangable name.

PR #730, opened 2020-03-02
Fix: Bug with cancel running process with error

### LINKED ISSUE #729: Cancel button is disabled when process has "Process currently is not running" state
It causes that user can't clean this error after manual cancellation of process in Flink.

lciolecki:
This close #729

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
@@ -23,7 +23,8 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
     SimpleStateStatus.Failed -> List(ProcessActionType.Deploy),
     SimpleStateStatus.Finished -> List(ProcessActionType.Deploy),
     SimpleStateStatus.Error -> List(ProcessActionType.Deploy),
-    SimpleStateStatus.RunningError -> List(ProcessActionType.Deploy, ProcessActionType.Cancel),
+    SimpleStateStatus.RunningWithError -> List(ProcessActionType.Deploy, ProcessActionType.Cancel),
+    SimpleStateStatus.DeployedWithError -> List(ProcessActionType.Deploy, ProcessActionType.Cancel),
     SimpleStateStatus.FailedToGet -> List(ProcessActionType.Deploy)
   )
 
@@ -39,7 +40,8 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
     SimpleStateStatus.Failed -> "/assets/states/failed.svg",
     SimpleStateStatus.Finished -> "/assets/states/success.svg",
     SimpleStateStatus.Error -> "/assets/states/error.svg",
-    SimpleStateStatus.RunningError -> "/assets/states/error.svg"
+    SimpleStateStatus.RunningWithError -> "/assets/states/deploy-failed.svg",
+    SimpleStateStatus.DeployedWithError -> "/assets/states/deploy-failed.svg"
   )
 
   val statusTooltipsMap: Map[StateStatus, String] = Map(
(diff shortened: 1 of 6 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
```diff
@@ -14,4 +14,5 @@ object SimpleStateStatus {
   val Canceled: StoppedStateStatus = StoppedStateStatus("CANCELED")
   val Failed: ErrorStateStatus = ErrorStateStatus("FAILED")
   val Error: ErrorStateStatus = ErrorStateStatus("ERROR")
+  val RunningError: ErrorStateStatus = ErrorStateStatus("RUNNING_ERROR")
```
arkadius: Maybe better name will be `RuninngWithError` or `RuninngWithWarning`? Because now it looks like it is problem during run like process was deployed but can't run.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
```diff
@@ -37,7 +38,8 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
     SimpleStateStatus.DuringCancel -> "/assets/states/stopping-running-animated.svg",
     SimpleStateStatus.Failed -> "/assets/states/failed.svg",
     SimpleStateStatus.Finished -> "/assets/states/success.svg",
-    SimpleStateStatus.Error -> "/assets/states/error.svg"
+    SimpleStateStatus.Error -> "/assets/states/error.svg",
+    SimpleStateStatus.RunningError -> "/assets/states/error.svg"
```
arkadius: Maybe we should use other icon to show that at the end this process is running like `deploy-failed.svg` for now and we should ask our ui-designers to add some `deploy-warning.svg`?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
```diff
@@ -51,7 +53,7 @@ object SimpleProcessStateDefinitionManager extends ProcessStateDefinitionManager
     SimpleStateStatus.DuringCancel -> "The process currently is being canceled.",
     SimpleStateStatus.Failed -> "There are some problems with checking state of process..",
     SimpleStateStatus.Finished -> "The process completed successfully.",
-    SimpleStateStatus.Error -> "There are some errors with process state. Please check if everything is okay with process!"
+    SimpleStateStatus.Error -> "There are some errors with process state. Please check if everything is okay with process!",
```
arkadius: You don't missed something here? Looks like you didn't handle `RunningError`

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
```diff
@@ -78,36 +78,42 @@ object ProcessStatus {
   }
 
   def simpleErrorShouldRunning(deployedVersionId: Long, user: String, previousState: Option[ProcessState]): ProcessStatus = simpleError(
+    errorStatus = SimpleStateStatus.RunningError,
```
arkadius: Here process is not running

lciolecki: ProcessState is not running but ProcessAction is deployed, so.. We should treat as running state.

 I've tested it by:
- deploy process 
- cancel process in flink

Then message should be 'Process should be running' and we should can do Cancle and Deploy action.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
lciolecki: I've added a missing tooltip and description.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleProcessStateDefinitionManager.scala
lciolecki: You are right - I will wait for incons.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
lciolecki: I've changed it to RunningWithError.

ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/ProcessStatus.scala
lciolecki: After consider I've added DeployedWithError status.

PR #732, opened 2020-03-02
Continue when process deploying takes too long

PPiedel:


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -226,6 +227,11 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, mainClassName:
         .body(program)
         .send()
         .flatMap(handleUnitResponse)
+        .recover({
+          //sometimes deploying takes too long, which causes TimeoutException while waiting for deploy response
+          //workaround for now, not the best solution though
+          case _: TimeoutException => Future.unit
```
arkadius: Please add logger.warn ... and maybe extend comment with some TODO of better solution like
`TODO: we should change logic of ManagementActor to always save action deploy (maybe we should rename it to events: deploy_attempt) and introduce new actions (events): deploy_success, deploy_failure(type of failure)`

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingProcessManagerSpec.scala
```diff
@@ -40,6 +40,17 @@ class FlinkStreamingProcessManagerSpec extends FunSuite with Matchers with Strea
     cancel(processId)
   }
 
+  ignore("continue on timeout exception during process deploy") {
```
arkadius: please add comment that it is manual test because is hard to make it automatic

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
PPiedel: Done

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingProcessManagerSpec.scala
PPiedel: Done

PR #741, opened 2020-03-03
[FIX] empty spel template should evaluate to empty string

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -161,6 +161,9 @@ class SpelExpressionParser(parser: org.springframework.expression.spel.standard.
     }
   }
 
+  private def shouldUseNullExpression(original: String): Boolean
+  = flavour != Template && StringUtils.isBlank(original)
```
alien11689: strange formatting ;)

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -510,6 +510,7 @@ class SpelExpressionSpec extends FunSuite with Matchers with EitherValues {
   test("evaluates expression with template context") {
     parseOrFail[String]("alamakota #{444}", ctx, SpelExpressionParser.Template).evaluateSyncToValue[String]() shouldBe "alamakota 444"
     parseOrFail[String]("alamakota #{444 + #obj.value} #{#mapValue.foo}", ctx, SpelExpressionParser.Template).evaluateSyncToValue[String]() shouldBe "alamakota 446 bar"
+    parseOrFail[String]("", ctx, SpelExpressionParser.Template).evaluateSyncToValue[String]() shouldBe ""
```
alien11689: Why don't you create separate test with name for such case?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
mproch: done

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
mproch: done

PR #745, opened 2020-03-04
Improvement: Add property extractor at MetaDataExtractor

lciolecki:


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/util/MetaDataExtractor.scala
```diff
@@ -16,4 +16,15 @@ object MetaDataExtractor {
   def extractTypeSpecificDataOrFail[T <: TypeSpecificData](metaData: MetaData)(implicit classTag: ClassTag[T]): T
     = extractTypeSpecificData(metaData).fold(_ => throw new IllegalArgumentException("Wrong process type"), identity)
 
+  def extractProperty[T](metaData: MetaData, property: String): Option[T] =
+    metaData
+      .additionalFields
+      .flatMap(
+        _.properties
+          .get(property)
+          .map(_.asInstanceOf[T])
```
mproch: Maybe let's add sth type-safer here? Either cast from shapeless or ClassTag + checking isInstanceOf - like in extractTypeSpecificData?

PR #749, opened 2020-03-05
Fix: error information about copy

### LINKED ISSUE #542: copy content of input fields behaviour and notifications
1. when you copy content of a input field in node details modal a notification in shown

image

2. when you do
- select all (^A) in input field
- type new value
- copy value (^A ^C)
 
warning is produced and the value is restet to initial

lciolecki:
This resolves #542

### DIFF
--- a/ui/client/containers/Visualization.js
+++ b/ui/client/containers/Visualization.js
@@ -43,7 +43,7 @@ class Visualization extends React.Component {
   bindCopyShortcut() {
     return (event) => {
       // Skip event triggered by writing selection to the clipboard.
-      if (this.isNotThisCopyEvent(event, copyNodeElementId)) {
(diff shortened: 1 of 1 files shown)


ui/client/containers/Visualization.js
```diff
@@ -52,6 +52,9 @@ class Visualization extends React.Component {
     }
   }
 
+  allowBindCopyShortcut = () =>
+    this.props.allModalsClosed && _.isEmpty(this.props.selectionState) === false
```
jedrz: Wouldn't `!_.isEmpty(this.props.selectionState)` be more readable?

lciolecki: Hmm, in future we want to rewrite this to TS, after that probably we will not use lodash ;)

PR #755, opened 2020-03-06
Avoid large number of threads for high parallelism in async mode

jedrz:
For high parallelism (assume = 10) there was created large amount of threads for a single task slot. That is there were 10 * 4 (default multiplier) = 40 threads and 40 * 10 = 400 for a job. On the other hand, for parallelism = 2 there were created only 2 * 4 = 8 threads for a single task slot and 8 * 2 = 16 for a job. Now we will create always the same number of threads threads for a single task and parallelism * workers threads for a job, that is for parallelism = 10 and workers = 8, there will be created 8 threads for a task and 80 for a job respectively.


jedrz:
No special reason. I decided it could be a sane default.

PR #756, opened 2020-03-06
Use parameter configuration in additional fields

PPiedel:
**Migration guide**
Additional properties configuration is now using power of editors and validators.
`additionalPropertiesConfig` is now ` Map[String, AdditionalPropertyConfig]`.

Sample migration looks like this:
Before:
```
  param1: {
    type: "select"
    default: "true"
    isRequired: true
    values: ["true", "false"]
  }
```
After:
```
param1: {
    label: "Just a sample parameter..."
    defaultValue: "true"
    editor: {
        type: "FixedValuesParameterEditor",
        possibleValues: [
            {"label": "true", "expression": "true"},
            {"label": "true", "expression": "true"}
        ]
    }
}
```
Here: label, default value and editor was configured. FixedValuesValidator will be determined by default based on editor.

It's slighty less concise, but you can use any SimpleParameterEditor you want.
Also validators can be defined, anything which inherits from ParameterValidator can be configured as property validator. Configuration is now way more flexible.


ui/client/components/graph/node-modal/editors/expression/Editor.ts
```diff
@@ -84,7 +85,7 @@ export const editors: Record<EditorType, EditorConfig> = {
   },
   [EditorType.STRING_PARAMETER_EDITOR]: {
     ...defaults,
-    editor: () => StringEditor,
+    editor: (param, displayRawEditor, expressionObj) => expressionObj?.language === "spel" ?  SpelStringEditor : StringEditor,
```
arkadius: Please add more generic support for other editors by passing language

ui/server/src/test/scala/pl/touk/nussknacker/ui/validation/AdditionalPropertiesValidatorTest.scala
```diff
@@ -49,39 +66,101 @@ class AdditionalPropertiesValidatorTest extends FunSuite with Matchers {
...
       ))
     ))
+
     val result = validator.validate(process)
 
-    checkOne(result, Seq("empty", "propReq"))
+    result.errors.processPropertiesErrors shouldBe List(
+      NodeValidationError(
+        "InvalidLiteralIntValue",
+        s"Property $reqFieldName ($label) has value of invalid type",
```
arkadius: Can you validate only type of error (without message). Let's make this tests less "concrete" and take advantage of error types.

lciolecki:
I wonder that FixedValuesParameterEditor can return some validators.. Now we have to duplicate code with posible values.. Maybe Editor should implement validators methods?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -16,12 +19,34 @@ import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{EmptyMand
  */
 @ConfiguredJsonCodec sealed trait ParameterValidator {
 
-  def isValid(paramName: String, expression: String)(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit]
+  def isValid(paramName: String, value: String, label: Option[String] = None)(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit]
```
lciolecki: Default value is ugly.. Better way is create two methods with different signature: one with label second without.

ui/client/components/graph/node-modal/AdditionalProperty.tsx
```diff
@@ -0,0 +1,51 @@
+import {get} from "lodash"
+import EditableEditor from "./editors/EditableEditor"
+import React from "react"
+import {PossibleValue} from "./editors/Validators";
```
lciolecki: Semicolons ;)

ui/client/components/graph/node-modal/editors/EditableEditor.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react"
-import ProcessUtils from "../../../../../common/ProcessUtils"
-import {DualEditorMode, editors, EditorType} from "./Editor"
-import SwitchIcon from "./SwitchIcon"
-import FixedValuesEditor from "./FixedValuesEditor"
+import ProcessUtils from "../../../../common/ProcessUtils"
+import {DualEditorMode, editors, EditorType} from "./expression/Editor"
+import SwitchIcon from "./expression/SwitchIcon"
+import FixedValuesEditor from "./expression/FixedValuesEditor"
 import _ from "lodash"
```
lciolecki: Please import only these methods which are used.

ui/client/components/graph/node-modal/editors/expression/SpelStringEditor.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const parseable = (expressionObj) => {
+  const expression = expressionObj.expression
+  const language = expressionObj.language
+  return stringPattern.test(expression) && language === "spel"
+}
+
+SpelStringEditor.switchableTo = (expressionObj) => parseable(expressionObj)
+SpelStringEditor.switchableToHint = () => "Switch to basic mode"
+SpelStringEditor.notSwitchableToHint = () => "Expression must be a simple string literal i.e. text surrounded by single or double quotation marks to switch to basic mode"
```
lciolecki: Without translation?

lciolecki:
As we have decied.. You have changed API for process properties so.. You have to add some information about it in MigrationGuide.. :)

ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
```diff
@@ -39,7 +39,7 @@ class ExpressionField extends React.Component<Props> {
 
     if (editor.type === EditorType.FIXED_VALUES_PARAMETER_EDITOR)
       return (
-        <EditableExpression
+        <EditableEditor
```
lciolecki: Maybe you can extract creating `EditableEditor` to another method?

PPiedel:
> I wonder that FixedValuesParameterEditor can return some validators.. Now we have to duplicate code with posible values.. Maybe Editor should implement validators methods?

It's intentionally done like that because we do not always have custom editor. Currently we don't have editor for integer, for example, but we want to define validator for int property. Providing validators based on editor is possible, but it brings another level of complexity.

ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
PPiedel: It would be not related to this change - the only thing which changed in this file is because of renaming :)

ui/client/components/graph/node-modal/editors/expression/Editor.ts
PPiedel: Done

ui/server/src/test/scala/pl/touk/nussknacker/ui/validation/AdditionalPropertiesValidatorTest.scala
PPiedel: Done

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
PPiedel: Removed default value

ui/client/components/graph/node-modal/AdditionalProperty.tsx
PPiedel: Done

ui/client/components/graph/node-modal/editors/EditableEditor.tsx
PPiedel: Done

ui/client/components/graph/node-modal/editors/expression/SpelStringEditor.tsx
PPiedel: Done

PPiedel:
> additionalPropertiesConfig

e

> As we have decied.. You have changed API for process properties so.. You have to add some information about it in MigrationGuide.. :)

Done

ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
lciolecki: Oka ;)

docs/MigrationGuide.md
```diff
@@ -11,6 +11,7 @@ To see biggest differences please consult the [changelog](Changelog.md).
  `processTypes`  configuration should be used from now on.
 * [#763](https://github.com/TouK/nussknacker/pull/763) Some API traits (ProcessManager, DictRegistry DictQueryService, CountsReporter) 
     now extend `java.lang.AutoCloseable`.
+* Old way or additional properties configuration should be replaced by the new one, which is now mapped to `Map[String, AdditionalPropertyConfig]`
```
lciolecki: Please move it to new section 

## Preparing version

PR #757, opened 2020-03-06
Add migration guide for changes in async execution config

jedrz:


MigrationGuide.md
```diff
@@ -1,21 +1,30 @@
-# Breaking changes
+# Migration guide
```
alien11689: Where will be the link to Migration guide? Should it be in README.md? Will it be visible in gitbook?

jedrz: You're right. Finally I decided to move migration guide to gitbook.

PR #758, opened 2020-03-06
Dicts fix: returning validation errors when fetching process

arkadius:
... using missing dict key + refactor: utils unification


ui/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
```diff
@@ -158,30 +185,24 @@ class DictsFlowTest extends FunSuite with ScalatestRouteTest with FailFastCirceS
       .as[String].right.value
   }
 
-  private def checkNoInvalidNodesDirect() = {
+  private def extractInvalidNodes: Json = {
```
lciolecki: Mayby you can map json to DTO JSON like `ProcessJson` and use this dto to asserts?

PR #762, opened 2020-03-06
[Refactor] Introduce ProcessingTypeDataProvider

mproch:
As first step towards reloadable models let's replace Map[ProcessingType, _] with dedicated trait so we can track usages more easily


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessingTypeDataProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+import net.ceedubs.ficus.readers.ValueReader
+import pl.touk.nussknacker.engine.ProcessingTypeData.ProcessingType
+import pl.touk.nussknacker.engine.{ProcessManagerProvider, ProcessingTypeConfig, ProcessingTypeData}
+
+/*
+  Currently, the only implementation is map-based, but in the future it will allow to reload ProcessingTypeData related stuff
+  without restarting the app
+  TODO: ensure by types (e.g. appropriate monad?? ;)) that all the methods except mapValues are invoked only in context of single request (or initialization)
+ */
+trait ProcessingTypeDataProvider[+T] {
```
arkadius: I don't know what exactly plan is on the end of refactoring road, but IMO current, `Reader` like interface, on this level will cause that finally we will have a lot switch cases inside real life implementations. We will decide if we should return T lazy or eager based on processing type or even based on level of transformation (added by `mapValues`). It will be much easier if we will have concrete `ProcessingTypeDataProvider` for each `ProcessingType` with concrete methods providing concrete information. Also IMO we should introduce own functors very very careful to not scary people from java ecosystem.

PR #764, opened 2020-03-09
Optional ending custom nodes

raphaelsolarski:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/CustomStreamTransformer.scala
```diff
@@ -26,6 +26,8 @@ abstract class CustomStreamTransformer {
   // TODO: remove after full switch to ContextTransformation API
   def canHaveManyInputs: Boolean = false
 
+  def canBeEnding: Boolean = false
```
mproch: maybe we should have it in ContextTransformation API instead of here?

raphaelsolarski: For me both options are ok. This information can be easily accessible both from `AbstractContextTransformation`/`AbstractContextTransformationDef` and `CustomStreamTransformer`/`CustomTransformerAdditionalData`.

I think that is a matter whether we want to support ending custom nodes only in brand new ContextTransformation API or maybe also in other projects where developers returns their own types to interpret by custom runtimes (without forcing to wrap with ContextTransformation).

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/splittedgraph/part.scala
```diff
@@ -24,6 +24,12 @@ object part {
     override type T = CustomNode
   }
 
+  case class EndingCustomNodePart(node: EndingNode[CustomNode]) extends SubsequentPart {
```
arkadius: Why not to use just `CustomNodePart` with `nextParts = empty` and `ends = empty`?

arkadius:
I don't see changes in `DefinitionPreparer`. We need to have ending custom nodes on the end, just like `services`. See this:
```
    val virtualGroups = List(
      List(inputs),
      List(base),
      List(enrichers, customTransformers) ++ subprocesses,
      List(services, sinks))
```

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -381,19 +401,19 @@ protected trait ProcessCompilerBase {
...
-            case TypedParameter(paramName, TypedExpressionMap(valueByBranch)) =>
-              valueByBranch.map {
-                case (branch, TypedExpression(_, _, typingInfo)) =>
-                  val expressionId = NodeTypingInfo.branchParameterExpressionId(paramName, branch)
-                  expressionId -> typingInfo
-              }
-          }.toMap
-          (typingInfo, obj)
-        }
+      createObject[T](nodeDefinition, outputVariableNameOpt, compiledParameters).map { obj =>
```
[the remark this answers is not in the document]
raphaelsolarski: I reverted few larger format changes.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -143,10 +149,10 @@ protected trait ProcessCompilerBase {
     //we use fold here (and not map/sequence), because we can compile part which starts from Join only when we
     //know compilation results (stored in BranchEndContexts) of all branches that end in this join
     val (result, _) = PartSort.sort(sources.toList).foldLeft(zeroAcc) { case ((resultSoFar, branchContexts), nextSourcePart) =>
-        val compiledPart = compile(nextSourcePart, branchContexts)
```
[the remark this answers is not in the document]
raphaelsolarski: as above

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/splittedgraph/part.scala
raphaelsolarski: I reused `CustomNodePart` at split stage, but  I'll leave `EndingCustomNodePart` at compiled graph structure till we make a try with flink runtime.

raphaelsolarski: After all I also reused CustomNodePart in compile graph.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -283,10 +288,25 @@ protected trait ProcessCompilerBase {
...
+                                   (implicit metaData: MetaData, nodeId: NodeId): CompilationResult[compiledgraph.part.CustomNodePart] = {
+      val (typingInfo, validatedNextCtx, compiledNode) = compileCustomNodeObject(data, Left(ctx), ending = true)
+      val nodeTypingInfo = Map(node.id -> NodeTypingInfo(ctx, typingInfo))
+
+      CompilationResult.map2(
+        CompilationResult(nodeTypingInfo, compiledNode),
+        CompilationResult(validatedNextCtx)
+      ) { (nodeInvoker, nextCtx) =>
+        compiledgraph.part.CustomNodePart(nodeInvoker, node, nextCtx, List.empty, List(NormalEnd(node.id)))
+//        compiledgraph.part.EndingCustomPart(nodeInvoker, node, nextCtx)
```
mproch: why is it commented out?

raphaelsolarski: fixed

PR #768, opened 2020-03-10
Update FE babel libs

lciolecki:
After update nodejs to latest version we have to update also babel libs.


ui/client/package.json
```diff
@@ -81,41 +81,41 @@
...
-    "@babel/plugin-syntax-dynamic-import": "^7.8.0",
-    "@babel/plugin-syntax-import-meta": "^7.8.0",
-    "@babel/preset-env": "^7.8.2",
-    "@babel/preset-react": "^7.8.0",
-    "@babel/preset-typescript": "^7.8.0",
-    "@testing-library/react": "^9.4.1",
-    "@types/classnames": "^2.2.9",
+    "@babel/cli": "^7.8.4",
+    "@babel/core": "^7.8.7",
+    "@babel/plugin-proposal-class-properties": "^7.8.3",
```
PPiedel: Maybe we should extract common versions, like 7.8.3, 7.8.7 etc 7.8.3 to some variable?
Then when someone will update this version all librarier will aoutomatically be updated

PR #772, opened 2020-03-10
Subprocess output returning map variable

nadberezny:
Allow subprocess outputs to return value.
Subprocess can behave as MapVariable and pass fields to process.

### DIFF
--- a/engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrarKafkaSpec.scala
+++ b/engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrarKafkaSpec.scala
@@ -10,7 +10,6 @@ import pl.touk.nussknacker.engine.build.GraphBuilder
 import pl.touk.nussknacker.engine.graph.EspProcess
 import pl.touk.nussknacker.engine.graph.exceptionhandler.ExceptionHandlerRef
 import pl.touk.nussknacker.engine.kafka.{KafkaConfig, KafkaSpec}
-import pl.touk.nussknacker.engine.process.helpers.ProcessTestHelpers
 import pl.touk.nussknacker.engine.process.helpers.SampleNodes.MockService
 import pl.touk.nussknacker.engine.spel
 import pl.touk.nussknacker.test.VeryPatientScalaFutures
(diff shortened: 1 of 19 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/ProcessNodesRewriter.scala
```diff
@@ -144,7 +144,13 @@ trait ExpressionRewriter {
       case n: Source =>
         n.copy(
           ref = n.ref.copy(parameters = rewriteParameters(n.ref.parameters)))
-      case _: BranchEndData | _: Split | _: SubprocessInputDefinition | _: SubprocessOutputDefinition | _: SubprocessOutput => data
+      case n: SubprocessOutputDefinition =>
+        n.copy(
```
lciolecki: Ugly formatting.. Better will be in one line?

engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/FlinkStreamingProcessRegistrarKafkaSpec.scala
```diff
@@ -47,7 +46,7 @@ class FlinkStreamingProcessRegistrarKafkaSpec
         .processorEnd("service", "logService", "all" -> "#input"))
     )
 
-    ProcessTestHelpers.processInvoker.invokeWithKafka(
+    helpers.ProcessTestHelpers.processInvoker.invokeWithKafka(
```
gadomsky: do we need that change actually?

nadberezny: we don't but it removes compiler warning

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/ProcessNodesRewriter.scala
nadberezny: Ugliness is a matter of tase probably :)
but back to this match clasue - I've kept the existing code formatting, for sake of style consistency

PR #779, opened 2020-03-11
Improvements: Editor NotBlank validator

lciolecki:


engine/interpreter/src/main/java/pl/touk/nussknacker/engine/definition/validator/adnotation/NotBlank.java
```diff
@@ -0,0 +1,14 @@
+package pl.touk.nussknacker.engine.definition.validator.adnotation;
```
mproch: Why is it here? And not in API?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -25,3 +24,12 @@ case object MandatoryValueValidator extends ParameterValidator {
   override def isValid(paramName: String, expression: String)(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] =
     if (StringUtils.isNotBlank(expression)) valid(Unit) else invalid(EmptyMandatoryParameter(paramName))
 }
+
+case object NotBlankValueValidator extends ParameterValidator {
```
mproch: Why is it here and not in intepreter?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -16,15 +16,18 @@ object Parameter {
 
   def apply[T: ClassTag](name: String): Parameter = Parameter(name, Typed[T], implicitly[ClassTag[T]].runtimeClass)
 
-  def apply(name: String, typ: TypingResult, runtimeClass: Class[_]): Parameter =
-    Parameter(name, typ, runtimeClass, editor = None, validators = List(MandatoryValueValidator), // we want to have mandatory parameters by default because it can protect us from NPE in some cases
+  def apply(name: String, typ: TypingResult, runtimeClass: Class[_], validators: List[ParameterValidator] = List.empty): Parameter =
+    Parameter(name, typ, runtimeClass, editor = None, validators = List(MandatoryValueValidator) ++ validators, // we want to have mandatory parameters by default because it can protect us from NPE in some cases
```
mproch: Let's add new method for this, without default MandatoryValueValidator, now it can be confusing...

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -141,6 +141,14 @@ object ProcessCompilationError {
       EmptyMandatoryParameter(paramName, nodeId.id)
   }
 
+  case class BlankParameter(paramName: String, nodeId: String)
```
mproch: I think we shouldn't have to change this class for each custom validator...  They should define their errors where they are defined...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotBlankValueValidatorExtractor.scala
```diff
@@ -0,0 +1,16 @@
+package pl.touk.nussknacker.engine.definition.validator
+
+import java.lang.reflect.Parameter
+
+import pl.touk.nussknacker.engine.api.definition.{NotBlankValueValidator, ParameterValidator}
+import pl.touk.nussknacker.engine.definition.validator.adnotation.NotBlank
+
+object NotBlankValueValidatorExtractor extends ValidatorExtractor {
```
mproch: maybe we can have some base class for simple validator extraction? E.g. in validator you would define it's annotation?

engine/interpreter/src/main/java/pl/touk/nussknacker/engine/definition/validator/adnotation/NotBlank.java
arkadius: please use standard annotation: https://docs.jboss.org/hibernate/beanvalidation/spec/2.0/api/javax/validation/constraints/NotBlank.html like we already used for NotNull

lciolecki: You are right..

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: TODO: in future..

engine/interpreter/src/main/java/pl/touk/nussknacker/engine/definition/validator/adnotation/NotBlank.java
lciolecki: This is annotation from hibernate javax validation.. Are you sure that we should use it?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
lciolecki: MandatoryValueValidator should be as default.. NotBlank has sense after field is set..

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotBlankValueValidatorExtractor.scala
lciolecki: We will do it in future..

engine/interpreter/src/main/java/pl/touk/nussknacker/engine/definition/validator/adnotation/NotBlank.java
lciolecki: Changed

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
lciolecki: Changed to `ErrorValidationParameter(validator, paramName, nodeId)`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
arkadius: I think that @mproch rather have an idea to add `apply` method with `validators: List[ParameterValidator]` and make sure that you will pass there correct list of validators. @mproch am I right? I don't think that adding `notBlank` to `Parameter` is a good idea. We will have now additional method for all combinations of validators? Optional/Mandatory are most important cases and it is why we have special methods for them...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotBlankValueValidatorExtractor.scala
arkadius: why? It will be only ~5 LOC. In this change you wrote the same thing again so IMO in this change you should extract duplicated code. The third developer will think "Why should I do that if noone before didn't?"

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -1,27 +1,39 @@
...
+    if (StringUtils.isNotBlank(expression)) valid(Unit) else invalid(ErrorValidationParameter(this, paramName))
+}
+
+case object NotBlankValueValidator extends ParameterValidator {
+
+  override def isValid(paramName: String, expression: String)(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] =
+    if (isExpressionNotBlank(expression)) valid(Unit) else invalid(ErrorValidationParameter(this, paramName))
+
+  private def isExpressionNotBlank(expression: String): Boolean =
+      StringUtils.trim(StringUtils.strip(StringUtils.trim(expression), "'")).length  > 0
```
arkadius: What about `123` case? I think that we should explicitly have such a logic in this validator:
```
if (trimmed expr matches: quote white space characters quote) false
else true // TODO: for now we correctly detect only literal expression with blank string - on this level (not evaluated expression) it is the only thing that we can do
```

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotBlankValueValidatorExtractor.scala
lciolecki: I've added `class AnnotationValidatorExtractor[AnnotationClass](parameterValidator: ParameterValidator) extends ValidatorExtractor `

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: I've changed a litle checking blank stirng.. But still for `"123"` it works.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -1,27 +1,39 @@
...
-import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{EmptyMandatoryParameter, NodeId}
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{ErrorValidationParameter, NodeId}
 
 /**
  * Extend this trait to configure new parameter validator which should be handled on FE.
  * Please remember that you have to also add your own `pl.touk.nussknacker.engine.definition.validator.ValidatorExtractor`
  * to `pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor` which should decide whether new validator
  * should appear in configuration for certain parameter
+  *
+  * TODO: Everyone should have chance create own validator
```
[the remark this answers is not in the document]
lciolecki: Changed.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -1,27 +1,39 @@
...
-    if (StringUtils.isNotBlank(expression)) valid(Unit) else invalid(EmptyMandatoryParameter(paramName))
+    if (StringUtils.isNotBlank(expression)) valid(Unit) else invalid(ErrorValidationParameter(this, paramName))
+}
+
+case object NotBlankValueValidator extends ParameterValidator {
+
+  override def isValid(paramName: String, expression: String)(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] =
+    if (isExpressionNotBlank(expression)) valid(Unit) else invalid(ErrorValidationParameter(this, paramName))
+
+  private def isExpressionNotBlank(expression: String): Boolean =
```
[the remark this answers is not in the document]
lciolecki: Okey, I will change to `NotBlankParameterValidator`. I don't validate value but expression.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
lciolecki: I've changed it.. I've create object for NotBlankParam with method where I put array of validators.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
lciolecki: All errors are defined at: `PrettyValidationErrors`. I've added `case class ErrorValidationParameter(validator: ParameterValidator, paramName: String, nodeId: String)`. No we create message and descripton checking type validator class.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/AnnotationValidatorExtractor.scala
```diff
@@ -0,0 +1,20 @@
...
+class AnnotationValidatorExtractor[AnnotationClass <: Annotation : ClassTag](parameterValidator: ParameterValidator) extends ValidatorExtractor {
+
+  override def extract(param: Parameter): Option[ParameterValidator] = {
+    param match {
+      case param if param.getAnnotation(toClass) != null => Some(parameterValidator)
+      case _ => None
+    }
+  }
+
+  protected def toClass: Class[AnnotationClass] =
```
arkadius: This should be done in factory method (`AnnotationValidatorExtractor.apply`). Class `AnnotationValidatorExtractor` should have explicit parameter `annotationClass: Class[_]`. It will be hard to use this extractor from java now. Also IMO we don't need to have generic `AnnotationClass`. It is overkill that will cause that if you will have in many places such wildcards: `AnnotationValidatorExtractor[_]`.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/NotBlankValueValidatorExtractor.scala
arkadius: Good job, I've left one comment regards to this extractor

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -13,7 +13,7 @@ import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{ErrorVali
  * to `pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor` which should decide whether new validator
  * should appear in configuration for certain parameter
   *
-  * TODO: Everyone should have chance create own validator
+  * TODO: It shouldn't be a sealed trait. We should allow everyone to create own ParameterValidator
```
PPiedel: Why leaving todo instead of removing sealed keyword?

PR #780, opened 2020-03-11
[FIX] join after filter works correctly

mproch:


ui/server/src/test/scala/pl/touk/nussknacker/ui/process/marshall/ProcessConverterSpec.scala
```diff
@@ -34,7 +35,7 @@ class ProcessConverterSpec extends FunSuite with Matchers with TableDrivenProper
 
   private val metaData = StreamMetaData(Some(2), Some(false))
 
-  val validation: ProcessValidation = {
+  lazy val validation: ProcessValidation = {
```
arkadius: why it need to be lazy?

PR #785, opened 2020-03-12
First/simplistic implementation of ModelData/ProcessingTypeData reload

mproch:


ui/server/src/test/scala/pl/touk/nussknacker/ui/integration/BaseFlowTest.scala
```diff
@@ -214,6 +217,56 @@ class BaseFlowTest extends FunSuite with ScalatestRouteTest with FailFastCirceSu
     }
   }
 
+  test("should reload ConfigCreator") {
```
arkadius: what about test for process with reloaded definition? would be nice to show in test that invoking `reload` in api doesn't change definition of proces - you still need to redeploy it...

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -611,3 +615,22 @@ class DatesTypesService extends Service with Serializable {
     ???
   }
 }
+
+//this is to simulate model reloading - we read parameters from file
+class DynamicService extends ServiceWithExplicitMethod {
+
+  private val fileWithDefinition = new File(Properties.tmpDir, "nkDynamicServiceProperties")
```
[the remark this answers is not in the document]
mproch: :+1: done

engine/flink/process/src/test/scala/pl/touk/nussknacker/engine/process/performance/PerformanceConfigCreator.scala
```diff
@@ -0,0 +1,7 @@
+package pl.touk.nussknacker.engine.process.performance
+
+import pl.touk.nussknacker.engine.testing.EmptyProcessConfigCreator
+
+class PerformanceConfigCreator extends EmptyProcessConfigCreator {
```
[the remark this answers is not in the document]
mproch: oh no... again I checked in too much :) removed...

ui/server/src/test/scala/pl/touk/nussknacker/ui/integration/BaseFlowTest.scala
mproch: Do you mean definition of process (json in DB)? Or process running in Flink?

arkadius: Let's say that in this reload change type returned by source - you can use now use brand new field "foo". I would like to see in tests that after reload I can introduce change in process which will use this new field and after redeploy it will handle it. Isn't it a purpose of this change? Definition reload is not the intention of user - he/she want to use this definition. I know that it works almost for sure but maybe we have some caching of part or definition somewhere or some problems with serialization. Without test it is hard to say...

mproch: You're right :+1: I added more elaborate test - is this what you had in mind?

PR #787, opened 2020-03-12
Additional field validators based on editor

PPiedel:
**Migration guide**
Additional properties configuration is now using power of editors and validators.
`additionalPropertiesConfig` is now ` Map[String, AdditionalPropertyConfig]`.

Sample migration looks like this:
Before:
```
  param1: {
    type: "select"
    default: "true"
    isRequired: true
    values: ["true", "false"]
  }
```
After:
```
param1: {
    label: "Just a sample parameter..."
    defaultValue: "true"
    editor: {
        type: "FixedValuesParameterEditor",
        possibleValues: [
            {"label": "true", "expression": "true"},
            {"label": "true", "expression": "true"}
        ]
    }
}
```
Here: label, default value and editor was configured. FixedValuesValidator will be determined by default based on editor.

It's slighty less concise, but you can use any SimpleParameterEditor you want.
Also validators can be defined, anything which inherits from ParameterValidator can be configured as property validator. Configuration is now way more flexible.

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
@@ -1,17 +1,17 @@
-package pl.touk.nussknacker.ui.definition.editor
+package pl.touk.nussknacker.engine.definition
 
 import java.time.temporal.ChronoUnit
 
-import pl.touk.nussknacker.engine.api.definition._
+import pl.touk.nussknacker.engine.api.definition.{DateParameterEditor, DateTimeParameterEditor, DualParameterEditor, DurationParameterEditor, FixedExpressionValue, FixedValuesParameterEditor, ParameterEditor, PeriodParameterEditor, StringParameterEditor, TimeParameterEditor}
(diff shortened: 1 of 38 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
```diff
@@ -0,0 +1,24 @@
+package pl.touk.nussknacker.engine.definition
+
+import java.lang.reflect.Parameter
+
+import pl.touk.nussknacker.engine.api.definition.{FixedExpressionValue, FixedValuesParameterEditor, ParameterEditor}
+
+object ParameterTypeEditorDeterminer {
+  def tryDetermine(p: Parameter): Option[ParameterEditor] = {
```
arkadius: Please add all cases but `com.cronutils.model.Cron` from `pl.touk.nussknacker.ui.definition.editor.ParameterTypeEditorDeterminer` and use this class also on ui side to avoid code split on both sides

PPiedel: Done

PPiedel:
I've merged this PR into #756

PR #791, opened 2020-03-12
Toolbars prepare part4

JulianWielga:


ui/client/components/ToolBox.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const processDefinitionData = useSelector(getProcessDefinitionData)
+  const processCategory = useSelector(getProcessCategory)
+
+  const nodesToAdd = useMemo(
+    () => ProcessDefitionUtils.getNodesToAddInCategory(processDefinitionData, processCategory),
+    [processDefinitionData, processCategory],
+  )
+
+  const nodeGroupIsEmpty = (nodeGroup) => {
+    return nodeGroup.possibleNodes.length == 0
```
PPiedel: could be inlined

ui/client/store/configureStore.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable i18next/no-literal-string */
```
arkadius: In this change you only rewritten this class to TS or it was changed anyway?

ui/client/components/ToolBox.tsx
```diff
@@ -0,0 +1,66 @@
...
+    if (!nodeGroupIsEmpty(nodeGroup)) {
+      dispatch(toggleToolboxGroup(nodeGroup.name))
+    }
+  }
+
+  return (
+    <div id="toolbox">
+      <div>
+        {nodesToAdd.map(nodeGroup => {
+          const label =
```
PPiedel: Maybe it should be separated component, with "label" logic inside it?

ui/client/components/graph/GraphContext.tsx
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable i18next/no-literal-string */
+import React, {createContext, PropsWithChildren, useContext} from "react"
+import Graph from "./Graph"
+
+const GraphContext = createContext<() => Graph>(null)
```
PPiedel: Maybe we could extract () => Graph as separate type?

ui/client/components/graph/GraphContext.tsx
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable i18next/no-literal-string */
```
alien11689: why is the whole file check disabled when it is needed only by one line?

ui/client/store/configureStore.ts
```diff
@@ -0,0 +1,35 @@
...
+  const store = createStore(
+    persistReducer(persistConfig, reducer),
+    composeWithDevTools(
+      applyMiddleware(analyticsMiddleware, thunk),
+    ),
+  )
+  const persistor = persistStore(store)
+
+  if (module.hot) {
+    // eslint-disable-next-line i18next/no-literal-string
```
alien11689: can you fix it by extracting '../reducers' to constant `moduleName`?

ui/client/components/ToolBox.tsx
```diff
@@ -0,0 +1,66 @@
+import React, {useMemo} from "react"
```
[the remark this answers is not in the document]
JulianWielga: no changes except simpler redux connection with hooks

ui/client/store/configureStore.ts
JulianWielga: like in commit message, added state persistency

ui/client/components/ToolBox.tsx
JulianWielga: it's not in scope of my change.

ui/client/components/ToolBox.tsx
JulianWielga: it's not in scope of my change.

ui/client/components/graph/GraphContext.tsx
JulianWielga: what will we get from it?

PR #793, opened 2020-03-12
Improvements: fe not blank validator

lciolecki:
- Fix: when FE validator for BE validator is not founded then use message/description from BE
- Add FE NotBlankValidator
- Add Translations for FE validators
- Update i18next
- Standarization validation message between FE and BE


ui/client/components/graph/node-modal/editors/expression/Editor.ts
```diff
@@ -56,11 +56,25 @@ export enum EditorType {
...
+const getValidators = (param: $TodoType) => {
+  if (param == null) { // By default we return mandatory parameter validator
+    return validators[HandledErrorType.MandatoryParameterValidator]()
+  }
+
+  return param.validators.reduce((results: Array<Error>, validator) => {
+    const frontendValidator = validators[validator.type]
+
+    if (frontendValidator instanceof Function) {
+      results.push(frontendValidator())
```
PPiedel: why we add Validator to Array<Error>?

lciolecki: Upss.. Types in TypeScript don't work very well.. :) I will change it!

ui/client/components/graph/node-modal/editors/Validators.ts
```diff
@@ -59,7 +67,9 @@ export function allValid(validators: Array<Validator>, values: Array<string>): b
...
 }
 
-export const validators: Record<ValidatorName, (errors?: Array<Error>, fieldName?: string) => Validator> = {
-  [ValidatorName.MandatoryParameterValidator]: () => mandatoryValueValidator,
-  [ValidatorName.ErrorValidator]: (errors, fieldName) => errorValidator(errors, fieldName),
+export const validators: Record<HandledErrorType, (errors?: Array<Error>, fieldName?: string) => Validator> = {
+  [HandledErrorType.MandatoryParameterValidator]: () => mandatoryValueValidator,
+  [HandledErrorType.NotBlankParameterValidator]: () => notBlankValueValidator,
+  [HandledErrorType.ErrorValidator]: (errors, fieldName) => errorValidator(errors, fieldName),
+  [HandledErrorType.WrongDateFormat]: (errors, fieldName) => errorValidator(errors, fieldName),
```
PPiedel: why WrongDateFormat is mapped to errorValidator?

lciolecki: removed

engine/api/src/main/java/pl/touk/nussknacker/engine/api/validation/LiteralInt.java
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.engine.api.validation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LiteralInt {
```
arkadius: I don't like idea to introduce annotation for each specific Validator. It will introduce a lot of boilerplate.
I see such options:
1. Introduce only one annotation `@LiteralValue` and in logic determining validators look at this annoatation + type of parameter
2. In logic determining validator to take a look if someone picked simple string editor + type is int and return `LiteralintValidtor` for such case

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -532,12 +533,16 @@ class SimpleTypesService extends Service with Serializable {
              rawParam: String,
 
              @ParamName("intParam")
-             @RawEditor intParam: Int,
+             @RawEditor
```
arkadius: Why you've used `RawEditor` for this? Raw editor is for expressions, IMO better option would be to pick simple string editor...

engine/api/src/main/java/pl/touk/nussknacker/engine/api/validation/LiteralInt.java
lciolecki: I've provided `@Literal` adnotation with type arg.

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
lciolecki: I've used it only for manual FE tests. In fact this param has already existed before my changes..

engine/api/src/main/java/pl/touk/nussknacker/engine/api/validation/Literal.java
```diff
@@ -0,0 +1,12 @@
+package pl.touk.nussknacker.engine.api.validation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Literal {
```
PPiedel: Maybe it would be overkill, but what about https://docs.oracle.com/javaee/7/api/javax/validation/constraints/Pattern.html ? it gives much more flexibility.

ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/PrettyValidationErrors.scala
```diff
@@ -42,13 +37,7 @@ object PrettyValidationErrors {
...
         node(s"Node parameters not filled", s"Please fill missing node parameters: : ${params.mkString(", ")}")
 
-      // Parameter validation error cases
-      case ErrorValidationParameter(validator, paramName, _) => validator match {
-        case MandatoryParameterValidator =>
-          validation(validator, s"No expression found for mandatory parameter", s"Please fill expression for this parameter", fieldName = Some(paramName))
-        case NotBlankParameterValidator =>
-          validation(validator, s"Blank expression for not blank parameter", s"Please fill expression for this parameter", fieldName = Some(paramName))
-      }
+      case pve: ParameterValidationError => handleParameterValidationError(pve)
```
PPiedel: Please do not use 3 letter abbreviations, use some meanigfull name instead :)

lciolecki: Man it's change from old PR (https://github.com/TouK/nussknacker/pull/795/files#diff-ea4056b249d590b498c1de2368661b86R40) which you've already have approved ;)

PPiedel: The comment is still up to date, but of course not crucial

engine/api/src/main/java/pl/touk/nussknacker/engine/api/validation/LiteralInt.java
arkadius: Annotation is ok, but `type` parameter inside it is not. You already have type of parameter. Not need to enforce developer to duplicate this information

engine/api/src/main/java/pl/touk/nussknacker/engine/api/validation/Literal.java
lciolecki: Pattern is good idea, but I think it should be done as separeted Validator. It is easier use: `        @Literal(`type` = LiteralType.Integer)`.

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
arkadius: I know purpose of this class :) It is "added value" for all devs. I just have in mind that it would be nice to change this editor to better present ability of our editors/validators

mproch:
:+1: but let's refactor Literal validators towards using generic regexp

PR #794, opened 2020-03-13
[Sync with master] Subprocess output returning map variable (#772)

nadberezny:


mproch:
we did this in other PR I think...

Issue #813, opened 2020-03-18
BE Literal Validators


lciolecki:
* Create more generic LiteralValidators instead of `LiteralIntValidator`
* Add support more literals: Number (Float, BigDecimal), Char

PR #816, opened 2020-03-19
Improvements: RegExpParameterValidator

lciolecki:
- Add BE and FE RegExpParameterValidator
- Replace LiteralIntValidator by RegExpParameterValidator
- Small refactor

This resolves https://github.com/TouK/nussknacker/issues/813

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
@@ -144,6 +144,8 @@ object ProcessCompilationError {
 
   case class EmptyMandatoryParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
+  case class InvalidIntegerLiteralParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
+
   case class MismatchParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
   case class MissingRequiredProperty(paramName: String, label: Option[String], nodeId: String)
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
@@ -144,6 +144,8 @@ object ProcessCompilationError {
 
   case class EmptyMandatoryParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
+  case class NotMatchParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
+
   case class MissingRequiredProperty(paramName: String, label: Option[String], nodeId: String)
     extends PartSubGraphCompilationError with InASingleNode
 
@@ -166,14 +168,6 @@ object ProcessCompilationError {
   object InvalidPropertyFixedValue {
     def apply(paramName: String, label: Option[String], value: String, values: List[String])(implicit nodeId: NodeId): PartSubGraphCompilationError =
       InvalidPropertyFixedValue(paramName, label, value, values, nodeId.id)
-  }
-
-  case class InvalidLiteralIntValue(paramName: String, label: Option[String], value: String, nodeId: String)
-    extends PartSubGraphCompilationError with InASingleNode
-
-  object InvalidLiteralIntValue {
-    def apply(paramName: String, label: Option[String], value: String)(implicit nodeId: NodeId): PartSubGraphCompilationError =
-      InvalidLiteralIntValue(paramName, label, value, nodeId.id)
   }
 
   case class OverwrittenVariable(variableName: String, nodeId: String)
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
@@ -144,7 +144,7 @@ object ProcessCompilationError {
 
   case class EmptyMandatoryParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
-  case class NotMatchParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
+  case class MismatchParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
   case class MissingRequiredProperty(paramName: String, label: Option[String], nodeId: String)
     extends PartSubGraphCompilationError with InASingleNode
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -108,15 +108,15 @@ case object LiteralParameterValidator {
     "This field value has to be an number",
     "Please fill field by proper number type"
   )
-  
+
+  /**
+    * A now we support only simple literals like: Int / Integer.
+    * Option / Optional / LazyParam are not allowed.
+    */
   def apply(clazz: Class[_]): Option[ParameterValidator] =
     clazz match {
       case clazz if clazz == classOf[Int] => Some(integerValidator)
-      case clazz if clazz == classOf[Option[Int]] => Some(integerValidator)
-      case clazz if clazz == classOf[Optional[Int]] => Some(integerValidator)
       case clazz if clazz == classOf[Integer] => Some(integerValidator)
-      case clazz if clazz == classOf[Option[Integer]] => Some(integerValidator)
-      case clazz if clazz == classOf[Optional[Integer]] => Some(integerValidator)
       case _ => None
     }
 
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -74,9 +74,12 @@ case class FixedValuesValidator(possibleValues: List[FixedExpressionValue]) exte
 }
 
 case class RegExpParameterValidator(pattern: String, message: String, description: String) extends ParameterValidator {
+
+  lazy val regexpPattern: Pattern = Pattern.compile(pattern)
+
   //Blank value should be not validate - we want to chain validators
   override def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
-    if (StringUtils.isBlank(value) || Pattern.compile(pattern).matcher(value).matches())
+    if (StringUtils.isBlank(value) || regexpPattern.matcher(value).matches())
       valid(Unit)
     else
       invalid(MismatchParameter(message, description, paramName, nodeId.id))
@@ -97,6 +100,7 @@ case object LiteralIntegerValidator extends ParameterValidator {
 }
 
 case object LiteralParameterValidator {
+
   lazy val integerValidator: ParameterValidator = LiteralIntegerValidator
 
   lazy val numberValidator: RegExpParameterValidator = RegExpParameterValidator(
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -1,24 +1,27 @@
 package pl.touk.nussknacker.engine.api.definition
+
 import java.util.regex.Pattern
 
-import pl.touk.nussknacker.engine.api.CirceUtil._
 import cats.data.Validated
 import cats.data.Validated.{invalid, valid}
 import io.circe.generic.extras.ConfiguredJsonCodec
 import org.apache.commons.lang3.StringUtils
-import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{BlankParameter, EmptyMandatoryParameter, InvalidPropertyFixedValue, MismatchParameter, NodeId}
 import pl.touk.nussknacker.engine.api.context.PartSubGraphCompilationError
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError._
 
 import scala.reflect.ClassTag
+import scala.util.Try
+
+import pl.touk.nussknacker.engine.api.CirceUtil._
 
 /**
- * Extend this trait to configure new parameter validator which should be handled on FE.
- * Please remember that you have to also add your own `pl.touk.nussknacker.engine.definition.validator.ValidatorExtractor`
- * to `pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor` which should decide whether new validator
- * should appear in configuration for certain parameter
- *
- * TODO: It shouldn't be a sealed trait. We should allow everyone to create own ParameterValidator
- */
+  * Extend this trait to configure new parameter validator which should be handled on FE.
+  * Please remember that you have to also add your own `pl.touk.nussknacker.engine.definition.validator.ValidatorExtractor`
+  * to `pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor` which should decide whether new validator
+  * should appear in configuration for certain parameter
+  *
+  * TODO: It shouldn't be a sealed trait. We should allow everyone to create own ParameterValidator
+  */
 @ConfiguredJsonCodec sealed trait ParameterValidator {
 
   def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit]
@@ -42,7 +45,7 @@ case object MandatoryParameterValidator extends ParameterValidator {
 
 case object NotBlankParameterValidator extends ParameterValidator {
 
-  private final val blankStringLiteralPattern: Pattern = Pattern.compile("'\\s*'")
+  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("'\\s*'")
 
   // TODO: for now we correctly detect only literal expression with blank string - on this level (not evaluated expression) it is the only thing that we can do
   override def isValid(paramName: String, expression: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] =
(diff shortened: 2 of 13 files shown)


ui/client/components/graph/node-modal/editors/Validators.ts
```diff
@@ -80,16 +69,28 @@ export const fixedValueValidator = (possibleValues: Array<PossibleValue>): Valid
...
   message: () => i18next.t("notBlankValueValidator.message", "This field value is required and can not be blank"),
   description: () => i18next.t("validator.description", "Please fill field value for this parameter"),
   handledErrorType: HandledErrorType.BlankParameter,
   validatorType: ValidatorType.Frontend,
 }
 
+export const regExpValueValidator = (pattern: string, message: string, description: string): Validator => ({
+  //Empty value should be not validate - we want to chain validators
+  isValid: value => {
+    console.log("Value", value)
```
mproch: without console.log?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/LiteralValidatorExtractor.scala
```diff
@@ -2,15 +2,14 @@ package pl.touk.nussknacker.engine.definition.validator
...
 import pl.touk.nussknacker.engine.api.validation.Literal
 
 object LiteralValidatorExtractor extends ValidatorExtractor {
   override def extract(param: Parameter): Option[ParameterValidator] =
     param.getAnnotation(classOf[Literal]) match {
-      case _:Literal => param.getType match {
-        case clazz if clazz == classOf[Int] => Some(LiteralIntValidator)
-        case clazz if clazz == classOf[Integer] => Some(LiteralIntValidator)
+      case _: Literal => param.getType match {
+        case clazz => LiteralParameterValidator(clazz)
```
mproch: I think this match is futile? You can just write LiteralParameterValidator(param.getType)?
And what about Option/Optional/LazyParameter? Will it work for them (I know it's kind of stupid to require Literal on LazyParameter or Option, but still - maybe it deserves at least some comment?)

engine/flink/management/sample/src/main/resources/reference.conf
```diff
@@ -14,7 +14,10 @@ additionalPropertiesConfig {
   intOptionalProperty: {
     validators: [
       {
-        type: "LiteralIntValidator"
+        type: "RegExpParameterValidator"
```
mproch: hmmm... this is sth we haven't thought of... :/ It's not so nice that here we have to write patterns in each occurence....

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -70,11 +70,27 @@ case class FixedValuesValidator(possibleValues: List[FixedExpressionValue]) exte
...
                       (implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
-
-    if (Try(value.toInt).isSuccess) valid(Unit) else invalid(ProcessCompilationError.InvalidLiteralIntValue(paramName, label, value))
+    if (Pattern.compile(pattern).matcher(value).matches()) valid(Unit) else invalid(NotMatchParameter(message, description, paramName, nodeId.id))
   }
 }
+
+case object LiteralParameterValidator {
+  val integerValidator: RegExpParameterValidator = RegExpParameterValidator(
+    "^-?[0-9]+$",
```
mproch: If we want to be really correct we would want to validate also length of this literal :P

ui/client/components/graph/node-modal/editors/expression/Editor.ts
```diff
@@ -73,9 +73,9 @@ const configureValidators = (paramConfig: $TodoType): Array<Validator> => {
   }
 
   return (paramConfig.validators || [])
-    .map(v => validators[v.type])
-    .filter(v => v != null)
-    .map(v => v(paramConfig.editor.possibleValues))
+    .map(v => ({fun: validators[v.type], args: omit(v, ["type"])}))
+    .filter(v => v.fun != null)
+    .map(v => v.fun(v.args))
```
mproch: this function doesn't look really readable? At very least we should write comment what is going on here, so that adding other validators would be easier?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -144,6 +144,8 @@ object ProcessCompilationError {
 
   case class EmptyMandatoryParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
 
+  case class NotMatchParameter(message: String, description: String, paramName: String, nodeId: String) extends ParameterValidationError
```
arkadius: Maybe mismatch?

ui/client/components/graph/node-modal/editors/Validators.ts
lciolecki: ups..

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -70,11 +70,27 @@ case class FixedValuesValidator(possibleValues: List[FixedExpressionValue]) exte
...
 }
 
-case object LiteralIntValidator extends ParameterValidator {
-
+case class RegExpParameterValidator(pattern: String, message: String, description: String) extends ParameterValidator {
   override def isValid(paramName: String, value: String, label: Option[String])
                       (implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
-
-    if (Try(value.toInt).isSuccess) valid(Unit) else invalid(ProcessCompilationError.InvalidLiteralIntValue(paramName, label, value))
+    if (Pattern.compile(pattern).matcher(value).matches()) valid(Unit) else invalid(NotMatchParameter(message, description, paramName, nodeId.id))
```
arkadius: Uh oh we have a bug here. I know that it was before this change, but we need to fix it. We should allow user to define `@Nullable param: Integer`. Validator shouldn't handle blank expression on BE and FE side. To make this validator composable we should handle mandatory checking only in `MandatoryValidator`. Can you write a test for that? Also I wonder what will happen in ``@Nullable param: Int` case (primitive int)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: I've added checking isBlank at RegExp ;)

ui/client/components/graph/node-modal/editors/expression/Editor.ts
lciolecki: Comment added

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
lciolecki: Changed

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
mproch: but I'm not talking about too short - I'm talking about too long number and checking for integer overflow :D

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Fix and tests: https://github.com/TouK/nussknacker/pull/816/commits/d30213d2f5cb86aa7922ab9d70beba7934200449

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -42,9 +43,9 @@ case object MandatoryParameterValidator extends ParameterValidator {
 
 case object NotBlankParameterValidator extends ParameterValidator {
 
-  private final val BlankStringLiteralPattern: Pattern = stringLiteralPattern("\\s*")
+  private def stringLiteralPattern(pattern: String): Pattern = Pattern.compile(s"'$pattern'")
```
PPiedel: Do we really need to wrap method Pattern.compile into another method ?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -70,11 +71,34 @@ case class FixedValuesValidator(possibleValues: List[FixedExpressionValue]) exte
...
 }
 
-case object LiteralIntValidator extends ParameterValidator {
-
+case class RegExpParameterValidator(pattern: String, message: String, description: String) extends ParameterValidator {
   override def isValid(paramName: String, value: String, label: Option[String])
                       (implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
-
-    if (Try(value.toInt).isSuccess) valid(Unit) else invalid(ProcessCompilationError.InvalidLiteralIntValue(paramName, label, value))
+    //Blank value should be not validate - we want to chain validators
```
PPiedel: I still do not understand this logic :/

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/LiteralValidatorExtractor.scala
lciolecki: It's not true.. When we have Integer param with RAW editr then we can put there expresion like: `#input.value` and more.. In that case LiteralValidator throws error.. So I think better way is use annotation when we want use Literal.. 

Ok, we don't support Option/Optional - these types are not Literal.. But we support `@Nullable` :)  I left some comment.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -57,7 +58,7 @@ case object NotBlankParameterValidator extends ParameterValidator {
     nodeId
   )
   private def isBlankStringLiteral(expression: String): Boolean =
-    BlankStringLiteralPattern.matcher(expression.trim).matches()
+    blankStringLiteralPattern.matcher(expression.trim).matches()
```
PPiedel: Maybe instead of using Pattern we should use here RegExpParameterValidator? I think we should keep logic with pattern checking in only one class. It would be clearer.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Hmm this should be catched on next level validation - I mean on typing? :p

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: What will happend when we have `@Nullable @Literal` param? :)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: If we use RegExp then we will have another ErrorValidation type.. Then at FE will have problem with matching error: empty, regexp.. For Mandatory and NotBlank validators we should have specials error types.

arkadius: ... also notblank have inverted logic - there is !matches instead of matches and it is surprisingly a big difference.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Hmm, changed.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: @mproch has right that for ints regexp are not the best choice  - let's say someone provide 111111111111111111111111111111111. The correct solution should be checking if `Try(expr.toInt).isSucces`... but regexp can be functional for handling `@Pattern` annotation :)

engine/flink/management/sample/src/main/resources/reference.conf
lciolecki: We can rollback LiteralIntValidator.. ;)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: Good change. One more notice: `Pattern.compile(pattern)` should be done once - factory method + replaced String with Pattern in fields or at least `lazy val`. Compiling is expensive...

engine/flink/management/sample/src/main/resources/reference.conf
arkadius: Indeed. Why it was changed? Now someone need to provide this boilerplate for handling common case (ints)? Configuration like this should be necessary only for custom patterns...

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
```diff
@@ -208,6 +210,54 @@ class ProcessValidatorSpec extends FunSuite with Matchers with Inside {
...
+        .source("id1", "source")
+        .customNode("customNodeId", "event", "withNullableLiteralIntegerParam", "nullableLiteralIntegerParam" -> "12")
+        .emptySink("emptySink", "sink")
+
+    validate(processWithInvalidExpression, baseDefinition).result should matchPattern {
+      case Valid(_) =>
+    }
+  }
+
+  test ("valid for Nullable Literal Integer param") {
```
arkadius: :+1: nice

engine/flink/management/sample/src/main/resources/reference.conf
lciolecki: Okey, I will roolback Integer verification ;)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Changed

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
lciolecki: Thanks!

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: I still see `Pattern.compile` inside `valid` method

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -56,25 +57,61 @@ case object NotBlankParameterValidator extends ParameterValidator {
...
+}
 
-  override def isValid(paramName: String, value: String, label: Option[String])
-                      (implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
+case object LiteralParameterValidator {
+  lazy val integerValidator: ParameterValidator = LiteralIntegerValidator
 
-    if (Try(value.toInt).isSuccess) valid(Unit) else invalid(ProcessCompilationError.InvalidLiteralIntValue(paramName, label, value))
-  }
+  lazy val numberValidator: RegExpParameterValidator = RegExpParameterValidator(
```
arkadius: Why you've left it here? It shouldn't be in public api. If you want to have it in sample, you should move it there. Also IMO it is not-real example now, because we recommend to use `LiteralIntegerValidator` for ints. Maybe e-mail regexp will be better choice in sample?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Ok I've moved it.. But in this case it doesn't matter..

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Number != Integer but Integer == Number.. It's here because we don't want to create many of RegExp validator for number. Creating compiler costs.. So I have create one validator and allow to reuse it. I think it's ok. But we can discuss about it :)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -56,25 +57,65 @@ case object NotBlankParameterValidator extends ParameterValidator {
...
+  override def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
+    if (StringUtils.isBlank(value) || regexpPattern.matcher(value).matches())
+      valid(Unit)
+    else
+      invalid(MismatchParameter(message, description, paramName, nodeId.id))
   }
 }
+
+case object LiteralIntegerValidator extends ParameterValidator {
+  //Blank value should be not validate - we want to chain validators
```
PPiedel: validate**d**?

lciolecki:
I've spoked with @PPiedel and we are agree with that we should have two strategies of validation:
- blank: validators: Mandatory / NotBlank
- not-blank: RegExp, LiteralInteger
RegExp and Literal's should be run only when value is not-blank.. But I think it should be done in another PR. Solution created by me is some fast fix.

PPiedel:
> I've spoked with @PPiedel and we are agree with that we should have two strategies of validation:
> 
> * blank: validators: Mandatory / NotBlank
> * not-blank: RegExp, LiteralInteger
> 
> RegExp and Literal's should be run only when value is not-blank.. But I think it should be done in another PR. Solution created by me is some fast fix.

I would divide it slightly different:
- validation of optional fields
- required fields validation
We came to solution, where different strategies should be applied to this two cases. 
Putting "blank" check in each validator is not a best solution, but works for now (as @lciolecki said, it's fast fix)

PR #821, opened 2020-03-20
[FIX] find context variables in nodes

gadomsky:
Fix for accessing variableTypes regardless of nodeId value. (Now, if nodeId contains e.g characters `[ ]` we get null


ui/client/common/ProcessUtils.js
```diff
@@ -44,7 +44,7 @@ class ProcessUtils {
 
   findAvailableVariables = (nodeId, process, processDefinition, fieldName, processCategory) => {
     const globalVariablesWithoutProcessCategory = this._findGlobalVariablesWithoutProcessCategory(processDefinition.globalVariables, processCategory)
-    const variablesFromValidation = _.get(process.validationResult, `variableTypes.${  nodeId}`)
+    const variablesFromValidation = process.validationResult.variableTypes[nodeId]
```
lciolecki: `_.(process, "validationResult.variableTypes.nodeId")`

gadomsky: done

PR #825, opened 2020-03-23
splitting, ts & cleanup (preparing for graph optimization)

JulianWielga:
I believe that there is no change in logic here.


ui/client/components/graph/NodeUtils.ts
```diff
@@ -1,32 +1,62 @@
...
+type Process = {
+  nodes: Node[],
+  edges: Edge[],
+  properties: $TodoType,
+}
+
 class NodeUtils {
 
-  isNode = (obj) => {
+  isNode = (obj: {}): obj is Node => {
```
lciolecki: I know it's refactor.. But can we do that kind of functions do as arrow function in one line? It looks much better.. :)

ui/client/components/graph/NodeUtils.ts
```diff
@@ -165,21 +201,27 @@ class NodeUtils {
...
       (groups) => (groups || []).map(group => ({
         ...group,
         nodes: nodeOperation(group.nodes),
       })),
-      processToDisplay)
+      processToDisplay,
+    )
   }
 
   _canHaveMoreInputs = (nodeTo, nodeInputs, processDefinitionData) => {
```
lciolecki: No we cane remove that prefix `_` for private / protected method..

ui/client/components/graph/NodeUtils.ts
```diff
@@ -1,32 +1,62 @@
+/* eslint-disable i18next/no-literal-string */
 import _ from "lodash"
```
lciolecki: As you have saide - we should import only used method from lodash ;)

ui/client/components/graph/Graph.js
```diff
@@ -152,39 +167,23 @@ class Graph extends React.Component {
     return this.state.exported
   }
 
-  validateConnection = (cellViewS, magnetS, cellViewT, magnetT) => {
-    const from = cellViewS.model.id
-    const to = cellViewT.model.id
-    return magnetT && NodeUtils.canMakeLink(from, to, this.props.processToDisplay, this.props.processDefinitionData)
+  validateConnection() {
```
mproch: why is it better than previous version? I'm just curious ;)

JulianWielga: I try to separate component props from inner joint/paper args

ui/client/components/graph/NodeUtils.ts
```diff
@@ -194,32 +236,41 @@ class NodeUtils {
...
     return this.edgesFromProcess(process).filter(e => e.from == nodeId)
   }
 
-  edgeId = (edge) => {
+  edgeId = (edge: Edge): string => {
     return `${edge.from}-${edge.to}`
   }
 
-  groupIncludesOneOfNodes = (nodeGroup, nodeIds)  => _.some(nodeGroup.nodes, (node) => _.includes(nodeIds, _.get(node, "id") || node))
+  groupIncludesOneOfNodes = (nodeGroup, nodeIds: NodeId[]) => _.some(
```
JulianWielga: something is wrong here. I can't identify the nodeGroup type.

ui/client/test/ValidationLabels-test.js
```diff
@@ -43,9 +42,7 @@ describe("test validation labels", () => {
 
       //when
       render(
-        <I18nextProvider i18n={i18n}>
-          <ValidationLabels validators={validators} values={[""]}/>
-        </I18nextProvider>
+        <ValidationLabels validators={validators} values={[""]}/>
```
lciolecki: Are you sure, doest it work after remove `I18nextProvider`?

PR #838, opened 2020-03-26
Fix checking status of Flink job

gadomsky:
Fix check status when deployment of new version is finished before cancelling previous job on Flink. (Cancelled version has later
last-modification date then the new running version)

### DIFF
--- a/engine/flink/management/src/test/scala/pl/touk/nussknacker/engine/management/FlinkRestManagerSpec.scala
+++ b/engine/flink/management/src/test/scala/pl/touk/nussknacker/engine/management/FlinkRestManagerSpec.scala
@@ -133,6 +133,15 @@ class FlinkRestManagerSpec extends FunSuite with Matchers with PatientScalaFutur
     ))
   }
 
+  test("return running status if cancelled job has last-modification date later then running job") {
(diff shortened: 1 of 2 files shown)


engine/flink/management/src/test/scala/pl/touk/nussknacker/engine/management/FlinkRestManagerSpec.scala
```diff
@@ -133,6 +133,15 @@ class FlinkRestManagerSpec extends FunSuite with Matchers with PatientScalaFutur
     ))
   }
 
+  test("return running status if cancelled job has last-modification date later then running job") {
+    statuses = List(JobOverview("2343", "p1", 20L, 10L, JobStatus.RUNNING), JobOverview("1111", "p1", 30L, 5L, JobStatus.CANCELED))
```
jedrz: Shouldn't we switch order in this list so that the entry with canceled could be skipped?

jedrz: I didn't notice we sort statuses by last modification earlier.

PR #841, opened 2020-03-30
Add namespaces to kafka sink/source

mpac123:
I added possibility to define a namespace via configuration files / environmental variable which influences topic names of Kafka's sources and sinks. The namespace, when defined, is appended to the topic name; otherwise, the name stays unaltered. 

This change will enable running multiple nussknacker instances on a single Kafka cluster.


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaSinkFactory.scala
```diff
@@ -29,8 +29,9 @@ class KafkaSinkFactory(config: KafkaConfig,
              @NotBlank
              topic: String
             )(metaData: MetaData): Sink = {
-    val serializationSchema = schemaFactory.create(topic, config)
-    new KafkaSink(topic, serializationSchema, s"${metaData.id}-$topic")
+    val _topic = NamespaceUtil.alterTopicNameIfNamespaced(topic, config)
```
lciolecki: Maybe `topicName` instead of `_topic`?

mpac123: Or better `namespacedTopic`? What do you think?

engine/flink/kafka-util/src/test/scala/pl/touk/nussknacker/engine/kafka/NamespaceUtilSpec.scala
```diff
@@ -0,0 +1,33 @@
+package pl.touk.nussknacker.engine.kafka
```
lciolecki: I think, we don't need this UnitTest. Everything is already tested in Integration Tests.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaSinkFactory.scala
lciolecki: But when topic is wihout namespace then it's not namespaced ;)

mpac123: yes, but `topicName` suggests that in this line we get the property `name` of the object `topic`, while both of them are simply strings - if we want to be so meticulous with naming

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/ObjectNaming.scala
```diff
@@ -0,0 +1,10 @@
+package pl.touk.nussknacker.engine.util.namespaces
```
mproch, line 1: maybe we should have it in api?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/NamingContext.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.engine.util.namespaces
+
+// This class can be extended in the future to provide more
+// information about context (e.g. process name, category)
+class NamingContext(usageKey: ObjectNamingUsageKey.UsageKey)
```
mproch, line 5: will it be possible to add context without changing this file? Probably not? Can we make it possible?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/ObjectNamingProvider.scala
```diff
@@ -0,0 +1,16 @@
+package pl.touk.nussknacker.engine.util.namespaces
+
+import com.typesafe.scalalogging.LazyLogging
+import pl.touk.nussknacker.engine.util.loader.ScalaServiceLoader
+
+trait ObjectNamingProvider extends LazyLogging with Serializable {
```
mproch, line 6: maybe some javadoc would be helpful?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/KafkaAvroSourceFactory.scala
```diff
@@ -14,20 +14,22 @@ import pl.touk.nussknacker.engine.avro.typed.AvroSchemaTypeDefinitionExtractor
...
+  extends KafkaSourceFactory[T](config, schemaFactory, timestampAssigner,
+    TestParsingUtils.newLineSplit, ObjectNamingProvider) {
 
   override protected def createSource(processMetaData: MetaData,
                                       topics: List[String],
                                       schema: KafkaDeserializationSchema[T]): KafkaSource = {
     val schemaRegistryClient = schemaRegistryClientFactory.createSchemaRegistryClient(config)
     new KafkaSource(consumerGroupId = processMetaData.id, topics, schema,
-      Some(AvroToJsonFormatter(schemaRegistryClient, topics.head, formatKey)))
+      Some(AvroToJsonFormatter(schemaRegistryClient, topics.head, formatKey)), ObjectNamingProvider)
```
mproch, line 32: why we use default here? And in all places other than ProcessConfigCreator?

mpac123: It's not necessarily default - `ObjectNamingProvider` uses `ScalaServiceLoader`, which will load the default `ObjectNaming` implementation `DefaultObjectNaming` only if there is no other implementation of `ObjectNaming` provided. I wanted to pass the provider to `KafkaSource` and `KafkaSink` only for testing purposes. I can move it higher up the tree than to `KafkaSourceFactory[T]` and `KafkaSinkFactory` as it is currently, but at some point `ObjectNamingProvider` will have to be passed anyway - so I suppose the question here is where is it reasonable to stop.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaSinkFactory.scala
```diff
@@ -29,8 +32,10 @@ class KafkaSinkFactory(config: KafkaConfig,
              @NotBlank
              topic: String
             )(metaData: MetaData): Sink = {
-    val serializationSchema = schemaFactory.create(topic, config)
-    new KafkaSink(topic, serializationSchema, s"${metaData.id}-$topic")
+    val objectNaming = objectNamingProvider.create(getClass.getClassLoader)
```
arkadius, line 35: I wonder if it is a good place to invoke `create()` and use service loader. We usually trying to keep it on very high level like in `ModelData` or `CompiledProcess.compile` just to make sure that we have separation of service loading (where many strange failures regards to classloaders can happen) and simple creation of objects. Maybe we should pass `ObjectNaming` instead of `ObjectNamingProvider` to `KafkaSourceFactory` etc?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
```diff
@@ -0,0 +1,10 @@
+package pl.touk.nussknacker.engine.api.namespaces
+
+trait ObjectNaming {
+  def prepareName(originalName: String, namingContext: NamingContext): String
+}
+
+case object DefaultObjectNaming extends ObjectNaming {
```
lciolecki, line 7: Maybe better will be NoObjectNaming or OriginalObjectNaming?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/NamingContext.scala
```diff
@@ -0,0 +1,12 @@
...
+
+// This class can be extended in the future to provide more
+// information about context (e.g. process name, category)
+// TODO: add context without changing this file
+class NamingContext(val usageKey: ObjectNamingUsageKey.UsageKey)
+
+object ObjectNamingUsageKey extends Enumeration {
+  type UsageKey = Value
+
+  val kafkaTopic: ObjectNamingUsageKey.Value = Value("kafka-topic")
```
lciolecki, line 11: `KafkaTopic: Value = Value("KafkaTopic")`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/NamingContext.scala
```diff
@@ -0,0 +1,12 @@
+package pl.touk.nussknacker.engine.api.namespaces
+
+// This class can be extended in the future to provide more
+// information about context (e.g. process name, category)
+// TODO: add context without changing this file
+class NamingContext(val usageKey: ObjectNamingUsageKey.UsageKey)
```
lciolecki, line 6: It can be case class.

engine/flink/management/sample/src/test/scala/pl/touk/nussknacker/engine/process/NamespacedKafkaSourceSinkTest.scala
```diff
@@ -0,0 +1,94 @@
...
+import pl.touk.nussknacker.engine.kafka._
+import pl.touk.nussknacker.engine.management.sample.DevProcessConfigCreator
+import pl.touk.nussknacker.engine.management.sample.signal.RemoveLockProcessSignalFactory
+import pl.touk.nussknacker.engine.process.compiler.FlinkProcessCompiler
+import pl.touk.nussknacker.engine.spel
+import pl.touk.nussknacker.engine.testing.LocalModelData
+
+
+
+class NamespacedKafkaSourceSinkTest extends FunSuite with BeforeAndAfterAll with KafkaSpec with Matchers {
```
lciolecki, line 25: To many enters ;)

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/ObjectNamingProvider.scala
```diff
@@ -0,0 +1,18 @@
+package pl.touk.nussknacker.engine.util.namespaces
```
lciolecki, line 1: I think it shoul be in api.

mpac123: It can't be without introducing a cyclic dependency. `ObjectNamingProvider` uses `ScalaServiceLoader` which is a member of `util` package - which, in turn, depends on `api` package.

arkadius: But `ObjectNaming` can be in api? It make sens that `ObjectNaming` is part of an api and we have provider for them in utils...

mpac123: Yes, @arkadius, that's what I did.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
mpac123: Hm, or maybe `IdentityObjectNaming`?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/ObjectNamingProvider.scala
arkadius: Indeed, my mistake

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
arkadius: I'm for `IdentityObjectNaming`

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/KafkaAvroSourceFactorySpec.scala
```diff
@@ -2,27 +2,32 @@ package pl.touk.nussknacker.engine.avro
...
   import MockSchemaRegistry._
+
   import collection.JavaConverters._
 
   // schema.registry.url have to be defined even for MockSchemaRegistryClient
-  override lazy val kafkaConfig = KafkaConfig(kafkaZookeeperServer.kafkaAddress,
-    Some(Map("schema.registry.url" -> "not_used")), None)
+  override lazy val config: Config = ConfigFactory.load()
+    .withValue("kafka.kafkaAddress", fromAnyRef(kafkaZookeeperServer.kafkaAddress))
+    .withValue("kafka.kafkaProperties.\"schema.registry.url\"", fromAnyRef("not_used"))
```
lciolecki, line 30: `"kafka.kafkaProperties.schema.registry.url"`

engine/demo/src/main/java/pl/touk/nussknacker/engine/javademo/DemoProcessConfigCreator.java
```diff
@@ -16,23 +9,22 @@
...
-import pl.touk.nussknacker.engine.kafka.KafkaConfig;
 import pl.touk.nussknacker.engine.kafka.KafkaSinkFactory;
 import pl.touk.nussknacker.engine.kafka.KafkaSourceFactory;
 import pl.touk.nussknacker.engine.kafka.serialization.SerializationSchemaFactory;
 import pl.touk.nussknacker.engine.kafka.serialization.schemas;
 import scala.Option;
 import scala.collection.JavaConverters;
 
+import java.io.IOException;
+import java.util.*;
```
lciolecki, line 26: Please import only these elements you need.

mpac123: Hm, looks like they are indeed needed in this class

PR #842, opened 2020-03-30
Allow to put model path by env

lciolecki:


nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -6,11 +6,14 @@ categoriesConfig: {
   "Default": "streaming"
 }
 
+modelClassPath: "model/genericModel.jar"
+modelClassPath: ${?MODEL_CLASS_PATH}
```
arkadius: Why u want to make it configurable? You can use docker volumes if you want to rewrite this model...?

lciolecki: This case is used when I have my own model and own development module with nussknacker-ui as dependencies. I want run nk from IDEA by `pl.touk.nussknacker.ui.NussknackerApp` with base configuration and with my model config.

PR #844, opened 2020-03-31
Improvements: Setup dev env

lciolecki:
Set default paths instead of using envs

### DIFF
--- a/nussknacker-dist/src/universal/conf/dev-application.conf
+++ b/nussknacker-dist/src/universal/conf/dev-application.conf
@@ -14,13 +14,14 @@ categoriesConfig: {
   "ServerRestApi": "request-response"
 }
 
-managementModelDir: "model"
+//By default we set path for development env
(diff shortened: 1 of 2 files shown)


nussknacker-dist/src/universal/conf/dev-application.conf
```diff
@@ -14,13 +14,14 @@ categoriesConfig: {
   "ServerRestApi": "request-response"
 }
 
-managementModelDir: "model"
```
arkadius, line 17: It should be able to run this configuration from docker. You want to enforce user using docker to pass all this values? Please keep in mind, that we are on of users of this config in docker - on our staging env.

lciolecki: As I remeber docker uses `docker-application.conf` file..  This is only dev configuration for IDEA. I have seen thath most of people have problem with run nk from IDEA. Most of problems are envs with these paths..

lciolecki:
I will change this PR. So for now I close it.

PR #854, opened 2020-04-08
FixedValueEditor in dual editor mode for enums

### LINKED ISSUE #853: Fixed value validators are too much restrictive
Currently if you define parameter's type as an Enum, there will be inferred simple, `FixedValuesParameterEditor` for it. This editor cause that will be added validations that the passed expression will be one of provided values.

This behavior won't be good for situations:
1. Someone already used in process other expression in parameter e.g. `#dicts.dict.VALUE` instead of `T(...).VALUE`
2. Someone want to provide value that is not known on process compile time e.g. `#input.value`

arkadius:
fixes #853

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
@@ -10,9 +10,11 @@ class ParameterTypeEditorDeterminer(val runtimeClass: Class[_]) extends Paramete
   override def determine(): Option[ParameterEditor] = {
     runtimeClass match {
       case klazz if klazz.isEnum => Some(
-        FixedValuesParameterEditor(
+        // We can pick here simple editor, but for compatibility reasons we choose dual editor instead
+        // - to be able to provide some other expression
+        DualParameterEditor(FixedValuesParameterEditor(
           possibleValues = runtimeClass.getEnumConstants.toList.map(extractEnumValue(runtimeClass))
-        )
+        ), DualEditorMode.SIMPLE)
       )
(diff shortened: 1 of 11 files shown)


ui/client/components/graph/node-modal/editors/expression/FixedValuesEditor.tsx
```diff
@@ -52,18 +52,17 @@ export default class FixedValuesEditor extends React.Component<Props> {
     const option = this.currentOption()
 
     return (
-      <React.Fragment>
+      <div className={`node-value-select ${className}`}>
```
mproch, line 55: is it necessary to fix editor behaviour?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ParameterTypeEditorDeterminer.scala
```diff
@@ -10,9 +10,11 @@ class ParameterTypeEditorDeterminer(val runtimeClass: Class[_]) extends Paramete
   override def determine(): Option[ParameterEditor] = {
     runtimeClass match {
       case klazz if klazz.isEnum => Some(
-        FixedValuesParameterEditor(
+        // We can pick here simple editor, but for compatibility reasons we choose dual editor instead
+        // - to be able to provide some other expression
+        DualParameterEditor(FixedValuesParameterEditor(
```
mproch, line 15: :+1: for comment

ui/client/components/graph/node-modal/editors/expression/FixedValuesEditor.tsx
arkadius: yep, otherwise was some problems with layout of rendered select box. it was a regression bug introduced during adding support for fixed value editor in process properties - after my change, in both situations it looks ok.

PR #867, opened 2020-04-17
Setting DB schema name support added.

trombka:


ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
```diff
@@ -24,35 +20,29 @@ object DatabaseServer {
...
-      .dataSource(new DatabaseDataSource(dbConfig.db))
-      .baselineOnMigrate(true)
+object DatabaseInitializer {
 
-    val flywayWithDbProfile = dbConfig.driver match {
-      case HsqldbProfile => flyway.locations("db/migration/hsql", "db/migration/common")
-      case PostgresProfile => flyway.locations("db/migration/postgres", "db/migration/common")
-    }
-    flywayWithDbProfile.load().migrate()
+  def initDatabase(path: String, config: Config): Unit = {
```
lciolecki, line 25: There should be only dbConfig.. Why You put path and config as separeted params?

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
```diff
@@ -24,35 +20,29 @@ object DatabaseServer {
...
-  override def setLoginTimeout(seconds: Int): Unit = ???
-  override def getParentLogger: Logger = ???
-  override def getLogWriter: PrintWriter = ???
+  def initDatabase(url: String, user: String, password: String, schema: Option[String] = None): Unit =
+    Flyway
+      .configure()
+      .dataSource(url, user, password)
+      .schemas(schema.toArray: _*)
+      .baselineOnMigrate(true)
+      .locations(
```
lciolecki, line 41: I think we should raise some exception when url doesn't match to these two patterns.

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
trombka: First of all the class named DbConfig is not a configuration class at all.It's a Slick database "handle". :)
Flyway uses a JDBC datasource to configure DB schema, but Slick doesn't expose the instance it uses internally. There was a tricky proxy here mimicking a java.sql.DataSource with a possible connection closing issue. 
Moreover there's no way to extract a db schema that Slick uses when connected to a database.
With this PR the FlyWay initializing method would have exactly the same signature as Slick initialization.

trombka: I'd suggest renaming DbConfig class. It seems confusing now.

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
trombka: I haven't found any generic purpose exception class throughout the source code that would pretty printed to the user in case of wrong configuration. I few places an IllegalArgumentException is thrown. I doubt it is a proper exception to be communicated to the user. Should I use it?

lciolecki: IllegalArgumentException seems to be good choice ;)

trombka: Done.

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
lciolecki: You are right. So ;) Let's do some cleaning here.. If you would be so kind, please do
- rename DbConfig to DbHandler
- create case class for DbConfig with necessary fields

After thath we can use everywhere DbConfig.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/DbTesting.scala
```diff
@@ -46,19 +55,16 @@ trait WithHsqlDbTesting
...
-  val db: DbConfig = DbConfig(JdbcBackend.Database.forURL(
-    url = s"jdbc:hsqldb:mem:esp;sql.syntax_ora=true",
-    driver = "org.hsqldb.jdbc.JDBCDriver",
-    user = "SA",
-    password = "",
-    //we don't use default because it uses too much connections and causes warning, fixed only in: https://github.com/slick/slick/commit/318843a1b81817d800ba14c9c749b6f2045b340c
-    executor = AsyncExecutor.default("Nussknacker-test", 20)
-  ), HsqldbProfile)
+  override val dbProfile = HsqldbProfile
+  override lazy val config: Config = ConfigFactory.parseString(
```
mproch, line 59: maybe parseMap would be better?

trombka: I've changed it.

ui/server/src/main/scala/pl/touk/nussknacker/ui/db/DatabaseInitializer.scala
trombka: To be done in a separate PR.

PR #874, opened 2020-04-22
CustomTransformers and Sinks should have access to ValidationContext

mproch:
Sometimes it would be nice to have access to ValidationContext in custom transformers or sinks - e.g. to prepare better serializers etc.


mproch:
Will implement it in different way probably...

PR #878, opened 2020-04-23
Add configurable namespaces to Flink ESP process names 

mpac123:
Flink ESP process name is now modified with `prepare` function of currently loaded instance of `ObjectNaming`.
This enables running multiple nussknacker instances on a single Flink cluster by e.g. appending unique namespace names to the process names.

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
@@ -4,7 +4,7 @@ import com.typesafe.config.Config
 
 trait ObjectNaming extends Serializable {
   def prepareName(originalName: String, config: Config, namingContext: NamingContext): String
-  def objectNamingParameters(originalName: String, config: Config, namingContext: NamingContext): ObjectNamingParameters
+  def objectNamingParameters(originalName: String, config: Config, namingContext: NamingContext): Option[ObjectNamingParameters]
 }
 
 trait ObjectNamingParameters {
@@ -13,11 +13,5 @@ trait ObjectNamingParameters {
 
 case object DefaultObjectNaming extends ObjectNaming {
   override def prepareName(originalName: String, config: Config, namingContext: NamingContext): String = originalName
-
-  override def objectNamingParameters(originalName: String, config: Config, namingContext: NamingContext): DefaultObjectNamingParameters
-  = DefaultObjectNamingParameters()
-}
-
-case class DefaultObjectNamingParameters() extends ObjectNamingParameters {
-  override def toTags: Map[String, String] = Map()
+  override def objectNamingParameters(originalName: String, config: Config, namingContext: NamingContext): Option[ObjectNamingParameters] = None
 }
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
@@ -3,11 +3,15 @@ package pl.touk.nussknacker.engine.api.namespaces
 import com.typesafe.config.Config
 
 trait ObjectNaming extends Serializable {
-  def prepareName(originalName: String, config: Config, namingContext: NamingContext): String
+  def prepareName(originalName: String, config: Config, namingContext: NamingContext): String = {
+    val separator = getSeparator(config,namingContext)
+    getNamespace(config, namingContext) match {
+      case None => originalName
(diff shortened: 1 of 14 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
```diff
@@ -4,6 +4,7 @@ import com.typesafe.config.Config
 
 trait ObjectNaming extends Serializable {
   def prepareName(originalName: String, config: Config, namingContext: NamingContext): String
+  def getNamespace(config: Config, namingContext: NamingContext): Option[String] = None
```
mproch, line 7: I think by adding this name we make some assumptions about prepareName method? Probably we want to specify them somehow? I mean, we assume that originalName + namespace will determine return value of prepareName?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
```diff
@@ -24,6 +26,17 @@ case class NkGlobalParameters(buildInfo: String,
       "user" -> processVersion.user
     ).filterNot(_._2 == null).asJava
   }
+
+  // This function is used in (not-legacy) `MetricUtils` to add additional tags to all types of metrics.
+  // If you add new tag, remember that it will also influence the measurement name. In order to avoid it,
+  // make sure you modify `telegraf.conf` accordingly.
+  def toTags: Map[String, String] = {
```
mproch, line 33: let's move this logic to MetricUtils, here we shouldn't put logic connected to metrics, just the data?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
mpac123: Hm yes, I made the `prepareName` function depend on `getNamespace` and `getSeparator` functions, do you think it's better this way? If somebody doesn't like the namespace + separator + originalName convention, they still can overwrite the `prepareName` function.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
mpac123: Good point, moved

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/metrics/MetricUtils.scala
```diff
@@ -91,7 +92,18 @@ class MetricUtils(runtimeContext: RuntimeContext) {
     }
   }
 
-
+  implicit class ExtendedNkGlobalParameters(val parameters: NkGlobalParameters) {
```
arkadius, line 95: IMO extension classes are great to handle situation when you have external library and want to add some useful methods, but in our internal classes would be better to introduce just `extractTags(globalParameters)` (it will be more straithforward). What do you think about that?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
```diff
@@ -11,6 +11,8 @@ import pl.touk.nussknacker.engine.api.ProcessVersion
 //Also, those configuration properties will be exposed via Flienk REST API/webconsole
 case class NkGlobalParameters(buildInfo: String,
                               processVersion: ProcessVersion,
+                              processOriginalName: String,
+                              namespace: Option[String],
```
arkadius, line 15: I wonder if it would be better to keep here only core params and for extensions add something like `rawParameters` here and extract typed object in place of extension. WDYT? Please consider situation: someone want to have some weird naming scenario like: name + namespace + mainPart. To achive that it will add here one more parameter `name: Option[String]`? Or better one: he want to have namespace as a list.

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/metrics/MetricUtils.scala
mpac123: Yes I think you're right, will do

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
mpac123: Yes, it should be clearer this way, I'll change it

PR #879, opened 2020-04-23
[Improvement] Use new metrics for demo, easier configuration

mproch:
- demo uses InfluxDB reporter and tags in Flink metrics
- old metrics can be used with globalParameters.useLegacyMetrics config
- exploring ExecutionConfig.globalParameters for passing configuration around


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
```diff
@@ -0,0 +1,34 @@
+package pl.touk.nussknacker.engine.flink.api
```
arkadius, line 1: It looks like all this flags are not Flink specific but only paring them from `ExecutionConfig`? Maybe it should be in common api for purpose of other engines?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
```diff
@@ -0,0 +1,34 @@
...
+      case Some(x) => x.toString
+      case null => "null"
+      case x => x.toString
+    }.asJava
+  }
+}
+
+object NKGlobalParameters {
+
+  def readFromContext(ec: ExecutionConfig): Option[NKGlobalParameters] = Option(ec.getGlobalJobParameters).collect {
```
arkadius, line 30: Shouldn't we read some of params like `useLegacyMetrics` directly from type safe's config and pass to global parameters only those that are determined on ui side (like `version`)? I wonder if this `getGlobalJobParameters` won't grow too much when we pass a lot of options (especially in nested structures, lists etc.)

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
```diff
@@ -0,0 +1,34 @@
...
+import java.util
+import scala.collection.JavaConverters._
+
+import org.apache.flink.api.common.ExecutionConfig
+import org.apache.flink.api.common.ExecutionConfig.GlobalJobParameters
+
+//we can use this class to pass config through RuntimeContext to places where it would be difficult to use otherwise
+//Also, those configuration properties will be exposed via Flienk REST API/webconsole
+//TODO: 
+case class NKGlobalParameters(buildInfo: String,
```
arkadius, line 12: Maybe `Nk` prefix to be concise with `Esp*` classes?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
arkadius: ... but then will be nice to have also `NkFlinkGlobalParameters(common: NkGlobalParameters, ...)` to be able to add some Flink's specific :)  If you prefer to just suffix it with `NkFlink...` it will be also fine.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
mproch: ok

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
mproch: I'll stick with Flink specific for now. I don't see much usage outside, and we would have to circumvent somehow subclassing GlobalJobParameters...

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
mproch: hmm... true that... I'll prepare sth

mproch: addressed in https://github.com/TouK/nussknacker/pull/884

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NKGlobalParameters.scala
mproch: addressed in https://github.com/TouK/nussknacker/pull/884

Crack:
👍 from me, I like this change.

Just note that passing metrics throuh Telegraf requires one of:
* Telegraf 1.14+ (it doesn't drop entire writes due to invalid metric values - influxdata/telegraf/issues/6124)
* flink-metrics-influxdb 1.10+ (it filters out invalid metric values - https://issues.apache.org/jira/browse/FLINK-12147)

PR #882, opened 2020-04-24
Explicit uids in operators - first step (turned off by default)

arkadius:
The purpose of this change is to make checkpoints recoverable after migration between Flink versions. For now we use generated uids which will change if layout of process has been changed or algorithm of generation of this uids will change.

Uids are set up in:
- sources
- sinks
- async functions
- custom nodes provided in generic model

Changes that are not mentioned in MigrationGuide, because there was no version cut:
- `FlinkSource.sourceStream` now takes `FlinkCustomNodeContext` instead of `MetaData`
- `FlinkSink.registerSink` now takes `FlinkCustomNodeContext` instead of `FlinkLazyParameterFunctionHelper`


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/compat/ExplicitUidInOperatorsCompat.scala
```diff
@@ -0,0 +1,67 @@
...
+  protected def explicitUidInStatefulOperators(nodeCtx: FlinkCustomNodeContext): Boolean =
+    ExplicitUidInOperatorsCompat.defaultExplicitUidInStatefulOperators(nodeCtx)
+
+}
+
+object ExplicitUidInOperatorsCompat {
+
+  def setUidToNodeIdIfNeed[T](explicitUidInStatefulOperators: Boolean, nodeCtx: FlinkCustomNodeContext)
+                             (stream: DataStream[T]): DataStream[T] = {
+    if (explicitUidInStatefulOperators)
```
alien11689, line 41: All 3 methods body are identical - can we extract method?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/compat/ExplicitUidInOperatorsCompat.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def setUidToNodeIdIfNeedJava[T](explicitUidInStatefulOperators: Boolean, nodeCtx: FlinkCustomNodeContext)
+                                 (stream: SingleOutputStreamOperator[T]): SingleOutputStreamOperator[T] = {
+    if (explicitUidInStatefulOperators)
+      stream.uid(nodeCtx.nodeId)
+    else
+      stream
+  }
+
+  // FIXME: replace with config when it be available in FlinkCustomNodeContext
```
alien11689, line 63: Who is gonna watch TODO and FIXME comments? Do we have some register different to `grep`?
I'd prefer to create an issue for fixinig/changing it and add issue id to FIXME comment.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkCustomNodeContext.scala
```diff
@@ -7,6 +7,7 @@ import scala.concurrent.duration.FiniteDuration
 import scala.reflect.ClassTag
 
 case class FlinkCustomNodeContext(metaData: MetaData,
+                                  // TODO: it can be used in state recovery - make sure that it won't change during renaming of nodes on gui
                                   nodeId: String,
```
alien11689, line 11: I don't know how many raw Strings are in NK, but maybe it would be a good idea to create a value object here?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/compat/ExplicitUidInOperatorsCompat.scala
arkadius: It is still draft change :) FIXME is only an information for my own

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkCustomNodeContext.scala
arkadius: It will be a painfull change. This context is used in all custom transformers in user's implementations. Also I think that it will be a good moment to add value object when we will split nodeId into nodeName and nodeId. For now it is mixed.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/compat/ExplicitUidInOperatorsCompat.scala
arkadius: They looks similar but Flink hasn't got common interface for all stream types.

arkadius:
> Don't we need some tests for this change?\
I am planning to switch this newly introduced feature flag on in integration tests. I think that it will cover some cases because we have verfiications that state is recoverable there. Have you any other idea what to test?

PR #892, opened 2020-04-28
[FIX] Validate generate data input, closes #1

### LINKED ISSUE #1: Validate amount of test data to generate in UI
Clicking on `Generate` button shows a form and it allows for request the amount of test data. But the number is not validated and when I provide `bla` or `1000000000000000000000000000000000` then any error message is not shown. File is generated and contains only one line:

```
The requested resource could not be found.
```

mproch:
Fixes #1

### DIFF
--- a/ui/client/components/modals/GenerateTestDataDialog.js
+++ b/ui/client/components/modals/GenerateTestDataDialog.js
@@ -24,6 +24,10 @@ class GenerateTestDataDialog extends React.Component {
     return HttpService.generateTestData(this.props.processId, this.state.testSampleSize, this.props.processToDisplay)
   }
 
+  onInputChange = (event) => {
+    this.setState({testSampleSize: event.target.value})
+  }
+
   render() {
     const validators = [literalIntegerValueValidator, mandatoryValueValidator]
     return (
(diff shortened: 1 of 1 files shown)


ui/client/components/modals/GenerateTestDataDialog.js
```diff
@@ -22,14 +25,23 @@ class GenerateTestDataDialog extends React.Component {
...
+        okBtnConfig={{disabled: !allValid(validators, [this.state.testSampleSize])}}
         type={Dialogs.types.generateTestData}
       >
         <p>Generate test data</p>
-        <input autoFocus={true} className="add-comment-on-save" value={this.state.testSampleSize} onChange={(e) => { this.setState({testSampleSize: e.target.value}) } }/>
+        <input
+          autoFocus={true}
+          className="add-comment-on-save"
+          value={this.state.testSampleSize}
+          onChange={(e) => { this.setState({testSampleSize: e.target.value}) } }
```
lciolecki, line 41: This should be defined as method at GenerateTestDataDialog.

mproch: ok, done

PR #913, opened 2020-05-01
Fix for: methods returning Futures

arkadius:
Futures were flatten during methods extraction instead what during method definition extraction

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/MethodDefinitionExtractor.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/MethodDefinitionExtractor.scala
@@ -6,7 +6,7 @@ import java.lang.reflect.Method
 import pl.touk.nussknacker.engine.api._
 import pl.touk.nussknacker.engine.api.definition._
 import pl.touk.nussknacker.engine.api.process.SingleNodeConfig
-import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypingResult, Unknown}
+import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypedClass, TypingResult, Unknown}
 import pl.touk.nussknacker.engine.definition.MethodDefinitionExtractor.{MethodDefinition, OrderedDependencies}
 import pl.touk.nussknacker.engine.definition.validator.ValidatorsExtractor
 import pl.touk.nussknacker.engine.types.EspTypeUtils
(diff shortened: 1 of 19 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/types/EspTypeUtils.scala
```diff
@@ -199,6 +193,13 @@ object EspTypeUtils {
       Typed.genericTypeClass(rawType, paramsType.getActualTypeArguments.toList.flatMap(extractClass))
     } else if (classOf[scala.collection.Iterable[_]].isAssignableFrom(rawType)) {
       Typed.genericTypeClass(rawType, paramsType.getActualTypeArguments.toList.flatMap(extractClass))
+    // All below are for AbstractMethodDefinitionExtractor purpose
```
mproch, line 196: Can we refactor it a bit? I think for all types except StateT we do Typed.genericTypeClass(rawType, paramsType.getActualTypeArguments.toList.flatMap(extractClass))?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/MethodDefinitionExtractor.scala
```diff
@@ -96,7 +96,14 @@ private[definition] trait AbstractMethodDefinitionExtractor[T] extends MethodDef
       Option(method.getAnnotation(classOf[MethodToInvoke])).map(_.returnType())
       .filterNot(_ == classOf[Object])
       .map[TypingResult](Typed(_))
-    val typeFromSignature = EspTypeUtils.getGenericType(method.getGenericReturnType).map(Typed(_))
+    val typeFromSignature = {
+      val rawType = EspTypeUtils.getReturnClassForMethod(method)
+      (expectedReturnType, rawType) match {
+        // uwrap Future, Source and so on
```
mproch, line 102: I wonder if we should have explicit list here? I don't think I understand the Source part - we don't handle it in inferGenericMonadType??

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/types/EspTypeUtils.scala
arkadius: Yep, I thought that It was explicitly specified because of some kind of optimization? But even if so, it may be true that it was premature optimalization... I will change it but in last PR: https://github.com/TouK/nussknacker/pull/918 to avoid fight with rebasing... Good catch!

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/MethodDefinitionExtractor.scala
arkadius: IMO it not necessary - if someone has monad with one generic param in `expectedReturnType` we can assume that we are interested of this generic param. Adding explicit list here will add not necessary coupling. WDYT? Regards to Source, from what I saw in the code, we never handle type extraction correctly - if you haven't specified `returnType`, the type was always `Unknown`.

mproch: Ok... let's leave it then... although I have some feeling that we have too many assumptions here...

PR #920, opened 2020-05-03
KafkaConfig programitically

arkadius:
… KafkaSource


engine/flink/generic/src/main/scala/pl/touk/nussknacker/genericmodel/GenericConfigCreator.scala
```diff
@@ -67,7 +65,7 @@ class GenericConfigCreator extends EmptyProcessConfigCreator {
   import pl.touk.nussknacker.engine.util.functions._
 
   override def expressionConfig(processObjectDependencies: ProcessObjectDependencies): ExpressionConfig = {
-    val kafkaConfig = processObjectDependencies.config.as[KafkaConfig]("kafka")
+    val kafkaConfig = KafkaConfig.parseConfig(processObjectDependencies.config, "kafka")
```
lciolecki, line 68: Hmm I think better way will be do some method like kafaConfig at ProcessObjectDependencies? Now we have many of copy paste `parseConfig`..

arkadius: We still keep our application `kafka` agnostic. I think that it will be less drawbacks from copy-paste this one-liner in some places then lost this loosely coupling...

lciolecki: Okej, so my proposition is do 2 methods in companion object KafkaConfig:
- `parseConfig(processObjectDependencies: ProcessObjectDependencies, namespace: String)`
- `parseConfig(processObjectDependencies: ProcessObjectDependencies)` with default  namespace "kafka"

PR #921, opened 2020-05-04
Branch parameters work better when editing nodes

mproch:


ui/client/reducers/graph.js
```diff
@@ -414,3 +421,54 @@ function createEdge(fromNode, toNode, edgeType, allEdges, processDefinitionData)
...
+                language: "spel",
+              },
+            })
+      })
+      return {
+        branchId: branchId,
+        parameters: newBranchParams,
+      }
+    })
+    delete node["branchParametersTemplate"]
```
arkadius, line 471: From my tests it looks like now we shouldn't delete template, because this node will be reused during another connection adding...

mproch: good catch, done

ui/client/reducers/graph.js
```diff
@@ -162,20 +164,25 @@ export function reducer(state, action) {
...
     case "NODES_CONNECTED": {
       const edge = createEdge(action.fromNode, action.toNode, action.edgeType, state.processToDisplay.edges, action.processDefinitionData)
+      const newEdges = _.concat(state.processToDisplay.edges, edge)
       return {
         ...state,
         processToDisplay: {
           ...state.processToDisplay,
-          edges: _.concat(state.processToDisplay.edges, edge),
+          nodes: state.processToDisplay.nodes.map(n => action.toNode.id === n.id ?
+            enrichNodeWithProcessDependentData(action.toNode, action.processDefinitionData, newEdges) : n),
```
arkadius: From my observation `action.toNode` is not the correct one to take (it hasn't got `branchParameters`). Everything works fine with `n`

PR #922, opened 2020-05-04
Split healthCheck into: app-is-not-running and deployed-process-failed

gskrobisz:
Split healthCheck into: app-is-not-running (healthCheck) and app-is-running-but-deployed-process-failed (processHealthCheck).


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -45,35 +46,53 @@ class AppResources(config: Config,
     }
   }
 
+  private def toEntity(healthCheckProcessResponse: HealthCheckProcessResponse): ResponseEntity = {
```
mproch, line 49: why do we need it here? isn't FailFastCirceSupport enough?

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -45,35 +46,53 @@ class AppResources(config: Config,
...
               case NonFatal(e) =>
                 logger.error("Failed to get statuses", e)
-                HttpResponse(status = StatusCodes.InternalServerError, entity = "Failed to retrieve job statuses")
+                val respEntity = toEntity(HealthCheckProcessResponse(ERROR, Some("Failed to retrieve job statuses")))
+                HttpResponse(status = StatusCodes.InternalServerError, entity = respEntity)
             }
           }
         }
-      } ~ path("sanityCheck")  {
+      } ~ path("healthCheck" / "process" / "validation")  {
```
mproch, line 86: let's try to be a bit more consistent. Either:
- validation and deployment 
- validated and deployed?

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala
```diff
@@ -50,30 +51,47 @@ class AppResources(config: Config,
       path("healthCheck") {
         get {
           complete {
-            notRunningProcessesThatShouldRun.map[HttpResponse] { set =>
+            Marshal(HealthCheckProcessResponse(OK))
```
lciolecki, line 54: You can extract this to some method like:

```
  private def healthCheckResponse(response: HealthCheckProcessResponse, statuCode: StatusCode) =
    Marshal(response)
      .to[ResponseEntity]
      .map(res => HttpResponse(status = statuCode, entity = res))

**Comment by lciolecki** [comment]
@gskrobisz add some info about this PR in Changelog.

File: ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala (hunk 3 above)
Comment (lciolecki) [reply ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala:54]: You can also do some method maped HealthCheckProcessResponseStatus to StatusCodes and use it here.

File: ui/server/src/main/scala/pl/touk/nussknacker/ui/api/AppResources.scala [hunk 4]
```diff
@@ -45,35 +46,50 @@ class AppResources(config: Config,
     }
   }
 
+  private def createHealthCheckHttpResponse(healthCheckResponse: HealthCheckProcessResponse): Future[HttpResponse] =
```
lciolecki, line 49:
```
status: HealthCheckProcessResponseStatus, message: Option[String]
```
Now you every thime repeat HealthCheckProcessResponse(..). We don't need it.

PR #924, opened 2020-05-04
Checking of method invocation parameter types

arkadius:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -1,22 +1,20 @@
...
 import pl.touk.nussknacker.engine.types.EspTypeUtils
 
 object TypeMethodReference {
-  def apply(methodReference: MethodReference, currentResults: List[TypingResult])(implicit settings: ClassExtractionSettings): Either[String, TypingResult] =
-    new TypeMethodReference(methodReference, currentResults).call
+  def apply(methodName: String, currentResults: List[TypingResult], params: List[TypingResult])(implicit settings: ClassExtractionSettings): Either[String, TypingResult] =
+    new TypeMethodReference(methodName, currentResults, params).call
 }
 
 // TODO:
```
mproch, line 13: shall we remove this comment now?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -27,39 +25,64 @@ class TypeMethodReference(methodReference: MethodReference, currentResults: List
...
+
+    validatedType match {
+      case Left(None) => Right(Unknown) // we use Left(None) to indicate situation when we want to skip further validations because of lack of some knowledge
+      case Left(Some(message)) => Left(message)
+      case Right(returnType) => Right(returnType)
+    }
+  }
+
+  private def validateClassDefinitionsNonEmpty(clazzDefinitions: List[ClazzDefinition]): Either[Option[String], List[ClazzDefinition]] =
+    if (clazzDefinitions.isEmpty) Left(None) else Right(clazzDefinitions)
```
mproch, line 49: does empty list represent Unknown here?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -27,39 +25,64 @@ class TypeMethodReference(methodReference: MethodReference, currentResults: List
...
+      case Nil => Left(Some(s"Unknown method '$methodName' in $displayableType"))
+      case methodInfoes => Right(methodInfoes)
     }
+  }
 
-  //TODO: we check only arity, but don't check if any of overloaded methods has matching signature
-  //this will lead to erros if we have different return types for different signatures!
-  private def typeFromMethodInfoes(methodInfoes: List[MethodInfo]): Either[String, TypingResult] =
-    methodInfoes.filter(_.parameters.size <= paramsCount) match {
+  private def validateMethodArity(methodInfoes: List[MethodInfo]): Either[Option[String], List[MethodInfo]] = {
```
mproch, line 62: is this necessary now?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -27,39 +25,64 @@ class TypeMethodReference(methodReference: MethodReference, currentResults: List
...
-        val typingResult = Typed(returnTypes.toSet)
-        Right(typingResult)
+        Right(nonEmpty)
     }
+  }
 
+  private def validateMethodParameterTypes(methodInfoes: List[MethodInfo]): Either[Option[String], List[TypingResult]] = {
+    val returnTypesForMatchingMethods = methodInfoes.flatMap { m =>
+      val allMatching = m.parameters.map(_.refClazz).zip(calledParams).forall {
+        case (declaredType, passedType) => passedType.canBeSubclassOf(declaredType)
```
mproch, line 74: can we have some test for cases like Long in signature and Int in parameter? I know we have CanBeSubclassDeterminer, but I'd fell safer :)
what about varargs? how does spel handle them?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -27,39 +25,64 @@ class TypeMethodReference(methodReference: MethodReference, currentResults: List
...
-        val returnTypes = nonEmpty.map(_.refClazz)
-        val typingResult = Typed(returnTypes.toSet)
-        Right(typingResult)
+        Right(nonEmpty)
     }
+  }
 
+  private def validateMethodParameterTypes(methodInfoes: List[MethodInfo]): Either[Option[String], List[TypingResult]] = {
+    val returnTypesForMatchingMethods = methodInfoes.flatMap { m =>
+      val allMatching = m.parameters.map(_.refClazz).zip(calledParams).forall {
```
mproch, line 73: this zip can omit some subtle edge cases when we have not same arity, but same types in common prefix?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: IMO we shall not remove part with arity, because we still checking only sublist of parameters when invocation is shorted than declaration. But indeed second part is to removal

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: Yep, I've described this monadic rule only in comment here `case Left(None)`. Is it enough or you think that we should add more descriptive type for left part?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: If I will change `validateMethodParameterTypes` a bit, it won't be necessary. It can be a good idea because there is more details in error...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: Indeed, there is hidden logic that it based on result of `validateMethodArity`. I can merge both methods so it will be more visible that we can't call methods passing less arguments than it is required

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: I can add more tests. Regards to varargs, I suppose that there is Array in the declaration and then types won't match. I will add test and some TODO for that...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: Ok, after I discovered `Map[String, MethodInfo]` now I know why was this comment. I've removed it, now we check all overloaded methods.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: I've removed `validateMethodArity` and added checking of sizes here.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
arkadius: I've added two ignored test cases + todo

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
mproch: No, it's ok, maybe a comment would be helpful

PR #932, opened 2020-05-06
[FIX] interpreter.open() was not called in standalone tests, closes #931

### LINKED ISSUE #931: open() is not called on service in tests from file for standalone runtime
**Describe the bug**
open() from StandaloneContextLifecycle/Lifecycle is not called for Service during test from file in UI for standalone runtime.

**To Reproduce**
1. Create model with Service implementing StandaloneContextLifecycle
2. In the Service override open(jobData: JobData, runtimeContext: StandaloneContext) and throw some exception in implementation or log some message
3. Build a process with the service and run test from file

mproch:
fixes #931

### DIFF
--- a/engine/standalone/engine/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/StandaloneProcessConfigCreator.scala
+++ b/engine/standalone/engine/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/StandaloneProcessConfigCreator.scala
@@ -1,6 +1,6 @@
 package pl.touk.nussknacker.engine.management.sample
 
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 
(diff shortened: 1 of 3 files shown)


engine/standalone/engine/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/StandaloneProcessConfigCreator.scala
```diff
@@ -97,12 +97,23 @@ object ProcessorService {
   def clear() = invocationsCount.set(0)
 }
 
-class ProcessorService extends Service {
+class ProcessorService extends Service with Lifecycle {
+
+  private var initialized = false
```
raphaelsolarski: Maybe `AtomicBoolean`?

mproch: done

PR #934, opened 2020-05-06
Save layout and groups to process metadata

JulianWielga:


ui/client/actions/nk/process.js
```diff
@@ -49,12 +50,16 @@ export function displayCurrentProcessVersion(processId) {
   return fetchProcessToDisplay(processId)
 }
 
-export function saveProcess(processId, processJson, comment) {
-  return (dispatch) => {
+export function saveProcess(comment) {
+  return (dispatch, getState) => {
+    dispatch({type: "APPEND_METADATA"})
```
JulianWielga, line 55: is this the right place and time? in this way it is not possible to save the layout itself

PR #937, opened 2020-05-07
Docker update latest option disabled for backports

arkadius:


.travis.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         - npm -v
       script:
         - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin;
-        - ./ciBuildDocker.sh --version=${BUILD_VERSION} --docker-publish-type=publishLocal --add-dev-model=$ADD_DEV_MODEL --docker-update-latest=false
+        - ./ciBuildDocker.sh --version=${BUILD_VERSION} --docker-publish-type=publishLocal --add-dev-model=$ADD_DEV_MODEL
```
raphaelsolarski, line 111: `--docker-update-latest` is by default set to `true` in `ciBuildDocker.sh`.
Do we need to tag this image as latest locally (I wonder because it is not published as `latest` outside in this scenario)?

arkadius: Good catch! I've removed default from there, added some comment and better protected from passing invalid value in `.travis.yml`. Also pushed the same HEAD to preview branch to make sure if "staging docker builds" pipeline is still ok after this change: https://travis-ci.org/github/TouK/nussknacker/builds/684327262. Thank you for the review!

release.sh
```diff
@@ -1,7 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -e
 
+if [[ "$BACKPORT" == "true" ]]; then
+  echo "Backport release - 'latest' tag won't be updated on docker"
+  dockerUpdateLatest="false"
```
lciolecki, line 7: Correct me if I'm wrong.. This backport we will use when eg. actual version is 5.4 and we want to do some changes to < 5.4 versions?

ciBuildDocker.sh
```diff
@@ -5,7 +5,7 @@ set -e
 version=`echo ${APPLICATION_VERSION} | sed 's/[^a-zA-Z0-9-]/\_/g' | awk '{print tolower($0)}'`
```
lciolecki, line 5: I wonder if we really need this file.. We can move all these logics to sbt and use it with envs..

release.sh
arkadius: Exactly let's say to 4.1. Then you don't want to change someones latest do lower version...

ciBuildDocker.sh
arkadius: Maybe we don't but this change is conservative one (I don't have ability to deeply test it).

PR #938, opened 2020-05-07
[Improvement] open/close only services that are used in process

mproch:


mproch:
We have to rethink it, in this version it will still open too many services...

PR #942, opened 2020-05-08
Travis ignore tagged sbt version

arkadius:


arkadius:
It seemed to be quick fix, but looks like we have sort of hacks for FE_ONLY (copied before_script in a few places)

PR #944, opened 2020-05-08
Feature: Pluggable SchemaRegistry Cache

lciolecki:
* Introduction to Cache
* Pluggable cache for CachedConfluentSchemaRegistryClient


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/DefaultSchemaCache.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class DefaultSchemaCache extends SchemaCache {
+
+  import scalacache.caffeine._
+  import scalacache.modes.sync._
+
+  private val caffeineClientCache: Cache[String, Entry[Schema]] =
+    Caffeine
+      .newBuilder()
+      .maximumSize(10000L)
```
arkadius, line 17: Can you extract it in constructor and set keep it in const in `CachedConfluentSchemaRegistryClient`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/SchemaCache.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.engine.avro.cache
+
+import java.io.Closeable
+
+import org.apache.avro.Schema
+
+import scala.concurrent.duration.Duration
+
+trait SchemaCache extends Closeable {
```
arkadius, line 9: I wonder if we need to specify that it is "schema" cache? Can it be just Simple cache with key as String and value as a generic parameter `T`? and move it to some util package?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/DefaultSchemaCache.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def getOrCreate(key: String, ttl: Option[Duration], opSchema: => Schema): Schema = {
+    val schema = cache.caching(key)(ttl) {
+      opSchema
+    }
+    schema
+  }
+
+  override def close(): Unit = {
+    caffeineClientCache.cleanUp()
+    cache.removeAll()
```
arkadius, line 32: From what I saw in scalacache / caffeine code, `cleanUp` should be enough. Why you've invoked also `removeAll`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/DefaultSchemaCache.scala
lciolecki: Good Idea!

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/SchemaCache.scala
lciolecki: Yep, we can do it - it will be some base Cache for nk.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryClient.scala
```diff
@@ -1,9 +1,11 @@
 package pl.touk.nussknacker.engine.avro.schemaregistry
 
+import java.io.Closeable
+
 import cats.data.Validated
 import org.apache.avro.Schema
 
-trait SchemaRegistryClient extends Serializable {
+trait SchemaRegistryClient extends Serializable with Closeable {
```
mproch, line 8: Where do we close it in our code? E.g. in SourceFactory, etc. ?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/cache/DefaultSchemaCache.scala
lciolecki: I've removed all close methods. GC should clean up all.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryClient.scala
lciolecki: 've removed all close methods. GC should clean up all.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
```diff
@@ -0,0 +1,79 @@
...
+  * @tparam T
+  */
+class DefaultCache[T](maximumSize: Long, expireAfterAccess: Option[Duration], registerStats: Boolean) extends Cache[T] {
+
+  import scalacache.caffeine._
+  import scalacache.modes.sync._
+
+  import scala.concurrent.duration._
+
+  private val stats = new ConcurrentStatsCounter
```
mproch, line 21: Why we don't just use .recordStats() as in: https://github.com/ben-manes/caffeine/wiki/Statistics?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClientFactory.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.avro.Schema
+import pl.touk.nussknacker.engine.kafka.KafkaConfig
+import pl.touk.nussknacker.engine.util.cache.DefaultCache
+
+import scala.concurrent.duration.Duration
+
+class CachedConfluentSchemaRegistryClientFactory(maximumSize: Long, latestSchemaTtl: Option[Duration], expireAfterAccess: Option[Duration])
+  extends ConfluentSchemaRegistryClientFactory {
+
+  //Cache engine is shared by many of CachedConfluentSchemaRegistryClient
```
mproch, line 12: let's add "the class is Serializable (e.g. during Flink process creation) and cache is not shared between serialized instances" so that no one is mistaken

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
lciolecki: It's more complicated.. Caffaine doesn't support ttl per key, we use only expierAfterAccess for all keys.

In our case we use two types of expiration:
- expierAfterAccess for all keys
- ttl for key from scalacache: https://github.com/cb372/scalacache/blob/master/modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala#L25

It's more complex detecting and cleanup cache. For that I have to use own stats counter.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClientFactory.scala
lciolecki: Done

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
```diff
@@ -0,0 +1,79 @@
...
+  implicit private val cache: CaffeineCache[T] =
+    CaffeineCache(caffeineClientCache)
+
+  override def getOrCreate(key: String, ttl: Option[Duration])(value: => T): T = {
+    if (ttl.exists(time => expireAfterAccess.exists(_.lt(time)))) {
+      throw new IllegalArgumentException(s"TTL (${ttl.get}) should be lower than ExpireAfterAccess (${expireAfterAccess.get}).")
+    }
+
+    cache
+      .doGet(key)
```
arkadius, line 43: This approach won't work correctly because of lack of "cirtical section" wrapping `get` and `put`. Consider situation:
1. Thread A doing doGet - return None
2. Thread B doing doGet - return None again
3. Thread A doing doPut
4. Thread B doing doPut
Shouldn't you use: https://cb372.github.io/scalacache/docs/memoization.html instead?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
arkadius: So maybe we should split this cache instance into two caches (one for parsed schemas) and one for (latest -> schema)? I don't like this idea of own counters handling - take a look at my notice about "cirtical section". There is a log of things that can go wrong.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
lciolecki: Look now.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
lciolecki: I've removed scalacache - there is only Caffeine.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
```diff
@@ -11,69 +10,37 @@ import scala.concurrent.duration.Duration
...
 
-    expireAfterAccess.foreach(expire => builder.expireAfterAccess(java.time.Duration.ofMillis(expire.toMillis)))
-
-    builder.build[String, Entry[T]]
-  }
-
-  implicit private val cache: CaffeineCache[T] =
-    CaffeineCache(caffeineClientCache)
+    expireAfterAccess
+      .foreach(expire => builder.expireAfterAccess(java.time.Duration.ofMillis(expire.toMillis)))
```
arkadius, line 27: We should use `expireAfterWrite`, because you can has many read accesses but still want to invalidate after a some delay from write...

lciolecki: I will add these three options..

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClient.scala
```diff
@@ -10,18 +10,17 @@ import pl.touk.nussknacker.engine.avro.schemaregistry.SchemaRegistryError
...
-class CachedConfluentSchemaRegistryClient(val client: CSchemaRegistryClient, schemaCache: Cache[Schema], latestSchemaTtl: Option[Duration]) extends ConfluentSchemaRegistryClient with LazyLogging {
+class CachedConfluentSchemaRegistryClient(val client: CSchemaRegistryClient, schemaCache: Cache[Schema], latestSchemaCache: Cache[Schema])
+  extends ConfluentSchemaRegistryClient with LazyLogging {
 
   private val latestCacheNamespace = "latest"
 
   override def getLatestSchema(subject: String): Validated[SchemaRegistryError, Schema] = {
     handleClientError {
-      schemaCache.getOrCreate(s"$subject-$latestCacheNamespace", latestSchemaTtl) {
+      latestSchemaCache.getOrCreate(s"$subject-$latestCacheNamespace") {
```
arkadius, line 23: you don't need this trick with `latestCacheNamespace` anymore - you can just pass `subject`

lciolecki: Done.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/DefaultCache.scala
lciolecki: You are right but.. Now I see thaht we should set expireAfterWrite for latest and expireAfterAccess for versions. I've added it.

PR #945, opened 2020-05-08
[FIX] workaround for problems with accessors on primitives/wrappers

mproch:


engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -607,6 +607,29 @@ class SpelExpressionSpec extends FunSuite with Matchers with EitherValues {
     parseWithDicts[Boolean]("#stringValue == #enum['one']", withObjVar, dicts) shouldBe 'invalid
   }
 
+  test("invokes methods on primitives correctly") {
+    def invokeAndCheck[T:TypeTag](expr: String, result: T): Unit = {
+      val parsed = parseOrFail[T](expr)
+      (1 to 5).foreach { _ =>
```
jedrz, line 613: Is this loop required?

mproch: I'll add comment. Expressions are often compiled to bytecode only after second usage. To make sure we're there I rounded up to 5 :)

PR #947, opened 2020-05-11
[Improvement] turn fatal on warnings on again

### LINKED ISSUE #709: Turn scala compiler option: fatal-warnings on again
As it was mentioned by @fijolekProjects in https://github.com/TouK/nussknacker/pull/342#discussion_r381509037 
, we can turn on fatal-warnings if we will add silencer plugin to sbt.

It can protect us from some runtime detected bug like `MatchError` in pattern matching.

mproch:
closes #709

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ import ReleaseTransformations._
 import scala.util.Try
 
 val scala211 = "2.11.12"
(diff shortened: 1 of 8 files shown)


mproch:
Done in different PR

PR #949, opened 2020-05-12
[Improvement] JVM options can be configured via env variable

### LINKED ISSUE #797: Make JVM parameters configurable
Right now application uses JVM defaults and there is no easy way to override that. It works for local development, but in production we should be able to customize JVM parameters. This would also help with memory usage od demo, which currently eats up 1-1,5 GB (default Xms/Xmx on my system), while it probably needs no more than 128 or 256 MB.

mproch:
Closes #797


lciolecki:
@mproch let's add Changelog info about this.

PR #953, opened 2020-05-12
[FIX] validate java vararg methods

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -71,4 +71,17 @@ class TypeMethodReference(methodName: String, currentResults: List[TypingResult]
...
+    val nonVarArgSize = method.parameters.size - 1
+    if (method.varArgs && calledParams.size >= nonVarArgSize) {
+      val nonVarArgParams = method.parameters.take(nonVarArgSize)
+      val nonVarArgCalledParams = calledParams.take(nonVarArgSize)
+      val nonVarArgMatching = nonVarArgParams.map(_.refClazz).zip(nonVarArgCalledParams).forall {
+        case (declaredType, passedType) => passedType.canBeSubclassOf(declaredType)
+      }
+      //TODO: we do not check var arg parameter types
+      if (nonVarArgMatching) Some(method.refClazz) else None
+    } else None
```
arkadius, line 84: can u add here braces:
```
} else {
  None
}
```
because can't see which `else` you handle here

mproch: ok

PR #954, opened 2020-05-12
Typed map fixes

### LINKED ISSUE #674: When using TypedMap Map methods are suggested but do not validate
TypedMap has inconsistent behaviour. When we suggest methods in UI we have methods like empty, containsKey etc. but they do not validate on backend.
`
Screenshot from 2020-02-11 09-25-23

arkadius:
fixes #674

### DIFF
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -49,6 +49,11 @@ val source = new KafkaAvroSourceFactory(schemaRegistryProvider, processObjectDep
 val avroFixedSourceFactory = FixedKafkaAvroSourceFactory[GenericData.Record](processObjectDependencies)
 ```
 
+## In version 0.1.2 (not released yet)
+
+* [#954](https://github.com/TouK/nussknacker/pull/954) `TypedMap` is not a case class wrapping scala Map anymore. If you have
+ done some pattern matching on it, you should use `case typedMap: TypedMap => typedMap.typedMap.asScala` instead.
+
 ## In version 0.1.1
 
 * [#930](https://github.com/TouK/nussknacker/pull/930) `DeeplyCheckingExceptionExtractor` was moved from `nussknacker-flink-util`
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -52,7 +52,7 @@ val avroFixedSourceFactory = FixedKafkaAvroSourceFactory[GenericData.Record](pro
 ## In version 0.1.2 (not released yet)
 
 * [#954](https://github.com/TouK/nussknacker/pull/954) `TypedMap` is not a case class wrapping scala Map anymore. If you have
- done some pattern matching on it, you should use `case typedMap: TypedMap => typedMap.typedMap.asScala` instead.
+ done some pattern matching on it, you should use `case typedMap: TypedMap => typedMap.asScala` instead.
 
 ## In version 0.1.1
 
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypedMap.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypedMap.scala
@@ -1,3 +1,21 @@
 package pl.touk.nussknacker.engine.api.typed
 
-case class TypedMap(fields: Map[String, Any])
+import java.util.Collections
+import java.{util => ju}
+
+class TypedMap(map: ju.Map[String, Any]) extends ju.HashMap[String, Any](map) {
+
+  def this() =
+    this(Collections.emptyMap())
+
+}
+
+object TypedMap {
+
+  import scala.collection.JavaConverters._
+
+  def apply(scalaFields: Map[String, Any]): TypedMap = {
+    new TypedMap(scalaFields.asJava)
+  }
+
+}
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
@@ -5,6 +5,7 @@ import pl.touk.nussknacker.engine.api.dict.DictInstance
 
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
+import scala.collection.JavaConverters._
 
 object typing {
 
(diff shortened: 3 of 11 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/MapLikePropertyTypeDeterminer.scala
```diff
@@ -0,0 +1,28 @@
+package pl.touk.nussknacker.engine.api.typed
+
+import java.util
+
+import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypedClass, TypingResult, Unknown}
+
+object MapLikePropertyTypeDeterminer {
```
mproch, line 7: I wonder, maybe it should be in spel package? It's currently used only in Typer as far as I see, and it's behaviour is heavily dependent on property accessors in spel?
We would be able to use Avro classes then?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/MapLikePropertyTypeDeterminer.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def mapLikePropertyType(typ: TypedClass): Option[TypingResult] = typ match {
```
mproch, line 9: can you add some doc what this method should return? So that when we add another case like avro we know what to do? :)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypedMap.scala
```diff
@@ -1,3 +1,63 @@
 package pl.touk.nussknacker.engine.api.typed
 
-case class TypedMap(fields: Map[String, Any])
\ No newline at end of file
+import java.{util => ju}
+
+case class TypedMap(fields: Map[String, Any]) extends ju.AbstractMap[String, Any] { self =>
```
mproch, line 5: I wonder - maybe we should use some map directly? Or is it for backwards compatibility?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -152,7 +152,7 @@ object typing {
           val fieldTypes = fields.map {
             case (k, v) => k -> fromInstance(v)
           }
-          TypedObjectTypingResult(fieldTypes, TypedClass(classOf[TypedMap], Nil))
+          TypedObjectTypingResult(fieldTypes)
```
mproch, line 155: Good catch :+1:

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/MapLikePropertyTypeDeterminer.scala
arkadius: Good point - moved to typer, it shouldn't cause problems with Avro classes

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/MapLikePropertyTypeDeterminer.scala
arkadius: Added

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypedMap.scala
arkadius: Indeed it will be less magic if we keep java Map for fields. Also I think that will be better if we will have equals working correctly with other maps - because of this I've replaced case class with plain serializable class.

I've added apply method for backward compatibility, and written some notice in MigrationGuide in case if someone use pattern matching on TypedMap

lciolecki:
@arkadius please update also Changelog ;)

docs/MigrationGuide.md
```diff
@@ -49,6 +49,11 @@ val source = new KafkaAvroSourceFactory(schemaRegistryProvider, processObjectDep
 val avroFixedSourceFactory = FixedKafkaAvroSourceFactory[GenericData.Record](processObjectDependencies)
 ```
 
+## In version 0.1.2 (not released yet)
```
mproch, line 52: Do we want it in 0.1.2? Maybe in 0.2.0? It's not so badly needed, and it's not fully backwards compatible...

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/sinks.scala
```diff
@@ -16,14 +16,10 @@ object sinks {
...
 
   case class GenericJsonSerialization(topic: String) extends SimpleSerializationSchema[Any](topic, element => {
-      val objToEncode = element match {
-        // TODO: would be safer if will be added expected type in Sink and during expression evaluation,
-        // would be performed conversion to it
-        case TypedMap(fields) => fields
-        case other => other
-      }
-      encoder.encode(objToEncode).spaces2
+    // TODO: would be safer if will be added expected type in Sink and during expression evaluation,
```
mproch, line 19: Is this comment still relevant?

docs/MigrationGuide.md
```diff
@@ -49,6 +49,11 @@ val source = new KafkaAvroSourceFactory(schemaRegistryProvider, processObjectDep
...
+
+* [#954](https://github.com/TouK/nussknacker/pull/954) `TypedMap` is not a case class wrapping scala Map anymore. If you have
+ done some pattern matching on it, you should use `case typedMap: TypedMap => typedMap.typedMap.asScala` instead.
```
mproch, line 55: typedMap.typedMap.asScala => typedMap.asScala?

docs/MigrationGuide.md
arkadius: IMO it is quite badly needed, because it is fix for two bugs:
- we suggest methods that can't be used
- we accept something in expressions and it blow up in runtime.
Compatibility is broken only on very advanced usages, when you extract scala map from `TypedMap` (i can add such extractor BTW but IMO it will mess a code a bit)

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/sinks.scala
arkadius: I think so - this sink still accept anything, nothing changed in this place

PR #955, opened 2020-05-13
[FIX] Empty inline lists have proper type

mproch:
I think we should treat Typed.empty like Nothing type - in particular, it's subclass of all classes :)


lciolecki:
@mproch please add some docs info about it ;)

PR #958, opened 2020-05-13
Improvement: avro schema enhancement

lciolecki:
- Changes at SchemaRegistryClient. Provided: getLatestFreshSchema and getFreshSchema methods
- Some enhancement at getting / parsing avro schema

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -18,7 +18,7 @@ Nussknacker versions
 * [#949](https://github.com/TouK/nussknacker/pull/949) JVM options can be configured via JDK_JAVA_OPTIONS env variable (in docker and standalone distribution) 
 * [#954](https://github.com/TouK/nussknacker/pull/954) Correct handling of types in empty inline lists 
 * [#944](https://github.com/TouK/nussknacker/pull/903) Pluggable SchemaRegistry Cache 
-* [#958](https://github.com/TouK/nussknacker/pull/958) Speedup getting avro schema
+* [#958](https://github.com/TouK/nussknacker/pull/958) Avro cache improvements
 
 0.1.1
 ------------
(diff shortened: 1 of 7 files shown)


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/AvroUtils.scala
```diff
@@ -29,12 +29,12 @@ class AvroUtils(schemaRegistryProvider: SchemaRegistryProvider[_]) extends Seria
 
   def latestKeySchema(topic: String): Schema =
     handleClientResponse {
-      schemaRegistryClient.getLatestSchema(AvroUtils.keySubject(topic))
+      schemaRegistryClient.getLatestFreshSchema(AvroUtils.keySubject(topic))
```
arkadius, line 32: This shouldn't be used in utils - they are invoked in engine and it is the only place where you should use cached version. Instead of this it should be used in `SchemaRegistryKafkaAvroProvider.typeDefinition`

docs/Changelog.md
```diff
@@ -18,6 +18,7 @@ Nussknacker versions
 * [#949](https://github.com/TouK/nussknacker/pull/949) JVM options can be configured via JDK_JAVA_OPTIONS env variable (in docker and standalone distribution) 
 * [#954](https://github.com/TouK/nussknacker/pull/954) Correct handling of types in empty inline lists 
 * [#944](https://github.com/TouK/nussknacker/pull/903) Pluggable SchemaRegistry Cache 
+* [#958](https://github.com/TouK/nussknacker/pull/958) Speedup getting avro schema
```
arkadius, line 21: It is not exactly "speedup", because without caching it was already fast :) Also it is very minor change, let's keep this list concise

lciolecki: Hmm, I think we should add all changes in changelog. Moreover I've added there some improvements to cache. So I can agree with you that this is not speed up, but still it's some good improvement ;) I will change description to: "Improvements Avro Cache"

PR #962, opened 2020-05-14
[Feature] Alternative method of computing counts

mproch:
Currently counts are measured as value_at_range_end - value_at_range_start. It does not work well when there are restarts which reset metric values. This change introduces alternative way (not enabled by default) which computes sum of non_negative_difference through range (more or less like computing integral). 

It can be much slower for long ranges and is not accurate during restarts, so third config option enables this mode only when there are restart in the range


Crack:
I will put it here, decide whether you want to include it in this change: we can make detection of restarts a bit faster by changing it from
```scala
      s"""SELECT derivative(value) FROM "source.count" WHERE
         |"process" = '$processName' AND env = '$env'
         | AND time >= ${from}s and time < ${to}s GROUP BY slot""".stripMargin
```
to
```scala
      s"""SELECT min(derivative) FROM (
         |  SELECT derivative(value) FROM "source.count" WHERE
         |  "process" = '$processName' AND env = '$env'
         |   AND time >= ${from}s and time < ${to}s GROUP BY slot)
         |)""".stripMargin
```
We reject partial results anyway, and are looking for negative value of derivative.

Unless we want to introduce a list of valid time ranges sometime in the future, then current method query could be used to find them.

Crack:
Another idea: add `run_id` tag to source count metric, which would be some UUID generated when metric is initialized in source task (or maybe just a compressed millisecond timestamp?). It would be similar to Flink's `job_id` it but would get regenerated also for restarts. This would enable us to query for continuity periods and min/max timestamps of each period.

We shouldn't include this UUID in node counts, or we'd risk hitting InfluxDB series cardinality limits.

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/ProcessBaseCounts.scala
```diff
@@ -0,0 +1,23 @@
+package pl.touk.nussknacker.processCounts.influxdb
+
+case class ProcessBaseCounts(nodes: Map[String, Long]) {
+
+  //node ids in influx are different than original ones, i.e influx converts spaces and dots to dashes '-'
```
Crack, line 5: InfluxDB allows for dots in names, it must be our code (or some library we use). InfluxDB itself happily accepts and returns dots (just tested in 1.8.3).

Quick test:

Session 1: `docker run --rm -itP --name influx influxdb:1.8.3`
Session 2: `docker exec -it influx influx -precision rfc3339`
```
> create database test;
> use test
Using database test
> insert test,a=a.b f=1 0
> select * from test
name: test
time                 a   f
----                 -   -
1970-01-01T00:00:00Z a.b 1
```

Session 3: `docker exec -it influx curl -XPOST -d 'q=select * from test' 'localhost:8086/query?db=test'`
```json
{"results":[{"statement_id":0,"series":[{"name":"test","columns":["time","a","f"],"values":[["1970-01-01T00:00:00Z","a.b",1]]}]}]}
```

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/InfluxCountsReporter.scala
```diff
@@ -3,47 +3,58 @@ package pl.touk.nussknacker.processCounts.influxdb
...
-  override def close(): Unit = influxBaseReporter.influxGenerator.close()
-
 }
 
 class InfluxCountsReporterCreator extends CountsReporterCreator {
 
   import net.ceedubs.ficus.Ficus._
   import net.ceedubs.ficus.readers.ArbitraryTypeReader._
-
+  import net.ceedubs.ficus.readers.EnumerationReader._
```
Crack, line 57: Is this needed here or is it a leftover after earlier experiments?

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/SimpleInfluxClient.scala
```diff
@@ -70,7 +68,7 @@ object InfluxSerie {
 
 }
 
-case class InfluxSerie(name: String, tags: Map[String, String], columns: List[String], values: List[List[Any]] = Nil) {
+case class InfluxSerie(name: String, tags: Option[Map[String, String]], columns: List[String], values: List[List[Any]] = Nil) {
```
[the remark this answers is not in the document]
mproch: :+1:

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/InfluxCountsReporter.scala
mproch: It is needed, QueryMode is enumeration

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/ProcessBaseCounts.scala
```diff
@@ -0,0 +1,23 @@
...
+  //that's wy we need reverse transformation
```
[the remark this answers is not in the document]
mproch: :+1:

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/InfluxConfig.scala
```diff
@@ -0,0 +1,19 @@
+package pl.touk.nussknacker.processCounts.influxdb
+
+case class InfluxConfig(influxUrl: String, user: String, password: String,
+                        database: String = "esp",
```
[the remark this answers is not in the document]
mproch: :+1:

engine/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/ProcessBaseCounts.scala
mproch: It think this was the case when we used to send metrics via graphite protocol.... I added comment to remove it when we remove useLegacyMetrics flag

piotrp: Oh, that explains it. I totally forgot about that abomination (despite the fact that we still use it)...

piotrp (approved):

PR #963, opened 2020-05-15
Improvements: Kafka tools refactor

lciolecki:
- Rename KafkaEspUtils to KafkaUtils
- Rename KafkaUtils from kafka tests to KafkaZookeeperUtils
- Extract BaseKafkaSinkFactory 
- Add some helper methods like: KafkaConfig.parseProcessObjectDependencies and KafkaUtils.prepareTopic

### DIFF
--- a/engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
+++ b/engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.engine.kafka
 
 import com.typesafe.config.Config
+import pl.touk.nussknacker.engine.api.process.ProcessObjectDependencies
 
 case class KafkaConfig(kafkaAddress: String,
                        kafkaProperties: Option[Map[String, String]],
@@ -22,4 +23,6 @@ object KafkaConfig {
     config.as[KafkaConfig](path)
   }
 
+  def parseProcessObjectDependencies(processObjectDependencies: ProcessObjectDependencies): KafkaConfig =
+    parseConfig(processObjectDependencies.config, "kafka")
 }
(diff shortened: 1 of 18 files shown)


engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaUtils.scala
```diff
@@ -17,17 +19,24 @@ import scala.concurrent.duration.Duration
...
 
   val defaultTimeoutMillis = 10000
 
+  private val kafkaTopicUsageKey = new NamingContext(KafkaUsageKey)
+
   def setClientId(props: Properties, id: String): Unit = {
     props.setProperty("client.id", sanitizeClientId(id))
   }
 
+  def prepareTopic(topic :String, processObjectDependencies: ProcessObjectDependencies): String =
```
mproch, line 35: prepareTopicName?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -22,4 +23,6 @@ object KafkaConfig {
     config.as[KafkaConfig](path)
   }
 
+  def parseProcessObjectDependencies(processObjectDependencies: ProcessObjectDependencies): KafkaConfig =
```
mproch, line 26: why we fix "kafka" here?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaUtils.scala
lciolecki: Right

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
lciolecki: Because it's default kafka namespace from processObjectDependencies.config

PR #969, opened 2020-05-18
[POC] Dynamic parameters

mproch:
This is NOT meant to be merged - more to discuss and try out some ideas


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -0,0 +1,62 @@
...
+/*
+  The idea is this would be low-level API, with more accessible DSLs/builders - also with annotation based configuration
+  I'd like to handle only this trait (or sth similar) in compiler - instead of MethodToInvoke, WithExplicitMethodToInvoke etc
+  Note: this trait is used without @MethodToInvoke. Usages with @MethodToInvoke can be implemented using GenericNodeTransformation
+  by NK itself (using more or less current DefinitionExtractor logic)
+ */
+object GenericNodeTransformation {
+
+  //TODO: what if we cannot determine parameters/context? With some "fatal validation error"?
+  type NodeTransformationDefinition = PartialFunction[List[(String, DefinedParameter)], TransformationStepResult]
```
arkadius, line 18: WDYT about adding some State monad to it? I would like to add some extra information between steps that will be used in each step computation. Consider situation: you have initial context `a: { b: Int, c: Int }, d: Int` and want to have some extractions on each step that will cause context stripping - e.g. you want to put `b: Int, c: Int` to `a` param additional variables and only `d: Int` to `rest` param.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -0,0 +1,62 @@
...
+  type InputContext = ValidationContext
+}
+
+trait BranchGenericNodeTransformation[T] extends GenericNodeTransformation[T] {
+  type InputContext = Map[String, ValidationContext]
+}
+
+
+//TODO: this is temporary solution, won't be able to handle adding parameters etc.
+trait WithExplicitMethodToInvokeTransformation extends WithExplicitMethodToInvoke {
```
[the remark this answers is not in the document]
mproch: As in comment - this is very temporary, I'll remove it before merge with sth better.
For now, I haven't touched ProcessCompiler/MethodExtractor to focus on the API and this is why this class exists...
I'd like to deprecate and remove WithExplicitMethodToInvoke after a while

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -0,0 +1,62 @@
...
+  self: SingleInputGenericNodeTransformation[_] =>
+
+  //TODO: we'd like to have proper type also here!
+  override def parameterDefinition: List[Parameter] = initialParameters
+
+  override def returnType: typing.TypingResult = Unknown
```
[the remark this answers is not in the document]
mproch: as in comment above - I want to replace it with sth better before merge ;)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def runtimeClass: Class[_] = classOf[Any]
```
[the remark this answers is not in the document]
mproch: yes, I also thought about it, but didn't want to make this change longer than absolutely necessary

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
mproch: I have a bit mixed feelings... I think it would be easier to ignore (if you don't need it) if you make it so:
PartialFunction[(List[(String, DefinedParameter), State)], TransformationStepResult]
but I wonder what State would be? I mean - we can make it some generic Map[String, Any] of course...

arkadius: `PartialFunction[(List[(String, DefinedParameter), State)], TransformationStepResult]` will be ok, but we will need to have some zero element for it then. For now I have some strange examples of `State` in mine use case. It was kind of `Optiona[ValidatationContext]` that at the beginning was `empty`, after initial step was `Some(fullListOfEnrichedAttributes)` and after extracting it to each parameter's it was reduced again to `Some(emptyListOfAttributes)`. I can imagine that it can be very different based on each case. I would totally prefer path dependent types way. For "normal users" we have high level API. Also someone can ignore state. and everything will looks pretty easy still IMO.

mproch: Option[_] seems like good zero :) I added state with dependent types and it looks pretty neat I think

mproch:
This spike will be continued in futher PRs

PR #971, opened 2020-05-18
Improvements: Pluggable Avro Fixed

lciolecki:
- More pluggable fixed avro provider


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/fixed/FixedKafkaAvroSchemaProvider.scala
```diff
@@ -22,18 +23,11 @@ import pl.touk.nussknacker.engine.kafka.{KafkaConfig, RecordFormatter}
   */
 class FixedKafkaAvroSchemaProvider[T: TypeInformation](val topic: String,
```
mproch, line 24: I wonder if we should move this class to .confluent package?

PR #972, opened 2020-05-18
set node layoutData after adding

JulianWielga:


lciolecki:
Can You add info with issue this PR resolves?

PR #974, opened 2020-05-20
Split Graph.js and EspNode.js to smaller parts with some TS

JulianWielga:


ui/client/components/graph/EspNode/getIconHref.ts
```diff
@@ -0,0 +1,9 @@
+import ProcessUtils from "../../../common/ProcessUtils"
```
lciolecki, line 1: Maybe it should be named as icons.ts?

ui/client/components/graph/ProcessGraph.tsx
```diff
@@ -0,0 +1,54 @@
+import {DropTarget} from "react-dnd"
```
lciolecki, line 1: Please add only one `/* eslint-disable i18next/no-literal-string */`

ui/client/components/modals/GenericModalDialog.tsx
```diff
@@ -29,47 +40,46 @@ class GenericModalDialog extends React.Component {
...
-      .then(this.closeDialog, this.removePending)
+    this.props.confirm(this.closeDialog).then(
+      this.closeDialog,
+      this.removePending,
+    )
   }
 
   renderOkBtn = () => {
-    return(
+    const STYLE = "zoom-in"
```
lciolecki, line 50: Please declare STYLE as const in GenericModalDialog.

ui/client/components/graph/EspNode/getIconHref.ts
JulianWielga: I prefer to name modules by main exported member. as long as module contains only one function I'd leave this name. "icons" says nothing about content.

ui/client/components/graph/ProcessGraph.tsx
JulianWielga: it's not right way. ignores are temporary and should attract attention. every of 3 strings should be fixed (removed) in near future

ui/client/components/modals/GenericModalDialog.tsx
JulianWielga: gives nothing positive but separates from real one time usage. will make separating button from this component harder in future.

ui/client/components/graph/EspNode/getIconHref.ts
lciolecki: `rect.ts` has also only one function.. So I would like to ask you for do it in one standard.

ui/client/components/modals/GenericModalDialog.tsx
lciolecki: Bu now it's create each time when you run renderOkBtn. When it will be as a constant class then will be declard once time.

JulianWielga: ...and it makes no difference, but makes readability worst

lciolecki: Only in your opinion?

PR #975, opened 2020-05-20
Feature: kafka avro sink

lciolecki:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/AvroUtils.scala
```diff
@@ -53,4 +52,7 @@ object AvroUtils {
 
   def parseSchema(avroSchema: String): Schema =
     parser.parse(avroSchema)
+
+  def createRecord(schema: Schema, data: Map[String, Any]): GenericData.Record =
```
arkadius, line 56: `record` method already exists

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactorySpec.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  test("should throw exception when output doesn't match to schema") {
+    assertThrowsWithParent[InvalidSinkOutput] {
+      val output = createOutput(FullNameV1.schema, FullNameV1.exampleData)
+      createSink(fullnameTopic, 2, output)
+    }
+  }
+
+  test("should not allow to create sink for #input v1 and #output v2") {
```
arkadius, line 60: It is not important that version where changed but that compatibility is broken. Rename tests

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/schema/FullName.scala
```diff
@@ -0,0 +1,79 @@
...
+
+object FullNameV2 extends TestSchema {
+  val stringSchema: String =
+    """{
+      |  "type": "record",
+      |  "namespace": "pl.touk.nussknacker.engine.avro.schema",
+      |  "name": "FullNameV2",
+      |  "fields": [
+      |    { "name": "first", "type": "string" },
+      |    { "name": "middle", "type": ["null", "string"], "default": null },
```
arkadius, line 72: It looks like compatible change. Why tests checks that is invalid to read V1 and save V2?

PR #986, opened 2020-05-28
Add schema registry to docker-compose-env file in demo

mpac123:


demo/docker/docker-compose-env.yml
```diff
@@ -60,6 +60,15 @@ services:
 #      - zookeeper
 #      - kafka
 
+  schema-registry:
```
lciolecki, line 63: Please do: schemaregistry

demo/docker/docker-compose-env.yml
```diff
@@ -60,6 +60,15 @@ services:
...
+    container_name: nussknacker_schema-registry
```
lciolecki, line 64: nussknacker_schemaregistry

demo/docker/docker-compose-env.yml
mpac123: done!

demo/docker/docker-compose-env.yml
mpac123: and also done :)

lciolecki:
And one more.. :) base-application.conf:
```
schemaRegistryUrl: "http://schema-registry:8081"

We should figure how to use SR from IDEA and flink containter.. I think, we should have another url  for Nussknacker GUI.

mpac123:
> And one more.. :) base-application.conf:
> 
> ```
> schemaRegistryUrl: "http://schema-registry:8081"
> ```
> 
> It's time when we should figure how to use SR from local nk and flink containter.. - not it doesnt' work.. I think, we should have another url for Nussknacker GUI.

By 'it's time' do you mean this PR? Because I'd say it's the thing to be fixed together with broken avro sources - wat do you think? I mean, for now I won't be even able to check if whatever solution I come up with works

Issue #959, opened 2020-05-14
Cannot copy any text from panels


alien11689:
**Describe the bug**
Cannot copy any text from tips or comment panel

**To Reproduce**
Try to right click on comment

**Environment (please complete the following information):**
 - OS: CentOS
 - Nussknacker version: 0.1.1
 - Flink version: Not important
 - Java version: 8
 - Scala version: 2.11
 - Method of ran of components: Not important
 - Browser type and version: chrome  81.0.4044.92

PR #993, opened 2020-05-29
Bugfix #959 & #832

JulianWielga:
Bugfix for #959 & #832

### DIFF
--- a/ui/client/components/graph/focusable.tsx
+++ b/ui/client/components/graph/focusable.tsx
@@ -1,0 +1,18 @@
+import React, {forwardRef, useCallback, MouseEventHandler} from "react"
+
+// eslint-disable-next-line react/display-name
+export const FocusableDiv = forwardRef<HTMLDivElement>((
+  {onClick, ...props}: React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
+  ref,
+) => {
+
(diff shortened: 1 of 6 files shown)


ui/client/containers/Visualization.js
```diff
@@ -58,7 +58,12 @@ class Visualization extends React.Component {
 
   bindPasteShortcut() {
     return (event) => this.props.actions.pasteSelection(
-      () => this.pasteSelection(event),
+      () => {
+        if (["INPUT", "TEXTAREA"].includes(event.target?.tagName)) {
```
mproch, line 62: does this mean if I copy nodes and "paste" them to textarea I'll get json?

ui/client/components/graph/focusable.tsx
```diff
@@ -0,0 +1,18 @@
+import React, {forwardRef, useCallback, MouseEventHandler} from "react"
```
mproch, line 1: What is it for? To be able to select sth on graph pane?

ui/client/containers/Visualization.js
JulianWielga: yes. if user has focus on input field then he wants to fill this input. pasting clipboard content somewhere else is confusing

ui/client/components/graph/focusable.tsx
JulianWielga: to grab focus e.g. from comment input when clicked on graph

PR #988, opened 2020-05-28
[Feature] json editor

Abrams27:
closes https://github.com/TouK/nussknacker/issues/894

Screenshot from 2020-05-28 15-37-11

PR #995, opened 2020-06-02
[Improvement] json validator

Abrams27:
json editor


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -132,6 +133,23 @@ case class MaximalNumberValidator(maximalNumber: BigDecimal) extends ParameterVa
   )
 }
 
+case object JsonValidator extends ParameterValidator {
+
+  //Blank value should be not validate - we want to chain validators
+  override def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] =
+    if (StringUtils.isBlank(value) || parse(value).isRight)
```
arkadius, line 140: 1. It can be tricky, but you won't get here `{"some": "json"}` but you will get `'{"some": "json"}'` (with `'`), because we transfer all values using spring expression language, especially using: https://docs.spring.io/spring/docs/4.3.10.RELEASE/spring-framework-reference/html/expressions.html#expressions-ref-literal
It is why I've mentioned that you need to do end-to-end tests that whole solution is working correctly. Maybe you can add some example class with json parameter to `SimpleTypesService`?
2. Can you return validation message in error?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/JsonValidatorExtractor.scala
```diff
@@ -0,0 +1,14 @@
+package pl.touk.nussknacker.engine.definition.validator
+
+import pl.touk.nussknacker.engine.api.definition.{JsonValidator, ParameterValidator}
+
+object JsonValidatorExtractor extends ValidatorExtractor {
+
+  override def extract(params: ValidatorExtractorParameters): Option[ParameterValidator] = {
+    params.extractedEditor match {
+      // TODO add json editor case
```
arkadius, line 9: Why TODO? Also I think that you only need to add new case to `AdditionalPropertyEditorValidatorDeterminer` and it should work.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
Abrams27: done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/JsonValidatorExtractor.scala
Abrams27: I tried to add new case only to `AdditionalPropertyEditorValidatorDeterminer` but it didnt work. `FixedValuesParameterEditor` has case in `AdditionalPropertyEditorValidatorDeterminer` but also in its own extractor, so I did the same.

arkadius: So why this class even exists? Why you've added extractor to the chain that extract nothing and has TODO inside? What is the purpose?

lciolecki: @arkadius is right. Please at first merge it: https://github.com/TouK/nussknacker/pull/988. After that uncomment validator extractor.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/validator/JsonValidatorExtractor.scala
```diff
@@ -1,13 +1,12 @@
...
-import pl.touk.nussknacker.engine.api.definition.{JsonValidator, ParameterValidator}
+import pl.touk.nussknacker.engine.api.definition.{JsonParameterEditor, JsonValidator, ParameterValidator}
 
 object JsonValidatorExtractor extends ValidatorExtractor {
 
   override def extract(params: ValidatorExtractorParameters): Option[ParameterValidator] = {
     params.extractedEditor match {
-      // TODO add json editor case
-      // case _ => Some(JsonValidator)
+      case Some(JsonParameterEditor) => Some(JsonValidator)
```
arkadius, line 9: IMO would be better to move this case to `FixedValueValidatorExtractor` and rename this class to `EditorBasedValidatorExtractor`.  Then we will have two looking almost identical classes: `AdditionalPropertyEditorValidatorDeterminer` and `EditorBasedValidatorExtractor`, and it will be easier to find this duplication of code. WDYT?

Abrams27: done

PR #996, opened 2020-06-03
[Feature] First version of GenericNodeTransformation

mproch:
This is first version of GenericNodeTransformation API, which aims to be basic low-level API for defining custom nodes. This PR is limited in scope:
- only single input supported
- limited support in UI
- API should be considered experimental
- only custom nodes supported (not sources/sinks)
- quite a lot of internal refactoring is needed. This PR does only what is absolutely necessary


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -0,0 +1,51 @@
...
+  //ValidationContext for single input, Map[String, ValidationContext] for joins
+  type InputContext
+
+  //State is arbitrary data that can be passed between steps of NodeTransformationDefinition
+  type State
+
+  //TODO: what if we cannot determine parameters/context? With some "fatal validation error"?
+  type NodeTransformationDefinition = PartialFunction[TransformationStep, TransformationStepResult]
+
+  def contextTransformation(context: InputContext, dependencies: List[NodeDependencyValue])(implicit nodeId: NodeId): NodeTransformationDefinition
```
arkadius, line 28: WDYT about adding `Future` to returned type?  We have in a quite near future example when we would like to call external service to determine parameter's definitions.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/evaluatedparam.scala
```diff
@@ -1,14 +1,26 @@
...
 import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 
 object evaluatedparam {
 
   case class TypedParameter(name: String, typedValue: TypedValue)
 
-  case class Parameter(name: String, expression: Expression, returnType: TypingResult,
-                       shouldBeWrappedWithScalaOption: Boolean, shouldBeWrappedWithJavaOptional: Boolean,
-                       typingInfo: ExpressionTypingInfo)
+  case class Parameter(typedExpression: TypedExpression, parameterDefinition: definition.Parameter) {
```
arkadius, line 11: Won't will cause some serialization problems? WDYT about replacing it with `apply()` method instead?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/DefinitionExtractor.scala
```diff
@@ -67,13 +74,67 @@ object DefinitionExtractor {
...
+
+  case class GenericNodeTransformationMethodDef(obj: GenericNodeTransformation[_], objectDefinition: ObjectDefinition) extends ObjectWithMethodDef {
+
+    override def invokeMethod(params: Map[String, Any], outputVariableNameOpt: Option[String], additional: Seq[AnyRef]): Any = {
+      val additionalParams = obj.nodeDependencies.map {
+        case TypedNodeDependency(klazz) =>
+          additional.find(klazz.isInstance).map(TypedNodeDependencyValue)
+            .getOrElse(throw new IllegalArgumentException(s"Failed to find dependency: $klazz"))
+        case OutputVariableNameDependency => outputVariableNameOpt.map(OutputVariableNameValue).getOrElse(throw new IllegalArgumentException("Output variable not defined"))
+        case _ => throw new IllegalArgumentException("Cannot handle this dependency...")
```
arkadius, line 120: Can you add "this" `other` in the message?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/evaluatedparam.scala
mproch: :+1: I added apply instead

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/DefinitionExtractor.scala
mproch: :+1: added

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
mproch: I thought about it, but I don't think it's a good idea. Because we'd either have to:
a) put Await somewhere nearby
b) pull Future along all compilation process

b) would be very difficult and a) wouldn't bring much benefit. I think compilation process is rather "rare" and can be considered synchronous. The schema registry case also shows that probably caching will also be involved anyway.

PR #998, opened 2020-06-04
Feature: Confluent Avro Serializer

lciolecki:
- Add confluent avro serializer


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/AbstractConfluentKafkaAvroSerializer.scala
```diff
@@ -0,0 +1,95 @@
...
+        val out = new ByteArrayOutputStream
+        out.write(AbstractKafkaAvroSerDe.MAGIC_BYTE)
+        out.write(ByteBuffer.allocate(AbstractKafkaAvroSerDe.idSize).putInt(schemaId).array)
+
+        data match {
+          case array: Array[Byte] => out.write(array)
+          case _ =>
+            val encoder = this.encoderFactory.directBinaryEncoder(out, null)
+
+            val value = data match {
```
arkadius, line 43: Please extract whole logic of schema evolution to some class like `AvroWriterSchemaEvolutionHandler` or sth similar, with one method `alignRecordToSchema(record, schema)`. We will be able to change implementation and compare if each works the same way.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/AbstractConfluentKafkaAvroSerializer.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+    * Convert serialization record to final schema.
+    * We try to set schema default value when record doesn't support field from schema.
+    *
+    * @param record
+    * @param schema
+    * @return
+    */
+  private def convertRecordToSchema(record: GenericRecord, schema: Schema): GenericData.Record = {
```
arkadius, line 78: IMO we should start with implementation that will do serialization -> deserilization (with schema evolution) -> serialization. After that we can do some tweaks if it won't be performed enough. If users will find some bugs, they will be able to switch to correct (slower) implementation. WDYT?

build.sbt
```diff
@@ -145,7 +145,7 @@ val forkSettings = Seq(
 
 val akkaV = "2.5.21" //same version as in Flink
 val flinkV = "1.9.1"
-val avroV = "1.8.2" // should be the same as in flink-avro
+val avroV = "1.9.1" // should be the same as in flink-avro
```
mproch, line 148: what about this comment?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/AbstractConfluentKafkaAvroSerializer.scala
mproch: :+1: to @arkadius remark, also - don't we need some recursion here if we want it to be correct? 
I would even suggest we leave handling this case (this whole default value business) for the future, and just and remark

build.sbt
lciolecki: `"org.apache.flink" % "flink-avro" % flinkV,` - right now is the same..

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
```diff
@@ -0,0 +1,106 @@
...
+        buffer.get(bytes, 0, length)
+        bytes
+      } else {
+        val start = buffer.position + buffer.arrayOffset
+        val binaryDecoder = decoderFactory.binaryDecoder(buffer.array, start, length, null)
+        val result = reader.read(record, binaryDecoder)
+        if (writerSchema.getType == Schema.Type.STRING) result.toString else result
+      }
+    } catch {
+      case exc@(_: RuntimeException | _: IOException) =>
```
arkadius, line 69: Why not to handle Exception instead?

build.sbt
lciolecki: You're right. A will rollback this.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
lciolecki: It was copied from confluent AbstractKafkaAvroSerializer ;) They catch only these two exceptions..

Issue #997, opened 2020-06-03
Improve Literal Validation process


lciolecki:
We have some problem with mandatory / not blank + else literal validators, because we have to skipp checking empty string in other validators.

Proposition of resolve this problem is provide two strategies of validation:

- require: mandatory / not empty validation 
- other validations

At first we should check required validation, if there is everything okay then we should run else validations.

We should do it on BE and FE.

PR #1000, opened 2020-06-08
Improve literal validation process with priorities

mpac123:
Concerns #997

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
@@ -14,15 +14,15 @@ case object OutputVariableNameDependency extends NodeDependency
 
 object Parameter {
 
-  def apply[T: ClassTag](name: String): Parameter = Parameter(name, Typed[T])
+  def apply[T: ClassTag](name: String, isRequiredParameter: Boolean = true): Parameter = Parameter(name, Typed[T], isRequiredParameter)
 
   // we want to have mandatory parameters by default because it can protect us from NPE in some cases)
-  def apply(name: String, typ: TypingResult): Parameter =
-    Parameter(name, typ, validators = List(MandatoryParameterValidator()))
+  def apply(name: String, typ: TypingResult, isRequiredParameter: Boolean): Parameter =
+    Parameter(name, typ, validators = List(MandatoryParameterValidator()), isRequiredParameter)
 
-  def apply(name: String, typ: TypingResult, validators: List[ParameterValidator]): Parameter =
+  def apply(name: String, typ: TypingResult, validators: List[ParameterValidator], isRequiredParameter: Boolean): Parameter =
     Parameter(name, typ, editor = None, validators = validators, additionalVariables = Map.empty,
-      branchParam = false, isLazyParameter = false, scalaOptionParameter = false, javaOptionalParameter = false)
+      branchParam = false, isLazyParameter = false, isRequiredParameter = isRequiredParameter, scalaOptionParameter = false, javaOptionalParameter = false)
 
   @deprecated("Passing runtimeClass to Parameter.apply is deprecated in favor of passing isLazyParameter", "0.1.0")
   def apply(name: String,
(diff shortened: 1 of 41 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -23,15 +23,25 @@ import pl.touk.nussknacker.engine.api.CirceUtil._
   */
 @ConfiguredJsonCodec sealed trait ParameterValidator {
 
-  def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit]
+  val priority: Long = 0
+  protected val skipIfBlank: Boolean
```
lciolecki, line 27: In most cases it will be `true`. So mayby we can set it true? :) And mayby better name will be skipIfNotRequired?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -23,15 +23,25 @@ import pl.touk.nussknacker.engine.api.CirceUtil._
...
 
+  final def isValid(paramName: String, value: String, label: Option[String])(implicit nodeId: NodeId): Validated[PartSubGraphCompilationError, Unit] = {
```
lciolecki, line 29: I think isValid should stay as before.. We can add validate method where we will check this one if..  But this is only my opinion ;)

lciolecki:
Good job! 

Some small improvements needed. What about change priority validators by they annotation?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
mpac123: But then everyone would have to remember to use this validate method for consistency, right? I don't know either, I was thinking this way would be clearer, but I may be wrong

mpac123: Ah, you meant to check it somewhere "higher"... Yeap, we could do it, but then we'd also have to rewrite tests testing the validators logic. I think it's better if the logic stays in one place.

lciolecki: Search Results
Translation result
English
Polish
Let's look for this case:
I want test one validator, for example Max Value. When I put there empty string then a I will get true.. Is it really true? I don't thinks so ;)

Each validator should reuturn good result alsow for empty string value.. 

Now we have the same trick with checking emptyString but we moved it to some abstract place.. 

WDYF?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
lciolecki: Ok, maybe you're right ;) But I think in most cases this skipping will be true.. @arkadius WDYT?

gadomsky: and WDYT about 
```
//pseudo code
trait NotBlanParameterValidator extends ParameterValidator {
 def validateWhenNotBlank(value) 
 final def validate(value) = if(value.isBlank) Valid else validateWhenNotBlank(value)  
}
```
and then instead of setting `skipIfBlank = true` we could extend `NotBlankParameterValidator` and implement it?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -44,14 +44,14 @@ object Parameter {
...
-      branchParam = false, isLazyParameter = false, scalaOptionParameter = false, javaOptionalParameter = false)
+      branchParam = false, isLazyParameter = false, isRequiredParameter = false, scalaOptionParameter = false, javaOptionalParameter = false)
 
 }
 
 object NotBlankParameter {
 
-  def apply(name: String, typ: TypingResult): Parameter =
-    Parameter(name, typ, validators = List(NotBlankParameterValidator))
+  def apply(name: String, typ: TypingResult, isRequiredParameter: Boolean = true): Parameter =
```
lciolecki, line 53: `NotBlankParametere` sould be always required?  I mean there shouldn't be possiblity to set this.

PR #1012, opened 2020-06-17
Improvements: Skipping nullable fields at avro typing

lciolecki:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -51,5 +53,18 @@ object AvroSchemaTypeDefinitionExtractor {
...
   }
+}
+
+object AvroSchemaTypeDefinitionExtractor {
 
+  private lazy val withoutOptionallyFieldsExtractor = new AvroSchemaTypeDefinitionExtractor(skippNullableFields = true)
+
+  private lazy val withOptionallyFieldsExtractor = new AvroSchemaTypeDefinitionExtractor(skippNullableFields = false)
+
+  def typeDefinition(schema: Schema, skippNullableFields: Boolean = false): TypingResult = {
```
mproch, line 64: let's not use default boolean parameters, they can be **really** misleading :)
Let's add method: typeDefinitionWithoutNullables to be clear?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryKafkaAvroProvider.scala
```diff
@@ -19,7 +17,7 @@ class SchemaRegistryKafkaAvroProvider[T](schemaRegistryProvider: SchemaRegistryP
     schemaRegistryProvider
```
mproch, line 17: let's add some comment that currently we're doing approximate type generation to avoid false positives in validation, but this should be improved in the future

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
lciolecki: Ok.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryKafkaAvroProvider.scala
lciolecki: Done

PR #1014, opened 2020-06-18
Feature: Avro Typing

lciolecki:
- Provided Avro Typing
- Removed Avro Helper
- Extended BestEffortAvroEncoder encode for `(Record, GenericContainer)` 
- Extended AvroSchemaTypeDefinitionExtractor typeDefinition for `Schema.Type.ENUM`
- Some refactor..

### DIFF
--- a/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
+++ b/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
@@ -8,7 +8,7 @@ import cats.data.Validated.{Invalid, Valid}
 import cats.data.{NonEmptyList, ValidatedNel}
 import cats.implicits._
 import org.apache.avro.generic.GenericData.EnumSymbol
-import org.apache.avro.generic.{GenericData, GenericRecordBuilder}
+import org.apache.avro.generic.{GenericContainer, GenericData, GenericRecordBuilder}
 import org.apache.avro.util.Utf8
 import org.apache.avro.{AvroRuntimeException, Schema}
 
@@ -23,6 +23,9 @@ object BestEffortAvroEncoder {
     (schema.getType, value) match {
       case (_, Some(nested)) =>
         encode(nested, schema)
+      //It's special situation when at sink we do map with #input as field
+      case (Schema.Type.RECORD, container: GenericContainer) =>
+        Valid(container)
       case (Schema.Type.RECORD, map: collection.Map[String@unchecked, _]) =>
         encodeRecord(map, schema)
       case (Schema.Type.RECORD, map: util.Map[String@unchecked, _]) =>
@@ -141,7 +144,7 @@ object BestEffortAvroEncoder {
     }
   }
 
-  private def error(str: String) = Invalid(NonEmptyList.of(str))
+  private def error(str: String): Invalid[NonEmptyList[String]] = Invalid(NonEmptyList.of(str))
 
   private def encodeString(str: String): Utf8 = {
     new Utf8(str)
--- a/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
+++ b/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
@@ -11,8 +11,9 @@ import org.apache.avro.generic.GenericData.EnumSymbol
 import org.apache.avro.generic.{GenericContainer, GenericData, GenericRecordBuilder}
 import org.apache.avro.util.Utf8
 import org.apache.avro.{AvroRuntimeException, Schema}
+import pl.touk.nussknacker.engine.avro.schema.{AvroSchemaEvolution, DefaultAvroSchemaEvolution}
 
-object BestEffortAvroEncoder {
+class BestEffortAvroEncoder(avroSchemaEvolution: AvroSchemaEvolution) {
 
   import scala.collection.JavaConverters._
 
@@ -119,7 +120,7 @@ object BestEffortAvroEncoder {
   }
 
   private def encodeGenericContainer(container: GenericContainer, schema: Schema): WithError[GenericContainer] = {
-    if (!container.getSchema.equals(schema)) {
+    if (!avroSchemaEvolution.canBeEvolved(container, schema)) {
       error(s"Not expected container: ${container.getSchema} for schema: $schema")
     } else {
       Valid(container)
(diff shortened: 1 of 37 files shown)


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -20,26 +20,28 @@ class AvroSchemaTypeDefinitionExtractor(skippNullableFields: Boolean) {
...
-          fields.map { field =>
-            field.name() -> typeDefinition(field.schema())
-          }.toMap,
-          Typed.typedClass[GenericRecord]
-        )
+        val fields = schema
+          .getFields
+          .asScala
+          //@TODO: Maybe Optional field should be marked when it has defaultValue?
+          .filterNot(field => skippNullableFields && field.schema().isNullable && field.hasDefaultValue)
```
mproch, line 30: why only skip when both default value and isNullable?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSink.scala
```diff
@@ -1,23 +1,36 @@
...
     dataStream
       .map(_.finalContext)
       .map(flinkNodeContext.lazyParameterHelper.lazyMapFunction(output))
-      .map(_.value)
+      .map(ctx => ctx.value match {
+          case data: java.util.Map[String, Any] => BestEffortAvroEncoder.encodeRecordOrError(data, schema)
+          case _: GenericRecord => ctx.value
+          case value => {
+            logger.error("Invalid output type error.", ctx)
+            throw InvalidSinkOutput(s"Invalid output type ${value.getClass}. Excepted java.util.Map or GenericRecord.")
```
mproch, line 28: I don't know if we should throw exception just like that here. It will cause whole job to restart - and since it's the fault of the event, it won't proceed. Maybe we should somehow use exceptionHandler (but we don't have access to it ATM?)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryKafkaAvroProvider.scala
```diff
@@ -15,10 +17,13 @@ class SchemaRegistryKafkaAvroProvider[T](schemaRegistryProvider: SchemaRegistryP
...
   //For typing we use all fields from schema (also optionally fields)
   override def typeDefinition: Validated[SchemaRegistryError, typing.TypingResult] =
+    fetchValueSchema.map(AvroSchemaTypeDefinitionExtractor.typeDefinition)
+
+  override def fetchValueSchema: Validated[SchemaRegistryError, Schema] =
     schemaRegistryProvider
       .createSchemaRegistryClient
-      .getFreshSchema(AvroUtils.valueSubject(topic), version)
-      .map(AvroSchemaTypeDefinitionExtractor.typeDefinition)
+      //There should be passed topic, we should use ConfluentUtils in confluent implementation..
```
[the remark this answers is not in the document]
lciolecki: Exactly.. It's only interface. I think we should have somethinkg like getValueFreshSchema or maybe getFreshSchema(topic: String, version, isKey: Boolean)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
lciolecki: When field is `union: [null, string]` you still should set value ;) So it's not definition of optional field..

lciolecki: For example PaymentV1.schema has field vat. When we remove setting null then avro throws exception:
```
Field vat type:UNION pos:5 not set and has no default value
Path in schema: --> vat

File: engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSink.scala (hunk 2 above)
Comment (lciolecki) [reply engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSink.scala:28]: I will remove exception and leave there log with TODO.

File: engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala [hunk 4]
```diff
@@ -23,6 +23,9 @@ object BestEffortAvroEncoder {
     (schema.getType, value) match {
       case (_, Some(nested)) =>
         encode(nested, schema)
+      //It's special situation when at sink we do map with #input as field
+      case (Schema.Type.RECORD, container: GenericContainer) =>
```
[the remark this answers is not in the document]
lciolecki: You're right. Thanks for catch that! Done.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
```diff
@@ -119,6 +118,14 @@ object BestEffortAvroEncoder {
     }
   }
 
+  private def encodeGenericContainer(container: GenericContainer, schema: Schema): WithError[GenericContainer] = {
+    if (!container.getSchema.equals(schema)) {
```
mproch, line 122: I think equals is too aggressive?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSink.scala
```diff
@@ -22,10 +22,11 @@ class KafkaAvroSink(topic: String, output: LazyParameter[Any], kafkaConfig: Kafk
       .map(flinkNodeContext.lazyParameterHelper.lazyMapFunction(output))
       .map(ctx => ctx.value match {
           case data: java.util.Map[String, Any] => BestEffortAvroEncoder.encodeRecordOrError(data, schema)
-          case _: GenericRecord => ctx.value
+          case record: GenericRecord => record
           case value => {
+            //TODO: We should better handle this situation by using EspExceptionHandler
             logger.error("Invalid output type error.", ctx)
-            throw InvalidSinkOutput(s"Invalid output type ${value.getClass}. Excepted java.util.Map or GenericRecord.")
+            value
```
mproch, line 29: I think value is not good choice as it will blow up in a moment? If we want to continue processing all we can do is log error and drop record?

PR #1016, opened 2020-06-19
Improvements: SchemaRegistryProvider

lciolecki:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
```diff
@@ -14,11 +16,17 @@ class ConfluentSchemaRegistryProvider[T: TypeInformation](val schemaRegistryClie
                                                           val deserializationSchemaFactory: KafkaVersionAwareDeserializationSchemaFactory[T],
                                                           val kafkaConfig: KafkaConfig,
                                                           val formatKey: Boolean) extends SchemaRegistryProvider[T] {
+
+  private lazy val schemaRegistryClient = schemaRegistryClientFactory.createSchemaRegistryClient(kafkaConfig)
```
mproch, line 20: It should probably be @transient if we don't want serialization problems?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
```diff
@@ -14,11 +16,17 @@ class ConfluentSchemaRegistryProvider[T: TypeInformation](val schemaRegistryClie
...
+
   def recordFormatter(topic: String): Option[RecordFormatter] =
     Some(ConfluentAvroToJsonFormatter(createSchemaRegistryClient, topic, formatKey))
 
   override def createSchemaRegistryClient: ConfluentSchemaRegistryClient =
-    schemaRegistryClientFactory.createSchemaRegistryClient(kafkaConfig)
+    schemaRegistryClient
+
+  override def getSchema(topic: String, version: Option[Int]): Validated[SchemaRegistryError, Schema] =
+    schemaRegistryClient.getSchema(ConfluentUtils.valueSubject(topic), version)
```
mproch, line 29: you change getFreshSchema to getSchema. Is it on purpose?

I also wonder if this is  good place to handle adding "-value". We expose "raw" SchemaRegistryClient as well, so it's possible now to do:

schemaRegistryProvider.schemaRegistryClientFactory.createSchemaRegistry().getSchema
and 
schemaRegistryProvider.getSchema 

and get different results - which is misleading...

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryProvider.scala
```diff
@@ -15,4 +17,13 @@ trait SchemaRegistryProvider[T] extends Serializable {
   def serializationSchemaFactory: KafkaVersionAwareSerializationSchemaFactory[Any]
 
   def recordFormatter(topic: String): Option[RecordFormatter]
+
+  /**
+    * Topic value schema
+    *
+    * @param topic
```
mproch, line 24: let's remove empty param annotations. Currently they are only noise

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
lciolecki: You're right. I've changed it again ;)

PR #1020, opened 2020-06-26
[Improvement] corrent context is passed for branch parameters on UI

mproch:
Currently suggestion engine in FE doesn't handle BranchParameters well. 
We refactor ExpressionSuggest a bit, so that list of variables is passed from outside. This allows to use other ValidationContext for branch parameters. 
It turns out we *already* pass ValidationContext for different branches for FE, so we use it, however this probably deserves refactoring so it's more explicit

### DIFF
--- a/ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
+++ b/ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
@@ -3,7 +3,7 @@ import React from "react"
 import ExpressionTestResults from "../../tests/ExpressionTestResults"
 import EditableEditor from "../EditableEditor"
 import {EditorType} from "./Editor"
-import {NodeType, UIParameter} from "../../../../../types"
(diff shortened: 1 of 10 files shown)


ui/client/components/graph/node-modal/editors/expression/ExpressionField.tsx
```diff
@@ -30,8 +31,10 @@ class ExpressionField extends React.Component<Props> {
     const {
       fieldName, fieldLabel, exprPath, isEditMode, editedNode, isMarked, showValidation, showSwitch,
       parameterDefinition, setNodeDataAt, testResultsToShow, testResultsToHide, toggleTestResult, renderFieldLabel, fieldType,
-      errors,
+      errors, variableTypes,
     } = this.props
+    console.log("name", fieldName, variableTypes)
```
arkadius, line 36: Please remove it

mproch: ups... sorry :) fixed

PR #1026, opened 2020-07-02
Configurable defaultUseAsyncInterpretation

arkadius:


engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessInterpreter.scala
```diff
@@ -55,7 +56,7 @@ object StandaloneProcessInterpreter {
       definitions,
       listeners,
       modelData.modelClassLoader.classLoader
-    ).andThen { compiledProcess =>
+    )(DefaultAsyncInterpretationValueDeterminer.DefaultValue).andThen { compiledProcess =>
```
mproch, line 59: Why is it default here?

ui/client/types/process.ts
```diff
@@ -29,4 +29,9 @@ export type ProcessDefinitionData = {
   nodesConfig?: $TodoType,
   nodesToAdd?: NodesGroup[],
   processDefinition?: $TodoType,
+  servicesDefinition?: ServicesDefinition
+}
+
+export type ServicesDefinition = {
```
mproch, line 35: Why ServicesDefinition? Currently it's indeed for services, but in the future we may apply async also for some other stuff?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionEvaluator.scala
```diff
@@ -28,7 +28,8 @@ object ExpressionEvaluator {
   def optimizedEvaluator(globalVariablesPreparer: GlobalVariablesPreparer,
                          listeners: Seq[ProcessListener],
                          metaData: MetaData,
-                         services: Map[String, ObjectWithMethodDef]): ExpressionEvaluator = {
+                         services: Map[String, ObjectWithMethodDef])
+                        (implicit defaultAsyncValue: DefaultAsyncInterpretationValue): ExpressionEvaluator = {
```
mproch, line 32: let's add TODO to remove this after getting rid of lazy values? Otherwise we'll forget about it :)

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessInterpreter.scala
arkadius: I found out that `StandaloneMetaData` has no `useAsyncInterpretation` field so I assumed that configuration is not needed and it can be passed here whatever value, because it won't be read. Am I wrong? 

Maybe we should hardcode here `true` value because standalone is always executed asynchronous? But then why it is `allowLazyVars = true` hardcoded in `StandaloneMetaData`. Shouldn't be `false`?

ui/client/types/process.ts
arkadius: On this level there are only "containers" for values so I was looking for some "container" that will be useful for other settings in the future... but indeed we can start with `defaultAsyncInterpretation: Boolean` and wrap it with container when we will have more then one similar field. WDYT?

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessInterpreter.scala
mproch: ok, I understand, it doesn't have any effect... probably it'll go away after removing lazy vals etc...

ui/client/types/process.ts
mproch: :+1:

PR #1037, opened 2020-07-06
Feature: Kafka Avro support namespace at topics

lciolecki:
* Added Kafko Avro Support for namespace at topics list // filtering by namespace prefix for example..
* Some refactor
* Add tests


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/namespaces/ObjectNaming.scala
```diff
@@ -5,6 +5,7 @@ import com.typesafe.config.Config
 trait ObjectNaming extends Serializable {
   def prepareName(originalName: String, config: Config, namingContext: NamingContext): String
   def objectNamingParameters(originalName: String, config: Config, namingContext: NamingContext): Option[ObjectNamingParameters]
+  def decodeName(preparedName: String, config: Config, namingContext: NamingContext): Option[String]
```
mpac123, line 8: Or maybe just `getOriginalName`?

PR #1039, opened 2020-07-06
ExecutionConfig preparation in a single place. 

arkadius:
+ `enableObjectReuse = false` in tests.


engine/flink/management/sample/src/test/scala/pl/touk/nussknacker/engine/process/NamespacedKafkaSourceSinkTest.scala
```diff
@@ -64,8 +64,8 @@ class NamespacedKafkaSourceSinkTest extends FunSuite with BeforeAndAfterAll with
     val config = ConfigFactory.load()
       .withValue("kafka.kafkaAddress", fromAnyRef(kafkaZookeeperServer.kafkaAddress))
       .withValue("namespace", fromAnyRef(namespaceName))
-    registrar = FlinkStreamingProcessRegistrar(new FlinkProcessCompiler(LocalModelData(config, configCreator,
-      objectNaming = new TestObjectNaming)), config)
+    val modelData = LocalModelData(config, configCreator, objectNaming = new TestObjectNaming)
+    registrar = FlinkStreamingProcessRegistrar(new FlinkProcessCompiler(modelData), config, ExecutionConfigPreparer.unOptimizedChain(modelData, None))
```
mproch, line 68: maybe we can have some helper in test-utils? This code is used quite often?

arkadius: It won't be easy. Currently we have only: kafka-test-utils, flink-test-utils and something like scalatest-utils. We don't have nk-test-utils. I suggest to back to this subject when we will have some wider perspective what should be in nk-test-utils. Maybe some wrapper for process tests without flink and kafka?

Issue #1035, opened 2020-07-06
Bump up Confluent package


lciolecki:
After bump up Avro to 1.9 - https://github.com/TouK/nussknacker/pull/1034 we can update confluent to newer version. From 5.5 confluent supports avro 1.9 version.

PR #1041, opened 2020-07-07
Improvements: Bump up confluent version

lciolecki:
Resolved https://github.com/TouK/nussknacker/issues/1035


build.sbt
```diff
@@ -78,11 +78,14 @@ lazy val publishSettings = Seq(
 )
 
 def nussknackerMergeStrategy: String => MergeStrategy = {
+  case PathList(ps@_*) if ps.last == "module-info.class" => MergeStrategy.first //after confluent bump up to 5.5
```
mproch, line 81: please, let's add TODO: figure how how to handle JDK modules
or sth like that,

build.sbt
```diff
@@ -192,7 +195,7 @@ val slickV = "3.3.2"
 val hsqldbV = "2.5.0"
 val postgresV = "42.2.12"
 val flywayV = "6.3.3"
-val confluentV = "5.4.1"
+val confluentV = "5.5.0"
```
mproch, line 198: which kafka version will be used now?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentUtils.scala
```diff
@@ -3,15 +3,19 @@ package pl.touk.nussknacker.engine.avro.schemaregistry.confluent
...
 import cats.data.Validated
+import io.confluent.kafka.schemaregistry.ParsedSchema
+import io.confluent.kafka.schemaregistry.avro.{AvroSchema, AvroSchemaProvider}
+import org.apache.avro.Schema
 import org.apache.kafka.common.errors.SerializationException
 
 object ConfluentUtils {
-  //Copied from AbstractKafkaAvroSerDe
+
+  private final val ValueSubjectPattern = "(.*)-value".r
```
Crack, line 13: Can we just use `endsWith`, to remove regexp parsing/execution?

lciolecki: It's not part of my change, but I agree with you: we can cache it. In fact we run it only at nk and once before deploy process.

build.sbt
lciolecki: Kafka version is still 2.2

lciolecki:
There is some problem with Krylo - we're waiting for @arkadius fiexes.

PR #1044, opened 2020-07-08
Flink 1.11

mproch:
Migration to 1.11 itself is pretty straightforward. The biggest problem is deprecation of current watermark mechanisms. In this PR we introduce own abstraction to easier handle Flink changes


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkProcessMain.scala
```diff
@@ -30,8 +29,8 @@ trait FlinkProcessMain[Env] extends FlinkRunner with LazyLogging {
       runProcess(env, modelData, process, processVersion, ExecutionConfigPreparer.defaultChain(modelData, buildInfo))
     } catch {
       // marker exception for graph optimalization
-      case ex: ProgramAbortException =>
-        throw ex
+      //case ex: ProgramAbortException =>
```
jedrz, line 32: Is it intended?

jedrz (approved):

arkadius (approved):

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkProcessMain.scala
mproch: Yes, in Flink 1.11 it works differently, I added not-so-nice fallback

PR #1060, opened 2020-07-10
Fixed specific record case

arkadius:
Now you don't need to provide version, also reader schema is determined based on class


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/ConfluentKafkaAvroDeserializationSchemaFactory.scala
```diff
@@ -26,28 +25,28 @@ trait ConfluentKafkaAvroDeserializerFactory extends ConfluentKafkaAvroSerializat
...
+      case _ =>
+        null
     }
 
-    deserializer.configure(kafkaConfig.kafkaProperties.getOrElse(Map.empty).asJava, isKey)
-    (deserializer.asInstanceOf[Deserializer[T]], typeInformation)
-  }
+    val deserializer = new ConfluentKafkaAvroDeserializer[T](kafkaConfig, schema, schemaRegistryClient, isKey = isKey, isSpecificRecord)
+
+    val typeInformation = schemaDeterminingStrategy match {
```
lciolecki, line 39: Maybe extract this to separeted method?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/ConfluentKafkaAvroDeserializationSchemaFactory.scala
```diff
@@ -26,28 +25,28 @@ trait ConfluentKafkaAvroDeserializerFactory extends ConfluentKafkaAvroSerializat
     val clazz = classTag[T].runtimeClass.asInstanceOf[Class[T]]
     val isSpecificRecord = classOf[SpecificRecord].isAssignableFrom(clazz)
```
lciolecki, line 27: Maybe we can extract these two matches to separaed methods?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/ConfluentKafkaAvroDeserializer.scala
```diff
@@ -15,22 +17,25 @@ import pl.touk.nussknacker.engine.avro.schemaregistry.confluent.client.Confluent
   *            fetching schema for deserializer has place at deploy moment. It can be happen when process has
   *            set latest version and deploy was run after new schema was added.
   */
-class ConfluentKafkaAvroDeserializer[T](schema: Schema, confluentSchemaRegistryClient: ConfluentSchemaRegistryClient, var isKey: Boolean, _useSpecificAvroReader: Boolean)
+class ConfluentKafkaAvroDeserializer[T](kafkaConfig: KafkaConfig, schema: Schema, confluentSchemaRegistryClient: ConfluentSchemaRegistryClient,
```
lciolecki, line 20: I know that is not part of your changes.. But maybe we can change:

```
abstract class AbstractConfluentKafkaAvroDeserializer(useSpecificAvroReader: Boolean) 
```
I think this `_useSpecificAvroReader: Boolean` looks ugly..

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/AvroSchemaEvolution.scala
```diff
@@ -4,7 +4,7 @@ import org.apache.avro.Schema
 import org.apache.avro.generic.GenericContainer
 
 trait AvroSchemaEvolution {
-  def alignRecordToSchema[T <: GenericContainer](record: T, schema: Schema): T
+  def alignRecordToSchema(record: GenericContainer, schema: Schema): Any
```
lciolecki, line 7: Why did you changed it?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/BasedOnVersionAvroSchemaDeterminer.scala
```diff
@@ -0,0 +1,35 @@
...
+  override def determineSchemaUsedInTyping: Validated[SchemaRegistryError, Schema] =
+    schemaRegistryClient
+      .getFreshSchema(topic, version, isKey = false)
+
+}
+
+/**
+ * This implementation uses concrete schema version during typing, but in runtime uses schema from record.
+ * It can be useful when you want to have basic typing before deploy but during runtime don't want to do schema evolution -
+ * just trying to use record as it is.
```
mproch, line 27: So this is the case when:
- I have #input of type schemaA 
- I want to to typed filtering on #input.field1
- I want to pass #input to sink, with all (possibly unknown at deployment time) fields?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
```diff
@@ -47,16 +47,11 @@ class DefaultAvroSchemaEvolution extends AvroSchemaEvolution with DatumReaderWri
...
-    * @param record
-    * @param writerSchema
-    * @param readerSchema
-    * @return
     */
-  protected def deserializePayloadToSchema(payload: Array[Byte], record: GenericContainer, writerSchema: Schema, readerSchema: Schema): Any = {
+  protected def deserializePayloadToSchema(payload: Array[Byte], writerSchema: Schema, readerSchema: Schema): Any = {
     try {
-      val reader = createDatumReader(record, writerSchema, readerSchema, useSchemaReflection = useSchemaReflection)
+      // We always want to create generic record at the end, because speecific can has other fields than expected
```
mproch, line 53: I don't understand this case. What happens with evolution for specific record?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/AvroSchemaDeterminer.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import cats.data.Validated
+import org.apache.avro.Schema
+import pl.touk.nussknacker.engine.avro.schemaregistry.SchemaRegistryError
+
+trait AvroSchemaDeterminer extends Serializable {
+
+  def determineSchemaUsedInTyping: Validated[SchemaRegistryError, Schema]
+
+  def determineSchemaInRuntime: Validated[SchemaRegistryError, Option[Schema]]
```
mproch, line 11: In all implementation it's either:
- Some(schema used in typing)
- None
Is it possible there will be other options? If not, maybe flag would be enough?
Also, maybe let's add comment why schema in runtime could be different?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/AvroSchemaDeterminer.scala
```diff
@@ -0,0 +1,13 @@
+package pl.touk.nussknacker.engine.avro
+
+import cats.data.Validated
+import org.apache.avro.Schema
+import pl.touk.nussknacker.engine.avro.schemaregistry.SchemaRegistryError
+
+trait AvroSchemaDeterminer extends Serializable {
+
+  def determineSchemaUsedInTyping: Validated[SchemaRegistryError, Schema]
```
lciolecki, line 9: I think on this leve we should have another type of error - SchemaRegistry is specific implementation.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
lciolecki: Why we return there `Any` if we want always GenericData.Record here?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
```diff
@@ -66,7 +61,7 @@ class DefaultAvroSchemaEvolution extends AvroSchemaEvolution with DatumReaderWri
       } else {
         val start = buffer.position() + buffer.arrayOffset
         val binaryDecoder = decoderFactory.binaryDecoder(buffer.array, start, length, null)
-        val result = reader.read(record, binaryDecoder)
+        val result = reader.read(null, binaryDecoder)
```
lciolecki, line 64: Why wo don't want reuse?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/BaseKafkaAvroSinkFactory.scala
```diff
@@ -2,45 +2,46 @@ package pl.touk.nussknacker.engine.avro.sink
...
 
   override def requiresOutput: Boolean = false
 
   // We currently not using nodeId but it is here in case if someone want to use in their own concrete implementation
   protected def createSink(preparedTopic: PreparedKafkaTopic,
+                           version: Option[Int],
                            output: LazyParameter[AnyRef],
                            kafkaConfig: KafkaConfig,
-                           kafkaAvroSchemaProvider: KafkaAvroSchemaProvider[_],
+                           schemaRegistryProvider: SchemaRegistryProvider[_],
```
lciolecki, line 26: We should pass there only serializationFactory. SchemaRegistryProvider is specific implementation. In some case for examples FixedSchema, there will be no SchemaRegistryProvider.. The same case is in Source.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/BaseKafkaAvroSourceFactory.scala
```diff
@@ -23,19 +25,22 @@ abstract class BaseKafkaAvroSourceFactory[T: ClassTag](processObjectDependencies
   // We currently not using processMetaData and nodeId but it is here in case if someone want to use e.g. some additional fields
   // in their own concrete implementation
   def createSource(preparedTopic: PreparedKafkaTopic,
+                   version: Option[Int],
                    kafkaConfig: KafkaConfig,
-                   kafkaAvroSchemaProvider: KafkaAvroSchemaProvider[T],
+                   schemaRegistryProvider: SchemaRegistryProvider[T],
```
lciolecki, line 30: The same as in Sink. There sould be only: deserializationSchemaFactory and recordFormatter.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/SpecificRecordKafkaAvroSourceFactory.scala
```diff
@@ -0,0 +1,31 @@
...
+import pl.touk.nussknacker.engine.api.typed.ReturningType
+import pl.touk.nussknacker.engine.api.{MetaData, MethodToInvoke, ParamName}
+import pl.touk.nussknacker.engine.avro.schemaregistry.{SchemaRegistryProvider, SpecificRecordEmbeddedSchemaDeterminer}
+import pl.touk.nussknacker.engine.kafka.KafkaFactory._
+import pl.touk.nussknacker.engine.kafka.source.KafkaSource
+import pl.touk.nussknacker.engine.kafka.{KafkaConfig, KafkaUtils}
+
+import scala.reflect._
+
+class SpecificRecordKafkaAvroSourceFactory[T <: SpecificRecord: ClassTag](schemaRegistryProvider: SchemaRegistryProvider[T], processObjectDependencies: ProcessObjectDependencies, timestampAssigner: Option[TimestampAssigner[T]])
```
lciolecki, line 18: Why we don't have SpecificKafkaAvroSinkFactory?

arkadius: On sink side it is not needed to specify that it is either generic or specific case - current `KafkaAvroSinkFactory` handles both cases. One thing that could be useful for backward compatibility is sink factory without version param and with `schemaDeterminer` = `UsingRecordSchemaInRuntimeAvroSchemaDeterminer` but it is quite easy to achive with current api so I won't add it here

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/DefaultAvroSchemaEvolution.scala
lciolecki: Hmm, we can check is specificrecord ;) But it's okey for me.

PR #1068, opened 2020-07-16
[Improvement] we can pick how lax we are during AvroSink output valid…

mproch:
We can control how strict avro sink is when it comes to validation:
- should we check that optional fields are present (even if they're null)
- should we check there are not redundant parameters?

TODO: handle default values properly


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+import pl.touk.nussknacker.engine.api.typed.{CanBeSubclassDeterminer, SingleSubclassRestriction, typing}
+import pl.touk.nussknacker.engine.api.typed.typing.{TypedObjectTypingResult, TypingResult}
+import pl.touk.nussknacker.engine.avro.KafkaAvroBaseTransformer.SinkValueParamName
+import pl.touk.nussknacker.engine.avro.typed.AvroSchemaTypeDefinitionExtractor
+
+object OutputValidator {
+
+  def validateOutput(value: TypingResult, schema: Schema, validationMode: ValidationMode)(implicit nodeId: NodeId): Validated[CustomNodeError, Unit] = {
+    val possibleTypes = AvroSchemaTypeDefinitionExtractor.ExtendedPossibleTypes
+    //FIXME: this still does not handle optional fields validation properly for acceptUnfilledOptional == true?
```
arkadius, line 16: 1. Can you change it to less offensive `TODO`, because it doens't look like it is very "hot" problem
2. I think that it is something missing in this comment. Am I right that you mean "not handle defined optional fields type type validation"? Or something like this? Because now it looks like we not handle this validation at all...

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+    val returnType = new AvroSchemaTypeDefinitionExtractor(skippNullableFields = validationMode.acceptUnfilledOptional).typeDefinition(schema, possibleTypes)
```
arkadius, line 17: I know that it is not part of your change, but `skipNullableFields` is wrong name because we skip `optional` fields. By optional we should understand field with default value. It can be filed without union with null but with default value.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  val strict: ValidationMode = ValidationMode("strict", "Strict mode", acceptUnfilledOptional = false, acceptRedundant = false)
+
+  val allowOptional: ValidationMode = ValidationMode("allowOptional", "No optional parameters validations", acceptUnfilledOptional = true, acceptRedundant = false)
+
+  val allowRedundantAndOptional: ValidationMode =  ValidationMode("allowRedundantAndOptional", "Allow redundant parameters", acceptUnfilledOptional = true, acceptRedundant = true)
+
+  val values: List[ValidationMode] = List(strict, allowOptional, allowRedundantAndOptional)
+
+  def byName(name: String): Option[ValidationMode] = values.find(_.name == name)
```
arkadius, line 15: I don't think that byName will be a good idea to identify mode, because we can change this label. WDYT about having in parameters value `(acceptUnfilledOptional,acceptRedundant)` or just  `acceptUnfilledOptional,acceptRedundant`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
```diff
@@ -50,19 +53,26 @@ class KafkaAvroSinkFactory(val schemaRegistryProvider: SchemaRegistryProvider, v
   private def initialParametersForNode(implicit nodeId: NodeId): WriterT[Id, List[ProcessCompilationError], List[Parameter]] =
     topicParam.map(List(
       Parameter.optional[CharSequence](KafkaAvroBaseTransformer.SinkKeyParamName).copy(isLazyParameter = true),
-      Parameter[AnyRef](KafkaAvroBaseTransformer.SinkValueParamName).copy(isLazyParameter = true), _))
+      Parameter[AnyRef](KafkaAvroBaseTransformer.SinkValueParamName).copy(isLazyParameter = true),
+      //TODO: default value
```
arkadius, line 57: First mode won't be the default one?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+    val restriction = new SingleSubclassRestriction {
+      override def canBeSubclassOf(givenClass: typing.SingleTypingResult, superclassCandidate: typing.SingleTypingResult): Boolean = (givenClass, superclassCandidate) match {
+        case (TypedObjectTypingResult(objFields, _), TypedObjectTypingResult(superFields, _)) if !validationMode.acceptRedundant =>
+          objFields.keys.forall(superFields.keys.toSet.contains)
```
arkadius, line 21: Nice idea, I like this approach - no new mode introduced :+1:

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
```diff
@@ -50,19 +53,26 @@ class KafkaAvroSinkFactory(val schemaRegistryProvider: SchemaRegistryProvider, v
   private def initialParametersForNode(implicit nodeId: NodeId): WriterT[Id, List[ProcessCompilationError], List[Parameter]] =
     topicParam.map(List(
       Parameter.optional[CharSequence](KafkaAvroBaseTransformer.SinkKeyParamName).copy(isLazyParameter = true),
-      Parameter[AnyRef](KafkaAvroBaseTransformer.SinkValueParamName).copy(isLazyParameter = true), _))
+      Parameter[AnyRef](KafkaAvroBaseTransformer.SinkValueParamName).copy(isLazyParameter = true),
```
arkadius, line 56: I wonder if better approach would be having this param in "endingParametersForNode". Because now you will got this parameter at the beginning and you force user to make some hard decision. If it would be at the end with some reasonable default, in 99% user won't be worried of fact that this parameter exist.

mproch: Well, ok, but OTOH it seems good to have it close to output...

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
mproch: yes, true

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
mproch: ok

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
mproch: ok

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
mproch: currently I have name and label to avoid such problems :)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
```diff
@@ -136,8 +136,8 @@ class BestEffortAvroEncoder(avroSchemaEvolution: AvroSchemaEvolution) {
     fields.asScala.map {
       case (fieldName, value) =>
         val field = schema.getField(fieldName)
-        if (field == null) {
-          error(s"Not expected field with name: $fieldName for schema: $schema")
+        if (field == null && !validationMode.acceptRedundant) {
+          error(s"Not expected field with name: $fieldName for schema: $schema and policy $validationMode does not allow redundant")
```
arkadius, line 140: How it will work? Can you add some test for that? In case of `field == null && validationMode.acceptRedundant` I see NPE in here: `field.schema()` :)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.engine.avro.encode
+
+case class ValidationMode private(name: String, label: String, acceptUnfilledOptional: Boolean, acceptRedundant: Boolean)
+
+object ValidationMode {
+
+  val strict: ValidationMode = ValidationMode("strict", "Strict mode", acceptUnfilledOptional = false, acceptRedundant = false)
```
lciolecki, line 7: these all static field should be also final ?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
arkadius: After thinking a while:
1. Shouldn't bidirectional check: `value.canBeSubclassOf(returnType) && returnType.canBeSubclassOf(value)` will be enough?
2. Maybe better approch would be just extends some method `checkObjectFields` in `CanBeSubclassDeterminer` instead of creating new abstraction. Current solution will be complicated in the future when we will want to change `&& restriction....` to `|| restriction` or sth similar

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
arkadius: Indeed I see now. IMO `acceptUnfilledOptional,acceptRedundant` will be better because it will be easier to migrate such value in case when our validation approach will slightly changed e.g. one flag will be replaced by another or we will add other combinations of modes ... but it is also fine :+1:

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
arkadius: @lciolecki In scala, final modifier for object's vals do nothing. `final` mean "won't be overridden" `val` mean "is contant"

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/OutputValidator.scala
mproch: 1. It also came to my mind, but I think cases like: String vs CharSequence, Int vs Long etc. will cause problems 
2. good idea, I'll try it.

mproch: done

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
mproch: done...

PR #1069, opened 2020-07-16
Avro sink: added optional key param

arkadius:
Plus others:
- renamed other params: topic -> Topic, output -> Output
- sink accept not only maps, but also primitives


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/serialization/KafkaAvroSerializationSchemaFactory.scala
```diff
@@ -1,46 +1,45 @@
...
 
-  protected def createKeySerializer(kafkaConfig: KafkaConfig): Serializer[AnyRef] =
-    new UUIDSerializer[AnyRef]
-
   protected def createValueSerializer(schemaOpt: Option[Schema], version: Option[Int], kafkaConfig: KafkaConfig): Serializer[AnyRef]
 
-  override def create(topic: String, version: Option[Int], schemaOpt: Option[NkSerializableAvroSchema], kafkaConfig: KafkaConfig): KafkaSerializationSchema[AnyRef] = {
-    new KafkaSerializationSchema[AnyRef] {
-      private lazy val keySerializer = createKeySerializer(kafkaConfig)
+  override def create(topic: String, version: Option[Int], schemaOpt: Option[NkSerializableAvroSchema], kafkaConfig: KafkaConfig): KafkaSerializationSchema[StringKeyedValue[AnyRef]] = {
```
mproch, line 35: I think this will make it difficult to add other key serializations without breaking API. Maybe we can move information about StringKeyedValue a bit higher?

PR #1072, opened 2020-07-17
[FIX] Global variables are handled corrently for generic transformations

mproch:


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesResources.scala
```diff
@@ -60,14 +55,21 @@ class NodesResources(val processRepository: FetchingProcessRepository[Future],
...
               implicit val metaData: MetaData = nodeData.processProperties.toMetaData(process.id)
+
+              val emptyCtx = GlobalVariablesPreparer(modelData.processWithObjectsDefinition.expressionConfig).emptyValidationContext(metaData)
+
+              //It's a bit tricky, because FE does not distinguish between global and local vars...
+              def prepareContext(variableTypes: Map[String, TypingResult]) = {
+                val localVars = variableTypes.filterNot(e => emptyCtx.globalVariables.keys.toSet.contains(e._1))
+                emptyCtx.copy(localVariables = localVars)
+              }
+              val validationContext = prepareContext(nodeData.variableTypes)
```
arkadius, line 70: This Context preparation logic starting to be long, can we extract some class taking  nodeData and returning ValidationContext configured by modelData? Or returned by modelData?

mproch: I extracted a bit. However, I think that the fact that there are global variables in map is specific behaviour of UI and we should not move this logic e.g. to engine

PR #1080, opened 2020-07-22
[FIX] We don't show id field errors in node id

mproch:


ui/client/components/graph/node-modal/NodeDetailsContent.js
```diff
@@ -130,6 +130,8 @@ export class NodeDetailsContent extends React.Component {
     }
   }
 
+  idField = () => this.createField("input", "Name", "id", true, [mandatoryValueValidator])
```
lciolecki, line 133: I think we should use component instead of this function..

PR #1081, opened 2020-07-22
[Improvement] Validate more nodes dynamically

mproch:
This PR extends dynamic validation to new types of nodes
- Standard (non-generic) Sources, Sinks and CustomNodes
- Services (Enrichers and Processors)

The validation logic is refactored a bit, so that it is unified between node validation and process compilation. Next steps:
- Validation/unification of validation of variable declarations, switches etc.
- Refactoring of Service handling so that we can introduce e.g. EagerService


arkadius (approved):
LGTM :)

PR #1084, opened 2020-07-23
[Refactor] Extract NodeCompiler from ProcessCompiler

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -0,0 +1,272 @@
...
+import pl.touk.nussknacker.engine.graph.node.SubprocessInputDefinition.SubprocessParameter
+import pl.touk.nussknacker.engine.graph.node._
+import pl.touk.nussknacker.engine.graph.{evaluatedparam, node}
+import pl.touk.nussknacker.engine.variables.GlobalVariablesPreparer
+import pl.touk.nussknacker.engine.{Interpreter, api}
+import shapeless.Typeable
+import shapeless.syntax.typeable._
+
+class NodeCompiler(definitions: ProcessDefinition[ObjectWithMethodDef],
+                   val objectParametersExpressionCompiler: ExpressionCompiler,
```
arkadius, line 32: Can you change visibility to protected?

mproch: done, also changed some unnecessary protected stuff to private...

PR #1088, opened 2020-07-23
MathUtils with common math functions (sum, max, min) 

arkadius:
with semantic similar to SpEL operators.


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+        override def onLongs(n1: Long, n2: Long): Long = n1 + n2
+        override def onBigIntegers(n1: BigInteger, n2: BigInteger): BigInteger = n1.add(n2)
+        override def onFloats(n1: Float, n2: Float): Float = n1 + n2
+        override def onDoubles(n1: Double, n2: Double): Double = n1 + n2
+        override def onBigDecimals(n1: java.math.BigDecimal, n2: java.math.BigDecimal): java.math.BigDecimal = n1.add(n2)
+      })
+    }
+  }
+  
+  private def withNotNullValues(n1: Number, n2: Number)(f: => Number)
```
mproch, line 59: Maybe we can make this (and withValuesWithTheSameType) protected? So that more cases can be added in models if needed (e.g. multiplication or sth else?)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/supertype/NumberTypesPromotionStrategy.scala
```diff
@@ -1,13 +1,43 @@
...
+  }
+
+  private def toSingleTypesSet(typ: TypingResult) : Either[Unknown.type, Set[SingleTypingResult]] =
+    typ match {
+      case s: SingleTypingResult => Right(Set(s))
+      case u: TypedUnion => Right(u.possibleTypes)
+      case Unknown => Left(Unknown)
+    }
+
+  final def promoteClasses(left: Class[_], right: Class[_]): TypingResult = {
```
mproch, line 32: Do we want TypingResult or Class/TypedClass here? In mathUtils we in fact rely on the fact that this method returns TypedClass?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      f
+    }
+  }
+
+  private def withValuesWithTheSameType(n1: Number, n2: Number)(handler: SameNumericTypeHandler)
+                                       (implicit promotionStrategy: NumberTypesPromotionStrategy): Number = {
+    val promotedClass = promotionStrategy.promoteClasses(n1.getClass, n2.getClass).asInstanceOf[SingleTypingResult].objType.klass
+    if (promotedClass == classOf[java.lang.Byte]) {
+      handler.onBytes(SpringNumberUtils.convertNumberToTargetClass(n1, classOf[java.lang.Byte]), SpringNumberUtils.convertNumberToTargetClass(n2, classOf[java.lang.Byte]))
```
mproch, line 74: maybe we can write val promoted1 = SpringNumberUtils.convertNumberToTargetClass(n1, promotedClass) just once?
and than do pattern matching?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/supertype/NumberTypesPromotionStrategy.scala
```diff
@@ -21,41 +51,60 @@ object NumberTypesPromotionStrategy {
 
   def isFloatingNumber(clazz: Class[_]): Boolean = FloatingNumbers.contains(clazz)
 
-  private val DecimalNumbers: mutable.LinkedHashMap[Class[_], Class[_]] = mutable.LinkedHashMap(
-    classOf[java.math.BigInteger] -> classOf[java.math.BigInteger],
-    classOf[java.lang.Long] -> classOf[java.lang.Long],
-    classOf[java.lang.Integer] -> classOf[java.lang.Integer],
-    classOf[java.lang.Short] -> classOf[java.lang.Integer],
-    classOf[java.lang.Byte] -> classOf[java.lang.Integer]
+  val DecimalNumbers: Seq[Class[_]] = IndexedSeq(
```
mproch, line 54: Let's add comment that order is important here :)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/supertype/NumberTypesPromotionStrategy.scala
arkadius: There is little trick here. For some promotions we can't be sure correct type. During random tests I found out that for power operator, Ints can be promoted to Longs or not. So in this special case this promotion strategy return union of types. So basically `promoteClasses` is rather "internal" method. Main method is `promote(left: TypingResult, right: TypingResult): TypingResult`. I could add additional abstraction level for `NumberTypesPromotionStrategy` for cases where during typing we will got only single case. WDYT?

mproch: Ok... I see now :) 
We can add one more layer - or just add comment/better handling in https://github.com/TouK/nussknacker/pull/1088/files#diff-7d2e0e302a77fb47e52efb8101378415R72

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
arkadius: Do you want to pattern mach on (promoted1, promoted2)? In this case we will lost type checking of `convertNumberToTargetClass` that it returns `T` from `Class<T>`. I'm not sure if it will be more mistake friendly... Or maybe you have some other idea?

mproch: yeah, true, probably there is not ideal solution here...

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/supertype/NumberTypesPromotionStrategy.scala
arkadius: I've added additional abstraction - now we have less downcasting - good point.

PR #1090, opened 2020-07-23
[Improvement] Publish branch-date-... to enable version ranges

mproch:


mproch:
Let's try to run successfully on JDK 11 instead...

PR #1091, opened 2020-07-24
[FIX]  handle union after split,switch,filter

mproch:
Also minor fixes in demo/docker

### DIFF
--- a/demo/docker/docker-compose-env.yml
+++ b/demo/docker/docker-compose-env.yml
@@ -1,4 +1,4 @@
-version: '3.0'
+version: '3.7'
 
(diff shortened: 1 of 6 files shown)


demo/docker/docker-compose-env.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.0'
+version: '3.7'
```
arkadius, line 1: Why you upgraded docker-compose version? We should have as lowest version as possible (until we use some features from higher version), because it cause that this docker-compose won't run with older dockers...

PR #1096, opened 2020-07-24
Try to tune sbt for jdk11...

mproch:


mproch:
Let's try to upgrade to JDK 11 instead...

PR #1099, opened 2020-07-29
Outer joins

arkadius:


engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/OuterJoinTransformerSpec.scala
```diff
@@ -0,0 +1,190 @@
...
+            new CoProcessFunctionInterceptor(super.prepareAggregatorFunction(aggregator, stateTimeout)) with LazyLogging {
+
+              @transient
+              private lazy val addedToStateCount = new AtomicInteger(0)
+
+              @transient
+              private lazy val stateLock = new Object
+
+              override protected def beforeProcessElement1(value: ValueWithContext[String]): Unit = {
+                logger.info(s"OuterJoin: processing element1 with key: ${value.value} - waiting on all elements added to state...")
```
mproch, line 143: logger.debug? here and in following occurrences?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/outer/OuterJoinTransformer.scala
```diff
@@ -0,0 +1,119 @@
...
+import pl.touk.nussknacker.engine.api.definition.{NodeDependency, OutputVariableNameDependency, Parameter, ParameterWithExtractor}
+import pl.touk.nussknacker.engine.flink.api.compat.ExplicitUidInOperatorsSupport
+import pl.touk.nussknacker.engine.flink.api.process.{FlinkCustomJoinTransformation, FlinkCustomNodeContext}
+import pl.touk.nussknacker.engine.flink.util.keyed.{StringKeyOnlyMapper, StringKeyedValue, StringKeyedValueMapper}
+import pl.touk.nussknacker.engine.flink.util.timestamp.TimestampAssignmentHelper
+import pl.touk.nussknacker.engine.flink.util.transformer.aggregate.{AggregateHelper, Aggregator}
+
+import scala.concurrent.duration.FiniteDuration
+
+class OuterJoinTransformer(timestampAssigner: Option[TimestampAssigner[TimestampedValue[ValueWithContext[AnyRef]]]])
```
mproch, line 24: I wonder if we need additional transformer, which handles situation when we don't evict JOINED, but we always take last one - currently there is simplified implementation in EnrichWithAdditionalDataTransformer in DevConfigCreator. I know it can be emulated with "inifite" window length, and "last" aggregation - but it's not v. elegant?

engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/OuterJoinTransformerSpec.scala
arkadius: The problem is that I'm not 100% sure, that it will work correctly. I've tried a couple of methods to do such synchronization and I'm not sure if the final one is working. Can we leave it as it is for some time? I have quite similar concept in other project, but there I have kafka source and I think that with collection source there can be some sequence that will cause deadlock ... :-/

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/outer/OuterJoinTransformer.scala
arkadius: IMO even if we will have fixed last synchronization we should always have some time-to-leave (window) to be sure that we don't have some leak of resources... but by this comment you want to say that I should add some abstraction with such will be easier to achieve this without too much code duplication?

mproch: Don't know TBH... we can leave it as it is now...

engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/OuterJoinTransformerSpec.scala
mproch: well... ok... ;)

PR #1110, opened 2020-08-05
Improvements: Change KavkaAvroSink params ordering

lciolecki:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
```diff
@@ -14,10 +14,14 @@ import pl.touk.nussknacker.engine.avro.schemaregistry.SchemaRegistryProvider
 import pl.touk.nussknacker.engine.avro.{KafkaAvroBaseTransformer, SchemaDeterminerErrorHandler}
 import pl.touk.nussknacker.engine.flink.api.process.FlinkSink
 
-class KafkaAvroSinkFactory(val schemaRegistryProvider: SchemaRegistryProvider,
-                           val processObjectDependencies: ProcessObjectDependencies)
+class KafkaAvroSinkFactory(val schemaRegistryProvider: SchemaRegistryProvider, val processObjectDependencies: ProcessObjectDependencies)
   extends BaseKafkaAvroSinkFactory with KafkaAvroBaseTransformer[FlinkSink] {
 
+  private val avroMessageParams = List(
```
arkadius, line 20: Why `validationModeParam` is not here? Maybe it should be named `paramsDeterminedAfterSchema`

PR #1126, opened 2020-08-06
Period souce and delay node

arkadius:


engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/DelayTransformer.scala
```diff
@@ -0,0 +1,83 @@
...
+import pl.touk.nussknacker.engine.api
+import pl.touk.nussknacker.engine.api._
+import pl.touk.nussknacker.engine.flink.api.compat.ExplicitUidInOperatorsSupport
+import pl.touk.nussknacker.engine.flink.api.process.{FlinkCustomNodeContext, FlinkCustomStreamTransformation}
+import pl.touk.nussknacker.engine.flink.util.keyed.StringKeyOnlyMapper
+
+object DelayTransformer extends CustomStreamTransformer with ExplicitUidInOperatorsSupport {
+
+  @MethodToInvoke(returnType = classOf[Void])
+  def invoke(@ParamName("key") key: LazyParameter[CharSequence],
```
mproch, line 21: I started to wonder - do we really need the key to be passed by user explicitly? 
We need it internally - but isn't it enough to generate random number or sth like that? 
We used similar element at one of the clients and they struggled to understand what the key should be for this element....

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/DelayTransformer.scala
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class DelayFunction(nodeId: String, delay: Duration)
+  extends KeyedProcessFunction[String, ValueWithContext[String], ValueWithContext[AnyRef]] {
+
+  type StateType = List[TimestampedValue[api.Context]]
+  type FlinkCtx = KeyedProcessFunction[String, ValueWithContext[String], ValueWithContext[AnyRef]]#Context
+  type FlinkTimerCtx = KeyedProcessFunction[String, ValueWithContext[String], ValueWithContext[AnyRef]]#OnTimerContext
+
+  private val descriptor = new ValueStateDescriptor[StateType]("state", classOf[StateType])
```
mproch, line 43: Wouldn't ListState be better?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/PeriodicSourceFactory.scala
```diff
@@ -0,0 +1,99 @@
...
+// TODO: add testing capabilities
+object PeriodicSourceFactory extends PeriodicSourceFactory(new MapAscendingTimestampExtractor(MapAscendingTimestampExtractor.DefaultTimestampField))
+
+class PeriodicSourceFactory(timestampAssigner: TimestampAssigner[jul.Map[String, AnyRef]]) extends FlinkSourceFactory[jul.Map[String, AnyRef]]  {
+
+  @MethodToInvoke
+  def create(@ParamName("period") period: Duration,
+             // TODO: @DefaultValue(1) instead of nullable
+             @ParamName("count") @Nullable nullableCount: Integer,
+             @ParamName("value") value: LazyParameter[jul.Map[String, AnyRef]]): Source[_] = {
```
mproch, line 32: does it have to be map? Or just AnyRef? I know you have MapAscendingTimestampExtractor below, but for constant time field maybe we could have just another field?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/PeriodicSourceFactory.scala
```diff
@@ -0,0 +1,99 @@
...
+      override def timestampAssignerForTest: Option[TimestampAssigner[jul.Map[String, AnyRef]]] = Some(timestampAssigner)
+
+      override val returnType: typing.TypingResult = value.returnType
+
+    }
+  }
+
+}
+
+// CheckpointedFunction to handle checkpoints of underlying stream - not sure if it is need
```
mproch, line 67: I don't think it is needed if we don't actually use state?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/DelayTransformer.scala
arkadius: I've made key optional, with default value - empty string. Also used MapState, so reading state won't hurt us even for one key

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/PeriodicSourceFactory.scala
arkadius: Good point, changed to AnyRef. Would be nice to have `@Placeholder({:})` which not works now :/

PR #1127, opened 2020-08-07
Fix metrics query for 1 hour window

jedrz:
Currently for newly deployed processes zero counts values are returned. What's more count values are incorrect (too small) because of the last aggregation.


arkadius:
Please add notice about severity and create 0.2-fixes branch with bugfixes to release

PR #1144, opened 2020-08-11
case insensitive sorting

### LINKED ISSUE #1138: Process sorting should be case insensitive
**Describe the bug**
As in description, as a user I expect natural sorting of `a, A, z, Z` and not by ASCII code, which is what I get now: `A, Z, a, z`.

**To Reproduce**
Open process list, create a few processes, sort them.

JulianWielga:
resolves #1138

### DIFF
--- a/ui/client/containers/ProcessesList.tsx
+++ b/ui/client/containers/ProcessesList.tsx
@@ -80,6 +80,8 @@ function useFiltersState(defaultQuery: Queries) {
   return {search, filters, setFilters}
 }
 
+const sortFunction = (a: string, b: string) => a.toLowerCase().localeCompare(b.toLowerCase())
+
 export function ProcessesList(props: BaseProcessesOwnProps) {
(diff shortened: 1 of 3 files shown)


ui/client/containers/ProcessesList.tsx
```diff
@@ -80,6 +80,8 @@ function useFiltersState(defaultQuery: Queries) {
   return {search, filters, setFilters}
 }
 
+const sortFunction = (a: string, b: string) => a.toLowerCase().localeCompare(b.toLowerCase())
```
Crack, line 83: Shouldn't it be simply a cached instance of  `Intl.Collator().compare`?

```javascript
> ["a", "A", "Z", "z", "1", "二", "一"].sort(Intl.Collator().compare)
[ "1", "a", "A", "z", "Z", "一", "二" ]
```

JulianWielga: sure, corrected

PR #1148, opened 2020-08-13
Fix: FE regexp for node id matching

lciolecki:


ui/client/common/ProcessUtils.js
```diff
@@ -50,14 +50,17 @@ class ProcessUtils {
...
     return _.transform(nodeResults || {}, function(result, value, key) {
-      const branch = key.match(new RegExp(`^\\$edge-(.*)-${nodeId}$`))
+      const branch = key.match(new RegExp(`^\\$edge-(.*)-${escapedNodeId}$`))
       if (branch && branch.length > 1) {
         result[branch[1]] = value.variableTypes
       }
     }, {})
   }
 
+  escapeId = (id) => id.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
mproch, line 62: escapeForRegexp? Or sth like that - because it will be different escaping for e.g. URL? And let's add link to source of this expression :)

PR #1150, opened 2020-08-13
[Improvement] Extract stuff non-compatible with Flink<1.10

mproch:


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -302,7 +306,7 @@ object flinkRestModel {
 
   @JsonCodec(decodeOnly = true) case class JobsResponse(jobs: List[JobOverview])
 
-  @JsonCodec(decodeOnly = true) case class JobOverview(jid: String, name: String, `last-modification`: Long, `start-time`: Long, state: JobStatus)
+  @JsonCodec(decodeOnly = true) case class JobOverview(jid: String, name: String, `last-modification`: Long, `start-time`: Long, state: String)
```
lciolecki, line 309: Why not JobStatus? Now we can put there any string? :(

mproch: I added comment

PR #1151, opened 2020-08-13
[Improvement] Upgrade typesafe config

mproch:
One of the main reasons is https://github.com/lightbend/config/pull/620

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -161,7 +161,7 @@ lazy val commonSettings =
       ),
       //here we add dependencies that we want to have fixed across all modules
       dependencyOverrides ++= Seq(
-        //currently Flink (1.9) uses 1.8.2 Avro version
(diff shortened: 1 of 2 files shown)


lciolecki:
Please add info at Changelog.md

build.sbt
```diff
@@ -162,7 +162,8 @@ lazy val commonSettings =
       //here we add dependencies that we want to have fixed across all modules
       dependencyOverrides ++= Seq(
         //currently Flink (1.9) uses 1.8.2 Avro version
```
lciolecki, line 164: This comment is now depricated?

mproch: 1.9 -> deprecated, but even now (Flink 1.11) 1.8.2 is used. It will remain so for a LONG time I guess (hive problems etc...)

PR #1155, opened 2020-08-18
metrics link

JulianWielga:


ui/client/containers/plainStyleLink.tsx
```diff
@@ -0,0 +1,15 @@
+import {css} from "emotion"
+import React from "react"
+import {Link, LinkProps} from "react-router-dom"
+
+export function PlainStyleLink(props: LinkProps) {
```
lciolecki, line 5: I think.. PlainStyleLink and ProcessLink is the same? Can we do one more generic component?

PR #1161, opened 2020-08-26
Bunch of fixes for NK 0.2 - mainly handling os specific records + MigrationGuide

arkadius:


alien11689 (approved):

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/BaseKafkaAvroSourceFactory.scala
```diff
@@ -24,23 +24,34 @@ abstract class BaseKafkaAvroSourceFactory[T: ClassTag](timestampAssigner: Option
...
-    new KafkaSource(
-      List(preparedTopic),
-      kafkaConfig,
-      deserializationSchemaFactory.create[T](schemaUsedInRuntime, kafkaConfig),
-      assignerToUse(kafkaConfig),
-      createRecordFormatter(preparedTopic.prepared),
-      TestParsingUtils.newLineSplit
-    ) with ReturningType {
-      override def returnType: typing.TypingResult = returnTypeDefinition
+    if (returnGenericAvroType) {
```
lciolecki, line 35: Mayby shorter:

````
override def returnType: typing.TypingResult = {
if (returnGenericAvroType) {
AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
} else {
super.returnType
}
}

lciolecki (approved):

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/BaseKafkaAvroSourceFactory.scala
arkadius: It won't work, because `ReturningType` is a marker interface and it will change a behaviour

PR #1166, opened 2020-08-31
[Improvement] Docker config refactor

mproch:
The idea is that we try to stick to following convention:
- default configuration of model is done in model.conf 
- default configuration of UI is done in defaultConfig.conf 
- we enable overrding properties with ENV variables (via CONFIG_FORCE)

In this PR initial refactor is done, with some documentation and sanity checks. 
In next PRs further default config refactor will be made


docs/Configuration.md
```diff
@@ -70,8 +70,16 @@ countsSettings {
   password: "admin"
 }
 
-```
-In the next sections we'll look at 
+```             
+###Default configurations
+Default configuration for UI is in [defaultConfig.conf](https://github.com/TouK/nussknacker/blob/staging/ui/server/src/main/resources/defaultConfig.conf).
```
arkadius, line 75: It would be nice to have some more clarified distinguishion of default configs for UI and for model. Maybe we should name it `defaultUiConfig.conf`?

docs/Configuration.md
```diff
@@ -70,8 +70,16 @@ countsSettings {
...
+We don't use ```reference.conf``` at the moment, as classloaders of model and ui are not separated, and we don't want UI config to be passed to model. 
+
+Default configuration of models is taken from ```model.conf``` files in model jar (see e.g. [model.conf](https://github.com/TouK/nussknacker/blob/staging/engine/flink/generic/src/main/resources/model.conf)).
```
arkadius, line 78: I don't exactly get it. How someone can override this default config? If it is really default config that can be overriden maybe better will be call it `defaultModelConfig.conf`?

docs/Configuration.md
```diff
@@ -70,8 +70,16 @@ countsSettings {
...
+You can also use ```reference.conf``` in model jars, however we found some problems with substitutions (see docs in [ModelConfigToLoad](https://github.com/TouK/nussknacker/blob/staging/engine/flink/generic/src/main/resources/model.conf)).
+ 
+All configurations can be overridden with environmental variables. Please consult [TypesafeConfig documentation](https://github.com/lightbend/config#optional-system-or-env-variable-overrides).
```
arkadius, line 81: And how to override only one property like `globalParameters.explicitUidInStatefulOperators`?

docs/Changelog.md
```diff
@@ -12,10 +12,13 @@ Nussknacker versions
 ------------------------
 * [#1128](https://github.com/TouK/nussknacker/pull/1128) Union-memo transformer
 * [#1054](https://github.com/TouK/nussknacker/pull/1054) Tabbed dark process list
-* [#1151](https://github.com/TouK/nussknacker/pull/1151) Upgrade Typesafe Config to latest version
+* Configuration improvements (library upgrade, conventions): 
+  [#1151](https://github.com/TouK/nussknacker/pull/1151), 
+  [#1151](https://github.com/TouK/nussknacker/pull/1166)
```
arkadius, line 17: you have typo here, should be `[#1166]`

mproch: thanks, fixed

docs/Configuration.md
mproch: I added more explanation + example

docs/Configuration.md
mproch: I added example

docs/Configuration.md
mproch: ok, renamed

docs/Configuration.md
mproch: I renamed as you suggested

arkadius (approved):

PR #1170, opened 2020-09-01
Defined*Parameter used in generic transformations holds TypedExpression instead of TypingResult

arkadius:


mproch (approved):
Please add also migration guide ;)

PR #1175, opened 2020-09-02
Fix for: BestEffortAvroEncoder haven't produced record with logical types for missing field with default values

arkadius:


lciolecki (approved):
LGTM

PR #1187, opened 2020-09-04
Fix docker flinkv

mproch:


nadberezny (changes requested):
This Flink bump requires chages in our Taskmanager memory config:
https://github.com/TouK/nussknacker/pull/1195/commits/561f9b152903d21ce2799ace53eb9a613575af7c

nadberezny (approved):

Issue #249, opened 2019-10-23
Showing inferred expression type


arkadius:
Would be nice if somewhere near expression will be shown type of inferred expression. Especially it will be helpful when someone define new variable and didn't know that inferred expression's type is unknown. This can produce a lot of errors in expressions using this variable.

This type can be shown only for validated processes.

witekw:
Shall we show it in the same way as parameter types under the label "Expression"?

![image](https://user-images.githubusercontent.com/632221/67776088-1f460e00-fa60-11e9-86e8-655bc395f66c.png)

![image](https://user-images.githubusercontent.com/632221/67776111-2705b280-fa60-11e9-8710-a9b0ce3fdf70.png)

Where do you plan to return this information (process API endpoint)?
Will it be available before saving the process?

arkadius:
Yes, IMO it may be presented the same way as parameter type.
We return it in validation result, in `variableTypes` field. When we will add validation in context of node, it will be simpler to extract this type, but for now it is also possible.

arkadius:
After thinking a while, presenting this information exactly like in parameter type can be confusing.
User may think that it is declared type, not inferred type. Declared type that he or she can't change and expression must produce the same type.
I think that the place may be right, but we need to think how to make it less ambiguous... Any idea? Maybe some little icon :information_source: with tooltip that the type is inferred?

arkadius:
Notice: At first step we can update it only after validation of process (closing modal and going back to them). But eventually it should be updated live during coding of expression

fi-:
It should look like this:
![inferredtype](https://user-images.githubusercontent.com/8850099/72062240-affa5f00-32d7-11ea-828e-0ce871d0c6a3.jpg)
height: 18px
bg color: #292929 
text color: #808080
text color (type): ideally every type could have a corresponding color, but on default we can stick with #6a8759 
font: consolas regular
the box should stretch to message width

for now we can't display inferred type in real time so meanwhile we will display an alternative message when we can't display the type yet.

nadberezny:
Resolved with:
https://github.com/TouK/nussknacker/pull/1191
https://github.com/TouK/nussknacker/pull/1222

PR #1191, opened 2020-09-04
Variable node validation * inferred type info [#249]

nadberezny:
1. Validate variable expression definition in Variable modal.
2. Display inferred type in variable expression definition.

TODO in next iteration:
0. Return typed expression in ValidationResult of DisplayableProcess
1. VariableBuilder validation & inferred type
2. Refactor of NodeDataValidator - validation logic should be moved to NodeCompiler

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
@@ -21,12 +21,15 @@ object TypeEncoders {
   private val encodeUnknown = JsonObject("refClazzName" -> fromString(classOf[Object].getName),
     "params" -> fromValues(Nil))
 
-  private def encodeTypingResult(result: TypingResult): Json = fromJsonObject((result match {
-    case single: SingleTypingResult => encodeSingleTypingResult(single)
-    case typing.Unknown => encodeUnknown
-    case TypedUnion(classes) =>
-      JsonObject("union" -> fromValues(classes.map(encodeTypingResult).toList))
-  }).+:(typeField -> fromString(TypingType.forType(result).toString)))
+  private def encodeTypingResult(result: TypingResult): Json = fromJsonObject(
+    (result match {
+      case single: SingleTypingResult => encodeSingleTypingResult(single)
+      case typing.Unknown => encodeUnknown
+      case TypedUnion(classes) =>
+        JsonObject("union" -> fromValues(classes.map(encodeTypingResult).toList))
+    })
+      .+:(typeField -> fromString(TypingType.forType(result).toString))
+      .+:("display" -> fromString(result.display)))
 
   private def encodeSingleTypingResult(result: SingleTypingResult): JsonObject = result match {
     case TypedObjectTypingResult(fields, objType) =>
--- a/ui/client/components/graph/node-modal/editors/expression/Editor.ts
+++ b/ui/client/components/graph/node-modal/editors/expression/Editor.ts
@@ -22,7 +22,6 @@ import TextareaEditor from "./TextareaEditor"
 import JsonEditor from "./JsonEditor"
 import DualParameterEditor from "./DualParameterEditor"
 
-export type ParamType = $TodoType
 type ValuesType = Array<string>
 type EditorProps = $TodoType
 
--- a/ui/client/components/graph/node-modal/editors/expression/Editor.ts
+++ b/ui/client/components/graph/node-modal/editors/expression/Editor.ts
@@ -22,7 +22,7 @@ import TextareaEditor from "./TextareaEditor"
 import JsonEditor from "./JsonEditor"
 import DualParameterEditor from "./DualParameterEditor"
 
-type ParamType = $TodoType
+export type ParamType = $TodoType
 type ValuesType = Array<string>
 type EditorProps = $TodoType
 
--- a/ui/client/components/modals/ValidationLabels.tsx
+++ b/ui/client/components/modals/ValidationLabels.tsx
@@ -35,6 +35,8 @@ export default function ValidationLabels(props: Props) {
     )
   )
 
+  // TODO: We're assuming that we have disjoint union of type info & validation errors, which is not always the case.
+  // It's possible that expression is valid and it's type is known, but a different type is expected.
   return (
     <div className={`validation-labels ${additionalClassName}`}>
       { isValid ? (
--- a/ui/client/components/modals/ValidationLabels.tsx
+++ b/ui/client/components/modals/ValidationLabels.tsx
@@ -1,25 +1,49 @@
 import React from "react"
-import {v4 as uuid4} from "uuid"
+import _ from "lodash"
 import {Validator, withoutDuplications} from "../graph/node-modal/editors/Validators"
 
 type Props = {
   validators: Array<Validator>,
   values: Array<string>,
   additionalClassName?: string,
+  validationLabelInfo?: string,
 }
 
 export default function ValidationLabels(props: Props) {
 
-  const {validators, values, additionalClassName} = props
+  type ValidationErrors = {
+    message: string,
+    description: string,
+  }
+
+  const {validators, values, additionalClassName, validationLabelInfo} = props
+
+  const validationErrors: ValidationErrors[] = withoutDuplications(validators)
+    .filter(v => !v.isValid(...values))
+    .map(validator => ({
+      message: validator.message && validator.message(),
+      description: validator.description && validator.description(),
+    }))
+
+  const isValid: boolean = _.isEmpty(validationErrors)
+
+  const renderErrorLablels = () => validationErrors.map(
+    (validationError, ix) => (
+      <span key={ix} className="validation-label-error" title={validationError.description}>
+        {validationError.message}
+      </span>
+    )
+  )
 
   return (
     <div className={`validation-labels ${additionalClassName}`}>
-      {withoutDuplications(validators).map(validator => validator.isValid(...values) ?
-        null : (
-          <span key={uuid4()} className="validation-label" title={validator.description && validator.description()}>
-            {validator.message()}
-          </span>
-        ))}
+      { isValid ? (
+        <span className="validation-label-info" title="Info">
+          {validationLabelInfo}
+        </span>
+      ) :
+        renderErrorLablels()
+      }
     </div>
   )
 }
--- a/ui/client/components/modals/ValidationLabels.tsx
+++ b/ui/client/components/modals/ValidationLabels.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import _ from "lodash"
+import {isEmpty} from "lodash"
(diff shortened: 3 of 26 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
```diff
@@ -21,12 +21,15 @@ object TypeEncoders {
...
-  }).+:(typeField -> fromString(TypingType.forType(result).toString)))
+  private def encodeTypingResult(result: TypingResult): Json = fromJsonObject(
+    (result match {
+      case single: SingleTypingResult => encodeSingleTypingResult(single)
+      case typing.Unknown => encodeUnknown
+      case TypedUnion(classes) =>
+        JsonObject("union" -> fromValues(classes.map(encodeTypingResult).toList))
+    })
+      .+:(typeField -> fromString(TypingType.forType(result).toString))
+      .+:("display" -> fromString(result.display)))
```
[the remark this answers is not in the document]
nadberezny: @arkadius `display` is a pretty version of ref.klass.getName; this is, instead of `java.lang.Long` `Long` is displayed; instead of `java.lang.Object` `Unknown` is displayed.
We could have independent implementation, only for FE on FE side, but in some cases it would be tricky to implement on FE.
eg. for `TypedClass#display` currently we're using Java reflection:
```
    override def display: String = {
      val className = ReflectUtils.fixedClassSimpleNameWithoutParentModule(ClassUtils.primitiveToWrapper(klass))
      if (params.nonEmpty)
        s"$className[${params.map(_.display).mkString(",")}]"
      else
        s"$className"
    }
```

But I see your point, that it would be nice to separate FE and BE resposibilities, but maybe for now using TypingResult#display is acceptable, and we address the issue in a separate PR.?

ui/client/types/definition.ts
```diff
@@ -7,17 +7,28 @@ export type UIObjectDefinition = {
     nodeConfig: SingleNodeConfig,
 }
 
-export type TypingResult = $TodoType
+export type TypingResult = {
```
mproch, line 10: This is not correct definition, I think. There is no "name" field, params and refClazzName are not present in all cases (e.g. union or dict instance)?

ui/client/components/modals/ValidationLabels.tsx
```diff
@@ -1,25 +1,49 @@
...
 
   return (
     <div className={`validation-labels ${additionalClassName}`}>
-      {withoutDuplications(validators).map(validator => validator.isValid(...values) ?
-        null : (
-          <span key={uuid4()} className="validation-label" title={validator.description && validator.description()}>
-            {validator.message()}
-          </span>
-        ))}
+      { isValid ? (
```
mproch, line 40: Here we assume we either have errors or validation info? Is it always the case? E.g. expression may type correctly, but is invalid for other reasone?

ui/client/components/modals/ValidationLabels.tsx
```diff
@@ -1,25 +1,49 @@
 import React from "react"
-import {v4 as uuid4} from "uuid"
+import _ from "lodash"
```
mproch, line 2: I think we try not to import _ anymore, but only what is needed

ui/client/components/graph/node-modal/editors/expression/Editor.ts
```diff
@@ -22,7 +22,7 @@ import TextareaEditor from "./TextareaEditor"
 import JsonEditor from "./JsonEditor"
 import DualParameterEditor from "./DualParameterEditor"
 
-type ParamType = $TodoType
+export type ParamType = $TodoType
```
mproch, line 25: Maybe let's move it to type definitions if we want to export it?

ui/client/reducers/nodeDetailsState.ts
```diff
@@ -1,9 +1,10 @@
 import {Action} from "../actions/reduxTypes"
 import {ValidationData} from "../actions/nk"
-import {NodeValidationError, UIParameter} from "../types"
+import {NodeValidationError, UIParameter, UITypedExpression} from "../types"
 
 export type NodeDetailsState = {
-    parameters? : Record<string, UIParameter>,
+    parameters? : UIParameter[],
+    typedExpressions?: UITypedExpression[],
```
mproch, line 7: Do we want to show typing results only when user is changing expressions? Or also when they open node modal? If so, we need to pass typedExpressions also in ValidationResult?

mproch (commented):
Look quite nice, I added some comments

ui/client/types/definition.ts
nadberezny: Yes, you're right, it should be defined this way:
```
TypingResult = {
    display: string,
    type: string,
    refClazzName?: string,
    params?: TypingResult[]
}
```

We could also type this as a union, but it would entail some cumbersone changes in few places

ui/client/components/modals/ValidationLabels.tsx
nadberezny: 👍

ui/client/components/graph/node-modal/editors/expression/Editor.ts
nadberezny: 👍

ui/client/reducers/nodeDetailsState.ts
nadberezny: Good point, I think it would be nice to display type info when modal is opened and no changes yet applied,
- will add type info to ValidatinResult

ui/client/components/modals/ValidationLabels.tsx
nadberezny: Ah, yes , that's right, I'll change the behaviour to diplay type info when it's correctly inferred, altogheter with errors

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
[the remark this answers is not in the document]
arkadius: How types of parameters are formatted now on FE?
I wonder if current implementation of displaying types in parameters is good enough to use for inferred types or just need to have some little tweaks to be fair enough? I don't want to make both implementations go far away from themself because at the end IMO we would like to have the same representation in both places. WDYT?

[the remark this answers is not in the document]
nadberezny: Ah, ok I see, indeed we have the same functionality implemented differently for parameters:
// NodeDetailsContent.js
```
  renderFieldLabel = (label) => {
    const parameter = this.findParamDefinitionByName(label)
    return (
      <div className="node-label" title={label}>{label}:
        {parameter ?
          <div className="labelFooter">{ProcessUtils.humanReadableType(parameter.typ.refClazzName)}</div> : null}
      </div>
    )
  }
```

I agree, it should be in one place. I'll change that.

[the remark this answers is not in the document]
nadberezny: @arkadius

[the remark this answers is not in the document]
nadberezny: @arkadius 
I've changed how type labels are displayed - now we're using information from BE, instead of `ProcessUtils.humanReadableType` in all places.
https://github.com/TouK/nussknacker/pull/1191/commits/ffab8ab0bcebe973411ff5c60e68e23e65a74bf3

!but, this changes how types are displayed:
Object -> Unknown
List -> List[Unknown]
List -> List[Long]

ui/client/types/definition.ts
nadberezny: @mproch 
https://github.com/TouK/nussknacker/pull/1191/commits/03b42f65903ce8692a19ee6c1a9dc122cb237ce5

ui/client/components/modals/ValidationLabels.tsx
nadberezny: @mproch 
https://github.com/TouK/nussknacker/pull/1191/commits/3c27991002c0d4bf98b865bcf54e0a057a258aa5

ui/client/components/modals/ValidationLabels.tsx
nadberezny: @mproch 
https://github.com/TouK/nussknacker/pull/1191/commits/1c0a0ce863576402dfc9259886c0786b62629dee

ui/client/components/graph/node-modal/editors/expression/Editor.ts
nadberezny: @mproch 
https://github.com/TouK/nussknacker/pull/1191/commits/6f6b606889c4cc90a05b21c6b8c8012a468b061e

ui/client/reducers/nodeDetailsState.ts
nadberezny: @mproch 
https://github.com/TouK/nussknacker/pull/1191/commits/d0abd3e3ead301d1a1b983ec531cf9ddb4025632

ui/client/types/definition.ts
mproch: Maybe some comment about union handling? (that is - that we don't handle it ATM)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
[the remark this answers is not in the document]
arkadius: 1. Now it is concise and it is fine
2. New types doesn't look too much verbose so it is IMO also fine
3. IMO `humanReadableType` method should stay. I would change it to `typeLabel` or sth like that and pass to it whole type info (not only value of display). Thanks to that we could change the implementation in one place. Maybe in the future we could create component rendering this type...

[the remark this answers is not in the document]
nadberezny: @arkadius 
Ad.3 The same thinkgwas suggested by @mproch 
I've restored #humanReadableType
https://github.com/TouK/nussknacker/pull/1191/commits/102900b3a62dfa10743ba361b0d6d3f6c5cf61ac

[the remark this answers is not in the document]
arkadius: Indeed! :+1:

arkadius (approved):
LGTM

ui/client/reducers/nodeDetailsState.ts
nadberezny: @mproch 
It turns out, that TypingResult is used in non-typed js files (non TS),
so I've changed type definition to algebraic type:
https://github.com/TouK/nussknacker/pull/1191/commits/0ee10e1583ce0d0c3562ae268b08d4d45bc791c1

Also TS, has very nice support for narrowing union types in if esle structures and type guard, so it won't be too much hassle to handle it.

PR #1193, opened 2020-09-05
Fix docker model config

nadberezny:


nussknacker-dist/src/universal/conf/base-application.conf
```diff
@@ -82,4 +82,42 @@ flinkEngineConfig {
...
+signalsTopic: ${?SIGNALS_TOPIC}
+
+kafkaAddress: "kafka:9092"
+kafkaAddress: ${?KAFKA_ADDRESS}
+
+// TODO: This won't work when nk is run outside container, it should be possible to add additional url
+schemaRegistryUrl: "http://schemaregistry:8081"
+schemaRegistryUrl: ${?SCHEMA_REGISTRY_URL}
+
+modelConfig {
```
arkadius, line 97: Why all this is reverted? Can u add more info in description of this PR?

PR #1200, opened 2020-09-07
Add ContextId to the list of optional service parameters

rafa-minimal:
The idea is to pass some custom generated context id (correlation id) when invoking standalone process so that the same context id is available when executing services.
    
This way we are able to log inbound and outbound HTTP requests with the same correlation id (context id).

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
@@ -8,7 +8,7 @@ object Context {
 
 }
 
-class ContextId(val id: String) extends AnyRef
+case class ContextId(value: String)
 
 case class Context(id: String, variables: Map[String, Any],
                    lazyContext: LazyContext, parentContext: Option[Context]) {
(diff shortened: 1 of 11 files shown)


arkadius (approved):
LGTM, good job!

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -8,6 +8,8 @@ object Context {
 
 }
 
+class ContextId(val id: String) extends AnyRef
```
arkadius, line 11: `extends AnyRef` doesn't give anything. Do you mean `extends AnyVal`? Scala value classes have some problems, but I don't remember what exactly. Also IMO would be nice to make this serializable and comparable in case of storing it in some map. Regards to that maybe would be better to use `case class`?

rafa-minimal: Thanks. Yes it was supposed to be a value class (`extends AnyVal`). Anyway, I've replaced it with case class.

PR #1207, opened 2020-09-08
Fix: errors for generic join with branch parameters were shown on the top of node

arkadius:


ui/client/components/graph/node-modal/NodeDetailsContent.js
```diff
@@ -652,7 +652,11 @@ export class NodeDetailsContent extends React.Component {
 
   availableFields = () => {
     if (this.props.dynamicParameterDefinitions) {
-      return this.props.dynamicParameterDefinitions.map(param => param.name)
+      const commonFields = ["id", "outputVar"]
+      const paramFields = this.props.dynamicParameterDefinitions.map(param => param.name)
+      const branchParamsFields = this.state?.editedNode?.branchParameters
+        .flatMap(branchParam => branchParam.parameters.map(param => branchErrorFieldName(param.name, branchParam.branchId)))
+      _.concat(commonFields, paramFields, branchParamsFields)
```
mproch, line 659: Don't you miss return here?

ui/client/components/graph/node-modal/NodeDetailsContent.js
```diff
@@ -652,7 +652,11 @@ export class NodeDetailsContent extends React.Component {
 
   availableFields = () => {
     if (this.props.dynamicParameterDefinitions) {
-      return this.props.dynamicParameterDefinitions.map(param => param.name)
+      const commonFields = ["id", "outputVar"]
```
mproch, line 655: can we extract common method with Join handling below?

arkadius: good point, extracted

ui/client/components/graph/node-modal/NodeDetailsContent.js
arkadius: good catch, added return

mproch (approved):

PR #1210, opened 2020-09-08
[Feature] Detecting dictId in Avro Schema

mproch:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -87,6 +88,8 @@ object AvroSchemaTypeDefinitionExtractor {
 
   val ExtendedPossibleTypes: Set[TypedClass] = DefaultPossibleTypes ++ Set(Typed.typedClass[java.util.Map[String, Any]])
 
+  val dictIdProperty = "nussknackerDictionaryId"
```
arkadius, line 91: Can it by `nkDictId`? `nussknacker` and `dictionary` are very bug-friendly words. Also we can have more similar properties and it can be not convenient in the future, WDYT?.

arkadius (commented):
Nice! One little comment added

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
mproch: :+1: I changed that

arkadius (approved):

PR #1211, opened 2020-09-08
fix combining SingleNodeConfig

gadomsky:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
```diff
@@ -83,3 +81,44 @@ object SingleNodeConfig {
...
 }
+
+trait WithSemiGroup[T] { self: T =>
+  def semigroup: Semigroup[T]
+}
+
+private object SemiGroupImplicits {
+  import cats.syntax.semigroup._
+
+  implicit def takeFirstIfNoSemi[T]: Semigroup[T] = Semigroup.instance[T] {
```
mproch, line 92: shouldn't this be resolved with low priority implicits? I think this would be more "idiomatic" ;)

mproch (commented):
Good catch, I added one comment

PR #1214, opened 2020-09-09
[Improvement] Unify extracting Validators and Editors

mproch:
- Editors and Validators are determined in EditorExtractor and ValidatorsExtractor in interpreter
- For "normal" methods they are used in MethodDefinitionExtractor
- For subprocess parameters they are used in UIProcessObjectFactory
- For GenericTransformation and WithExplicitMethodToInvoke current behaviour is retained
- Further work: 
  - use config to override validators/editors in GenericTransformation/WithExplicitMethodToInvoke
  - do some more unification of handling parameters for methods and subprocesses
  - move determining default value also to interpreter and add defaultValue to Parameter


arkadius (approved):
LGTM

PR #1217, opened 2020-09-11
[FIX] Use param names from @ParamName

mproch:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/ParameterExtractor.scala
```diff
@@ -13,13 +13,14 @@ import pl.touk.nussknacker.engine.types.EspTypeUtils
...
   //TODO: extract more logic to be handled by ParameterData etc. so that it can be reused in UIProcessObjectsFactory to determine subprocess data...
-  def extractParameter(p: java.lang.reflect.Parameter, parameterConfig: ParameterConfig): Parameter = {
+  def extractParameter(p: java.lang.reflect.Parameter, nodeConfig: SingleNodeConfig): Parameter = {
     val nodeParamNames = Option(p.getAnnotation(classOf[ParamName]))
       .map(_.value())
     val branchParamName = Option(p.getAnnotation(classOf[BranchParamName]))
       .map(_.value())
     val name = (nodeParamNames orElse branchParamName)
       .getOrElse(throwIllegalArgument(p, isBranch = false, "missing @ParamName or @BranchParamName annotation"))
+    val parameterConfig = nodeConfig.params.flatMap(_.get(name)).getOrElse(ParameterConfig.empty)
```
nadberezny, line 23: we already have helper function for this
SingleNodeConfig
def paramConfig(name: String): ParameterConfig = params.flatMap(_.get(name)).getOrElse(ParameterConfig.empty)

mproch: :+1: thanks, I knew I've seen it somewhere :D

nadberezny (approved):
looks good 👍

PR #1234, opened 2020-09-18
[FIX] Admin/Services parameters work

mproch:


nadberezny (approved):
Looks good, and works fine
👍

PR #1243, opened 2020-09-23
Static TypeInformation based on ValidationContext

mproch:


mproch:
I close this PR in favour of #1338

PR #1255, opened 2020-10-02
Improvements: Some small FE things

lciolecki:


mproch (approved):
Please add note to MigrationGuide/Changelog that Metrics are configured as customtab?

PR #1256, opened 2020-10-05
Fix: metrics tab

lciolecki:


nussknacker-dist/src/universal/conf/dev-application.conf
```diff
@@ -74,7 +74,7 @@ environmentAlert: {
 }
 
 customTabs = [
-  {name: "Metrics", url: ${base.grafanaUrl}, id: "metrics"},
+  {name: "Metrics", url: ${proxyUrl}/metrics, id: "metrics"},
```
mproch, line 77: I think it should be:   {name: "Metrics", url: ${metricsSettings.url}, id: "metrics"},

mproch (approved):
:+1:

PR #1259, opened 2020-10-08
[FIX] use defaults from defaultUiConfig when initializing akka

mproch:


arkadius (approved):

arkadius:
Title of PR is misleading. It looks like it will be used only defaultUiConfig.

PR #1269, opened 2020-10-19
Fix subprocess output context:

nadberezny:
1. Pass parent context for subprocess output fields compilation


mproch (approved):
Let's write a test for that...

PR #1276, opened 2020-10-23
Access job data from custom nodes

jedrz:
Job data is already available from services via RuntimeInjectedJobData trait.


mproch (approved):
:+1:

PR #1277, opened 2020-10-23
improve CaseClassSerializer

gadomsky:


mproch (approved):
Doesn't look nice, but guess we have to live with it...

PR #1288, opened 2020-11-02
[Feautre] Default ObjectNaming using simple namespaces

mproch:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/DefaultNamespacedObjectNaming.scala
```diff
@@ -0,0 +1,69 @@
...
+    regexMap.getOrElse(namespace, s"${namespace}_(.*)".r)
+
+}
+
+case class DefaultNamespacedObjectNamingParameters(originalName: String,
+                                                   namespace: String) extends ObjectNamingParameters {
+  override def toTags: Map[String, String] = {
+    Map(
+      "originalProcessName" -> originalName,
+      "namespace" -> namespace
```
lciolecki, line 66: `NamespacePath` ? We can also move `originalProcessName`  to some const.

lciolecki (approved):

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/namespaces/DefaultNamespacedObjectNaming.scala
mproch: I extracted constants, I used separate one for namespaceTag, as it should not be necessarily the same as config path

PR #1290, opened 2020-11-05
introduce custom serializers provider

gadomsky:


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/util/SerializersProvider.scala
```diff
@@ -0,0 +1,14 @@
...
+import pl.touk.nussknacker.engine.util.loader.ScalaServiceLoader
+
+trait SerializersProvider {
+  def serializers(): List[SerializerWithSpecifiedClass[_]]
+}
+
+object SerializersProvider {
+  def load(): List[SerializerWithSpecifiedClass[_]] = ScalaServiceLoader
+    .load[SerializersProvider](getClass.getClassLoader)
+    .flatMap(_.serializers())
```
lciolecki, line 13: What happend when there is no providers? I mean what about some default?

lciolecki:
Good improvement! Please add some information about Changelog and Documentation ;)

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/util/SerializersProvider.scala
gadomsky: When you provide no implementations ScalaServiceLoader.load() will return empty list so its ok.
Defaults are provided in code as it was before.
If you want to override them, it is also possible, because custom serilizers loaded here are registered afterwards

mproch (approved):
:+1:

PR #1303, opened 2020-11-12
AdditionalInfo for TypedObjectTypingResult

mproch:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -44,7 +45,8 @@ object typing {
 
   }
 
-  case class TypedObjectTypingResult(fields: Map[String, TypingResult], objType: TypedClass) extends SingleTypingResult {
+  case class TypedObjectTypingResult(fields: Map[String, TypingResult],
```
arkadius, line 48: Why not `additionalInfo: Map[String, AdditionalDataValue] = Map.empty`?

mproch: in most cases it will be empty, didn't want to put too much into json... Or you think I should handle it on encoder's level?

arkadius: I haven't idea that it is because of lighter json purpose. IMO if we want to have lighter json we should put this logic to encoder. I agree with you that we should encode empty map as empty there.

mproch: :+1: I fixed it

arkadius (approved):
LGTM

arkadius:
Can u add notice to MigrationGuide?

PR #1304, opened 2020-11-12
More generic SerializersRegistrar API in place of SerializersProvider…

arkadius:
…, moved AvroSerializersRegistrar to avro utils as a service


build.sbt
```diff
@@ -544,7 +544,7 @@ lazy val avroFlinkUtil = (project in engine("flink/avro-util")).
       )
     }
   )
-  .dependsOn(kafkaFlinkUtil, kafkaTestUtil % "test", flinkTestUtil % "test", process % "test")
```
mproch, line 547: maybe we can have SerializersRegistrar in flink-api?

gadomsky (approved):
LGTM

mproch (approved):
:+1: I added one minor coment

build.sbt
arkadius: fixed

PR #1314, opened 2020-11-15
Benchmark for various avro serialization round-trips

arkadius:


engine/benchmarks/src/main/scala/pl/touk/nussknacker/engine/benchmarks/avro/AvroSamples.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.flink.formats.avro.typeutils.LogicalTypesGenericRecordBuilder
+import pl.touk.nussknacker.engine.avro.AvroUtils
+
+object AvroSamples {
+
+  val sampleSchemaId = 123
+
+  private val fieldsCount = 10
+
+  val sampleSchema: Schema = AvroUtils.parseSchema(s"""{
```
mproch, line 14: Can't we use Schema.Builder instead of raw strings? :)

mproch (approved):

engine/benchmarks/src/main/scala/pl/touk/nussknacker/engine/benchmarks/avro/AvroSamples.scala
arkadius: raw string is easier to copy-paste from somewhere :)

PR #1316, opened 2020-11-16
Kafka avro tests refactor: added MockConfluentSchemaRegistryClientFactory...

arkadius:
…+ correctly passing schema id to RuntimeSchemaData


engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/KafkaAvroSpecMixin.scala
```diff
@@ -46,14 +48,20 @@ trait KafkaAvroSpecMixin extends FunSuite with FlinkSpec with KafkaSpec with Mat
   protected def kafkaTopicNamespace: String = getClass.getSimpleName
 
   // schema.registry.url have to be defined even for MockSchemaRegistryClient
```
mproch, line 50: I think this comment should be moved somewhere else ;)

mproch (approved):

PR #1321, opened 2020-11-16
Exception handler accessible via custom node context

jedrz:
Report events that could not be encoded to Avro.


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/compiler/CompiledProcessWithDeps.scala
```diff
@@ -20,7 +21,7 @@ import scala.concurrent.duration.FiniteDuration
 
 class CompiledProcessWithDeps(compiledProcess: CompiledProcess,
                               val jobData: JobData,
-                              val exceptionHandler: FlinkEspExceptionHandler,
+                              exceptionHandler: FlinkEspExceptionHandler,
```
mproch, line 24: I wonder, maybe we should make it () => FlinkEspExceptionHandler? Or add a comment that it has to be opened separately? (open in this clas does not handle it)

mproch (commented):
I added some comments, I also wonder how to make it more comprehensible what and when is/should be opened/closed in CompiledProcessWithDeps - maybe some comments?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSink.scala
```diff
@@ -59,4 +48,31 @@ class KafkaAvroSink(preparedTopic: PreparedKafkaTopic, versionOption: SchemaVers
...
   }
+
+  class EncodeAvroRecordFunction(flinkNodeContext: FlinkCustomNodeContext)
+    extends RichMapFunction[ValueWithContext[KeyedValue[AnyRef, AnyRef]], KeyedValue[AnyRef, AnyRef]] {
+
+    private val nodeId = flinkNodeContext.nodeId
+    private val exceptionHandlerPreparer = flinkNodeContext.exceptionHandlerPreparer
+    private var exceptionHandler: FlinkEspExceptionHandler = _
+
+    override def open(parameters: Configuration): Unit = {
```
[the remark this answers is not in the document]
jedrz: Extracted to helper trait.

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/RecodingExceptionHandler.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.avro
+
+import java.util.concurrent.CopyOnWriteArrayList
+
+import org.apache.flink.api.common.restartstrategy.RestartStrategies
+import pl.touk.nussknacker.engine.api.exception.{EspExceptionInfo, NonTransientException}
+import pl.touk.nussknacker.engine.flink.api.exception.{FlinkEspExceptionConsumer, FlinkEspExceptionHandler}
+import pl.touk.nussknacker.engine.flink.util.exception.ConsumingNonTransientExceptions
+
+class RecodingExceptionHandler extends FlinkEspExceptionHandler
```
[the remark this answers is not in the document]
jedrz: Done.

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/compiler/CompiledProcessWithDeps.scala
```diff
@@ -54,5 +55,13 @@ class CompiledProcessWithDeps(compiledProcess: CompiledProcess,
   val lazyInterpreterDeps: LazyInterpreterDependencies = compiledProcess.lazyInterpreterDeps
 
   val sources: NonEmptyList[PotentiallyStartPart] = compiledProcess.parts.sources
+
+  def restartStrategy: RestartStrategies.RestartStrategyConfiguration = exceptionHandler.restartStrategy
```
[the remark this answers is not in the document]
jedrz: Done.

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/compiler/CompiledProcessWithDeps.scala
jedrz: Done, comment added. `exceptionHandler` so it can be only accessed via `prepareExceptionHandler`.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/exception/WithFlinkEspExceptionHandler.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.flink.api.common.functions.{RichFunction, RuntimeContext}
+import org.apache.flink.configuration.Configuration
+
+/**
+  * Helper for using exception handler.
+  *
+  * Be aware that super.open and super.close are not called.
+  */
+trait WithFlinkEspExceptionHandler {
+  self: RichFunction =>
```
gadomsky, line 12: I wonder if it's a good practice to do something like this?
If you implement multiple traits like this in your RichFunction you will have valid scala code, but actually methods from only one trait will be executed.
So this trait cannot be used with e.g `LazyParameterInterpreterFunction`

mproch: But do we have better way to do this?

jedrz: That's why I added a comment that open and close from base class are never called.

gadomsky: If there is no better way, I guess I will have to live with it :)

mproch (approved):

PR #1331, opened 2020-11-20
Caching loggers for LoggingListener

mproch:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/LoggingListener.scala
```diff
@@ -1,5 +1,8 @@
 package pl.touk.nussknacker.engine.util
 
+import java.util.concurrent.ConcurrentHashMap
```
arkadius, line 3: Please add Changelog entry

arkadius (approved):

PR #1336, opened 2020-11-23
[Improvement] Aggregate metrics

mproch:


engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -56,6 +86,7 @@ trait AggregatorFunctionMixin { self: StateHolder[TreeMap[Long, AnyRef]] =>
 
     updateState(newState, newElementInStateTimestamp + timeWindowLengthMillis, timeService)
     handleElementAddedToState(newElementInStateTimestamp, newElement, value.context, timeService, out)
+    sizeHistogram.update(newState.size)
```
arkadius, line 89: I think that we should at least add a comment that is not valid metric. Someone could think that it holds histogram of current snapshot of states for all keys, but in fact it is biased by keys that update state more frequently then other ones. Also this histogram won't come down to zero even if all states will be evictated. 
Maybe better approach would be to add extra state value `timestampOfLastHistogramUpdate` and make sure that is updated in some interval and also add `update` invocation to `evictStates`?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -42,9 +70,11 @@ trait AggregatorFunctionMixin { self: StateHolder[TreeMap[Long, AnyRef]] =>
   protected def handleNewElementAdded(value: ValueWithContext[StringKeyedValue[AnyRef]],
                                       timestamp: Long, timeService: TimerService,
                                       out: Collector[ValueWithContext[AnyRef]]): Unit = {
+    val start = System.nanoTime()
     val newState = addElementToState(value, timestamp, timeService, out)
     val finalVal = computeFinalValue(newState, timestamp)
     out.collect(ValueWithContext(finalVal, value.context))
+    timeHistogram.update(System.nanoTime() - start)
```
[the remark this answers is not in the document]
mproch: :+1:

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
mproch: I added comment and changed name, to reflect that I'm interested in how many buckets were deserialized. 
It will come to zero after 10 secs of no events ;)

arkadius (approved):

PR #1337, opened 2020-11-23
Omitting zeros for sum aggregate to avoid unnecessary timers registrations

arkadius:


engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -50,13 +50,20 @@ trait AggregatorFunctionMixin { self: StateHolder[TreeMap[Long, AnyRef]] =>
...
+    if (aggregator.isNeutralForAccumulator(newElement)) {
+      stateForTimestampToSave(readStateOrInitial(), newElementInStateTimestamp)
+    } else {
+      val newState = computeNewState(newElementInStateTimestamp, newElement)
 
-    updateState(newState, newElementInStateTimestamp + timeWindowLengthMillis, timeService)
-    handleElementAddedToState(newElementInStateTimestamp, newElement, value.context, timeService, out)
-    newState
+      updateState(newState, newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+      handleElementAddedToState(newElementInStateTimestamp, newElement, value.context, timeService, out)
```
mproch, line 64: is it obvious that handleElementAddedToState is not called? Looking at implementations I'm not 100% sure...

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -50,13 +50,20 @@ trait AggregatorFunctionMixin { self: StateHolder[TreeMap[Long, AnyRef]] =>
...
                                   out: Collector[ValueWithContext[AnyRef]]): TreeMap[Long, aggregator.Aggregate] = {
-    val newElementInStateTimestamp = computeTimestampToStore(timestamp)
     val newElement = value.value.value.asInstanceOf[aggregator.Element]
-    val newState = computeNewState(newElementInStateTimestamp, newElement)
+    val newElementInStateTimestamp = computeTimestampToStore(timestamp)
+
+    // For neutral element, we do not create bucket, add to it element, and move timers to avoid unnecessary buckets
+    // in our state and unnecessary timers registration
+    if (aggregator.isNeutralForAccumulator(newElement)) {
+      stateForTimestampToSave(readStateOrInitial(), newElementInStateTimestamp)
```
mproch, line 59: in computeNewState we also remove some older buckets. Here we don't do it, if there is a period of zeroes after longer period of non-zeros, it will make state larger than necessary?

arkadius:
Closing, because we do not want that this optimization will complicate other scenarios.

PR #1343, opened 2020-12-04
Aggregator computes stored type

mproch:


engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -102,27 +107,28 @@ object aggregates {
...
-    override type Aggregate = AnyRef
+    override type Aggregate = Option[AnyRef]
 
     override type Element = AnyRef
 
-    override def zero: Aggregate = Marker
+    override def zero: Aggregate = None
 
-    override def addElement(el: Element, agg: Aggregate): Aggregate = if (agg == zero) el else agg
+    override def addElement(el: Element, agg: Aggregate): Aggregate = if (agg.isEmpty) Some(el) else agg
```
arkadius, line 123: It will break state for this aggregator, can you add some notice to migrationguide?

mproch: :+1:

arkadius (approved):

PR #1346, opened 2020-12-07
FlinkRangeMap type class - to be able to switch map state impl

arkadius:


mproch (approved):
:+1: let's try this approach :)

PR #1350, opened 2020-12-15
Higher default precision for HyperLogLog, some tests

mproch:


engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/HyperLogLogPlusAggregatorSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    val higherPrecision = HyperLogLogPlusAggregator(10, 15)
+    runForData(higherPrecision, set1) shouldBe (20, 47)
+    runForData(higherPrecision, set2) shouldBe (100, 200)
+    runForData(higherPrecision, set3) shouldBe (1033, 693)
+
+  }
+
+  private def generateRandomData(unique: Int, randomRepeat: Int) = {
+    Random.shuffle((1 to unique).flatMap(k => (0 to Random.nextInt(randomRepeat)).map(_ => k.toString)))
```
arkadius, line 30: 1. Why you use strings instead of ints?
2. Why you need to use  `Random.nextInt` instead of fixed numer of repeats or even one repeat?

engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/HyperLogLogPlusAggregatorSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  private def runForData(agg: HyperLogLogPlusAggregator, data: Iterable[AnyRef]): (Long, Int) = {
+    val total = data.foldRight(agg.zero)(agg.addElement)
+    (agg.result(total), total.wrapped.getBytes.length)
```
arkadius, line 35: Can you extract local variable better naming what is hidden in `total.wrapped.getBytes.length`? It is memory used by hyper log log plus?

mproch: done

engine/flink/model-util/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/HyperLogLogPlusAggregatorSpec.scala
mproch: I added comment, shuffling and number of occurrences should not matter - added just to check that ;)

arkadius (approved):

PR #1352, opened 2020-12-17
introduce possibility to represent Avro type string as Java String

gadomsky:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.avro.specific.{SpecificData, SpecificDatumReader}
+import org.apache.avro.util.Utf8
+
+import scala.util.{Properties, Try}
+
+trait StringForcingDatumReader[T] extends GenericDatumReader[T]  {
+  self: GenericDatumReader[T] =>
+
+  override def readString(old: Any, expected: Schema, in: Decoder): AnyRef = super.readString(old, expected, in) match {
+    case data: Utf8 if AvroStringSettings.forceUsingStringForStringSchema => data.toString
```
arkadius, line 17: why it is still needed if original implementation already check result of `findStringClass`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  lazy val forceUsingStringForStringSchema: Boolean = Properties.envOrNone(envName)
+    .map(str => Try(str.toBoolean).getOrElse {
+      logger.warn(s"$envName is not valid boolean value. Using default $default value")
+      default
+    }).getOrElse(default)
+}
+
+object StringForcingDatumReader {
+  def forGenericDatumReader[T](writer: Schema, reader: Schema, data: GenericData): GenericDatumReader[T] =
```
arkadius, line 39: Maybe we can just skip "for" in name? Because "string forcing datum reader for generic datum reader" looks weird. Maybe just `StringForcingDatumReader.genericDatumReader`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
```diff
@@ -0,0 +1,47 @@
...
+    else super.findStringClass(schema)
+  }
+}
+
+object AvroStringSettings extends LazyLogging {
+  val default: Boolean = false
+  val envName = "AVRO_USE_STRING_FOR_STRING_TYPE"
+
+  lazy val forceUsingStringForStringSchema: Boolean = Properties.envOrNone(envName)
+    .map(str => Try(str.toBoolean).getOrElse {
```
arkadius, line 32: Wdyt about failing fast here? It is case when smb has typo in boolean. I think that he/she want to know it ASAP instead wait until smb use string field incorrectly.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
gadomsky: i'ts not needed anymore, I forgot to clean it up :)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
gadomsky: you're rigth it really sound wird - changed

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReader.scala
gadomsky: good idea - changed

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReaderSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+import org.apache.avro.generic.{GenericDatumWriter, GenericRecord}
+import org.apache.avro.io.{DecoderFactory, EncoderFactory}
+import org.apache.avro.Schema
+import org.scalatest.{EitherValues, FunSpec, Matchers}
+import pl.touk.nussknacker.engine.avro.{AvroStringSettingsInTests, AvroUtils}
+import pl.touk.nussknacker.engine.avro.encode.{BestEffortAvroEncoder, ValidationMode}
+
+
+class StringForcingDatumReaderSpec extends FunSpec with Matchers with EitherValues {
```
arkadius, line 12: It would be nice to see here not only encoding records via `BestEffortAvroEncoder` but also creating it from scratch using `new LogicalTypesGenericRecordBuilder(schema).set("foo", new Utf8(...)) ... / "")` with corner cases like: schema with default string field but not filled during building records. Will it work correctly? Or `new GenericRecord(otherRecord, deepCopy = true)` <- AFAIK it won't work and will be nice to show it in tests. Also would be nice to see what is inside record after encoding - even in round trips, I mean: `encodedRecord.get("foo") should be instanceOf[String]`

arkadius (approved):
LGTM, I've added one "nice to have" comment

engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/schema/StringForcingDatumReaderSpec.scala
gadomsky: `BestEffortAvroEncoder` was here only to create GenericRecord, actually it is not imporant how we create it. What's important here is reading it by DatumReader after serialization, and that is what I test.
I agree on the other hand, that it is worth writing test for provding default value from schema by LogicalTypesGenericRecordBuilder - it acctualy does use DatumReader (and in consequent StringForcingDatumReader)

`GenericData.Record::new(deepCopy=true)` also does not use DatumReader underneath, so I believe this change doesnt affect this operation

PR #1357, opened 2020-12-23
Pass run mode to nodes

jedrz:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
```diff
@@ -0,0 +1,24 @@
...
+  * - Test - process is run in test mode to collect results for test data.
+  * - Verification - process is run to verify compatibility e.g. with state of running process in older version.
+  * - ServiceQuery - one of the services is queried by Nussknacker.
+  * - Definition - process config creator is used extract component definitions. Components should not be invoked.
+  */
+sealed trait RunMode
+
+object RunMode {
+
+  case object Engine extends RunMode
```
arkadius, line 15: WDYT about changing the name to "Normal"? `Engine` is not exactly correct because tests are usually run on the same engine.

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompiler.scala
```diff
@@ -46,7 +46,9 @@ class FlinkProcessCompiler(creator: ProcessConfigCreator,
   def compileProcess(process: EspProcess,
                      processVersion: ProcessVersion,
                      deploymentData: DeploymentData,
-                     resultCollector: ResultCollector)(userCodeClassLoader: ClassLoader): FlinkProcessCompilerData = {
+                     resultCollector: ResultCollector,
+                     runMode: RunMode)
```
arkadius, line 50: It is sad that we need to pass this runMode during compilation :(. It should be done only in interpreter but to make it happen we should do a lot rework in definition / implementation split and proposed approach is the best for now.

arkadius (approved):
LGTM but maybe @mproch could take a look also?

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
```diff
@@ -17,7 +18,8 @@ import scala.util.control.NonFatal
 
 private[registrar] class SyncInterpretationFunction(val compiledProcessWithDepsProvider: ClassLoader => FlinkProcessCompilerData,
                                                     val node: SplittedNode[_<:NodeData],
-                                                    validationContext: ValidationContext, useIOMonad: Boolean)
+                                                    validationContext: ValidationContext, useIOMonad: Boolean,
+                                                    runMode: RunMode)
```
mproch, line 22: Let's use the one from compiledProcessWithDeps?

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerData.scala
```diff
@@ -33,7 +33,8 @@ class FlinkProcessCompilerData(compiledProcess: ProcessCompilerData,
                                exceptionHandler: FlinkEspExceptionHandler,
                                val signalSenders: FlinkProcessSignalSenderProvider,
                                val asyncExecutionContextPreparer: AsyncExecutionContextPreparer,
-                               val processTimeout: FiniteDuration
+                               val processTimeout: FiniteDuration,
+                               runMode: RunMode
```
mproch, line 37: Let's put val here instead of FlinkProcessCompiler?

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -284,7 +287,7 @@ object FlinkProcessRegistrar {
       ScalaServiceLoader.load[FlinkCompatibilityProvider](getClass.getClassLoader)
         .headOption.map(_.createExecutionEnvPreparer(config, prepareExecutionConfig, compiler.diskStateBackendSupport))
         .getOrElse(new DefaultStreamExecutionEnvPreparer(checkpointConfig, rocksDBStateBackendConfig, prepareExecutionConfig))
-    new FlinkProcessRegistrar(compiler.compileProcess, defaultStreamExecutionEnvPreparer, eventTimeMetricDuration)
+    new FlinkProcessRegistrar(compiler.compileProcess, defaultStreamExecutionEnvPreparer, eventTimeMetricDuration, compiler.runMode)
```
mproch, line 290: Let's use runMode directly from FlinkProcessCompilerData

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
```diff
@@ -0,0 +1,18 @@
+package pl.touk.nussknacker.engine.api.process
+
+/**
+  * Specifies the mode a custom node is invoked. It can be one of the following values:
+  * <ul>
+  * <li>Normal - process is executed inside Flink or other engine. Real implementation should be used.</li>
```
mproch, line 6: Let's add comment that this mode is also used for validations, test data generations etc.

mproch (approved):
:+1: Added one documentation comment

PR #1359, opened 2020-12-28
Improvements: Variable output validation

lciolecki:
* Add validation variable name
* Changes in presentation error message

Before:
output-1
After:
output-2
output-3

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -13,6 +13,7 @@ Nussknacker versions
 * More precise TypeInformation generation
     * [#1338](https://github.com/TouK/nussknacker/pull/1338) Defining TypeInformation based on TypingResult
     * [#1343](https://github.com/TouK/nussknacker/pull/1343) Aggregators compute stored types    
+    * [#1343](https://github.com/TouK/nussknacker/pull/1359) Improvements in variable output validation 
 
 
 0.3.1 (not released yet) 
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
@@ -180,12 +180,19 @@ object ProcessCompilationError {
       InvalidPropertyFixedValue(paramName, label, value, values, nodeId.id)
   }
 
-  case class OverwrittenVariable(variableName: String, nodeId: String)
+  case class OverwrittenVariable(variableName: String, nodeId: String, paramName: Option[String])
     extends PartSubGraphCompilationError with InASingleNode
 
   object OverwrittenVariable {
-    def apply(variableName: String)(implicit nodeId: NodeId): PartSubGraphCompilationError =
-      OverwrittenVariable(variableName, nodeId.id)
+    def apply(variableName: String)(implicit nodeId: NodeId, paramName: Option[String] = None): PartSubGraphCompilationError =
+      OverwrittenVariable(variableName, nodeId.id, paramName)
+  }
+
+  case class InvalidVariableOutputName(name: String, nodeId: String, paramName: Option[String]) extends PartSubGraphCompilationError with InASingleNode
+
+  object InvalidVariableOutputName {
+    def apply(variableName: String)(implicit nodeId: NodeId, paramName: Option[String]): PartSubGraphCompilationError =
+      InvalidVariableOutputName(variableName, nodeId.id, paramName)
   }
 
   case class NoParentContext(nodeId: String) extends PartSubGraphCompilationError with InASingleNode
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
@@ -1,17 +1,19 @@
 package pl.touk.nussknacker.engine.api.context
-
 import cats.data.Validated.{Invalid, Valid}
 import cats.data._
-import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{NoParentContext, NodeId, OverwrittenVariable}
+import cats.implicits._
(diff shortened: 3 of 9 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/PrettyValidationErrors.scala
```diff
@@ -47,7 +47,8 @@ object PrettyValidationErrors {
       case RedundantParameters(params, _) => node(s"Redundant parameters", s"Please omit redundant parameters: ${params.mkString(", ")}")
       case WrongParameters(requiredParameters, passedParameters, _) =>
         node(s"Wrong parameters", s"Please provide ${requiredParameters.mkString(", ")} instead of ${passedParameters.mkString(", ")}")
-      case OverwrittenVariable(varName, _) => node(s"Variable $varName is already defined", "You cannot overwrite variables")
+      case OverwrittenVariable(varName, _, paramName) => node(s"Variable $varName is already defined", "You cannot overwrite variables", fieldName = paramName)
+      case InvalidVariableOutputName(varName, _, paramName) => node(s"Variable $varName is not valid.", "Please use only letters, numbers or '_'.", fieldName = paramName)
```
mproch, line 51: "Variable name $varName is not valid" -> I think it would be more clear

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
```diff
@@ -26,11 +30,15 @@ case class ValidationContext(localVariables: Map[String, TypingResult] = Map.emp
   def contains(name: String): Boolean = variables.contains(name)
 
   def withVariable(name: String, value: TypingResult)
-                  (implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, ValidationContext] =
-    if (variables.contains(name))
+                  (implicit nodeId: NodeId, paramName: Option[String] = None): ValidatedNel[PartSubGraphCompilationError, ValidationContext] = {
+    if (variables.contains(name)) {
       Invalid(OverwrittenVariable(name)).toValidatedNel
-    else
+    } else if (!VariableNamePattern.matcher(name).matches()) {
```
mproch, line 36: I think this else if is misleading, let's make those validations independently and then just combine errors... (I know that if variable is defined, it name will be valid, just for sake of readability)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
```diff
@@ -15,6 +17,8 @@ case class ValidationContext(localVariables: Map[String, TypingResult] = Map.emp
                              globalVariables: Map[String, TypingResult] = Map.empty,
                              parent: Option[ValidationContext] = None) {
 
+  val VariableNamePattern: Pattern = "^[a-zA-Z0-9_]*$".r.pattern
```
mproch, line 20: Maybe we should use javax.lang.model.SourceVersion.isIdentifier? e.g. identifiers cannot start with digits?

mproch (commented):
:+1: good idea, I added some small comments

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
lciolecki: You're right - done.

mproch (approved):

docs/Changelog.md
```diff
@@ -13,6 +13,7 @@ Nussknacker versions
 * More precise TypeInformation generation
     * [#1338](https://github.com/TouK/nussknacker/pull/1338) Defining TypeInformation based on TypingResult
     * [#1343](https://github.com/TouK/nussknacker/pull/1343) Aggregators compute stored types    
+    * [#1343](https://github.com/TouK/nussknacker/pull/1359) Improvements variable output validation
```
nadberezny, line 16: ..typo

mprovements **in** variable output validation 
or 
Improve variable output validation

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -180,12 +180,19 @@ object ProcessCompilationError {
       InvalidPropertyFixedValue(paramName, label, value, values, nodeId.id)
   }
 
-  case class OverwrittenVariable(variableName: String, nodeId: String)
+  case class OverwrittenVariable(variableName: String, nodeId: String, paramName: Option[String])
```
nadberezny, line 183: imo fieldName would be better,

also I'm wondering if we shouldn't have a separate PartSubGraphCompilationError for a case when we override field names, like OverwrittenFieldName

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
```diff
@@ -26,23 +28,29 @@ case class ValidationContext(localVariables: Map[String, TypingResult] = Map.emp
...
   def withVariable(name: String, value: TypingResult)
-                  (implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, ValidationContext] =
-    if (variables.contains(name))
-      Invalid(OverwrittenVariable(name)).toValidatedNel
-    else
-      Valid(copy(localVariables = localVariables + (name -> value)))
+                  (implicit nodeId: NodeId, paramName: Option[String] = None): ValidatedNel[PartSubGraphCompilationError, ValidationContext] = {
+
+    List(validateVariableExists(name), validateVariableFormat(name))
+      .sequence_
```
nadberezny, line 34: Traversable.sequence can be used instead of NestedFoldableOps.sequence_

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
nadberezny: also I'm not sure if making paramName (aka fieldName) implicit is a good idea
this is not a sort of dependency that you pass over?
wdyf?

PR #1363, opened 2020-12-30
Open needed services

mproch:


engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/registrar/AsyncInterpretationFunction.scala
```diff
@@ -35,6 +36,9 @@ private[registrar] class AsyncInterpretationFunction(val compiledProcessWithDeps
       getRuntimeContext.getExecutionConfig.getParallelism)
   }
 
+
+  override def nodesUsed: List[NodeData] = SplittedNodesCollector.collectNodes(node).map(_.data)
```
arkadius, line 40: I know that it is not a part of this change, but can you add a comment in `SplittedNodesCollector.collectNodes` that this logic of collecting should be the same as in `ProcessSplitter`? Also maybe we should move this collecting logic to splitter to make sure that those things will stay consistent?

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/WithCompiledProcessDeps.scala
```diff
@@ -3,26 +3,29 @@ package pl.touk.nussknacker.engine.process
...
 import org.apache.flink.configuration.Configuration
 import pl.touk.nussknacker.engine.flink.api.exception.FlinkEspExceptionHandler
+import pl.touk.nussknacker.engine.graph.node.NodeData
 import pl.touk.nussknacker.engine.process.compiler.CompiledProcessWithDeps
 
 trait WithCompiledProcessDeps extends RichFunction {
 
   def compiledProcessWithDepsProvider: ClassLoader => CompiledProcessWithDeps
 
+  def nodesUsed: List[NodeData]
```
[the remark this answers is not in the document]
mproch: :+1: I refactored it a bit

engine/flink/process/src/main/scala/pl/touk/nussknacker/engine/process/registrar/AsyncInterpretationFunction.scala
mproch: I added comment for now

arkadius (approved):
LGTM

PR #1365, opened 2020-12-30
[Improvement] store compiled classes during build

mproch:


.github/workflows/pr.yml
```diff
@@ -6,11 +6,44 @@ on:
...
+        with:
+          name: build-target
+          path: target.tgz
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    needs: [build]
     steps:
       - uses: actions/checkout@v1
+      - uses: olafurpg/setup-scala@v10
```
lciolecki, line 40: Hmm, I wonder can we move `olafurpg/setup-scala@v10` to some var? Right now use it in a few places?

lciolecki (approved):

.github/workflows/pr.yml
mproch: I hope we'll manage to extract some common stuff in next PRs...

PR #1368, opened 2021-01-04
[Build] Publish via githubactions - first version

mproch:
Things that are left for future PRs:
- computing coverage
- adding devModel for staging
- cross publishing
- building FE only once and caching artifacts

Refactor of workflows:
- GithubActions do not handle composite actions or other form of reusing parts of workflow definition (e.g. no support for YAML anchors...). There is composite run step: https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action, but one can only use "run" as steps and not other actions

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,9 +1,15 @@
-name: PR Build
+name: "Build and Test"
 on:
   pull_request:
     branches:
       - master
       - staging
+  push:
+    branches:
+      - master
+      - staging
+env:
+  CROSS_BUILD: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || startsWith(github.head_ref, 'preview') }}
 
 jobs:
   build:
@@ -105,7 +111,7 @@ jobs:
       - name: Integration tests
         shell: bash
         run: ./ciRunSbt.sh flinkProcessManager/it:test engineStandalone/it:test
-  slowTest:
+  slowTests:
     name: Slow tests
     runs-on: ubuntu-latest
     needs: [ build ]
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
-        run: cd ./demo/docker && echo "NUSSKNACKER_VERSION=$GITHUB_SHA" > .env && ./testQuickstart.sh
+        run: cd ./demo/docker && echo "NUSSKNACKER_VERSION=ghactionstest-$GITHUB_SHA" > .env && ./testQuickstart.sh
+
+#TODO: extract to different workflow?
+  publish:
+    runs-on: ubuntu-latest
+    needs: [build, tests, crossCompile, integrationTests, slowTests, frontendTests, dockerTest]
+    if: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || startsWith(github.head_ref, 'preview') ||  github.head_ref == 'publish_githubactions' }}
+    steps:
+      - name: debug
```
lciolecki, line 199: Do we need this debug?

.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
-      - name: setup node
-        shell: bash
-        run: (cd ui/client; npm ci)
       - name: Prepare docker
         shell: bash
         run: echo "version in ThisBuild := \"$GITHUB_SHA\"" > version.sbt && ./ciRunSbt.sh dist/docker:publishLocal
       - name: Test docker
         shell: bash
-        run: cd ./demo/docker && echo "NUSSKNACKER_VERSION=$GITHUB_SHA" > .env && ./testQuickstart.sh
+        run: cd ./demo/docker && echo "NUSSKNACKER_VERSION=ghactionstest-$GITHUB_SHA" > .env && ./testQuickstart.sh
```
lciolecki, line 191: Any env from  file `.env` is used only when specific env is not declared. So enough will be only declare `NUSSKNACKER_VERSION=ghactionstest-$GITHUB_SHA`.

.github/workflows/pr.yml
```diff
@@ -1,9 +1,15 @@
...
   pull_request:
     branches:
       - master
       - staging
+  push:
+    branches:
+      - master
+      - staging
+env:
+  CROSS_BUILD: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' }}
```
lciolecki, line 12: As I understand: we want do cross build only when we merge our branch to master or staging?

lciolecki: I think there should be also checking preview branch like below - maybe we can move this piece of code to separate fragment?

.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "$GITHUB_CONTEXT"
+
+      - uses: actions/checkout@v1
+      - name: Cache npm
+        uses: actions/cache@v2
+        with:
+          path: |
+            **/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
```
lciolecki, line 210: Nice trick ;)

.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build
+        shell: bash
+        env:
+          #TODO: shouldn't be needed...
+          nexusUrl: https://oss.sonatype.org/content/repositories/snapshots
```
lciolecki, line 241: Can we declare nexurUrl as secret same us user / password?  Or at least as env.

.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
+          nexusUser: ${{ secrets.SONATYPE_USER }}
+          nexusPassword: ${{ secrets.SONATYPE_PASSWORD }}
+        #TODO: handle version better
+        run: sbt -J-Xms1000M -J-Xmx1000M "set version in ThisBuild := (version in ThisBuild).value.replace(\"-SNAPSHOT\", \"-$(date -I)-$GITHUB_RUN_NUMBER-$GITHUB_SHA-ghactionstest-SNAPSHOT\")" publish dist/docker:publish
```
lciolecki, line 245: Yeap, it's ugly.. But for now we can leave it.

build.sbt
```diff
@@ -18,28 +18,34 @@ lazy val supportedScalaVersions = List(scala212, scala211)
 val silencerV_2_12 = "1.6.0"
 val silencerV = "1.7.0"
 
+def propOrEnv(name: String, default: String = null) = Option(System.getProperty(name)).orElse(sys.env.get(name)).getOrElse(default)
```
lciolecki, line 21: `propOrEnv(name).getOrElse(default)`

build.sbt
```diff
@@ -18,28 +18,34 @@ lazy val supportedScalaVersions = List(scala212, scala211)
...
+def propOrEnv(name: String) = Option(System.getProperty(name)).orElse(sys.env.get(name))
```
lciolecki, line 22: Hmm.. How does it work? For example: it looks for `includeFlinkAndScala` in evns? I think it's ugly..  I mean when we run ` propOrEnv("includeFlinkAndScala")` we should look for property `includeFlinkAndScala` or env `INCLUDE_FLINK_AND_SCALA`. But I didn't look to method  `sys.env.get`

.github/workflows/pr.yml
```diff
@@ -177,12 +183,64 @@ jobs:
...
+          path: |
+            ~/.ivy2/cache
+            ~/.sbt
+          key: ${{ runner.os }}-ivy2-${{ hashFiles('**/*.sbt') }}
+          restore-keys: ${{ runner.os }}-sbt
+      - uses: olafurpg/setup-scala@v10
+      - uses: actions/download-artifact@v2
+        with:
+          name: build-target
+      - name: Untar artifacts
```
lciolecki, line 229: Do we really need it?

.github/workflows/pr.yml
mproch: what do you mean "as env"? It's in env now :) I think it should be just default

lciolecki: I mean: `${{ secrets.SONATYPE_URL }}`

.github/workflows/pr.yml
mproch: was just following tutorials :)

.github/workflows/pr.yml
mproch: :+1:  true, thanks

.github/workflows/pr.yml
mproch: noo, removed :+1:

.github/workflows/pr.yml
mproch: :)

build.sbt
mproch: I added TODO to get rid of properties and switch to ENV

.github/workflows/pr.yml
mproch: well, caching makes build a bit faster, tar is needed because upload files one by one is really slow as there is awful lot of them

build.sbt
mproch: :+1:

lciolecki (approved):
LGTM

.github/workflows/pr.yml
```diff
@@ -1,9 +1,15 @@
...
   pull_request:
     branches:
       - master
       - staging
+  push:
+    branches:
+      - master
+      - staging
+env:
+  CROSS_BUILD: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || startsWith(github.head_ref, 'preview') }}
```
lciolecki, line 12: Right now, you don't use this env. Maybe we can use it in `if` statement at `crossCompile` job ?

mproch: We use it in ciRunSbt.sh - all tests && publishing will be done with "+" then

PR #1373, opened 2021-01-12
Abitility to load custom process config

jedrz:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/ConfigLoader.scala
```diff
@@ -0,0 +1,24 @@
...
+/**
+  * Loads base process config. Resolved config should be available when a process is run
+  * (resolved config is provided to the process by a process manager).
+  *
+  * Config is resolved once when loading model data. Implement this trait when you want to add dynamic configuration
+  * that is fetched from external source, for example OpenAPI specification.
+  *
+  * This trait is optional to implement.
+  */
+trait ConfigLoader {
```
arkadius, line 14: 1. I see a problem that it is not clear what exactly happen during config loading. Part of logic is in `ModelConfigToLoad.load`, part will be here. WDYT about extracting new object e.g. `DefaultModelConfigLoader` which will replace  `ConfigLoader.identity` and will do what is done in `ModelConfigToLoad.load`? If smb want to reuse it in theirs own implementation, he/she can just invoke it in any place.
2. Also IMO we should rename it to `ModelConfigLoader` because we have many configs in our projects and this name is confusing

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/ModelConfigLoader.scala
```diff
@@ -0,0 +1,31 @@
...
+  * This trait is optional to implement. Default implementations complements base config with configuration
+  * available in JAR's resources.
+  */
+trait ModelConfigLoader extends Serializable {
+
+  /**
+    *  Resolves full config used inside [[pl.touk.nussknacker.engine.api.process.ProcessConfigCreator]]. Invoked both
+    *  in NK and running process.
+    */
+  def resolveFullConfig(configPassedInExecution: Config, classLoader: ClassLoader): Config
```
mproch, line 21: I think we should provide default implementation out of the box here. I think it should be very conscious decision, that defaultModelConfig etc. will not be used.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/ModelConfigLoader.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+    * Resolves partial config that is passed in execution during e.g. process deployment on Flink cluster. When running
+    * a process on engine (e.g. on Flink) this config is used to construct full config.
+    *
+    * It is used for performance reasons to reduce serialized configuration size inside deployed processes.
+    */
+  def resolveConfigPassedInExecution(baseModelConfig: Config, classLoader: ClassLoader): Config
```
mproch, line 29: I have two problems with this method:
- I don't understand name. maybe sth along lines resolveConfigDuringModelAnalysis and resolveConfigDuringExecution?
- here baseModelConfig does not contain defaultModelConfig etc.? I think it makes it more difficult to use and error prone... I mean, for me in 99% cases I will want to use defaults during resolution. Now it's not clear how to do it...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
```diff
@@ -0,0 +1,63 @@
...
+    * <ol>
+    * <li> param configPassedInExecution resolved by [[resolveConfigToPassInExecution]], by default part of main config
+    * file of nussknacker (in path processTypes.{type_name}.modelConfig)</li>
+    * <li>defaultModelConfig.conf from model jar</li>
+    * <li>reference.conf from model jar</li
+    * </ol
+    *
+    * Default implementation should rather not be changed.
+    */
+  def resolveConfigDuringExecution(configPassedInExecution: Config, classLoader: ClassLoader): Config = {
```
arkadius, line 51: Can you extract new protected method e.g. `resolveConfigUsingDefaults` which will be invoked by this method? IMO it will make the usage of this method more easy to understand in concrete implementations.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
```diff
@@ -0,0 +1,63 @@
...
+    *
+    * Method used for performance reasons to reduce serialized configuration size inside deployed processes. By default
+    * config from main nussknacker file at path: processTypes.{type_name}.modelConfig is used.
+    *
+    * @param configDuringModelAnalysis default config resolved by [[resolveConfigDuringExecution]] for configuration
+    *                                  from processTypes.{type_name}.modelConfig
+    * @return transformation of config part (processTypes.{type_name}.modelConfig) that is later passed to a running
+    *         process (see e.g. FlinkProcessCompiler, ModelConfigToLoad).
+    */
+  def resolveConfigToPassInExecution(configDuringModelAnalysis: Config, classLoader: ClassLoader): ConfigToPassInExecution
```
arkadius, line 35: Why this method return class with `transform` instead just return result of `transform` or eagerly wrapped result of transformation? I can't read it from documentation :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigToLoad.scala
```diff
@@ -0,0 +1,17 @@
...
+import com.typesafe.config.Config
+
+/**
+  * Used by a running process on Flink to recreate full process config using only config passed during process
+  * deployment.
+  *
+  * This class with config is serialized in Flink graph so we separate it from the rest of configs to keep it
+  * small (see e.g. FlinkProcessCompiler), as it's only part that has to be passed to Flink.
+  */
+case class ModelConfigToLoad(inputConfig: Config, modelConfigLoader: ModelConfigLoader) {
```
arkadius, line 12: Why this class still exists? Why not to `loadConfig` directly in `ModelData`? Because of backward compatibility or there are some other (serialization?) reasons?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
```diff
@@ -84,9 +93,21 @@ trait ModelData extends AutoCloseable {
...
-  def processConfigFromConfiguration: ModelConfigToLoad
+  protected def inputConfig: Config
+
+  protected def modelConfigLoader: ModelConfigLoader
+
+  lazy val processConfig: Config = modelConfigLoader.resolveConfigDuringExecution(configToPassInExecution, modelClassLoader.classLoader)
+
+  // Config to pass in execution (see FlinkProcessManager).
+  private lazy val configToPassInExecution: Config = {
+    val defaultConfigDuringExecution = modelConfigLoader.resolveConfigDuringExecution(inputConfig, modelClassLoader.classLoader)
```
arkadius, line 104: I'm a little bit supprised that we here firstly invoke `resolveConfigDuringExecution` and then `resolveConfigToPassInExecution`. because we are not yet "during execution". And am I right that during execution config passed in execution is resolved one more time using `resolveConfigDuringExecution`?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
jedrz: The main reason is to distinguish which config can be modified since `configDuringModelAnalysis` should be read only and the one passed inside `transform` can be only changed.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigToLoad.scala
jedrz: Yes, we can get rid of it. It was left because of serialization reasons though I see now that serialVersionUID has been changed. Anyway I need to check if a Flink job can start from a previous savepoint. If not then the only way not to break compatiblity would be leaving `ModelConfigToLoad` as is and removing `ModelConfigLoader.resolveConfigDuringExecution` method.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
jedrz: `resolveConfigDuringExecution` is invoked twice because:
1. `defaultConfigDuringExecution` (I will rename it to `configDuringModelAnalysis`) is `processTypes.{type_name}.modelConfig` + `defaultModelConfig.conf` + `reference.conf`.
2. In order to modify config send to Flink process (`processTypes.{type_name}.modelConfig`) we would like to have access to all the config above. Some urls/paths/etc. can be defined in resources that are possibly needed to modify config.

arkadius: Why not to: 
1. invoke here only `resolveConfigToPassInExecution` which by default will resolve this by merging `processTypes.{type_name}.modelConfig + defaultModelConfig.conf + reference.conf`
2. after that do step 3.
Result will be the same but it will be more straightforward: only two steps,  neither of them done more then once

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
arkadius: But typesafe config is immutable....

jedrz: `ConfigToPassInExecution` has been replaced with a value class that is used in the second step to resolve final config.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
jedrz: Currently all the methods are called only once. Since `resolveConfigUsingDefaults` is extracted I think it makes it more clear.

docs/MigrationGuide.md
```diff
@@ -16,8 +16,8 @@ To see biggest differences please consult the [changelog](Changelog.md).
   and also requires restarting TaskManager when changing the value.
 * [#1361](https://github.com/TouK/nussknacker/pull/1361) Lazy variables are removed, you should use standard enrichers for those cases.
   Their handling has been source of many problems and they made it harder to reason about the exeuction of process.   
-* [#1373](https://github.com/TouK/nussknacker/pull/1373) Use plain config to create `ClassLoaderModelData`,
-  wrapping with ModelConfigToLoad by yourself is not needed.
+* [#1373](https://github.com/TouK/nussknacker/pull/1373) Creating `ClassLoaderModelData` directly is not allowed, use
+  `ModelData.apply` with plain config, wrapping with ModelConfigToLoad by yourself is not needed.
```
arkadius, line 20: This comment is outdated after removal of `ModelConfigToLoad`

arkadius (approved):
LGTM, one minor comment regards do docs left.

docs/MigrationGuide.md
jedrz: I meant that after updating to newer NK versions, wrapping config with `ModelConfigToLoad` is not needed anymore :)

arkadius: ahh indeed. my misunderstanding

mproch (approved):
:+1:

PR #1377, opened 2021-01-15
[Improvement] Can redeploy scheduled job

mproch:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
```diff
@@ -182,9 +183,12 @@ class PeriodicProcessManager(delegate: ProcessManager,
 case class ScheduledStatus(nextRunAt: LocalDateTime) extends CustomStateStatus("SCHEDULED") {
   override def isFollowingDeployAction: Boolean = true
   override def isRunning: Boolean = true
+  override def canDeploy: Boolean = true
```
jedrz, line 186: I think it should be also changed in `PeriodicProcessStateDefinitionManager`

mproch: :+1: thanks for spotting! I decided to remove canDeploy as it's not actually needed...

jedrz (approved):

PR #1379, opened 2021-01-17
Preview/custom actions fe fix

nadberezny:
1. Fix Custom actions button process state
2. Fix ManagementActor StateStatus compare

### DIFF
--- a/ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
+++ b/ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
@@ -26,7 +26,7 @@ export default function CustomActionButton(props: Props) {
       : <DefaultIcon/>
 
   const statusName = processStatus?.name
-  const isDisabled = processStatus === null ? true : !action.allowedStateStatusNames.includes(statusName)
+  const isDisabled = !action.allowedStateStatusNames.includes(statusName)
 
   const toolTip = isDisabled
       ? t("panels.actions.custom-action.tooltips.disabled", "Disabled for {{statusName}} status.", {statusName})
--- a/ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
+++ b/ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
@@ -26,7 +26,7 @@ export default function CustomActionButton(props: Props) {
       : <DefaultIcon/>
 
   const statusName = processStatus.name
-  const isDisabled = !action.allowedProcessStates.map(s => s.name).includes(statusName)
+  const isDisabled = !action.allowedProcessStates.includes(statusName)
 
   const toolTip = isDisabled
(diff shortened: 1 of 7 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -125,7 +125,7 @@ class ManagementActor(managers: ProcessingTypeDataProvider[ProcessManager],
               manager.customActions.find(_.name == actionName) match {
                 case Some(customAction) =>
                   getProcessStatus(id).flatMap {
-                    case Some(status) if customAction.allowedProcessStates.contains(status.status) =>
+                    case Some(status) if customAction.allowedProcessStates.map(_.name).contains(status.status.name) =>
```
mproch, line 128: Let's have only names in allowedProcessStates, with comment that we cannot use full state as it can contain attributes that are changing

nadberezny: :+1:

nadberezny: done:
https://github.com/TouK/nussknacker/pull/1379/commits/4dd43719a61b99daf936b4de63f3ad1d5c798f97

nadberezny: @mproch:
https://github.com/TouK/nussknacker/pull/1379/commits/4b2f08e02fbe9ffaa793b2958828f908ef7a348c

Disable (instead of hiding) CustomActionButton when state is not loaded

ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
```diff
@@ -25,8 +25,8 @@ export default function CustomActionButton(props: Props) {
       ? <img alt={`custom-action-${action.name}`} src={action.icon} />
       : <DefaultIcon/>
 
-  const statusName = processStatus.name
-  const isDisabled = !action.allowedProcessStates.map(s => s.name).includes(statusName)
+  const statusName = processStatus?.name
+  const isDisabled = processStatus === null ? true : !action.allowedStateStatusNames.includes(statusName)
```
mproch, line 29: is it needed? I mean, action.allowedStateStatusNames.includes(null) is false? or not? :)

mproch (approved):
I added minor question, but overall LGTM :+1:

ui/client/components/toolbars/status/buttons/CustomActionButton.tsx
nadberezny: :+1:   ur right, it's unnecessary, it's false anyways

PR #1380, opened 2021-01-18
Compile sink endResult expression in NodeCompiler

gadomsky:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -120,13 +120,17 @@ class NodeCompiler(definitions: ProcessDefinition[ObjectWithMethodDef],
...
+      .map(expr => compileExpression(expr, ctx, Typed[Any], outputVar = None).compiledObject.map(_ => ()))
+      .getOrElse(Valid(()))
+    val compilationResult = definitions.sinkFactories.get(ref.typ) match {
       case Some(definition) =>
         compileObjectWithTransformation[api.process.Sink](sink, Left(ctx), None, definition._1, (_: Any) => Valid(ctx))
       case None =>
         val error = invalid(MissingSinkFactory(sink.ref.typ)).toValidatedNel
         NodeCompilationResult(Map.empty[String, ExpressionTypingInfo], None, Valid(ctx), error)
     }
+    compilationResult.copy(compiledObject = endResultExprValidation.ap(compilationResult.compiledObject.map(self => _ => self)))
```
mproch, line 133: Let's add comment why it's here (that is - we validate expression just to display errors, but we ignore compiled expression, as it's already compiled elsewhere...)

We can also write:
    ```ValidatedNelApplicative.map2(endResultExprValidation, compilationResult.compiledObject) {
      case (_, obj) => obj
    }```
- for me it's more clear than ```self => _ => self```, but just a matter of taste ;)

mproch (approved):

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
mproch: :+1:

gadomsky: You're right, it is acctualy a bit confusing, why we ignore result of this expression compilation - comments added.

For me, `Validated::ap` is a little simpler than `App::map2` - but it's just me. I took youre advice, and changed as you wrote.

PR #1393, opened 2021-01-24
Fix: Github actions publish preview branches

lciolecki:


.github/workflows/pr.yml
```diff
@@ -14,7 +14,7 @@ on:
 env:
   #we use this variable in ciRunSbt.sh
   #NOTE: for publishing we use different settings, we don't use ciRunSbt.sh there
-  CROSS_BUILD: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' }}
+  CROSS_BUILD: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/preview') }}
```
mproch, line 17: Do we want to run all tests on preview branches? Publish is done for all versions regardless of this setting

.github/workflows/pr.yml
```diff
@@ -221,7 +221,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build, tests, crossCompile, integrationTests, slowTests, frontendTests, dockerTest]
     #TODO: should we publish everything on all those branches?
-    if: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || github.ref == 'refs/heads/demo' || startsWith(github.head_ref, 'preview') }}
+    if: ${{ github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || github.ref == 'refs/heads/demo' || startsWith(github.ref, 'refs/heads/preview') || startsWith(github.head_ref, 'preview') }}
```
mproch, line 224: If you want the build to be triggered also on pushes to preview branch, you have to change also triggers around line 7-8

.github/workflows/pr.yml
lciolecki: In fact we do some snapshot release, so we should test it.

mproch (approved):

PR #1396, opened 2021-01-27
Typed QueryableState

mproch:
This is prototype of being able to acces generic state via queryable state in Flink. 
We have to be able to compute TypeInformation for given node for this to work.
Unfortunately, we need https://issues.apache.org/jira/browse/FLINK-21138 to be merged before we can really use it, as we still use Kryo serialization in many cases...


mproch:
Closing for now, we'll have to wait for Flink 1.11.4 etc. to make it usable

PR #1400, opened 2021-01-29
Replace Future with cats.IO (or other monad) in interpreter

mproch:


engine/benchmarks/src/main/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/OneParamInterpreterBenchmark.scala
```diff
@@ -23,34 +25,66 @@ class OneParamInterpreterBenchmark {
...
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  def benchmarkAsyncIO(): AnyRef = {
+    interpreterIO(Context(""), ExecutionContext.Implicits.global).unsafeRunSync()
   }
 
 
 
 }
-object OneParamService extends Service {
+class OneParamService(realAsync: Boolean) extends Service {
```
arkadius, line 82: Both are real ;) Maybe `instantlyCompletedFuture`? Like it is called in java api: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completedFuture-U-

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
```diff
@@ -36,7 +36,9 @@ case class NkGlobalParameters(buildInfo: String,
 //this is part of global parameters that is parsed with typesafe Config (e.g. from application.conf/model.conf)
 case class ConfigGlobalParameters(useLegacyMetrics: Option[Boolean],
                                   explicitUidInStatefulOperators: Option[Boolean],
-                                  useTypingResultTypeInformation: Option[Boolean])
+                                  useTypingResultTypeInformation: Option[Boolean],
+                                 //TODO: temporary, until we confirm which method works better
```
arkadius, line 40: 1. How you will now that one of this method works better?
2. IMO Use I/O is confusing, because it can mean some other I/O. Maybe useIOMonad?

engine/benchmarks/src/main/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/OneParamInterpreterBenchmark.scala
mproch: :+1:

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
mproch: 1. I changed comment :)
2. :+1:

arkadius (approved):

PR #1401, opened 2021-01-29
Improvements: allowed actions support on BE

lciolecki:
This PR provides verification on REST API that we can do one of the following actions: archive / unarchive / deploy / cancel. 

Verification is doing on:
- checking allowed actions from actual process state - we have to configure ProcessStateDefinitionManager
- checking archivization information 
- checking process type (subproces)


ui/server/src/main/resources/db/migration/postgres/V1_028__AddMissingArchiveAction.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE "process_actions" DROP CONSTRAINT "process_deployment_info_check";
```
mproch, line 1: Is there any difference between this and h2 version? If not, let's put it to common?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -0,0 +1,115 @@
+package pl.touk.nussknacker.ui.process
```
mproch, line 1: Can we add description what should go here?
I wonder, maybe it would be better to merge ProcessService with ManagementService? The boundaries are not so clear  for me. If there is good reason for separation, let's write it in some comment?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/exception/ProcessIllegalAction.scala
```diff
@@ -0,0 +1,19 @@
...
+import pl.touk.nussknacker.engine.api.deployment.ProcessActionType.ProcessActionType
+import pl.touk.nussknacker.engine.api.deployment.ProcessState
+import pl.touk.nussknacker.restmodel.process.ProcessIdWithName
+import pl.touk.nussknacker.ui.IllegalOperationError
+
+case class ProcessIllegalAction(message: String) extends Exception(message) with IllegalOperationError
+
+object ProcessIllegalAction {
+  def apply(action: ProcessActionType, processIdWithName: ProcessIdWithName, state: ProcessState): ProcessIllegalAction =
+    ProcessIllegalAction(s"Forbidden action: $action for process: ${processIdWithName.name.value} with state: ${state.status.name}, allowed actions: ${state.allowedActions.mkString(",")}.")
```
mproch, line 12: Action ... is not allowed in state ...
?

ui/server/src/main/resources/db/migration/postgres/V1_028__AddMissingArchiveAction.sql
lciolecki: Oo  good catch! I didn't know about it.. ;)

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -0,0 +1,134 @@
...
+import akka.actor.ActorRef
+import akka.pattern.ask
+import java.time
+
+/**
+  * ProcessService provides functionality for archive, unarchive, deploy, cancel process.
+  * Each action includes verification based on actual process state and checking process is subprocess / archived.
+  */
+class ProcessService(managerActor: ActorRef,
+                     duration: time.Duration,
```
mproch, line 27: timeLimit?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  import scala.concurrent.duration._
+
+  private implicit val timeout: Timeout = Timeout(duration.toMillis millis)
+
+  /**
+    * Handling error at retrieving status from manager is created at ManagementActor
+    */
+  def getProcessState(processIdWithName: ProcessIdWithName)(implicit ec: ExecutionContext, user: LoggedUser): Future[ProcessState] = {
+    implicit val timeout: Timeout = Timeout(1 minute)
```
mproch, line 43: Why we use fixed timeout here, instead of the one passed in constructor?

mproch (approved):
Overall ok, I added minor comments :)

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
lciolecki: Oh.. this is a copy from old one.. You're right.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
lciolecki: actorRequestLimit?

PR #1411, opened 2021-02-08
UX Improvements

lciolecki:
screenshot-localhost_3000-2021 02 17-12_43_09

### DIFF
--- a/ui/client/components/Process/ProcessLastAction.tsx
+++ b/ui/client/components/Process/ProcessLastAction.tsx
@@ -1,0 +1,30 @@
+import React from "react"
+import {ProcessActionType, ProcessType} from "./types"
+import {formatRelatively} from "../../common/DateUtils"
+
+import {Glyphicon} from "react-bootstrap"
+import {useTranslation} from "react-i18next"
+
+type Props = {
+  process: ProcessType,
+}
+
+export default function ProcessLastAction(props: Props): JSX.Element {
+  const {t} = useTranslation()
+  const {process} = props
+
+  const processAction = (lastAction: ProcessActionType): JSX.Element => {
+    const performedAt = formatRelatively(lastAction.performedAt)
+    const user = lastAction.user
+
+    return (
+      <span title={t("processLastAction.title", "Performed by {{user}} at {{performedAt}}.", {performedAt, user})}>
+        {lastAction.action}
+      </span>
+    )
+  }
+
+  return process?.lastAction ?
+    processAction(process.lastAction) :
+    <Glyphicon glyph="minus" title={t("processLastAction.noActionTitle", "No action occurred.")}/>
+}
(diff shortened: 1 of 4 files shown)


ui/client/containers/Processes.tsx
```diff
@@ -49,14 +56,16 @@ const ElementsRenderer: RowsRenderer = ({processes, getProcesses, statuses}) =>
...
-const sortable = ["name", "category", "modifyDate", "createdAt", "createdBy"]
+const sortable = ["name", "category", "modifyDate", "createdAt", "createdBy", "actionDate", "actionUser"]
 const filterable: Filterable = ["name", "processCategory", "createdBy"]
 const columns = [
   {key: "name", label: "Name"},
   {key: "category", label: "Category"},
   {key: "createdBy", label: "Created by"},
   {key: "createdAt", label: "Created at"},
   {key: "modifyDate", label: "Last modification"},
+  {key: "actionDate", label: "Last action date"},
```
mproch, line 67: Is it understandable? I mean, one does not see what the action actually was? I know that it's somehow reflected in state, but I wonder if we can make it more clear? Wouldn't some tooltips be enough?

lciolecki: Hmm, I can add some tooltip with information: Process was deploy/canceled/archived/unarchived at  / by?

mproch: I was more thinking about tooltip with action data by status icon

ui/client/components/Process/ProcessLastAction.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default function ProcessLastAction(props: Props): JSX.Element {
+  const {t} = useTranslation()
+  const {process} = props
+
+  const processAction = (lastAction: ProcessActionType): JSX.Element => {
+    const performedAt = formatRelatively(lastAction.performedAt)
+    const user = lastAction.user
+
+    return (
+      <span title={t("processLastAction.title", "Action run {{performedAt}} by {{user}}.", {performedAt, user})}>
```
mproch, line 21: Performed by {{user}} at {{performedAt}}
?

ui/client/containers/Archive.tsx
```diff
@@ -18,7 +18,8 @@ const ElementsRenderer: RowsRenderer = ({processes}) => processes.map(process =>
     <Td column="category">{process.processCategory}</Td>
     <Td column="createdBy" className="centered-column" value={process.createdBy}>{process.createdBy}</Td>
     <Td column="createdAt" className="centered-column" value={process.createdAt}><Date date={process.createdAt}/></Td>
-    <Td column="modifyDate" className="centered-column" value={process.modificationDate}><Date date={process.modificationDate}/></Td>
+    <Td column="actionDate" className="centered-column" value={process?.lastAction?.performedAt}><Date date={process?.lastAction?.performedAt}/></Td>
```
mproch, line 21: Is it obvious that for archived processes last action is Archive? Now it is so, but if we add other actions??

mproch (approved):
:+1: I added just one minor comment

ui/client/containers/Archive.tsx
lciolecki: Right now we assume that after archive we can do only unarchvie :) If want change this behavior in future, then we should return this information in another fields.  Right now we can leave it.

PR #1412, opened 2021-02-09
[Refactor] Simplify QueryService

mproch:
Main aim is to use NodeCompiler when preparing service to invoke, so that logic (especially WRT dynamic, eager services) can be reused


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
```diff
@@ -17,61 +28,74 @@ class ServiceQuery(modelData: ModelData) {
...
-      collector.cleanResults()
+      val collector = QueryServiceInvocationCollector(serviceName)
+        .enable(TestRunId(UUID.randomUUID().toString))
+
+      val variablesPreparer = GlobalVariablesPreparer(definitions.expressionConfig)
+      val validationContext = variablesPreparer.emptyValidationContext(metaData)
+
+      val compiled = compiler.compileService(ServiceRef(serviceName, params), validationContext, None, Some(_ => collector))(NodeId(""), metaData)
+      compiled.compiledObject.map { service =>
+          service.invoke(ctx, ExpressionEvaluator.unOptimizedEvaluator(variablesPreparer))._2.map(QueryResult(_, collector.getResults))
```
jedrz, line 62: Shouldn't `evaluator` be used here?

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/ServiceRoutesSpec.scala
```diff
@@ -70,23 +69,24 @@ class ServiceRoutesSpec extends FunSuite with Matchers with ScalatestRouteTest w
       """.stripMargin)
     Post("/service/streaming/enricher", entity) ~> serviceRoutes.securedRoute ~> check {
       status shouldEqual StatusCodes.InternalServerError
-      entityAs[JsonThrowable].message shouldEqual Some("NonEmptyList(ExpressionParseError(EL1041E: After parsing a valid expression, there is still more data in the expression: 'spell',defaultNodeId,Some(param),not valid spell expression))")
-      entityAs[JsonThrowable].className shouldEqual classOf[ParametersCompilationException].getCanonicalName
+      entityAs[JsonThrowable].message shouldEqual Some("MissingParameters(Set(tariffType),)")
```
jedrz, line 72: What about adding tariffType parameter to the invocation so the error should remain more or less the same?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
mproch: :+1: done

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/ServiceRoutesSpec.scala
mproch: :+1: done

jedrz (approved):

PR #1416, opened 2021-02-12
OAuth2 refactoring and authorizations' cache

trombka:


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.Decoder
+import pl.touk.nussknacker.engine.util.cache.{CacheConfig, DefaultAsyncCache, ExpiryConfig}
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+
+import scala.concurrent.duration.{Deadline, FiniteDuration, MILLISECONDS}
+import scala.concurrent.{ExecutionContext, Future}
+
+class BaseOAuth2Service[
```
lciolecki, line 11: `CachedOAuth2Service`?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
```diff
@@ -0,0 +1,51 @@
...
+](clientApi: OAuth2ClientApi[ProfileResponse, AccessTokenResponse],
+  createLoggedUser: (ProfileResponse) => LoggedUser,
+  configuration: OAuth2Configuration)
+ (implicit ec: ExecutionContext)
+  extends BaseOAuth2Service[ProfileResponse, AccessTokenResponse](clientApi, createLoggedUser, configuration)
+    with LazyLogging {
+
+  protected val introspectAccessToken: Boolean = configuration.jwt.exists(_.accessTokenIsJwt)
+
+  protected val jwtValidator: Option[JwtValidator] = configuration.jwt.map(_.authServerPublicKey).map(new JwtValidator(_))
```
lciolecki, line 36: private instead of protected?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
```diff
@@ -0,0 +1,51 @@
...
+  AccessTokenResponse <: StandardAccessTokenResponse : Decoder,
+  DecodedJwtAccessToken <: JwtStandardClaims : Decoder
+](clientApi: OAuth2ClientApi[ProfileResponse, AccessTokenResponse],
+  createLoggedUser: (ProfileResponse) => LoggedUser,
+  configuration: OAuth2Configuration)
+ (implicit ec: ExecutionContext)
+  extends BaseOAuth2Service[ProfileResponse, AccessTokenResponse](clientApi, createLoggedUser, configuration)
+    with LazyLogging {
+
+  protected val introspectAccessToken: Boolean = configuration.jwt.exists(_.accessTokenIsJwt)
```
lciolecki, line 34: private instead of protected?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
```diff
@@ -0,0 +1,51 @@
...
+
+class JwtOAuth2Service[
+  ProfileResponse: Decoder,
+  AccessTokenResponse <: StandardAccessTokenResponse : Decoder,
+  DecodedJwtAccessToken <: JwtStandardClaims : Decoder
+](clientApi: OAuth2ClientApi[ProfileResponse, AccessTokenResponse],
+  createLoggedUser: (ProfileResponse) => LoggedUser,
+  configuration: OAuth2Configuration)
+ (implicit ec: ExecutionContext)
+  extends BaseOAuth2Service[ProfileResponse, AccessTokenResponse](clientApi, createLoggedUser, configuration)
```
lciolecki, line 31: Please use composition over inheritance - `delegate: BaseOAuth2Service[ProfileResponse, AccessTokenResponse]`.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectService.scala
```diff
@@ -0,0 +1,33 @@
...
+
+class OpenIdConnectService[
+  ProfileResponse: Decoder,
+  AccessTokenResponse <: OpenIdConnectAccessTokenResponse : Decoder,
+  DecodedJwtAccessToken <: JwtStandardClaims : Decoder
+](clientApi: OAuth2ClientApi[ProfileResponse, AccessTokenResponse],
+  createLoggedUser: (ProfileResponse) => LoggedUser,
+  configuration: OAuth2Configuration)
+ (implicit ec: ExecutionContext)
+  extends JwtOAuth2Service[ProfileResponse, AccessTokenResponse, DecodedJwtAccessToken](clientApi, createLoggedUser, configuration)
```
lciolecki, line 23: Please use composition over inheritance - `delegate: JwtOAuth2Service`.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
```diff
@@ -0,0 +1,56 @@
...
+  ProfileResponse: Decoder,
+  AccessTokenResponse <: StandardAccessTokenResponse : Decoder
+](clientApi: OAuth2ClientApi[ProfileResponse, AccessTokenResponse],
+  createLoggedUser: (ProfileResponse) => LoggedUser,
+  configuration: OAuth2Configuration)
+ (implicit ec: ExecutionContext) extends OAuth2Service with LazyLogging {
+
+  protected val authorizationsCache = new DefaultAsyncCache[String, (LoggedUser, Deadline)](CacheConfig(new ExpiryConfig[String, (LoggedUser, Deadline)]() {
```
lciolecki, line 19: private instead of protected?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/DefaultOAuth2ServiceFactory.scala
```diff
@@ -63,7 +44,7 @@ object DefaultOAuth2ServiceFactoryWithProfileFormat {
       throw new Exception("profileFormat is missing in the authentication configuration")
     } match {
       case ProfileFormat.GITHUB => new DefaultOAuth2ServiceFactoryWithProfileFormat[GitHubProfileResponse](GitHubProfile)
-      case ProfileFormat.AUTH0 => new DefaultOAuth2ServiceFactoryWithProfileFormat[Auth0ProfileResponse](Auth0Profile)
+      case ProfileFormat.DEFAULT => new DefaultOAuth2ServiceFactoryWithProfileFormat[DefaultProfileResponse](DefaultProfile)
```
lciolecki, line 47: Why right now it's default? Isn't it still AUTH0?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2Configuration.scala
```diff
@@ -79,7 +78,8 @@ object JwtConfiguration {
 
   implicit val jwtConfigurationVR: ValueReader[JwtConfiguration] = ValueReader.relative(_.rootAs[JwtConfig])
 
-  private case class JwtConfig(enabled: Boolean = false,
+  private case class JwtConfig(accessTokenIsJwt: Boolean = false,
```
lciolecki, line 81: Defaults on class is ugly way.. Better is create object with `apply(publicKey: Option[String],publicKeyFile: Option[String], certificate: Option[String])` which creates `JwtConfig(accessTokenIsJwt = false, userinfoFromIdToken = false, publicKey, publicKeyFile, certificate)`.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectService.scala
```diff
@@ -0,0 +1,33 @@
...
+    with LazyLogging {
+
+  protected val useIdToken: Boolean = configuration.jwt.exists(_.userinfoFromIdToken)
```
lciolecki, line 26: private instead of protected?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/cache/Cache.scala
```diff
@@ -1,20 +1,44 @@
...
   val defaultMaximumSize = 10000L
-}
\ No newline at end of file
+
+  /**
+   * @param maximumSize the maximum elements number can contain cache
+   * @param expireAfterAccess the expiration time from last action (read / write)
+   * @param expireAfterWrite the expiration time after value was written to cache
+   */
+  def apply(maximumSize: Long = CacheConfig.defaultMaximumSize,
```
lciolecki, line 36: Formatting.

lciolecki (changes requested):
Very good change !:) I left some questions and improvements in comments.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
```diff
@@ -0,0 +1,56 @@
...
+    override def expireAfterWriteFn(key: String, value: (LoggedUser, Deadline), now: Deadline): Option[Deadline] = Some(value._2)
+  }))
+
+  import pl.touk.nussknacker.ui.security.oauth2.BaseOAuth2Service._
+
+  def authenticate(code: String): Future[OAuth2AuthenticateData] = {
+    clientApi.accessTokenRequest(code).map { resp: AccessTokenResponse =>
```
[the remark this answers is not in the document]
trombka: It is not the authorization code that is cached here, but an access token mapped to a LoggedUser instance.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
trombka: The validator is also used for ID token validation in the OIDC subclass.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
trombka: I think it is useful for a subclass to have access to this field in case it overrides any method using it e.g. getProfile.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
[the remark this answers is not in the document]
lciolecki: You're right - I see it now ;) But what happend when ` getProfile(resp).map((createExpiringLoggedUser _).tupled)` returns error? :) I know that is small chance of that, but still server can return 501 / 503..  Maybe better way is return there response and cached response at first time on `authorize`?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
lciolecki: So let's do access by method?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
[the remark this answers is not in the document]
trombka: Do you think there is a case for a service without even a short cache? With no cache, every request to the Nussknacker API triggers a query to a remote authorization server, which might be remote or have a quota.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/DefaultOAuth2ServiceFactory.scala
pjagielski: maybe explicitly OIDC?

trombka: The default profile is meant to be in the format of `/userinfo` response of OIDC which is widely accepted. Auth0 delivers this endpoint too.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2Configuration.scala
trombka: This is a configuration class for Typesafe Config. Will it handle `apply`? No such pattern in the existing code.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
trombka: Why?! It is a `val`. It is like a Java a getter already.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2Configuration.scala
trombka: And it is a `case class` which is already Scala's syntactic sugar for the `class` and `apply` in a companion object.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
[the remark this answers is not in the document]
lciolecki: 1. I can imagine that case.. But you're right it's unlikely.But still it's good practice to separate getting from cache.
2. I think this class should be named `CacheOAuth2Service` with possibility to inject `authorizationsCache` by constructor. Right now you support only one cache - `DefaultAsyncCache`. What about use db as a cache? Cache in memory will not work with more nk instances.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
lciolecki: I think OIDC subclass should use `JwtOAuth2Service` as underlying / delegate. After that this field should be hided.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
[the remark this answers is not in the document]
trombka: You're right about most of the above.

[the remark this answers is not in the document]
trombka: I tried to convert it your way, but it is easier said than done.
With the decorator pattern, cache entries would have to be created from the results of methods of the decorated object only.
The first problem is with token expiry which in the case of JWT would need to be returned from the `authorize` method, this would be justified.
But another problem is with OIDC. It is nice to use an id token that comes with the access token response and not to call the `/userinfo` endpoint later. Again, to cache with a decorator, we would need to have the user profile returned from the service's method along with an expiry. In my opinion, this would make the pattern misused by changing an interface to fit this exact decorator.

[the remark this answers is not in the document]
trombka: Maybe we should not think of it as a cache, but an authorization store?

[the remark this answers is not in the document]
trombka: I think I have managed it. There is CachingOAuth2Service now, which is a decorator. However, it required a new service interface and a backward compatible wrapper.

trombka: Renamed.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
trombka: OIDC needs a JWT parser/validator for the ID token, too.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectService.scala
trombka: If we allow people to implement their own implementation of OAuth2 and inject it by SPI it desirable to give a chance not to write it all from scratch.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
```diff
@@ -0,0 +1,56 @@
...
+import io.circe.generic.JsonCodec
+import io.circe.generic.extras.{Configuration, ConfiguredJsonCodec, JsonKey}
+import pl.touk.nussknacker.ui.security.api.{LoggedUser, RulesSet}
+import pl.touk.nussknacker.ui.security.oauth2.OAuth2Profile.getUserRoles
+
+import scala.concurrent.duration.Deadline
+
+@ConfiguredJsonCodec case class OpenIdConnectUserInfo
+(
+  @JsonKey("sub") subject: Some[String],
```
mproch, line 12: Either Option[String] or just String? :)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
```diff
@@ -0,0 +1,56 @@
...
+  @JsonKey("middle_name") middleName: Option[String],
+  nickname: Option[String],
+  @JsonKey("preferred_username") preferredUsername: Option[String],
+  profile: Option[String],
+  picture: Option[String],
+  website: Option[String],
+  email: Option[String],
+  @JsonKey("email_verified") emailVerified: Option[Boolean],
+  gender: Option[String],
+  @JsonKey("birth_date") birthDate: Option[String], //ISO 8601
```
mproch, line 26: If we know it's ISO maybe we can just use decoded version?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
```diff
@@ -1,19 +1,55 @@
...
 import scala.concurrent.{ExecutionContext, Future}
 
+trait OAuth2AuthorizationData {
+  val accessToken: String
+  val expirationPeriod: Option[FiniteDuration]
+  val tokenType: String
+  val refreshToken: Option[String]
+}
+
+trait OAuth2NewService[+UserInfoData, +AuthorizationData <: OAuth2AuthorizationData] {
```
mproch, line 16: I wonder if it makes sense to maintain backwards compatibility here. We'll be left with this name forever. For non-essential API we usually just add migration guide?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
trombka: This class implements the `JwtStandardClaim` interface. In a JWT the field is optional, but in the context of OIDC, it becomes mandatory.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
trombka: True, but changing this will make the PR even bigger. :) I thought we could do it in two steps.

trombka: I think I can change `OAuthService` into `OAuthOldService` and `OAuthNewService` into `OAuthService` (and do all the subsequent changes) so that the simplest migration would be to import the old class with an alias.

mproch: :+1:

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
mproch: ok, I understand, nice idea :+1: though a comment why it's like that would be useful...

trombka: I have added a comment.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
trombka: Replaced with `LocalDate`

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
trombka: Done. I've also put an entry in the migration guide.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/BaseOAuth2Service.scala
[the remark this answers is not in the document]
trombka: It is only few milliseconds after the UI receives a token when it uses it with the first request and I think it is better to fail fast when an authorization server is not available. Moreover, we can handle such a case more appropriately in the UI when it is on the `/authentication` endpoint and not any other.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CachingOAuth2Service.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(UserInfoData, Option[Deadline])] = {
+    authorizationsCache.getOrCreate(accessToken) {
+      delegate.checkAuthorizationAndObtainUserinfo(accessToken).map {
+        case (userInfo, expiration) => (userInfo, expiration.getOrElse(Deadline.now + defaultExpiration()))
+      }
+    }.map { case (userInfo, expiration) => (userInfo, Some(expiration)) }
+  }
+
+  private def defaultExpiration() = configuration.defaultTokenExpirationTime
```
lciolecki, line 39: It can be lazy val.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2OldServiceFactory.scala
```diff
@@ -0,0 +1,27 @@
...
+  def authenticate(code: String): Future[OAuth2AuthenticateData]
+  def authorize(token: String): Future[LoggedUser]
+
+  final override def obtainAuthorizationAndUserInfo(authorizationCode: String): Future[(OAuth2AuthenticateData, Option[LoggedUser])] =
+    authenticate(authorizationCode).map((_, None))
+  final override def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(LoggedUser, Option[Deadline])] =
+    authorize(accessToken).map((_, None))
+}
+
+case class OAuth2AuthenticateData(access_token: String, token_type: String, refresh_token: Option[String]) extends OAuth2AuthorizationData {
```
lciolecki, line 22: `access_token` => `accessToken`, `token_type` => `tokenType`,  `refresh_token` => `refreshToken` ?

engine/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactorySpec.scala
```diff
@@ -18,10 +18,11 @@ class ExampleOAuth2ServiceFactorySpec extends FlatSpec with Matchers with Patien
   import io.circe.syntax._
 
   import ExecutionContext.Implicits.global
+//  import OAuth2ServiceFactory.BackwardCompatibleOAuth2ServiceWrapper
```
lciolecki, line 21: Do we need it?

lciolecki (approved):
LGTM, but let's look at some comments below.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CachingOAuth2Service.scala
trombka: True. Changed.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2OldServiceFactory.scala
trombka: It is a class from the previous interface. I think it would be an unnecessary migration burden since we want to drop this interface at all.

engine/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactorySpec.scala
trombka: Of course not.

PR #1417, opened 2021-02-12
React fast refresh

JulianWielga:


ui/client/components/TogglePanel.js
```diff
@@ -1,24 +1,26 @@
...
-    type: PropTypes.oneOf(["RIGHT", "LEFT"]).isRequired,
-    isOpened: PropTypes.bool.isRequired,
-    onToggle: PropTypes.func.isRequired,
-  }
+export default function TogglePanel(props) {
+  const {t} = useTranslation()
+  const {isOpened, onToggle, type} = props
+  const left = type === "LEFT" ? isOpened : !isOpened
+  const iconFile = `arrows/arrow-${left ? "left" : "right"}.svg`
+  const title = type === "LEFT" ?
```
lciolecki, line 14: Let's do this `LEFT` as enum.

lciolecki (approved):
LGTM, but please create LEFT/RIGHT as enum type.

ui/client/components/TogglePanel.js
JulianWielga: enum here gives nothing, only two well known values. string literal types works.

PR #1423, opened 2021-02-16
[Logs] Remove some unnecessary logs in tests

mproch:


engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -52,14 +57,19 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
 
   private def kafkaConfig(config: Config) = KafkaConfig.parseConfig(config)
 
-  override def sinkFactories(processObjectDependencies: ProcessObjectDependencies): Map[String, WithCategories[SinkFactory]] =
+  override def sinkFactories(processObjectDependencies: ProcessObjectDependencies): Map[String, WithCategories[SinkFactory]] = {
+    val confluentFactory: ConfluentSchemaRegistryClientFactory = if (processObjectDependencies.config.hasPath(DevProcessConfigCreator.emptyMockedSchemaRegistryProperty)) {
```
lciolecki, line 61: What will happend when `withMockedConfluent: false` in config?

lciolecki (approved):
This feature with setting `ConfluentSchemaRegistryClientFactory` it's ugly.. But right now I don't have any idea how do it in another way ;)

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
mproch: ok, I made it a bit better ;)

PR #1430, opened 2021-02-22
Global ExecutionContext and local Sttp backend removed

trombka:
I have found `ExecutionContext.Implicits.global` in a few places in the security package and changed it to implicit parameters so that we could use the main (from Akka) NK execution context.

I have also removed a locally created Sttp backend in favour of one created from Akka.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -250,6 +250,7 @@ val jbcryptV = "0.4"
 val cronParserV = "9.1.3"
 val javaxValidationApiV = "2.0.1.Final"
 val caffeineCacheV = "2.8.2"
+val sttpV = "2.2.3"
 
 lazy val dockerSettings = {
   val workingDir = "/opt/nussknacker"
@@ -826,7 +827,6 @@ lazy val httpUtils = (project in engine("httpUtils")).
   settings(
     name := "nussknacker-http-utils",
     libraryDependencies ++= {
-      val sttpV = "2.2.3"
       Seq(
         "io.circe" %% "circe-core" % circeV,
         "io.circe" %% "circe-parser" % circeV,
@@ -925,6 +925,7 @@ lazy val ui = (project in file("ui/server"))
         "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpV % "test",
         "com.typesafe.akka" %% "akka-testkit" % akkaV % "test",
         "de.heikoseeberger" %% "akka-http-circe" % akkaHttpCirceV,
+        "com.softwaremill.sttp.client" %% "akka-http-backend" % sttpV,
 
         "ch.qos.logback" % "logback-core" % logbackV,
         "ch.qos.logback" % "logback-classic" % logbackV,
@@ -945,7 +946,7 @@ lazy val ui = (project in file("ui/server"))
 
         "com.typesafe.slick" %% "slick-testkit" % slickV % "test",
         "com.whisk" %% "docker-testkit-scalatest" % "0.9.8" % "test",
-        "com.whisk" %% "docker-testkit-impl-spotify" % "0.9.8" % "test"
+        "com.whisk" %% "docker-testkit-impl-spotify" % "0.9.8" % "test",
       )
     }
   )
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -834,6 +834,7 @@ lazy val httpUtils = (project in engine("httpUtils")).
         "org.scala-lang.modules" %% "scala-parser-combinators" % scalaParsersV, // scalaxb deps
         "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV,
         "com.softwaremill.sttp.client" %% "core" % sttpV,
+        "com.softwaremill.sttp.client" %% "akka-http-backend" % sttpV,
```
mproch, line 837: I don't think we want to introduce akka dependency here, as this module is often used in model.
Let's add it only where it's needed (security or ui)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicHttpAuthenticator.scala
```diff
@@ -7,10 +7,9 @@ import pl.touk.nussknacker.engine.util.cache.DefaultCache
 import pl.touk.nussknacker.ui.security.api.{DefaultAuthenticationConfiguration, LoggedUser, RulesSet}
 import pl.touk.nussknacker.ui.security.basicauth.BasicHttpAuthenticator.{EncryptedPassword, PlainPassword, UserWithPassword}
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
-class BasicHttpAuthenticator(configuration: DefaultAuthenticationConfiguration, allCategories: List[String]) extends SecurityDirectives.AsyncAuthenticator[LoggedUser] {
+class BasicHttpAuthenticator(configuration: DefaultAuthenticationConfiguration, allCategories: List[String])(implicit ec: ExecutionContext) extends SecurityDirectives.AsyncAuthenticator[LoggedUser] {
```
mproch, line 12: I think here we don't need ec at all, just replace Future { ... } with Future.successful

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticatorFactory.scala
```diff
@@ -6,7 +6,9 @@ import com.typesafe.config.Config
 import pl.touk.nussknacker.ui.security.api.AuthenticatorFactory.{AuthenticatorData, LoggedUserAuth}
 import pl.touk.nussknacker.ui.security.api.{AuthenticatorFactory, DefaultAuthenticationConfiguration}
 
-class BasicAuthenticatorFactory extends AuthenticatorFactory with Directives {
+import scala.concurrent.ExecutionContext
+
+class BasicAuthenticatorFactory(implicit ec: ExecutionContext) extends AuthenticatorFactory with Directives {
```
mproch, line 11: I think we don't need it, see comment in BasicHttpAuthenticator

mproch (changes requested):
overall very needed change, I added a few remarks, most important is in buid.sbt

build.sbt
trombka: Moved to UI.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicHttpAuthenticator.scala
trombka: Removed

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticatorFactory.scala
trombka: Removed

mproch (approved):

PR #1432, opened 2021-02-22
node with removed type display errors

### LINKED ISSUE #1270: Editing node with removed type
When we try to edit node with removed type then we see error 500 page. In that case should see popup with information about non exists element type. Maybe we can show also node params with disabled edition possibility?

JulianWielga:
fixes #1270


lciolecki (approved):
LGTM

PR #1433, opened 2021-02-24
[Feature] Introduce DeploymentData to be passed to execution

mproch:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/JobData.scala
```diff
@@ -1,3 +1,5 @@
 package pl.touk.nussknacker.engine.api
 
-case class JobData(metaData: MetaData, processVersion: ProcessVersion)
+import pl.touk.nussknacker.engine.api.deployment.DeploymentVersion
+
+case class JobData(metaData: MetaData, processVersion: ProcessVersion, deploymentVersion: DeploymentVersion)
```
arkadius, line 5: I wonder if it is a good time to start using more human readable name for this data. It doesn't look like a version at all ;) Or at least add here to todo that it should be renamed...

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ExternalDeploymentId.scala
```diff
@@ -0,0 +1,12 @@
+package pl.touk.nussknacker.engine.api.deployment
+
+import io.circe.generic.extras.semiauto.{deriveUnwrappedDecoder, deriveUnwrappedEncoder}
+import io.circe.{Decoder, Encoder}
+
+//id generated by external system - e.g. Flink JobID
+final case class ExternalDeploymentId(value: String) extends AnyVal
```
arkadius, line 7: :+1: for introducing new name to clarify that it is not and id passed from ui. But maybe instead of external we could name it engine? Because external looks like it is defined by some other system

mproch: For me it's external - outside NK?

arkadius: If you will have embedded engine that will generate own id would it be still external? ... but it is only suggestion so you can skip it if you want.

arkadius (approved):

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/JobData.scala
mproch: :+1: I changed to DeploymentData

PR #1438, opened 2021-02-26
Posibility to inject Context in ServiceQuery to make testing services easier

gadomsky:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
```diff
@@ -24,14 +24,12 @@ import pl.touk.nussknacker.engine.variables.GlobalVariablesPreparer
 import scala.concurrent.{ExecutionContext, Future}
 
 // TODO: Processes using Flink's RuntimeContex, ex. metrics throws NPE, but in another thread, so service works.
-class ServiceQuery(modelData: ModelData) {
+class ServiceQuery(modelData: ModelData, ctx: Context = Context("")) {
```
mproch, line 27: Do we want it here, or in the invoke method?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
```diff
@@ -54,7 +52,7 @@ class ServiceQuery(modelData: ModelData) {
       val collector = QueryServiceInvocationCollector(Some(TestRunId(UUID.randomUUID().toString)), serviceName)
 
       val variablesPreparer = GlobalVariablesPreparer(definitions.expressionConfig)
-      val validationContext = variablesPreparer.emptyValidationContext(metaData)
+      val validationContext = variablesPreparer.validationContextWithLocalVariables(metaData, ctx.variables.mapValues(Typed.fromInstance))
```
mproch, line 55: Maybe it would be better to pass ValidationContext explicitly (or just have option to do it).  Typed.fromInstance can be inprecise...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
gadomsky: You're right, it would be better to have it in the `invoke` method, but it scala syntax does not think so..

Unless we ged rid of varargs paramter from `def invoke(serviceName: String, args: (String, Expression)*)` or make `context` implicit arg

gadomsky: I assumed thats this parameter should have default value..

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/util/service/query/ServiceQuery.scala
gadomsky: :+1: 
ok, but mabye we should pass `localVariables: Map[String, (Any, TypingResult)]` instead of `Context` and `ValidationContext` ? 
WDYT?

mproch: ok for me :)

mproch (approved):

PR #1441, opened 2021-03-01
Fix expiration units: milliseconds -> seconds

trombka:
I got it wrong from the specification before.


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CirceDurationConversions.scala
```diff
@@ -2,7 +2,7 @@ package pl.touk.nussknacker.ui.security.oauth2
 
 import io.circe._
 
-import scala.concurrent.duration.{Deadline, FiniteDuration, MILLISECONDS}
+import scala.concurrent.duration.{Deadline, FiniteDuration, SECONDS}
 
 protected[oauth2] trait CirceDurationConversions {
```
mproch, line 7: I think this trait/codecs should be named e.g. OAuth codecs or EpochSecondsCodecs, because it's not obvious without looking at implementation how they work?

mproch (approved):

PR #1449, opened 2021-03-04
[Feature] Make running local NK UI easier

mproch:
Helpers to make running UI with fixed model easier. This is especially needed when developing custom node transformers - we want to see result on UI, without packaging model into jar every time.

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
@@ -37,6 +37,8 @@ class ProcessManagerStub extends ProcessManager {
 
 }
 
+//This provider can be used for testing. Override methods to implement more complex behaviour
+//Provider is registered via ServiceLoader, so it can be used e.g. to run simple docker configuration
 class ProcessManagerProviderStub extends ProcessManagerProvider {
 
   override def createProcessManager(modelData: ModelData, config: Config): ProcessManager = new ProcessManagerStub
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
@@ -1,0 +1,52 @@
+package pl.touk.nussknacker.engine.testing
+
+import com.typesafe.config.Config
+import pl.touk.nussknacker.engine.api.deployment.simple.SimpleProcessStateDefinitionManager
+import pl.touk.nussknacker.engine.api.{ProcessVersion, StreamMetaData, TypeSpecificData}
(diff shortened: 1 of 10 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
```diff
@@ -11,7 +11,8 @@ class WebResources(publicPath: String) extends Directives {
   private lazy val mainContentFile = {
     val tempMainContentFile = Files.createTempFile("nussknacker", "main.html").toFile
     tempMainContentFile.deleteOnExit()
-    val content = IOUtils.toString(getClass.getClassLoader.getResourceAsStream("web/static/main.html"))
+    val data = getClass.getClassLoader.getResourceAsStream("web/static/main.html")
+    val content = Option(data).map(IOUtils.toString).getOrElse("")
```
arkadius, line 15: Can you write a comment that this fallback is in case if you hasn't got client app in resources and want to at least has working backend. Am I right that it is that case? Maybe we could also add some logging for this case because someone can spend a while figuring out what is wrong when he/she skip fe preparation.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
```diff
@@ -0,0 +1,52 @@
...
+import pl.touk.nussknacker.engine.api.deployment.simple.SimpleProcessStateDefinitionManager
+import pl.touk.nussknacker.engine.api.{ProcessVersion, StreamMetaData, TypeSpecificData}
+import pl.touk.nussknacker.engine.api.deployment.{CustomAction, CustomActionError, CustomActionNotImplemented, CustomActionRequest, CustomActionResult, ProcessDeploymentData, ProcessManager, ProcessState, ProcessStateDefinitionManager, SavepointResult, TestProcess, User}
+import pl.touk.nussknacker.engine.api.process.ProcessName
+import pl.touk.nussknacker.engine.{ModelData, ProcessManagerProvider}
+import pl.touk.nussknacker.engine.api.queryablestate.QueryableClient
+
+import scala.concurrent.Future
+
+class ProcessManagerStub extends ProcessManager {
```
arkadius, line 13: I wonder if we could use it from docker e.g. for purpose of cypress tests or in case if someone don't want to setup flink and don't want to see nasty errors during fetch status

arkadius (commented):
LGTM, two minor comments added.

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
mproch: I added log.error - outside development this really should not happen

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/testing/ProcessManagerStub.scala
mproch: I think we can, I added it also to ServiceLoader - loading is by name so it should be ok

arkadius (approved):

PR #1452, opened 2021-03-05
[FIX] Handle Services with dynamic parameters on FE

nadberezny:


ui/client/components/graph/node-modal/ParametersUtils.ts
```diff
@@ -12,7 +12,7 @@ const findUnusedParameters = (parameters: Array<Parameter>, definitions: Array<U
   return parameters.filter(param => !definitions.find(def => def.name == param.name))
 }
 
-//Currently we want to handle adjustment only for CustomNode,Source and Sink, TODO: add other types when they will be handled
+// TODO: add other types when they will be handled
```
mproch, line 15: I think we don't have any cases left :) 
https://github.com/TouK/nussknacker/blob/staging/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala#L82 => let's add comment here, that if someone wants to add new implementation, they must also add case here? 
Also, can we change `propertiesPath` to `parametersPath`? I think it was bad name from the beginning (of course I'm the culprit...)

nadberezny: :+1: done
https://github.com/TouK/nussknacker/pull/1452/commits/cbac7c5b790f341fa424fb689a4e5b849caaf84e

mproch (approved):
thanks :+1:

arkadius (approved):

PR #1455, opened 2021-03-07
Upgrade project dependencies

piotrp:
Some dependency upgrades. Scala 2.11 holds us from upgrading circe and cats. Flink with its frozen Akka version makes it impossible to use a newer Akka HTTP.

NK extensions use nussknacker-ui assembly as their runtime so we should try keeping packaged dependencies up to date.


piotrp:
This reminded me how fragile Scala libraries are...
* scala-collection-compat 2.4+ breaks Silencer, which needs to be upgraded
* but that requires newer Scala version, because Silencer is built only for specific compiler versions
* and I can't upgrade Scala to 2.12.13 because this breaks scoverage

mproch (approved):

PR #1463, opened 2021-03-11
FIX: do not send cancel request to Flink when canceling not running p…

gadomsky:
…rocess

### DIFF
--- a/engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkProcessManager.scala
+++ b/engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkProcessManager.scala
@@ -65,7 +65,9 @@ abstract class FlinkProcessManager(modelData: ModelData, shouldVerifyBeforeDeplo
 
   override def cancel(processName: ProcessName, user: User): Future[Unit] = {
     findJobStatus(processName).flatMap {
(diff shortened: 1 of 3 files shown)


lciolecki (approved):

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkProcessManager.scala
```diff
@@ -65,7 +65,8 @@ abstract class FlinkProcessManager(modelData: ModelData, shouldVerifyBeforeDeplo
 
   override def cancel(processName: ProcessName, user: User): Future[Unit] = {
     findJobStatus(processName).flatMap {
-      case Some(ProcessState(Some(deploymentId), _, _, actions, _, _, _, _, _, _)) if actions.contains(ProcessActionType.Cancel) =>
+      // Flink does not allow to cancel 'completed' job
```
lciolecki, line 68: Maybe not 'completed' but 'not runnint'?

gadomsky: Ok. 'completed' was supposed to refer to `Completed Job List` in Flink UI Console

mproch (approved):

PR #1464, opened 2021-03-12
[Improvement] Small aggregate refactor, more possibilities WRT triggers

mproch:
1. Some helper methods, comments
2. It's possible to define tumbling window emitted on each event

This PR is preparation for adding session windows


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
```diff
@@ -49,40 +49,48 @@ object keyed {
...
   }
 
-  class StringKeyedValueMapper(protected val lazyParameterHelper: FlinkLazyParameterFunctionHelper, key: LazyParameter[CharSequence], value: LazyParameter[AnyRef])
-    extends BaseKeyedValueMapper[String] {
 
-    private lazy val interpreter = prepareInterpreter(key.map(transformKey), value)
+
+  class StringKeyedValueMapper[T<:AnyRef:TypeTag](protected val lazyParameterHelper: FlinkLazyParameterFunctionHelper,
+                                                  key: LazyParameter[CharSequence],
+                                                  value: LazyParameterInterpreter => LazyParameter[T])
```
arkadius, line 80: Parameter that is a function taking parameter interpreter and returning lazy parameter that will be interpreted by some other parameter interpreter? It looks like we need at least comment in this place ;)

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/TransformStateTransformer.scala
```diff
@@ -55,7 +56,7 @@ class TransformStateFunction(protected val lazyParameterHelper: FlinkLazyParamet
                              transformWhenParam: LazyParameter[java.lang.Boolean],
                              newValueParam: LazyParameter[AnyRef],
                              stateTimeout: FiniteDuration)
-  extends LatelyEvictableStateFunction[ValueWithContext[CharSequence], ValueWithContext[AnyRef], GenericState]
+  extends LatelyEvictableStateFunction[ValueWithContext[String], ValueWithContext[AnyRef], GenericState]
```
arkadius, line 59: Why this type change is needed here?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -78,7 +78,8 @@ object aggregates {
 
     override def mergeAggregates(agg1: Aggregate, agg2: Aggregate): Aggregate = agg1 ++ agg2
 
-    override def result(finalAggregate: Aggregate): AnyRef = new java.util.ArrayList[Any](finalAggregate.asJava)
+    //we revert elements only during final results computation
```
arkadius, line 81: Woops. Will this change break a business logic in processes that was made on assumption that the order is different?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
```diff
@@ -49,40 +49,48 @@ object keyed {
...
     override protected def interpret(ctx: Context): KeyedValue[AnyRef, AnyRef] = interpreter(ctx)
 
   }
 
-  class StringKeyedValueMapper(protected val lazyParameterHelper: FlinkLazyParameterFunctionHelper, key: LazyParameter[CharSequence], value: LazyParameter[AnyRef])
-    extends BaseKeyedValueMapper[String] {
 
-    private lazy val interpreter = prepareInterpreter(key.map(transformKey), value)
+
+  class StringKeyedValueMapper[T<:AnyRef:TypeTag](protected val lazyParameterHelper: FlinkLazyParameterFunctionHelper,
```
arkadius, line 78: It looks like we need an entry in `MigrationGuide` for those changes.

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/sampleTransformers.scala
```diff
@@ -157,10 +157,18 @@ object sampleTransformers {
                   )), defaultMode = DualEditorMode.SIMPLE) aggregator: Aggregator,
                 @ParamName("aggregateBy") aggregateBy: LazyParameter[AnyRef],
                 @ParamName("windowLength") length: java.time.Duration,
-                @ParamName("emitExtraWindowWhenNoData") emitExtraWindowWhenNoData: Boolean,
+                @ParamName("emitWhen")
+                @DualEditor(simpleEditor = new SimpleEditor(
+                  `type` = SimpleEditorType.FIXED_VALUES_EDITOR,
+                  possibleValues = Array(
+                    new LabeledExpression(label = "On each event", expression = "'OnEvent'"),
```
arkadius, line 164: WDYT about extracting java enum and use here this `T(...).VALUE` approach like we have in `aggregator`? It will be consistent + less code parsing this parameter.

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
mproch: I added comment :)

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/TransformStateTransformer.scala
mproch: I added comment and changed to T..

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
mproch: Good point, I will defer it to next PR. For now, I added TODO

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
mproch: Added

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/sampleTransformers.scala
mproch: :+1: done

arkadius (approved):

PR #1465, opened 2021-03-12
flexible users file location - fs path, classpath or url

trombka:


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
```diff
@@ -62,6 +62,15 @@ object AuthenticationConfiguration {
                         categories: List[String] = List.empty,
                         permissions: List[Permission] = List.empty,
                         globalPermissions: List[GlobalPermission] = List.empty)
+
+  private def parseUsers(usersLocation: String) = {
```
mproch, line 66: Good idea. Can we either:
- extract it to utils and merge somehow with logic in CustomFicusInstances.urlValueReader 
- use sth like ResourceLoader (I know that solution below is somewhat nicer, as it doesn't require us to write "file://"

mproch (approved):

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ClasspathURLStreamHandlerFactory.scala
```diff
@@ -0,0 +1,14 @@
+package pl.touk.nussknacker.engine.util.config
+
+import java.net.{URL, URLConnection, URLStreamHandler, URLStreamHandlerFactory}
+
+object ClasspathURLStreamHandlerFactory extends URLStreamHandlerFactory {
+  // Change to java.net.spi.URLStreamHandlerProvider when migrated to Java 9+
+  URL.setURLStreamHandlerFactory(this)
```
mproch, line 7: I think this is not safe. This does not give guarantees that it will be called only once per JVM (we can have different classloaders for UI and testing, also in Flink there are many jobs).
What's more, when using in Flink it can lead to classloader leak (URL is in root classloader, while this object lives in job classloader)

mproch (changes requested):

mproch (approved):

PR #1478, opened 2021-03-22
Avro encoder: Handle CharSequence enums

nadberezny:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/encode/BestEffortAvroEncoder.scala
```diff
@@ -41,11 +40,9 @@ class BestEffortAvroEncoder(avroSchemaEvolution: AvroSchemaEvolution, validation
       case (Schema.Type.RECORD, map: util.Map[String@unchecked, _]) =>
         encodeRecord(map, schema)
       case (Schema.Type.ENUM, str: String) =>
```
mproch, line 42: String is also CharSequence, so this case is no longer needed?

nadberezny: :+1:

nadberezny: https://github.com/TouK/nussknacker/pull/1478/commits/d9f02339d455c170674a7394e040159257b43406

mproch (approved):

PR #1482, opened 2021-03-22
optional outputVariable for StandaloneCustomTransformer

Filemon279:
Changing `createTransformation(outputVariable: String)` to `createTransformation(outputVariable: Option[String])` makes it possible to move extracting of `outputVariable` (or not extracting it) from `StandaloneProcessInterpreter` to `createTransformation`. 
That allows to create `StandaloneCustomTransformer` with no return value.


mproch (commented):
Good idea, can you add test/sample to show everything works ok?

Filemon279:
I have added some tests :+1:

lciolecki (approved):
LGTM

PR #1484, opened 2021-03-23
Feature: Allow admin force delete process

lciolecki:
On FE e2e tests we would like to allow for fast delete process.


mproch (approved):

lciolecki:
@JulianWielga fix it on FE side.

PR #1490, opened 2021-03-24
FIX: send cancel request to Flink when job is in non terminal state

gadomsky:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -94,6 +94,8 @@ final case class AllowDeployStateStatus(name: String) extends StateStatus
 
 final case class NotEstablishedStateStatus(name: String) extends StateStatus
 
+final case class UndeterminedStateStatus(name: String) extends StateStatus
+
```
lciolecki, line 98: `override def isUndetermined: Boolean = true` and we can change `ps.status.isInstanceOf[NotEstablishedStateStatus])` for `ps.isUndetermined`?

gadomsky: but `NotEstablishedStatus` means is not undetermined. It would make no sense... We can tell it is not established, cause we know exactly what status it is..? Or do I understand it wrong?

lciolecki: You're right.. :)

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -107,6 +112,26 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, mainClassName:
...
+      Future.successful(())
+    },
+    whenDuplicates = { _ =>
+      //TODO cancel all these jobs
+      logger.warn(s"Trying to cancel ${processName.value} which maps to multiple jobs running in Flink")
+      Future.successful(())
+    },
+    whenSingle = overview => {
+      val status = mapJobStatus(overview)
+      if (processStateDefinitionManager.statusActions(status).contains(ProcessActionType.Cancel) && isNotFinished(overview)) {
```
lciolecki, line 127: Let's move this if logic to `override protected def cancel(deploymentId: ExternalDeploymentId): Future[Unit]` ?

lciolecki (approved):
LGTM - good change!

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
gadomsky: Please see the code now

lciolecki: Great!

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -107,6 +112,29 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, mainClassName:
     }
   }
 
+  override def cancel(processName: ProcessName, user: User): Future[Unit] = {
+    def doCancel(overview: JobOverview) = {
+      val status = mapJobStatus(overview)
+      if (processStateDefinitionManager.statusActions(status).contains(ProcessActionType.Cancel) && isNotFinished(overview)) {
+        cancel(ExternalDeploymentId(overview.jid))
+      } else {
+        logger.warn(s"Trying to cancel ${processName.value} which is in status $status")
```
lciolecki, line 121: `s"Trying to cancel terminated process ${processName.value} which is in status $status".` ?

mproch (approved):

PR #1493, opened 2021-03-25
FIX: cancel FAILED periodic process

gadomsky:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
```diff
@@ -81,10 +81,10 @@ class PeriodicProcessManager(val delegate: ProcessManager,
     }
   }
 
-  private def cancelIfAlreadyDeployed(processVersion: ProcessVersion, user: User): Future[Unit] = {
+  private def cancelIfAlreadyDeployedOrFailed(processVersion: ProcessVersion, user: User): Future[Unit] = {
     findJobStatus(processVersion.processName)
       .map {
-        case Some(s) => s.isDeployed //for periodic status isDeployed=true means deployed or scheduled
+        case Some(s) => s.isDeployed || s.status.isFailed //for periodic status isDeployed=true means deployed or scheduled
```
mproch, line 87: I wonder - do we really have any states where we shouldn't deactivate? I mean, maybe it should be findJobStatus(...).map(_.isDefined)?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
```diff
@@ -96,22 +96,46 @@ class PeriodicProcessManager(val delegate: ProcessManager,
...
 
-  override def cancel(name: ProcessName, user: User): Future[Unit] = {
-    service.deactivate(name).flatMap {
-      _ => delegate.cancel(name, user)
+  private def deactivate[T](name: ProcessName)(callback: => Future[T]): Future[T] = {
+    def handleFailed(processState: Option[ProcessState], periodicDeploymentOpt: Option[PeriodicProcessDeployment]): Future[Unit] = {
+      import PeriodicProcessDeploymentStatus._
+      periodicDeploymentOpt.map { periodicDeployment =>
+        (processState.map(_.status), periodicDeployment.state.status) match {
+          case (Some(status), Failed | Deployed) if status.isFailed => service.markFailed(periodicDeployment, processState)
```
mproch, line 112: I wonder, maybe we should move this logic to PeriodicProcessService and just invoke:
handleFinished(deployedProcess: PeriodicProcessDeployment, processState: Option[ProcessState])
but "without reschedule" here? Because there is also case, when periodicDeployment is finished, and we want to mark as finished?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
gadomsky: :+1:

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
gadomsky: Thanks, I actually forgot about this case. 
Fixed and documented with test.

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -93,27 +103,37 @@ class PeriodicProcessService(delegateProcessManager: ProcessManager,
...
           // but was made invalid when rescheduling again.
           logger.error(s"Wrong periodic property, error: $error. Deactivating $deployment")
-          deactivateInternal(process.processVersion.processName)
+          deactivateAction(process.processVersion.processName)
       }
     } yield callback
   }
 
-  private def markFailed(deployment: PeriodicProcessDeployment, state: Option[ProcessState]): RepositoryAction[Unit] = {
+  def markFailed(deployment: PeriodicProcessDeployment, state: Option[ProcessState]): Future[Unit] = markFailedAction(deployment, state).run
```
mproch, line 116: I think it no longer needs to be public?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -58,7 +60,10 @@ class PeriodicProcessService(delegateProcessManager: ProcessManager,
 
     def handleSingleProcess(deployedProcess: PeriodicProcessDeployment): Future[Unit] = {
       delegateProcessManager.findJobStatus(deployedProcess.periodicProcess.processVersion.processName).flatMap { state =>
-        handleFinished(deployedProcess, state).runWithCallbacks
+        handleFinishedAction(deployedProcess, state)
+          .flatMap { finishedSuccessfully =>
+            if (finishedSuccessfully) reschedule(deployedProcess, state) else scheduledProcessesRepository.monad.pure(()).emptyCallback
```
mproch, line 65: maybe let's call it "needsReschedule"? 
```
 case _ => scheduledProcessesRepository.monad.pure(()).finished(successfully = false)
```  => this is not really unsuccessful, just unfinished? :)

mproch (commented):
:+1: almost fine for me, I added just two small naming comments

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
gadomsky: it is actually unsed now... I will delete it

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
gadomsky: yes, it means unfinished in this case.

I've change name of this type alias to NeedsReschedule

mproch (approved):
LGTM

PR #1497, opened 2021-03-30
[Feature] Handling multiple schedules in periodic manager - first version

mproch:
It's possible to define multiple named schedules (e.g. run with different parameters). Currently we don't handle running multiple Flink jobs with same ProcessName well, so there is limitation - we schedule only one process simultaneously (other schedules for same process have to wait).


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicPropertyExtractor.scala
```diff
@@ -3,16 +3,19 @@ package pl.touk.nussknacker.engine.management.periodic
...
 import java.time.Clock
 
 trait PeriodicPropertyExtractor {
-  def apply(processDeploymentData: ProcessDeploymentData): Either[String, PeriodicProperty]
+  def apply(processDeploymentData: ProcessDeploymentData): Either[String, Map[String, PeriodicProperty]]
 }
 
 object PeriodicPropertyExtractor {
 
+  val defaultScheduleName = "schedule"
```
arkadius, line 17: What's the trick with this default schedule? Will it be gone?

engine/flink/management/periodic/src/main/resources/db/batch_periodic/migration/hsql/V1_104__add_schedule_name.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "periodic_process_deployments" ADD COLUMN "schedule_name" VARCHAR(16777216) default 'schedule' not null ;
```
lciolecki, line 1: 1. It's the same as postgresql? So we can put it in default migration?
2. Do we realy need there `VARCHAR(16777216)` It looks like 255 / 512 ?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -32,21 +34,26 @@ class PeriodicProcessService(delegateProcessManager: ProcessManager,
...
-        jarManager.prepareDeploymentWithJar(processVersion, processJson).flatMap { deploymentWithJarData =>
-          scheduledProcessesRepository.create(deploymentWithJarData, periodicProperty, runAt).flatMap { data =>
-            handleEvent(ScheduledEvent(data, firstSchedule = true))
-          }.run
-        }
-      case Right(None) =>
-        Future.failed[Unit](new PeriodicProcessException(s"No future date determined by $periodicProperty"))
-      case Left(error) =>
-        Future.failed[Unit](new PeriodicProcessException(error))
+    //FIXME:
```
lciolecki, line 40: FIXME: what ?:)

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessDeploymentsTable.scala
```diff
@@ -26,13 +26,15 @@ trait PeriodicProcessDeploymentsTableFactory extends PeriodicProcessesTableFacto
...
+    def scheduleName: Rep[String] = column[String]("schedule_name", NotNull)
+
     def deployedAt: Rep[Option[LocalDateTime]] = column[Option[LocalDateTime]]("deployed_at")
 
     def completedAt: Rep[Option[LocalDateTime]] = column[Option[LocalDateTime]]("completed_at")
 
     def status: Rep[PeriodicProcessDeploymentStatus] = column[PeriodicProcessDeploymentStatus]("status", NotNull)
 
-    override def * : ProvenShape[PeriodicProcessDeploymentEntity] = (id, periodicProcessId, createdAt, runAt, deployedAt, completedAt, status) <>
+    override def * : ProvenShape[PeriodicProcessDeploymentEntity] = (id, periodicProcessId, createdAt, runAt, scheduleName, deployedAt, completedAt, status) <>
```
lciolecki, line 37: I wonder about deployments.. Right now only periodic process has deployments? What about normal process? - I know that is not part of your changes but.. :)

lciolecki (approved):
LGTM, but please look for my comments.

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicPropertyExtractor.scala
mproch: Mostly for backward compatibility and for simple cases where we have only one schedule

arkadius: I wonder if we can get rid of backward compatibility, make this multiple schedules approach as easy that it won't be needed maintenance of two schemes of creation of schedules?

engine/flink/management/periodic/src/main/resources/db/batch_periodic/migration/hsql/V1_104__add_schedule_name.sql
mproch: I changed it...

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
mproch: I removed it...

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicPropertyExtractor.scala
mproch: I ended up having sealed trait... I still think in many cases single schedule would be better...

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -34,26 +36,59 @@ class PeriodicProcessService(delegateProcessManager: ProcessManager,
...
 
-  def schedule(periodicProperty: PeriodicProperty,
+  def schedule(schedule: BasePeriodicProperty,
                processVersion: ProcessVersion,
                processJson: String): Future[Unit] = {
-    periodicProperty.nextRunAt(clock) match {
-      case Right(Some(runAt)) =>
-        logger.info("Scheduling periodic process: {} on {}", processVersion, runAt)
+    findInitialScheduleDates(schedule) match {
+      case Right(scheduleDates) =>
```
gadomsky, line 43: what about the case when all cron expressions have no match in future? do we want to allow to schedule such a process?

mproch: :+1: I forgot to keep previous behaviour... I added test

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProperty.scala
```diff
@@ -25,6 +31,9 @@ object PeriodicProperty {
   def nextRunAt(clock: Clock): Either[String, Option[LocalDateTime]]
 }
 
+@JsonCodec case class ComplexPeriodicProperty(schedules: Map[String, PeriodicProperty]) extends BasePeriodicProperty
```
arkadius, line 34: Can it be `MultipleSchedulesProperty`? I can't see any complexity here ;)

arkadius (commented):
LGTM, one naming issue I've found. I know that it will need some migration but IMO it will cost us less if we do it now then spend time talking why it is named that way.

arkadius (approved):
LGTM but let's someone with more production experiences take a look

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProperty.scala
mproch: done

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProperty.scala
```diff
@@ -11,11 +11,17 @@ import io.circe.generic.extras.{Configuration, ConfiguredJsonCodec}
...
+}
+
+@ConfiguredJsonCodec sealed trait BasePeriodicProperty
+
 object PeriodicProperty {
   implicit val configuration: Configuration = Configuration.default.withDefaults
 }
 
-@ConfiguredJsonCodec sealed trait PeriodicProperty {
+@ConfiguredJsonCodec sealed trait PeriodicProperty extends BasePeriodicProperty {
```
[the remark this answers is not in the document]
mproch: done

gadomsky (approved):
LGTM

PR #1506, opened 2021-04-07
Publish standalone/app as docker image

kdutkowski:


engine/standalone/app/src/universal/conf/application.conf
```diff
@@ -0,0 +1,8 @@
+standaloneConfig {
```
arkadius, line 1: I think that we need to add comment here, that purpose of this config is to handle some very basic case when someone mount own very simple model via volumes - model doesn't need any configuration. For more advanced usage, developer should write own config.

engine/standalone/app/src/universal/bin/nussknacker-standalone-entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+MODELS_DIR="$NUSSKNACKER_DIR/models"
+
+CONFIG_FILE=${NUSSKNACKER_CONFIG_FILE-${2-"$CONF_DIR/application.conf"}}
+LOG_FILE=${NUSSKNACKER_LOG_FILE-${3-"$CONF_DIR/docker-logback.xml"}}
+APPLICATION_APP=${NUSSKNACKER_APPLICATION_APP-${4-"pl.touk.nussknacker.engine.standalone.http.StandaloneHttpApp"}}
+USER=${DAEMON_USER-${5-"daemon"}}
+GROUP=${DAEMON_GROUP-${6-"daemon"}}
+
+mkdir -p ${STORAGE_DIR}/logs
+mkdir -p ${STORAGE_DIR}/db
```
arkadius, line 16: standalone doesn't have db as nk. The only db is `/standaloneProcesses` directory

build.sbt
```diff
@@ -384,9 +390,25 @@ lazy val engineStandalone = (project in engine("standalone/engine")).
   ).
   dependsOn(interpreter % "provided", standaloneUtil, httpUtils % "provided", testUtil % "it,test")
 
+lazy val standaloneDockerSettings = {
+  val workingDir = "/opt/nussknacker"
+
+  commonDockerSettings ++ Seq(
+    dockerEntrypoint := Seq(s"$workingDir/bin/nussknacker-standalone-entrypoint.sh", dockerPort.toString),
+    dockerExposedVolumes := Seq(s"$workingDir/storage", s"$workingDir/data"),
```
arkadius, line 398: The only storage is `/standaloneProcesses`, BTW maybe we should reconfigure it to some `/opt/standalone/processes` to be consistent with nk convention?

build.sbt
```diff
@@ -384,9 +390,25 @@ lazy val engineStandalone = (project in engine("standalone/engine")).
   ).
   dependsOn(interpreter % "provided", standaloneUtil, httpUtils % "provided", testUtil % "it,test")
 
+lazy val standaloneDockerSettings = {
+  val workingDir = "/opt/nussknacker"
+
+  commonDockerSettings ++ Seq(
+    dockerEntrypoint := Seq(s"$workingDir/bin/nussknacker-standalone-entrypoint.sh", dockerPort.toString),
```
arkadius, line 397: I think that this port argument isn't handled. I see that by default exposed ports are:
```
      - "8070:8070"
      - "8071:8080"
```
IMO we don't need to "expose" it by entrypoint but can use standard Dockerfile expose directive. It will help developers find out on which ports are available and some comment will help to find out which one is for what.

arkadius (commented):
Nice work

build.sbt
kdutkowski: I'll add both processes and management ports through `dockerExposedPorts`

build.sbt
kdutkowski: Or maybe `/opt/processes` is enough? It's a path inside standalone container

kdutkowski: `"/opt/standalone/processes"` or `"/opt/nussknacker/processes"`? Or maybe ``"/opt/nussknacker-standalone/processes"`` ;)

kdutkowski: Don't we need `"$workingDir/storage"` for `logs` directory?

build.sbt
```diff
@@ -257,23 +258,12 @@ val javaxValidationApiV = "2.0.1.Final"
 val caffeineCacheV = "2.8.8"
 val sttpV = "2.2.9"
 
-lazy val dockerSettings = {
-  val workingDir = "/opt/nussknacker"
-
+lazy val commonDockerSettings = {
```
mproch, line 264: dockerPort and authentication related envVars should be only in UI part

build.sbt
```diff
@@ -297,7 +287,23 @@ lazy val dockerSettings = {
       List(dockerVersion, updateLatest, latestBranch, dockerTagName)
         .map(tag => alias.withTag(tag.map(sanitize)))
         .distinct
-    },
+    }
+  )
+}
+
+lazy val distDockerSettings = {
+  val workingDir = "/opt/nussknacker"
```
mproch, line 297: workingDir, defaultLinuxInstallLocation and exposedVolumes can be in common?

kdutkowski: Well, after Arek's comments I added a /processes volume specific to standalone. So now `exposedVolumes` is not the same for both images. Plus I need workingDir to construct entrypoint path, which is specific for each image. 

As for `defaultLinuxInstallLocation` it probably could, or even should be common. But it depends on workingDir, so I'd have to either make it a global constant in the file or define common settings as a function of workingDir instead of a simple Seq. Do you think it's worth it?

engine/standalone/app/src/universal/conf/application.conf
```diff
@@ -1,8 +1,10 @@
+; This config works in a simple scenario when the model provided via volumes doesn't need any configuration values
```
arkadius, line 1: Please use hash instead of semicolon. It's not windows ini file! ;) BTW this is good stuff worth to use not only in scala, take a look at full feature list: https://github.com/lightbend/config#features-of-hocon

build.sbt
arkadius: `/opt/nussknacker/processes` is fine :+1: . Regards to storage - you've right, we have logs inside so it is ok. BTW logs are not stored there because we have only console appender (in nk and standalone) but lets it stay as it is. Maybe it is for some advanced configurations of logging.

engine/standalone/app/src/universal/bin/nussknacker-standalone-entrypoint.sh
```diff
@@ -2,6 +2,7 @@
 
 NUSSKNACKER_DIR=`dirname "$0" | xargs -I{} readlink -f {}/..`
 export STORAGE_DIR="$NUSSKNACKER_DIR/storage"
+PROCESSES_DIR="$NUSSKNACKER_DIR/processes"
```
arkadius, line 5: IMO process should be nested in storage. It is the category of files that you would like to keep on some external volume and be sure that after container recreation it won't disappear

arkadius (approved):
LGTM

PR #1510, opened 2021-04-08
Refactoring: FlinkSource API and test source API.

gskrobisz:


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+import pl.touk.nussknacker.engine.api.context.ValidationContext
+import pl.touk.nussknacker.engine.api.typed.typing
+
+/**
+  * FlinkContextInitializer provides definition of Context returned within DataStream by FlinkSource
+  * and map function that transforms raw input to Context.
+  *
+  * @tparam T - type of event that is generated by raw source.
+  */
+abstract class FlinkContextInitializer[T] extends Serializable {
```
arkadius, line 15: IMO those two classes should be in the same file/package as `FlinkIntermediateRawSource` because they are connected. Reading docs you don't know here what is "raw source".

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def validationContext(context: ValidationContext, name: String, result: typing.TypingResult)(implicit nodeId: NodeId): ValidationContext
```
arkadius, line 17: 1. Why we need this `name` here? Is it a name in case when we produce only one variable?
2. Why we need `result` here? What is it?
3. Shouldn't both be passed as a parameters of constructor in very specific case whene source produces only one variable?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/InitContextFunction.scala
```diff
@@ -0,0 +1,21 @@
...
+import pl.touk.nussknacker.engine.api.{Context, VariableConstants}
+
+/**
+  * Maps raw source event to Context where default "input" variable represents the event.
+  *
+  * @param processId - id of the process or flink job, used to setup name of created Context
+  * @param taskName - name of the task within the process or flink job, used to setup name of created Context
+  * @tparam T - type of event that is generated by raw source
+  */
+case class InitContextFunction[T](processId: String, taskName: String) extends RichMapFunction[T, Context] with ContextInitializingFunction {
```
arkadius, line 14: It looks like `BasicContextInitializingFunction`. IMO should be in the same file as `BasicFlinkContextInitializer` because they are connected.

arkadius (commented):
Good change, some comments added.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -37,12 +37,12 @@ private[confluent] class ConfluentAvroMessageReader(schemaRegistryClient: Schema
...
 
   // TODO: This implementation won't handle separator escaping
-  def readMessage(str: String, keySchema: Schema, valueSchema: Schema): ProducerRecord[Array[Byte], Array[Byte]] = {
+  def readMessage(str: String, keySchema: Schema, valueSchema: Schema): ConsumerRecord[Array[Byte], Array[Byte]] = {
     try {
       if (!parseKey) {
         val value = jsonToAvro(str, valueSchema)
         val serializedValue = serializeImpl(valueSubject, value)
-        new ProducerRecord(topic, serializedValue)
+        new ConsumerRecord(topic, 0, 0L, Array[Byte](), serializedValue)
```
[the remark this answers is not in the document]
gskrobisz: Fixed.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
```diff
@@ -4,55 +4,91 @@ import org.apache.flink.api.common.typeinfo.TypeInformation
...
+/**
+  * Source with typical source stream trasformations:
+  *  - adds source using provided SourceFunction (here raw source produces raw data)
+  *  - sets UID
+  *  - assigns timestamp and watermarks
+  *  - initializes Context that is streamed within output DataStream (here raw data are mapped to Context)
+  * It separates raw event data produced by SourceFunction and data released to the stream as Context variables.
+  * By default it uses simple implementation of initializer, see [[pl.touk.nussknacker.engine.flink.util.context.BasicFlinkContextInitializer]].
+  *
+  * @tparam T - type of event that is generated by this source. This is needed to handle e.g. syntax suggestions in UI
```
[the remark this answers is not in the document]
gskrobisz: Changed T -> Raw and fixed description.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/InitContextFunction.scala
gskrobisz: Fixed. Moved `InitContextFunction` to the same file as `BasicFlinkContextInitializer`.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
gskrobisz: Fixed. Moved to `FlinkSource` and `FlinkIntermediateRawSource` file.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
gskrobisz: 1 & 2. Changed name -> outputVariableName and result -> outputVariableType and fixed documentation. Those parameters represent default one variable produced by the source.
3. For multiple variables see `GenericSourceWithCustomVariablesSample`. There is one `val customContextInitilalizer` which is used in two places:
- creating the source
- contextTransformation - This is the place where I know all variables and their types (especially in avro sources), and pass them as parameters to the method of already created object. 
I see two alternatives:
1. pass a List of variable name-and-type pairs
2. initialize `var customContextInitilalizer` with new FlinkContextInitializer inside contextTransformation
WDYT?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
```diff
@@ -1,58 +1,125 @@
...
 
 /**
   * Source with methods specific for Flink
-  * @tparam T - type of event that is generated by this source. This is needed to handle e.g. syntax suggestions in UI
+  *
+  * @tparam Raw - type of raw event that is generated by flink source function.
   */
-trait FlinkSource[T] extends Source[T] {
+//TODO: remove type T parameter
+trait FlinkSource[Raw] extends Source[Raw] {
```
arkadius, line 24: I think that here shouldn't be Raw - Raw is only in case of `FlinkIntermediateRawSource` / `FlinkContextInitializer`. In simple, basic case this `T` is final output variable type

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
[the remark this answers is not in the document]
arkadius: Sorry for being not precise, by "this description is correct?" I was meant the only "This is needed to handle e.g. syntax suggestions in UI" part. You removed it so it wasn't correct, right? (I'm asking because I don't know for sure). Thanks for very precise description BTW.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
arkadius: I see that not exaclty. Have you pushed everything? Because I still see both `FlinkIntermediateRawSource` and `FlinkContextInitializer` in `FlinkSource.scala`. Can you extract them to separate `FlinkIntermediateRawSource` file like you wrote in comment?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
arkadius: Ok, I see now. WDYT about passing to this `validationContext` method everything what we have in `TransformationStep`? I know that there are type dependent types but we can have `parameters: List[(String, BaseDefinedParameter)], state: Option[Any]` and someone could do some pattern matching on it. WDYT? Thanks to that we will have the whole power of `GenericNodeTransformation` in this place.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/InitContextFunction.scala
arkadius: Ok, and what about chnaging a name `InitContextFunction` -> `BasicContextInitializingFunction`? WDYT?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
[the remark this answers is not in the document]
gskrobisz: I've reverted this part about syntax suggestions, it is used like that for sources with explicit MethodToInvoke annotation.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/FlinkContextInitializer.scala
gskrobisz: Fixed. Extracted both to separate file.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/util/context/InitContextFunction.scala
gskrobisz: Fixed, name is changed.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkSource.scala
gskrobisz: Good point. Reverted to `FlinkSource[T]`.

arkadius (approved):

PR #1512, opened 2021-04-08
Add kafka source with additional variable with metadata.

gskrobisz:


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
```diff
@@ -0,0 +1,34 @@
...
+  def withType[K: ClassTag]: typing.TypingResult = {
+    val keyTypingResult = Typed[K]
+    TypedObjectTypingResult(fields(keyTypingResult), objType(keyTypingResult))
+  }
+
+  def withType(keyTypingResult: typing.TypingResult): typing.TypingResult = {
+    TypedObjectTypingResult(fields(keyTypingResult), objType(keyTypingResult))
+  }
+
+  private def fields(keyTypingResult: typing.TypingResult): Map[String, TypingResult] = {
```
mproch, line 21: maybe we could just introduce case class here?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaGenericNodeSourceFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+                                                              formatter: RecordFormatter,
+                                                              processObjectDependencies: ProcessObjectDependencies)
+  extends BaseKafkaSourceFactory[ConsumerRecord[K, V]](deserializationSchemaFactory, timestampAssigner, formatter, processObjectDependencies)
+    with SingleInputGenericNodeTransformation[FlinkSource[ConsumerRecord[K, V]]] {
+
+  protected val customContextInitializer: KafkaGenericContextInitializer[K, V, DefinedParameter, State] = new KafkaGenericContextInitializer[K, V, DefinedParameter, State]
+
+  override type State = Nothing
+
+  override def initialParameters: List[Parameter] = Parameter[String](KafkaGenericNodeSourceFactory.TopicParamName)
```
mproch, line 41: So this is only for "generic" topic, without avro and the likes?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
arkadius: We already have `InputMeta` case class. The problem here is that we mix type of key with primitive fields. IMO this way is ok, to resolve this problem. This companion object is in the same file as class so probability of desynchornization is minimal.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
```diff
@@ -0,0 +1,88 @@
...
+@nowarn("cat=deprecation")
+class ConsumerRecordToJsonFormatter extends RecordFormatter {
+
+  private val cs: Charset = StandardCharsets.UTF_8
+
+  // TODO: add better key-value encoding and decoding
+  implicit val consumerRecordEncoder: Encoder[ConsumerRecord[Array[Byte], Array[Byte]]] = {
+    val encode: Encoder[Any] = BestEffortJsonEncoder(failOnUnkown = false).circeEncoder
+    new Encoder[ConsumerRecord[Array[Byte], Array[Byte]]] {
+      override def apply(a: ConsumerRecord[Array[Byte], Array[Byte]]): Json = Json.obj(
```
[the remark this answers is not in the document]
arkadius: I aggree that we should just encode/decode only fields that are used in `InputMeta`.  If it can be handled by json semiauto devriation, it would be perfect. Be aware @Grzesiek, that this jsons sometimes will be hand-written by users - if they need to prepare own tests inputs.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaGenericNodeSourceFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+  * that allows custom ValidationContext and Context transformations, which are provided by [[pl.touk.nussknacker.engine.kafka.source.KafkaGenericContextInitializer]]
+  *
+  * @param deserializationSchemaFactory - produces KafkaDeserializationSchema for [[pl.touk.nussknacker.engine.kafka.source.KafkaSource]]
+  * @param timestampAssigner            - provides timestampAsigner and WatermarkStrategy to KafkaSource
+  * @param formatter                    - support for test data parser and generator
+  * @param processObjectDependencies
+  * @tparam K - type of key of kafka event that is generated by raw source (SourceFunction).
+  * @tparam V - type of value of kafka event that is generated by raw source (SourceFunction).
+  */
+class KafkaGenericNodeSourceFactory[K: ClassTag, V: ClassTag](deserializationSchemaFactory: KafkaDeserializationSchemaFactory[ConsumerRecord[K, V]],
```
arkadius, line 30: 1. Why you declare here class tags? I can't find usage of them.
2. I don't like `genericnode` in this name.  IMO developers without background that we initially didn't have generic nodes and meantime they were introduced, will be confused why this class is more generic then some other. I have other options:
a) let's skip phasing out of "old" `KafkaSourceFactory` and call this just `KafkaSourceFactory`
b) let's call it `KafkaSourceFactory_v2`
c)  let's call it `KafkaSourceFactoryWithMeta` or sth similar to name purpose of this class instead how it is implemented.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
```diff
@@ -26,6 +26,8 @@ import scala.reflect.ClassTag
   *   - KafkaSourceFactory - `topic` parameter has to be passed on frontend
   *   - SingleTopicKafkaSourceFactory - topic is defined on level of configuration
   *
+  * Used to produce simple sources with single input variable.
```
arkadius, line 29: Do we add depreaction for this class? Why someone wil use it if he/she has better approach?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSchemaFactory.scala
```diff
@@ -0,0 +1,60 @@
...
+          record.serializedValueSize(),
+          key,
+          value,
+          record.headers()
+        )
+      }
+
+      override def isEndOfStream(nextElement: ConsumerRecord[K, V]): Boolean = false
+
+      override def getProducedType: TypeInformation[ConsumerRecord[K, V]] = TypeInformation.of(clazz)
```
arkadius, line 56: What about this? Have you find out the way how to implement it correctly? We need to have at least TODO here, because after this change our effort to avoid kryo in serialization of value will go to trash

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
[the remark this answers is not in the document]
gskrobisz: Fixed (and introduced case class `SerializableConsumerRecord`)

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
gskrobisz: Fixed (and introduced case class SerializableConsumerRecord)

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaGenericNodeSourceFactory.scala
gskrobisz: 1. Class tags are required by `KafkaContextInitializer` to set proper `ValidationContext`
2. I've chosen option a), hence only one `KafkaSourceFactory`

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
gskrobisz: Fixed, skipped phasing out and replaced `KafkaSourceFactory`

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
```diff
@@ -1,104 +1,76 @@
...
+  * @param processObjectDependencies
+  * @tparam K - type of key of kafka event that is generated by raw source (SourceFunction).
+  * @tparam V - type of value of kafka event that is generated by raw source (SourceFunction).
   * */
-class KafkaSourceFactory[T: ClassTag](deserializationSchemaFactory: KafkaDeserializationSchemaFactory[T],
-                                      timestampAssigner: Option[TimestampWatermarkHandler[T]],
-                                      formatter: RecordFormatter,
-                                      processObjectDependencies: ProcessObjectDependencies)
-  extends BaseKafkaSourceFactory(deserializationSchemaFactory, timestampAssigner, formatter, processObjectDependencies) {
+class KafkaSourceFactory[K: ClassTag, V: ClassTag](deserializationSchemaFactory: KafkaDeserializationSchemaFactory[ConsumerRecord[K, V]],
```
arkadius, line 37: I see that after this change we will lost `BaseKafkaSourceFactory`, that had ability to specify more than one topic. Maybe we don't need it anymore but we should think a moment if we really don't need it and if it will be hard to implement it on user side.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
```diff
@@ -0,0 +1,87 @@
...
+    new ConsumerRecord(topic, partition, offset, timestamp,
+      TimestampType.NO_TIMESTAMP_TYPE, ConsumerRecord.NULL_CHECKSUM.longValue(),
+      ConsumerRecord.NULL_SIZE, ConsumerRecord.NULL_SIZE,
+      key, value, headers
+    )
+  }
+
+  def from[K, V](topic: String, record: SerializableConsumerRecord[K, V]): ConsumerRecord[K, V] = {
+    createConsumerRecord(
+      record.topic.getOrElse(topic),
```
arkadius, line 78: It is counter-intuitive that topic specified in record has higher priority than `record.topic`. WHy this method is needed? Why not to just `serializableConsumerRecord.copy(topic = Some(otherTopic)).toConsumerRecord)`?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSchemaFactory.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      @transient
+      private lazy val keyDeserializer = createKeyDeserializer(kafkaConfig)
+      @transient
+      private lazy val valueDeserializer = createValueDeserializer(kafkaConfig)
+
+      override def deserialize(record: ConsumerRecord[Array[Byte], Array[Byte]]): ConsumerRecord[K, V] = {
+        val key = keyDeserializer.deserialize(record.topic(), record.key())
+        val value = valueDeserializer.deserialize(record.topic(), record.value())
+        new ConsumerRecord[K, V](
```
arkadius, line 44: Why you haven't used constructor with `leaderEpoch` and why not to also put it to `InputMeta`?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
```diff
@@ -0,0 +1,87 @@
...
+  override protected def testDataSplit: TestDataSplit = TestParsingUtils.newLineSplit
+
+}
+
+
+case class SerializableConsumerRecord[K, V](key: Option[K], value: V, topic: Option[String], partition: Option[Int], offset: Option[Long], timestamp: Option[Long], headers: Option[Map[String, Option[String]]])
+
+object SerializableConsumerRecord {
+
+  def createConsumerRecord[K, V](topic: String, partition: Int, offset: Long, timestamp: Long, key: K, value: V, headers: Headers): ConsumerRecord[K, V] = {
```
arkadius, line 68: Am I right that it is only used on data extracted from `SerializableConsumerRecord`? If it so, why not too have just a method `toKafkaConsumerRecord` inside `SerializableConsumerRecord`.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
arkadius: Also I see now that this method has cheated me - take a look for full invocation: looking on `SerializableConsumerRecord from (....)` you will expect that the result will be `SerializableConsumerRecord`, not kafka `ConsumerRecord`

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
```diff
@@ -0,0 +1,87 @@
...
+    val serializableRecord = SerializableConsumerRecord(
+      Option(deserializedRecord.key()),
+      deserializedRecord.value(),
+      Option(deserializedRecord.topic()),
+      Option(deserializedRecord.partition()),
+      Option(deserializedRecord.offset()),
+      Option(deserializedRecord.timestamp()),
+      Option(ConsumerRecordUtils.toMap(deserializedRecord.headers()).mapValues(s => Option(s)))
+    )
+    implicitly[Encoder[SerializableConsumerRecord[K, V]]].apply(serializableRecord).noSpaces.getBytes(StandardCharsets.UTF_8)
```
arkadius, line 39: Why you explicitly derive encoder, make it implicit and after that implicitly derive it again? After making it implicit, you can use just `serializableRecord.asJson` or if you don't like implicit extensions (Ops syntax) you can remove implicit from `consumerRecordEncoder` declaration and use just `consumerRecordEncoder(serializableRecord)`

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/sources.scala
```diff
@@ -25,8 +27,15 @@ object sources {
 
   import collection.JavaConverters._
 
-  class GenericJsonSourceFactory(processObjectDependencies: ProcessObjectDependencies) extends KafkaSourceFactory[java.util.Map[_, _]](
-    JsonMapDeserialization, None, JsonRecordFormatter, processObjectDependencies)
+  class EspValueDeserializaitionSchemaFactory[T](schema: DeserializationSchema[T]) extends ConsumerRecordDeserializationSchemaFactory[String, T]{
```
arkadius, line 30: Can you name it more descriptive like `FixedValueDeserializaitionSchemaFactory` and move it to the same package as `ConsumerRecordDeserializationSchemaFactory`? Also can you rename `T` to `V` and `schema` to `valueSchema`?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
gskrobisz: In our project we already at least have one reference to `BaseKafkaSourceFactory`. Here I assume that multi-topic option can be achieved by implementing proper `override def initialParameters` (and now I see that in `def implementation` I could expose method to extract topics from `params` to allow this multi-topic).

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSchemaFactory.scala
gskrobisz: WDYT about adding `timestampType` (with values `NO_TIMESTAMP_TYPE(-1, "NoTimestampType"), CREATE_TIME(0, "CreateTime"), LOG_APPEND_TIME(1, "LogAppendTime")`)?

arkadius: IMO more information is always better option then less in case of meta data :+1:

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
gskrobisz: More information in `MigrationGuide.md`.

docs/MigrationGuide.md
```diff
@@ -58,10 +58,21 @@ To see biggest differences please consult the [changelog](Changelog.md).
...
+* [#1512](https://github.com/TouK/nussknacker/pull/1512) Replaced `KafkaSourceFactory` with source based on `GenericNodeTransformation`, which gives access to setup of `ValidationContext` and `Context` initialization.
+  It uses `KafkaContextInitializer` to initialize `Context` with additional variable containing kafka event metadata. Source factory requires deserialization to `ConsumerRecord` (see `ConsumerRecordDeserializationSchemaFactory`).
+  To migrate `KafkaSourceFactory`:
+  - provide deserializer factory:
+    - use `ConsumerRecordDeserializationSchemaFactory` with current `DeserializationSchema` as a value deserializer, add key deserializer
+    - or use `FixedValueDeserializaitionSchemaFactory` with simple key-as-string deserializer
+  - current `RecordFormater` should be sufficient for value-only serialization, or use `ConsumerRecordToJsonFormatter` for metadata serialization
+  - provide timestampAssigner that is able to extract time from `ConsumerRecord[K, V]`
+  Also:
+  - removed `BaseKafkaSourceFactory` with multiple topics support: use `KafkaSourceFactory` instead, see test "source with two input topics"
```
arkadius, line 74: can you fix the order? should it be `see "source with two input topics" test case`

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
gskrobisz: Fixed according to our discussion.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatter.scala
gskrobisz: Fixed, according to our discussion.

docs/MigrationGuide.md
gskrobisz: Fixed.

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
```diff
@@ -20,3 +20,12 @@ trait TypeInformationDetection extends Serializable {
   def forValueWithContext[T](validationContext: ValidationContext, value: TypingResult): TypeInformation[ValueWithContext[T]]
 
 }
+
+/**
+  * Trait that allows for providing more details TypeInformation when TypingResult is known.
+  */
+trait TypeInformationDetectionForTypingResult extends TypeInformationDetection {
```
arkadius, line 27: 1. Why not to merge it with TypeInformationDetection? 2. Maybe it could have better name because now it looks like it only detect type for typing result (fot context and so on - no)

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
```diff
@@ -0,0 +1,102 @@
...
+      override def createSerializer(config: ExecutionConfig): TypeSerializer[InputMeta[K]] =
+        new ScalaCaseClassSerializer[InputMeta[K]](classOf[InputMeta[K]], fieldTypes.map(_.createSerializer(config)).toArray)
+    }
+  }
+}
+
+/**
+  * Customisation for TypeInformationDetection that provides type information for BaseInputMeta.
+  */
+class InputMetaAwareTypeInformationCustomisation extends TypingResultAwareTypeInformationCustomisation {
```
arkadius, line 96: Shouldn't it be used in generic/demo/dev model?

arkadius (approved):
LGTM, have some comments to type info detection because I didn't looked into it so far, maybe @mproch also take a look on it

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
mproch: :+1: If it extends TypeInformationDetection I think it's better to move this method to forType?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
mproch: :+1: Currently it does not seem to be used anywhere? At least we should register it with ServiceLoader?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
gskrobisz: Fixed. Removed TypeInformationDetectionForTypingResult, method forType moved to TypeInformationDetection.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
gskrobisz: Fixed. See `META-INF/services` in `flinkKafkaUtil` and `process`. Example test scenario is in `InputMetaDeserializationSpec`.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSchemaFactory.scala
```diff
@@ -0,0 +1,68 @@
...
+  * Produces deserialization schema that describes how to turn the Kafka raw [[org.apache.kafka.clients.consumer.ConsumerRecord]]
+  * (with raw key-value of type Array[Byte]) into deserialized ConsumerRecord[K, V] (with proper key-value types).
+  * It allows the source to provide event value AND event metadata to the stream.
+  *
+  * Deprecated annotations for checksum field mapping, to fulfill constructor requirements.
+  *
+  * @tparam K - type of key of deserialized ConsumerRecord
+  * @tparam V - type of value of deserialized ConsumerRecord
+  */
+@silent("deprecated")
```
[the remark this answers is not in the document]
gskrobisz: Fixed. ConsumerRecord marks field `checksum` as deprecated. We do not need to use this field, therefore here I replaced it with default NULL_CHECKSUM value.

PR #1514, opened 2021-04-09
[Improvement] Show DeploymentData in Flink console

mproch:


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/NkGlobalParameters.scala
```diff
@@ -10,25 +9,32 @@ import org.apache.flink.api.common.ExecutionConfig.GlobalJobParameters
...
 
   //here we decide which configuration properties should be shown in REST API etc.
   //For now it will be only deployment information
-  //NOTE: this information is used in FlinkRestManager - any changes here should be reflected there
+  //NOTE: some of the information is used in FlinkRestManager - any changes here should be reflected there
   override def toMap: util.Map[String, String] = {
-    //we wrap in HashMap because .asJava creates not-serializable map in 2.11
-    new util.HashMap(Map[String, String](
+    val deploymentInfo = deploymentData.additionalDeploymentData.map {
+      case (k, v) => s"deployment.properties.$k" -> v
```
arkadius, line 27: I wonder how to handle situation when it will came something big in those properties? WDYT? I know that we have the same problem in `configMap` but there are rather static values and here someone can put e.g. something generated on manager site. Maybe we should extract protected method like `extractDeploymentDataToMap`? But still we have case class that shouldn't be extendend. Maybe we should have `AbstractNkGlobalParameters` and `NkGlobalParameters` as a standard implementation?

mproch: I extracted part of serialization to strategy.

arkadius (approved):

PR #1515, opened 2021-04-09
Avro sink enhancements

nadberezny:
1. Remove unnecessary parameter: ValidationMode
2. Refactor: pass transformation state instead of var mutation

### DIFF
--- a/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactoryWithEditor.scala
+++ b/engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactoryWithEditor.scala
@@ -9,8 +9,9 @@ import pl.touk.nussknacker.engine.api.process.{ProcessObjectDependencies, SinkFa
 import pl.touk.nussknacker.engine.api.{LazyParameter, MetaData}
 import pl.touk.nussknacker.engine.avro.schemaregistry.SchemaRegistryProvider
 import pl.touk.nussknacker.engine.avro.{KafkaAvroBaseTransformer, SchemaDeterminerErrorHandler}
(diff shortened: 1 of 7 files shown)


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactoryWithEditor.scala
```diff
@@ -85,12 +85,11 @@ class KafkaAvroSinkFactoryWithEditor(val schemaRegistryProvider: SchemaRegistryP
...
     val schemaUsedInRuntime = schemaDeterminer.toRuntimeSchema(schemaData)
 
-    val validationMode = extractValidationMode(params(SinkValidationModeParameterName).asInstanceOf[String])
-    val sinkValue = AvroSinkValue.applyUnsafe(sinkValueParameter.get, parameterValues = params)
+    val sinkValue = AvroSinkValue.applyUnsafe(state.sinkValueParameter, parameterValues = params)
     val key = params(SinkKeyParamName).asInstanceOf[LazyParameter[CharSequence]]
 
     new KafkaAvroSink(preparedTopic, versionOption, key, sinkValue, kafkaConfig, schemaRegistryProvider.serializationSchemaFactory,
-      schemaData.serializableSchema, schemaUsedInRuntime.map(_.serializableSchema), clientId, validationMode)
+      schemaData.serializableSchema, schemaUsedInRuntime.map(_.serializableSchema), clientId, ValidationMode.strict)
```
mproch, line 92: I wonder, does it have any effect/causes problem when "Latest version" is chosen? E.g. when we want to handle compatible changes?

arkadius: I don't see issue with it. The idea is that our schema evolution mechanism handle this situation. From user perspective he/she should see definition of latest version. I see other problem - with handling defaults but will left comment in the place where it exists.

arkadius (approved):

PR #1516, opened 2021-04-09
Process rename update

JulianWielga:


lciolecki (approved):
LGTM

ui/client/actions/nk/processUtils/saveProcess.ts
```diff
@@ -1,33 +1,35 @@
...
   return async (dispatch, getState) => {
     const state = getState()
-    const processId = getProcessId(state)
     const processJson = getProcessToDisplay(state)
 
     // save changes before rename and force same processId everywhere
-    await HttpService.saveProcess(processId, {...processJson, id: processId}, comment)
+    await HttpService.saveProcess(processJson.id, processJson, comment)
+
+    const isRenamed = isProcessRenamed(state) && await doRenameProcess(processJson.id, processJson.newId)
```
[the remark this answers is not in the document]
mproch: it's not await as in Scala :) 
It's syntactic sugar on promises - there's no blocking, promises are passed up the invocation chain, so you have to take care of them...

ui/client/components/graph/node-modal/NodeDetailsContent.js
```diff
@@ -537,7 +537,7 @@ export class NodeDetailsContent extends React.Component {
       (newValue) => this.setNodeDataAt(fieldProperty, newValue, defaultValue),
       readonly,
       this.isMarked(fieldProperty),
-      key,
+      key || fieldProperty || fieldLabel,
```
mproch, line 540: Why is it needed here?

ui/client/components/modals/SaveProcessDialog.tsx
```diff
@@ -2,7 +2,7 @@ import React, {ChangeEvent, useCallback, useState} from "react"
 import {useTranslation} from "react-i18next"
 import {useDispatch, useSelector} from "react-redux"
 import {saveProcess} from "../../actions/nk"
-import {getProcessId, getProcessToDisplay, isProcessRenamed} from "../../reducers/selectors/graph"
+import {getProcessId, getProcessNewId, isProcessRenamed} from "../../reducers/selectors/graph"
```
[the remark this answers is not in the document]
mproch: yes, "process.id" is due to "historical" reasons, let's use processName in selectors etc.

ui/client/components/toolbars/status/ProcessInfo.tsx
```diff
@@ -122,7 +128,13 @@ class ProcessInfo extends React.Component<OwnProps & StateProps, State> {
                   {icon}
                 </div>
                 <div className={"process-info-text"}>
-                  <div className={"process-name"}>{process.name}</div>
+                  {isRenamePending ?
+                    (
+                      <div className="process-name process-name-rename" title={process.name}>{nextId}*</div>
```
mproch, line 133: What effect do we want to achieve with this "title"?

ui/client/types/process.ts
```diff
@@ -4,6 +4,7 @@ import {ValidationResult} from "./validation"
 
 export type Process = {
   id: string,
+  newId?: string,
```
mproch, line 7: do we have to put it here? currently this type reflects the data that come from the backend...

ui/client/actions/nk/processUtils/saveProcess.ts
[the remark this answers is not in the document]
lciolecki: Ups ;)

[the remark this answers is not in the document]
JulianWielga: just rename of selector or avoid using process.id?

ui/client/components/modals/SaveProcessDialog.tsx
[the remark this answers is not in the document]
JulianWielga: every iteration in react needs key's for elements. in this function key is the last argument and it's most time omitted in usages. I just added some fallback to almost always set and probably unique values to avoid react warnings.

ui/client/components/graph/node-modal/NodeDetailsContent.js
JulianWielga: wanted some place with "old" value. there is "*" postfix in renamed (not yet saved) value. I think it's nice feature to show this as hint and only old value is enough in my opinion

JulianWielga: every iteration in react needs key's for elements. in this function key is the last argument and it's most time omitted in usages. I just added some fallback to almost always set and probably unique values to avoid react warnings.

ui/client/components/toolbars/status/ProcessInfo.tsx
JulianWielga: wanted some place with "old" value. there is "*" postfix in renamed (not yet saved) value. I think it's nice feature to show this as hint and only old value is enough in my opinion

PR #1524, opened 2021-04-13
prepare process status panel for toolbars changes

JulianWielga:


ui/client/components/toolbars/status/ProcessInfo.tsx
```diff
@@ -69,15 +57,16 @@ class ProcessInfo extends React.Component<OwnProps & StateProps, State> {
       )
     }
 
+    const size = 32
```
lciolecki, line 60: Maybe it can be a prop?

lciolecki (approved):
LGTM

ui/client/components/toolbars/status/ProcessInfo.tsx
JulianWielga: i've removed this from props, never changed, never used

PR #1533, opened 2021-04-19
Fix: Update process with same json

lciolecki:
After moved rename to process properties we can observe that when we change name then system create new version.. In fact we always send rename request and update process request.. If we change 2-3 times name then we can observe that system creates new version of process - but in json we don't see differences.. I debug mode I saw that lastProcessVersion json contains white spaces like `\n` but in BE tests I couldn't reproduce it.. 

This MR fix this situation.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
@@ -55,6 +55,7 @@ trait ProcessRepository[F[_]] {
 class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTypeDataProvider[Int])
   extends ProcessRepository[DB] with EspTables with LazyLogging with CommentActions with ProcessDBQueryRepository[DB] {
 
+  import io.circe.parser._
   import profile.api._
 
   // FIXME: It's temporary way.. After merge and refactor process repositories we can remove it.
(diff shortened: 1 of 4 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -98,12 +98,18 @@ class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTy
       case CustomProcess(mainClass) => (None, Some(mainClass))
     }
 
+    def normalizeJsonString(jsonString: String) = jsonString.filterNot(_.isWhitespace)
```
mproch, line 101: let's compare parsed json instead (i.e. Json class). Otherwise we will still have possible issues - like changing order of fields in json object etc.  I think we can actually pass Json instead of raw string in some places (although I would do large refactor ATM :))

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -98,13 +99,32 @@ class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTy
...
+      case Left(_) => Left(InvalidProcessJson(s"Invalid raw json string: $jsonString."))
+      case Right(json) => Right(json.noSpaces)
     }
 
+    def createProcessVersionEntityData(id: Int, processingType: ProcessingType, json: Option[String], maybeMainClass: Option[String]) = ProcessVersionEntityData(
+      id = id, processId = processId.value, json =json, mainClass = maybeMainClass, createDate = DateUtils.toTimestamp(now),
+      user = loggedUser.username, modelVersion = modelVersion.forType(processingType)
+    )
+
+    //In some reasons in lastVersion.json is string with whitespaces..
```
mproch, line 113: "we compare Json representation to ignore formatting differences"?

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -98,13 +99,32 @@ class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTy
...
+    def isLastVersionContainsSameJson(lastVersion: ProcessVersionEntityData, maybeJson: Option[String]) =
+      lastVersion.json.map(normalizeJsonString) == maybeJson.map(normalizeJsonString)
+
+    def versionToInsert(latestProcessVersion: Option[ProcessVersionEntityData], processesVersionCount: Int, processingType: ProcessingType) =
+      (latestProcessVersion, maybeJson) match {
+        case (Some(version), _) if isLastVersionContainsSameJson(version, maybeJson) && version.mainClass == maybeMainClass =>
+          Right(None)
+        case (_, Some(json)) =>
+          normalizeJsonString(json)
```
mproch, line 122: I think we don't need this, as we normalize in isLastVersionContainsSameJson anyway?

mproch (approved):
:+1: I added two minor comments

PR #1535, opened 2021-04-20
UUID available in GenericConfigCreator

mproch:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/functions/util.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import pl.touk.nussknacker.engine.api.{Documentation, ParamName}
+import pl.touk.nussknacker.engine.util.MathUtils
+
+import java.util.UUID
+
+object util {
+
+  @Documentation(description = "Generate unique identifier (https://en.wikipedia.org/wiki/Universally_unique_identifier)")
+  def uuid: String = UUID.randomUUID().toString
```
lciolecki, line 11: Maybe we can add to methods: `uuid` and `uuidAsString`?  or `stringUuid`?

lciolecki (approved):
LGTM

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/functions/util.scala
mproch: I would rather add rawUuid if it's needed. I wouldn't want to confuse users with UUID details in java :)

PR #1536, opened 2021-04-20
Fix: Create multiple versions for empty process

lciolecki:


lciolecki:
@mproch approved it ;)

mproch (approved):

PR #1537, opened 2021-04-20
Default InfluxDb reporter for standalone-app metrics

arkadius:


engine/standalone/app/src/main/scala/pl/touk/nussknacker/engine/standalone/http/StandaloneHttpApp.scala
```diff
@@ -46,17 +47,19 @@ object StandaloneHttpApp extends Directives with FailFastCirceSupport with LazyL
...
     val metricRegistry = new MetricRegistry
     val metricReporters = loadMetricsReporters()
-    metricReporters.foreach { reporter =>
-      reporter.createAndRunReporter(metricRegistry, config)
+    if (metricReporters.nonEmpty) {
+      metricReporters.foreach { reporter =>
+        reporter.createAndRunReporter(metricRegistry, config)
+      }
+    } else {
+      StandaloneInfluxDbReporter.createAndRunReporterIfConfigured(metricRegistry, config)
```
mproch, line 62: I think we can just have 
```(metricReporters :+ StandaloneInfluxDbReporter).foreach ... ```
and have StandaloneInfluxDbReporter as standalone reporter. If it's not configured - it will not run

kdutkowski (approved):

engine/standalone/app/src/main/scala/pl/touk/nussknacker/engine/standalone/http/StandaloneHttpApp.scala
arkadius: I thought about this options, but then figure out that our influx configuration is not the only influx configuration that someone can want and maybe more secure option would be to let user decide what reporters configurations is the best for he/she. At the end he/she always can reuse `StandaloneInfluxDbReporter` in their code. E.g. consider some additional tags. Current implementation already has some specific tags like `env` and `type` which assumes that someone has multi-environment setup and multi application type.

mproch: oook

PR #1538, opened 2021-04-21
TopicSelectionStrategy on KafkaAvroBaseTransformer

pjagielski:


arkadius (approved):

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/TopicSelectionStrategy.scala
```diff
@@ -0,0 +1,28 @@
+package pl.touk.nussknacker.engine.avro
+
+import cats.data.Validated
+import pl.touk.nussknacker.engine.avro.schemaregistry.{SchemaRegistryClient, SchemaRegistryError}
+
+import java.util.regex.Pattern
+
+trait TopicSelectionStrategy {
+
+  val schemaRegistryClient: SchemaRegistryClient
```
mproch, line 10: I think this is not needed here? Let's either remove it, or put schemaRegistryClient to getTopics parameter (and make Strategy stateless)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/KafkaAvroBaseTransformer.scala
```diff
@@ -25,14 +25,16 @@ trait KafkaAvroBaseTransformer[T] extends SingleInputGenericNodeTransformation[T
 
   def schemaRegistryProvider: SchemaRegistryProvider
 
+  def topicSelectionStrategy(schemaRegistryClient: SchemaRegistryClient): TopicSelectionStrategy = new AllTopicsSelectionStrategy(schemaRegistryClient)
```
mproch, line 28: If we make it def topicSelectionStrategy: TopicSelectionStrategy,
and pass schemaRegistryClient as parameter to strategy method, we'll enable having strategy as field in subclasses

arkadius: woops good point :+1: , I forgotten that this `getTopicParam` is invoken quite often.

pjagielski: moved as param

mproch (approved):

PR #1542, opened 2021-04-21
validate if topic exist or autocreate enabled in broker

gadomsky:


engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+        .describeConfigs(List(new ConfigResource(ConfigResource.Type.BROKER, id)).asJava,
+          new DescribeConfigsOptions().timeoutMs(config.adminClientTimeout.toMillis.toInt))
+        .values()
+        .values()
+        .asScala
+        .map(_.get())
+        .headOption
+        .map(_.get("auto.create.topics.enable").value())
+        .map(_.toBoolean)
+        .get
```
arkadius, line 81: why here is get?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+
+          if (notExistingTopics.isEmpty)
+            Valid(topics)
+          else
+            Invalid(TopicExistenceValidationException.notExistingTopics(notExistingTopics))
+        }
+      })
+
+
+  private def isAutoCreateEnabled: Validated[TopicExistenceValidationException, Boolean] = catchedErrorToInvalid {
```
arkadius, line 66: `TopicExistenceValidationException` is in which case? It looks like in some technical problem right? I wonder if it is a good idea to use validated for technical exceptions. Why not to handle it transparently in a `Future` like other exceptions and use `Validated` only for business validation that topic doesn't exits.

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+  private lazy val kafkaAdminClient = {
+    val properties = new Properties()
+    properties.setProperty("bootstrap.servers", kafkaConfig.kafkaAddress)
+    logger.info(kafkaConfig.kafkaAddress)
+    AdminClient.create(properties)
+  }
+
+  def validateTopics(topics: List[String]): Validated[TopicExistenceValidationException, List[String]] =
+    isAutoCreateEnabled
+      .andThen(if (_) Valid(topics) else {
```
arkadius, line 50: This `if (_)` looks magic. Can you replace it with named function argument?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+        catchedErrorToInvalid {
+          topicListCache.getOrCreate {
```
arkadius, line 52: I wonder if it would be a bad idea to always has scenario:
- try cache
- if not exist in cache, fetch fresh topics
Thanks to that approach we will have 0 latency for scenario when someone just have added topic and fast response for optimistic case when topic already exit.

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaUtils.scala
```diff
@@ -32,6 +31,13 @@ object KafkaUtils extends LazyLogging {
     props.setProperty("client.id", sanitizeClientId(id))
   }
 
+  def validateTopicsExistence(topics: List[PreparedKafkaTopic], kafkaConfig: KafkaConfig): Unit = {
+    new CachedTopicsExistenceValidator(
+      kafkaConfig = kafkaConfig,
+      config = CachedTopicsExistenceValidatorConfig.DefaultConfig)
```
mproch, line 37: Can we decode validator config from kafkaConfig? or some other config? Currently it's quite cumbersome to override config without coding?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  final def validateTopic(topic: String): Validated[TopicExistenceValidationException, String] = validateTopics(List(topic)).map(_.head)
+
+  def validateTopics(topics: List[String]): Validated[TopicExistenceValidationException, List[String]]
+}
+
+trait WithCachedTopicsExistenceValidator extends TopicsExistenceValidator {
+  protected val kafkaConfig: KafkaConfig
+  protected val validatorConfig: CachedTopicsExistenceValidatorConfig = DefaultConfig
+  private lazy val validator = new CachedTopicsExistenceValidator(kafkaConfig, validatorConfig)
```
mproch, line 28: does this has to be private? It makes quite difficult to provide own validator?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      if (notExistingTopics.isEmpty)
+        Valid(topics)
+      else
+        Invalid(new TopicExistenceValidationException(notExistingTopics))
+    }
+  }
+
+  def validationDisabled: Boolean = !kafkaConfig.kafkaEspProperties.flatMap(_.get(TopicExistenceValidationEnabledPropertyName))
+    .map(_.toBoolean).getOrElse(ValidationEnabledDefault)
```
mproch, line 76: maybe it's time to change kafkaEspProperties to Config?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  def validateTopics(topics: List[String]): Validated[TopicExistenceValidationException, List[String]] = {
+    if (validationDisabled || isAutoCreateEnabled) {
+      Valid(topics)
+    } else {
+      val existingTopics = topicListCache.getOrCreate {
+        usingAdminClient {
+          _.listTopics(new ListTopicsOptions().timeoutMs(config.adminClientTimeout.toMillis.toInt))
+            .names().get().asScala.toSet
```
mproch, line 63: why here .toSet and then .toList? :)

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+        Valid(topics)
+      else
+        Invalid(new TopicExistenceValidationException(notExistingTopics))
+    }
+  }
+
+  def validationDisabled: Boolean = !kafkaConfig.kafkaEspProperties.flatMap(_.get(TopicExistenceValidationEnabledPropertyName))
+    .map(_.toBoolean).getOrElse(ValidationEnabledDefault)
+
+  private def isAutoCreateEnabled: Boolean = autoCreateSettingCache.getOrCreate {
```
mproch, line 78: We'd rather have a test for that? this code looks a bit quirky - and I'd suspect it may be a bit more kafka version dependent than most of other codes

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+        .values()
+        .asScala
+        .map(_.get())
+        .head.get(AutoCreateTopicPropertyName)
+        .value()
+        .toBoolean
+    }
+  }
+
+  private def usingAdminClient[T](adminClientOperation: Admin => T): T = {
```
mproch, line 97: I wonder, maybe we can extract some of it to KafkaUtils?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+    }
+    usingAdminClient {
+      _
+        .describeConfigs(List(new ConfigResource(ConfigResource.Type.BROKER, id)).asJava,
+          new DescribeConfigsOptions().timeoutMs(config.adminClientTimeout.toMillis.toInt))
+        .values()
+        .values()
+        .asScala
+        .map(_.get())
+        .head.get(AutoCreateTopicPropertyName)
```
mproch, line 91: Why head here? Maybe let's add at least comment here - that because we request ConfigResource.Type.BROKER config resource, we expect only one?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -0,0 +1,107 @@
...
+        Invalid(new TopicExistenceValidationException(notExistingTopics))
+    }
+  }
+
+  def validationDisabled: Boolean = !kafkaConfig.kafkaEspProperties.flatMap(_.get(TopicExistenceValidationEnabledPropertyName))
+    .map(_.toBoolean).getOrElse(ValidationEnabledDefault)
+
+  private def isAutoCreateEnabled: Boolean = autoCreateSettingCache.getOrCreate {
+    val timeout = config.adminClientTimeout.toMillis.toInt
+    val id = usingAdminClient {
```
mproch, line 80: is this id "randomBrokerFromClusterId"? If so, can we name it as such? :)

mproch (commented):
I added some minor comments

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: fixed

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: fixed

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: fixed

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: I'm not quite sure about this solution, because it will cause that cache will be totally pointless when topic is not created..
I've added it in speparate commit, though.

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaUtils.scala
gadomsky: I've added field for that in KafkaConfig

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: good point, changed to protected

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: moved this to dedicated config

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: I want to have List in API, so I've changed also cache to List

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: I've added integration test for this validator, will it be enough?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: good point

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: comment addded

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
gadomsky: fixed

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
mproch: :+1:

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -41,3 +46,21 @@ object KafkaConfig {
   def parseProcessObjectDependencies(processObjectDependencies: ProcessObjectDependencies): KafkaConfig =
     parseConfig(processObjectDependencies.config, defaultGlobalKafkaConfigPath)
 }
+
+case class TopicsExistenceValidationConfig(enabled: Boolean, validatorConfig: Option[CachedTopicsExistenceValidatorConfig] = None) {
+  def getValidatorConfig = validatorConfig.getOrElse(CachedTopicsExistenceValidatorConfig.DefaultConfig)
```
mproch, line 51: can't we just put DefaultConfig as default argument?

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -41,3 +46,21 @@ object KafkaConfig {
...
+}
+
+object TopicsExistenceValidationConfig {
+  import net.ceedubs.ficus.Ficus._
+  import net.ceedubs.ficus.readers.ArbitraryTypeReader._
+  // scala 2.11 needs it
+  implicit val valueReader: ValueReader[Option[TopicsExistenceValidationConfig]] = new ValueReader[Option[TopicsExistenceValidationConfig]] {
+    override def read(config: Config, path: String): Option[TopicsExistenceValidationConfig] = {
```
mproch, line 59: let's just use OptionReader explicitly?

mproch (approved):
I have two small comments, overall LGTM

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
gadomsky: done

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
gadomsky: I didn't know such class exist
done

engine/kafka/src/main/scala/pl/touk/nussknacker/engine/kafka/validator/TopicsExistenceValidator.scala
```diff
@@ -22,84 +22,78 @@ trait TopicsExistenceValidator {
...
+  }
 
-  private def isAutoCreateEnabled: Validated[TopicExistenceValidationException, Boolean] = catchedErrorToInvalid {
-    autoCreateSettingCache.getOrCreate {
-      val timeout = config.adminClientTimeout.toMillis.toInt
-      val id = kafkaAdminClient
-        .describeCluster(new DescribeClusterOptions().timeoutMs(timeout)).nodes().get().asScala.head.id().toString
-      kafkaAdminClient
-        .describeConfigs(List(new ConfigResource(ConfigResource.Type.BROKER, id)).asJava,
+  private def isAutoCreateEnabled: Boolean = autoCreateSettingCache.getOrCreate {
```
arkadius, line 73: Can you add parameterless `()` so it will be visible that this is not clean function?

arkadius (approved):
LGTM some, minor syntax comments added

PR #1543, opened 2021-04-22
[Feature] ComponentProvider

mproch:
Introduces new way of providing Components (Services, SourceFactories, etc.)
Use cases:
- OpenAPI integration
- DB (SQL and noSQL) integration
- configurable aggregations etc.

### DIFF
--- a/demo/docker/testQuickstart.sh
+++ b/demo/docker/testQuickstart.sh
@@ -57,6 +57,7 @@ elif [[ $CODE == 400 ]]; then
   echo "Process has already exists in db."
 else
   echo "Process creation failed with $CODE"
+  docker logs nussknacker_app
   exit 1
 fi
 
@@ -65,6 +66,7 @@ if [[ $CODE == 200 ]]; then
   echo "Counts queried"
 else
   echo "Counts query failed with $CODE"
+  docker logs nussknacker_app
   exit 1
 fi
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -206,9 +206,39 @@ Configuration of processes has few common keys:
 If configuration does not contain `checkpointConfig`, `checkpointInterval` and process does not contain `checkpointInterval` in its properties then checkpoints are not enabled for process.
 
 The rest of model configuration depends on your needs - all the properties defined here will be passed to ```ProcessConfigCreator``` as explained in [API](API.md) documentation.
+                                                   
+###Configuration of component providers
 
-###Configuration of services
-In model configuration you can also define some attributes of services. These include:
+```
+  componentProviders {
+    sqlHsql {
+      providerType: sql
+      jdbcUrl: jdbc:hsql...//
+      admin/pass
+    }
+    sqlOracle {
+      providerType: sql
+      jdbcUrl: jdbc:oracle...//
+      admin/pass
+    }
+    prinzH20 {
+      providerType: prinzH20
+      h2oLocation:
+    }
+    prinzMlFlow {
+      //this is not needed, as configuration is named just as provider
+      //providerType: prinzMLFlow
+      mlFlowUrl:
+    }
+    //not needed in our domain
+    aggregation {
+      disabled: true
+    }
+  }
+```
+
+###Configuration of UI attributes of components
+In model configuration you can also define some UI attributes of components. These include:
 * default values of fields
 * icons
 
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
@@ -1,0 +1,119 @@
+package pl.touk.nussknacker.engine.api.component
+
+import com.typesafe.config.Config
+import pl.touk.nussknacker.engine.api.process.{ProcessObjectDependencies, SingleNodeConfig}
+
+
+/*
+  Service, SourceFactory, SinkFactory, CustomStreamTransformer
+ */
+trait Component
+
+
+/*
+  componentProviders {
+    sqlHsql {
+      providerType: sql
+      jdbcUrl: jdbc:hsql...//
+      admin/pass
+    }
+    sqlOracle {
+      providerType: sql
+      jdbcUrl: jdbc:oracle...//
+      admin/pass
+    }
+    prinzH20 {
+      providerType: prinzH20
+      h2oLocation:
+    }
+    prinzMlFlow {
+      //this is not needed, as configuration is named just as provider
+      //providerType: prinzMLFlow
+      mlFlowUrl:
+    }
+    openapiService1 {
+      providerType: openapi
+      url: http://host1
+    }
+    openapiService2 {
+      providerType: openapi
+      url: http://host2
+      componentsConfig {
+        //we want to ignore particular service here
+        serviceA {
+          disabled: true
+        }
+      }
+    }
+    //not needed in our domain
+    aggregation {
+      disabled: false
+    }
+    configurableFilter {
+      componentsConfig: {
+        filterWithA {
+          component: customFilter
+          customProperty1: A
+        },
+        filterWithB {
+          component: customFilter
+          customProperty1: B
+        }
+      }
+    }
+  }
+ */
+case class ComponentProviderConfig(providerType: Option[String],
+                                   disabled: Boolean = false,
+                                    //TODO: do we need it?
+                                    elementPrefix: Option[String],
+                                    componentsConfig: Map[String, ComponentConfig],
+                                    categories: List[String],
+                                    config: Config)
+
+case class ComponentConfig(component: Option[String],
+                            disabled: Boolean = false,
+                            config: Config,
+                           //TODO: do we want it here??
+                            uiConfig: Option[SingleNodeConfig])
+
+
+//TODO: figures out how to handle versioning, use e.g. semver4j?
+case class NussknackerVersion(value: String)
+
+object ComponentData {
+
+  def simple(name: String, component: Component): ComponentDefinition = ComponentDefinition(name, enabledByDefault = true, (_, c) => (component, c))
+
+}
+
+//In some cases Nussknacker config admin can be 
+case class ComponentDefinition(name: String, enabledByDefault: Boolean,
+                         //we return Option[SingleNodeConfig], to make it possible to configure node config (e.g defaults) for
+                         create: (Config, Option[SingleNodeConfig]) => (Component, Option[SingleNodeConfig]))
+
+//case class Component
+
+/**
+  * Implementations should be registered with ServiceLoader mechanism. Each provider can be configured multiple times
+  * (e.g. differnent DBs, different OpenAPI registrars and so on.
+  */
+trait ComponentProvider {
+
+  def providerName: String
+
+  //for e.g. OpenAPI we don't want to resolve swagger on Flink (it can be in different network location, or have lower HA guarantees)
+  def resolveConfigForExecution(config: Config): Config
+
+  def create(config: Config, dependencies: ProcessObjectDependencies): List[ComponentDefinition]
+
+  def isCompatible(version: NussknackerVersion): Boolean
+
+}
+
+
+
+
+
+
+
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
@@ -2,12 +2,42 @@ package pl.touk.nussknacker.engine.api.component
 
 import com.typesafe.config.{Config, ConfigFactory}
 import com.vdurmont.semver4j.Semver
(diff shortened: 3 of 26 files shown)


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+    }
+  }
+ */
+case class ComponentProviderConfig(provider: Option[String],
+                                   disabled: Boolean = false,
+                                    //TODO: do we need it?
+                                    elementPrefix: Option[String],
+                                    componentsConfig: Map[String, ComponentConfig],
```
arkadius, line 70: Why it need to be specified on this level? Why provider just not have raw `Config` and on their own decide how to internally handle it? The same notice regards to `elementPrefix` and `categories`. It is because you introduce standarized way to configure that? What about situation when someone will create provider that will have some groups of components and want to have configs for each groups of components?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -0,0 +1,116 @@
...
+        },
+        filterWithB {
+          component: customFilter
+          customProperty1: B
+        }
+      }
+    }
+  }
+ */
+case class ComponentProviderConfig(provider: Option[String],
```
arkadius, line 66: What is `provider` here? Shouldn't it be as a key in map of provider configs? Why it is optional?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
mproch: Yes, the main goal is to introduce some standardized way of handling it. 
We can discuss how far should it go, but I'd like to have some structure. Of course, each componentProvider will have its own config, but I think this should be just a list of fixed properties - I hope in the future ComponentProvider will also have list of "ConfigProperties" which admin will be able to configure via UI. 
It would be easier to design such UI if we had standard way of grouping etc.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
mproch: My idea was to make easy things simple and compex thigns possible:
`
    openapiService1 {
      provider: openapi //we have different configurations of openapi, each is named
      url: http://host1
    }
    singleService {
      //provider: singleService //if have only one configuration of provider, name is taken from name of configuration
    }
`
I think in many cases `singleService` way will be ok

arkadius: ok, let's name it `providerType`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
arkadius: Ok, I know what was my main anxiety reason. I would like to:
- Have one common way to adding postprocessing of definition - we need to think about how this configuration to merge with current configuration of nodes, maybe we should move this configuration to existing configuration or other way?
- Make this configuration separate from core nk application, let's say that we have:
```
ComponentDefinitionPosprocessor {
  def postprocess(componentName): String
  def postprocess(param: Parameter): Parameter
  def postprocess(...) 
}
```
and it will be executed after each step of `GenericNodeContextTransformation`
If you have such an interface it is not important where you have such configuration and how you split it - if it will be configured for each component or it will be defined other way.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
mproch: :+1:

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
mproch: I simplified initial design, component configuration refactor is deferred to next PRs...

docs/Configuration.md
```diff
@@ -206,9 +206,39 @@ Configuration of processes has few common keys:
...
+                                                   
+###Configuration of component providers
 
-###Configuration of services
-In model configuration you can also define some attributes of services. These include:
+```
+  componentProviders {
+    sqlHsql {
+      providerType: sql
+      jdbcUrl: jdbc:hsql...//
```
arkadius, line 216: after taking a look to examples in code, it looks like those properties should be nested in `config`. Maybe we should make it that it won't be nested? WDYT?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+  def apply(classLoader: ClassLoader): ComponentExtractor = {
+    val version = NussknackerVersion(new Semver(BuildInfo.version))
+    ComponentExtractor(classLoader, version)
+  }
+
+}
+
+case class ComponentExtractor(classLoader: ClassLoader, nussknackerVersion: NussknackerVersion) {
+
+  private val providers = ScalaServiceLoader.load[ComponentProvider](classLoader).map(p => p.providerName -> p).toMap
```
arkadius, line 25: WDYT about making it lazy val? IMO it is not intutitive that creation of extractor already load providers

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  private def loadCorrectComponents(config: Config): Map[String, (ComponentProviderConfig, ComponentProvider)] = {
+    import net.ceedubs.ficus.readers.ArbitraryTypeReader._
+    val componentsConfig = config.getAs[Map[String, ComponentProviderConfig]](componentConfigPath).getOrElse(Map.empty)
+    componentsConfig.filterNot(_._2.disabled).map {
+      case (name, providerConfig: ComponentProviderConfig) =>
+        val providerName = providerConfig.providerType.getOrElse(name)
+        val provider = providers.getOrElse(providerName, throw new IllegalArgumentException(s"Provider $providerName not found"))
```
arkadius, line 33: can you add " for component: $name" for wider context of exception?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+        if (!provider.isCompatible(nussknackerVersion)) {
+          throw new IllegalArgumentException(s"Component provider $name (of type $providerName) is not compatible with $nussknackerVersion, please disable it explicitly ")
```
arkadius, line 35: "please use correct component provider version or disable it explicitly"

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  private def extractOneProviderConfig(config: ComponentProviderConfig, provider: ComponentProvider, processObjectDependencies: ProcessObjectDependencies) = {
+    val components = provider.create(config.config, processObjectDependencies).map { cd =>
+      val finalName = config.componentPrefix.map(_ + cd.name).getOrElse(cd.name)
+      finalName -> cd
+    }.toMap
+    components.mapValues(k => WithCategories(k.component, config.categories: _*))
+  }
+
+  private def reduceUnique[T](list: Iterable[Map[String, T]]): Map[String, T] = list.foldLeft(Map.empty[String, T]) {
```
arkadius, line 64: Can you move it to utils? Maybe in pl.touk.nussknacker.engine.util.Implicits ?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+import pl.touk.nussknacker.engine.component.ComponentExtractor.componentConfigPath
+import pl.touk.nussknacker.engine.util.loader.ScalaServiceLoader
+import pl.touk.nussknacker.engine.version.BuildInfo
+
+object ComponentExtractor {
+
+  val componentConfigPath = "components"
+
+  def apply(classLoader: ClassLoader): ComponentExtractor = {
+    val version = NussknackerVersion(new Semver(BuildInfo.version))
```
arkadius, line 17: Can you move this line to `NussknackerVersion()` or mabye we can have `case object NussknackerVersion exteds NussknackerVersion(new Semver(BuildInfo.version))` (don't know if it will work correclty with value classes)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -0,0 +1,47 @@
...
+  //on Flink (it can be in different network location, or have lower HA guarantees), @see ModelConfigLoader
+  def resolveConfigForExecution(config: Config): Config
+
+  def create(config: Config, dependencies: ProcessObjectDependencies): List[ComponentDefinition]
+
+  def isCompatible(version: NussknackerVersion): Boolean
+
+}
+
+case class NussknackerVersion(value: Semver) extends AnyVal
```
arkadius, line 38: There is a problem in usage of value classes from java. Can u verify that and if there is no way to use it from java, replace it with simple case class and add notice in CONTRIBUTING.md ?

arkadius (commented):
Good change, some minor comments added

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -0,0 +1,47 @@
...
+/**
+  * Implementations should be registered with ServiceLoader mechanism. Each provider can be configured multiple times
+  * (e.g. differnent DBs, different OpenAPI registrars and so on.
+  */
+trait ComponentProvider {
+
+  def providerName: String
+
+  //in some cases e.g. external model/service registry we don't want to resolve registry settings
+  //on Flink (it can be in different network location, or have lower HA guarantees), @see ModelConfigLoader
```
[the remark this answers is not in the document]
mproch: :+1:

demo/docker/testQuickstart.sh
```diff
@@ -49,6 +49,9 @@ waitForOK() {
 
 waitForOK "api/processes" "Checking Frontend API response.." "Frontend not started" "app"
 
+echo "Creating fake process"
```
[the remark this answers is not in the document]
mproch: :+1: yeah, I forgot to clean up...

docs/Configuration.md
mproch: it's a bit tricky, but OK - I did it :+1:

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
mproch: :+1:

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
mproch: :+1:

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
mproch: :+1:

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
mproch: :+1:

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
mproch: I added "current"

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
mproch: :+1:  unfortunately, you're right :(

arkadius (approved):
Good change, let's merge it

PR #1546, opened 2021-04-23
[Feature] Unions in standalone

mproch:


engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessInterpreter.scala
```diff
@@ -141,36 +141,83 @@ object StandaloneProcessInterpreter {
...
+            processCompilerData.interpreter.interpret[Future](node, processCompilerData.metaData, one)
+          }).flatMap { list =>
+            val foldedResults = foldResults(list.map(_.swap.leftMap(NonEmptyList.one)))
+            val interpreterd = foldedResults.map { resultList =>
+              resultList.groupBy(_.reference).map {
+                case (pr, ir) => interpretationInvoke(partsInvokers)(pr, ir)
+              }.toList
+            }
+            interpreterd match {
+              case Right(other) => Future.sequence(other).map(foldResults)
```
arkadius, line 158: why not to move `Future.sequence(other).map(foldResults)` into mapping of `foldedResults`? It will be visible then that you here only transforming either[future] into future[either]. Also maybe we should use for-comprehension for futures and/or extract method for all this folding to make it more human readable?

engine/standalone/engine/src/test/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessConfigCreator.scala
```diff
@@ -242,12 +242,13 @@ class StandaloneFilterWithLog(filterExpression: LazyParameter[java.lang.Boolean]
...
-      (ctx: engine.api.Context, ec: ExecutionContext) => {
+      (ctxs: List[engine.api.Context], ec: ExecutionContext) => {
         implicit val ecc: ExecutionContext = ec
-        exprInterpreter(ec, ctx).flatMap(exprResult => {
-          if(exprResult.filterExpression) continuation(ctx, ec)
-          else Future(Right(List(InterpretationResult(DeadEndReference(nodeId.id), exprResult, ctx))))
-        })
+        Future.sequence(ctxs.map(ctx => exprInterpreter(ec, ctx).map((_, ctx)))).flatMap { runInformations =>
+          val (pass, skip) = runInformations.partition(_._1.filterExpression)
+          val skipped = skip.map(info => EndResult(InterpretationResult(DeadEndReference(nodeId.id), info._1, info._2)))
```
arkadius, line 249: can you use:
```
map {
  case (xxx, yyy) => 
}
``` syntax instead of _1 and _2 so it will be more clear what happens here without using IDE?

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessInterpreter.scala
mproch: I refactored a bit, should be clearer a bit now

engine/standalone/engine/src/test/scala/pl/touk/nussknacker/engine/standalone/StandaloneProcessConfigCreator.scala
mproch: :+1:

arkadius (approved):
LGTM

PR #1551, opened 2021-04-27
Rename bugs fixed

JulianWielga:


ui/client/types/process.ts
```diff
@@ -5,7 +5,6 @@ import {EditorProps} from "../components/graph/node-modal/editors/expression/Edi
 
 export type Process = {
   id: string,
-  unsavedNewName?: string,
```
JulianWielga, line 8: @mproch so I throw it out anyway ;)

mproch: :+1: круто!

lciolecki (approved):
Great!

PR #1557, opened 2021-04-30
Standalone modules refactor: engine (used by manager) not using (mode…

arkadius:
…l) utils anymore


engine/standalone/app/src/main/scala/pl/touk/nussknacker/engine/standalone/http/DefaultResponseEncoder.scala
```diff
@@ -1,4 +1,4 @@
-package pl.touk.nussknacker.engine.standalone.utils
+package pl.touk.nussknacker.engine.standalone.http
```
mproch, line 1: I wonder if it's ok to put this class to app. I'd like to be able to use only engine without having to use app

arkadius: It can be on engine side - I've moved it.

mproch (approved):

PR #1562, opened 2021-05-05
Some of the akhq routes were incorrectly resolved as  default (nk) route

ZbyszekMM:
Some of the akhq routes were incorrectly resolved as  default (nk) route, just a little nuisance.

### DIFF
--- a/demo/docker/nginx/nginx.conf
+++ b/demo/docker/nginx/nginx.conf
@@ -14,6 +14,9 @@ http {
       proxy_pass http://akhq:8080/akhq/;
     }
 
+    location /ui/ {
+      proxy_pass http://akhq:8080/akhq/ui/;
+    }
     # this is the internal Docker DNS, cache only for 30s
     # This setting + substitution pattern used below is because we want ot have separated compose for app and environment.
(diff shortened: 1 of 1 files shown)


lciolecki (approved):
Good job!

demo/docker/nginx/nginx.conf
```diff
@@ -14,6 +14,9 @@ http {
       proxy_pass http://akhq:8080/akhq/;
     }
 
+    location /ui/ {
+      proxy_pass http://akhq:8080/akhq/ui/;
+    }
```
gadomsky, line 19: If we had Apache we could handle relative redirects with `ProxyPassReverse`, but nginx does not have it.
But I found on the Web https://www.nginx.com/resources/wiki/start/topics/examples/likeapache/
Would it be any good in this case?

mproch (changes requested):
Let's upgrade to AKHQ 0.17 with fix instead:
https://github.com/tchiotludo/akhq/issues/469

ZbyszekMM:
Performed upgrade of akhq to 0.17 in a separate PR

PR #1568, opened 2021-05-07
Use String in Avro by default

mproch:


engine/flink/avro-util/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
```diff
@@ -171,16 +171,14 @@ class AvroSchemaSpelExpressionSpec extends FunSpec with Matchers {
...
   it("should recognize avro type string as String") {
-    AvroStringSettingsInTests.whenEnabled {
-      val schema = wrapWithRecordSchema(
-        """[
-          |  { "name": "stringField", "type": "string" }
-          |]""".stripMargin)
+    val schema = wrapWithRecordSchema(
+      """[
+        |  { "name": "stringField", "type": "string" }
+        |]""".stripMargin)
```
gadomsky, line 177: maybe we could expand this tests to also check types MAP and ENUM?
or test it somewhere else?

arkadius (approved):

PR #1631, opened 2021-05-13
[Improvement] Load config from multiple files, don't load application.conf in model

mproch:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ConfigFactoryExt.scala
```diff
@@ -4,8 +4,12 @@ import com.typesafe.config.{Config, ConfigFactory}
 
 import java.io.File
 import java.net.URI
+import scala.util.Try
 
 object ConfigFactoryExt {
+
+  private val configLocationProperty = "config.locations"
```
arkadius, line 11: 1. Maybe we should do it public?
2. Maybe it should be "nk.config.locations". To clarify that it is our location

nussknacker-dist/src/universal/bin/run.sh
```diff
@@ -48,11 +48,11 @@ cd $WORKING_DIR
 set -x
 if [[ "${RUN_IN_BACKGROUND}" == "true" ]]; then
   echo "Starting Nussknacker in background"
-  exec java $JDK_JAVA_OPTIONS -Dconfig.override_with_env_vars=true -Dlogback.configurationFile=$LOGBACK_FILE -Dconfig.file=$CONFIG_FILE -cp "$CLASSPATH" pl.touk.nussknacker.ui.NussknackerApp >> $LOG_FILE 2>&1 &
+  exec java $JDK_JAVA_OPTIONS -Dconfig.override_with_env_vars=true -Dlogback.configurationFile=$LOGBACK_FILE -Dconfig.locations=$CONFIG_FILE -cp "$CLASSPATH" pl.touk.nussknacker.ui.NussknackerApp >> $LOG_FILE 2>&1 &
```
arkadius, line 51: Can u also write a notice in `MigrationGuide.md` in case if someone has its own wrapper e.g. in systemd unit?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ConfigFactoryExt.scala
mproch: :+1:

nussknacker-dist/src/universal/bin/run.sh
mproch: :+1:

arkadius (approved):

PR #1643, opened 2021-05-17
[FIX] handling overrides on main config level

mproch:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/ConfigFactoryExt.scala
```diff
@@ -19,17 +19,19 @@ object ConfigFactoryExt {
     }
   }
 
-  def load(locationString: String = System.getProperty(configLocationProperty)): Config = {
+  def load(locationString: String = System.getProperty(configLocationProperty), classLoader: ClassLoader = getClass.getClassLoader): Config = {
```
arkadius, line 22: I don't like this default: `getClass.getClassLoader`. Picking correct classloader is tricky thing and someone can omit it by mistake. Maybe we should have explicitly named method for this option? Also the default shouldn't be context classloader?

mproch: I removed default. In fact we use it only at the start of application, so "any" classloader will do - but you're right, default classloader can be error prone :)

arkadius (approved):

arkadius (approved):

PR #1650, opened 2021-05-18
Fix for: union inside spel's projection and selection wasn't handled …

arkadius:
…by suggester


ui/client/components/graph/node-modal/editors/expression/ExpressionSuggester.js
```diff
@@ -207,12 +207,34 @@ export default class ExpressionSuggester {
...
       //TODO: currently we don't assume nested selections/projections
       const focusedClazz = this._findRootClazz(properties, this._variables)
-      return (focusedClazz || {}).params ? focusedClazz.params[0] : null
+      return this._getFirstParameterType(focusedClazz)
     } else {
       return null
     }
   }
 
+  _getFirstParameterType = (typ) => {
```
mproch, line 216: We have v. nice unit tests for this class, can you add one more? :)

arkadius: Test added

mproch (approved):
I think we're coming to the point where we'll need to invoke BE to get good code completion :)

PR #1651, opened 2021-05-18
Avro source returns additional variable with metadata.

gskrobisz:


gskrobisz:
Changes and concepts:
1) `KafkaAvroSourceFactory`
 - has key and value type parameters
 - [no change] supports Generic and Specific records
 - produces KafkaSource[ConsumerRecord[K, V]]
 - requires event deserialization to ConsumerRecord[K, V]
 - uses `KafkaContextInitializer` to initialize context with #input and #inputMeta
2) `SpecificRecordKafkaAvroSourceFactory`
 - extends `KafkaAvroSourceFactory`
 - returns suggestions for topics like it is in generic version (KafkaAvroSourceFactory)
 - [no change] supports only value-only deserialization
3) Source factory has one point where key or value schema is determined
 - key schema -> BasedOnVersionWithFallbackAvroSchemaDeterminer
 - value schema -> BasedOnVersionAvroSchemaDeterminer
 - that one schema is used in all places where event's raw data are interpreted: deserialization and test data formatting
 - deserializer factory uses determined schema, test data RecordFormatter uses created KafkaDeserializalizationSchema
 - for key schema: when SchemaRegistryClient cannot provide valid schema for given topic and version, runtime schema uses default FallbackSchemaStrategy
 - for key schema: when value-only deserializer is provided, runtime schema uses default FallbackSchemaStrategy
 - there is one point of definition for default FallbackSchemaStrategy (`KafkaAvroBaseTransformer.defaultFallbackStrategyInRuntime`)
 - `RuntimeSchemaData` contains info on fallback strategy
4) Deserialization:
 - `KafkaAvroDeserializationSchemaFactory` indicates if it is value-only deserialization (maybe value-only deserialization should be removed)
 - value-only deserializers use fallback strategy for key deserialization
 - problem: does not handle scenario when writer schema id is used (Option[RuntimeSchemaData] == None)
 - fixed generate-and-parse test data for `ConfluentJsonPayloadDeserializer`
 - removed `TupleAvroKeyValueKafkaAvroDeserializerSchemaFactory`
5) Test data and RecordFormatter
 - uses the same deserialization and schemas (and fallback strategies) as in event processing
 - uses SerializableConsumerRecord to read/write json content
 - problem: two different readers are used in ConfluentAvroMessageReader
6) Tests:
 - new tests for key deserialization (Generic and Specific, AvroPayload and JsonPayload, value-only and key-value deserialization)
 - new tests for JsonPayload
 - problem: failing tests for scenario when writer schema id is used

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
```diff
@@ -23,24 +29,59 @@ private[confluent] class ConfluentAvroMessageFormatter(schemaRegistryClient: Sch
 
   schemaRegistry = schemaRegistryClient
 
-  def writeTo(data: Array[Byte], output: PrintStream): Unit = {
-    val obj = deserialize(data)
-    val schema = AvroSchemaUtils.getSchema(obj)
+  def genericToJson(obj: AnyRef, schema: Schema): Json = {
+    toJson(obj, schema, new GenericDatumWriter[AnyRef](schema))
```
arkadius, line 33: I know that it was before this way, but imo we should use `AvroAvroUtils.xxxData.genericDatumWriter` instead of using constructor directly. We have some fixes in our `AvroAvroUtils` regards to logical type handling, `GenericRecordWithSchemaId`, `StringForcing` and so on and would be nice to keep it consistent

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatter.scala
```diff
@@ -24,64 +29,62 @@ private[confluent] class ConfluentAvroToJsonFormatter(schemaRegistryClientFactor
...
+  private def determineSchemaIdAndContent[T: ClassTag](obj: T, rawData: Array[Byte], schemaDataOpt: Option[RuntimeSchemaData]): (Option[Int], Option[Json]) = {
+    schemaDataOpt match {
+      case Some(RuntimeSchemaData(_, _, Some(runtimeFallbackStrategy))) =>
+        // fallback
+        val jsonOpt = Some(runtimeFallbackStrategy.toJson(obj.asInstanceOf[runtimeFallbackStrategy.T]))
+        (None, jsonOpt)
+      case _ if AvroUtils.isSpecificRecord[T] =>
+        // SpecificRecord
+        val schemaIdOpt = schemaDataOpt.flatMap(_.schemaIdOpt)
+        val jsonOpt = schemaDataOpt.map(runtimeSchema => formatter.specificToJson(obj.asInstanceOf[AnyRef], runtimeSchema.schema))
```
arkadius, line 55: instead of dispatching between specific record/generic record case here, it would be nice to have `AvroUtils.createDatumWriter` that will do this dispatching on their own.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
```diff
@@ -23,24 +29,59 @@ private[confluent] class ConfluentAvroMessageFormatter(schemaRegistryClient: Sch
...
   }
+
+  def serializeRecord[K, V](record: ConsumerRecord[K, V],
+                            keySchemaIdOpt: Option[Int],
+                            keyJsonOpt: Option[Json],
+                            valueSchemaIdOpt: Option[Int],
+                            valueJsonOpt: Option[Json]
+                           ): Array[Byte] = {
+
+    val serializableRecord = SerializableConsumerRecord(
```
arkadius, line 67: Can you create factory method `SerializableConsumerRecord.apply(consumerRecord, keySchemaIdOpt, valueSchemaIdOpt)` to have this option wrapping logic in one place?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
```diff
@@ -1,15 +1,21 @@
...
+import io.confluent.kafka.serializers.AbstractKafkaAvroDeserializer
+import org.apache.avro.{AvroRuntimeException, Schema}
 import org.apache.avro.generic.GenericDatumWriter
-import org.apache.avro.io.EncoderFactory
+import org.apache.avro.io.{DatumWriter, EncoderFactory}
+import org.apache.avro.specific.SpecificDatumWriter
+import org.apache.kafka.clients.consumer.ConsumerRecord
 import org.apache.kafka.common.errors.SerializationException
+import pl.touk.nussknacker.engine.kafka.ConsumerRecordUtils
+import pl.touk.nussknacker.engine.kafka.consumerrecord.SerializableConsumerRecord
```
arkadius, line 18: I wonder if after changes with `RecordFormatterFactory` it would be possible in avro case to use just `ConsumerRecordToJsonFormatter` and only has some avro specific logic in the factory?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -54,15 +54,27 @@ private[confluent] class ConfluentAvroMessageReader(schemaRegistryClient: Schema
...
           val valueString = if (keyIndex + 1 > str.length) "" else str.substring(keyIndex + 1)
-          val serializedKey = if (useStringForKey) {
-            keyString.getBytes(StandardCharsets.UTF_8)
-          } else {
-            val key = jsonToAvro(keyString, keySchema)
-            serializeImpl(keySubject, key, new AvroSchema(keySchema))
+          if (keySchema != null && keyString.length == 0) {
+            // key schema is provided but there is no content
+            throw new SerializationException("No key content found in line " + str)
+          } else if (keySchema == null) {
```
arkadius, line 60: Can you change `null` to  `Option`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
```diff
@@ -37,15 +37,15 @@ object ConfluentSchemaRegistryProvider extends Serializable {
   def apply(schemaRegistryClientFactory: ConfluentSchemaRegistryClientFactory): ConfluentSchemaRegistryProvider =
     avroPayload(
       schemaRegistryClientFactory,
-      formatKey = false
+      formatKey = true
```
arkadius, line 40: Why not to just remove formatKey flag? Or you want to make it in next PR with better test generator?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/serialization/KafkaAvroDeserializationSchemaFactory.scala
```diff
@@ -103,26 +65,38 @@ abstract class KafkaAvroKeyValueDeserializationSchemaFactory
...
+          record.leaderEpoch()
+        )
       }
 
-      override def isEndOfStream(nextElement: O): Boolean = false
+      override def isEndOfStream(nextElement: ConsumerRecord[K, V]): Boolean = false
+
+      override def getProducedType: TypeInformation[ConsumerRecord[K, V]] = {
+        if (kafkaConfig.useStringForKey) {
+          val clazz = classTag[ConsumerRecord[String, V]].runtimeClass.asInstanceOf[Class[ConsumerRecord[K, V]]]
```
arkadius, line 91: 1. Can't you just do `classOf[ConsumerRecord[String, V]]`? without classTag? Also IMO there is no need to have this if-else here because this class object is always the same because generic type erassure.
2. Can you left TODO that it would be nice to build `TypeInformation` for `ConsumerRecord` using `createKeyTypeInfo` and `createValueTypeInfo`?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/KafkaAvroSourceFactory.scala
```diff
@@ -1,35 +1,38 @@
...
-                                         timestampAssigner: Option[TimestampWatermarkHandler[T]])
-  extends BaseKafkaAvroSourceFactory[T](timestampAssigner) with KafkaAvroBaseTransformer[FlinkSource[T]]{
+class KafkaAvroSourceFactory[K:ClassTag, V:ClassTag](val schemaRegistryProvider: SchemaRegistryProvider,
+                                                     val processObjectDependencies: ProcessObjectDependencies,
+                                                     timestampAssigner: Option[TimestampWatermarkHandler[ConsumerRecord[K, V]]])
+  extends BaseKafkaAvroSourceFactory[K, V](timestampAssigner) with KafkaAvroBaseTransformer[FlinkSource[ConsumerRecord[K, V]]]{
 
   override type State = KafkaAvroSourceFactoryState
 
+  var kafkaContextInitializer: KafkaContextInitializer[K, V, DefinedParameter, State] = _
```
arkadius, line 31: Nooo pleeease don't use vars ;-). You have `State` for purpose of passing created objects during context transformation...

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/SpecificRecordKafkaAvroSourceFactory.scala
```diff
@@ -1,46 +1,74 @@
...
+  override def initialParameters: List[Parameter] = Nil
 
-    val valueSchemaDeterminer = new SpecificRecordEmbeddedSchemaDeterminer(classTag[T].runtimeClass.asInstanceOf[Class[_ <: SpecificRecord]])
-    val valueSchemaData = valueSchemaDeterminer.determineSchemaUsedInTyping.valueOr(SchemaDeterminerErrorHandler.handleSchemaRegistryErrorAndThrowException)
-    val valueSchemaUsedInRuntime = valueSchemaDeterminer.toRuntimeSchema(valueSchemaData)
+  override def contextTransformation(context: ValidationContext, dependencies: List[NodeDependencyValue])(implicit nodeId: NodeId): NodeTransformationDefinition =
+    topicParamStep orElse {
+      case step@TransformationStep((TopicParamName, DefinedEagerParameter(topic:String, _)) :: Nil, _) =>
+        val preparedTopic = prepareTopic(topic)
+        val versionOption = LatestSchemaVersion
```
arkadius, line 37: Can you replace overloading of `prepareValueSchemaDeterminer(topics, version)` with fixed `LatestSchemaVersion` with inline creation of determiner as it was before? `LatestSchemaVersion` is bad because we want to use schema that is inside generated classes and it can be other than latest.

arkadius, line 37: Can you extract method in `KafkaAvroBaseTransformer` for this duplicated code? Let it take `valueSchemaDeterminer` and other things...

arkadius: also after changes with passing of schemadata in state I see that `SpecificRecordEmbeddedSchemaDeterminer` is not needed anymore - you just pass schemadata

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
gskrobisz: Good point, it is moved to separate PR.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatter.scala
gskrobisz: Good point, it is moved to separate PR.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
gskrobisz: That exact change I do in separate PR. Here goes the minimal change that introduces #inputMeta & ConsumerRecord to avro sources.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
gskrobisz: This is temporary change to keep coherence with unchanged `readSchemaId`. It is removed in next PR that improves test data formatter.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
gskrobisz: That changes to Formatter are moved to separate PR.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
gskrobisz: I like the idea, will try that with other changes to Formatter that are moved to separate PR.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
gskrobisz: Changed null to Option.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentSchemaRegistryProvider.scala
gskrobisz: Removed `formatKey` params.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/serialization/KafkaAvroDeserializationSchemaFactory.scala
gskrobisz: Fixed and added some TODO explanation.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/KafkaAvroSourceFactory.scala
gskrobisz: Fixed. Now the initializer is passed through the `State`.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/SpecificRecordKafkaAvroSourceFactory.scala
gskrobisz: Brought back the inline creation. Added some comment with explanation.

arkadius (approved):

PR #1655, opened 2021-05-18
Fix for regression: dict labels wasn't completed correctly

arkadius:


JulianWielga (approved):
it would be nice to have tests but ...

PR #1669, opened 2021-05-20
Improvements: add EmptyInvocationCollector for tests

lciolecki:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/test/EmptyInvocationCollector.scala
```diff
@@ -0,0 +1,14 @@
+package pl.touk.nussknacker.engine.api.test
+
+import pl.touk.nussknacker.engine.api.test.InvocationCollectors.{ServiceInvocationCollector, ToCollect}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+//This implementation is used only for tests
```
mproch, line 7: This implementation **should be** used only for unit tests of services

mproch (approved):
:+1: I have one minor doc comment

Issue #1154, opened 2020-08-17
Undo button doesn't work


lciolecki:
Undo button at process edit window doesn't work. 

Url: https://staging.nussknacker.io/visualization/avroLCL?businessView=false

FE Exception: 

`vendors.js?f3e1ab588ef371e85c35:107 Uncaught TypeError: Cannot read property 'map' of undefined
    at D (main.js?f3e1ab588ef371e85c35:1)
    at main.js?f3e1ab588ef371e85c35:1
    at vendors.js?f3e1ab588ef371e85c35:91
    at Array.reduce (<anonymous>)
    at vendors.js?f3e1ab588ef371e85c35:91
    at gt (vendors.js?f3e1ab588ef371e85c35:9)
    at Function.En.reduce (vendors.js?f3e1ab588ef371e85c35:9)
    at e (main.js?f3e1ab588ef371e85c35:1)
    at e (main.js?f3e1ab588ef371e85c35:1)
    at vendors.js?f3e1ab588ef371e85c35:14`

PR #1681, opened 2021-05-24
Bugfix/node keyboard delete

JulianWielga:
1. also fixed (most annoying) some bugs with undo/redo and better business view switch
2. Closed https://github.com/TouK/nussknacker/issues/1154


ui/client/components/toolbars/view/buttons/BussinesViewSwitch.tsx
```diff
@@ -12,7 +11,7 @@ import {ReactComponent as Icon} from "../../../../assets/img/toolbarButtons/busi
 type Props = StateProps
 
 function BussinesViewSwitch(props: Props) {
-  const {businessView, businessViewChanged, fetchProcessToDisplay, nothingToSave, processId, versionId} = props
+  const {businessView, fetchProcessToDisplay, nothingToSave, processId, versionId} = props
   const {t} = useTranslation()
   return (
     <ToolbarButton
```
lciolecki, line 17: Right new we don't need info about business view?

lciolecki (approved):

ui/client/components/toolbars/view/buttons/BussinesViewSwitch.tsx
JulianWielga: we have this info inside process, no need to add it separately and then sync.

PR #1700, opened 2021-05-27
Fixed SPI for AuthenticationProvider

trombka:
With my previous PR to the security module, I added implicit arguments to the constructor of `OAuth2AuthenticatorFactory`. This makes it impossible to load its subclasses with SPI.
I moved the `ec` argument to the `createAuthenticator` and pulled down STTP backend creation into the authenticator factory class.

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatorFactory.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatorFactory.scala
@@ -3,6 +3,9 @@ package pl.touk.nussknacker.ui.security.api
 import akka.http.scaladsl.server.directives.AuthenticationDirective
 import akka.http.scaladsl.server.Route
 import com.typesafe.config.Config
+import sttp.client.{NothingT, SttpBackend}
+
+import scala.concurrent.{ExecutionContext, Future}
 
 trait AuthenticatorFactory {
   import AuthenticatorFactory._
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -914,7 +914,8 @@ lazy val httpUtils = (project in engine("httpUtils")).
         "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV,
         "com.softwaremill.sttp.client" %% "core" % sttpV,
         "com.softwaremill.sttp.client" %% "async-http-client-backend-future" % sttpV,
-        "com.softwaremill.sttp.client" %% "circe" % sttpV
+        "com.softwaremill.sttp.client" %% "circe" % sttpV,
+        "com.softwaremill.sttp.client" %% "akka-http-backend" % sttpV,
```
arkadius, line 918: Why this dependency is in `http-util` instead of `security`? `http-util` is used by model extensions and we don't want to have akka in model ...

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatorFactory.scala
```diff
@@ -3,15 +3,18 @@ package pl.touk.nussknacker.ui.security.api
 import akka.http.scaladsl.server.directives.AuthenticationDirective
 import akka.http.scaladsl.server.Route
 import com.typesafe.config.Config
+import sttp.client.{NothingT, SttpBackend}
 
-trait AuthenticatorFactory {
+import scala.concurrent.{ExecutionContext, Future}
+
+trait AuthenticatorFactory[-T] {
```
mproch, line 10: I think it's too generic. You can implement AuthenticatorFactory[String], it will be loaded with ServiceLoader and class cast exception will be thrown. I think it will be easier to just use AuthenticatorContext here... If sb doesn't want to use sttpBackend, they just won't use it...

mproch (approved):

arkadius (approved):

PR #1723, opened 2021-05-31
Grafana dashboard corrections

ZbyszekMM:
Changed treatment of null metric values on the graph. Instead of extrapolating values  with a straight  line ("connect"), simply do not show anything in the time period when there were no metrics reported. The result is hopefully on the attached pic.
image

Additionally modified  the "max lag" graph , we were showing stacked sum of max lags as reported by each Flink parallelized task. This is roughly the same value shown and stacked as many times as the Flink task parallelism level. Decided to show only one value, commented the graph accordingly. 

image.

An interesting new effect is visible - every 4 minutes reporting (hopefully only reporting) stops for ca 20s, I get this effect on both my local machine and client (N..a) deployment.

### DIFF
--- a/demo/docker/grafana/dashboards/Flink-ESP.json
+++ b/demo/docker/grafana/dashboards/Flink-ESP.json
@@ -1429,7 +1429,7 @@
       "dashLength": 10,
       "dashes": false,
       "datasource": "influx",
-      "description": "IMPORTANT. \nThis metrics does NOT show the overall lag on Kafka source; a largest lag among all the source partitions is shown. ",
+      "description": "IMPORTANT. \nThis metric does NOT show the overall lag on Kafka source; a largest lag among all the source partitions is shown.",
       "editable": true,
       "error": false,
       "fill": 1,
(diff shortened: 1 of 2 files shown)


arkadius (approved):
Good job, let's merge it!

demo/docker/grafana/dashboards/Flink-ESP.json
```diff
@@ -1429,6 +1429,7 @@
       "dashLength": 10,
       "dashes": false,
       "datasource": "influx",
+      "description": "IMPORTANT. \nThis metrics does NOT show the overall lag on Kafka source; a largest lag among all the source partitions is shown. ",
```
piotrp, line 1432: "This metrics" -> "This metric", and while we're changing it we may remove unnecessary space at end :D

piotrp (changes requested):

ZbyszekMM:
Applied piotrp review comments.
Additional changes: the core of this PR is not to invent values when there are no metrics reported. In majority of cases the graphs look much more credible. There is one side effect though - from time to time, time difference between consecutive metrics is slightly above 10s resulting in "null value" and disconnected line on the graph. This is disturbing to the viewer, because one might think something wrong is going here. This effect becomes less and less visible when the observation "interval"  is higher, becoming not visible at all with interval set to 3 minutes. 
So, seeing some value in the observation interval of 10s, I set the default one to 30s and added 3 minute one. 
Finally, tuned up telegraf config to let it deal with a bit higher than very low number of metrics (_Metric buffer overflow; XYZ metrics have been dropped_ .

piotrp (approved):

PR #1728, opened 2021-06-01
Move useStringForKey to KafkaConfig.

gskrobisz:


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClientFactory.scala
```diff
@@ -15,7 +15,7 @@ class CachedConfluentSchemaRegistryClientFactory(maximumSize: Long, latestSchema
   //Cache engines are shared by many of CachedConfluentSchemaRegistryClient
   @transient private lazy val caches = new SchemaRegistryCaches(maximumSize, latestSchemaExpirationTime, schemaExpirationTime, versionsCacheExpirationTime)
 
-  override def createSchemaRegistryClient(kafkaConfig: KafkaConfig): ConfluentSchemaRegistryClient = {
+  override def create(kafkaConfig: KafkaConfig): ConfluentSchemaRegistryClient = {
```
arkadius, line 18: What is the purpose of renaming this method? I would prefer to keep it named old way instead of resolve problems after code migration.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -24,7 +26,8 @@ import pl.touk.nussknacker.engine.avro.schemaregistry.confluent.ConfluentUtils
   * @param parseKey if key should be parsed
   * @param keySeparator key separator
   */
-private[confluent] class ConfluentAvroMessageReader(schemaRegistryClient: SchemaRegistryClient,
+private[confluent] class ConfluentAvroMessageReader(kafkaConfig: KafkaConfig,
```
arkadius, line 29: WDYT about passing  `useStringForKey` instead of full config here? It would be consistent with current approach for `parseKey` and `keySeparator`

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClientFactory.scala
arkadius: Ok, I see now that it is for purpose of being consistent with `DeserializationSchemaFactory` and so on

arkadius (approved):
I've left one minor comment in ConfluentAvroMessageReader. Good job

PR #1741, opened 2021-06-07
[Feature] Errors can be logged to Kafka

mproch:


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaExceptionSerializationSchema.scala
```diff
@@ -0,0 +1,69 @@
...
+import pl.touk.nussknacker.engine.api.MetaData
+import pl.touk.nussknacker.engine.api.exception.{EspExceptionInfo, NonTransientException}
+
+import java.io.{PrintWriter, StringWriter}
+import java.net.InetAddress
+import java.nio.charset.StandardCharsets
+import java.time.ZoneId
+import scala.io.Source
+
+case class KafkaExceptionSerializationSchema(metaData: MetaData, consumerConfig: KafkaExceptionConsumerConfig) {
```
lciolecki, line 15: I wonder why assume errors will be always log as json? What about log errors as avro, etc?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/sharedproducer/KafkaSharedProducer.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def sendToKafka(producerRecord: ProducerRecord[Array[Byte], Array[Byte]])(implicit ec: ExecutionContext): Future[Unit] = {
+    KafkaUtils.sendToKafka(producerRecord)(producer).map(_ => ())
+  }
+
+  override protected def sharedServiceHolder: SharedServiceHolder[KafkaProducerCreator.Binary, _] = SharedKafkaProducerHolder
+
+  // See FlinkKafkaProducer.close. Producer is closed again in case of an exception.
+  override def internalClose(): Unit = {
+    try {
```
lciolecki, line 40: There we can try do some recurention function to try/catch?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaExceptionSerializationSchema.scala
mproch: I added comment to add further formats in the future

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/sharedproducer/KafkaSharedProducer.scala
mproch: For now I removed this code

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
```diff
@@ -0,0 +1,69 @@
...
+  private lazy val hostName = InetAddress.getLocalHost.getHostName
+
+  def apply(metaData: MetaData, exceptionInfo: EspExceptionInfo[NonTransientException], config: KafkaExceptionConsumerConfig): KafkaExceptionInfo = {
+    new KafkaExceptionInfo(
+      metaData.id,
+      exceptionInfo.nodeId,
+      Option(exceptionInfo.throwable.message),
+      Option(exceptionInfo.throwable.input),
+      optional(exceptionInfo.context, config.includeInputEvent).map(_.toString),
+      serializeStackTrace(config.stackTraceLengthLimit, exceptionInfo.throwable),
```
arkadius, line 49: I see one problem here. We have two possible cases of `NonTransientException` - one is class which extends  `NonTransientException` and the other one case (used more often), when we wrap some `Exception` with `NonTransientException`. In the second case IMO will be seen in stacktrace only `Unknown exception`, because default stack trace limit is set to low limit (10). I wonder if we should have some extraction logic that will extract "interesting" exception. This handler could skip `NonTransientException` that is only the wrapper and some other common "wrapper" exceptions like `java.util.concurrent.{CompletionException, ExecutionException}`

mproch: We already skip some common stuff in `DeeplyCheckingExceptionExtractor` - but only to get to NonTransientException. I think that detecting "real" cause can be not so simple. 
I think we can raise default limit (like, to 50?) and do some work to throw NonTransientException when we expect they can happen (e.g. with spel evaluation) instead of catching them in DefaultExceptionHandler
WDYT?

arkadius: ahh, indeed! yes, please raise default limit

arkadius (approved):

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
mproch: done

PR #1760, opened 2021-06-11
Some default setup improvements

mproch:
Mainly RocksDB setup: exporting metrics, default values. 
Also added RocksDB panels to dashboard (as separate row, can be hidden by default)
Screenshot from 2021-06-22 09-54-37

### DIFF
--- a/demo/docker/flink/flink-conf.yaml
+++ b/demo/docker/flink/flink-conf.yaml
@@ -29,8 +29,8 @@ taskmanager.hostname: taskmanager
 
 taskmanager.data.port: 6126
 
-taskmanager.memory.task.heap.size: 1256m
-taskmanager.memory.managed.size: 1204m
+taskmanager.memory.task.heap.size: 256m
+taskmanager.memory.managed.size: 204m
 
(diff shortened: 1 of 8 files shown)


demo/docker/flink/flink-conf.yaml
```diff
@@ -73,6 +73,11 @@ state.backend: filesystem
 state.checkpoints.dir: file:///opt/flink/data/checkpoints
 state.savepoints.dir: file:///opt/flink/data/savepoints
 
+state.backend.rocksdb.metrics.estimate-num-keys: true
+state.backend.rocksdb.metrics.estimate-live-data-size: true
+#We can have many jobs per cluster, in such setting managed memory is not easy to tune
+state.backend.rocksdb.memory.managed: false
```
arkadius, line 79: Please add e.g. `state.backend.rocksdb.writebuffer.size: 1000MB` + comment.

mproch: I increased to 256m and added comment, I think in docker-compose setup it should be enough...

arkadius (approved):

PR #1762, opened 2021-06-14
Refactor to use Using.resource

avan1235:
Introduce Using.resource usages as an alternative to Java Try-With-Resources.
Additionally, refactor all found possible usages of this method in project.


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/exception/WithResources.scala
```diff
@@ -0,0 +1,30 @@
+package pl.touk.nussknacker.engine.util.exception
+
+object WithResources {
+
+  // based on Oracle try-with-resources explanation from
+  // https://www.oracle.com/technical-resources/articles/java/trywithresources.html
+  def use[Resource <: AutoCloseable, Result](resource: Resource)
+                                            (consumer: Resource => Result): Result = {
+
+    var maybeThrown: Throwable = null
```
lciolecki, line 10: Please try do this `try in try`.. as recursion function, after that you should be able to remove `var maybeThrown: Throwable = null`.

piotrp:
Looks nice, but can you change it to use backported `scala.util.Using` (eg. from https://github.com/bigwheel/util-backports/)? That way we can reduce the amount of code we need in our utils.

avan1235:
Ok, refactored to use external library as it seems to be small enough and not provide unused dependencies. Maybe it will have better compatibility with Scala 2.11 than was my version as I saw some problems in compilation

avan1235:
What merge strategy should I define for the conflicting packages in your opinion?

piotrp:
None 🤦 

`util-backports` isn't needed, `scala.util.Using` is already provided by `scala-collection-compat` so if it's missing in some module we should use that implementation.

avan1235:
Yeah, you are right, I didn't see the Using class in logs but it is provided. Already fixed

piotrp (approved):

piotrp:
👍, will merge when tests pass. During review I found an unclosed `Source.fromFile` in `OAuth2Configuration` that could benefit from your changes - you can add it to this PR if you want, if not i will fix it later.

PR #1765, opened 2021-06-14
Improvements: Typer better way to extract method reference

lciolecki:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -155,12 +155,10 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
           }
         }
 
-      case e: MethodReference => typeChildren(validationContext, node, current.popStack) { typedParams =>
-        TypeMethodReference(e.getName, current.stack, typedParams) match {
-          case Right(typingResult) => Valid(typingResult)
-          case Left(errorMsg) => if(strictMethodsChecking) invalid(errorMsg) else Valid(Unknown)
+      case e: MethodReference =>
+        current.stackHead.map(_ => extractMethodReference(e, validationContext, node, current.popStack)).getOrElse {
```
mproch, line 159: you're popping stack twice... Let's maybe use .stackHead explicitly?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -155,12 +155,10 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
+          invalid(s"Invalid method reference: $e.")
```
mproch, line 160: Does MethodReference have nice .toString? Maybe let's use: .toStringAST

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
lciolecki: Yep.. I've changed it.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
lciolecki: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -155,12 +155,12 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
           }
         }
 
-      case e: MethodReference => typeChildren(validationContext, node, current.popStack) { typedParams =>
-        TypeMethodReference(e.getName, current.stack, typedParams) match {
-          case Right(typingResult) => Valid(typingResult)
-          case Left(errorMsg) => if(strictMethodsChecking) invalid(errorMsg) else Valid(Unknown)
+      case e: MethodReference =>
+        if (current.isEmptyStack) {
```
mproch, line 159: maybe let's remove popStack and stackTail altogether? And force to use stackHead with proper pattern matching? 
Now in `extractMethodReference` you assume it will be invoked after checking the stack is not empty - which is a bit error prone?

lciolecki: Ugh.. As awlays, you're right :)

mproch (approved):

PR #1769, opened 2021-06-14
Fix variable suggestions for KafkaSourceFactory with explicit key-val…

gskrobisz:
…ue types.

### DIFF
--- a/engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
+++ b/engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
@@ -12,7 +12,7 @@ import pl.touk.nussknacker.engine.api.context.{ProcessCompilationError, Validati
 import pl.touk.nussknacker.engine.api.context.transformation.{BaseDefinedParameter, DefinedEagerParameter, DefinedSingleParameter, NodeDependencyValue, SingleInputGenericNodeTransformation}
 import pl.touk.nussknacker.engine.api.definition._
 import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypedClass}
-import pl.touk.nussknacker.engine.definition.WithExplicitPossibleProducedVariableTypes
+import pl.touk.nussknacker.engine.definition.WithExplicitTypesToExtract
 import pl.touk.nussknacker.engine.kafka.source.KafkaSourceFactory.KafkaSourceFactoryState
 import pl.touk.nussknacker.engine.kafka.validator.WithCachedTopicsExistenceValidator
 
@@ -45,7 +45,7 @@ class KafkaSourceFactory[K: ClassTag, V: ClassTag](deserializationSchemaFactory:
   extends FlinkSourceFactory[ConsumerRecord[K, V]]
     with SingleInputGenericNodeTransformation[FlinkSource[ConsumerRecord[K, V]]]
     with WithCachedTopicsExistenceValidator
-    with WithExplicitPossibleProducedVariableTypes {
+    with WithExplicitTypesToExtract {
 
   protected val topicNameSeparator = ","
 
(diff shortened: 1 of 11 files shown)


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
```diff
@@ -42,10 +42,26 @@ class KafkaSourceFactory[K: ClassTag, V: ClassTag](deserializationSchemaFactory:
...
                                                    processObjectDependencies: ProcessObjectDependencies)
-  extends FlinkSourceFactory[ConsumerRecord[K, V]] with SingleInputGenericNodeTransformation[FlinkSource[ConsumerRecord[K, V]]] with WithCachedTopicsExistenceValidator {
+  extends FlinkSourceFactory[ConsumerRecord[K, V]]
+    with SingleInputGenericNodeTransformation[FlinkSource[ConsumerRecord[K, V]]]
+    with WithCachedTopicsExistenceValidator
+    with WithExplicitPossibleProducedVariableTypes {
 
   protected val topicNameSeparator = ","
 
+  protected val keyTypingResult: TypedClass = Typed.typedClass[K]
```
lciolecki, line 52: Why you assume there always be typed class?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/DefinitionExtractor.scala
```diff
@@ -32,7 +32,10 @@ class DefinitionExtractor[T](methodDefinitionExtractor: MethodDefinitionExtracto
     (obj match {
       //TODO: how validators/editors in NodeConfig should be handled for GenericNodeTransformation/WithExplicitMethodToInvoke?
       case e:GenericNodeTransformation[_] =>
-        val returnType = if (e.nodeDependencies.contains(OutputVariableNameDependency)) Unknown else Typed[Void]
+        val returnType = e match {
```
mproch, line 35: we shouldn't do it here. This should be possible return type of this component - current usage with passing union of key and value types does not make sense here. Let's rather use it in TypesInformation.extract method

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/WithExplicitPossibleProducedVariableTypes.scala
```diff
@@ -0,0 +1,7 @@
+package pl.touk.nussknacker.engine.definition
+
+import pl.touk.nussknacker.engine.api.typed.typing.TypedClass
+
+trait WithExplicitPossibleProducedVariableTypes {
+  def possibleVariableClasses: Set[TypedClass]
```
mproch, line 6: let's rename it somehow. These are not necessarily exact variable classes. Maybe sth along the lines of: 
`trait DeclaringExplicitTypesToExtract {
   def typesToExtract: Set[TypedClass]
}
`?

mproch (changes requested):

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/KafkaSourceFactory.scala
gskrobisz: Here K and V are usually a specific class or some generic Map/TypedMap. For specific classes this is the same as previous assumption with Typed[K] and Typed[V]. For generic Map/TypedMap the returning type is overwritten by TypedObjectTypingResult which has explicit fields definition.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/DefinitionExtractor.scala
arkadius: By component you mean "Component" class? IMO we shouldn't do it too much generic. Take a look that this `possibleVariableClasses` takes no arguments. If we would like to do it correctly, we should probable traverse the whole scenario (not definition) and take a look what classes are returned. Also "codefull" approach of declaring components is deprecated. Isn't it?

gskrobisz: I moved it to `TypesInformation.extract`.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/WithExplicitPossibleProducedVariableTypes.scala
gskrobisz: Renamed.

arkadius (approved):
please move the trait to api module

mproch (approved):

PR #1772, opened 2021-06-15
Fix for Spel validation: Use not existing method reference

lciolecki:
TODO:
- Add tests
- Rethrow exception with content of expression for easier further diagnostics
- Add entry in changelog


arkadius:
I've changed the title on more descriptive. Is it correct now? Or maybe it is more general problem when passed arguments lists is wider than expected?

mproch (approved):

arkadius (approved):

arkadius (approved):

PR #1774, opened 2021-06-15
focus fix

fi-:
focus-fix1
focus-fix2


lciolecki:
We will do it in another way - more ts.

PR #1780, opened 2021-06-15
More flexible authenticator customization by using named SPI services.

trombka:
I have found out that the extension engine for authenticators is hardly useful.
You can override an authenticator, but you cannot add a new method name since they are defined in an enumeration. So in the case of a totally new authenticator, it is not possible to pass to FE a method name other than `Other`.

The PR drops the enumeration in favor of the NamedServiceProvider interface.
I have also renamed `DefaultAuthenticationConfiguration` to `Basic...` since it seems the only usage case and the class is not extendable.

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
@@ -1,21 +1,15 @@
 package pl.touk.nussknacker.ui.security.api
 
-import java.io.File
-import java.net.{URI, URL}
-import java.security.PublicKey
+import java.net.URI
 
-import com.typesafe.config.{Config, ConfigFactory}
-import pl.touk.nussknacker.engine.util.cache.CacheConfig
+import com.typesafe.config.Config
 import pl.touk.nussknacker.engine.util.config.ConfigFactoryExt
 import pl.touk.nussknacker.ui.security.api.AuthenticationConfiguration.{ConfigRule, ConfigUser}
-import pl.touk.nussknacker.ui.security.api.AuthenticationMethod.AuthenticationMethod
 import pl.touk.nussknacker.ui.security.api.GlobalPermission.GlobalPermission
 import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
-import scala.concurrent.duration._
-
 trait AuthenticationConfiguration {
-  def method: AuthenticationMethod
+  def name: String
   def usersFile: URI
 
   val userConfig: Config = ConfigFactoryExt.parseUri(usersFile)
@@ -23,14 +17,6 @@ trait AuthenticationConfiguration {
   lazy val users: List[ConfigUser] = AuthenticationConfiguration.getUsers(userConfig)
 
   lazy val rules: List[ConfigRule] = AuthenticationConfiguration.getRules(userConfig)
-}
-
-object AuthenticationMethod extends Enumeration {
-  type AuthenticationMethod = Value
-
-  val BasicAuth = Value("BasicAuth")
-  val OAuth2 = Value("OAuth2")
-  val Other = Value("Other")
 }
 
 object AuthenticationConfiguration {
@@ -42,8 +28,6 @@ object AuthenticationConfiguration {
   val methodConfigPath = s"$authenticationConfigPath.method"
   val usersConfigurationPath = "users"
   val rulesConfigurationPath = "rules"
-
-  def parseMethod(config: Config): AuthenticationMethod = config.as[AuthenticationMethod](methodConfigPath)
 
   def getUsers(config: Config): List[ConfigUser] = config.as[List[ConfigUser]](usersConfigurationPath)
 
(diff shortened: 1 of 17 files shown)


lciolecki:
In general it's ok. But.. You forgot about FE changes, mostly there: https://github.com/TouK/nussknacker/blob/staging/ui/client/containers/Auth/StrategySelector.tsx#L16.

trombka:
FE changes are here #1792

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticationConfiguration.scala
```diff
@@ -0,0 +1,30 @@
...
+import pl.touk.nussknacker.ui.security.api.{AuthenticationConfiguration, CachingHashesConfig}
+
+import java.net.URI
+
+case class BasicAuthenticationConfiguration(usersFile: URI,
+                                            cachingHashes: Option[CachingHashesConfig]) extends AuthenticationConfiguration {
+
+  override def method: String = BasicAuthenticationConfiguration.name
+
+  def cachingHashesOrDefault: CachingHashesConfig = cachingHashes.getOrElse(CachingHashesConfig.defaultConfig)
```
mproch, line 13: I know it's not your change, but this can be misleading - defaultConfig should be "merged" with the one configured by user...

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
```diff
@@ -19,7 +18,7 @@ trait AuthenticationConfiguration {
   def authSeverPublicKey: Option[PublicKey] = Option.empty
   def idTokenNonceVerificationRequired: Boolean
   def implicitGrantEnabled: Boolean
-  def method: AuthenticationMethod
+  def method: String
```
mproch, line 21: Could we use "name" from NamedServiceProvider here? Is there a situation where it can (and should) be different?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/loader/ScalaServiceLoader.scala
```diff
@@ -34,12 +32,12 @@ object ScalaServiceLoader {
     }
   }
 
-  def loadNamed[T<:NamedServiceProvider:ClassTag](name: String, classLoader: ClassLoader = Thread.currentThread().getContextClassLoader): T = {
+  def loadNamed[T<:NamedServiceProvider:ClassTag](name: String, default: => List[T] = Nil, classLoader: ClassLoader = Thread.currentThread().getContextClassLoader): T = {
     val available = load[T](classLoader)
     val className = implicitly[ClassTag[T]].runtimeClass.getName
-    Multiplicity(available.filter(_.name == name)) match {
+    Multiplicity(available.filter(_.name == name)).orElse(Multiplicity(default.filter(_.name == name))) match {
```
mproch, line 38: can we just have `available = load[T](classloader) ++ default` ?

mproch (commented):
Looks good, I added some minor comments

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/loader/ScalaServiceLoader.scala
trombka: It would make undistinguishable the two situations:
- when there is a single implementation loaded through SPI and a built-in one, both with the same name;
- when there are multiple implementations loaded through SPI with the same name.
The latter is treated as a misconfiguration.

mproch: for me it's ok - I mean, I would treat first one also as misconfiguration, I don't see any reason to do that

trombka: But you may just want to override a  built-in auth method without adding totally new behavior in the frontend app.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
trombka: Renamed

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticationConfiguration.scala
trombka: I do not fully understand what you mean by "merged" and do not want to handle it now since the PR has a different purpose. I just moved the `CachingHashaesConfig` class to the `basicauth` package where the class is solely used.

mproch: ok, let's leave it for other time ;)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationProvider.scala
```diff
@@ -19,12 +20,11 @@ trait AuthenticationProvider {
...
-    val loaded = ScalaServiceLoader.loadClass[AuthenticationProvider](classLoader) {
-      AuthenticationConfiguration.parseMethod(config) match {
-        case AuthenticationMethod.OAuth2 => OAuth2AuthenticationProvider()
-        case _ => BasicAuthenticationProvider()
-      }
-    }
+    val loaded = ScalaServiceLoader.loadNamed[AuthenticationProvider](
+      config.getString(AuthenticationConfiguration.methodConfigPath),
+      classLoader,
+      List(OAuth2AuthenticationProvider(), BasicAuthenticationProvider())
```
mproch, line 26: I wonder - do we need these defaults? Can't we just register those providers with ServiceLoader?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
```diff
@@ -1,21 +1,15 @@
...
 import pl.touk.nussknacker.ui.security.api.AuthenticationConfiguration.{ConfigRule, ConfigUser}
-import pl.touk.nussknacker.ui.security.api.AuthenticationMethod.AuthenticationMethod
 import pl.touk.nussknacker.ui.security.api.GlobalPermission.GlobalPermission
 import pl.touk.nussknacker.ui.security.api.Permission.Permission
 
-import scala.concurrent.duration._
-
 trait AuthenticationConfiguration {
-  def method: AuthenticationMethod
+  def name: String
```
mproch, line 12: I think we don't need it at all now. We only use name from AuthenticationProvider?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationProvider.scala
trombka: I have replaced the defaults with a services file.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
trombka: We cannot use the name from an AuthenticationProvider directly, because we need the name value in an AuthenticationResources instance and it is the AuthenticationProvider that creates AuthenticationResources and not the other way round. Instead, we could use a constant like `BasicAuthenticationProvider.name` there. Would it be better?

mproch (approved):

PR #1784, opened 2021-06-16
Some conceptual docs moved to docs-bak

arkadius:


arkadius:
closing because deeper refactoring of guides needed

PR #1785, opened 2021-06-16
[FIX] detecting erroneous projection/selection

mproch:


arkadius (approved):
LGTM but maybe let's merge https://github.com/TouK/nussknacker/pull/1772 first?

PR #1790, opened 2021-06-17
PeriodicProcessEvent: move deployment to sealed trait & make listeners closable

gadomsky:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessManager.scala
```diff
@@ -43,14 +43,16 @@ object PeriodicProcessManager {
...
+    val listener = listenerFactory.create(originalConfig)
+    val service = new PeriodicProcessService(delegate, jarManager, scheduledProcessesRepository, listener, additionalDeploymentDataProvider, clock)
     system.actorOf(DeploymentActor.props(service, periodicBatchConfig.deployInterval))
     system.actorOf(RescheduleFinishedActor.props(service, periodicBatchConfig.rescheduleCheckInterval))
 
     val toClose = () => {
       Await.ready(system.terminate(), 10 seconds)
       db.close()
       Await.ready(backend.close(), 10 seconds)
+      listener.close()
```
mproch, line 55: maybe it's better to close listener first? And/or wrap in some try/catch?

gadomsky: hmm, but as I can read in comments in the code:

> Listener is invoked during DB transaction,

Doesn't it imply that listers should be closed after db? 

As for wraping it in try catch, I will do that

mproch: I'd say that we should close in reverse order than we create. But - I was more thinking about order WRT akka/backend

mproch (approved):

PR #1794, opened 2021-06-17
Kafka topic parameter name configurable

arkadius:


gskrobisz (approved):
LGTM

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/SpecificRecordKafkaAvroSourceFactory.scala
```diff
@@ -39,7 +34,7 @@ class SpecificRecordKafkaAvroSourceFactory[V <: SpecificRecord: ClassTag](schema
 
         prepareSourceFinalResults(preparedTopic, valueValidationResult, context, dependencies, step.parameters)
 
-      case step@TransformationStep((TopicParamName, _) :: Nil, _) =>
-        prepareSourceFinalErrors(context, dependencies, step.parameters, List(CustomNodeError("Topic/Version is not defined", Some(TopicParamName))))
+      case step@TransformationStep((`topicParamName`, _) :: Nil, _) =>
+        prepareSourceFinalErrors(context, dependencies, step.parameters, List(CustomNodeError("Topic/Version is not defined", Some(topicParamName))))
```
lciolecki, line 38: What about `"Topic/Version`? Maybe you cane do something like this: `s"${topicParamName.capitalize} / version is not defined"` ?

arkadius: Good point but there was the other problem. These errors shouldn't be returned at all (take a look like it is done in generic avro source factory), because if there are some problems during topic validation, there are reported separately. Also we used  "Version" / "Schema version" parameter reference which wasn't declared. I've removed them.

PR #1798, opened 2021-06-18
Preview/java 8 fix

arkadius:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/types/EspTypeUtils.scala
```diff
@@ -135,14 +133,14 @@ object EspTypeUtils {
...
+      case t: ParameterizedType => extractGenericMonadReturnType(t)
       case t => None
     }
   }
 
   // This method should be used only for method's and field's return type - for method's parameters such unwrapping has no sense
-  private def extractGenericMonadReturnType(genericReturnType: ParameterizedTypeImpl): Option[TypingResult] = {
-    val rawType = genericReturnType.getRawType
+  private def extractGenericMonadReturnType(genericReturnType: ParameterizedType): Option[TypingResult] = {
+    val rawType = genericReturnType.getRawType.asInstanceOf[Class[_]]
```
mproch, line 143: I wonder - are we really sure it'll always be the case? Or maybe it'd be safer to write:
` case t: ParametrizedType if t.getRawType.isInstanceOf[Class[_]]` or sth like that?

build.sbt
```diff
@@ -163,16 +163,15 @@ lazy val commonSettings =
         "-language:existentials",
         "-Ypartial-unification",
         //Flink image is available only for jdk8
```
mproch, line 165: Let's write `we remain JDK8 compatible until 0.4.0 release` or sth like that

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/types/EspTypeUtils.scala
arkadius: changed

build.sbt
arkadius: By mistake missed this one flink's reference - replaced with other message. I don't want to specify if it will be exact 0.4.0 version here, because we haven't discussed it yet. Maybe we should check before how compatibility with java8 is broken in other similar projects?

mproch (approved):
LGTM

dswiecki (approved):

arkadius:
merged

PR #1799, opened 2021-06-18
Refactoring of AvroToJsonFormatter - part 2

gskrobisz:
Formatter is able to serialize key and value with schema ids and full ConsumerRecord metadata.

Continuation of https://github.com/TouK/nussknacker/pull/1783


engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+    * Step 1: Deserialize raw kafka event to GenericRecord/SpecificRecord domain.
+    * Step 2: Create Encoders that use ConfluentAvroMessageFormatter to convert avro object to json
+    * Step 3: Encode event's data with schema id's with derived encoder.
+    * @tparam K - key type passed from KafkaAvroSourceFactory, used to determine which datumReaderWriter use (e.g. specific or generic)
+    * @tparam V - value type passed from KafkaAvroSourceFactory, used to determine which datumReaderWriter use (e.g. specific or generic)
+    */
+  protected def doFormatRecord[K: ClassTag, V: ClassTag](kafkaConfig: KafkaConfig,
```
arkadius, line 44: Won't be easier to create `ConfluentAvroToJsonFormatter` class instead of passing such many parameters around?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,119 @@
...
+        keyOpt.map(keyJson => reader.readJson[K](keyJson, Option(keySchema), ConfluentUtils.keySubject(topic))).getOrElse(Array.emptyByteArray)
+      }
+      val valueSchema = reader.schemaById(record.valueSchemaId)
+      val valueBytes = reader.readJson[V](value, Option(valueSchema), ConfluentUtils.valueSubject(topic))
+      (keyBytes, valueBytes)
+    }
+    record.consumerRecord.toKafkaConsumerRecord(topic, serializeKeyValue)
+  }
+
+  private def createKeyEncoder[K: ClassTag](key: K, useStringForKey: Boolean, messageFormatter: ConfluentAvroMessageFormatter): Encoder[K] = {
```
arkadius, line 97: 1. Why key is passed here instead of using from argument in `apply`?  2. Can encoders be created once in formatter class (see comment above) and used many times (for each record)?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,119 @@
...
+  protected def doParseRecord[K: ClassTag, V: ClassTag](kafkaConfig: KafkaConfig,
+                                                        createReader: String => ConfluentAvroMessageReader)
+                                                       (topic: String, bytes: Array[Byte]): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    implicit val serializableRecordDecoder: Decoder[SerializableConsumerRecord[Json, Json]] = deriveDecoder
+    val consumerRecordDecoder: Decoder[AvroSerializableConsumerRecord[Json, Json]] = deriveDecoder
+    val record = CirceUtil.decodeJsonUnsafe(bytes)(consumerRecordDecoder)
+
+    def serializeKeyValue(keyOpt: Option[Json], value: Json): (Array[Byte], Array[Byte]) = {
+      val reader = createReader(topic)
+      val keyBytes = if (kafkaConfig.useStringForKey) {
```
arkadius, line 79: I don't exactly understand why this code has such many cases. when `useStringForKey` is true, shouldn't you expect that keyOpt is always  `Some(j) if j.isString`?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+  *
+  * @param deserializationSchema - schema used to convert raw kafka event to serializable representation (see SerializableConsumerRecord)
+  * @param serializeKeyValue - converts key and value from K-V domain to Array[Byte]
+  * @tparam K - event key type
+  * @tparam V - event value type
+  */
+class ConsumerRecordToJsonFormatterFactory[K:Encoder:Decoder, V:Encoder:Decoder] extends RecordFormatterFactory {
+
+  override def create[KK: ClassTag, VV: ClassTag](kafkaConfig: KafkaConfig, kafkaSourceDeserializationSchema: KafkaDeserializationSchema[ConsumerRecord[KK, VV]]): RecordFormatter = {
+    new RecordFormatter {
```
arkadius, line 25: Can you extract it to separate class? Let it be in the same file. Currently it is hard to override this class. I know that "prefer composition over inheritance" rule but sometimes... ;)

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
gskrobisz: Created `ConfluentAvroToJsonFormatter`

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
gskrobisz: I see it necessary to handle null and emptyByteArray key values.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
gskrobisz: 1 & 2 Fixed.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
```diff
@@ -22,9 +23,18 @@ private[confluent] class ConfluentAvroMessageFormatter(schemaRegistryClient: Sch
 
   schemaRegistry = schemaRegistryClient
 
-  def writeTo(data: Array[Byte], output: PrintStream): Unit = {
-    val obj = deserialize(data)
+  def getSchemaIdOpt(bytes: Array[Byte]): Option[Int] = {
```
arkadius, line 26: Why this method is here? It looks strange when some class has multiple public methods that are not connected. WDYT about moving it to `ConfluentUtils.readIdOpt`? Or maybe just inline invocation: `Option(bytes).map(ConfluentUtils.readInt)`. Why this check: `avroContent.nonEmpty` is necessary? Why not to just throw exception like in this inlined example?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -1,94 +1,64 @@
...
 
-  private val keySubject = ConfluentUtils.keySubject(topic)
-
-  private val valueSubject = ConfluentUtils.valueSubject(topic)
-
   private val decoderFactory = DecoderFactory.get
 
-  // TODO: This implementation won't handle separator escaping
-  def readMessage(str: String, keySchema: Option[Schema], valueSchema: Option[Schema]): ConsumerRecord[Array[Byte], Array[Byte]] = {
+  def schemaById(schemaId: Int): Schema = {
```
arkadius, line 29: This reader doesn't read schemaById, it is just fetched from schema registry. Can you move this to place of invocation?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -1,94 +1,64 @@
...
-              keyString.getBytes(StandardCharsets.UTF_8)
-            } else {
-              val key = jsonToAvro(keyString, keySchema.get)
-              serializeImpl(keySubject, key, new AvroSchema(keySchema.get))
-            }
-            val value = jsonToAvro(valueString, valueSchema.get)
-            val serializedValue = serializeImpl(valueSubject, value, new AvroSchema(valueSchema.get))
-            new ConsumerRecord(topic, 0, 0L, serializedKey, serializedValue)
-        }
+      schemaOpt match {
```
arkadius, line 36: Returning `emptyByteArray` is not too much intuitive. Can we make this class easyly handle only defined schema and handle all edge cases in one places - in formatter?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,121 @@
...
+  override protected def parseRecord(topic: String, bytes: Array[Byte]): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    val record = CirceUtil.decodeJsonUnsafe(bytes)(consumerRecordDecoder)
+
+    def serializeKeyValue(keyOpt: Option[Json], value: Json): (Array[Byte], Array[Byte]) = {
+      val reader = createReader(topic)
+      val keyBytes = if (kafkaConfig.useStringForKey) {
+        keyOpt match {
+          // we handle strings this way because we want to keep result value compact and JString is formatted in quotes
+          case Some(j) if j.isString => j.asString.get.getBytes(StandardCharsets.UTF_8)
+          case Some(j) => j.noSpaces.getBytes(StandardCharsets.UTF_8)
```
arkadius, line 71: Why this case must be handled? Don't we always have raw "String" for this flag?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,121 @@
...
+  protected val consumerRecordEncoder: Encoder[AvroSerializableConsumerRecord[K, V]] = deriveEncoder
+
+}
+
+object ConfluentAvroToJsonFormatter {
+
+  def createKeyEncoder[K: ClassTag](useStringForKey: Boolean, messageFormatter: ConfluentAvroMessageFormatter): Encoder[K] = {
+    new Encoder[K] {
+      override def apply(key: K): Json = if (useStringForKey) {
+        Json.fromString(key.asInstanceOf[String])
```
arkadius, line 102: Is this code is correct? Shouldn't `serializeKeyValue(encoder.apply(key), ...)` always return `key`? If so, correct would be just `key.asInstanceOf[String]` ... I wonder if we can have `serializeKey` near `encode` and the same for `value`. Maybe it could be in the class instead of companion object?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -1,94 +1,64 @@
...
-  *
   * @param schemaRegistryClient schema registry client
   * @param topic topic
-  * @param keySeparator key separator
   */
-private[confluent] class ConfluentAvroMessageReader(schemaRegistryClient: SchemaRegistryClient,
-                                                    topic: String,
-                                                    useStringForKey: Boolean,
-                                                    keySeparator: String)
+private[confluent] class ConfluentAvroMessageReader(schemaRegistryClient: SchemaRegistryClient, topic: String)
```
arkadius, line 22: `topic` is not used, after removing it you can in factory pass just `reader` and not create reader for each record

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
gskrobisz: Fixed.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
gskrobisz: Fixed

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
gskrobisz: True, fixed.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageFormatter.scala
gskrobisz: Fixed.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
gskrobisz: Fixed

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
gskrobisz: Fixed

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -1,94 +1,53 @@
...
-            }
-            val value = jsonToAvro(valueString, valueSchema.get)
-            val serializedValue = serializeImpl(valueSubject, value, new AvroSchema(valueSchema.get))
-            new ConsumerRecord(topic, 0, 0L, serializedKey, serializedValue)
-        }
-      }
+      val avroObj = jsonToAvro[T](jsonObj, schema)
+      val serializedValue = serializeImpl(subject, avroObj, new AvroSchema(schema))
+      serializedValue
     } catch {
```
arkadius, line 32: can you remove this line and return result instantly?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroMessageReader.scala
```diff
@@ -1,94 +1,53 @@
...
+      case ex: Exception =>
         throw new SerializationException("Error reading from input", ex)
     }
   }
 
-  private def jsonToAvro(jsonString: String, schema: Schema) = {
+  private def jsonToAvro[T: ClassTag](jsonObj: Json, schema: Schema): AnyRef = {
     try {
-      val reader = createDatumReader(schema, schema, useSchemaReflection = false, useSpecificAvroReader = false)
+      val jsonString = jsonObj.noSpaces
```
arkadius, line 40: can you move this line outside the try-catch block and use it in handling exception - will be visible, that this method can't throw exception

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/formatter/ConfluentAvroToJsonFormatterFactory.scala
```diff
@@ -0,0 +1,120 @@
...
+    def serializeKeyValue(keyOpt: Option[Json], value: Json): (Array[Byte], Array[Byte]) = {
+      val keyBytes = if (kafkaConfig.useStringForKey) {
+        keyOpt match {
+          // we handle strings this way because we want to keep result value compact and JString is formatted in quotes
+          case Some(j) if j.isString => j.asString.get.getBytes(StandardCharsets.UTF_8)
+          case None => null
+        }
+      } else {
+        val keySchema = record.keySchemaId.map(id => getSchemaById(id)).getOrElse(throw new IllegalArgumentException("Error reading schema: empty schema id"))
+        keyOpt.map(keyJson => messageReader.readJson[K](keyJson, keySchema, ConfluentUtils.keySubject(topic))).getOrElse(Array.emptyByteArray)
```
arkadius, line 76: Instead of returning empty array, please throw exception with message showing that for avro keys we expected defined, properly formatted key

arkadius (approved):

PR #1804, opened 2021-06-18
Fix for: SpecificRecordKafkaAvroSourceFactory produced GenericRecord …

arkadius:
…instead of Specific (getters and other additional methods were missing)


gskrobisz (approved):
LGTM

PR #1809, opened 2021-06-21
Performance optimization for aggregates: do not update state if added…

arkadius:
… element is neutral for current state

### DIFF
--- a/engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
+++ b/engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
@@ -98,11 +98,15 @@ trait AggregatorFunctionMixin[MapT[K,V]] { self: StateHolder[MapT[Long, AnyRef]]
                                   out: Collector[ValueWithContext[AnyRef]]): MapT[Long, aggregator.Aggregate] = {
     val newElementInStateTimestamp = computeTimestampToStore(timestamp)
     val newElement = value.value.value.asInstanceOf[aggregator.Element]
-    val newState = computeNewState(newElementInStateTimestamp, newElement)
+    val (newState, stateWasChanged) = computeNewState(newElementInStateTimestamp, newElement)
 
-    updateState(newState.asInstanceOf[MapT[Long, AnyRef]], newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+    if (stateWasChanged) {
+      updateState(newState.asInstanceOf[MapT[Long, AnyRef]], newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+      retrievedBucketsHistogram.update(newState.toScalaMapRO.size)
+    } else {
+      doMoveEvictionTime(newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+    }
     handleElementAddedToState(newElementInStateTimestamp, newElement, value.context, timeService, out)
-    retrievedBucketsHistogram.update(newState.toScalaMapRO.size)
     newState
   }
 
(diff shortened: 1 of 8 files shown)


engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/CardinalityAggregator.scala
```diff
@@ -84,5 +87,20 @@ private[aggregate] abstract class CardinalityWrapper extends Value with KryoSeri
     in.read(bytes)
     wrapped = readFromBytes(bytes)
   }
+
+  def canEqual(other: Any): Boolean = other.isInstanceOf[CardinalityWrapper]
+
+  override def equals(other: Any): Boolean = other match {
```
mproch, line 93: can we use some builder for that, e.g. from commons? Is this null safe?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -48,6 +50,9 @@ object aggregates {
 
     override def zero: Number = null
 
+    override def isNeutralForAccumulator(element: Element, currentAggregate: Aggregate): Boolean =
+      addElement(element, currentAggregate) == currentAggregate
```
mproch, line 54: maybe this should be the default implementation?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -98,11 +98,15 @@ trait AggregatorFunctionMixin[MapT[K,V]] { self: StateHolder[MapT[Long, AnyRef]]
                                   out: Collector[ValueWithContext[AnyRef]]): MapT[Long, aggregator.Aggregate] = {
     val newElementInStateTimestamp = computeTimestampToStore(timestamp)
     val newElement = value.value.value.asInstanceOf[aggregator.Element]
-    val newState = computeNewState(newElementInStateTimestamp, newElement)
+    val (newState, stateWasChanged) = computeNewState(newElementInStateTimestamp, newElement)
 
-    updateState(newState.asInstanceOf[MapT[Long, AnyRef]], newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+    if (stateWasChanged) {
+      updateState(newState.asInstanceOf[MapT[Long, AnyRef]], newElementInStateTimestamp + timeWindowLengthMillis, timeService)
+      retrievedBucketsHistogram.update(newState.toScalaMapRO.size)
```
mproch, line 105: I think retrievedBucketsHistogram name suggests we count retrieved, not updated? Maybe we can add separate metric for that?

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/CardinalityAggregator.scala
arkadius: This generated by Idea equals is null safe because null isn't handled by: `case that: CardinalityWrapper =>` but by default case. hashcode wasn't null-safe - I've replaced with `Objects.hash`.

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
arkadius: Fixed

engine/flink/model-util/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
arkadius: Ahh I forgot about comment how it works. Let's keep only `retrievedBucketsHistogram` - not sure if `updatedBucketsHistogram` is really needed

mproch (approved):
:+1:

Issue #1156, opened 2020-08-20
Demo should work on generic model


arkadius:
The idea is to make user will start with generic model and to show how to extend it to make it usable with some specific domain (like currently implemented model for financial domain)

Steps to make it happen:
- Some script that will add transaction/client/alert schema to schema registry
- Replace `transactionAmountAggregator`/`eventsCounter` with some generic aggregation or add some ability to extend given model with some extra nodes
- Replace `clientService` with service integrating with openapi?
- Remove demo module or some classes?

PR #1817, opened 2021-06-22
Generic model in demo

dswiecki:


nussknacker-dist/src/universal/conf/demo-application.conf
```diff
@@ -1,17 +1,19 @@
 base: { include "base-application.conf" }
 
-environment: "demo"
+environment: "local"
```
mproch, line 3: Do we need this file anymore? I think it can be removed?

demo/docker/testData/transactions-value.json
```diff
@@ -0,0 +1,15 @@
+{
```
mproch, line 1: what aboud alerts-value? Can we add schemas automatically

nussknacker-dist/src/universal/conf/demo-application.conf
dswiecki: Removed

demo/docker/testData/transactions-value.json
dswiecki: I've created files with Schema Registry definition for all of the subjects and added script to load them into the Schema Registry

arkadius (approved):
LGTM

dswiecki:
Closes #1156

mproch (approved):
:+1:

PR #1839, opened 2021-06-28
Change explicitUidInStatefulOperators to true by default

dswiecki:


engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/compat/ExplicitUidInOperatorsSupport.scala
```diff
@@ -67,13 +68,10 @@ object ExplicitUidInOperatorsSupport {
       stream
   }
 
-  @PublicEvolving // default behaviour will be switched to true in some future
```
mproch, line 70: let's keep PublicEvolving?

dswiecki: Done

mproch (approved):

PR #1847, opened 2021-06-29
Add comment when updating process with the same version

jedrz:
Just to be sure the process was migrated/synchronized. Currently when there is no difference between processes, on remote enviroment no comment is added, but on local environment a notification is displayed about successful migration.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.ui.process.repository
 
 import java.time.LocalDateTime
+
 import cats.data._
 import cats.syntax.either._
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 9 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -78,22 +81,31 @@ class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTy
...
+            .map(_.right.map(_.newVersion))
         }
       }
     }
 
     insertAction
   }
 
-  def updateProcess(updateProcessAction: UpdateProcessAction)(implicit loggedUser: LoggedUser): DB[XError[Option[ProcessVersionEntityData]]] =
+  def updateProcess(updateProcessAction: UpdateProcessAction)(implicit loggedUser: LoggedUser): DB[XError[ProcessUpdated]] = {
```
lciolecki, line 94: I wonder about move adding comment in transaction way at ProcessService. WDYT?

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/ProcessPosting.scala
```diff
@@ -19,9 +19,9 @@ class ProcessPosting {
     toRequest(ProcessConverter.toDisplayable(ProcessCanonizer.canonize(process), TestProcessingTypes.Streaming))
   }
 
-  def toEntityAsProcessToSave(process: EspProcess): RequestEntity = {
+  def toEntityAsProcessToSave(process: EspProcess, comment: String = ""): RequestEntity = {
```
lciolecki, line 22: `comment` as `Option[String]`?

lciolecki (approved):
LGTM, but please look at below comments.

ui/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/ProcessPosting.scala
jedrz: It's required to pass comment in the request body, that's the reason it's mandatory here.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
jedrz: I added an comment about extracting logic.

PR #1850, opened 2021-06-29
Add delayed sources (generic and avro)

gskrobisz:


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/DelayedFlinkKafkaConsumer.scala
```diff
@@ -0,0 +1,120 @@
...
+    while (delay > latency) {
+      val eventDelay = delay - latency
+      val sleepTime = Math.min(maxSleepTime, eventDelay)
+
+      val logMessage = s"Sleeping for $sleepTime ms of total $eventDelay ms for ${records.size()} events. Max event timestamp is $maxEventTimestamp, fetcher delay is $delay."
+
+      if (sleepTime < maxSleepTime) {
+        logger.debug(logMessage)
+      } else {
+        logger.info(logMessage)
```
arkadius, line 111: Why in case when sleepTime exceed maxSleepTime we only log it as an info and sleep nonetheless? Shouldn't be at least some todo/comment here? or maybe we should sleep only maxSleepTime?

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/DelayedKafkaAvroSourceFactory.scala
```diff
@@ -0,0 +1,82 @@
...
+        case Valid((valueRuntimeSchema, _)) =>
+          val definition = valueRuntimeSchema.map(_.schema.getFields.asScala.map(f => (f.name(), ())).toMap ).getOrElse(Map.empty).asJava
+          val delayValidationErrors = Option(delay.asInstanceOf[java.lang.Long]).map(d => validateDelay(d)).getOrElse(Nil)
+          val timestampValidationErrors = Option(field.asInstanceOf[String]).map(f => validateTimestampField(f, definition)).getOrElse(Nil)
+          val errors = delayValidationErrors ++ timestampValidationErrors
+          prepareSourceFinalResults(preparedTopic, valueValidationResult, context, dependencies, step.parameters, errors)
+        case Invalid(exc) =>
+          prepareSourceFinalErrors(context, dependencies, step.parameters, List(exc))
+      }
+    case step@TransformationStep((`topicParamName`, _) :: (SchemaVersionParamName, _) :: (TypeDefinitionParamName, _) :: (TimestampFieldParamName, _) :: (DelayParameterName, _) :: Nil, _) =>
```
arkadius, line 56: Why this timestamp field is needed? Until now we have timstamp assignment logic controlled by timestamphandler. Shouldn't be done the same way here? Or maybe timestamp assignment logic should be configurable for all type of kafka sources the same way?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/BaseKafkaDelayedSourceFactory.scala
```diff
@@ -0,0 +1,26 @@
+package pl.touk.nussknacker.engine.kafka.generic
+
+import org.apache.flink.streaming.connectors.kafka.KafkaDeserializationSchema
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import pl.touk.nussknacker.engine.flink.api.process.FlinkContextInitializer
+import pl.touk.nussknacker.engine.flink.api.timestampwatermark.TimestampWatermarkHandler
+import pl.touk.nussknacker.engine.kafka.{KafkaConfig, PreparedKafkaTopic, RecordFormatter}
+import pl.touk.nussknacker.engine.kafka.source.KafkaSource
+
+trait BaseKafkaDelayedSourceFactory {
```
arkadius, line 10: IMO we should write some descriptive comment why we need this feature instead of using just delay component. I guess that it is usefull in case when user don't want to have big state with scheduled mssages and just want to have this "saved" it on kafka topic. If so, it looks like very technical optimization details and I wonder if we can done this without involving user - e.g. by some graph optimization checking if there is a sequence of nodes: source -> delay and in this case replace source by this source.

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/DelayedKafkaAvroSourceFactory.scala
```diff
@@ -0,0 +1,82 @@
...
+                                      deserializationSchema: KafkaDeserializationSchema[ConsumerRecord[K, V]],
+                                      timestampAssigner: Option[TimestampWatermarkHandler[ConsumerRecord[K, V]]],
+                                      formatter: RecordFormatter,
+                                      flinkContextInitializer: FlinkContextInitializer[ConsumerRecord[K, V]]): KafkaSource[ConsumerRecord[K, V]] = {
+    val timestampFieldName = extractTimestampField(params)
+    val extractTimestamp: (ConsumerRecord[K, V], Long) => Long = (consumerRecord, kafkaEventTimestamp) => {
+      Option(timestampFieldName)
+        .map(f => consumerRecord.value().asInstanceOf[GenericRecord].get(f).asInstanceOf[Long])
+        .getOrElse(kafkaEventTimestamp)
+    }
```
arkadius, line 74: 1. We should keep logic of timestamp extraction consistent - if user define timestamp extraction strategy, it should override timestamp assignment logic. The best would be passing timestamp assigner to `DelayedFlinkKafkaConsumer`
2. Why not to use LazyParameter for timestamp instead of fixing this implementation to `GenericRecord`? 
3. Let's add TODO that this timestamp extraction logic configured in parameter (instead of events normaliation in ETLs + configured timestamp assigner) should be also reused in normal kafka (avro/json) sources

arkadius (changes requested):
Two critical comments added:
- regards do description of technical details
- regards to timestamp extraction

engine/flink/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/DelayedKafkaAvroSourceFactory.scala
gskrobisz: We want to allow users to calculate delay related to their selected field that represents "business timestamp".

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/BaseKafkaDelayedSourceFactory.scala
```diff
@@ -0,0 +1,46 @@
...
+                                               flinkContextInitializer: FlinkContextInitializer[ConsumerRecord[K, V]],
+                                               delay: Long): KafkaSource[ConsumerRecord[K, V]] = {
+    new KafkaSource[ConsumerRecord[K, V]](preparedTopics, kafkaConfig, deserializationSchema, timestampAssigner, formatter) {
+      override val contextInitializer: FlinkContextInitializer[ConsumerRecord[K, V]] = flinkContextInitializer
+      override protected def createFlinkSource(consumerGroupId: String) =
+        new DelayedFlinkKafkaConsumer(preparedTopics, deserializationSchema, kafkaConfig, consumerGroupId, delay)
+    }
+  }
+
+  protected def prepareTimestampAssigner[K, V](kafkaConfig: KafkaConfig, extract: ConsumerRecord[K, V] => Long): TimestampWatermarkHandler[ConsumerRecord[K, V]] = {
```
arkadius, line 37: Why not to add this method to `StandardTimestampWatermarkHandler` instead of duplicating it?

arkadius (approved):
LGTM, one "nice to have" comment added.

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/KafkaDelayedSourceFactory.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  val delayValidators = List(MinimalNumberValidator(0), MaximalNumberValidator(Long.MaxValue))
+
+  final val DelayParameterName = "delayInMillis"
+
+  final val DelayParameter = Parameter.optional(DelayParameterName, Typed[java.lang.Long])
+
+  final val TimestampFieldParamName = "timestampField"
+
+  // TODO: consider changing to lazy parameter
```
arkadius, line 21: Can you extend this comment to "and add the same parameter also in "not delayed" kafka sources?

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/BaseKafkaDelayedSourceFactory.scala
gskrobisz: Method removed. Needed to revert the merge of timestampAssigner (used to modify created stream that has ready source function -> `FlinkIntermediateRawSource`) and timestampExtractor (used to create source function with KafkaFetcher and addDiscoveredPartitions). It still seems that those extractors can be combined but it should be created with more caution (this is the same WatermarkStrategy but in the moment of event emission it is null and extractor uses raw kafka timestamp)

engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/generic/KafkaDelayedSourceFactory.scala
gskrobisz: Comment added.

gskrobisz:
Closing PR, changes moved to tested preview-branch, see https://github.com/TouK/nussknacker/pull/1894

PR #1868, opened 2021-07-01
Category-permission mapping extracted from AuthenticationProvider

trombka:
Solves https://github.com/TouK/nussknacker/blob/staging/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationProvider.scala#L15
AuthenticationProvider no longer delivers a LoggedUser but an AuthenticatedUser instance with a roles collection. Permission mapping is moved to an additional LoggedUser apply method.

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -15,10 +15,6 @@ sealed trait LoggedUser {
 }
 
 object LoggedUser {
-  def apply(authenticatedUser: AuthenticatedUser, config: Config, processCategories: List[String]): LoggedUser = {
-    apply(authenticatedUser, AuthenticationConfiguration.getRules(config), processCategories)
-  }
-
   def apply(authenticatedUser: AuthenticatedUser, rules: List[ConfigRule], processCategories: List[String]): LoggedUser = {
     val rulesSet = RulesSet.getOnlyMatchingRules(authenticatedUser.roles, rules, processCategories)
     apply(id = authenticatedUser.id, username = authenticatedUser.username, rulesSet = rulesSet)
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
@@ -1,7 +1,11 @@
 package pl.touk.nussknacker.ui.security.api
 
+import com.typesafe.config.Config
+import pl.touk.nussknacker.ui.security.api.AuthenticationConfiguration.ConfigRule
 import pl.touk.nussknacker.ui.security.api.GlobalPermission.GlobalPermission
 import pl.touk.nussknacker.ui.security.api.Permission.Permission
+
+
 
 sealed trait LoggedUser {
   val id: String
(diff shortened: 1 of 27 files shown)


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -11,19 +15,28 @@ sealed trait LoggedUser {
...
-             categoryPermissions: Map[String, Set[Permission]] = Map.empty,
-             globalPermissions: List[GlobalPermission] = Nil,
-             isAdmin: Boolean = false): LoggedUser = {
-     if (isAdmin) {
-       AdminUser(id, username)
-     } else {
-       CommonUser(id, username, categoryPermissions, globalPermissions)
-     }
+  def apply(authenticatedUser: AuthenticatedUser, config: Config, processCategories: List[String]): LoggedUser = {
+    apply(authenticatedUser, AuthenticationConfiguration.getRules(config), processCategories)
```
mproch, line 19: we read config with each invocation here. Could we do this only once at application start?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -11,19 +15,28 @@ sealed trait LoggedUser {
...
+            isAdmin: Boolean = false): LoggedUser = {
+    if (isAdmin) {
+      AdminUser(id, username)
+    } else {
+      CommonUser(id, username, categoryPermissions, globalPermissions)
+    }
   }
 
-  def apply(id: String, username: String, rulesSet: RulesSet): LoggedUser = {
+  def apply(id: String, username: String, roles: List[String], rulesSet: RulesSet): LoggedUser = {
```
mproch, line 39: roles are not used here?

engine/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactorySpec.scala
```diff
@@ -65,18 +65,22 @@ class ExampleOAuth2ServiceFactorySpec extends FlatSpec with Matchers with Patien
     }.futureValue
   }
 
+/* These test probably test the behaviour that is not desired any more
+i.e. assigning category permissions directly by an AuthenticationProvider
+and not by a roles-to-category-permission mapping
```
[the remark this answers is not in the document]
trombka: Removed. I have also added a test in LoggedUserTest to test the roles to permissions mapping.

engine/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactory.scala
```diff
@@ -88,6 +94,15 @@ object ExampleOAuth2ServiceFactory {
       None
     )
 
+  val testRules: List[AuthenticationConfiguration.ConfigRule] = AuthenticationConfiguration.getRules(testConfig.usersFile)
+//  def testRules = List(
```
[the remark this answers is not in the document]
trombka: Removed.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
trombka: Right. Garbage. I thought of passing roles along with permissions in LoggedUser but then dropped the idea. Removed.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
trombka: I have moved config reading to the main application class.

trombka: A separate commit so you can see the difference.

mproch (approved):

PR #1885, opened 2021-07-06
Improvements: Process Toolbars

lciolecki:
- Added encodeing process name in urls

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/service/ProcessToolbarService.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/service/ProcessToolbarService.scala
@@ -2,6 +2,7 @@ package pl.touk.nussknacker.ui.service
 
 import com.typesafe.config.Config
 import io.circe.generic.JsonCodec
+import pl.touk.nussknacker.engine.util.UriUtils
 import pl.touk.nussknacker.restmodel.processdetails.BaseProcessDetails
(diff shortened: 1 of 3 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/service/ProcessToolbarService.scala
```diff
@@ -92,10 +93,13 @@ private [service] object ToolbarHelper {
   def createProcessToolbarId(config: ProcessToolbarsConfig, process: BaseProcessDetails[_]): String =
     s"${config.uuidCode}-${if(process.isArchived) "archived" else "not-archived"}-${if(process.isSubprocess) "subprocess" else "process"}"
 
-  def fillByProcessData(text: String, process: BaseProcessDetails[_]): String =
+  def fillByProcessData(text: String, process: BaseProcessDetails[_], urlOption: Boolean = false): String = {
+    val processName = if (urlOption) UriUtils.encodeURIComponent(process.name) else process.name
```
piotrp, line 97: Is it ok to pass unescaped process name to `name` and `title`? Won't it break if we have names with eg. `"`?

lciolecki: Right now at fe w don't escape process name and we don't , so I think it will be ok.

piotrp (approved):

PR #1892, opened 2021-07-07
Split openapi component from generic model

dswiecki:


nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -6,14 +6,14 @@ categoriesConfig: {
   "Default": "streaming"
 }
 
-modelClassPath: "model/genericModel.jar"
+modelClassPath: ["model/genericModel.jar"]
```
arkadius, line 9: why `model/openapi.jar` is not set in default classpath?

build.sbt
```diff
@@ -345,7 +345,8 @@ lazy val dist = {
       mappings in Universal ++= Seq(
         (crossTarget in generic).value / "genericModel.jar" -> "model/genericModel.jar",
         (crossTarget in flinkProcessManager).value / s"nussknacker-flink-manager.jar" -> "managers/nussknacker-flink-manager.jar",
-        (crossTarget in engineStandalone).value / s"nussknacker-standalone-manager.jar" -> "managers/nussknacker-standalone-manager.jar"
+        (crossTarget in engineStandalone).value / s"nussknacker-standalone-manager.jar" -> "managers/nussknacker-standalone-manager.jar",
+        (crossTarget in openapi).value / "openapi.jar" -> "model/openapi.jar"
```
arkadius, line 349: WDYT about some nested directory for additional component providers like `model/components` I see two benedfits:
- it will be clear that there are not models but just plugins with extra components
- in the futher we could add some discovery mechanizm that will check this directory and add all plugins to model's classpath

dswiecki: Done

nussknacker-dist/src/universal/conf/application.conf
dswiecki: It separates model in base configuration from components, also demo shows how to extend with your own components

arkadius (approved):

PR #1895, opened 2021-07-08
Static method validation

mbz-touk:
Added functionality for validating classes passed as TypeReference with flag for enabling/disabling that functionality.

### DIFF
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
@@ -15,7 +15,7 @@ case class ExpressionConfig(globalProcessVariables: Map[String, WithCategories[A
                             dictionaries: Map[String, WithCategories[DictDefinition]] = Map.empty,
                             hideMetaVariable: Boolean = false,
                             strictMethodsChecking: Boolean = true,
-                            referenceTypeValidating: Boolean = false
+                            staticMethodInvocationsChecking: Boolean = false
                            )
 
 object ExpressionConfig {
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
@@ -14,7 +14,8 @@ case class ExpressionConfig(globalProcessVariables: Map[String, WithCategories[A
                             //      DictInstance have the same dictId as DictDefinition
                             dictionaries: Map[String, WithCategories[DictDefinition]] = Map.empty,
                             hideMetaVariable: Boolean = false,
-                            strictMethodsChecking: Boolean = true
+                            strictMethodsChecking: Boolean = true,
+                            referenceTypeValidating: Boolean = false
                            )
 
 object ExpressionConfig {
--- a/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
+++ b/engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
@@ -25,6 +25,8 @@ object typing {
 
     def display: String
 
+    def javaClassName :String
+
   }
 
   sealed trait KnownTypingResult extends TypingResult
@@ -57,6 +59,7 @@ object typing {
 
     override def display: String = fields.map { case (name, typ) => s"$name: ${typ.display}"}.mkString("{", ", ", "}")
 
+    override def javaClassName: String = ???
   }
 
   case class TypedDict(dictId: String, valueType: SingleTypingResult) extends SingleTypingResult {
@@ -67,6 +70,7 @@ object typing {
 
     override def display: String = s"Dict(id=$dictId)"
 
+    override def javaClassName: String = ???
   }
 
   case class TypedTaggedValue(underlying: SingleTypingResult, tag: String) extends SingleTypingResult {
@@ -75,6 +79,7 @@ object typing {
 
     override def display: String = s"${underlying.display} @ $tag"
 
+    override def javaClassName: String = ???
   }
 
   // Unknown is representation of TypedUnion of all possible types
@@ -82,6 +87,7 @@ object typing {
 
     override val display = "Unknown"
 
+    override def javaClassName: String = ???
   }
 
   // constructor is package protected because you should use Typed.apply to be sure that possibleTypes.size > 1
@@ -94,10 +100,13 @@ object typing {
       case many => many.map(_.display).mkString(" | ")
     }
 
+    override def javaClassName: String = ???
   }
 
   //TODO: make sure parameter list has right size - can be filled with Unknown if needed
   case class TypedClass private[typing] (klass: Class[_], params: List[TypingResult]) extends SingleTypingResult {
+
+    override def javaClassName :String = ReflectUtils.fixedClassSimpleNameWithoutParentModule(ClassUtils.primitiveToWrapper(klass))
 
     //TODO: should we use simple name here?
     override def display: String = {
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -2,43 +2,34 @@ package pl.touk.nussknacker.engine
 
 import org.springframework.expression.spel.SpelNode
 import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+import pl.touk.nussknacker.engine.spel.ast.SpelAst.RichSpelNode
 
 
 case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = Set()) {
 
-  def validateTypeReference(spelNode: SpelNode) : Boolean = {
+  def validateTypeReference(spelNode: SpelNode): Boolean = {
 
+    val spelNodeChildAST = spelNode.children.headOption.getOrElse(throw new Exception("SpelNode has no children")).toStringAST
 
-    if(spelNode.getChildCount < 1) {
-      //todo: przechwycić błąd w razie jakby nie miało jednego child node'a
-    }
-
-    val spelNodeChildAST = spelNode.getChild(0).toStringAST
-
-    typeDefinitions.foreach(typeDefinition => {
-      if(typeDefinition.clazzName.display.equals(spelNodeChildAST))
-        return true
-    })
-    false
+    typeDefinitions.exists(typeDefinition => typeDefinition.clazzName.display.equals(spelNodeChildAST))
   }
 
-  def displayBasicInfo(spelNode : SpelNode) = {
+  def displayBasicInfo: String = {
 
-    println("spelNode: ")
-    println(spelNode.toStringAST)
-    println(spelNode.getChild(0).toStringAST)
-    println
+    val newLine = "\n"
+    val tab = "\t"
+    val basicInfo: StringBuilder = new StringBuilder()
+
     typeDefinitions.foreach(element => {
 
-      println(element.clazzName.display)
-      println("\tmethods: ")
+      basicInfo.append(newLine ++ element.clazzName.display ++ newLine ++ tab ++ "methods: " ++ newLine)
       element.methods.foreach(method => {
-        println("\t\t"+method._1)
-        println("\t\t"+method._2)
-        println
+        basicInfo.append(
+          newLine ++ tab ++ tab ++ method._1 ++ tab ++ method._2.toString ++ newLine
+        )
       })
-      println
     })
+    basicInfo.result()
 
   }
 }
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -18,7 +18,7 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
 
     val newLine = "\n"
     val tab = "\t"
-    val basicInfo: StringBuilder = new StringBuilder()
+    val basicInfo = new StringBuilder()
 
     typeDefinitions.foreach(element => {
 
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -1,17 +1,26 @@
 package pl.touk.nussknacker.engine
 
+import cats.data.{NonEmptyList, Validated}
+import cats.data.Validated.{Invalid, Valid}
 import org.springframework.expression.spel.SpelNode
+import pl.touk.nussknacker.engine.api.expression.ExpressionParseError
+import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+import pl.touk.nussknacker.engine.spel.TypedNode
 import pl.touk.nussknacker.engine.spel.ast.SpelAst.RichSpelNode
 
 
-case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = Set()) {
+class TypeDefinitionSet(typeDefinitions: Map[String, TypeInfos.ClazzDefinition]) {
 
-  def validateTypeReference(spelNode: SpelNode): Boolean = {
+  def validateTypeReference(spelNode: SpelNode): Validated[NonEmptyList[ExpressionParseError], SpelNode] = {
 
     val spelNodeChildAST = spelNode.children.headOption.getOrElse(throw new Exception("SpelNode has no children")).toStringAST
 
-    typeDefinitions.exists(typeDefinition => typeDefinition.clazzName.display.equals(spelNodeChildAST))
+    if(typeDefinitions.contains(spelNodeChildAST)) {
+      Valid(spelNode)
+    } else {
+      Invalid(NonEmptyList.of(ExpressionParseError("Class is not allowed to be passed as TypeReference")))
+    }
   }
 
   def displayBasicInfo: String = {
@@ -20,8 +29,8 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
     val tab = "\t"
     val basicInfo = new StringBuilder()
 
-    typeDefinitions.foreach(element => {
-
+    typeDefinitions.keySet.foreach(key => {
+      val element = typeDefinitions.getOrElse(key, throw new Exception(s"No value found for key ${key}"))
       basicInfo.append(newLine ++ element.clazzName.display ++ newLine ++ tab ++ "methods: " ++ newLine)
       element.methods.foreach(method => {
         basicInfo.append(
@@ -31,5 +40,15 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
     })
     basicInfo.result()
 
+  }
+}
+
+object TypeDefinitionSet {
+
+  def apply(typeDefinitionSet: Set[TypeInfos.ClazzDefinition] = Set()): TypeDefinitionSet = {
+
+    val clazzDefinitionMap = typeDefinitionSet.map(clazzDefinition => clazzDefinition.clazzName.javaClassName -> clazzDefinition).toMap
+
+    new TypeDefinitionSet(clazzDefinitionMap)
   }
 }
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -8,7 +8,6 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
 
   def validateTypeReference(spelNode: SpelNode) : Boolean = {
 
-    //todo: dodac logike zamieniajaca mi na zbior typow
 
     if(spelNode.getChildCount < 1) {
       //todo: przechwycić błąd w razie jakby nie miało jednego child node'a
@@ -21,18 +20,14 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
       if(typeDefinition.clazzName.display.equals(spelNodeChildAST))
         return true
     })
-    //todo: zamienic na petle ktora sie przerywa jak znajdzie
     false
   }
 
   def displayBasicInfo(spelNode : SpelNode) = {
-    //    uproszczona logika:
-    //      typeDefinitions.contains(spelNode)
-    println("spelNode")
+
+    println("spelNode: ")
     println(spelNode.toStringAST)
     println(spelNode.getChild(0).toStringAST)
-    println
-    println("======================")
     println
     typeDefinitions.foreach(element => {
 
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -1,0 +1,50 @@
+package pl.touk.nussknacker.engine
+
+import org.springframework.expression.spel.SpelNode
+import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+
+
+case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = Set()) {
+
+  def validateTypeReference(spelNode: SpelNode) : Boolean = {
+
+    //todo: dodac logike zamieniajaca mi na zbior typow
+
+    if(spelNode.getChildCount < 1) {
+      //todo: przechwycić błąd w razie jakby nie miało jednego child node'a
+    }
+
+    val spelNodeChildAST = spelNode.getChild(0).toStringAST
+
+    typeDefinitions.foreach(typeDefinition => {
+      println(typeDefinition.clazzName.display)
+      if(typeDefinition.clazzName.display.equals(spelNodeChildAST))
+        return true
+    })
+    //todo: zamienic na petle ktora sie przerywa jak znajdzie
+    false
+  }
+
+  def displayBasicInfo(spelNode : SpelNode) = {
+    //    uproszczona logika:
+    //      typeDefinitions.contains(spelNode)
+    println("spelNode")
+    println(spelNode.toStringAST)
+    println(spelNode.getChild(0).toStringAST)
+    println
+    println("======================")
+    println
+    typeDefinitions.foreach(element => {
+
+      println(element.clazzName.display)
+      println("\tmethods: ")
+      element.methods.foreach(method => {
+        println("\t\t"+method._1)
+        println("\t\t"+method._2)
+        println
+      })
+      println
+    })
+
+  }
+}
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
@@ -16,7 +16,6 @@ case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = S
     val spelNodeChildAST = spelNode.getChild(0).toStringAST
 
     typeDefinitions.foreach(typeDefinition => {
-      println(typeDefinition.clazzName.display)
       if(typeDefinition.clazzName.display.equals(spelNodeChildAST))
         return true
     })
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ExpressionCompiler.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ExpressionCompiler.scala
@@ -42,9 +42,9 @@ object ExpressionCompiler {
                       optimizeCompilation: Boolean, settings: ClassExtractionSettings, typeDefinitionSet: TypeDefinitionSet): ExpressionCompiler = {
     val defaultParsers = Seq(
       SpelExpressionParser.default(loader, dictRegistry, optimizeCompilation, expressionConfig.strictTypeChecking,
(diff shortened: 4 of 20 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -244,8 +245,11 @@ protected trait ProcessCompilerBase {
 
 object ProcessValidator {
 
-  def default(definitions: ProcessDefinition[ObjectWithMethodDef], dictRegistry: DictRegistry, classLoader: ClassLoader = getClass.getClassLoader): ProcessValidator = {
-    val expressionCompiler = ExpressionCompiler.withoutOptimization(classLoader, dictRegistry, definitions.expressionConfig, definitions.settings)
+  def default(definitions: ProcessDefinition[ObjectWithMethodDef], dictRegistry: DictRegistry,
+              classLoader: ClassLoader = getClass.getClassLoader): ProcessValidator = {
+    val typeDefinitionSet = new TypeDefinitionSet(ProcessDefinitionExtractor.extractTypes(definitions))
```
dswiecki, line 250: dla case class mozesz pominac new

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ExpressionCompiler.scala
```diff
@@ -18,28 +18,33 @@ import pl.touk.nussknacker.engine.spel.SpelExpressionParser
...
+      TypeDefinitionSet(modelData.typeDefinitions))
   }
 
   private def default(loader: ClassLoader, dictRegistry: DictRegistry, expressionConfig: ExpressionDefinition[ObjectMetadata],
-                      optimizeCompilation: Boolean, settings: ClassExtractionSettings): ExpressionCompiler = {
+                      optimizeCompilation: Boolean, settings: ClassExtractionSettings, typeDefinitionSet: TypeDefinitionSet): ExpressionCompiler = {
     val defaultParsers = Seq(
-      SpelExpressionParser.default(loader, dictRegistry, optimizeCompilation, expressionConfig.strictTypeChecking, expressionConfig.globalImports, SpelExpressionParser.Standard, expressionConfig.strictMethodsChecking)(settings),
-      SpelExpressionParser.default(loader, dictRegistry, optimizeCompilation, expressionConfig.strictTypeChecking, expressionConfig.globalImports, SpelExpressionParser.Template, expressionConfig.strictMethodsChecking)(settings),
+      SpelExpressionParser.default(loader, dictRegistry, optimizeCompilation, expressionConfig.strictTypeChecking,
```
dswiecki, line 44: imo do przemyslenia czy nie przekazywac calego expressionConfig do SqlExpressionParser.default

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -28,11 +29,14 @@ import pl.touk.nussknacker.engine.types.EspTypeUtils
...
 import scala.annotation.tailrec
 import scala.reflect.runtime._
-import scala.util.{Success, Failure, Try}
+import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
 private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: CommonSupertypeFinder,
-                          dictTyper: SpelDictTyper, strictMethodsChecking: Boolean)(implicit settings: ClassExtractionSettings) extends LazyLogging {
+                          dictTyper: SpelDictTyper, strictMethodsChecking: Boolean,
+                          typeDefinitionSet: TypeDefinitionSet = TypeDefinitionSet(),
```
dswiecki, line 37: imo kolejnosc typeDefinitionSet i referenceTypeValidating powinna byc odwrotna, to pierwsze ma defaulta

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/TyperSpec.scala
```diff
@@ -29,10 +30,11 @@ class TyperSpec extends FunSuite with Matchers {
 
   private val strictTypeChecking = false
   private val strictMethodsChecking = false
+  private val referenceTypeValidating = false
   private val classResolutionStrategy = SupertypeClassResolutionStrategy.Union
   private val commonSupertypeFinder = new CommonSupertypeFinder(classResolutionStrategy, strictTypeChecking)
   private val dict = new SimpleDictRegistry(Map.empty)
-  private val typer = new Typer(this.getClass.getClassLoader, commonSupertypeFinder, new KeysDictTyper(dict), strictMethodsChecking)(ClassExtractionSettings.Default)
+  private val typer = new Typer(this.getClass.getClassLoader, commonSupertypeFinder, new KeysDictTyper(dict), strictMethodsChecking, typeDefinitionSet = TypeDefinitionSet(), referenceTypeValidating)(ClassExtractionSettings.Default)
```
dswiecki, line 37: typeDefinitionSet ma przeciez default

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,44 @@
+package pl.touk.nussknacker.engine
+
+import org.springframework.expression.spel.SpelNode
+import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+
+
+case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = Set()) {
+
+  def validateTypeReference(spelNode: SpelNode) : Boolean = {
```
lciolecki, line 9: It's scala - try to not use return :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,44 @@
...
+
+
+    if(spelNode.getChildCount < 1) {
+      //todo: przechwycić błąd w razie jakby nie miało jednego child node'a
```
lciolecki, line 13: Comments in english.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    val spelNodeChildAST = spelNode.getChild(0).toStringAST
```
lciolecki, line 16: You can do something like this: `spelNode.children..headOption.getOrElse(throw Exception(""))`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    typeDefinitions.foreach(typeDefinition => {
```
lciolecki, line 18: You can use `typeDefinitions.exists()`. After that you shouldn't return true / false.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,44 @@
...
+      if(typeDefinition.clazzName.display.equals(spelNodeChildAST))
+        return true
+    })
+    false
+  }
+
+  def displayBasicInfo(spelNode : SpelNode) = {
```
lciolecki, line 25: Are you sure this method should printing ? I think it should return string.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -251,7 +255,19 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
         case _ => invalid("Invalid ternary operator") // shouldn't happen
       }
       //TODO: what should be here?
```
lciolecki, line 257: Right now you can remove this comment.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -251,7 +255,19 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
-      case e: TypeReference => fixed(Unknown)
+
+      case e: TypeReference => {
+
+        if(!referenceTypeValidating) {
```
lciolecki, line 261: Try to not use return. Look how we do it above in this class.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -251,7 +255,19 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
+          return valid(Unknown)
+        }
+
+        if(typeDefinitionSet.validateTypeReference(e:SpelNode)) {
+          return valid(Unknown)
+        }
+        return invalid("class is not allowed to be passed as TypeReference")
```
lciolecki, line 268: Message should start with upper case ;)

lciolecki (changes requested):
Welcome in our nussknacker world! :) God job and keep it! Please look at my comments.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: I understand but here I don't see how I could avoid it, other then by using a flag which I would return at the end, and that would create a redundancy by searching the whole set every time even after it finds a match.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
mbz-touk: ok, poprawiłem dzięki

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
mbz-touk: fixed

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/TyperSpec.scala
mbz-touk: już zmieniłem, dzięki

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: ok I changed it, thanks, looks mich better

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: changed it, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: changed it, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
lciolecki: Do something like this:
```
(referenceTypeValidating, typeDefinitionSet.validateTypeReference(e)) {
case (false, _) => valid(Unknown)
case (true, true) => valid(Unknown)
_ => invalid("Class is not allowed to be passed as TypeReference")
}
```
After that you shouldn't use return :)

mbz-touk: I think I fixed it. Please take a look again when you find the time if that's what you meant

lciolecki: Ok, It looks better now.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
```diff
@@ -14,7 +14,8 @@ case class ExpressionConfig(globalProcessVariables: Map[String, WithCategories[A
                             //      DictInstance have the same dictId as DictDefinition
                             dictionaries: Map[String, WithCategories[DictDefinition]] = Map.empty,
                             hideMetaVariable: Boolean = false,
-                            strictMethodsChecking: Boolean = true
+                            strictMethodsChecking: Boolean = true,
+                            referenceTypeValidating: Boolean = false
```
arkadius, line 18: I would prefer single  flag for both type refernce and theirs method checking. Also WDYT about having "static" in the name? Now this name is very spel implementation details specific. From user perspecitve it is something like just "staticMethodInvocationsChecking"

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -250,8 +254,14 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
         case _ => invalid("Invalid ternary operator") // shouldn't happen
       }
-      //TODO: what should be here?
-      case e: TypeReference => fixed(Unknown)
+
+      case e: TypeReference => {
+        (referenceTypeValidating, typeDefinitionSet.validateTypeReference(e)) match {
+          case (false, _) => valid(Unknown)
+          case (true, true) => valid(Unknown)
+          case _ => invalid("Class is not allowed to be passed as TypeReference")
```
arkadius, line 262: Can we move this invalid to `validateTypeReference`? `Validated` is better return type for method `validate`. Also validate is more aware about context of validation so is more close to message.

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
```diff
@@ -93,11 +93,51 @@ class ProcessValidatorSpec extends FunSuite with Matchers with Inside {
     ObjectDefinition.noParam,
     ExpressionDefinition(
       Map("processHelper" -> ObjectDefinition(List(), Typed(ProcessHelper.getClass), List("cat1"), SingleNodeConfig.zero)),
-      List.empty, LanguageConfiguration.default, optimizeCompilation = false, strictTypeChecking = true, dictionaries = Map.empty, hideMetaVariable = false, strictMethodsChecking = true
+      List.empty, LanguageConfiguration.default, optimizeCompilation = false, strictTypeChecking = true, dictionaries = Map.empty, hideMetaVariable = false, strictMethodsChecking = true, referenceTypeValidating = true
     ),
     ClassExtractionSettings.Default
   )
 
+  test("TypeReference validation T(MathContext), should succeed") {
```
arkadius, line 101: Why validation for `MathContext` is succeed? Let's give this scenario some more "user perspective" name. The same below

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,35 @@
+package pl.touk.nussknacker.engine
+
+import org.springframework.expression.spel.SpelNode
+import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+import pl.touk.nussknacker.engine.spel.ast.SpelAst.RichSpelNode
+
+
+case class TypeDefinitionSet(typeDefinitions: Set[TypeInfos.ClazzDefinition] = Set()) {
```
arkadius, line 8: Let's add factory method `apply(typeDefinitions: Set[TypeInfos.ClazzDefinition])` but in implementation let's use `Map`...

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def validateTypeReference(spelNode: SpelNode): Boolean = {
+
+    val spelNodeChildAST = spelNode.children.headOption.getOrElse(throw new Exception("SpelNode has no children")).toStringAST
+
+    typeDefinitions.exists(typeDefinition => typeDefinition.clazzName.display.equals(spelNodeChildAST))
```
arkadius, line 14: `clazzName.display` is not a good choice. Display is only for presentation - not for operational usage. Let's add some method like `javaClassName` to `TypedClass`  that will do `ReflectUtils.fixedClassSimpleNameWithoutParentModule(ClassUtils.primitiveToWrapper(klass))`

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
mbz-touk: Changed the flag name, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
mbz-touk: I put some of the logic into validateMethod so it returns Validated but I'm not certain if that's exactly what you meant so I would appreciate it if you took another look at it

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
mbz-touk: Changed the tests and their names so they should be more intuitive now

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: Switched to Map and added the apply method, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mbz-touk: Added the javaClassName method, thanks

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -67,6 +70,7 @@ object typing {
 
     override def display: String = s"Dict(id=$dictId)"
 
+    override def javaClassName: String = ???
```
lciolecki, line 73: Missing implementation? Maybe this method should return: Option[String] and there shuld be just null?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,54 @@
...
+    if(typeDefinitions.contains(spelNodeChildAST)) {
+      Valid(spelNode)
+    } else {
+      Invalid(NonEmptyList.of(ExpressionParseError("Class is not allowed to be passed as TypeReference")))
+    }
+  }
+
+  def displayBasicInfo: String = {
+
+    val newLine = "\n"
```
lciolecki, line 28: I wonder about use: `System.lineSeparator()`? Can you check it? Try use the same thing in tab separator.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -250,8 +254,20 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
           }
         case _ => invalid("Invalid ternary operator") // shouldn't happen
       }
-      //TODO: what should be here?
-      case e: TypeReference => fixed(Unknown)
+
+      case e: TypeReference => {
+        //todo ponizsze ma zadzialac
```
lciolecki, line 259: Comments in English ;)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -250,8 +254,20 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
+        //"T(LocalDateTime).now()"
+        //todo ponizsze nie
+        //"T(LocalDateTime).nowqwe()"
+        print(typeDefinitionSet.displayBasicInfo)
```
lciolecki, line 263: Please remove this print and use logger.debug instead of.

lciolecki (approved):
It's ok for me. But please look at my comments.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
maciej-brzezinski: true, I was going to change it but I wasn't sure how, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
maciej-brzezinski: also changed TypingResult to TypedClass in ClazzDefinition and implemented the method there

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
maciej-brzezinski: Changed the logic there, turns out those implementations weren't necessary I only needed that method in TypedClass

arkadius: Please push the change because I still see those ???

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -94,11 +100,14 @@ object typing {
       case many => many.map(_.display).mkString(" | ")
     }
 
+    override def javaClassName: String = ???
   }
 
   //TODO: make sure parameter list has right size - can be filled with Unknown if needed
   case class TypedClass private[typing] (klass: Class[_], params: List[TypingResult]) extends SingleTypingResult {
 
+    override def javaClassName :String = ReflectUtils.fixedClassSimpleNameWithoutParentModule(ClassUtils.primitiveToWrapper(klass))
```
arkadius, line 109: we can't use simple name here because AFAIK spel expects full name, right? why tests doesn't failed?

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
```diff
@@ -93,11 +93,51 @@ class ProcessValidatorSpec extends FunSuite with Matchers with Inside {
...
       Map("processHelper" -> ObjectDefinition(List(), Typed(ProcessHelper.getClass), List("cat1"), SingleNodeConfig.zero)),
-      List.empty, LanguageConfiguration.default, optimizeCompilation = false, strictTypeChecking = true, dictionaries = Map.empty, hideMetaVariable = false, strictMethodsChecking = true
+      List.empty, LanguageConfiguration.default, optimizeCompilation = false, strictTypeChecking = true, dictionaries = Map.empty, hideMetaVariable = false, strictMethodsChecking = true, staticMethodInvocationsChecking = true
     ),
     ClassExtractionSettings.Default
   )
 
+  test("Invocation of Type Reference of valid class, should succeed") {
+
+    val typeReferenceWithValidClass = "T(MathContext).staticMethod()"
```
arkadius, line 103: 1. What does make `MathContext` available for typing? It is available because of `Decimal`? Can we use some more straightforward type like `String` + write comment that `String` is always available or sth like that?
2. This test is tests invalid case - AFIK spel need full class name (with package). Even if this expression will validate, it will thow exception during evaluation. I think that we need to add more tests in `SpelExpressionSpec` that would check that not only epxression is valid but also would evaluate

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,54 @@
...
+import pl.touk.nussknacker.engine.spel.ast.SpelAst.RichSpelNode
+
+
+class TypeDefinitionSet(typeDefinitions: Map[String, TypeInfos.ClazzDefinition]) {
+
+  def validateTypeReference(spelNode: SpelNode): Validated[NonEmptyList[ExpressionParseError], SpelNode] = {
+
+    val spelNodeChildAST = spelNode.children.headOption.getOrElse(throw new Exception("SpelNode has no children")).toStringAST
+
+    if(typeDefinitions.contains(spelNodeChildAST)) {
```
arkadius, line 19: I've missed one thing - spell need to have full class name or simple class name if you defined global imports for that class. I think that we need to replace `contains` with `find` invoked on `Set` and make checking global imports aware.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
```diff
@@ -0,0 +1,54 @@
...
+import pl.touk.nussknacker.engine.api.expression.ExpressionParseError
+import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
+import pl.touk.nussknacker.engine.definition.{DefinitionExtractor, ProcessDefinitionExtractor, TypeInfos}
+import pl.touk.nussknacker.engine.spel.TypedNode
+import pl.touk.nussknacker.engine.spel.ast.SpelAst.RichSpelNode
+
+
+class TypeDefinitionSet(typeDefinitions: Map[String, TypeInfos.ClazzDefinition]) {
+
+  def validateTypeReference(spelNode: SpelNode): Validated[NonEmptyList[ExpressionParseError], SpelNode] = {
```
mproch, line 15: let's have TypeReference as parameter to this method

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
maciej-brzezinski: You are right, i got a little fixated on that pattern matching there, it's much better this way. Thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mproch: I wonder, maybe it would be easier to just evaluate this node (using getValue) and obtain "real" class?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
mproch: do we need this method at all? Except for debugging/println? :)

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -206,7 +206,9 @@ object SpelExpressionParser extends LazyLogging {
               strictTypeChecking: Boolean,
               imports: List[String],
               flavour: Flavour,
-              strictMethodsChecking: Boolean)
+              strictMethodsChecking: Boolean,
+              staticMethodInvocationsChecking: Boolean,
+              typeDefinitionSet: TypeDefinitionSet = TypeDefinitionSet(Set()))
```
mproch, line 211: I don't think this default helps - in all reasonable cases we should pass explicit list

mproch (changes requested):

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
maciej-brzezinski: Changed it, thanks

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
maciej-brzezinski: Apologies, must have missed it. Should be ok now

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
maciej-brzezinski: I deleted it, thanks

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
maciej-brzezinski: Made some changes, please take another look

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
maciej-brzezinski: changed to String

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
maciej-brzezinski: Implemented it. Thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/TypeDefinitionSet.scala
maciej-brzezinski: Changed it, thanks

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompilerData.scala
```diff
@@ -40,7 +39,8 @@ object ProcessCompilerData {
     val dictRegistryFactory = loadDictRegistry(userCodeClassLoader)
     val dictRegistry = dictRegistryFactory.createEngineDictRegistry(definitions.expressionConfig.dictionaries)
 
-    val expressionCompiler = ExpressionCompiler.withOptimization(userCodeClassLoader, dictRegistry, definitions.expressionConfig, definitions.settings)
+    val typeDefinitionSet = TypeDefinitionSet(ProcessDefinitionExtractor.extractTypes(definitions))
```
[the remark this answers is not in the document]
maciej-brzezinski: I'm not sure I understand. Would need to discuss it with you more in depth

PR #1904, opened 2021-07-12
Rename splitStateToDisk to spillStateToDisk

dswiecki:


ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
```diff
@@ -0,0 +1,34 @@
...
+}
+
+object V1_030__SpillStateToDisk {
+
+  final val emptyAdditionalBranches = Json.fromValues(List.empty)
+
+  private[migration] def renameSpillStateToDisk(jsonProcess: Json): Option[Json] = {
+    val typeSpecificDataCursor = jsonProcess.hcursor.downField("metaData").downField("typeSpecificData")
+    val updatedTypeSpecificData = typeSpecificDataCursor
+      .withFocus(typeSpecificData => {
```
arkadius, line 19: Please use `.withFocus { typeSpecificData => ` (without curly braces). The same below.

ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
```diff
@@ -0,0 +1,34 @@
...
+        val spillStateToDisk = typeSpecificData.hcursor.downField("splitStateToDisk").focus
```
arkadius, line 20: IMO variable should be named `oldSpillStateToDisk` or `splitStateToDisk` because now it looks like you want to extract correct value instead of old one

ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
```diff
@@ -0,0 +1,34 @@
...
+        spillStateToDisk match {
+          case Some(oldValue) =>
+            return typeSpecificDataCursor.withFocus(json => {
+              val jsonWithNewField = json.mapObject(_.add("spillStateToDisk", oldValue))
+              val jsonWithoutOldField = jsonWithNewField.hcursor.downField("splitStateToDisk").delete.top.get
```
arkadius, line 25: Why not to do it in builder:
```
json.mapObject(_.add("spillStateToDisk", oldValue))
  .hcursor.downField("splitStateToDisk").delete.top.get
```
currently it is easy to introduce a bug using in second line other `json` variable then expected

arkadius (approved):
Added some minor comments, LGTM

ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
dswiecki: Done

ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
dswiecki: Done

ui/server/src/main/scala/db/migration/V1_030__SpillStateToDisk.scala
dswiecki: Done

PR #1907, opened 2021-07-13
[FIX] Leaner OpenAPI assembly

mproch:
"optional, provided" does not seem to work....


dswiecki (approved):
Great ;) Now jar does not include engine package

PR #1921, opened 2021-07-19
Rename ProcessManager to DeploymentManager

mproch:


docs/Architecture.md
```diff
@@ -10,7 +10,7 @@ Nussknacker consists of three parts:
 
 
 ##Engine
-Engine consists of various modules that enable creation of processes building blocks in UI and interpretation of process diagrams. Engines implement `ProcessManagerProvider`.
+Engine consists of various modules that enable creation of processes building blocks in UI and interpretation of process diagrams. Engines implement `DeploymentManagerProvider`.
```
arkadius, line 13: We should change this chapter. It should be sth like "scenario deployment". We should write that "Scenario is deployed on engine (like Flink) using DeploymentManager"

build.sbt
```diff
@@ -1139,4 +1139,4 @@ lazy val root = (project in file("."))
   )
 
 addCommandAlias("assemblySamples", ";flinkManagementSample/assembly;standaloneSample/assembly;generic/assembly")
-addCommandAlias("assemblyEngines", ";flinkProcessManager/assembly;engineStandalone/assembly")
+addCommandAlias("assemblyEngines", ";flinkDeploymentManager/assembly;engineStandalone/assembly")
```
arkadius, line 1142: IMO we should also rename task to "assemblyDeploymentManagers" or "assemblyDeployers". Also I think that we should scan for usage of word "engine" in such context. I don't mind that is used in package name or module name but if was used as an alias to DeploymentManager - it should be changed.

arkadius (approved):

PR #1924, opened 2021-07-19
Fix for: cypress used root user in unified mode which caused bad file permissions

arkadius:


dswiecki (approved):
Great, it even caused problem with lack of permission during removing generated file

PR #1927, opened 2021-07-20
Rename outer-join to single-side-join

mproch:


arkadius (approved):

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -165,7 +165,7 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
     "aggregate" -> categories(SlidingAggregateTransformerV2),
     "union" -> categories(UnionTransformer),
     "union-memo" -> categories(UnionWithMemoTransformer),
-    "outer-join" -> categories(OuterJoinTransformer),
```
dswiecki, line 168: Should we support name migration for such nodes ?

PR #1930, opened 2021-07-21
Fix 'counts' functionality for custom nodes. Only incoming events should be counted

gadomsky:
This PR also coherents presenting results of 'tests from file'. New variables in context are preseneted in the following node in graph. Before this was true for all kind of nodes besides custom nodes.

### DIFF
--- a/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
+++ b/engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
@@ -57,9 +57,9 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
 
   private def interpretNode(node: Node, ctx: Context): F[List[InterpretationResult]] = {
     implicit val nodeImplicit: Node = node
-    listeners.foreach(_.nodeEntered(node.id, ctx, metaData))
     node match {
       case Source(_, next) =>
+        listeners.foreach(_.nodeEntered(node.id, ctx, metaData))
         interpretNext(next, ctx)
       case VariableBuilder(_, varName, Right(fields), next) =>
         val variable = createOrUpdateVariable(ctx, varName, fields)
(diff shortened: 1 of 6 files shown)


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -57,9 +57,9 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
 
   private def interpretNode(node: Node, ctx: Context): F[List[InterpretationResult]] = {
     implicit val nodeImplicit: Node = node
-    listeners.foreach(_.nodeEntered(node.id, ctx, metaData))
     node match {
       case Source(_, next) =>
+        listeners.foreach(_.nodeEntered(node.id, ctx, metaData))
```
arkadius, line 62: Doesn't we count twice `nodeEntered` for sources? You have it here and in `interpretNext`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -154,11 +154,13 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
     }
   }
 
-  private def interpretNext(next: Next, ctx: Context): F[List[InterpretationResult]] =
+  private def interpretNext(next: Next, ctx: Context): F[List[InterpretationResult]] = {
+    listeners.foreach(_.nodeEntered(next.id, ctx, metaData))
```
arkadius, line 158: IMO better would be leave notifing `nodeEntered` where it was before and add "additional" notify in PartRef handling with additional comment. WDYT? Also I have doubts if there are cases where we double `nodeEntered` for custom node? Once during `PartRef` handling and second during going though `case CustomNode(...) => `? Do you know it if is really possible or maybe we can have:
```
      case CustomNode(_, next) =>
        throw new IllegalStateException("Custom node should be cut off during process splitting")
```
to make sure that it is situation which will never happen

gadomsky: Please note, that in `intepretNext` we invoke listener for **next**.id 
I have added test which confirms that.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
gadomsky: No, in `interpretNext` I invoke this method for **next** node, not current one

For what I know, it is not possible to have another node before soruce in the graph - only in this case it would be counted twice.

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
gadomsky: @arkadius please see the code now

arkadius (approved):

PR #1933, opened 2021-07-21
Separate names for a BE's authentication provider and a FE's authentication strategy

trombka:
The idea is that a frontend authentication strategy is no longer selected by the name of a backend authentication provider. Now it is the latter to select the proper FE strategy through a settings DTO. Additionally, a DummyAutheticationProvider has been added and an anonymous access configuration for the Basic Auth as well.


ui/client/reducers/settings.ts
```diff
@@ -33,16 +33,23 @@ export type SettingsState = {
 }
 
 export type BaseAuthenticationSettings = {
-  backend?: string
+  provider?: string
+  strategy?: string
 }
 
-export type AuthenticationSettings = BaseAuthenticationSettings | RemoteAuthenticationSettings | OAuth2Settings
+export type AuthenticationSettings = BaseAuthenticationSettings | BrowserAuthenticationSettings | RemoteAuthenticationSettings | OAuth2Settings
```
mproch, line 40: why do we need these BaseAuthenticationSettings here?

trombka: To allow nulls for the provider and strategy fields in the initial state.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
```diff
@@ -11,6 +11,7 @@ import pl.touk.nussknacker.ui.security.api.Permission.Permission
 trait AuthenticationConfiguration {
   def name: String
   def usersFile: URI
+  def anonymousUserRole: Option[String]
```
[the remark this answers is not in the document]
trombka: Removed.

mproch (approved):

PR #1945, opened 2021-07-23
Renaming process to scenario on FE - fixes

mproch:


dswiecki:
Something looks wired regarding branches ? why are there changes in version.sbt ?

ui/client/components/QueriedStateTable.js
```diff
@@ -24,17 +25,19 @@ class QueriedStateTable extends React.Component {
   }
 
   render() {
+    const {t} = useTranslation()
```
JulianWielga, line 28: no.
hooks only in functional components.  https://reactjs.org/docs/hooks-overview.html
you need hoc (withTranslation) or component (Trans) here. https://react.i18next.com/latest/withtranslation-hoc
or rewrite to FC and use hook.

ui/client/components/tips/Errors.js
```diff
@@ -65,20 +68,20 @@ export default class Errors extends React.Component {
         <div className={"node-error-links"}>
           <NodeErrorsLinkSection
             nodeIds={_.concat(otherNodeErrorIds, _.isEmpty(propertiesErrors) ? [] : "properties")}
-            message={"Errors in: "}
+            message={i18next.t("errors.errorsIn", "Errors in: ")}
```
JulianWielga, line 71: better move to FC (and tsx) with hook

JulianWielga (changes requested):

ui/client/components/QueriedStateTable.js
mproch: thanks for catching, used hoc for now (this should be small change to merge into 0.4.0)

ui/client/components/tips/Errors.js
mproch: In next PR, this one should be small to be merged into 0.4.0, and for now we don't have >1 language version anyway...

JulianWielga (approved):

PR #1955, opened 2021-07-23
Ui config loading fix

arkadius:
… config entries inside defaultUiConfig.conf wasn't able to use in base application conf passed via nussknacker.config.locations


mproch (approved):
:+1:  Thanks, it makes loading UI config much clearer now :)

PR #1966, opened 2021-07-27
Disable flag on dynamic property access

apanczyk:


engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
```diff
@@ -180,6 +180,43 @@ class ProcessValidatorSpec extends FunSuite with Matchers with Inside {
...
+
+    val compilationResult = validate(correctProcess, baseDefinitionCopy)
+
+    compilationResult.result should matchPattern {
+      case Valid(_) =>
+    }
+  }
+
+
+  test("Valid dynamic property access when disabled") {
```
mproch, line 204: "Invalid dynamic property access..."?

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
```diff
@@ -180,6 +180,43 @@ class ProcessValidatorSpec extends FunSuite with Matchers with Inside {
...
+    val correctProcess = EspProcessBuilder
+      .id("process1")
+      .exceptionHandler()
+      .source("id1", "source")
+      .filter("filter1", "#input.plainValue == 1")
```
mproch, line 209: What about this filter? Do we want to disable access here or not?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
```diff
@@ -17,7 +17,8 @@ case class ExpressionConfig(globalProcessVariables: Map[String, WithCategories[A
                             hideMetaVariable: Boolean = false,
                             strictMethodsChecking: Boolean = true,
                             staticMethodInvocationsChecking: Boolean = false,
-                            disableMethodExecutionForUnknown: Boolean = false
+                            disableMethodExecutionForUnknown: Boolean = false,
+                            disableDynamicPropertyAccess: Boolean = true
                            )
```
mproch, line 22: I wonder, maybe it's time to introduce some "SpelCompilationFlags/Settings" class? To make passing all those flags around easier?

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/ProcessValidatorSpec.scala
apanczyk: Well, it's not necessary. Already changed that. As first I thought, it will be better to show that problem is with 2nd filter, but you're right.

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
apanczyk: Great idea imo. It starts to look trashy. Should we add this class here, or just create another task/branch?

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -132,9 +133,9 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
       case e: Indexer =>
         current.stack match {
           case TypedClass(clazz, param :: Nil) :: Nil if clazz.isAssignableFrom(classOf[java.util.List[_]]) => valid(param)
-          case TypedClass(clazz, keyParam :: valueParam :: Nil):: Nil if clazz.isAssignableFrom(classOf[java.util.Map[_, _]]) => valid(valueParam)
+          case TypedClass(clazz, keyParam :: valueParam :: Nil) :: Nil if clazz.isAssignableFrom(classOf[java.util.Map[_, _]]) => valid(valueParam)
           case (d: TypedDict) :: Nil => dictTyper.typeDictValue(d, e).map(toResult)
-          case _ => valid(Unknown)
+          case _ => if(disableDynamicPropertyAccess) invalid("Dynamic property access is not allowed") else valid(Unknown)
```
arkadius, line 138: We have multiple subclasses of `TypingResult`. I see that under this default there are: 1) `TypedObjectTypingResult` 2) `TypedTaggedValue` 3)`TypedUnion` and 4) `Unknown`. It looks like `TypedTaggedValue` can be handled like it would be without tag

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/process/ExpressionConfig.scala
arkadius: I would do that in other change. I have other doubt - using negating logic causes that we are fixing ourselves on opt-out logic. I wonder if it would be better to name it `dynamicPropertyAccessAllowed`. This comment is a little bit late, because I see now that the same is in `disableMethodExecutionForUnknown` (it was first precedent)

mproch: :+1: let's leave it for next change

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -102,6 +103,16 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
         throw new SpelCompilationException(node, e)
     }
 
+    def indexerValidator(e: Indexer, stack: List[TypingResultWithContext]): NodeTypingResult = {
```
arkadius, line 106: maybe better name would be `typeIndexer`? like other methods in this class e.g. `typeNode`, `typeExpression`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -102,6 +103,16 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
+      stack match {
+        case TypingResultWithContext(TypedClass(clazz, param :: Nil), _) :: Nil if clazz.isAssignableFrom(classOf[java.util.List[_]]) => valid(param)
+        case TypingResultWithContext(TypedClass(clazz, keyParam :: valueParam :: Nil), _):: Nil if clazz.isAssignableFrom(classOf[java.util.Map[_, _]]) => valid(valueParam)
+        case TypingResultWithContext(d: TypedDict, _) :: Nil => dictTyper.typeDictValue(d, e).map(typ => toResult(typ))
```
arkadius, line 110: It can be just `.map(toResult)`

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
apanczyk: Done

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
apanczyk: Changed

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -102,6 +103,16 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
         throw new SpelCompilationException(node, e)
     }
 
+    def typeIndexer(e: Indexer, stack: List[TypingResultWithContext]): NodeTypingResult = {
+      stack match {
+        case TypingResultWithContext(TypedClass(clazz, param :: Nil), _) :: Nil if clazz.isAssignableFrom(classOf[java.util.List[_]]) => valid(param)
+        case TypingResultWithContext(TypedClass(clazz, keyParam :: valueParam :: Nil), _):: Nil if clazz.isAssignableFrom(classOf[java.util.Map[_, _]]) => valid(valueParam)
+        case TypingResultWithContext(d: TypedDict, _) :: Nil => dictTyper.typeDictValue(d, e).map(toResult)
+        case TypingResultWithContext(TypedTaggedValue(underlying, _), _) :: Nil => typeIndexer(e, TypingResultWithContext(underlying.objType) :: Nil)
+        case _ => if(dynamicPropertyAccessAllowed) valid(Unknown) else invalid("Dynamic property access is not allowed")
```
mproch, line 112: I think we need to handle one more case - Union. This probably will amount to having sequence on all Union types

apanczyk: I just appended my match expression but I'm not sure about my solution. Does it fit your idea?

mproch (approved):
LGTM

arkadius (approved):

PR #1967, opened 2021-07-27
[FIX] Handle initial parameters status in KafkaSinkWithEditor properly

mproch:
Peek 2021-07-28 07-20


dswiecki (approved):
theoretically, we could write test in KafkaAvroSinkFactoryWithEditorIntegrationTest but I see there's no setup for scenario without params

PR #1968, opened 2021-07-27
Handle optional JSON encoding properly

mproch:
Current mechanism works on Flink, but when testing, the results look like this:

Screenshot from 2021-07-29 12-29-50

Also, we need a way to encode classes like `InputMeta` properly


engine/flink/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/source/InputMeta.scala
```diff
@@ -98,4 +102,17 @@ class InputMetaTypeInformationCustomisation extends TypingResultAwareTypeInforma
       InputMeta.typeInformation(originalDetection.forType(a.fields(InputMeta.keyParameterName)))
   }
 
+}
+
+class InputMetaToJson extends ToJsonEncoder {
```
dswiecki, line 107: I miss some InputMetaToJsonSpec with inputMeta encoding test

mproch: I added one

dswiecki (approved):
LGTM

PR #1970, opened 2021-07-27
Fix/migration comments

Filemon279:
Fix for:
- adding comment on empty migration, when no migration was applied `ProcessModelMigrator` proceed with action `UpdateProcessAction` with comment `Migrations applied: `, it was visible in scenario comments section.
- applying the same migration over and over when migration changed nothing. Since we set not to increase version when json's are equal (useful for saving visual changes) migration has been trying to apply migrations on every NK restart. Now every migration increases version.

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
@@ -14,10 +14,11 @@ case class MigrationResult(process: CanonicalProcess, migrationsApplied: List[Pr
 
   def id : String = process.metaData.id
 
-  def toUpdateAction(processId: ProcessId): UpdateProcessAction = UpdateProcessAction(processId,
-    GraphProcess(ProcessMarshaller.toJson(process).noSpaces),
-    s"Migrations applied: ${migrationsApplied.map(_.description).mkString(", ")}",
-    true
+  def toUpdateAction(processId: ProcessId): UpdateProcessAction = UpdateProcessAction(
+    id = processId,
+    deploymentData = GraphProcess(ProcessMarshaller.toJson(process).noSpaces),
+    comment = s"Migrations applied: ${migrationsApplied.map(_.description).mkString(", ")}",
+    increaseVersionWhenJsonNotChanged = true
   )
 
 }
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
@@ -29,27 +29,22 @@ class ProcessModelMigrator(migrations: ProcessingTypeDataProvider[ProcessMigrati
     for {
       migrations <- migrations.forType(processDetails.processingType)
       displayable <- processDetails.json
-      migrated <- migrateWithMigrations(migrations, ProcessConverter.fromDisplayable(displayable), processDetails.modelVersion)
-    } yield migrated
+      migrationsToApply = findMigrationsToApply(migrations, processDetails.modelVersion) if !migrationsToApply.isEmpty
+    } yield migrateWithMigrations(ProcessConverter.fromDisplayable(displayable), migrationsToApply)
   }
 
-  private def migrateWithMigrations(migrations: ProcessMigrations, process: CanonicalProcess, modelVersion: Option[Int])
-    : Option[MigrationResult] = {
-
-    val migrationsToApply = migrations.processMigrations.toList.sortBy(_._1).dropWhile {
+  private def findMigrationsToApply(migrations: ProcessMigrations, modelVersion: Option[Int]): List[ProcessMigration] = {
+    migrations.processMigrations.toList.sortBy(_._1).dropWhile {
       case (migrationNumber, _) => migrationNumber <= modelVersion.getOrElse(0)
     }.map(_._2)
-
-    if(migrationsToApply.isEmpty) None
-    else {
-      val resultProcess = migrationsToApply.foldLeft(process) {
-        case (processToConvert, migration) => migration.migrateProcess(processToConvert)
-      }
-      Some(MigrationResult(resultProcess, migrationsToApply))
-    }
-
   }
 
+  private def migrateWithMigrations(process: CanonicalProcess, migrationsToApply: List[ProcessMigration]): MigrationResult = {
+    val resultProcess = migrationsToApply.foldLeft(process) {
+      case (processToConvert, migration) => migration.migrateProcess(processToConvert)
+    }
+    MigrationResult(resultProcess, migrationsToApply)
+  }
 
 }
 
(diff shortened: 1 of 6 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
```diff
@@ -16,7 +16,8 @@ case class MigrationResult(process: CanonicalProcess, migrationsApplied: List[Pr
 
   def toUpdateAction(processId: ProcessId): UpdateProcessAction = UpdateProcessAction(processId,
     GraphProcess(ProcessMarshaller.toJson(process).noSpaces),
-    s"Migrations applied: ${migrationsApplied.map(_.description).mkString(", ")}"
+    s"Migrations applied: ${migrationsApplied.map(_.description).mkString(", ")}",
+    true
```
arkadius, line 20: Can u use named parameter here and in other cases of usage to make easier visible what this argument do without navigating to method declaration? BTW Intellij Idea highlighting such case as a code style issue

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -106,7 +106,7 @@ class DBProcessRepository(val dbConfig: DbConfig, val modelVersion: ProcessingTy
     }
   }
 
-  private def updateProcessInternal(processId: ProcessId, processDeploymentData: ProcessDeploymentData)(implicit loggedUser: LoggedUser): DB[XError[ProcessUpdated]] = {
+  private def updateProcessInternal(processId: ProcessId, processDeploymentData: ProcessDeploymentData, forceIncreaseVersion: Boolean)(implicit loggedUser: LoggedUser): DB[XError[ProcessUpdated]] = {
```
arkadius, line 109: It doesn't look like this argument cause that version increase is always forced. It looks rather like `increaseVersionWhenJsonChanged`?

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigratorSpec.scala
```diff
@@ -34,6 +34,16 @@ class ProcessModelMigratorSpec extends FlatSpec with BeforeAndAfterEach with Pat
     processor shouldBe ServiceRef(ProcessTestData.otherExistingServiceId, List())
   }
 
+  it should "migration should not return empty migration result" in {
+
+    val migrationResultOpt: Option[MigrationResult] = migrateByVersionsOpt(Some(1), 1)
+
+    migrationResultOpt.map(o => println(o.toUpdateAction(ProcessId(1L))))
```
arkadius, line 41: Please don't use system out. It will be hard to reduce numer of such logs by simple changes in logback configuration. Can it be printed using logger?

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigratorSpec.scala
```diff
@@ -34,6 +34,16 @@ class ProcessModelMigratorSpec extends FlatSpec with BeforeAndAfterEach with Pat
...
+
+    migrationResultOpt shouldNot be('defined)
```
arkadius, line 43: Easier would be `shouldBe empty`

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepositorySpec.scala
```diff
@@ -189,24 +189,46 @@ class DBFetchingProcessRepositorySpec
...
+      .source("s", "")
+      .emptySink("s2", "")
+
+    saveProcess(espProcess, now)
+
+    val latestProcessVersion = fetchLatestProcessVersion(processName)
+    latestProcessVersion.id shouldBe 1
+    updateProcess(latestProcessVersion.copy(json = someJson), false).newVersion.get.id shouldBe 2
+    //without force
+    updateProcess(latestProcessVersion.copy(json = someJson), false).newVersion shouldNot be('defined)
```
arkadius, line 216: The same as above

arkadius (approved):
Good job, some minor comments added. Maybe co-author of bug that u've repaired can take a look also?

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigratorSpec.scala
Filemon279: Hah, it should be removed - I missed it while cleaning after debugging. Deleted.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
Filemon279: Maybe `increaseVersionWhenJsonNotChanged` ?

arkadius: Ah, u've right - should be negation there! LGTM.

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigratorSpec.scala
Filemon279: Changed.
Personally I like using this format for ex. `be('defined)` since It implies that I'm dealing with optionals at the first glance.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
Filemon279: You are right, I've added named parameters for readability.

Filemon279:
Thanks @arkadius for good points - I have changed some lines according to your comments

ui/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigratorSpec.scala
arkadius: you've right but you have non necessary negation, but maybe it is subjective thing

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
```diff
@@ -27,20 +29,25 @@ class ProcessModelMigrator(migrations: ProcessingTypeDataProvider[ProcessMigrati
...
-    : MigrationResult = {
+    : Option[MigrationResult] = {
 
     val migrationsToApply = migrations.processMigrations.toList.sortBy(_._1).dropWhile {
       case (migrationNumber, _) => migrationNumber <= modelVersion.getOrElse(0)
     }.map(_._2)
 
-    val resultProcess = migrationsToApply.foldLeft(process) {
-      case (processToConvert, migration) => migration.migrateProcess(processToConvert)
+    if(migrationsToApply.isEmpty) None
```
mproch, line 43: It would look a little bit nicer if we splitted this method to:
- 'findMigrationsToApply' - returning list/option[list]
- `migrateWithMigrations`
But it's minor thing, so feel free to ignore this comment

Filemon279: So nicer it is :)

PR #1971, opened 2021-07-27
Fix for: state verification mechanism didn't work with rocksdb

arkadius:


nussknacker-dist/src/universal/conf/dev-application.conf
```diff
@@ -28,6 +28,9 @@ flinkDeploymentConfig {
   restUrl: ${?FLINK_REST_URL}
   queryableStateProxyUrl: "taskmanager:9069"
   queryableStateProxyUrl: ${?FLINK_QUERYABLE_STATE_PROXY_URL}
+  # It is disabled by default for easier dev env setup - see ui/README.md for details
+  shouldVerifyBeforeDeploy: false
```
mproch, line 32: I wonder, maybe it should be default here if we provide dev setup for `false` anyway?

mproch (approved):

PR #1974, opened 2021-07-28
Fix for: Node dropped on edge wasn't horizontally aligned

arkadius:
Peek 2021-07-28 16-27

Until this: https://github.com/TouK/nussknacker/pull/1976 won't be done, behavior could looks strange for case when node is dropped on the free space in graph


ui/client/components/graph/ProcessGraph.tsx
```diff
@@ -21,7 +21,9 @@ const spec = {
   drop: (props, monitor, component: Graph) => {
     const clientOffset = monitor.getClientOffset()
     const relOffset = component.processGraphPaper.clientToLocalPoint(clientOffset)
-    component.addNode(monitor.getItem(), relOffset)
+    // to make node vertically aligned
```
mproch, line 24: vertically or horizontally?? ;)

mproch (approved):

PR #1980, opened 2021-07-29
Remove subprocessVersions - from FE

mproch:


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -41,20 +41,18 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
...
-                  UIProcessObjectsFactory.prepareUIProcessObjects(
-                    processingTypeData.modelData,
-                    processingTypeData.deploymentManager,
-                    user,
-                    subprocesses,
-                    isSubprocess,
-                    processCategoryService)
-                )
-              }
+          get {
```
dswiecki, line 44: where is change post->get in UI client ? or is it transparent in that part ?

mproch: It's in `HttpService.ts`

dswiecki: Somehow I missed this file

dswiecki (approved):

PR #1995, opened 2021-08-02
Docker installation guide improvements

arkadius:


build.sbt
```diff
@@ -299,6 +299,7 @@ lazy val distDockerSettings = {
     dockerEntrypoint := Seq(s"$workingDir/bin/nussknacker-entrypoint.sh", dockerPort.toString),
     dockerExposedPorts := Seq(dockerPort),
     dockerEnvVars := Map(
+      // TODO: move to defaultUiConfig.conf as a one place for defining defaults
```
mproch, line 302: :+1: good point

docs/installation_configuration_guide/Installation.md
```diff
@@ -9,7 +9,74 @@ documentation for details on how to configure them for optimal performance .
...
+
+### Base Image
+
+As a base image we uses `openjdk:11-jdk`. See [Open JDK's Docker hub](https://hub.docker.com/_/openjdk) for more details.
+
+### Properties
+
+Available Nussknacker image properties:
+
+| Property name  | Type     | Default value | Description
```
mproch, line 22: Are these properties of docker image, or of default configuration? Some of these are docker-specific, but most are from defaultUiConfig - maybe we can keep those separately/distinguished as they are the same for e.g. tgz distribution?

mproch (approved):

PR #2001, opened 2021-08-03
Configurable standard tabs

mproch:
- GlobalPermission can be an arbitrary string
- All top tabs can be defined now


ui/server/src/test/scala/pl/touk/nussknacker/ui/api/UsersResourcesSpec.scala
```diff
@@ -16,23 +17,21 @@ class UsersResourcesSpec extends FunSuite with ScalatestRouteTest with FailFastC
     getUser(isAdmin = false) ~> check {
       status shouldBe StatusCodes.OK
       responseAs[Json] shouldBe Json.obj(
-        "id" -> Json.fromString("1"),
-        "username" -> Json.fromString("user"),
+        "id" -> fromString("1"),
+        "username" -> fromString("user"),
         "isAdmin" -> Json.fromBoolean(false),
```
dswiecki, line 22: Lets be consistent about impots

dswiecki:
It's ok for me however probably in the future we should somehow restrict user without AdminTab permission e.g., Writer from typing explicit path and opening /admin tab

dswiecki (approved):

PR #2005, opened 2021-08-04
Drop Scala 2.11 and JDK8 - initial

mproch:


build.sbt
```diff
@@ -13,15 +13,14 @@ import scala.util.Try
...
 // Warning: Flink doesn't work correctly with 2.12.11
 // Warning: 2.12.13 + crossVersion break sbt-scoverage: https://github.com/scoverage/sbt-scoverage/issues/319
 val scala212 = "2.12.10"
-lazy val supportedScalaVersions = List(scala212, scala211)
+lazy val supportedScalaVersions = List(scala212)
 
 // Silencer must be compatible with exact scala version - see compatibility matrix: https://search.maven.org/search?q=silencer-plugin
 // Silencer 1.7.x require Scala 2.12.11 (see warning above)
-// Silencer (and all '@silent' annotations) can be removed after we drop support for Scala 2.11
+// Silencer (and all '@silent' annotations) can be removed after we can upgrade to 2.12.13...
```
dswiecki, line 23: previous comment said it can be done after we drop 2.11 ;) can't we do it for sure ?

dswiecki (approved):

build.sbt
mproch: Well... it turns out to be not so easy...

PR #2009, opened 2021-08-06
using openapi service from dev setup

arkadius:


nussknacker-dist/src/universal/conf/dev-application.conf
```diff
@@ -35,16 +35,16 @@ scenarioTypes {
   "streaming": {
     deploymentConfig: ${flinkDeploymentConfig}
     modelConfig: {
-      classPath: [ ${managementModelDir}"/managementSample.jar" ]
+      classPath: [ ${genericModelDir}"/genericModel.jar" ]
```
dswiecki, line 38: but here we miss openapi.jar (and sql.jar) and its not trivial to setup this path

arkadius: It is quite trivial ;) - I've added the change after which it is available after dev run. Should be merged with: https://github.com/TouK/nussknacker-quickstart/pull/4 Take a look

arkadius: After talk with Maciek, I've copied `nussknacker.conf` content to `dev-application.conf` to make dev setup lest related to quickstart and avoid some additional code on quickstart side. We need to fix one thing in model config resolution yet, but I'll do this in separate PR

mproch (approved):

PR #2025, opened 2021-08-09
Fix url for adminApp & processes when serving NK on not empty path

Filemon279:


arkadius (approved):
@JulianWielga said that it is ok, so let's merge it

JulianWielga (approved):

PR #2036, opened 2021-08-13
Pipeline GH Action for microbenchmarks

apanczyk:


apanczyk:
Completed benchmark with SampleSpelBenchmark: https://github.com/apanczyk/nussknacker/actions/runs/1126883805
Failed benchmark scheduled at 00.00 UTC (already fixed): https://github.com/apanczyk/nussknacker/actions/runs/1118308151

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,24 @@
+name: Benchmarks
+on:
+  schedule:
+#   Cron uses UTC - now this time is 00:00, during winter it won't be
```
mproch, line 4: let's remove this comment, we don't really care about exact hour...

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: "openjdk@1.11"
+      - name: Run benchmarks
+        run: sbt "benchmarks/jmh:run -i 8 -wi 3 -f1 -t2 .SampleSpelBenchmark -rf csv"
```
mproch, line 19: I wonder, maybe it's better to use just one thread? (-t1) 
And let's check how long will it take for all benchmarks to run?

apanczyk: At first glance there are minor/no differences between -t1 or -t2. With same parameters full benchmarks should be completed about 40 minutes https://github.com/apanczyk/nussknacker/actions/runs/1127313049.

dswiecki:
LGTM!

dswiecki (approved):

PR #2038, opened 2021-08-13
[FIX] Disable possibility to add double edge - on FE and BE

mproch:


dswiecki:
@mproch Unfortunately I'am not able to add second edge to any node - first works at glance but second one not

ui/client/components/graph/NodeUtils.ts
```diff
@@ -200,13 +200,15 @@ class NodeUtils {
```
dswiecki, line 200: @mproch Unfortunately I'am not able to add second edge to any node - first works at glance but second one not

dswiecki: Fixed

dswiecki (approved):

PR #2039, opened 2021-08-13
glossary related changes to the main Configuration document + sections reordering

ZbyszekMM:
I suggest moving the HoCon section to the end of this document. When I want to read about configuration, I want to have an overview of what awaits me first ( 3 areas to configure), rather than about configuration language

I use the convention that  names of "parts" of NK (cannot say _component_  here - Designer, Executor, Deployment Manager start with a capital letter. Thia, in this context I should have written  _model_ starting with lower case - because it is not a "part". I'll change to lowercase after a review.

I  adopted the convention that there are no links to other documents from the Installation and Configuration area in this document. If they stayed, it would not be clear whether you would be taken to the Glossary or a relevant document.

Finally, I did not add a Scenario Type glossary term  yet. This is a non-critical concept for now, and I don't think we have a good definition in the <internal> Glossary.


docs/installation_configuration_guide/Configuration.md
```diff
@@ -1,40 +1,38 @@
...
-
 ## Configuration areas
 
 Nussknacker configuration is divided into several configuration areas, each area addressing a specific aspect of using Nussknacker:
 
-* Designer configuration
-* Scenario types configuration
-    * Deployment manager/executor configuration
-    * Model configuration
+* [Designer](https://docs.nussknacker.io/about/GLOSSARY#nussknacker-designer) configuration
```
dswiecki, line 7: it should relative link not absolute

dswiecki: here and in another places

mproch (approved):

PR #2042, opened 2021-08-13
redirectUri made optional in the OAuth2 configuration

trombka:
… in favor of dynamic evaluation in the FE

The redirect URI is unnecessary in configuration. When the frontend needs to call the authorization endpoint it already knows its own location, so there is no need to take this value from the backend settings. Though I left such an option for a yet-unknown reason.

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2AuthenticationResources.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2AuthenticationResources.scala
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.ui.security.oauth2
 
 import akka.http.scaladsl.marshalling.ToResponseMarshallable
+import akka.http.scaladsl.model.StatusCodes.NotFound
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, HttpResponse, StatusCodes}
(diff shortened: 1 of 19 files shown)


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2AuthenticationResources.scala
```diff
@@ -36,17 +37,28 @@ class OAuth2AuthenticationResources(realm: String, service: OAuth2Service[Authen
 
   override lazy val additionalRoute: Route =
     pathEnd {
-      parameters('code) { authorizeToken =>
+      parameters('code,  'redirect_uri.?) { (authorizationCode, redirectUri) =>
         get {
-          complete {
-            oAuth2Authenticate(authorizeToken)
+          xor(redirectUri, configuration.redirectUri.map(_.toString)).map { redirectUri =>
```
arkadius, line 42: You can just do `redirectUri.orElse(configuration.redirectUri.map(_.toString))`

arkadius (commented):
LGTM, just please small thing in code

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2AuthenticationResources.scala
trombka: It is supposed to be an eXclusive OR.

arkadius (approved):

PR #2046, opened 2021-08-16
Extra functions in generic model

dswiecki:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/functions/math.scala
```diff
@@ -0,0 +1,62 @@
+package pl.touk.nussknacker.engine.util.functions
+
+import pl.touk.nussknacker.engine.api.{Documentation, ParamName}
+
+object math {
+
+  @Documentation(description = "Returns the greater of two {@code double} values.")
+  def maxDouble(@ParamName("a") a: Double, @ParamName("b") b: Double): Double =
```
mproch, line 8: Let's make one function with Number as argument and deal with conversions in implementation?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/functions/math.scala
```diff
@@ -0,0 +1,62 @@
...
+  @Documentation(description = "Returns the smaller of two {@code int} values.")
+  def minFloat(@ParamName("a") a: Float, @ParamName("b") b: Float): Float =
+    Math.min(a, b)
+
+  @Documentation(description = "Returns the value of the first argument raised to the power of the second argument")
+  def pow(@ParamName("a") a: Double, @ParamName("b") b: Double): Double =
+    Math.pow(a, b)
+
+  @Documentation(description = "Returns the absolute value of a {@code int} value.")
+  def absInt(@ParamName("a") a: Long): Long =
```
mproch, line 44: Let's make one function with Number as argument and deal with conversions in implementation?

mproch (approved):

PR #2068, opened 2021-08-17
Permit non-absolute URI config values

trombka:


engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/URIExt.scala
```diff
@@ -0,0 +1,13 @@
+package pl.touk.nussknacker.engine.util.config
+
+import java.io.File
+import java.net.URI
+
+class URIExt(uri: URI) {
+  def withFileSchemeDefault: URI = if (uri.isAbsolute) uri else new File(uri.getSchemeSpecificPart).toURI
+}
+
+object URIExt {
```
arkadius, line 10: You can just use `implicit class` instead: https://docs.scala-lang.org/overviews/core/implicit-classes.html

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/URIExt.scala
```diff
@@ -0,0 +1,13 @@
+package pl.touk.nussknacker.engine.util.config
+
+import java.io.File
+import java.net.URI
+
+class URIExt(uri: URI) {
+  def withFileSchemeDefault: URI = if (uri.isAbsolute) uri else new File(uri.getSchemeSpecificPart).toURI
```
arkadius, line 7: Can you add a test for that? What case will work other way after this change? Is it the case when `new URI(value)` throws an exception?

trombka: Before:
Any URI in a config class would default to a file if no scheme were specified.
After:
The standard Ficus reader is used for a URI in a config class. If you want a default scheme, use the extension method.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/URIExt.scala
trombka: Changed.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/config/URIExt.scala
trombka: I have added a test.

arkadius (approved):

Issue #1435, opened 2021-02-25
Simplify OAuth2 configuration


trombka:
Currently, the OAuth2 configuration is quite complicated and requires much knowledge of the standard. 

In a few places, the properties seem to be duplicated e.g we need to explicitly set `authorizeParams.response_type = code` even though we set `accessTokenParams.grant_type = authorization_code`.

When talking about OAuth2 nowadays most people have in mind not only the primary specification but lots of flavors and de facto standards based on it. In most cases, an authorization server handles OIDC probably with the Discovery. But even if not, it is for instance very uncommon to use URLs other than `.../authorize` and `.../token`.

I think this:
```
authentication: {
  method: "oauth2"
  issuer: "http://authorization.server.url"
  clientId: "XXXXXXXXXXXX"
}
```
is all the information required to work with a server handling OIDC + Discovery + PKCE.
But let's drop the PKCE for a moment and stay with the `clientSecret` property.

So describing the enhancement in one sentence:
The `issuer`, `clientId`, and `clientSecret` properties _must_ be sufficient to work with an OIDC Discovery compliant server and probably _should_ be sufficient in most other cases.

PR #2070, opened 2021-08-18
A separate authentication provider for OIDC inheriting from the standard OAuth2 one.

trombka:
Solves: #1435

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -103,6 +103,7 @@ def modelMergeStrategy: String => MergeStrategy = {
 }
 
 def uiMergeStrategy: String => MergeStrategy = {
+  case PathList(ps@_*) if ps.last == "module-info.class" => MergeStrategy.discard
   case PathList(ps@_*) if ps.last == "NumberUtils.class" => MergeStrategy.first //TODO: shade Spring EL?
   case PathList("org", "apache", "commons", "logging", _ @ _*) => MergeStrategy.first //TODO: shade Spring EL?
   case PathList(ps@_*) if ps.last == "io.netty.versions.properties" => MergeStrategy.first //Netty has buildTime here, which is different for different modules :/
@@ -217,7 +218,12 @@ lazy val commonSettings =
         "io.netty" % "netty-handler" % nettyV,
         "io.netty" % "netty-codec" % nettyV,
         "io.netty" % "netty-transport-native-epoll" % nettyV,
-      )
+
+        // Jackson is used by openapi and jwks-rsa
+        "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonV,
+        "com.fasterxml.jackson.core" % "jackson-core" % jacksonV,
+        "com.fasterxml.jackson.core" % "jackson-databind" % jacksonV,
+       )
     )
 
 val akkaV = "2.5.21" //same version as in Flink
@@ -235,7 +241,7 @@ val argonautV = "6.2.1"
 val circeV = "0.11.2"
 val circeJava8V = "0.11.1"
 val jwtCirceV = "4.0.0"
-val jacksonV = "2.9.2"
+val jacksonV = "2.11.3"
 val catsV = "1.5.0"
 val scalaParsersV = "1.0.4"
 val slf4jV = "1.7.30"
@@ -835,6 +841,7 @@ lazy val security = (project in engine("security")).
         "io.circe" %% "circe-core" % circeV,
         "com.pauldijou" %% "jwt-circe" % jwtCirceV,
         "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV,
+        "com.auth0" % "jwks-rsa" % "0.19.0", // a tool library for reading a remote JWK store, not an Auth0 service dependency
         "com.dimafeng" %% "testcontainers-scala-scalatest" % testcontainersScalaV % "it,test",
         "com.github.dasniko" % "testcontainers-keycloak" % "1.6.0" % "it,test"
       )
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
@@ -1,0 +1,77 @@
+package pl.touk.nussknacker.ui.security.oidc
+
+import com.auth0.jwk.{JwkProvider, JwkProviderBuilder}
+import pl.touk.nussknacker.engine.util.config.URIExtensions
+import pl.touk.nussknacker.ui.security.oauth2.{JwtConfiguration, OAuth2Configuration}
+import pl.touk.nussknacker.ui.security.oauth2.ProfileFormat.OIDC
+import sttp.client.{NothingT, SttpBackend}
+import sttp.model.MediaType
+
+import java.net.URI
+import java.security.PublicKey
+import java.util.NoSuchElementException
+import scala.concurrent.{ExecutionContext, Future}
+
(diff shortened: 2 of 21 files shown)


build.sbt
```diff
@@ -235,7 +241,7 @@ val argonautV = "6.2.1"
 val circeV = "0.11.2"
 val circeJava8V = "0.11.1"
 val jwtCirceV = "4.0.0"
-val jacksonV = "2.9.2"
+val jacksonV = "2.11.3"
```
dswiecki, line 244: there's already 2.11.4 but why not use 2.12.x ?

build.sbt
```diff
@@ -217,7 +218,12 @@ lazy val commonSettings =
         "io.netty" % "netty-handler" % nettyV,
         "io.netty" % "netty-codec" % nettyV,
         "io.netty" % "netty-transport-native-epoll" % nettyV,
-      )
+
+        // Jackson is used by openapi and jwks-rsa
```
dswiecki, line 222: i'm not sure if these libraries should be in common, can we be more precise so that we avoid adding them to unnecessary modules ?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationProvider.scala
```diff
@@ -0,0 +1,25 @@
...
+  import net.ceedubs.ficus.readers.ArbitraryTypeReader._
+  import pl.touk.nussknacker.engine.util.config.CustomFicusInstances._
+  import pl.touk.nussknacker.ui.security.api.AuthenticationConfiguration.authenticationConfigPath
+
+  def name: String = "Oidc"
+
+  override def createAuthenticationResources(config: Config, classLoader: ClassLoader)(implicit ec: ExecutionContext, sttpBackend: SttpBackend[Future, Nothing, NothingT]): OAuth2AuthenticationResources = {
+    new OidcAuthenticationResources(
+      realm,
+      config.as[OidcAuthenticationConfiguration](authenticationConfigPath).withDiscovery
```
dswiecki, line 22: do I understand correctly that right now we'll try to discover config based on issuer even though authorizationEndpoint is provided ? do we want it to work that way ?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcService.scala
```diff
@@ -0,0 +1,13 @@
+package pl.touk.nussknacker.ui.security.oidc
+
+import pl.touk.nussknacker.ui.security.oauth2.{DefaultJwtAccessToken, DefaultOpenIdConnectAuthorizationData, JwtValidator, OAuth2ClientApi, OpenIdConnectService, OpenIdConnectUserInfo}
+import sttp.client.{NothingT, SttpBackend}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class OidcService(configuration: OidcAuthenticationConfiguration)
```
dswiecki, line 8: what's the difference in terms of intended use between OidcService and OpenIdConnectService ?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
```diff
@@ -0,0 +1,77 @@
...
+                                           // from the OIDC Discovery or in case it is not supported at all.
+                                           // They may be relative to the issuer.
+                                           authorizationEndpoint: Option[URI] = None,
+                                           tokenEndpoint: Option[URI] = None,
+                                           userinfoEndpoint: Option[URI] = None,
+                                           jwksUri: Option[URI] = None
+                                          ) extends URIExtensions {
+
+  lazy val oAuth2Configuration: OAuth2Configuration = OAuth2Configuration(
+    usersFile = this.usersFile,
```
dswiecki, line 35: probably some of this `this.` references are redundant

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcService.scala
trombka: The OidcService is concrete - with case classes specified for JWT parsing.
The OpenIdConnectService is generic - you can pass your own type for JWT parsing. Maybe the name could be better here.

My aim generally is to have a generic implementation in the oauth2 package (in case someone needs to write pseudo-authentication with OAuth2 with own requirements) and a specific one in the oidc package

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationProvider.scala
trombka: Yes.
OpenID Discovery provides configuration for any endpoint required in the flow. We cannot skip querying if only one endpoint is provided in the local configuration to override the remote one. And check if all of them are provided seems to be a very improbable case.

build.sbt
trombka: But this is the depenedencyOverrides field. We want it here to have the same version in any submodule when imported there.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
trombka: Though, I thought it would be more descriptive in this quite complex transformation. Shall I remove it?

build.sbt
trombka: 2.11.3 was the largest of all the transitive dependencies. The library is not used in the NK code.

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcService.scala
mproch: Let's change name to sth more meaningful... Or with a comment...

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcDiscovery.scala
```diff
@@ -0,0 +1,45 @@
...
+import sttp.client.{NothingT, SttpBackend, UriContext, basicRequest}
+import sttp.model.MediaType
+
+import java.net.URI
+import scala.concurrent.duration.{Duration, SECONDS}
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+object OidcDiscovery extends LazyLogging {
+
+  implicit val uriDecoder: Decoder[URI] = Decoder.decodeString.map(URI.create)
```
mproch, line 17: maybe we can put it e.g. in CirceUtil.codecs?

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcDiscovery.scala
```diff
@@ -0,0 +1,45 @@
...
+  def apply(issuer: URI)(implicit ec: ExecutionContext, sttpBackend: SttpBackend[Future, Nothing, NothingT]): Option[OidcDiscovery] =
+    Await.result(
+      basicRequest
+        .contentType(MediaType.ApplicationJson)
+        .get(uri"$issuer/.well-known/openid-configuration")
+        .response(asJson[OidcDiscovery]).send(),
+      Duration(30, SECONDS)
+    ).body match {
+      case Right(v) => Some(v)
+      case Left(err) =>
```
mproch, line 29: Maybe we can handle other errors (e.g. timeout, unresolved etc.) in the same way?

mproch (approved):

PR #2093, opened 2021-08-26
Appending roles from the roles claim in the ID token.

trombka:
This adds a configuration field to enable user roles from the claims provided by an OIDC issuer in an ID token or userinfo. Since there is no standard claim for the purpose, a claim name must be provided probably in collision-resistant form like URI.

Example:
```
authentication {
  method: "Oidc"
  ...
  rolesClaim: "urn:namespace:roles"
  ...
}
```

### DIFF
--- a/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
+++ b/engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
@@ -1,7 +1,10 @@
 package pl.touk.nussknacker.ui.security.oauth2
 
+import io.circe.Decoder
+
 import java.time.LocalDate
 import io.circe.generic.extras.{Configuration, ConfiguredJsonCodec, JsonKey}
+import io.circe.generic.extras.semiauto.deriveDecoder
 import io.circe.java8.time.{JavaTimeDecoders, JavaTimeEncoders}
 import pl.touk.nussknacker.ui.security.api.AuthenticatedUser
 import pl.touk.nussknacker.ui.security.oauth2.OAuth2Profile.getUserRoles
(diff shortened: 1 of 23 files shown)


engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
```diff
@@ -40,19 +43,35 @@ import scala.concurrent.duration.Deadline
...
   val jwtId: Option[String] = None
   val notBefore: Option[Deadline] = None
 }
 
 object OpenIdConnectUserInfo extends EpochSecondsCodecs with JavaTimeDecoders with JavaTimeEncoders with EitherCodecs {
-  implicit val config: Configuration = Configuration.default
+  implicit val config: Configuration = Configuration.default.withDefaults
+
+  lazy val decoder: Decoder[OpenIdConnectUserInfo] = deriveDecoder[OpenIdConnectUserInfo]
+  def decoderWithCustomRolesClaim(rolesClaim: String): Decoder[OpenIdConnectUserInfo] =
```
mproch, line 60: cool :)

engine/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationResources.scala
```diff
@@ -6,7 +6,7 @@ import sttp.client.{NothingT, SttpBackend}
 import scala.concurrent.{ExecutionContext, Future}
 
 
-class OidcAuthenticationResources(realm: String, configuration: OidcAuthenticationConfiguration)(implicit ec: ExecutionContext, sttpBackend: SttpBackend[Future, Nothing, NothingT])
+class OidcAuthenticationResources(override val name: String, realm: String, configuration: OidcAuthenticationConfiguration)(implicit ec: ExecutionContext, sttpBackend: SttpBackend[Future, Nothing, NothingT])
```
mproch, line 9: I wonder, maybe it would be better to just pass it as constructor parameter? Now it's not quite clear for me, looking at the code, when and how this should be overridden

trombka: Hmmm... I don't get it. `override` does not mean that a value or method is to be overridden, but that it overrides a value or method from a superclass.

mproch (approved):

PR #2095, opened 2021-08-30
Remove BusinessView

mproch:


dswiecki (approved):
I've sent you two places to remove businessView in PM

PR #2098, opened 2021-08-31
Correct timestamps in test from file for Kafka sources

### LINKED ISSUE #2074: Tumbling window aggregation raises RuntimeException when default KafkaSource
**Describe the bug**
EDIT: Problem occurs for Test From File. Have not registered this error when a job is running and consuming real events.

A combination of basic KafkaSource with default timestampAssigner and with `StreamRecord.hasTimestamp == false` (e.g. staging -> kafka-json) and CustomTransformer with tumbling window aggregation (e.g. staging -> aggregate-tumbling) raises exception 
`java.lang.RuntimeException: Record has Long.MIN_VALUE timestamp (= no timestamp marker). Is the time characteristic set to 'ProcessingTime', or did you forget to call 'DataStream.assignTimestampsAndWatermarks(...)'?`

EDIT: problem occurs also for avro sources.

KafkaSource uses deprecated `BoundedOutOfOrderPreviousElementAssigner` when no other assigner is provided. Here when hasTimestamp == false the default prev timestamp is `Long.MIN_VALUE` (see also parameters of `TimestampsAndWatermarksOperator.processElement`). When `TumblingEventTimeWindows.assignWindows` is executed it checks if timestamp is equal to `Long.MIN_VALUE` and if so it raises the exception mentioned above.

**To Reproduce**
1.Configure scenario with components as described above. 

2.Run test from file.

Reproduced on staging.nussknacker.io

**Environment (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - Nussknacker version 0.4.0
 - Browser type and version FF v92, Chrome v92

mproch:
Closes #2074

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -756,7 +756,7 @@ lazy val flinkTestUtil = (project in engine("flink/test-util")).
         "org.apache.flink" % "flink-metrics-dropwizard" % flinkV
       )
     }
-  ).dependsOn(testUtil, queryableState, flinkUtil)
+  ).dependsOn(testUtil, queryableState, flinkUtil, interpreter)
 
 lazy val standaloneUtil = (project in engine("standalone/util")).
   settings(commonSettings).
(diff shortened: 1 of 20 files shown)


gskrobisz (approved):
LGTM

PR #2105, opened 2021-08-31
Handling Flink's job initializing state as a DURING_DEPLOY instead of…

arkadius:
… RUNNING


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -86,7 +86,9 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, mainClassName:
   //NOTE: Flink <1.10 compatibility - protected to make it easier to work with Flink 1.9, JobStatus changed package, so we use String in case class
   protected def mapJobStatus(overview: JobOverview): StateStatus = {
     toJobStatus(overview) match {
-      case JobStatus.RUNNING => FlinkStateStatus.Running
+      case JobStatus.RUNNING if overview.tasks.running == overview.tasks.total => FlinkStateStatus.Running
```
mproch, line 89: I think we should count also finished as running? Also, for batch jobs there may be quirks around scheduling tasks lazily

arkadius: I wonder what is the case for streaming that some task is finished? But "finished" also doesn't look like "during deploy" so maybe it is a good way to handle that - changed.
Regards to second one I've extracted protected methods which can be overridden in the same way as it was before change. Is it ok for now or sth need to be changed now in `PeriodicDeploymentManager`?

engine/flink/management/src/test/scala/pl/touk/nussknacker/engine/management/FlinkRestManagerSpec.scala
```diff
@@ -275,8 +275,8 @@ class FlinkRestManagerSpec extends FunSuite with Matchers with PatientScalaFutur
     val version = 15L
     val user = "user1"
 
-    statuses = List(JobOverview(jid, processName.value, 40L, 10L, JobStatus.FINISHED.name()),
-      JobOverview("1111", "p1", 35L, 30L, JobStatus.FINISHED.name()))
+    statuses = List(JobOverview(jid, processName.value, 40L, 10L, JobStatus.FINISHED.name(), baseTasksOverview().copy(finished = 1)),
```
mproch, line 278: maybe let's add more parameters (with defaults) to baseTasksOverview?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -95,11 +96,19 @@ class FlinkRestManager(config: FlinkConfig, modelData: ModelData, mainClassName:
...
-      case JobStatus.INITIALIZING => FlinkStateStatus.DuringDeploy
     }
 
   }
 
+  protected def ensureTasksRunning(overview: JobOverview): Boolean = {
+    overview.tasks.running + overview.tasks.finished == overview.tasks.total
+  }
+
+  protected def isDuringDeploy(s: JobStatus): Boolean = {
```
mproch, line 110: canBeDuringDeploy? now it also returns true if job is running stably

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingDeploymentManagerSlotsCountSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+      logger.debug(s"Checking if: $processId, have expected state status: $expectedStatus, current job status: $jobStatus, attempt: $attempt")
+
+      jobStatus.map(_.status.name) shouldBe Some(expectedStatus.name)
+      if (attempt != attempts - 1) {
+        Thread.sleep(checkInterval.toMillis)
+      }
+    }
+  }
+
+  private def cancel(processId: String): Unit = {
```
mproch, line 54: maybe we can extract it to StreamingDockerTest?

engine/flink/management/src/test/scala/pl/touk/nussknacker/engine/management/FlinkRestManagerSpec.scala
arkadius: changed

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
arkadius: changed

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingDeploymentManagerSlotsCountSpec.scala
arkadius: extracted

mproch (approved):

PR #2108, opened 2021-09-02
Enhance default ClassExtractionSettings

mproch:


engine/api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterEditor.scala
```diff
@@ -63,6 +63,10 @@ object PeriodParameterEditor {
   }
 }
 
+/* To use this editor you have to:
```
arkadius, line 66: Thanks :+1:

engine/flink/test-util/src/main/scala/pl/touk/nussknacker/engine/flink/test/ClassExtractionBaseTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    val parsed =  parse(IOUtils.toString(getClass.getResourceAsStream(outputResource))).right.get
+    val decoded = decode(parsed)
+
+    //println(types.asJson.spaces2)
+    checkGeneratedClasses(types, decoded)
+  }
+
+  //We don't do 'types shouldBe decoded' to avoid unreadable messages
+  private def checkGeneratedClasses(types: Set[ClazzDefinition], decoded: Set[ClazzDefinition]): Unit = {
+    types.size shouldBe decoded.size
```
arkadius, line 33: IMO we shouldn't fail so fast. Consider case when one new class is added. It would be nice to see full details of added class. Maybe in case of some differences would be better to always print:
- recently added classes
- recently removed classes
- recently modified methods inside each modified class
WDYT?

mproch: :+1: , for methods I think 
`methods.keys shouldBe decodedMethods.keys` is sufficient

arkadius (approved):
Good job

PR #2110, opened 2021-09-02
Remove groups

mproch:


ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
```diff
@@ -154,15 +154,10 @@ class ProcessValidation(validators: ProcessingTypeDataProvider[ProcessValidator]
...
 
   private def validateDuplicates(displayable: DisplayableProcess): ValidationResult = {
-    val groupIds = displayable.metaData.additionalFields
-      .flatMap(a => a.cast[ProcessAdditionalFields])
-      .toList
-      .flatMap(_.groups)
-      .map(_.id)
     val nodeIds = displayable.nodes.map(_.id)
 
     //in theory it would be possible to have group named like one of nodes inside, but it's not worth complicating logic...
```
dswiecki, line 159: hmm it this comment still valid ? even more, can we have duplicates right now ?

ui/client/reducers/graph/reducer.ts
```diff
@@ -322,7 +283,7 @@ const reducer = mergeReducers(
     processToDisplay: {
       nodes,
       properties: {
-        additionalFields: {groups},
+        additionalFields: {},
```
dswiecki, line 286: is explicit setting additionalFields still required ?

dswiecki (approved):

ui/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
mproch: :+1: comment is not valid, but we can have duplicates defined in FE

ui/client/reducers/graph/reducer.ts
mproch: I'd leave it for next refactor....

PR #2133, opened 2021-09-06
Hide SQL Variable from generic model

mproch:


engine/flink/generic/src/main/resources/defaultModelConfig.conf
```diff
@@ -55,5 +55,12 @@
     previousValue: {
       docsUrl: "https://nussknacker.io/designingProcesses/FlinkCustomTransformers.html#previousvalue"
     }
+    #SQL variable is not documented and ready for wide usage, so we hide it from UI
+    sqlVariable {
+      category: hidden
```
arkadius, line 60: How someone will be able to unhide it? By overriding category with "base"? Can you write it down in comment? Also maybe we should add a note in migrationguide?

mproch: :+1: I added comment & entry in guide

arkadius (approved):

PR #2150, opened 2021-09-07
Handle multiple one time schedules in batch periodic engine

jedrz:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -194,15 +201,31 @@ class PeriodicProcessService(delegateDeploymentManager: DeploymentManager,
...
 
   /**
-    * Returns latest deployment. It can be in any status (consult [[PeriodicProcessDeploymentStatus]]).
-    *
-    * For multiple schedules only the oldest one is returned.
-    */
+   * Returns latest deployment. It can be in any status (consult [[PeriodicProcessDeploymentStatus]]).
+   * For multiple schedules only the effective one is returned.
+   */
   def getLatestDeployment(processName: ProcessName): Future[Option[PeriodicProcessDeployment]] = {
```
mproch, line 207: Is it enough to handle e.g. 'deactivate' method properly?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -194,15 +201,31 @@ class PeriodicProcessService(delegateDeploymentManager: DeploymentManager,
   }
 
   /**
-    * Returns latest deployment. It can be in any status (consult [[PeriodicProcessDeploymentStatus]]).
-    *
-    * For multiple schedules only the oldest one is returned.
-    */
+   * Returns latest deployment. It can be in any status (consult [[PeriodicProcessDeploymentStatus]]).
+   * For multiple schedules only the effective one is returned.
```
mproch, line 205: can we write what 'effective' means? :)

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -194,15 +201,31 @@ class PeriodicProcessService(delegateDeploymentManager: DeploymentManager,
...
   def getLatestDeployment(processName: ProcessName): Future[Option[PeriodicProcessDeployment]] = {
     implicit val localDateOrdering: Ordering[LocalDateTime] = Ordering.by(identity[ChronoLocalDateTime[_]])
-    //TODO: so far we return only one
+
     scheduledProcessesRepository.getLatestDeploymentForEachSchedule(processName)
-      .map(_.sortBy(_.runAt).headOption).run
+      .map(_.sortBy(_.runAt)).run
+      .map {
+        case Nil => None
+        case singleDeployment :: Nil => Some(singleDeployment)
```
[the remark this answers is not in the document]
jedrz: You're right. I think the pattern matching code can be removed at all.

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
jedrz: I think it should be fine. In deactivate method, latest deployment is passed to listener and in case of a failed job, it could be passed to the listener instead of newer deployment in current behaviour... Maybe we should pass to listeners all latest deployments for each schedule.

mproch (approved):

PR #2165, opened 2021-09-10
Drop legacy metrics mode

mproch:


lciolecki:
You have forgotten about changes in documentation?

arkadius:
> You have forgotten about changes in documentation?

I see that it was undocumented feature so it looks like the answer is NO

arkadius (approved):

lciolecki:
Ok, you're right. There is some information in MigrationGuide..

PR #2176, opened 2021-09-13
Allow to enrich process config on initial schedule and each deployment

jedrz:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/service/ProcessConfigEnricher.scala
```diff
@@ -0,0 +1,68 @@
...
+  trait ProcessConfigEnricherInputData {
+    def processJson: String
+    def modelConfigJson: String
+
+    def canonicalProcess: CanonicalProcess = {
+      ProcessMarshaller.fromJson(processJson).valueOr(err => throw new IllegalArgumentException(err.msg))
+    }
+
+    def modelConfig: Config = {
+      ConfigFactory.parseString(modelConfigJson)
```
mproch, line 35: Should it be parse or load? Where/how do we resolve substitutions/defaults?

jedrz: It's rather parse. This config is sent to Flink and I think it should be only extended since factory (`ProcessConfigEnricherFactory`) has access to configuration of deployment manager. What do you think? Anyway, I think I should rename it to `InputConfigDuringExecution` here and in other classes.

mproch (approved):

PR #2180, opened 2021-09-13
Generate openApi for standalone processes (without response)

Filemon279:
- Add `openApiDefinition` method for source which describe request body schema for given process. 
- Interpreter extended with method to collect request/response schema and produce openApi for process.
- Generate full openApi for given pair (ProcessName, Interpreter).


engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+  def generateOpenApi(pathWithInterpreter: List[(String, StandaloneProcessInterpreter)], oApiInfo: OApiInfo, serverDescription: OApiServer): String = {
+    val scenarioDefinitions: Map[String, Json] = pathWithInterpreter
+      .flatMap(a => a._2.produceOpenApiDefinition().map(oApi => a._1 -> oApi))
+      .map {
+        case (path, interpreter) => "/" + path -> interpreter
+      }.toMap
+    OApiDocumentation(OPEN_API_VERSION, oApiInfo, List(serverDescription), scenarioDefinitions.asJson).asJson.spaces2
+  }
+
+  private def preetyPrintMap(m: Map[String, String]): String = m.map(v => s"<b>${v._1}</b>: ${v._2}").mkString(" <br/> ")
```
mproch, line 40: Why don't we just use markdown as per: https://swagger.io/specification/#rich-text-formatting?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypedMapUtils.scala
```diff
@@ -0,0 +1,30 @@
+package pl.touk.nussknacker.engine.standalone.utils.typed
+
+import io.circe.{Json, JsonObject}
+import pl.touk.nussknacker.engine.api.typed.TypedMap
+
+import scala.collection.JavaConverters._
+
+object TypedMapUtils {
+
+    def jsonToTypedMap(json: Json): TypedMap =
```
mproch, line 10: Where do we need/use this?

Filemon279: I am working on tests to demonstrate how and when we need to use it. It is going to be a source which expect JSON in request body (schema defined in source field). This utils helps to change `Json` to `TypedMap`

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
Filemon279: Good idea. I will change it that way

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
```diff
@@ -0,0 +1,76 @@
+package pl.touk.nussknacker.engine.standalone.openapi
+
+import io.circe.Json
+import io.circe.syntax._
+import io.circe.generic.auto._
+import pl.touk.nussknacker.engine.standalone.StandaloneProcessInterpreter
+import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
+
+object StandaloneOpenApiGenerator {
```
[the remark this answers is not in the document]
Filemon279: We agreed not to use external library since model we produce is very simple and it seems that it is easier to manage/change it manually. I am open for discussion/suggestions if any

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
```diff
@@ -0,0 +1,33 @@
+package pl.touk.nussknacker.engine.standalone.utils.typed
+
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.Json
+import io.circe.syntax._
+import pl.touk.nussknacker.engine.api.typed.typing.{TypedClass, TypedObjectTypingResult, TypingResult}
+
+class TypingResultToJsonSchemaConverter(exampleValues: Map[String, Json]) extends LazyLogging {
+
+  def typingResultToJsonSchema(typingResult: TypingResult, paramName: Option[String] = None): Map[String, Any] = typingResult match {
```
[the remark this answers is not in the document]
Filemon279: Result is `Map[String, Any]` and I parsed it to `Json` one level further. I've check it and I think we could do parsing to Json here as you suggested.

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
Filemon279: Done

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypedMapUtils.scala
Filemon279: I've added `JsonSchemaStandaloneSourceFactory` which uses those utils

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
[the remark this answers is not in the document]
Filemon279: I have updated it to return `Json`

build.sbt
```diff
@@ -785,7 +786,8 @@ lazy val standaloneUtil = (project in engine("standalone/util")).
 lazy val standaloneApi = (project in engine("standalone/api")).
   settings(commonSettings).
   settings(
-    name := "nussknacker-standalone-api"
+    name := "nussknacker-standalone-api",
+    libraryDependencies += "com.github.erosb" % "everit-json-schema" % everitSchemaV
```
dswiecki, line 790: is it used ?

Filemon279: Yes - `JsonSchemaStandaloneSourceFactory`

Filemon279: It is used to compose schema validator and validate json from request

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+      tags,
+      processName,
+      preetyPrintMap(description),
+      requestDefinition,
+      responseDefinition
+    )
+    jsonEncoder.encode(postOpenApiDefinition)
+  }
+
+  def generateOpenApi(pathWithInterpreter: List[(String, StandaloneProcessInterpreter)], oApiInfo: OApiInfo, serverDescription: OApiServer): String = {
```
arkadius, line 31: 1. Can u split this method into one that returns documentation and the second one that render it as a json? In case if someone would like to change this definition before returning it or would like to verify sth in tests
2. Do you have plans to use it in `ProcessRoute` in some next PR or maybe at least let's write a TODO that we should use it there to make this feature available without need of changing docker image of this standalone-app

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+      override def responseEncoder: Option[ResponseEncoder[TypedMap]] = Option(new ResponseEncoder[TypedMap] {
+        override def toJsonResponse(input: TypedMap, result: List[Any]): GenericResultType[Json] = {
+          val jsonResult = result.map(jsonEncoder.encode)
+            .headOption
+            .getOrElse(throw new IllegalArgumentException(s"Process did not return any result"))
+          Right(jsonResult)
+        }
+      })
+
+      private def validateAndReturnTypedMap(parameters: String): TypedMap = {
```
arkadius, line 60: It would be nice to have easy ability to use this class with some slight modifications that won't require copy-pasting the whole code. For example someone would like to turn on `PrimitiveValidationStrategy.LENIENT` mode. Can you make this method protected and/or extract some protected methods like `prepareValidator`, `jsonToTypedMap` to make it possible?

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  private def generatePostOApiDefinition(tags: List[String], processName: String, description: String, requestSchema: Json, responseSchema: Json) = Map(
+    "post" -> Map(
+      "tags" -> tags,
+      "summary" -> processName,
+      "description" -> description,
+      "consumes" -> List("application/json"),
+      "produces" -> List("application/json"),
+      "requestBody" -> generateOApiRequestBody(requestSchema),
+      "responses" -> generateOApiResponse(responseSchema) //TODO
```
arkadius, line 72: Shouldn't TODO be in `produceOpenApiDefinition` ? It looks like you have everything prepared correctly at this point. Only in upper level there is some work to do. Also can u add some comment about what u think that should be done to make this feature available?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypedMapUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+      jsonObjectToTypedMap(json.asObject.getOrElse(JsonObject.empty))
+
+    private def jsonToAny(json: Json): Any = {
+      json.fold(
+        jsonNull = null,
+        jsonBoolean = identity,
+        jsonNumber = d => BigDecimal(d.toDouble), //prevent scientific notation
```
arkadius, line 17: Can u use `java.math.BigDecimal` instead? We have very limited support for scala types (if we have any?). In the other hand, spel have very nice support for standard java types - for example it has in flight conversions between bigdecimal and integer and many others

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypedMapUtils.scala
mproch: maybe let's move it to util? so that we can replace 
`pl.touk.nussknacker.engine.kafka.generic.source.jsonToMapn` with it in the future?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
```diff
@@ -0,0 +1,39 @@
+package pl.touk.nussknacker.engine.standalone.utils.typed
+
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.Json
+import io.circe.syntax._
+import pl.touk.nussknacker.engine.api.typed.typing.{TypedClass, TypedObjectTypingResult, TypingResult}
+import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
+
+class TypingResultToJsonSchemaConverter(exampleValues: Map[String, Json]) extends LazyLogging {
```
arkadius, line 9: 1) Where this class is used? 2) I don't see usage of logger

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
```diff
@@ -0,0 +1,39 @@
...
+        Map("type" -> "boolean".asJson)
+      case TypedClass(klass, _) if klass == classOf[java.lang.Integer] =>
+        Map("type" -> "integer".asJson)
+      case TypedClass(klass, _) if klass == classOf[java.lang.Number] =>
+        Map("type" -> "number".asJson)
+      case TypedClass(klass, params) if klass == classOf[java.util.List[_]] =>
+        Map("type" -> "array".asJson, "items" -> params.flatMap(p => translateTypeResult(p)).toMap.asJson)
+      case TypedClass(klass, params) if klass == classOf[java.util.Map[_, _]] =>
+        Map("type" -> "object".asJson, "additionalProperties" -> params.flatMap(p => translateTypeResult(p)).toMap.asJson)
+      case _ => Map()
```
mproch, line 35: What about e.g. Union, Dicts, etc? Maybe we should pattern match on SingleTypingResult?

build.sbt
mproch: Wouldn't it be easier to just use swagger-model?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
```diff
@@ -0,0 +1,39 @@
...
+    }) ++ example.fold(Map.empty[String, Json])(e => Map("example" -> e.asJson))
```
arkadius, line 36: Can you replace `fold` with `map().getOrElse()`? We rather avoid folding of option in our code. It is quite hard to imagine how option is folded for newcomers. Take a look at this blog post: https://nurkiewicz.com/2014/06/optionfold-considered-unreadable.html

engine/standalone/app/src/test/scala/pl/touk/nussknacker/engine/standalone/http/TestConfigCreator.scala
```diff
@@ -19,7 +19,8 @@ class TestConfigCreator extends EmptyProcessConfigCreator {
   override def sourceFactories(processObjectDependencies: ProcessObjectDependencies): Map[String, WithCategories[SourceFactory[_]]] = Map(
     "request1-post-source" -> WithCategories(new JsonStandaloneSourceFactory[Request]),
     "request1-get-source" -> WithCategories(RequestGetSourceFactory),
-    "genericGetSource" -> WithCategories(new TypedMapStandaloneSourceFactory)
+    "genericGetSource" -> WithCategories(new TypedMapStandaloneSourceFactory),
```
mproch, line 22: Can we remove/merge it with JsonSchemaStandaloneSourceFactory? It's more or less the same, but for GET

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  private val jsonEncoder = BestEffortJsonEncoder(failOnUnkown = true, getClass.getClassLoader)
+
+  private[standalone] def generateScenarioDefinition(processName: String,
+                                                     requestDefinition: Json,
+                                                     responseDefinition: Json,
+                                                     description: Map[String, String],
+                                                     tags: List[String]
+                                                    ): Json = {
+   val postOpenApiDefinition = generatePostOApiDefinition(
```
mproch, line 21: Why do we handle only POST?
If we don't want to implement it, we should at least return None for GET etc.?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/SchemaTypingResult.scala
```diff
@@ -0,0 +1,80 @@
...
+      (key, value) <- jobject.toVector
+    } yield (key, value)
+  }.toMap
+
+  def getKeyNames(json: Json): List[String] = for {
+    jobject <- json.asObject.toList
+    k <- jobject.keys
+  } yield k
+
+  def resolveJsonTypingResult(json: Json): TypingResult = {
```
arkadius, line 58: WDYT about keeping code that prepares definition and implementaion (`TypedMapUtils`) in one class to avoid desynchronization of this code? Take a look at `InputMeta` for example. Also would be nice to use some important keyword in the name of this class. Like `Json` , `TypedMap` maybe just `JsonToTypedMapConverter`? `TypedMapUtils` doesn't saying too much. I know that someone can take a look at package name but I'm a big fun of using find by name IDE feature.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+        val jsonObject = new JSONObject(parameters)
+        schema.validate(jsonObject)
```
mproch, line 62: maybe we can use https://github.com/circe/circe-json-schema?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+import pl.touk.nussknacker.engine.api.typed.{TypedMap, _}
+import pl.touk.nussknacker.engine.api.{CirceUtil, MethodToInvoke, ParamName}
+import pl.touk.nussknacker.engine.standalone.api.types.GenericResultType
+import pl.touk.nussknacker.engine.standalone.api.{ResponseEncoder, StandalonePostSource, StandaloneSourceFactory}
+import pl.touk.nussknacker.engine.standalone.utils.typed.{SchemaTypingResult, TypedMapUtils}
+import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
+
+import java.nio.charset.StandardCharsets
+
+class JsonSchemaStandaloneSourceFactory extends StandaloneSourceFactory[TypedMap] {
```
mproch, line 19: Shouldn't it override openApiDefinition?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
```diff
@@ -0,0 +1,33 @@
...
+        Map("type" -> "string".asJson)
+      case TypedClass(klass, _) if klass == classOf[java.lang.Boolean] =>
+        Map("type" -> "boolean".asJson)
+      case TypedClass(klass, _) if klass == classOf[java.lang.Integer] =>
+        Map("type" -> "integer".asJson)
+      case TypedClass(klass, _) if klass == classOf[java.lang.Number] =>
+        Map("type" -> "number".asJson)
+      case TypedClass(klass, params) if klass == classOf[java.util.List[_]] =>
+        Map("type" -> "array".asJson, "items" -> params.flatMap(p => translateTypeResult(p)).toMap.asJson)
+      case TypedClass(klass, params) if klass == classOf[java.util.Map[_, _]] =>
```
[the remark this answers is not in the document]
Filemon279: I have decided to remove this class since we are going towards sources defined by schema and had no use in this MR.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypedMapUtils.scala
Filemon279: I have moved (and renamed) this file to utils

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
Filemon279: 1. Good idea, please find updated class
2. I didn't planned to do it in this PR - though I have added comment as you suggested

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
Filemon279: Thanks, that is great idea. I have added way to define your own schema with `prepareSchema`, validator with `validator` and way to parse json to typeMap with `jsonToTypeMap`.

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
Filemon279: You're right, moved. I've also added comment with my thoughts about openApi for sinks (above `generateOpenApiDefinition`)

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
Filemon279: As I wrote to mproch I changed my conception a little bit and decided to remove this class. As you observed correctly it had no use in this PR.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
Filemon279: Replaced. That is true, it does not come naturally to read/understand it quickly.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/TypingResultToJsonSchemaConverter.scala
Filemon279: As mentioned earlier I have removed this file from PR.

engine/standalone/app/src/test/scala/pl/touk/nussknacker/engine/standalone/http/TestConfigCreator.scala
Filemon279: I do not see point in merging it with `jsonSchemaSource` tests since it not provides any extra value. I do not want to remove it either since it is the only test for generic GET requests. It is not complex so I suggest to keep it.

engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/openapi/StandaloneOpenApiGenerator.scala
Filemon279: You are right, I have added a slot to put GET definition in the future but marked it with TODO for now.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
Filemon279: We have much more flexibility with `everit` itself (circe-json-schema uses `everit` under the hood). What circe-json-schema really does for us is change `circe.Json` to `JSONObject` and call everit validator.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
Filemon279: Yes it should! Thanks. Updated and I added a test for that.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/SchemaTypingResult.scala
Filemon279: Perhaps that is good idea, for now on I just updated `TypedMapUtils` to more meaningful `JsonToTypedMapConverter` as you suggested.

build.sbt
Filemon279: You mean to replace `everit` with `swagger-model`?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
arkadius: Thanks, it is almost ok, but I think that we need to make `StandalonePostSource`  top level class (can be in the same file). Currently, it is anonymous class so nobody can extend it and override those methods.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/typed/SchemaTypingResult.scala
arkadius: It looks that it is not a good "in the future", but it is needed now - you already made those two places of code desynchronized changing scala BigDecimal to java BigDeicimal. Take a look at `SchemaTypingResult`

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
```diff
@@ -0,0 +1,29 @@
...
+object JsonToTypedMapConverter {
+
+  def jsonToTypedMap(json: Json): TypedMap =
+    jsonObjectToTypedMap(json.asObject.getOrElse(JsonObject.empty))
+
+  private def jsonToAny(json: Json): Any = {
+    json.fold(
+      jsonNull = null,
+      jsonBoolean = identity,
+      jsonNumber = d => java.math.BigDecimal.valueOf(d.toDouble), //prevent scientific notation
```
arkadius, line 16: I've added a test that will show that this implementation loose precision. Because of this bug: https://github.com/circe/circe/issues/1101 it will fail even if you change it, but after upgrade to: https://github.com/TouK/nussknacker/pull/2208/files it will work correctly. IMO we should handle this that way: `d.toBigDecimal.map(_.bigDecimal).getOrElse(d.toDouble)`. I'm not sure about `getOrElse` part - it handles the case when decimal overflows `BigDecimal`

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
Filemon279: Ah, I missed that. Moved to top level class.

arkadius (approved):

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  private def jsonValuesToTypingResult(namedSchema: Map[String, Schema]): TypingResult = {
+    TypedObjectTypingResult(namedSchema.mapValues(resolveJsonTypingResult).toList)
+  }
+
+  def resolveJsonTypingResult(schema: Schema): TypingResult = {
+    schema match {
+      case s:ArraySchema => getArrayTypingResult(s)
+      case s:ObjectSchema => parseObjectSchema(s)
+      case s:NumberSchema => if(s.requiresInteger()) Typed.typedClass[Integer] else Typed.typedClass[java.math.BigDecimal]
```
mproch, line 68: I think it integer in json schema should be rather Long for us?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
```diff
@@ -0,0 +1,86 @@
...
+      case _:BooleanSchema => Typed.typedClass[Boolean]
+      case _:TrueSchema => Typed.typedClass[Boolean]
+      case _:FalseSchema => Typed.typedClass[Boolean]
+      case _:NullSchema => Typed.typedClass[Null]
```
mproch, line 72: Do we handle Null nicely?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def jsonToTypedMap(json: Json): TypedMap =
+    jsonObjectToTypedMap(json.asObject.getOrElse(JsonObject.empty))
+
+  private def jsonToAny(json: Json): Any = {
+    json.fold(
+      jsonNull = null,
+      jsonBoolean = identity,
+      jsonNumber = d => {
+        //circe 0.14.1 still looses precision as was in https://github.com/circe/circe/issues/1101.
```
mproch, line 21: Does it? The issue seems resolved for quite a long time?

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
```diff
@@ -0,0 +1,91 @@
...
+  private val schema: Schema = prepareSchema(rawSchema)
+
+  override def parse(parameters: Array[Byte]): TypedMap = {
+    val parametersString = new String(parameters, StandardCharsets.UTF_8)
+    validateAndReturnTypedMap(parametersString)
+  }
+
+  override def openApiDefinition: Option[OpenApiSourceDefinition] = {
+    val json = decodeJsonWithError(schemaStr)
+    val properties = metaData.additionalFields.map(_.properties).getOrElse(Map.empty)
```
mproch, line 54: are we sure all properties should be in external definition?

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
Filemon279: In NK we are using circe `0.14.1` and still `JsonToTypedMapConverterTest` fails if we use `d.toBigDecimal`.
You can observe it by looking at the output for `.toBigDecimal` in this tests.
I was sure it was fixed but sadly I observe different behavior.

engine/standalone/util/src/main/scala/pl/touk/nussknacker/engine/standalone/utils/JsonSchemaStandaloneSourceFactory.scala
Filemon279: True, I moved part of generating description to `protected val openApiDescription` which you can override if you will.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
Filemon279: Yes, we can work with this `null` in NK scenarios nicely.

engine/util/src/main/scala/pl/touk/nussknacker/engine/util/typing/JsonToTypedMapConverter.scala
Filemon279: :+1:

PR #2183, opened 2021-09-13
Fix KeyedRecordFlatMapper performance:

nadberezny:
1. Do not use reflection, get typing from AvroSinkRecordValue


nadberezny:
https://github.com/TouK/nussknacker/pull/2185

PR #2191, opened 2021-09-14
Fix KeyedRecordFlatMapper performance:

nadberezny:
1. Do not map LazyParameter on each event.
2. Do not use refletcion; get typing from SinkValue instead.


arkadius (approved):
LGTM

PR #2221, opened 2021-09-20
array class handling in spel parser

### LINKED ISSUE #970: Handle arrays more properly
Currently we have method and fields returing Array filtered out thanks to this rule:
```
      // Arrays are not supported for now
      ClassMemberPredicate(ClassPredicate { case _ => true }, {
        case m: Method => m.getReturnType.isArray
        case f: Field => f.getType.isArray
      }),
```
it cause that e.g. `String.split` is not available.

Commenting this code out make it available and usable again because we have such a logic in `Typer`:
`          case _ => valid(Unknown)`

Proposed changes:
- remove this rule from method exclude list
- make `Indexer`, `Projection` and `Selection` works with arrays (SpEL evaluation AFAIK works)
- return error in `Indexer` when is used in place where it shouldn't be used
- make suggester completion show human friendly type instead of `[Ljava.lang.String;`

maciej-brzezinski:
closes #970


engine/flink/test-util/src/main/scala/pl/touk/nussknacker/engine/flink/test/ClassExtractionBaseTest.scala
```diff
@@ -28,7 +28,7 @@ trait ClassExtractionBaseTest extends FunSuite with Matchers {
 
     //printFoundClasses(types)
     //println(types.asJson.spaces2)
-    checkGeneratedClasses(types, decoded)
+    //checkGeneratedClasses(types, decoded)
```
maciej-brzezinski, line 31: until we update allowed classes and methods snapshot for including arrays

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/util/ReflectUtils.scala
```diff
@@ -4,7 +4,9 @@ object ReflectUtils {
 
   def fixedClassSimpleNameWithoutParentModule(clazz: Class[_]): String = {
     // fix for https://issues.scala-lang.org/browse/SI-8110
-    clazz.getName
+
+    val className = if(clazz.isArray) clazz.getCanonicalName else clazz.getName
+    className
```
maciej-brzezinski, line 9: could we change all to canonical name? from what i know it would change $ sign to dot in nested classes and only would not work with anonymous inner classes

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/types/EspTypeUtilsSpec.scala
```diff
@@ -262,7 +262,8 @@ class EspTypeUtilsSpec extends FunSuite with Matchers with OptionValues {
 
   test("should hide some ugly presented methods") {
     val classDef = singleClassDefinition[ClassWithWeirdTypesInMethods]().value
-    classDef.methods.get("methodReturningArray") shouldBe empty
+    classDef.methods.get("methodReturningArray") shouldBe
```
dswiecki, line 265: probably this check should be done in another test - this one checks hidden methods

maciej-brzezinski: I understand, thought just removing the test would be confusing so I modified it but I get that it doesn't really fit in here

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/util/ReflectUtils.scala
dswiecki: could we just use getSimpleName ?, it look as if this issue mentioned above is long gone, unfortunately we don't have unit tests for this util method

maciej-brzezinski: if it's as you say and if it's enough information for us in case of e.g. nested classes than I think we could 
I can add the tests

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -129,6 +129,9 @@ object typing {
       throw new IllegalArgumentException("Cannot have typed class of Any, use Unknown")
     } else if (klass.isPrimitive) {
       TypedClass(ClassUtils.primitiveToWrapper(klass), parameters)
+    } else if (klass.isArray) {
+      //to not have separate class for each array, we pass Array of Objects
+      TypedClass(classOf[Array[Object]], List(TypedClass(ClassUtils.primitiveToWrapper(klass.getComponentType), List())))
```
mproch, line 134: Typed.typedClass(klass.getComponentType) => to handle nested arrays ;)

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/util/ReflectUtils.scala
arkadius: Yep, after upgrade to scala 2.12 we can remove this helper. How `TypedClass` that is array of Strings will be present now? Am I right that it will be presented as a `Object[][String]` IMO it should be rather `Array[String]`. Other option is to present as a `String[]` but I think that sticking with Java/C notation is not a benefit for us - let's make this types more human readable. Can u write a test for that?

engine/flink/test-util/src/main/scala/pl/touk/nussknacker/engine/flink/test/ClassExtractionBaseTest.scala
arkadius: so it is still a draft?

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -125,13 +124,13 @@ object typing {
     def typedClass[T: ClassTag]: TypedClass = typedClass(toRuntime[T])
 
     //TODO: make it more safe??
-    def typedClass(klass: Class[_], parameters: List[TypingResult] = Nil): TypedClass = if (klass == classOf[Any]) {
-      throw new IllegalArgumentException("Cannot have typed class of Any, use Unknown")
-    } else if (klass.isPrimitive) {
-      TypedClass(ClassUtils.primitiveToWrapper(klass), parameters)
-    } else if (klass.isArray) {
+    def typedClass(klass: Class[_], parameters: List[TypingResult] = Nil): TypedClass =
```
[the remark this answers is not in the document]
maciej-brzezinski: yes of course, forgot i pushed a version without it

engine/flink/test-util/src/main/scala/pl/touk/nussknacker/engine/flink/test/ClassExtractionBaseTest.scala
maciej-brzezinski: changed it back with updated classes and methods, unfortunately still not sure why it doesn't pass the test, parameters from methods extracted from file are parsed as unknown where parameters from model.processWithObjectsDefinition are known classes 
i will write update once i fixed it

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/util/ReflectUtils.scala
maciej-brzezinski: added the tests for display, please take a look if that's what you meant

engine/interpreter/src/test/scala/pl/touk/nussknacker/engine/types/EspTypeUtilsSpec.scala
maciej-brzezinski: removed the test, you are right, there wasn't much sense for it noce arrays are included

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
maciej-brzezinski: changed it

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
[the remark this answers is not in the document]
maciej-brzezinski: returned it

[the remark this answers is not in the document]
maciej-brzezinski: returned it

engine/api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -125,13 +123,24 @@ object typing {
...
+        TypedClass(ClassUtils.primitiveToWrapper(klass), parameters)
+      } else if (klass.isArray) {
+        //to not have separate class for each array, we pass Array of Objects
+        if (klass.getComponentType == classOf[Object]) {
+          TypedClass(klass, parameters)
+        } else parameters match {
+          case Nil =>
+            Typed.typedClass(classOf[Array[Object]], List(Typed(klass.getComponentType)))
+          case _: List[TypingResult] =>
+            throw new IllegalArgumentException("Array parameter type passed twice")
```
mproch, line 139: We could add passed types (both from parameters and from componentType) in message?

mproch (approved):

PR #2222, opened 2021-09-20
Preview/gh actions improvements part2

arkadius:


.github/workflows/pr.yml
```diff
@@ -215,30 +215,35 @@ jobs:
     name: FrontendTests
     runs-on: ubuntu-latest
     needs: [ changes ]
-    if: ${{ needs.changes.outputs.fe_changes_count > 0 || github.ref == 'refs/heads/staging' }}
     steps:
       - name: Cancel previous runs
+        if: ${{ needs.changes.outputs.fe_changes_count > 0 || github.ref == 'refs/heads/staging' }}
```
mproch, line 220: maybe we can set some var, e.g. env and not repeat it so many times?

.github/workflows/pr.yml
```diff
@@ -372,14 +383,14 @@ jobs:
           echo "GIT_SOURCE_BRANCH=$GIT_SOURCE_BRANCH" >> $GITHUB_ENV
           SANITIZED_BRANCH=`echo ${GIT_SOURCE_BRANCH} | sed 's/[^a-zA-Z0-9.\-_]/\_/g' | awk '{print tolower($0)}'`
           echo "VERSION_SUFFIX=-$SANITIZED_BRANCH-$(date -I)-$GITHUB_RUN_NUMBER-$GITHUB_SHA" >> $GITHUB_ENV
-      - name: Build
+      - name: Publish with scaladocs
+        if: ${{ github.ref == 'refs/heads/staging' }}
```
mproch, line 387: What about release branches?

mproch (approved):

PR #2232, opened 2021-09-21
GH actions improvement: not building image if changes are fe-only

arkadius:


.github/workflows/pr.yml
```diff
@@ -121,7 +136,7 @@ jobs:
     name: CrossCompile
     runs-on: ubuntu-latest
     needs: [ build, changes ]
-    if: ${{ needs.changes.outputs.all_changes_count > needs.changes.outputs.fe_changes_count || github.ref == 'refs/heads/staging' }}
+    if: ${{ needs.changes.outputs.all_changes_count > needs.changes.outputs.fe_changes_count || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/master' || github.ref == 'refs/heads/demo' || startsWith(github.ref, 'refs/heads/preview') || startsWith(github.ref, 'refs/heads/release') }}
```
lciolecki, line 139: May be we can define array of urls and simplify this if?

PR #2247, opened 2021-09-23
Fix scenario status when job manager is down

dswiecki:


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ManagementActor.scala
```diff
@@ -209,11 +209,12 @@ class ManagementActor(managers: ProcessingTypeDataProvider[DeploymentManager],
             ProcessStatus.simpleErrorMismatchDeployedVersion(ver.versionId, action.processVersionId, action.user, processState)
           case Some(ver) if ver.versionId == action.processVersionId =>
             state
+          case None if !state.isDeployed =>
```
mproch, line 212: let's just change line 206 to `case _ if !state.isDeployed => `?

dswiecki: yeah, i've been thinking about it ;)

mproch (approved):

mproch (approved):

PR #2252, opened 2021-09-23
GH actions: build FE once

arkadius:


build.sbt
```diff
@@ -1000,7 +1001,13 @@ lazy val ui = (project in file("ui/server"))
   .settings(
     name := "nussknacker-ui",
     buildUi :=  {
-      runNpm("run build", "Client build failed", (compile / crossTarget).value)
+      if (useFEBuildTarget) {
+        val feDistDirectory = file("ui/client/dist")
+        val feDistFiles: Seq[File] = (feDistDirectory ** "*").get()
+        IO.copy(feDistFiles pair Path.rebase(feDistDirectory, (compile / crossTarget).value / "classes" / "web" / "static"), CopyOptions.apply(overwrite = true, preserveLastModified = true, preserveExecutable = false))
```
mproch, line 1007: Maybe we can just have this code regardless of useFEBuildTarget and remove OUTPUT_PATH settings to be consistent?

.github/workflows/pr.yml
```diff
@@ -89,6 +89,47 @@ jobs:
...
           name: build-target
           path: target.tgz
+
+  build-fe:
+    name: BuildFrontend
+    runs-on: ubuntu-latest
+    needs: [ changes ]
+    env:
+      # We can't just use conditional jobs mechanism ('if' directive) because 'publish' job depends on this one.
+      shouldPerformFrontendBuild: ${{ needs.changes.outputs.fe_changes_count > 0 }}
```
mproch, line 99: where do we use this variable?

arkadius: removed - thanks for notice

mproch (approved):

PR #2274, opened 2021-09-27
Spel selection for ^ and $ operand

dswiecki:


engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -301,19 +300,25 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
     })
   }
 
+  //currently there is no better way than to check ast string starting with $ or ^
+  private def resolveSelectionTypingResult(node: Selection, parentType: TypingResultWithContext, childElementType: TypingResult) = {
+    val isSingleElementSelection = List("$", "^").map(node.toStringAST.startsWith(_)).foldLeft(false)(_ || _)
```
mproch, line 305: We can use TokenKind.SELECT_LAST/FIRST

mproch (approved):
I added small comment

engine/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
dswiecki: Unfortunately its package private

PR #2281, opened 2021-09-28
Fix Flink's restart strategy config reader

nadberezny:


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/RestartStrategyFromConfiguration.scala
```diff
@@ -61,9 +61,15 @@ object RestartStrategyFromConfiguration {
...
     config.entrySet().asScala.foreach { entry =>
-      flinkConfig.setString(s"$restartStrategyFlinkConfigPrefix.$strategy.${entry.getKey}", entry.getValue.render())
+      flinkConfig.setString(s"$restartStrategyFlinkConfigPrefix.$strategy.${entry.getKey}", toString(entry.getValue))
     }
     RestartStrategies.fromConfiguration(flinkConfig).asScala
       .getOrElse(throw new IllegalArgumentException(s"Failed to find configured restart strategy: $strategy"))
   }
+
+  private def toString(configValue: ConfigValue): String =
+    if (configValue.valueType() == ConfigValueType.STRING)
```
arkadius, line 71: WDYT about `configValue.render(ConfigRenderOptions.defaults().setJson(false)` instead of this if-else. Or it won't work?

engine/flink/util/src/test/scala/pl/touk/nussknacker/engine/flink/util/exception/ConfigurableExceptionhandlerSpec.scala
```diff
@@ -21,7 +21,7 @@ class ConfigurableExceptionhandlerSpec extends FunSuite with Matchers {
     //it's difficult to mock RuntimeContext for metrics so we switch it off..
     "exceptionHandler.withRateMeter" -> false,
     "restartStrategy.default.strategy" -> "fixed-delay",
-    "restartStrategy.default.delay" -> 10,
+    "restartStrategy.default.delay" -> "10 ms",
```
arkadius, line 24: Numerical delay will still work? If not, it will be needed entry in migration guide

nadberezny: Yes, numerical delay still will be working.

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/RestartStrategyFromConfiguration.scala
nadberezny: It won't work when value in hocon config will be quoted. The result will be:
`restart-strategy.fixed-delay.delay -> ""10 ms""`

arkadius (approved):

PR #2295, opened 2021-10-01
Better exception handling for LazyParameter

mproch:


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
```diff
@@ -53,8 +54,7 @@ object keyed {
 
   }
 
-  /* TODO, FIXME: Errors on interpret should be handled with FlinkEspExceptionHandler; see KeyedRecordFlatMapper */
-  abstract class BaseKeyedValueMapper[OutputKey <: AnyRef: TypeTag, OutputValue <:AnyRef: TypeTag] extends RichMapFunction[Context, ValueWithContext[KeyedValue[OutputKey, OutputValue]]] with LazyParameterInterpreterFunction {
+  abstract class BaseKeyedValueMapper[OutputKey <: AnyRef: TypeTag, OutputValue <:AnyRef: TypeTag] extends RichFlatMapFunction[Context, ValueWithContext[KeyedValue[OutputKey, OutputValue]]] with LazyParameterInterpreterFunction {
```
arkadius, line 57: Shouldn't we made some changes also in `KeyedRecordFlatMapper` after this change so it can look like other `LazyParameterInterpreterFunction`s ?

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/transformer/CustomFilter.scala
```diff
@@ -11,7 +11,7 @@ case object CustomFilter extends CustomStreamTransformer {
   def execute(@ParamName("expression") expression: LazyParameter[java.lang.Boolean]): FlinkCustomStreamTransformation
   = FlinkCustomStreamTransformation((start: DataStream[Context], ctx: FlinkCustomNodeContext) =>
     start
-      .filter(ctx.lazyParameterHelper.lazyFilterFunction(expression))
+      .filter(new LazyParameterFilterFunction(expression, ctx.lazyParameterHelper))
```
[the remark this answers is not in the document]
mproch: forgot to cleanup previous experiments, I'll fix it...

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
mproch: true :+1:

PR #2301, opened 2021-10-05
GenericNodeTransformation.initialParameters removal

arkadius:


mproch (approved):

engine/components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseLookupEnricher.scala
```diff
@@ -61,10 +61,6 @@ class DatabaseLookupEnricher(dBPoolConfig: DBPoolConfig, dbMetaDataProvider: DbM
       QueryArguments(
         QueryArgument(index = 1, value = params(KeyValueParamName)) :: Nil)
 
-  override def initialParameters: List[Parameter] = {
```
dswiecki, line 64: hmm will it work ?

arkadius: why not? currently this initial parameters step was implemented everywhere twice - for purpose of parameters used in ours component templates (aka NodeToAdd) and in node validation. first place was needed because it was the only place where defaults were respected. after previous PR that I've made it is not needed anymore - now node validation also deliver this defaults.

arkadius: Ahh I know what you mean by this comment now. I see now that `DatabaseLookupEnricher.initialParameters` wasnt used only in component templates but it was a cycle of dependencies: `DatabaseLookupEnricher.contextTransformation` uses super class method `DatabaseQueryEnricher.initialStep` which uses `DatabaseLookupEnricher.initialParameters` which overrides `DatabaseQueryEnricher.initialParameters`. Nice mess. Would be gratefull for more detailed comments in the future :P

PR #2377, opened 2021-10-21
Remove clazz from SourceFactory

mproch:


api/src/main/scala/pl/touk/nussknacker/engine/api/process/Source.scala
```diff
@@ -38,9 +38,7 @@ trait TestDataGenerator { self: Source[_] with SourceTestSupport[_] =>
   * IMPORTANT lifecycle notice:
   * Implementations of this class *must not* allocate resources (connections, file handles etc.)
   */
-trait SourceFactory[+T] extends Serializable with Component {
-  def clazz : Class[_]
-}
+trait SourceFactory[+T] extends Serializable with Component
```
arkadius, line 41: `T` is still needed? Do we have a plans to also remove it in the future? Maybe some TODO to clean it up?

arkadius (approved):

api/src/main/scala/pl/touk/nussknacker/engine/api/process/Source.scala
mproch: :+1: Very good remark, I added TODO

PR #2386, opened 2021-10-25
Sort scenario list by status

dswiecki:


ui/client/containers/Processes.tsx
```diff
@@ -33,7 +33,7 @@ const ElementsRenderer: RowsRenderer = ({processes, statuses}) => {
         <Td column="lastAction" className="centered-column" value={process?.lastAction?.performedAt}>
           <ProcessLastAction process={process}/>
         </Td>
-        <Td column="status" className="status-column">
+        <Td column="status" className="status-column" value={process.state.status.name}>
```
JulianWielga, line 36: maybe we need some number value (and some mapping) here?
this way we could get better order

JulianWielga (approved):

ui/client/containers/Processes.tsx
dswiecki: Ok I'll add this in another PR

PR #2394, opened 2021-10-26
nu-engine application launcher

maciej-brzezinski:


engine/base/kafka/src/universal/conf/logback.xml
```diff
@@ -0,0 +1,59 @@
...
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>${CONSOLE_THRESHOLD_LEVEL:-INFO}</level>
+        </filter>
+        <encoder>
+            <Pattern>%d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</Pattern>
+        </encoder>
+    </appender>
+
+    <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>${LOGS_DIR:-./logs}/nussknacker-designer.log</file>
```
mproch, line 14: I don't think it should be nussknacker-designer :)

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  val scenarioInterpreter = prepareScenarioInterpreter
+
+  Using.resource(scenarioInterpreter) { interpreter =>
+    interpreter.run()
+  }
+
+  private def parseArgs = {
+    if (args.length < 1) {
+      System.err.println("Missing scenario_file_location argument!")
```
mproch, line 36: Do we want to add it also to normal log?

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
```diff
@@ -0,0 +1,80 @@
...
+    val modelData = ModelData(modelConfig, ModelClassLoader(modelConfig.as[List[URL]]("classPath")))
+    // TODO Use correct MetricProvider
+    val preparer = new EngineRuntimeContextPreparer(NoOpMetricsProvider)
+    // TODO Pass correct ProcessVersion and DeploymentData
+    val jobData = JobData(scenario.metaData, ProcessVersion.empty, DeploymentData.empty)
+
+    val exceptionHandler = new UncaughtExceptionHandler {
+      override def uncaughtException(t: Thread, e: Throwable): Unit = {
+        // TODO Is it ok?
+        throw e
```
mproch, line 74: Can we add some error to log?

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
```diff
@@ -0,0 +1,80 @@
...
+  private def prepareScenarioInterpreter = {
+    // TODO Should it be extracted from designer?
+    val engineConfig = ConfigFactory.load(ConfigFactory.parseFile(new File(System.getProperty("nussknacker.config.locations"))))
+    val modelConfig: Config = engineConfig.getConfig("modelConfig")
+
+    val modelData = ModelData(modelConfig, ModelClassLoader(modelConfig.as[List[URL]]("classPath")))
+    // TODO Use correct MetricProvider
+    val preparer = new EngineRuntimeContextPreparer(NoOpMetricsProvider)
+    // TODO Pass correct ProcessVersion and DeploymentData
+    val jobData = JobData(scenario.metaData, ProcessVersion.empty, DeploymentData.empty)
```
mproch, line 69: I wonder, do we want to pass it in other args? Or in some JSON containing those data together with the scenario?

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
```diff
@@ -0,0 +1,80 @@
...
+      sys.exit(3)
+    }
+  }
+
+  private def prepareScenarioInterpreter = {
+    // TODO Should it be extracted from designer?
+    val engineConfig = ConfigFactory.load(ConfigFactory.parseFile(new File(System.getProperty("nussknacker.config.locations"))))
+    val modelConfig: Config = engineConfig.getConfig("modelConfig")
+
+    val modelData = ModelData(modelConfig, ModelClassLoader(modelConfig.as[List[URL]]("classPath")))
```
mproch, line 65: I wonder, do we need a separate classloader here? Maybe it'd be easier to just assume a flat classpath?

build.sbt
```diff
@@ -827,11 +827,46 @@ lazy val baseEngineRuntime = (project in engine("base/runtime")).
...
 
-lazy val kafkaBaseEngineRuntime = (project in engine("base/kafka")).
+lazy val kafkaBaseEngineRuntime: Project = (project in engine("base/kafka")).
   settings(commonSettings).
+  settings(publishAssemblySettings: _*).
+  enablePlugins(SbtNativePackager, JavaServerAppPackaging).
   settings(
-    name := "nussknacker-baseengine-runtime-kafka"
+    name := "nussknacker-baseengine-runtime-kafka",
+    assembly / assemblyOption := (assembly / assemblyOption).value.copy(includeScala = true, level = Level.Info),
```
mproch, line 852: do we want assembly if we'll have .tgz and docker distributions?

build.sbt
```diff
@@ -827,11 +827,46 @@ lazy val baseEngineRuntime = (project in engine("base/runtime")).
...
+      "commons-io" % "commons-io" % commonsIOV
+    )
+  ).dependsOn(interpreter % "it", kafkaUtil % "it", testUtil % "it", kafkaTestUtil % "it")
+
 
-lazy val kafkaBaseEngineRuntime = (project in engine("base/kafka")).
+lazy val kafkaBaseEngineRuntime: Project = (project in engine("base/kafka")).
   settings(commonSettings).
+  settings(publishAssemblySettings: _*).
+  enablePlugins(SbtNativePackager, JavaServerAppPackaging).
```
mproch, line 849: Can we add at least TODO for docker distribution?

engine/base/nu-kafka-engine-bin-test/src/it/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineBinTest.scala
```diff
@@ -0,0 +1,89 @@
...
+    kafkaClient.createTopic(outputTopic, 1)
+
+    val scenario = EspProcessBuilder
+        .id("test")
+        .exceptionHandler()
+        .source("source", "kafka-json", "topic" -> s"'$inputTopic'")
+        .emptySink("sink", "kafka-json", "topic" -> s"'$outputTopic'", "value" -> "#input")
+    val jsonFile = saveScenarioToTmp(scenario)
+
+    val engineExitCodePromise = Promise[Int]
```
mproch, line 39: I wonder, will it be possible to use dockerPublishLocal + TestContainers for this test in the future?

engine/base/nu-kafka-engine-bin-test/src/it/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineBinTest.scala
```diff
@@ -0,0 +1,89 @@
...
+  private def saveScenarioToTmp(scenario: EspProcess) = {
+    val canonicalScenario = ProcessCanonizer.canonize(scenario)
+    val json = ProcessMarshaller.toJson(canonicalScenario)
+    val jsonFile = File.createTempFile(getClass.getSimpleName, ".json")
+    jsonFile.deleteOnExit()
+    FileUtils.write(jsonFile, json.toString(), StandardCharsets.UTF_8)
+    jsonFile
+  }
+
+  private def shellScriptPath = {
```
mproch, line 78: can we add result types?

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
```diff
@@ -0,0 +1,80 @@
...
+      System.err.println("Missing scenario_file_location argument!")
+      System.err.println("")
+      System.err.println("Usage: ./run.sh scenario_file_location.json")
+      sys.exit(1)
+    }
+
+    Path.of(args(0))
+  }
+
+  private def parseScenario = {
```
mproch, line 45: can we add result types?

build.sbt
arkadius: added

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
arkadius: not sure if it is a good approach. normal logs are also printed on stdout as it we also have in designer so we will have it duplicated

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
arkadius: I like current approach because we have explicitly splitted core engine classpath from component (providers). After joining them we could miss some dependency problem. Also we are not loading jars that are not used like openapi and sql in this example. I think that we could think about auto-discovered classpath based on some conventions but IMO it is more general issue (not this engine specific) WDYT?

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
arkadius: I would do that but regards to YAGNIY currently see no usage of this additional data. Can you provide some test scenario that will show usage of them? Also don't think that without fixing an issue with components we can go to the point in test that will show that it works correctly

engine/base/nu-kafka-engine-bin-test/src/it/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineBinTest.scala
arkadius: I would like to do this in separate PR

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
arkadius: added

build.sbt
arkadius: cleared

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
arkadius: Also am I right that we should do `sys.exit` after logging? I've found your note about that in test - added

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
mproch: On the module level - it's good to have them separated, I agree. But in runtime - I don't see that much benefit from creating a custom classloader. But we can leave it, it's not much fuss

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
mproch: If we are to pass them as the next arguments, I agree we don't need it now. But if we would like more complex JSON structure of first argument - then maybe it's better to do it now

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineApp.scala
mproch: Hmm... I started to look at our logback, and isn't it so, that we will have duplicate content in logs when using 'in background'? One in .out, one in .log?

engine/base/nu-kafka-engine-bin-test/src/it/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaEngineBinTest.scala
mproch: :+1:

mproch (approved):

PR #2406, opened 2021-11-02
Metrics in base engine

mproch:
Metrics in Kafka engine, on par with Flink engine metrics (the main difference is that dot (.) is used instead of _ as a separator)


engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/SourceMetrics.scala
```diff
@@ -0,0 +1,43 @@
...
+class SourceMetrics(metricsProvider: MetricsProviderForScenario) {
+
+  private val sourceMetrics = collection.concurrent.TrieMap[SourceId, MetricsForOneSource]()
+
+  def markElement(sourceId: SourceId, elementTimestamp: Long): Unit = {
+    sourceMetrics.getOrElseUpdate(sourceId, new MetricsForOneSource(sourceId)).process(elementTimestamp)
+  }
+
+  protected def now(): Long = {
+    System.currentTimeMillis()
```
arkadius, line 16: Can we just pass a `Clock` as a field to constructor? Then will be available fixed clock and won't be needed inheritance.

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/RateMeterExceptionConsumer.scala
```diff
@@ -0,0 +1,21 @@
+package pl.touk.nussknacker.engine.baseengine.metrics
+
+import cats.data.NonEmptyList
+import pl.touk.nussknacker.engine.api.Lifecycle
+import pl.touk.nussknacker.engine.api.exception.EspExceptionConsumer
+import pl.touk.nussknacker.engine.api.runtimecontext.EngineRuntimeContext
+import pl.touk.nussknacker.engine.util.exception.GenericRateMeterExceptionConsumer
+import pl.touk.nussknacker.engine.util.metrics.{InstantRateMeterWithCount, MetricsProviderForScenario, RateMeter}
+
+class RateMeterExceptionConsumer extends GenericRateMeterExceptionConsumer with Lifecycle {
```
arkadius, line 10: Current name looks like interface that is implemented by `GenericRateMeterExceptionConsumer` - not otherwise

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/dropwizard/BaseEngineMetrics.scala
```diff
@@ -0,0 +1,37 @@
...
+      }
+    } else {
+      BaseEngineInfluxDbReporter.createAndRunReporterIfConfigured(metricRegistry, config)
+    }
+    metricRegistry
+  }
+
+  private def loadMetricsReporters(): List[MetricsReporter] = {
+    try {
+      val reporters = ScalaServiceLoader.load[MetricsReporter](Thread.currentThread().getContextClassLoader)
```
arkadius, line 28: Why it was needed to by loaded via ServiceLoader? (just asking)

utils/util/src/main/scala/pl/touk/nussknacker/engine/util/exception/GenericRateMeterExceptionConsumer.scala
```diff
@@ -1,29 +1,17 @@
 package pl.touk.nussknacker.engine.util.exception
 
-import cats.data.NonEmptyList
 import pl.touk.nussknacker.engine.api.exception.{EspExceptionConsumer, EspExceptionInfo, NonTransientException}
-import pl.touk.nussknacker.engine.util.metrics.RateMeter
 
-trait GenericRateMeterExceptionConsumer extends EspExceptionConsumer {
+trait GenericRateMeterExceptionConsumer extends EspExceptionConsumer with ExceptionRateMeter {
```
arkadius, line 5: 1. We count exceptions in both exception consumer and  `NodeCountingListener` ? is it ok? Won't be it doubled / shadowed? 
2. Can't it be composed instead of inherited? :-(

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/NodeCountingListener.scala
```diff
@@ -0,0 +1,40 @@
+package pl.touk.nussknacker.engine.baseengine.metrics
+
+import cats.data.NonEmptyList
+import pl.touk.nussknacker.engine.api.exception.EspExceptionInfo
+import pl.touk.nussknacker.engine.api.runtimecontext.EngineRuntimeContext
+import pl.touk.nussknacker.engine.api.{Context, EmptyProcessListener, Lifecycle, MetaData}
+import pl.touk.nussknacker.engine.util.exception.ExceptionRateMeter
+import pl.touk.nussknacker.engine.util.metrics.{Counter, InstantRateMeterWithCount, MetricIdentifier, MetricsProviderForScenario, RateMeter}
+
+class NodeCountingListener extends EmptyProcessListener with Lifecycle with ExceptionRateMeter {
```
arkadius, line 10: Can `ExceptionRateMeter` be composed instead of inherited?

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/SourceMetrics.scala
mproch: :+1: I forgot about it :)

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/RateMeterExceptionConsumer.scala
mproch: I cleaned up a little bit - I've left only RateMeterExceptionConsumer for Flink with comment why it's needed there

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/dropwizard/BaseEngineMetrics.scala
mproch: TBH I just copied it from StandaloneApp. The idea was to be able to add other providers with custom jars, not sure if it was much used, can change it, but how would you like it to work?

utils/util/src/main/scala/pl/touk/nussknacker/engine/util/exception/GenericRateMeterExceptionConsumer.scala
mproch: I cleaned up a bit, left RateExceptionConsumer only in Flink with comment

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/NodeCountingListener.scala
mproch: :+1:

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/KafkaSingleScenarioTaskRun.scala
```diff
@@ -79,11 +80,13 @@ class KafkaSingleScenarioTaskRun(metaData: MetaData,
...
 
 
-  private def deserializeRecords(records: ConsumerRecords[Array[Byte], Array[Byte]]): List[(SourceId, Context)] = {
+  private def prepareRecords(records: ConsumerRecords[Array[Byte], Array[Byte]]): List[(SourceId, Context)] = {
     sourceToTopic.toList.flatMap {
       case (topic, sources) =>
         val forTopic = records.records(topic).asScala.toList
-        sources.mapValues(source => forTopic.map(source.deserialize(context, _))).toList.flatMap {
+        //TODO: try to handle source metrics in more generic way?
+        sources.keys.foreach(sourceId => forTopic.foreach(record => sourceMetrics.markElement(sourceId, record.timestamp())))
```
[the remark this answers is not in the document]
arkadius: 2. Ahh, I missed that thing - it's fine as it is
3. So maybe we should pass (sourceId, forTopic.map(_.timestamp()) and inside `MetricsForOneSource.process(timestamps: List[Long])` do `timestamps.max`? WDYT about my performance considerations?

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/RateMeterExceptionConsumer.scala
```diff
@@ -1,26 +1,34 @@
...
-import pl.touk.nussknacker.engine.util.metrics.{InstantRateMeterWithCount, RateMeter}
+import pl.touk.nussknacker.engine.util.exception.ExceptionRateMeter
+import pl.touk.nussknacker.engine.util.metrics.WithMetrics
 
-class RateMeterExceptionConsumer(val underlying: FlinkEspExceptionConsumer) extends FlinkEspExceptionConsumer with WithMetrics with GenericRateMeterExceptionConsumer with Lifecycle  {
+//For Flink we count metrics in ExceptionHandler, not e.g. in NodeCountListener, so that it's easier to
+//handle errors consistently in CustomStreamTransformers
+class RateMeterExceptionConsumer(val underlying: FlinkEspExceptionConsumer) extends FlinkEspExceptionConsumer with WithMetrics {
+
+  private var exceptionRateMeter: ExceptionRateMeter = _
```
arkadius, line 13: Can it be `transient`? Not sure if it needed but looks like it can be needed.

engine/base/kafka/src/main/scala/pl/touk/nussknacker/engine/baseengine/kafka/KafkaSingleScenarioTaskRun.scala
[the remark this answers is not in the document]
mproch: The max timestamp is not enough as we compute histogram there. Actually, implementation was not thread-safe, so I changed it a little bit. The current implementation is in fact not using synchronization - AFAIK histogram and other structures are using lock-free implementations, so performance shouldn't be a problem.

engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/exception/RateMeterExceptionConsumer.scala
mproch: Don't think it needs to be, but it won't hurt, added :+1:

arkadius (approved):

PR #2408, opened 2021-11-03
Improvements: Simple components deduplication

lciolecki:
- Provided componentId at SingleComponentConfig
- Allow to override componentId by SingleComponentConfig at ComponentElementList
- Simple deduplication components

TODO: 
- Add validation for deduplicated components witch attributes: name, componentType, componentGroupName when application is starting

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
@@ -83,7 +83,7 @@ case class ComponentExtractor(classLoader: ClassLoader, nussknackerVersion: Nuss
   private def extractOneProviderConfig(config: ComponentProviderConfig, provider: ComponentProvider, processObjectDependencies: ProcessObjectDependencies): List[(String, WithCategories[Component])] = {
     provider.create(config.config, processObjectDependencies).map { cd =>
       val finalName = config.componentPrefix.map(_ + cd.name).getOrElse(cd.name)
-      finalName -> WithCategories(cd.component, config.categories, SingleComponentConfig.zero.copy(docsUrl = cd.docsUrl, icon = cd.icon))
+      finalName -> WithCategories(cd.component, config.categories, SingleComponentConfig.zero.copy(docsUrl = cd.docsUrl, icon = cd.icon, componentId = Some(ComponentId.create(finalName))))
     }
   }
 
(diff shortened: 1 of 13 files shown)


ui/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/definition/package.scala
```diff
@@ -64,10 +65,11 @@ package object definition {
   import pl.touk.nussknacker.engine.graph.NodeDataCodec._
   object ComponentTemplate {
     def create(`type`: ComponentType, node: NodeData, categories: List[String], branchParametersTemplate: List[evaluatedparam.Parameter] = List.empty): ComponentTemplate =
-      ComponentTemplate(`type`, `type`.toString, node, categories, branchParametersTemplate)
+      ComponentTemplate(`type`, `type`.toString, node, categories, branchParametersTemplate, componentId = None)
   }
 
-  @JsonCodec(encodeOnly = true) case class ComponentTemplate(`type`: ComponentType, label: String, node: NodeData, categories: List[String], branchParametersTemplate: List[evaluatedparam.Parameter] = List.empty)
+  //TODO: componentId is temporary field and it is work around for components duplication across multiple scenario types
+  @JsonCodec(encodeOnly = true) case class ComponentTemplate(`type`: ComponentType, label: String, node: NodeData, categories: List[String], branchParametersTemplate: List[evaluatedparam.Parameter] = List.empty, componentId: Option[ComponentId])
```
arkadius, line 72: Why `componentId` is needed in `ComponentTemplate`? Template is for creating node based on component and should has similar fields to be used in `Node`.  During writting it, I've realized that it is not a `Component` `Template`. It is `NodeTemplate` based on `Component`

lciolecki: Let's look at DefaultComponentService 78 line.  To fetch all components we based on `UIProcessObjectsFactory.prepareUIProcessObjects` which is preparing ComponentTempalte - we use this DTO to create our ComponentListElement.

But right now I see we can use `componentsConfig`, because we put here merged componentsConfig (fixed - config, dynamic - from ProcessCreator). WDT?

interpreter/src/main/scala/pl/touk/nussknacker/engine/component/ComponentExtractor.scala
```diff
@@ -80,7 +80,7 @@ case class ComponentExtractor(classLoader: ClassLoader, nussknackerVersion: Nuss
   private def extractOneProviderConfig(config: ComponentProviderConfig, provider: ComponentProvider, processObjectDependencies: ProcessObjectDependencies): List[(String, WithCategories[Component])] = {
     provider.create(config.config, processObjectDependencies).map { cd =>
       val finalName = config.componentPrefix.map(_ + cd.name).getOrElse(cd.name)
-      finalName -> WithCategories(cd.component, config.categories, SingleComponentConfig.zero.copy(docsUrl = cd.docsUrl, icon = cd.icon))
+      finalName -> WithCategories(cd.component, config.categories, SingleComponentConfig.zero.copy(docsUrl = cd.docsUrl, icon = cd.icon, componentId = Some(ComponentId.create(finalName))))
```
arkadius, line 83: Ahh I've take a look into how those components are configured and it is not that way as I was thinking. Take a look at `dev-application.conf` - they are defined for each `scenarioType`. I'm thinking about passing there `None` and let's user define componentId in `componentsUiConfig`. But I see now one problem with `componentsUiConfig`. It is map `label -> config` and it should be rather `(label, componentType) -> config` to make sure that `kafka-avro` source and sink will work together. But maybe we can left it how it is and think about it when this problem appear?

arkadius (approved):

PR #2409, opened 2021-11-03
GenericNodeTransformation - checking validators during steps evaluati…

arkadius:
…on instead of at the end - it was too late because implementation was invoked but it shouldn't be

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
@@ -1,6 +1,5 @@
 package pl.touk.nussknacker.engine.compile
 
-import cats.data.{NonEmptyList, Validated}
 import cats.data.Validated.{invalid, valid}
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{MissingParameters, NodeId, RedundantParameters}
 import pl.touk.nussknacker.engine.api.context._
@@ -12,9 +11,9 @@ object Validations {
   import cats.data.ValidatedNel
   import cats.implicits._
 
-  def validateParameters[T >: PartSubGraphCompilationError <: ProcessCompilationError](parameterDefinitions: List[Parameter],
-                                                                                       parameters: List[evaluatedparam.Parameter])
-                                                                                      (implicit nodeId: NodeId): ValidatedNel[T, Unit] = {
+  def validateParameters(parameterDefinitions: List[Parameter],
+                         parameters: List[evaluatedparam.Parameter])
+                        (implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, Unit] = {
     val definedParamNamesSet = parameterDefinitions.map(_.name).toSet
     val usedParamNamesSet = parameters.map(_.name).toSet
 
@@ -29,9 +28,9 @@ object Validations {
       ).mapN { (_, _, _) => Unit }
   }
 
-  def validateSubProcessParameters[T >: PartSubGraphCompilationError <: ProcessCompilationError](definedParamNamesSet: Set[String],
-                                                                                                 usedParamNamesSet: Set[String])
-                                                                                                (implicit nodeId: NodeId): ValidatedNel[T, Unit] = {
+  def validateSubProcessParameters(definedParamNamesSet: Set[String],
+                                   usedParamNamesSet: Set[String])
+                                  (implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, Unit] = {
     val validatedRedundant = validateRedundancy(definedParamNamesSet, usedParamNamesSet)
     val validatedMissing = validateMissingness(definedParamNamesSet, usedParamNamesSet)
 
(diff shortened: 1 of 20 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
```diff
@@ -56,14 +57,21 @@ object Validations {
...
-    val validationResults = paramWithValidatorList.map {
-      case (param, validator) => validator.isValid(param.name, param.expression.expression, None).toValidatedNel
-    }
+      paramDefinition <- definitionsMap.get(param.name)
+      paramValidationResult = validateParameterWithCustomValidators(paramDefinition, param)
+    } yield paramValidationResult
     validationResults.sequence.map(_ => Unit)
   }
+
+  def validateParameterWithCustomValidators[T >: PartSubGraphCompilationError <: ProcessCompilationError](paramDefinition: Parameter,
```
mproch, line 69: I wonder, could it be a method in Parameter? Or do we want to keep them anemic ;) 
Also, I think 'custom' is misleading here - they are in most cases rather standard?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/ConfluentSchemaRegistryClient.scala
```diff
@@ -29,6 +32,35 @@ trait ConfluentSchemaRegistryClient extends SchemaRegistryClient with LazyLoggin
     }
 }
 
+class DefaultConfluentSchemaRegistryClient(override val client: CSchemaRegistryClient) extends ConfluentSchemaRegistryClient {
```
mproch, line 35: I don't think I fully understand why is this needed for this change?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/StandardParameterEnrichment.scala
```diff
@@ -21,13 +21,22 @@ object StandardParameterEnrichment {
...
+    val finalValidators = determineValidatorss(original, parameterConfig, parameterData, finalEditor)
     val isOptional = original.isOptional
     val finalDefaultValue = original.defaultValue.orElse(DefaultValueDeterminerChain.determineParameterDefaultValue(
       DefaultValueDeterminerParameters(parameterData, isOptional, parameterConfig, finalEditor)))
     original.copy(editor = finalEditor, validators = finalValidators, defaultValue = finalDefaultValue)
   }
 
+  private def determineValidatorss(original: Parameter, parameterConfig: ParameterConfig, parameterData: ParameterData, finalEditor: Option[ParameterEditor]) = {
+    // In case if someone forget to specify validators and used only Parameter() or Parameter.optional(), we add missing forgotten validators
+    // It can be tricky, because we have no option to specify that we don't want those "standard" validators
```
mproch, line 33: So, if someone specifies own validators we skip 'standard' ones? Maybe it would be better to add explicitly some flag, like 'skipStandardValidators'? If you need this change now for some reasons - let's add it to changelog/migration guide, as this can be tricky change in behaviour?

mproch (commented):
In general I like it, good change. Added some minor remarks

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/ConfluentSchemaRegistryClient.scala
arkadius: Because in tests was used cached implementation and each test case was changing behaviour of other test cases

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/StandardParameterEnrichment.scala
arkadius: We currently have such a case: in `GenericTypedJsonSourceFactory` we have parameter `type` with `JsonParameterEditor` which causes adding `JsonValidator` but it is overriden by explicit list of validators. I personally don't like adding those "standard" validators at all and don't want to adding extra fields to `Parameter` for that. WTYT about removing this implication: `JsonParameterEditor` ->  `JsonValidator` and let's developer decide if he/she want to add only editor or both editor and validator?

mproch: ok, I looked a bit at it, and I'm in favour of not adding these validators at all ;)

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
arkadius: fixed

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/StandardParameterEnrichment.scala
arkadius: Changed but for now I only removed `JsonValidator` from those `EditorBasedValidatorExtractor` - must be turned on explicitly because I see a pain for developers to defined both editor and validator for "fixed-values" case. At the end we should rather add something like `ParameterEditor.validators` method and in `Paramter` collect both `validators` and `editor.validators` but don't want to change it in this PR

mproch (approved):

PR #2443, opened 2021-11-05
fix: module federation context overwrite

JulianWielga:
loading module a second time (and each subsequent) was causing  context overwrite, which resulted in overwriting and bad generation of material-ui styles


lciolecki (approved):
I knew, you will fix  it ;)

PR #2453, opened 2021-11-09
[FEATURE] Allow providing custom actions for periodic scenarios from…

gadomsky:
… outside of PeriodicDeploymentManager

### DIFF
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicCustomActionsProviderFactory.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicCustomActionsProviderFactory.scala
@@ -1,0 +1,26 @@
+package pl.touk.nussknacker.engine.management.periodic
+
+import pl.touk.nussknacker.engine.api.deployment._
+import pl.touk.nussknacker.engine.management.periodic.db.PeriodicProcessesRepository
+
+import scala.concurrent.Future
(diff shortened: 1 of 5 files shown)


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicCustomActionsProviderFactory.scala
```diff
@@ -0,0 +1,26 @@
+package pl.touk.nussknacker.engine.management.periodic
+
+import pl.touk.nussknacker.engine.api.deployment._
+import pl.touk.nussknacker.engine.management.periodic.db.PeriodicProcessesRepository
+
+import scala.concurrent.Future
+
+trait PeriodicCustomActionsProviderFactory {
```
gskrobisz, line 8: This pattern could be reused for other `DeploymentManager`s. What do you think about renaming/moving it to `pl.touk.nussknacker.engine.management.CustomActionsProviderFactory` ?

gadomsky: But, the point is to have access to `PeriodicProcessesRepository` and `PeriodicProcessService`, cause some implementations of custom actions may need that

gskrobisz (commented):
LGTM

arkadius (approved):

PR #2454, opened 2021-11-09
Feature: Serve fe submoduless app from akka

lciolecki:


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
```diff
@@ -1,36 +1,73 @@
...
-  }
+  private val mainFilesCache = new DefaultCache[String, File](CacheConfig())
+
+  private val defaultMainFilePath = "web/static/main.html"
 
   val route: Route =
-    pathPrefix("static") {
+    pathPrefix("submodules" / Segment /) { moduleName =>
+      (get & submodule(moduleName)) { _ =>
+        extractRequest { matched =>
```
mproch, line 22: Why can't we use getFromResourceDirectory? Do we really need index.html in submodules?

lciolecki: I've changed it, right now we don't wan't so complex routing..

PR #2457, opened 2021-11-09
Get output schema from properties

Filemon279:


engine/standalone/engine/src/main/scala/pl/touk/nussknacker/engine/standalone/StandaloneScenarioEngine.scala
```diff
@@ -101,15 +103,20 @@ object StandaloneScenarioEngine {
...
-    * I think we should generate openApi response definition for 'sinks with schema' (to be done) only.
-    * That way we can ensure that we are generating expected response.
-    * */
+    * TODO : move inputSchema and outputSchema to one place
+    * It is better to have both schemas in one place (properties or some new/custom place)
+    *  */
+    def getSchemaOutputProperty: Json = {
+      context.jobData.metaData.additionalFields.flatMap(_.properties.get(OutputSchemaProperty)) match {
+        case None => Map("type" -> "object".asJson, "properties" -> Json.Null).asJson
+        case Some(outputSchemaStr) => CirceUtil.decodeJsonUnsafe[Json](outputSchemaStr, "Provided json is not valid")
```
mproch, line 112: Can we validate it's correct schema?

Filemon279: We could but it is not the place we want to handle it.
I have planned to handle it in the sink - create schema (check if its valid) and validate sink output expression to match created schema.

mproch (approved):

PR #2459, opened 2021-11-09
Kafka components unflinkization next step

arkadius:


build.sbt
```diff
@@ -827,7 +827,7 @@ lazy val baseEngineComponents = (project in engine("base/components")).
   settings(assemblySampleSettings("baseEngineComponents.jar"): _*).
   settings(
     name := "nussknacker-baseengine-components",
-  ).dependsOn(baseEngineApi)
+  ).dependsOn(baseEngineApi, avroUtil)
```
mproch, line 830: doesn't seem to be needed? :)

arkadius: it is part of longer story :D (forgot to setup Draft)

api/src/main/scala/pl/touk/nussknacker/engine/api/process/Source.scala
```diff
@@ -43,12 +47,18 @@ trait SourceFactory[+T] extends Serializable with Component
 
 object SourceFactory {
 
-  def noParam[T: ClassTag](source: Source[T]): SourceFactory[T] =
-    new NoParamSourceFactory[T](source)
+  def noParam[T: ClassTag](source: Source[T], inputType: TypingResult): SourceFactory[T] =
```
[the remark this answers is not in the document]
arkadius: good point, wasn't before my change, but I can clean it

mproch (approved):
:+1:

PR #2463, opened 2021-11-10
Register Kafka metrics in Kafka engine

mproch:


arkadius (approved):

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/dropwizard/DropwizardMetricsProviderForScenario.scala
```diff
@@ -17,36 +17,39 @@ class DropwizardMetricsProviderForScenario(scenarioId: String, metricRegistry: M
...
   override def histogram(metricIdentifier: MetricIdentifier, instantTimerWindowInSeconds: Long = 10): Histogram = {
-    val histogram = register(metricIdentifier, new metrics5.Histogram(new SlidingTimeWindowReservoir(instantTimerWindowInSeconds, TimeUnit.SECONDS)))
+    val reservoir = new SlidingTimeWindowReservoir(instantTimerWindowInSeconds, TimeUnit.SECONDS)
+    val histogram = register(metricIdentifier, new metrics5.Histogram(reservoir), acceptConflicts = true)
     histogram.update _
   }
 
   //we want to be safe in concurrent conditions...
-  def register[T <: Metric](id: MetricIdentifier, metric: T): T = {
+  private def register[T <: Metric](id: MetricIdentifier, metric: T, acceptConflicts: Boolean): T = {
```
arkadius, line 37: I wonder if it will be better to call it `reuseIfExisting`? Because now I need to think a while why there is "conflict" that is acceptable for counters. Do this solution has some unwritten assumptions? It will work in case if we have two concurrent processes updating the same counter?

arkadius (changes requested):

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/metrics/dropwizard/DropwizardMetricsProviderForScenario.scala
mproch: :+1: changed name, added a comment

arkadius (approved):

PR #2474, opened 2021-11-12
Handle errors on branches after split independently

gadomsky:


arkadius (approved):
LGTM but let's @mproch take a look too

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -220,29 +210,38 @@ object Interpreter {
...
+  }
+
   //Interpreter can be invoked with various effects, we require MonadError capabilities and ability to convert service invocation results
   trait InterpreterShape[F[_]] {
 
-    def monadError: MonadError[F, Throwable]
+    def monad: Monad[F]
 
-    def fromFuture[T](implicit ec: ExecutionContext): Future[T] => F[T]
+    def fromFuture[T, E](handleError: Throwable => E)(implicit ec: ExecutionContext): Future[T] => F[Either[T, E]]
```
mproch, line 225: Can we just make 
`fromFuture[T](implicit ec:ExecutionContext): F[Either[T, Throwable]]` and do `transform` as `map` on F?

mproch (approved):
:+1: please add also an entry in the changelog?

mproch (approved):

PR #2481, opened 2021-11-15
Fix: component's config icon url normalization

lciolecki:


ui/client/components/toolbars/creator/ComponentIcon.tsx
```diff
@@ -17,13 +17,14 @@ export const getComponentIconSrc = memoize((node: NodeType, processDefinitionDat
   if (node) {
     const nodeComponentId = ProcessUtils.findNodeConfigName(node)
     const componentConfig = processDefinitionData.componentsConfig?.[nodeComponentId]
-    const iconFromConfig = componentConfig?.icon
-    const iconBasedOnType = node.type && `${node.type}.svg`
+
+    const iconFromConfig = componentConfig?.icon && absoluteBePath(componentConfig.icon)
+    const iconBasedOnType = node.type && absoluteBePath(`/assets/components/${node.type}.svg`)
```
JulianWielga, line 22: why call `absoluteBePath` twice?

JulianWielga (changes requested):

docs/MigrationGuide.md
```diff
@@ -30,6 +30,7 @@ To see the biggest differences please consult the [changelog](Changelog.md).
     * `nodesToAdd` was renamed to `componentGroups`
     * `posibleNode` was renamed to `components`
     * `nodesConfig` was renamed to `componentsConfig`
+    *  config `icon` property from `componentsConfig` right now should contain relative path to icon e.g. `/assets/components/Filter.svg` (before was just `Filter.svg`) or url (with `http` / `https`)
```
arkadius, line 33: Should be "relatvie to ..." base path? Because before was more relative than it is now. Now is more "absolute" than   before ;-) Also why u want to change it? Someone would want to use other location of icon?

ui/client/components/toolbars/creator/ComponentIcon.tsx
lciolecki: It doesn't matter.. It will be done once.. But, ok I can do a little change calling absolutePath once. IMO right now is clearer and there is less code.

docs/MigrationGuide.md
lciolecki: Hmm, you're right - thanks ;) We want to have one convention to declare path for urls / icons / images.. In other places we use base path with `/assets/*` or just url with `http` / `https`.

`DefaultComponentService`  uses to figure icon's url two components: 

- `componentsUiConfig` it uses relative path e.g. `Filter.svg`
-  `DefaultsComponentIcon` it uses  absolute path `/assets/components/Filter.svg`

Let's see https://staging.nussknacker.io/api/component, right now components api provides three standards of declaration icons: relative, absolute and url.. I want normalize it and use only absolute and url.

lciolecki: And right now if you provide in componentsUiConfig in icon full url, e.g. `http://nussknacker.io/icons/SomeIcon.svg` it will be not work, beacuse of `/assets/components/${icon}`.

arkadius: now description is almost ok, but ... by "base path" someone could think that it should contain also something what we have configured in `http.publicPath` configuration property. After talk with @JulianWielga , he want in the future have this `publicPath` included in all urls returned from BE (in configuration of icon it probable it shouldn't appear because it is already in `publicPath.publicPath`). Reassuming my first proposal that it should be "realtive to " is IMO better then "contains base path". But maybe not "relative to base path" but "relative to http.publicPath". WDYT?

lciolecki: I don't have any idea how to do it in understanding way, so I will agree :)

JulianWielga (approved):

arkadius (approved):

PR #2486, opened 2021-11-15
ContextIdGenerator

arkadius:


engine/base/api/src/main/scala/pl/touk/nussknacker/engine/baseengine/api/customComponentTypes.scala
```diff
@@ -20,7 +21,8 @@ object customComponentTypes {
     def transform[From[_]](implicit tag: TypeTag[From[Any]]): ValidatedNel[ProcessCompilationError, From ~> Target]
   }
 
-  case class CustomComponentContext[F[_]](nodeId: String, interpreter: LazyParameterInterpreter, capabilityTransformer: CapabilityTransformer[F])
+  // Maybe we should open components other then services (sources, custom nodes, sinks) instead of passing  ContextIdGenerator here?
```
mproch, line 24: :+1: I think we should consider going in that direction...

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/ScenarioInterpreterFactory.scala
```diff
@@ -171,7 +171,8 @@ object ScenarioInterpreterFactory {
     private def compileWithCompilationErrors(node: SplittedNode[_], validationContext: ValidationContext): ValidatedNel[ProcessCompilationError, Node] =
       processCompilerData.subPartCompiler.compile(node, validationContext)(compiledProcess.metaData).result
 
-    private def customComponentContext(nodeId: String) = CustomComponentContext[F](nodeId, lazyParameterInterpreter, capabilityTransformer)
+    private def customComponentContext(nodeId: String) = CustomComponentContext[F](nodeId, lazyParameterInterpreter, capabilityTransformer,
```
mproch, line 174: :/ let's add at least a TODO to remove this duplication of creating IdGenerator

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/RichLifecycleFunction.scala
```diff
@@ -0,0 +1,30 @@
+package pl.touk.nussknacker.engine.flink.api.process
+
+import org.apache.flink.api.common.functions.{AbstractRichFunction, MapFunction, RuntimeContext}
+import org.apache.flink.configuration.Configuration
+import pl.touk.nussknacker.engine.api.Lifecycle
+import pl.touk.nussknacker.engine.api.runtimecontext.EngineRuntimeContext
+
+abstract class RichLifecycleFunction extends AbstractRichFunction {
```
mproch, line 8: Do we use it ATM?

engine/base/api/src/main/scala/pl/touk/nussknacker/engine/baseengine/api/runtimecontext/EngineRuntimeContextPreparer.scala
```diff
@@ -19,5 +19,9 @@ class EngineRuntimeContextPreparer(metricRegistryForScenario: String => MetricsP
 
 case class BaseEngineRuntimeContext(jobData: JobData,
                                     metricsProvider: MetricsProviderForScenario with AutoCloseable) extends EngineRuntimeContext with AutoCloseable {
+
+  override def contextIdGenerator(nodeId: String): ContextIdGenerator = new IncContextIdGenerator(jobData.metaData.id + "-" + nodeId)
```
mproch, line 23: I wonder, maybe we can have a constructor for IncContextIdGenerator which takes JobData & nodeId?

engine/base/runtime/src/main/scala/pl/touk/nussknacker/engine/baseengine/ScenarioInterpreterFactory.scala
arkadius: added

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/RichLifecycleFunction.scala
arkadius: no, I can add it in next change when it will be needed (wanted to have in those change only functional changes)

engine/base/api/src/main/scala/pl/touk/nussknacker/engine/baseengine/api/runtimecontext/EngineRuntimeContextPreparer.scala
arkadius: Added factory method `withProcessIdNodeIdPrefix` - don't want to have `apply` because it is not the only one (trivial) option for this prefix. For flink it can confuse someone and can choose the bad one option (without subtask id)

mproch (approved):
:+1: pls, add sth for ChangeLog/MigrationGuide?

PR #2489, opened 2021-11-16
finalize components list

JulianWielga:


piotrp:
Review for attached screenshot:
1. We shouldn't have column titles that don't fit, maybe let's rename "Usage c..." to something else (Uses? Use count?) and/or make it wider?
2. Type column shows which panel group component belongs to, maybe this should be renamed to "Group"? Plural "type" seems strange, but maybe it's just me

ui/client/components/toolbars/creator/ComponentIcon.tsx
```diff
@@ -13,12 +13,20 @@ export const preloadImage = memoize((href: string) => new Promise<string>(resolv
   return image.onload = () => resolve(href)
 }))
 
-export const getComponentIconSrc = memoize((node: NodeType, processDefinitionData: ProcessDefinitionData) => {
-  if (node) {
-    const nodeComponentId = ProcessUtils.findNodeConfigName(node)
-    const componentConfig = processDefinitionData.componentsConfig?.[nodeComponentId]
+function _getComponentIconSrc(nodeOrPath: string): string
```
lciolecki, line 16: Why you change it? Using `/api/components` you should use only `absoluteBePath(icon)`. Right now it's to complicated.. :/

lciolecki:
1. I think right now fonts are a little bit different? Don't you think so?
2. Right now it's good moment to change module name from `tools` to `components`, let's do it.

lciolecki (approved):
LGTM

PR #2490, opened 2021-11-16
KafkaSourceFactory unflinkization

arkadius:


engine/flink/kafka-util/src/main/resources/META-INF/services/pl.touk.nussknacker.engine.util.json.ToJsonEncoder
```diff
@@ -1 +1 @@
-pl.touk.nussknacker.engine.kafka.source.flink.InputMetaToJson
```
mproch, line 1: Let's move also this file to generic kafka-util

mproch (commented):
:+1: cool, looks good, added only one minor remark + I think we'll need changelog/migration guide

mproch (approved):

arkadius:
> +1 cool, looks good, added only one minor remark + I think we'll need changelog/migration guide

Added something ... I hope that it will help someone

PR #2496, opened 2021-11-18
KafkaAvroSourceFactory and KafkaAvroSinkFactory unflinkization

arkadius:


engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -76,15 +77,15 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
       "monitor" -> categories(SinkFactory.noParam(EmptySink)),
       "communicationSink" -> categories(DynamicParametersSink),
       "kafka-string" -> all(new KafkaSinkFactory(new SimpleSerializationSchema[AnyRef](_, String.valueOf), processObjectDependencies) with FlinkKafkaSinkFactory),
-      "kafka-avro" -> all(new KafkaAvroSinkFactoryWithEditor(schemaRegistryProvider, processObjectDependencies))
+      "kafka-avro" -> all(new BaseKafkaAvroSinkFactoryWithEditor(schemaRegistryProvider, processObjectDependencies) with FlinkKafkaAvroSinkFactory)
```
mproch, line 80: We have it (BaseKafka... with FlinkKafkaAvro...) in a few places. Can't we have a class (non-anonymous) for that?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import pl.touk.nussknacker.engine.api.LazyParameter
+import pl.touk.nussknacker.engine.api.process.{Sink, SinkFactory}
+import pl.touk.nussknacker.engine.avro.RuntimeSchemaData
+import pl.touk.nussknacker.engine.avro.encode.ValidationMode
+import pl.touk.nussknacker.engine.kafka.serialization.KafkaSerializationSchema
+import pl.touk.nussknacker.engine.kafka.{KafkaConfig, PreparedKafkaTopic}
+import pl.touk.nussknacker.engine.util.KeyedValue
+
+trait KafkaAvroSinkFactory extends SinkFactory {
```
mproch, line 11: I wonder, do we need inheritance here? It seems that everything is passed in arguments?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/KafkaAvroSourceFactory.scala
```diff
@@ -118,19 +114,20 @@ class KafkaAvroSourceFactory[K: ClassTag, V: ClassTag](val schemaRegistryProvide
...
 
     // - avro payload formatter requires to format test data with writer schema, id of writer schema comes with event
     // - for json payload event does not come with writer schema id
-    val formatterSchema = schemaRegistryProvider.deserializationSchemaFactory.create[K, V](kafkaConfig, None, None)
-    val recordFormatter = schemaRegistryProvider.recordFormatterFactory.create[K, V](kafkaConfig, formatterSchema)
+    val formatterSchema = schemaRegistryProvider.deserializationSchemaFactory.create[K, V](kafkaConfig, None, None)(keyClassTag, valueClassTag)
+    val recordFormatter = schemaRegistryProvider.recordFormatterFactory.create[K, V](kafkaConfig, formatterSchema)(keyClassTag, valueClassTag)
 
-    createSource(params, dependencies, finalState, List(preparedTopic), kafkaConfig, deserializationSchema, timestampAssigner, recordFormatter, kafkaContextInitializer)
+    createSource(params, dependencies, finalState, List(preparedTopic), kafkaConfig, deserializationSchema, recordFormatter, kafkaContextInitializer)
```
mproch, line 130: I wonder, do we need inheritance here? It seems that everything is passed in arguments?

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
arkadius: IMO current solution causes that engine specific implementation is clearly separated. Developer of next engine-specific component using similar solution won't use engine specific things in places where it is not necessary. It is very close to idea to keep definition separated from imlementation. When we add  BaseFlinkKafkaAvroSinkFactoryWithEditor it will cause that it will be a bag for "everything". I'd rather want to do the same separation in sources - but haven't done it because we have `timestampAssigner` in constructor.

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
arkadius: I don't understand. `KafkaAvroSinkFactory extends SinkFactory` just to not add `extends SinkFactory` in all concrete implementations. How u would like to it in other way?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/KafkaAvroSourceFactory.scala
arkadius: You mean by replacing it with composition? Yes, I am a great fun of this. But have no idea how to make it not hurting too much during migration. New parameter like `KafkaAvroSourceImplementationFactory`? It looks weird because this new extracted class is a real factory and all the code in `KafkaAvroSourceFactory` is kind of `KafkaAvroSourceFactoryDefinitionAndImplementationFacade` :/. Any idea?

engine/flink/management/sample/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
mproch: ok, when I looked further I realised I just want composition - which we discuss in other place ;)

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/source/KafkaAvroSourceFactory.scala
mproch: Yes, sth like ImplementationFactory, the same with Sink (see comment in KafkaAvroSinkFactory). I think we made so many changes that one more won't hurt that much. What's more, I think that Avro components are less frequently customised/changed than 'generic' (where e.g. deployments use their own model)

arkadius: The true benefit was only when I've change it in both KafkaAvroSourceFactory and KafkaSourceFactory - changed. A lot of duplicated code disappear - take a look. Tomorrow I will do the same with Sinks

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactory.scala
arkadius: Replaced by composition - like in sources case

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactoryImpl.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import pl.touk.nussknacker.engine.api.LazyParameter
+import pl.touk.nussknacker.engine.api.process.Sink
+import pl.touk.nussknacker.engine.avro.RuntimeSchemaData
+import pl.touk.nussknacker.engine.avro.encode.ValidationMode
+import pl.touk.nussknacker.engine.kafka.serialization.KafkaSerializationSchema
+import pl.touk.nussknacker.engine.kafka.{KafkaConfig, PreparedKafkaTopic}
+import pl.touk.nussknacker.engine.util.KeyedValue
+
+trait KafkaAvroSinkFactoryImpl {
```
mproch, line 11: I think it's more KafkaAvroSinkImplementationFactory :) (also in other places)

mproch (commented):
:+1: good job, I added one naming comment, also - I think we need sth in migration/changelog?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/sink/KafkaAvroSinkFactoryImpl.scala
arkadius: good point - fixed

mproch (approved):

dswiecki (approved):

PR #2499, opened 2021-11-18
Remove exception handler Part 2 (remove exception handler ref)

nadberezny:


mproch (approved):
:+1: But please correct UI tests :)

mproch (approved):

PR #2501, opened 2021-11-18
Lite basic model, basic streaming lite components

mproch:


ui/client/cypress/integration/components.ts
```diff
@@ -68,11 +68,12 @@ describe("Components list", () => {
     cy.get("@rows").filter(`:contains("Server")`).should("have.lengthOf", 1)
   })
 
-  it("should allow filtering by usage", () => {
+    /*
+  test("should allow filtering by usage", () => {
```
dswiecki, line 72: Do I  understand correctly ? Do we want to rewrite this test to be less backed dependent ? But for its disabled/commented out

dswiecki (approved):

ui/client/cypress/integration/components.ts
mproch: yes, I'd like to rewrite it, but for now I'd rather skip it...

maciej-brzezinski (approved):

PR #2543, opened 2021-11-24
Injecting helpers as additional variables

mproch:


api/src/main/java/pl/touk/nussknacker/engine/api/AdditionalVariable.java
```diff
@@ -12,4 +12,6 @@
     String name();
 
     Class clazz();
+
+    boolean initializeValue() default false;
```
arkadius, line 16: by `initialize value` you mean that initialization of value will be done by Nussknacker or user should initialize value in expression? maybe a little bit more descriptive name showing who "initialize"? Maybe "initializedByRuntime"?

api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -105,3 +106,19 @@ case class Parameter(name: String,
...
+  def typingResult: TypingResult
+}
+
+object AdditionalVariableProvidedInRuntime {
+  def apply[T:TypeTag]: AdditionalVariableProvidedInRuntime = AdditionalVariableProvidedInRuntime(Typed.fromDetailedType[T])
+}
+
+case class AdditionalVariableProvidedInRuntime(typingResult: TypingResult) extends AdditionalVariable
+
+object AdditionalVariableWithFixedValue {
```
arkadius, line 120: Like it :+1:

api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -105,3 +106,19 @@ case class Parameter(name: String,
...
+
+sealed trait AdditionalVariable {
+  def typingResult: TypingResult
+}
+
+object AdditionalVariableProvidedInRuntime {
+  def apply[T:TypeTag]: AdditionalVariableProvidedInRuntime = AdditionalVariableProvidedInRuntime(Typed.fromDetailedType[T])
+}
+
+case class AdditionalVariableProvidedInRuntime(typingResult: TypingResult) extends AdditionalVariable
```
arkadius, line 118: IMO "fixed value" below is also "provided in runtime" when it will be used in lazy parameter. Maybe "provided by component" hmm...

api/src/main/java/pl/touk/nussknacker/engine/api/AdditionalVariable.java
mproch: :+1: done

api/src/main/java/pl/touk/nussknacker/engine/api/AdditionalVariable.java
```diff
@@ -12,4 +12,6 @@
     String name();
 
     Class clazz();
+
+    boolean initializedByRuntime() default false;
```
arkadius, line 16: I see problems with this name of parameter:
1. Consider situation: someone provide additional variable "#MATH` and value of parameter `#MATH.pow(2, 2)` and some dependent parameters based on this value. How it is initialized by runtime? It will be initialized on designer side during validation and in runtime during compilation.
2. AdditionalVariableProvided*InRuntime* is opposite to `initializedByRuntime = true` and name is quite similar.
Maybe we should name this `initializedDuringCompilation`? Any other proposal?

mproch: I think we don't need this flag at all :) If the parameter is lazy, then it's always false, if it's eager - it's always true...

arkadius: good point, let's change it that way - at the end the whole eager parameter is "initialized during compilation" ;)

arkadius (approved):

PR #2547, opened 2021-11-26
Lite kafka runtime directory cleanups

arkadius:


mproch (approved):

dswiecki (approved):

engine/base/kafka/integration-test/src/it/scala/pl/touk/nussknacker/engine/baseengine/kafka/NuKafkaRuntimeBinTest.scala
```diff
@@ -57,6 +57,8 @@ class NuKafkaRuntimeBinTest extends FunSuite with KafkaSpec with Matchers with L
         """{"foo":"ping"}""".stripMargin
       kafkaClient.sendMessage(inputTopic, input).futureValue
 
+      // we check if future not completed with failure instantly to not shadow true failure by consume timeout
+      runtimeExitCodeFuture.value.flatMap(_.failed.toOption)  shouldBe empty
```
mproch, line 61: I wonder, maybe `runtimeExitCodeFuture.onFailure { _ => throw new AssertionError } ` would be more reliable? Because in the line above it may still theoretically happen that runtime fails afterward?

PR #2554, opened 2021-11-26
Flink components cleanup + fixes in dev (staging) setup

arkadius:


build.sbt
```diff
@@ -825,14 +825,14 @@ lazy val liteEngineApi = (project in lite("api")).
 
 lazy val liteBaseComponents = (project in lite("components/base")).
   settings(commonSettings).
-  settings(assemblyNoScala("liteBaseComponents.jar"): _*).
+  settings(assemblyNoScala("base.jar"): _*).
```
mproch, line 828: maybe it'll be safer to leave lite/flink in names? I know it'll go to different dirs but still it can be misleading?

arkadius: In dist build ? you mean `components/flink/flinkBase.jar` ?

mproch (approved):

PR #2563, opened 2021-11-29
Improvemetns: Disable override component id for base and fragment com…

lciolecki:
…ponents

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentIdProvider.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentIdProvider.scala
@@ -13,15 +13,15 @@ trait ComponentIdProvider {
 }
 
 class DefaultComponentIdProvider(configs: Map[ProcessingType, ComponentsUiConfig]) extends ComponentIdProvider {
-  override def createComponentId(processingType: ProcessingType, name: String, componentType: ComponentType): ComponentId =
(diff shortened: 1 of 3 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentIdProvider.scala
```diff
@@ -13,15 +13,15 @@ trait ComponentIdProvider {
...
-      ComponentId.forBaseComponent(componentType)
-    } else {
-      val defaultComponentId = ComponentId.default(processingType, name, componentType)
-      val overriddenComponentId = getOverriddenComponentId(processingType, name, defaultComponentId)
-      overriddenComponentId
-    }
+  override def createComponentId(processingType: ProcessingType, name: String, componentType: ComponentType): ComponentId = {
+    val defaultComponentId = ComponentId.default(processingType, name, componentType)
+
+    //We assume that base nad fragment component's id can't be overridden
```
mproch, line 19: I think we should add 'currently' for fragments. It's just an implementation limitation, not sth we're doing by design? What's more - maybe we should throw some exception/warning if use tries to do it?

lciolecki: I'v added throwing exception.

mproch (approved):

PR #2564, opened 2021-11-29
Refactor union part 1 - flink union

dswiecki:
Squash


docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,21 +98,25 @@ There can be at most one edge of type `Default`, and it gets all records that do
...
 
-Union node defines new stream which is union of all branches. In this new stream there is only one variable; it's name is defined by 'Output' parameter; it's value is: 
+Union node defines new stream which is union of all branches. It's name is defined by 'Output' parameter; it's value is: 
 ```$json
 {
-  "key": `value of key expression for given event`,
-  "branch1": `value expression when event comes from branch1, otherwise null`,
-  "branch2": `value expression when event comes from branch2, otherwise null`,
-  ...
+  "unionOutput": `value expression`,
```
arkadius, line 107: Nope - it doesn't. It is rather:
```
{
  field1: field1 value in value expression
(...)
  fieldN: fieldN value in value expression
}
```
Please reformat it on your way

engine/flink/components/base/src/it/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformerSpec.scala
```diff
@@ -62,10 +62,35 @@ class UnionTransformerSpec extends FunSuite with BeforeAndAfterAll with Matchers
...
+            ),
+            BranchBarId -> List(
+              "value" -> "{a: '123'}"
+            )
+          )
+        )
+        .processorEnd("end", "mockService", "all" -> s"#$OutVariableName.a")
+    ))
+
+    val data = List("10")
```
mproch, line 85: can we have a bit larger test sample? ;)

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformer.scala
```diff
@@ -1,55 +1,44 @@
...
-      val newType = TypedObjectTypingResult(
-        (UnionTransformer.KeyField -> Typed[String]) :: inputContexts.map {
-          case (branchId, _) =>
-            ContextTransformation.sanitizeBranchName(branchId) -> valueByBranchId(branchId).returnType
-        }.toList
-      )
-      ValidationContext(Map(variableName -> newType), Map.empty, parent)
+      ValidationContext(Map(variableName -> branchReturnTypes.head), Map.empty, parent)
+    }.andThen { vc =>
+      Validated.cond(branchReturnTypes.toSet.size == 1, vc, CannotCreateObjectError("All branch values must be of the same type", nodeId.id)).toValidatedNel
```
mproch, line 27: Can we have some validation test, checking if this works ok? I also wonder - what about cases like `.canBeSubclassOf`? Is equality always enough?

engine/flink/components/base/src/it/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformerSpec.scala
mproch, line 85: Those union tests are a bit simplistic (testing only one branch?) Maybe we can make `BaseSampleConfigCreator` accept a `Map[SourceId, Data]` and use it to test multiple sources?

mproch (commented):
Added minor comments

engine/flink/components/base/src/it/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformerSpec.scala
```diff
@@ -35,37 +35,117 @@ class UnionTransformerSpec extends FunSuite with BeforeAndAfterAll with Matchers
...
+    val scenario = EspProcess(MetaData("sample-union", StreamMetaData()), NonEmptyList.of[SourceNode](
+      GraphBuilder.source("start-foo", "source")
+        .branchEnd(BranchFooId, UnionNodeId),
+      GraphBuilder.source("start-bar", "source")
+        .branchEnd(BranchBarId, UnionNodeId),
+      GraphBuilder
+        .branch(UnionNodeId, "union", Some(OutVariableName),
+          List(
+            BranchFooId -> List("value" -> "#CONV.toNumber(#input).intValue"),
+            BranchBarId -> List("value" -> "#CONV.toNumber(#input).intValue / (#CONV.toNumber(#input).intValue % 4)")
```
mproch, line 134: let's use sth like `#input.length - 2` (or input == 0) so we can have simpler setup, without CONF

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,21 +98,18 @@ There can be at most one edge of type `Default`, and it gets all records that do
...
 
-Union merges multiple branches into one stream. For each incoming branch two parameters are configured:
-- key - it's value should be of type `String`, definex how elements from branches will be matched together
-- value - this is the output value which will be put the field with name the same as branch id
+Union merges multiple branches into one stream. For each incoming branch only one parameter is configured:
+- value - this is the output value which will be put in the result steam
 
-Union node defines new stream which is union of all branches. In this new stream there is only one variable; it's name is defined by 'Output' parameter; it's value is: 
+Union node defines new stream which is union of all branches. This new steam name is defined by 'Output' parameter;
+If you want to distinguish between branches in output stream you can do this using e.g. map as value of branch
```
arkadius, line 105: It should be rather "If you want to distinguish between branches in output stream you can e.g. add extra field in value map. Value expression will then look like for branch1: 
```$json
{ branch: "branch1", unifiedField1: <expression>, unifiedFieldN: <expression> }
```
and for branch2:
```$json
{ branch: "branch2", unifiedField1: <expression>, unifiedFieldN: <expression> }
```
After this, you can check the branch using e.g. `#outputVarible.branch == 'branch1'` expression

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformer.scala
dswiecki: I've added test

engine/flink/components/base/src/it/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformerSpec.scala
dswiecki: removed conv dependency, done it other way

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,27 +98,45 @@ There can be at most one edge of type `Default`, and it gets all records that do
 
 ![union_window](img/union_window.png)
 
-Union merges multiple branches into one stream. For each incoming branch two parameters are configured:
-- key - it's value should be of type `String`, definex how elements from branches will be matched together
-- value - this is the output value which will be put the field with name the same as branch id
+Union merges multiple branches into one branch. Events from the incoming branch are passed to the output branch without an attempt to combine them or match them.
```
ZbyszekMM, line 101: ... combine them or match them --> combine or match them.

Get 3rd opinion on which version sounds better

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,27 +98,45 @@ There can be at most one edge of type `Default`, and it gets all records that do
...
+- Output Variable Name - the name of the variable containing results of the merge.
+- Output Expression - the value of this expression will be passed to the output branch. The output expression is defined separately for each input branch.
+
+Please note, that the #input variable used in the Output expression field refers to the content of the respective incoming branch.
+
+## UnionMemo
+![union_memo_window](img/union_memo_window.png)
+
+Works similarly to Union, but also memoize values for each branches. Memoized state will be cleared when there was no
+subsequent events from any branch during `stateTimeout`.
```
ZbyszekMM, line 120: There are 2 key differences from the Union (memoization and merge logic) -> I would not say that UnionMemo is similar to Union. 

_Similarly to Union, UnionMemo node merges branches into one branch. There are however important differences in the way UnionMemo works:

- events from the incoming branches are matched together based on some key value
- data which arrived from any of the incoming branch will be memoized by the UnionMemo node for time duration defined in stateTimeout.
- events are emitted after the memoization time (stateTimeout) elapses (????) [on every incoming event ?????\]
- the event time for the emitted events is set to ...........??????
 

-

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,27 +98,45 @@ There can be at most one edge of type `Default`, and it gets all records that do
...
+
+Example:
+![union_memo_example](img/union_memo_example.png)
+
+
+UnionMemo merges multiple branches into one stream. For each incoming branch two parameters are configured:
+- key - it's value should be of type `String`, defines how elements from branches will be matched together
+- value - the value of this expression which will be put the field with name the same as branch id
+
+Union node defines new stream which is union of all branches. In this new stream there is only one variable; it's name is defined by 'Output variable name' parameter; it's value is:
```
ZbyszekMM, line 130: ---> I wouild remove sentence "Union (memo) node defines ... ; by now it should be clear. Next sentence would run something like _#input variable is no longer available downstream the UnionMemo, a new variable whose name is defined by "Output variable name' parameter will be available instead:_

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -98,27 +98,45 @@ There can be at most one edge of type `Default`, and it gets all records that do
...
 ```$json
 {
   "key": `value of key expression for given event`,
   "branch1": `value expression when event comes from branch1, otherwise null`,
   "branch2": `value expression when event comes from branch2, otherwise null`,
+  "branch3": `value expression when event comes from branch3, otherwise null`,
```
ZbyszekMM, line 136: What if all branches received an event for the given key in the "stateTimeout" period? Is it correct to state the following:

...
"branchN": value of the expression configured for this branch if event for the given key arrived from this branch in the stateTimeout period; otherwise null". 

The above means that the there will be between 1 and N  "branch", "value" pairs in the output event where N denotes number of branches merged by the unionMemo node.

docs/MigrationGuide.md
```diff
@@ -35,6 +35,10 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   It means that `LiteSource` also takes this parameter and have a new method `createTransformation`.
 * [#2635](https://github.com/TouK/nussknacker/pull/2635) `ContextInitializer.initContext` now takes `ContextIdGenerator` instead of `nodeId` and returns just a function
   with strategy of context initialization instead of serializable function with `Lifecycle`. To use it with Flink engine, use `FlinkContextInitializingFunction` wrapper.
+* [#2564](https://github.com/TouK/nussknacker/pull/2564/files) Flink union now takes only 'Output expression' parameters for branches (previously 'value' parameter), output variable must be of the same type,
```
arkadius, line 38: Please extract new section `Scenario authoring changes` and split this entry into two parts: for users and for developers (about `UnionParametersMigration`)

docs/scenarios_authoring/BasicNodes.md
dswiecki: done

docs/scenarios_authoring/BasicNodes.md
dswiecki: done

docs/scenarios_authoring/BasicNodes.md
dswiecki: done

docs/MigrationGuide.md
dswiecki: done

arkadius (approved):
LGTM

PR #2574, opened 2021-11-30
Lite Kafka runtime - integration test for avro

arkadius:


engine/lite/kafka/integration-test/src/it/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeDockerAvroTest.scala
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  test("avro ping-pong should work") {
+    val schemaRegistryClient = new CachedSchemaRegistryClient(mappedSchemaRegistryAddress, 10)
+    val parsedAvroSchema = ConfluentUtils.convertToAvroSchema(NuKafkaRuntimeTestSamples.avroPingSchema)
+    val inputSchemaId = schemaRegistryClient.register(ConfluentUtils.valueSubject(fixture.inputTopic), parsedAvroSchema)
+    val outputSchemaId = schemaRegistryClient.register(ConfluentUtils.valueSubject(fixture.outputTopic), parsedAvroSchema)
+    val valueBytes = ConfluentUtils.serializeRecordToBytesArray(NuKafkaRuntimeTestSamples.avroPingRecord, inputSchemaId)
+    kafkaClient.sendRawMessage(fixture.inputTopic, "fooKey".getBytes, valueBytes).futureValue
+    consumeFirstError shouldBe empty
```
mproch, line 50: so, it basically waits for 5 secs? I wonder, maybe it would be better to move it after consuming messages and wait shorter (1s?) ? When there is a problem the test will take longer to fail, but it will be shorter in normal circumstances?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentUtils.scala
```diff
@@ -56,4 +59,28 @@ object ConfluentUtils extends LazyLogging {
       .parsePayloadToByteBuffer(bytes)
       .valueOr(exc => throw new SerializationException(exc.getMessage, exc))
       .getInt
+
+  def serializeRecordToBytesArray(record: GenericRecord, schemaId: Int): Array[Byte] = {
```
mproch, line 63: I wonder, can't we use AbstractConfluentKafkaAvroSerializer somehow and not duplicate this low-level stuff?

arkadius: I would like to move in other direction - to stop using `AbstractConfluentKafkaAvroSerializer` and instead of this just use simple methods to serialize/deserialize records. Now its takes too long to figure out how to (de)serialize avro after a while of not being in this subject or for newcomers in this subject. Also currently we already have a part of serializing logic in `ConfluentUtils` - `MagicByte` and `readId`.

engine/lite/kafka/integration-test/src/it/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeDockerAvroTest.scala
arkadius: It won't be so easy - if there will be sth in error, it will be overriden by information that there is timeout waiting on output. I can move it to both `withClue` for checking output and after checking output to won those 5 secs. It won't look too pretty but I've right - time of tests is important

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentUtils.scala
mproch: hmm... oook...

engine/lite/kafka/integration-test/src/it/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeDockerAvroTest.scala
arkadius: ehh `withClue` is not nice unfortunately :-( : 
1. it handles only scaltest exceptions and in case of error we have `TimeoutException` here
2. it evaluates clue eagerly
I've changed 5s -> 1s and added this check in finally block - it is a little bit tricky but don't want to add our own, better `withClue`

mproch (approved):

mproch (approved):

PR #2575, opened 2021-11-30
Components extras

JulianWielga:
- clear name filter button
- set group filter by cell click
- debounced query setting
- elegant loading indicator


ui/submodules/packages/components/src/components/unavailableViewPlaceholder.tsx
```diff
@@ -0,0 +1,17 @@
+import { Link, Typography } from "@mui/material";
+import React from "react";
+import { Link as RouterLink } from "react-router-dom";
+import { useHistory } from "../common";
+
+export function UnavailableViewPlaceholder(): JSX.Element {
+    const history = useHistory();
+    const to = history[history.length - 2] || history[0] || `/`;
```
lciolecki, line 8: What is it?

lciolecki (approved):

ui/submodules/packages/components/src/components/unavailableViewPlaceholder.tsx
JulianWielga: store history and go back to only reasonable locations (not outside) or app root

PR #2578, opened 2021-11-30
Update migration guide

dswiecki:


mproch (approved):
I wonder, maybe we should merge it also to 1.1?

PR #2579, opened 2021-11-30
Metrics unregistration

arkadius:


engine/flink/util/src/main/scala/pl/touk/nussknacker/engine/flink/util/metrics/FlinkMetricsProviderForScenario.scala
```diff
@@ -55,6 +53,10 @@ class FlinkMetricsProviderForScenario(runtimeContext: RuntimeContext) extends Me
     group.histogram(name, histogram)
   }
 
+  override def remove(metricIdentifier: MetricIdentifier): Unit = {
+    // Not needed handling of situation of metrics removal
```
mproch, line 57: Let's throw NotImplementedException or sth, in case sb wants to actually use it ;)

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/KafkaMetricsRegistrar.scala
```diff
@@ -0,0 +1,34 @@
+package pl.touk.nussknacker.engine.lite.kafka
+
+import cats.data.NonEmptyList
+import org.apache.kafka.common.{Metric, MetricName}
+import pl.touk.nussknacker.engine.util.metrics.{Gauge, MetricIdentifier, MetricsProviderForScenario}
+
+import scala.collection.mutable
+import scala.jdk.CollectionConverters.mapAsScalaMapConverter
+
+//We have to pass taskId, as we will need
```
mproch, line 10:
```suggestion
//We have to pass taskId, as we need different tags. `metrics` map passed in constructor is mutable (by Kafka), so we 
//create own set of registered metrics to remove them correctly
```

mproch (approved):
Added two minor comments

PR #2581, opened 2021-12-01
Extract common metrics code

mproch:
Count and Rate meters for nodes are handled by a common listener. Exception rate meter is still handled separately with ExceptionHandler in Flink


utils/util/src/main/scala/pl/touk/nussknacker/engine/util/service/TimeMeasuringService.scala
```diff
@@ -20,7 +20,7 @@ trait TimeMeasuringService extends LazyLogging { self: Service =>
   protected def metricName: NonEmptyList[String] = NonEmptyList.of("service")
 
   //TODO: add metrics eagerly during open, so that we don't need this map
-  @transient lazy val metrics : collection.concurrent.TrieMap[String, EspTimer] = collection.concurrent.TrieMap()
+  private val metrics = new SafeLazyMetrics[String, EspTimer]
```
arkadius, line 23: I wonder if we can go one step further and make this `TimeMeasuringService` just a composited `AsyncOperationTimeMeter` or sth  similar because it looks like it could be used with any operation (not only service invokation) - it would make it a little bit easier to understand...

mproch: I did it, but left TimeMeasuringService, as it's more convenient to mix a trait when you have to handle lifecycle anyway...

arkadius (approved):

PR #2582, opened 2021-12-01
lite kafka runtime: auto reset offset turned off

arkadius:


mproch (approved):
:+1: Thx :)

utils/kafka-util/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaUtils.scala
```diff
@@ -89,11 +89,6 @@ object KafkaUtils extends LazyLogging {
     props.setProperty("bootstrap.servers", config.kafkaAddress)
     props.put("key.serializer", classOf[ByteArraySerializer])
     props.put("value.serializer", classOf[ByteArraySerializer])
-    props.setProperty("acks", "all")
```
mproch, line 92: Let's add entry in Changelog/Migration, as someone may rely on this code in deployments...

PR #2586, opened 2021-12-01
Run EmbeddedDeploymentManager in dev mode

mproch:


mproch:
Will be handled by #2567

PR #2600, opened 2021-12-03
Lite source api refactor - Input => Context transformation done by interprerter

arkadius:


mproch (approved):
:+1: Nice, can you also add sth to changelog/migration?

PR #2611, opened 2021-12-06
Fix: Ability to handleExceptionDuringTransformation in GenericNodeTra…

arkadius:
…nsformation with default fallbackFinalResult - it was handled before by initialParameters

### DIFF
--- a/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
+++ b/api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
@@ -2,7 +2,7 @@ package pl.touk.nussknacker.engine.api.context.transformation
 
 import cats.data.ValidatedNel
 import com.typesafe.scalalogging.LazyLogging
-import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.NodeId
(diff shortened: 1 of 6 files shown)


api/src/main/scala/pl/touk/nussknacker/engine/api/context/transformation/GenericNodeTransformation.scala
```diff
@@ -42,7 +42,18 @@ trait GenericNodeTransformation[T] {
 
   // FinalResult which will be used if some TransformationStep won't be handled inside contextTransformation.
   // Especially useful for cases when evaluation of some parameter will fail
-  def fallbackFinalResult(step: TransformationStep, inputContext: InputContext, outputVariable: Option[String])(implicit nodeId: NodeId): FinalResults = {
+  def handleUnmatchedTransformationStep(step: TransformationStep, inputContext: InputContext, outputVariable: Option[String])(implicit nodeId: NodeId): FinalResults = {
```
mproch, line 45: I wonder (it's more a remark for the previous change), maybe if there are no former exceptions (i.e. it's not the case when evaluation of a parameter fails) we should add some error here? Because if no other error occurred, then sth is definitely wrong :)

mproch (approved):

PR #2613, opened 2021-12-06
Lite kafka event timestamp handling

maciej-brzezinski:


engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/KafkaSingleScenarioTaskRun.scala
```diff
@@ -125,9 +125,21 @@ class KafkaSingleScenarioTaskRun(taskId: String,
   }
 
   private def sendOutputToKafka(output: ResultType[interpreterTypes.EndResult[ProducerRecord[Array[Byte], Array[Byte]]]]): Future[_] = {
-    val results = output.value.map(_.result)
+
+    val resultsWithEventTimestamp = output.value.map(endResult => {
+      val contextEventTimestamp = endResult.context.variables(VariableConstants.EventTimestampVariableName).asInstanceOf[Long]
```
mproch, line 130: `endResult.context.get[Long]` - I think we should be prepared for the case where the variable is missing and just leave null (`orElse(contextEventTimestamp).orNull`) in line 135

engine/lite/kafka/api/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/api/LiteKafkaSource.scala
```diff
@@ -9,4 +10,10 @@ trait LiteKafkaSource extends BaseLiteSource[ConsumerRecord[Array[Byte], Array[B
 
   def topics: List[String]
 
+  def transform(record: ConsumerRecord[Array[Byte], Array[Byte]]): Context = {
```
mproch, line 13: what about `LiteKafkaSourceImpl`? I wonder - should it be the single implementation? This addition should be more of a fallback

maciej-brzezinski: you are right, that's why some other tests didn't pass, I added it 
but if I delete the method here i think i need to add it in TestComponentProvider so maybe it should stay here as well?

engine/lite/kafka/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/kafka/KafkaTransactionalScenarioInterpreterTest.scala
```diff
@@ -46,6 +46,28 @@ class KafkaTransactionalScenarioInterpreterTest extends fixture.FunSuite with Ka
...
+    val inputTimestamp = System.currentTimeMillis()
+
+    runScenarioWithoutErrors(fixture, scenario) {
+      val input = "test-input"
+      kafkaClient.sendRawMessage(
+        inputTopic,
+        key = null,
+        content = input.getBytes(),
+        timestamp = inputTimestamp
+      ).futureValue
```
dswiecki, line 63: why do you invoke futureValue ?

engine/lite/kafka/api/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/api/LiteKafkaSource.scala
```diff
@@ -9,4 +10,10 @@ trait LiteKafkaSource extends BaseLiteSource[ConsumerRecord[Array[Byte], Array[B
...
+
+    Context(contextIdGenerator.nextContextId())
```
dswiecki, line 15: why do you add this variable this explicitly and it test impl `super.transform(record)` implicitly ?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/KafkaSingleScenarioTaskRun.scala
maciej-brzezinski: fixed it, thx

engine/lite/kafka/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/kafka/KafkaTransactionalScenarioInterpreterTest.scala
maciej-brzezinski: removed it, it was just in case but turned out unnecessary

mproch (approved):

PR #2625, opened 2021-12-07
Remove remaining Standalone name usages

mproch:


arkadius (approved):
Please update changelog/migrationguide

PR #2633, opened 2021-12-08
Test from file - lite-streaming + EmbeddedDeploymentManager

mproch:


engine/lite/api/src/main/scala/pl/touk/nussknacker/engine/lite/api/utils/sources.scala
```diff
@@ -0,0 +1,39 @@
...
+import pl.touk.nussknacker.engine.api.runtimecontext.{ContextIdGenerator, EngineRuntimeContext}
+import pl.touk.nussknacker.engine.api.{Context, Lifecycle}
+import pl.touk.nussknacker.engine.lite.api.commonTypes.ErrorType
+import pl.touk.nussknacker.engine.lite.api.customComponentTypes
+import pl.touk.nussknacker.engine.lite.api.customComponentTypes.LiteSource
+
+import scala.language.higherKinds
+import scala.util.Try
+
+object sources {
```
arkadius, line 16: Can we use just package and one file in it? Until there is no need to have some package scope scala "type"s, those objects introduce a lot of problems e.g. you can't have two the same objects in the same packages when you split sth between two modules - we have such problems during moving classes between flink and non-flink modules

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaComponentProvider.scala
```diff
@@ -34,10 +34,10 @@ class LiteKafkaComponentProvider extends ComponentProvider {
   override def create(config: Config, dependencies: ProcessObjectDependencies): List[ComponentDefinition] = List(
     ComponentDefinition("kafka-json", new KafkaSinkFactory(GenericJsonSerialization(_), dependencies, LiteKafkaSinkImplFactory)),
     ComponentDefinition("kafka-json", new KafkaSourceFactory[String, java.util.Map[_, _]](
-      ConsumerRecordDeserializationSchemaFactory.fixedValueDeserialization(deserializeToMap), jsonFormatterFactory, dependencies, new LiteKafkaSourceImplFactory)),
```
arkadius, line 37: What impact have this change? Why it is done other way than in `FlinkKafkaComponentProvider`?

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaSourceImplFactory.scala
```diff
@@ -50,4 +51,11 @@ class LiteKafkaSourceImpl[K, V](contextInitializer: ContextInitializer[ConsumerR
     initializerFun(deserialized)
   }
 
+  override def generateTestData(size: Int): Array[Byte] = formatter.generateTestData(topics, size, kafkaConfig)
```
arkadius, line 54: Don't like that we still have it in source implementation instead of factory :-( Because of this I can't see what is the real difference in flink and lite implementations

interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ScenarioParser.scala
```diff
@@ -0,0 +1,19 @@
+package pl.touk.nussknacker.engine.marshall
+
+import cats.data.Validated.{Invalid, Valid}
+import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
+import pl.touk.nussknacker.engine.canonize.ProcessCanonizer
+import pl.touk.nussknacker.engine.graph.EspProcess
+
+object ScenarioParser {
```
arkadius, line 8: Thanks for extracting it +1 Can u replace other places of this code duplication?

engine/lite/api/src/main/scala/pl/touk/nussknacker/engine/lite/api/utils/sources.scala
mproch: :+1: done

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaComponentProvider.scala
mproch: It turned out it's actually not needed, I reverted this change...

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaSourceImplFactory.scala
mproch: I added trait for common `generateTestData` and added comments on testDataParser implementations

interpreter/src/main/scala/pl/touk/nussknacker/engine/marshall/ScenarioParser.scala
mproch: done in https://github.com/TouK/nussknacker/pull/2633/commits/65e1f25b0cf1bbc6cfea9cd4c26375075c11d50b

arkadius (approved):

PR #2635, opened 2021-12-08
Lite kafka source failure handling

arkadius:


engine/lite/kafka/api/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/api/LiteKafkaSource.scala
```diff
@@ -1,31 +1,36 @@
...
-  private var context: EngineRuntimeContext = _
+  protected var context: EngineRuntimeContext = _
 
   override def open(context: EngineRuntimeContext): Unit = {
     this.context = context
   }
 
-  override def createTransformation[F[_] : Monad](evaluateLazyParameter: customComponentTypes.CustomComponentContext[F]): ConsumerRecord[Array[Byte], Array[Byte]] => ValidatedNel[ErrorType, Context] = record =>
-    Valid(deserialize(context, record))
+  protected def nextContextId: ContextId
```
mproch, line 26: I don't think I like this exposing nextContextId, but I don't have a better idea now, so let's leave it...

mproch (approved):

engine/lite/kafka/api/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/api/LiteKafkaSource.scala
arkadius: I also don't like it. The problem was that `ContextInitializingFunction` was coupled with `ContextIdGenerator` and it was done because of Flink serialization. I've decoupled them and it looks better now. Can you take a look?

PR #2650, opened 2021-12-15
E2E test for embedded

mproch:


e2e-tests/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
+version: '3.7'
+
+services:
+  designer:
+    container_name: e2e_designer
+    image: touk/nussknacker:${NUSSKNACKER_VERSION-staging-latest}
```
dswiecki, line 6: probably we can omit fallback part

e2e-tests/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    environment:
+      CONFIG_FILE: ${NUSSKNACKER_CONFIG_FILE-/opt/nussknacker/conf/dev-application.conf}
```
dswiecki, line 8: and here do we set this or use fallback ?

e2e-tests/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    - --reserve-memory
+    - 0M
+    - --overprovisioned
+    - --node-id
+    - '0'
+    - --kafka-addr
+    - PLAINTEXT://0.0.0.0:9092,OUTSIDE://0.0.0.0:3032
+    - --advertise-kafka-addr
+    - PLAINTEXT://kafka:9092,OUTSIDE://localhost:3032
+    # NOTE: Please use the latest version here!
```
dswiecki, line 48: using latest is a bad practice in general ;)

e2e-tests/test/helpers.bash
```diff
@@ -0,0 +1,33 @@
...
+read_message() {
+  docker run --network=$NETWORK edenhill/kcat:1.7.0 -C -b $KAFKA_ADDRESS -t $1 -o beginning -c 1 -e | sed s/[^[:print:]]//g
+}
+
+prepare_deployed_scenario() {
+  NAME=$1
+  SUFFIX=$2
+
+  curl --fail -u admin:admin -X POST $DESIGNER_URL/api/processes/$NAME/LiteStreaming
+  cat test/data/scenario.json | sed s/##NAME##/$NAME/g | sed s/##SUFFIX##/$SUFFIX/g | curl --fail -u admin:admin -F process=@- $DESIGNER_URL/api/processes/import/$NAME | (echo '{ "comment": "created by test", "process": '; cat; echo '}') | curl --fail -u admin:admin -X PUT -H "Content-type: application/json" $DESIGNER_URL/api/processes/$NAME -d @-
```
dswiecki, line 31: maybe split this into few vars so it would be easier to reason about in case of some failure

e2e-tests/docker-compose.yml
mproch: :+1:

e2e-tests/docker-compose.yml
mproch: :+1:

e2e-tests/docker-compose.yml
mproch: I removed this comment...

e2e-tests/test/helpers.bash
mproch: :+1:

dswiecki (approved):

PR #2651, opened 2021-12-15
Support passing output from fragments which clear context

gadomsky:


interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -68,7 +68,7 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
         interpretNext(next, newCtx.pushNewContext(vars))
       case SubprocessEnd(_, varName, fields, next) =>
         val updatedCtx = createOrUpdateVariable(ctx, varName, fields)
-        val parentContext = ctx.popContext
+        val parentContext = ctx.parentContext.getOrElse(updatedCtx.copy(variables = Map.empty))
```
mproch, line 71: can we add some comment/todo why we reuse this context instead of creating a new one?

engine/flink/components/base/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/TransformersTest.scala
```diff
@@ -151,6 +154,43 @@ class TransformersTest extends FunSuite with FlinkSpec with Matchers with Inside
     aggregateVariables shouldBe List(3, 5)
   }
 
+  test("sum tumbling aggregate when in fragment") {
```
mproch, line 157: Can we add a test for validation that shows that e.g. input is not accessible after fragment?

mproch (approved):
:+1: thanks

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
gadomsky: done

docs/scenarios_authoring/Fragments.md
```diff
@@ -23,3 +23,4 @@ Fragment can define zero, one or more outputs. Each of them has a name, main sce
 - They cannot be nested (i.e. fragment cannot invoke other fragment).
 - They cannot pass output variables to the main scenario. This may change in the future.
 - When inputs/outputs of fragment change, scenarios using it have to be corrected manually.
+- If fragment use some component which clears context, context will be also cleared in the main scenario.
```
dswiecki, line 26: @ZbyszekMM do you get what context is ?

dswiecki: We should either describe what context is (here above or in Intro) or refer to I rather simply as `variables`

gadomsky: @dswiecki I agree, `context` term could be obscure to readers of this doc.
See new changes if they are sufficient

PR #2654, opened 2021-12-15
Embeddeddm persistent

arkadius:


engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManager.scala
```diff
@@ -52,34 +53,47 @@ object EmbeddedDeploymentManager extends LazyLogging {
...
+                                processingTypeDeploymentService: ProcessingTypeDeploymentService,
                                 handleUnexpectedError: (ProcessVersion, Throwable) => Unit)(implicit ec: ExecutionContext) extends BaseDeploymentManager with LazyLogging {
 
   private val metricRegistry = LiteEngineMetrics.prepareRegistry(engineConfig)
 
   private val contextPreparer = new LiteEngineRuntimeContextPreparer(new DropwizardMetricsProviderFactory(metricRegistry))
 
-  @volatile private var interpreters = Map[ProcessName, ScenarioInterpretationData]()
+  @volatile private var interpreters: Map[ProcessName, ScenarioInterpretationData] = {
+    val deployedScenarios = Await.result(processingTypeDeploymentService.getDeployedScenarios, 10.seconds)
```
mproch, line 64: let's extract `10.seconds` to some constant... so it's more visible. OTOH, I wonder if this method will ever be invoked without await ;)

mproch (commented):
:+1: For this change, let's merge refactor first...

engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManager.scala
arkadius: constant extracted + rebased to staging

mproch: :+1:

engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManager.scala
```diff
@@ -52,34 +53,49 @@ object EmbeddedDeploymentManager extends LazyLogging {
...
+  private def deployScenario(processVersion: ProcessVersion, deploymentData: DeploymentData, parsedResolvedScenario: EspProcess) = {
+    val jobData = JobData(parsedResolvedScenario.metaData, processVersion, deploymentData)
+    val interpreter = new KafkaTransactionalScenarioInterpreter(parsedResolvedScenario, jobData, modelData, contextPreparer)
+    val result = interpreter.run()
+    result.onComplete {
+      case Failure(exception) => handleUnexpectedError(processVersion, exception)
+      case Success(_) => //closed without problems
     }
+
+    val deploymentId = UUID.randomUUID().toString
```
mproch, line 95: I wondered a bit if we should reuse the previous deploymentId when we restart the designer? But maybe it's just new deployment then...

PR #2678, opened 2021-12-22
Lite kafka event timestamp union

maciej-brzezinski:

### DIFF
--- a/api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
+++ b/api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
@@ -64,7 +64,7 @@ case class Context(id: String, variables: Map[String, Any], parentContext: Optio
   def popContext : Context =
     parentContext.getOrElse(throw new RuntimeException("No parent context available"))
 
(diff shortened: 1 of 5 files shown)


api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -64,4 +64,7 @@ case class Context(id: String, variables: Map[String, Any], parentContext: Optio
   def popContext : Context =
     parentContext.getOrElse(throw new RuntimeException("No parent context available"))
 
+  def clearRedundantVariablesExcept(variablesToLeave: Set[String]) : Context =
```
mproch, line 67: Can we have a method like `clearUserVariables`? or sth like that? And have EventTimestampVariableName in that method?

maciej-brzezinski: changed it, please check if it's okay now

mproch: Let's add a comment that it leaves technical variables, hidden from user

mproch (approved):
:+1: let's just add comment to Context method

PR #2679, opened 2021-12-22
Contributing guide cleanups:

arkadius:


CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
 # Contributing
 
-## Setting up development environment
+## Hot to be involved?
```
mproch, line 3:
```suggestion
## How to get involved?
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
...
+### Building
+
+#### Building tarball
+```sbt dist/Universal/packageZipTarball```
+
+#### Building stage - stage is a local directory with all the files laid out as they would be in the final distribution
+```sbt dist/Universal/stage```
+
+#### Publish jars to local repository
+```sbt dist/Universal/publishLocal```
```
mproch, line 123: what does this really do in the case of dist? maybe let's add dist/Docker/publishLocal instead?

CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
...
+cp -r client/.federated-types/nussknackerUi submodules/types/@remote
+cd submodules
+npm ci
+CI=true npm run build
+cd ../..
+sbt copyUiSubmodulesDist
+```
+It will produce submodules static assets and copy them to `./ui/server/target/scala-XXX/classes/web/submodules/` that make them accessible via http://localhost:8080/submodules/*
+
+2. If you want to test verification mechanism, you need to make directory with savepoints available from your dev host. You can use `./bindSavepointsDirLocally.sh` script for that.
```
mproch, line 111:
```suggestion
2. If you want to test the verification mechanism (used during redeployment of Flink scenarios), you need to make a directory with savepoints available from your dev host. You can use `./bindSavepointsDirLocally.sh` script for that.
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
...
+   At the end you need to turn `FLINK_SHOULD_VERIFY_BEFORE_DEPLOY` flag on in environment variables.
+
+### Building
```
mproch, line 114: Let's add a reference to readme in ui/client - I think it's ok that it's in separate file, but reference link would be useful

CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
 # Contributing
 
-## Setting up development environment
+## Hot to be involved?
+
+All forms of contribution are welcome, including:
+- Bug reports, proposals: https://github.com/TouK/nussknacker/issues/new/choose
+- Pull requests: https://github.com/TouK/nussknacker/pulls
+
+## Releasing strategy
```
mproch, line 9: :+1:

CONTRIBUTING.md
```diff
@@ -1,12 +1,141 @@
...
+
+We are trying to have regular release (about once a two months) in similar to [Kafka manner](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan)
```
mproch, line 11:
```suggestion
We are trying to have regular releases (about once every two months), similar to [Kafka way](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan).
```

mproch (commented):
:+1: cool, added some minor comments

CONTRIBUTING.md
arkadius: Don't know what original poster was in mind ;) will change

CONTRIBUTING.md
arkadius: It was in troubleshooting above but good idea - will be more intuitive if it will be in building section. I've added precise steps and reference

mproch (approved):

PR #2682, opened 2021-12-23
Improvement: Request Response Clinet handle errors

lciolecki:


engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/management/RequestResponseProcessClient.scala
```diff
@@ -113,5 +118,21 @@ class HttpRequestResponseClient(managementUrl: String)(implicit backend: SttpBac
 
   override def close(): Unit = Await.result(backend.close(), Duration(10, TimeUnit.SECONDS))
 
+  private def handleUnitResponse(action: String)(response: Response[Either[String, String]]): Future[Unit] = (response.code, response.body) match {
```
mproch, line 121: Can you move it to httputils, extracting also from flink client?

dswiecki (commented):
Please add changelog entry

lciolecki:
> Please add changelog entry

Earlier.. @arkadius said that we shouldn't add everything to changelog.. IMO this change is not so important..

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/management/RequestResponseProcessClient.scala
lciolecki: Done, can you look right now?

mproch (approved):

PR #2689, opened 2021-12-27
EmbeddedDeploymentManager: handling scenario incompatible with compon…

arkadius:


engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManager.scala
```diff
@@ -88,26 +88,30 @@ class EmbeddedDeploymentManager(modelData: ModelData, engineConfig: Config,
 
   private def deployScenario(processVersion: ProcessVersion, deploymentData: DeploymentData, parsedResolvedScenario: EspProcess) = {
     val jobData = JobData(parsedResolvedScenario.metaData, processVersion, deploymentData)
-    val interpreter = new KafkaTransactionalScenarioInterpreter(parsedResolvedScenario, jobData, modelData, contextPreparer)
-    val result = interpreter.run()
-    result.onComplete {
-      case Failure(exception) => handleUnexpectedError(processVersion, exception)
-      case Success(_) => //closed without problems
+    val interpreterTry = Try {
```
mproch, line 91: I wonder if we want to handle all errors in this way? And I wonder if we should do it only during starting? Now if sth is wrong with the scenario, we fail deployment and we assume 'it didn't happen'. Now deployment will succeed, but the scenario will be in some failed state.

arkadius: 1. all errors: probable no. In some situations probable would be better to not handle failures and make whole designer blow up, but in the other hand, I think that number of situations of failures when error should be isolated to scenario scope is greater then the only problem, when scenario is incompatible. I would recommend to catch all exceptions and when we consider that it was catched too much, we could add some exceptions to this rule
2. only during starting: I think that we should have consistent behavior for both failures during starting and during deployment, and yes - I know that this change will modify also case of failures during handling deploy button, but for now I can't see any danger in it. Do you see some? Can you specify some test case that should be ended be handling failure in the old way?

mproch: ad 1. oook...
ad 2. no particular test case, but I think it's better to tell the user it's not ok immediately, rather than say 'Scenario was deployed' if we know it's not... But if you want we can leave it as it is...

arkadius: Ad 2. ok, got it, I will throw immediately for  deploy triggered by the button

arkadius: fixed

mproch: thx

mproch (approved):
:+1:

mproch (approved):

PR #2690, opened 2021-12-27
Streaming-lite scenario metadata

arkadius:


api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -46,16 +46,31 @@ sealed trait ScenarioSpecificData extends TypeSpecificData
...
+
+}
+
+case class LiteStreamMetaData(parallelism: Option[Int] = None) extends BaseStreamMetaData[LiteStreamMetaData] {
+
+  override def withParallelism(newParallelism: Option[Int]): LiteStreamMetaData = copy(parallelism = newParallelism)
+
+}
+
+sealed trait BaseStreamMetaData[T <: BaseStreamMetaData[T]] extends ScenarioSpecificData {
```
mproch, line 70: I wonder if we benefit from this trait (except for easier builder). Maybe for lite we won't have parallelism but sth else in the end?

arkadius: I don't see too many. One that I have in mind is that if we want to have similar expression of using streaming and streaming-lite, we should try to have consistent options for them. So if we will change the way how those number of workers is determined in streaming-lite, we should think about similar change in streaming. But we can do it also without type checking ... so I removed this parent class.

mproch (approved):
:+1: I wonder if we need MigrationGuide for this? Also, rebase pls

PR #2694, opened 2021-12-29
Ability to provide custom ConversionService for SpelExpressions

arkadius:


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -212,7 +209,8 @@ object SpelExpressionParser extends LazyLogging {
               typeDefinitionSet: TypeDefinitionSet,
               methodExecutionForUnknownAllowed: Boolean,
               dynamicPropertyAccessAllowed: Boolean,
-              spelExpressionExcludeList: SpelExpressionExcludeList)
+              spelExpressionExcludeList: SpelExpressionExcludeList,
+              conversionService: ConversionService = DefaultSpelConversionsProvider.getConversionService)
```
mproch, line 213: I wonder if we should provide default value here? It may happen that we're using this method in > 1 place?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/internal/DefaultSpelConversionsProvider.scala
```diff
@@ -16,9 +16,11 @@ import java.util.{Currency, Locale, UUID}
   * is lack of conversions between String and java time API (ZoneId, ZoneOffset, etc.), it has only support for legacy
   * java Date/Calendar API
   */
-object NuConversionServiceFactory {
+object DefaultSpelConversionsProvider extends DefaultSpelConversionsProvider
 
-  def prepareConversionService: ConversionService = {
+class DefaultSpelConversionsProvider extends SpelConversionsProvider {
```
mproch, line 21: Do we need both class and object? If they are not loaded via ServiceLoader?

arkadius: It is in case if someone would like to extends this class and add some additional covnersions

mproch (approved):

PR #2695, opened 2021-12-29
Add NodeComponentInfo to NuExceptionInfo

bartektartanus:


api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
```diff
@@ -2,4 +2,4 @@ package pl.touk.nussknacker.engine.api.exception
 
 import pl.touk.nussknacker.engine.api.Context
 
-case class NuExceptionInfo[T <: Throwable](nodeId: Option[String], throwable: T, context: Context) extends Serializable
\ No newline at end of file
+case class NuExceptionInfo[T <: Throwable](nodeId: Option[String], componentName: Option[String], componentType: Option[String], throwable: T, context: Context) extends Serializable
```
arkadius, line 5: Please use `ComponentType` for type

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -44,7 +44,19 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
 
   private implicit def nodeToId(implicit node: Node): NodeId = NodeId(node.id)
 
-  private def handleError(node: Node, ctx: Context): Throwable => NuExceptionInfo[_ <: Throwable] = NuExceptionInfo(Some(node.id), _, ctx)
+  private def handleError(node: Node, ctx: Context): Throwable => NuExceptionInfo[_ <: Throwable] = {
+    val componentName = node match {
```
arkadius, line 48: We already have sth similar to this in `DefaultComponentIdProvider.nodeToComponentId`. It would be nice to have both things in one place

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
bartektartanus: 1. Ok
2. For basic components (like Filter or Switch) the `componentName` was actually empty, but `componentType` was `Some("filter")`. Maybe for basic component we can use the same value for `componentName` and `componentType`, WDYT?

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
bartektartanus: `DefaultComponentIdProvider` needs `NodeData` from `engine.graph` package, while in my case `Node` is from `engine.compiledgraph`. So it looks similar but I'm not sure if I should put them in one place...

bartektartanus: I've extracted it into separate class in `interpreter` module.

api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentType.scala
```diff
@@ -49,6 +50,7 @@ object ComponentType extends Enumeration {
     case _: CustomNodeData => Some(CustomNode)
     case _: Enricher => Some(Enricher)
     case _: Processor => Some(Processor)
+    case _: BranchEndData => Some(BranchEnd)
```
bartektartanus, line 53: Test is failing because I've added this new type - not sure then if this is really needed? 

Is `BranchEnd` a component?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkLazyParameterFunctionHelper.scala
```diff
@@ -4,12 +4,13 @@ import org.apache.flink.api.common.functions.{RuntimeContext, _}
...
 import org.apache.flink.util.Collector
 import pl.touk.nussknacker.engine.api._
+import pl.touk.nussknacker.engine.api.exception.ExceptionComponentInfo
 import pl.touk.nussknacker.engine.flink.api.exception.ExceptionHandler
 
 /*
   This is helper class that allows to evaluate LazyParameter[T] in Flink functions.
  */
-class FlinkLazyParameterFunctionHelper(val nodeId: String,
+class FlinkLazyParameterFunctionHelper(val exceptionComponentInfo: ExceptionComponentInfo,
```
arkadius, line 13: This class should be named other way. `nodeId` wasn't used only for porpose of exception handling. In the `FlinkLazyParameterFunctionHelper` we have utilities for component implementations. Maybe the name should rather shows "what this class is" instead of "how it is usage". IMO this class is basically bunch of crusial identifiers related to component usage (which we call "node"). WTYT about `NodeIdentifiers` or `NodeComponentId`. Any other ideas?

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
```diff
@@ -1,5 +1,8 @@
 package pl.touk.nussknacker.engine.api.exception
 
 import pl.touk.nussknacker.engine.api.Context
+import pl.touk.nussknacker.engine.api.component.ComponentType.ComponentType
 
-case class NuExceptionInfo[T <: Throwable](nodeId: Option[String], throwable: T, context: Context) extends Serializable
\ No newline at end of file
+case class NuExceptionInfo[T <: Throwable](componentInfo: Option[ExceptionComponentInfo], throwable: T, context: Context) extends Serializable
+
+case class ExceptionComponentInfo(nodeId: String, name: String, typ: ComponentType)
```
arkadius, line 8: `name` and `typ` is related with component and `nodeId` is related with usage of this component (node). It would be nice to group it somehow. I see two options: 1) Let's wrap name and typ with some class like `ComponentIdData` (could be placed somewhere near `ComponentId` and `ComponentType` classes 2) At least prefix names of field with `component`

api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentType.scala
arkadius: No, `BranchEnd` is not an component. It is special node that make special topologies of our graphs possible (mainly joins) to be written as a tree structure.

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
arkadius: 2. IMO same value is ok. In the far future I would like to replace base components with components with type = custom and name = e.g. filter so this option is close to this vision.

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
arkadius: :+1: we can start with this approach and in the future think about some kind of merging both things. Can u add comment to this class that this logic should be kept synchronized with `DefaultComponentIdProvider` logic and TODO that it should be in one place in the future?

engine/flink/api/src/main/scala/pl/touk/nussknacker/engine/flink/api/process/FlinkLazyParameterFunctionHelper.scala
bartektartanus: `NodeIdentifiers` suggests that there are ids of many different nodes. `NodeComponentId` is good :)

api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentType.scala
bartektartanus: Removed 👍

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
bartektartanus: Ok, looks like first option is better suited in this case.

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
bartektartanus: TODO added :)

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
bartektartanus: Done :)

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
```diff
@@ -1,5 +1,16 @@
 package pl.touk.nussknacker.engine.api.exception
 
 import pl.touk.nussknacker.engine.api.Context
+import pl.touk.nussknacker.engine.api.component.ComponentType.ComponentType
 
-case class NuExceptionInfo[T <: Throwable](nodeId: Option[String], throwable: T, context: Context) extends Serializable
\ No newline at end of file
+case class NuExceptionInfo[T <: Throwable](nodeComponentId: Option[NodeComponentInfo], throwable: T, context: Context) extends Serializable
```
arkadius, line 6: should be `nodeComponentInfo` after class rename

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -174,7 +177,7 @@ class FlinkProcessRegistrar(compileProcess: (EspProcess, ProcessVersion, Deploym
       val contextTypeInformation = typeInformationDetection.forContext(part.validationContext)
 
       val start = source
-        .sourceStream(env, nodeContext(part.id, Left(ValidationContext.empty)))
+        .sourceStream(env, nodeContext(NodeComponentInfo(part.id, "source", ComponentType.Source), Left(ValidationContext.empty)))
```
arkadius, line 180: Why it is hardcoded name. We should hardcode it only for base component types. I wondering if we can do it more explicitly to someone not copy this code. Can u add `NodeComponentInfo.forBaseNode(nodeId, componentType)` (without name)?

api/src/main/scala/pl/touk/nussknacker/engine/api/exception/NuExceptionInfo.scala
bartektartanus: Fixed

arkadius (approved):

arkadius:
Please add changelog/migration guide

PR #2700, opened 2021-12-30
Closing interpreter after exception during open - moved compilation f…

arkadius:


arkadius:
closing in favour of #2701

PR #2701, opened 2021-12-30
Closing interpreter after exception during open - moved compilation f…

arkadius:


engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/metrics/SourceMetrics.scala
```diff
@@ -7,11 +7,14 @@ import pl.touk.nussknacker.engine.util.metrics.common.OneSourceMetrics
...
 @ThreadSafe
-class SourceMetrics(metricsProvider: MetricsProviderForScenario,
-                    sourceIds: Iterable[SourceId]) {
+class SourceMetrics(sourceIds: Iterable[SourceId]) {
 
   private val sourceMetrics = sourceIds.map(sourceId =>
-    sourceId -> new OneSourceMetrics(metricsProvider, sourceId.value)).toMap
+    sourceId -> new OneSourceMetrics(sourceId.value)).toMap
+
+  def registerOwnMetrics(metricsProvider: MetricsProviderForScenario): Unit = {
```
mproch, line 15: I wonder, maybe it would be simpler to initialize sourceMetrics map in this method? And leave OneSourceMetrics as it is?

mproch (approved):
:+1: Good job debugging this stuff ;)

PR #2702, opened 2022-01-03
[FIX] KafkaSourceFactory compilation fix (for errors)

gadomsky:


mproch (approved):
Do you think we can have a test for that somewhere?

gadomsky:
> Do you think we can have a test for that somewhere?

of course :) I was just checking if any test will fail after this change. And it did.

PR #2712, opened 2022-01-04
Add During Deploy state for Embedded DeploymentManager

kdutkowski:

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -34,7 +34,6 @@ Nussknacker versions
 * [#2684](https://github.com/TouK/nussknacker/pull/2684) Handled 'Restarting' state in Embedded DeploymentManager when the embedded scenario is failing
 * [#2686](https://github.com/TouK/nussknacker/pull/2686) Rename `ServiceWithStaticParameters` to `EagerServiceWithStaticParameters` to avoid confusion about lazy and eager parameters used by default
 * [#2695](https://github.com/TouK/nussknacker/pull/2695) Replaced `nodeId` with `NodeComponentInfo` in `NuExceptionInfo`
-* [#2712](https://github.com/TouK/nussknacker/pull/2712) Embedded DeploymentManager now starts in During Deploy state before it changes to Running
   
 1.1.1 (Not released yet)
 --------------------
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -34,7 +34,8 @@ Nussknacker versions
 * [#2684](https://github.com/TouK/nussknacker/pull/2684) Handled 'Restarting' state in Embedded DeploymentManager when the embedded scenario is failing
 * [#2686](https://github.com/TouK/nussknacker/pull/2686) Rename `ServiceWithStaticParameters` to `EagerServiceWithStaticParameters` to avoid confusion about lazy and eager parameters used by default
 * [#2695](https://github.com/TouK/nussknacker/pull/2695) Replaced `nodeId` with `NodeComponentInfo` in `NuExceptionInfo`
-
+* [#2712](https://github.com/TouK/nussknacker/pull/2712) Embedded DeploymentManager now starts in During Deploy state before it changes to Running
+  
 1.1.1 (Not released yet)
 --------------------
 * [#2660](https://github.com/TouK/nussknacker/pull/2660) Fix for handling errors after split in async mode
--- a/engine/lite/embeddedDeploymentManager/src/test/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManagerRestartTest.scala
+++ b/engine/lite/embeddedDeploymentManager/src/test/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManagerRestartTest.scala
@@ -22,8 +22,6 @@ class EmbeddedDeploymentManagerRestartTest extends BaseEmbeddedDeploymentManager
       fixture.deployScenario(scenario)
     }
 
-    manager.findJobStatus(name).futureValue.map(_.status) shouldBe Some(SimpleStateStatus.Running)
-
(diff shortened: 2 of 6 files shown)


engine/lite/embeddedDeploymentManager/src/test/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManagerRestartTest.scala
```diff
@@ -22,8 +22,6 @@ class EmbeddedDeploymentManagerRestartTest extends BaseEmbeddedDeploymentManager
       fixture.deployScenario(scenario)
     }
 
-    manager.findJobStatus(name).futureValue.map(_.status) shouldBe Some(SimpleStateStatus.Running)
-
     kafkaZookeeperServer.kafkaServer.shutdown()
```
dswiecki, line 25: wouldn't it now be clearer if this shutdown and await would be done after Restarting check - right befeore line 34

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/LoopUntilClosed.scala
```diff
@@ -102,9 +103,7 @@ class LoopUntilClosed(prepareSingleRunner: () => Task, waitAfterFailureDelay: Fi
           tryWithInterruptedHandle {
             status = Restarting
             Thread.sleep(delayToWait)
-          } {
-            status = Running
-          }
+          } { }
```
mproch, line 106: after this change, we'll be switching Restarting => Running only after successful init()? I think it's ok, just wanted to make it clear

mproch (commented):
Looks good, we just need to make checking "during deploy" deterministic ;)

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/LoopUntilClosed.scala
kdutkowski: Yep, I thought it will be simpler to make it consistent regardless of whether restarting takes place or not.

engine/lite/embeddedDeploymentManager/src/test/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManagerRestartTest.scala
kdutkowski: I'm not sure what you mean by that, the shutdown needs to complete for this assertion to pass `jobStatus.map(_.status) shouldBe Some(EmbeddedStateStatus.Restarting)`

dswiecki: hmm I think I checked it last week and this assertion passed regardless of shutdown placement

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/LoopUntilClosed.scala
```diff
@@ -19,8 +19,8 @@ class TaskRunner(taskName: String,
                  singleRun: String => Task,
                  terminationTimeout: Duration,
                  waitAfterFailureDelay: FiniteDuration) extends AutoCloseable with LazyLogging {
-  def status(): TaskStatus = tasks.find(p => p.status == Restarting).map(_.status
-  ).getOrElse(Running)
+  def status(): TaskStatus = tasks
```
arkadius, line 22: I would be for precedence (from most important to lower)
- restarting
- during_deploy
- running
IMO restarting is more important than during deploy because it is some failure situation

docs/Changelog.md
```diff
@@ -34,7 +34,8 @@ Nussknacker versions
 * [#2684](https://github.com/TouK/nussknacker/pull/2684) Handled 'Restarting' state in Embedded DeploymentManager when the embedded scenario is failing
 * [#2686](https://github.com/TouK/nussknacker/pull/2686) Rename `ServiceWithStaticParameters` to `EagerServiceWithStaticParameters` to avoid confusion about lazy and eager parameters used by default
 * [#2695](https://github.com/TouK/nussknacker/pull/2695) Replaced `nodeId` with `NodeComponentInfo` in `NuExceptionInfo`
-
+* [#2712](https://github.com/TouK/nussknacker/pull/2712) Embedded DeploymentManager now starts in During Deploy state before it changes to Running
```
arkadius, line 37: Please remove this changelog entry, embedded DM is still an undocumented feature - let's not mess changelog, it can be confusing for users.

engine/lite/embeddedDeploymentManager/src/test/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManagerRestartTest.scala
kdutkowski: Could be that because of the timeout which was set too low, the deployment went into Restarting status regardless of the shutdown... should be better now.

docs/Changelog.md
arkadius: Good question! looks like @mproch leaked this hidden feature ;-) But despite this, our changelog should be written from perspective of user who has knowledge about what was in  version N-1 and want to know what's new is in version N. Would he/she be interested in fact that "Embedded DeploymentManager" was implemented in a few steps by us? I propose to just add this PR link to entry about "Embedded DeploymentManager" instead of adding new entry

arkadius (approved):

PR #2717, opened 2022-01-06
cypress tests fixed

JulianWielga:
retry location query check. 
mask version info on snapshots.


arkadius (approved):
thank u for making tests more stable

PR #2726, opened 2022-01-07
Lite metrics: using instanceId instead of host tag

arkadius:


engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/streaming/embedded/EmbeddedDeploymentManager.scala
```diff
@@ -59,7 +59,9 @@ class EmbeddedDeploymentManager(modelData: ModelData, engineConfig: Config,
 
   private val retrieveDeployedScenariosTimeout = 10.seconds
 
-  private val metricRegistry = LiteEngineMetrics.prepareRegistry(engineConfig)
+  // TODO: better would be to use some global instance id - be default it could be:
+  //  instance id of designer (hostname:port by default) + scenario type (will identify deployment manager)
```
mproch, line 63: These metrics are scenario-based, so I think hostname + port should be ok?

engine/lite/kafka/runtime/src/universal/bin/nu-kafka-engine-entrypoint.sh
```diff
@@ -14,6 +14,10 @@ LOGBACK_FILE=${LOGBACK_FILE-"$CONF_DIR/docker-logback.xml"}
 
 export KAFKA_ADDRESS=${KAFKA_ADDRESS:-localhost:9092}
 export KAFKA_TOPIC=${KAFKA_TOPIC:-"errors"}
+# For k8s deployments we crop POD_NAME to last part which is an id of replica (hash) to make metrics tags shorten
```
mproch, line 17: Let's just set INSTANCE_ID env in pod definition? I think the less logic we have in this entrypoint, the better

engine/lite/request-response/app/src/main/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpApp.scala
```diff
@@ -25,13 +26,17 @@ object RequestResponseHttpApp extends Directives with FailFastCirceSupport with
...
-  val metricRegistry = LiteEngineMetrics.prepareRegistry(config)
-
-  val requestResponseApp = new RequestResponseHttpApp(config, metricRegistry)
-
   val managementPort = Try(args(0).toInt).getOrElse(8070)
   val processesPort = Try(args(1).toInt).getOrElse(8080)
 
+  lazy val hostname = sys.env.getOrElse(
+    "HOSTNAME",
+    throw new IllegalStateException("HOSTNAME environment variable unavailable"))
```
mproch, line 34: I'm afraid this will make running on windows a bit more difficult ;) but probably we can live with it...

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/metrics/dropwizard/LiteMetricRegistryFactory.scala
```diff
@@ -60,7 +59,15 @@ object LiteEngineMetrics extends LazyLogging {
...
+                                instanceId: Option[String],
                                 environment: String,
                                 additionalTags: Map[String, String] = Map.empty)
 }
+
+object LiteMetricRegistryFactory {
+
+  def usingHostnameAsDefaultInstanceId = new LiteMetricRegistryFactory(sys.env.getOrElse(
+    "HOSTNAME",
+    throw new IllegalStateException("HOSTNAME environment variable unavailable")))
```
mproch, line 71: Can we also have some method for adding a suffix (like port etc.), so we don't have to put this HOSTNAME logic twice?

engine/lite/kafka/runtime/src/universal/bin/nu-kafka-engine-entrypoint.sh
arkadius: I want to, but looks like there is no option to do it - I can add comment

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -66,12 +67,19 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
     val preparer = new LiteEngineRuntimeContextPreparer(new DropwizardMetricsProviderFactory(metricRegistry))
     // TODO Pass correct ProcessVersion and DeploymentData
     val jobData = JobData(scenario.metaData, ProcessVersion.empty, DeploymentData.empty)
 
     KafkaTransactionalScenarioInterpreter(scenario, jobData, modelData, preparer)
   }
 
+  private def prepareMetricRegistry(engineConfig: Config) = {
+    lazy val instanceId = sys.env.get("INSTANCE_ID") orElse sys.env.get("HOSTNAME") getOrElse {
+      throw new IllegalStateException("Cannot determine instanceId: neither INSTANCE_ID nor HOSTNAME environment variables unavailable")
```
dswiecki, line 80: can we do here some generic fallback - I wonder if throwing exception would be to harsh ?

engine/lite/request-response/app/src/main/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpApp.scala
arkadius: I forgot about this case - I've added some solution for that.

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: I've changed it a bit - now it is fallback to "localhost".

mproch (approved):
:+1:

dswiecki (approved):

PR #2734, opened 2022-01-11
K8s DM: readiness and liveness

arkadius:


utils/util/src/main/scala/pl/touk/nussknacker/engine/util/loader/ModelClassLoader.scala
```diff
@@ -1,18 +1,28 @@
...
 import java.net.{URL, URLClassLoader}
 
 case class ModelClassLoader private(classLoader: ClassLoader,
-                                    urls: List[URL])
+                                    urls: List[URL]) {
+
+  override def toString: String = s"ModelClassLoader(${toString(classLoader)})"
 
-object ModelClassLoader {
+  private def toString(classLoader: ClassLoader): String = classLoader match {
```
mproch, line 12: just curious, why do we need it instead of URL list?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -26,16 +26,24 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
+  val runtimeConfig = ConfigFactory.load(ConfigFactoryExt.parseUnresolved(classLoader = getClass.getClassLoader))
 
+  val system = ActorSystem("nu-kafka-runtime", runtimeConfig)
+  import system.dispatcher
+
+  val scenarioInterpreter = prepareScenarioInterpreter(runtimeConfig)
   Runtime.getRuntime.addShutdownHook(new Thread() {
     override def run(): Unit = {
       logger.info("Closing KafkaTransactionalScenarioInterpreter")
       scenarioInterpreter.close()
```
mproch, line 38: should we close also the ActorSystem here?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -26,16 +26,24 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
     }
   })
 
-  Await.result(scenarioInterpreter.run(), Duration.Inf)
+  Await.result(for {
+    _ <- new HealthCheckServerRunner(system, scenarioInterpreter).start()
```
mproch, line 43: We don't need to close it somehow? if e.g. exception happens in run?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/HealthCheckServerRunner.scala
```diff
@@ -0,0 +1,59 @@
...
+import akka.pattern._
+import akka.util.Timeout
+import com.typesafe.scalalogging.LazyLogging
+import pl.touk.nussknacker.engine.lite.kafka.KafkaScenarioInterpreterStatusCheckerActor.GetStatus
+import pl.touk.nussknacker.engine.lite.kafka.TaskStatus.{Running, TaskStatus}
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class HealthCheckServerRunner(system: ActorSystem, scenarioInterpreter: KafkaTransactionalScenarioInterpreter) {
```
mproch, line 15: Cannot help thinking it's a bit of an overkill, but let's leave it :)

mproch (commented):
I wonder, maybe it would be good to also add checks configuration to K8sDeploymentManager, so we know it works as intended?

utils/util/src/main/scala/pl/touk/nussknacker/engine/util/loader/ModelClassLoader.scala
arkadius: Because I want to see what exactly classloader was used (what hierarchy). Currently it shows `UrlClassloader@hash`. I have a  with compilation of scenario. After a while of digging I figure out that the problem was because compilation was done before topic schema registration - we are checking it not only on designer side, but also on runtime side. IMO it can be helpful in the future. In the past I have a few situations that I was leaving a breakpoint to see what is a hierarchy of those classloaders

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: Good point, added

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: Good point, added

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/HealthCheckServerRunner.scala
arkadius: Why overkill? It is very light library - it depends only on akka-http. Amount of code is less or equal to writting it on our side. Benefits are: 
- we can easily configure combination of checks in application.conf
- we stick to some akka standard solution instead of writting our new one (standard port, urls, we can use some standard checks), good documentation: https://doc.akka.io/docs/akka-management/current/index.html#overview
- we can turn on security if we want
- some additional extensions available that could be useful
- it will be easy to switch to our own made solution

arkadius:
> I wonder, maybe it would be good to also add checks configuration to K8sDeploymentManager, so we know it works as intended?
I've added. I've only checked it manually by turning on `Waiting30SecsCheck` readiness-check and looking into logs / deployment status.

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/HealthCheckServerRunner.scala
mproch: ok ;) :+1:

utils/util/src/main/scala/pl/touk/nussknacker/engine/util/loader/ModelClassLoader.scala
mproch: :+1:

mproch (approved):

PR #2737, opened 2022-01-12
EitherValuesDetailedMessage - similar to EitherValues but printing information about left value when isLeft

arkadius:


mproch (approved):
cool :+1:

Issue #1989, opened 2021-07-30
Error message for Custom scenario contains invalid process link


piotrp:
**Describe the bug**
When custom scenario fails and Nussknacker displays "Deployed scenarios not running (probably failed): ..." message along with a link to scenario, but it's hardcoded to use `/visualization/{scenario name}`, and custom scenarios aren't editable in this view. Clicking on this link results in:

![obraz](https://user-images.githubusercontent.com/1409807/127638226-410bdb51-c491-43b8-8008-09b11e66b8cb.png)

Example link: https://staging.nussknacker.io/visualization/customProcess1

**To Reproduce**
Deploy a custom scenario that will fail when deployed.

Nussknacker shouldn't link to custom scenarios, and when going to visualization should display a sane error message telling that this process is not editable.

**Environment (please complete the following information):**
 - Nussknacker: staging@360bcb1b775e384de9cf2d11af15fcbbca72eb02

PR #2741, opened 2022-01-12
Improvements: Remove CustomProcess

lciolecki:
- Remove BE custom process elements
- Remove FE custom process tab
- Replace ProcessDeploymentData by GraphProcess (as VO)
- Replace scenario jsonString by GraphProcess

Closes: https://github.com/TouK/nussknacker/issues/1989


api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/GraphProcess.scala
```diff
@@ -2,17 +2,18 @@ package pl.touk.nussknacker.engine.api.deployment
...
-//
-//}
+object GraphProcess {
 
-final case class GraphProcess(processAsJson: String) extends AnyVal
+  //TODO: Add json validation here?
+  def apply(jsonString: String): GraphProcess = {
+    val json = parse(jsonString) match {
+      case Left(_) => throw new IllegalArgumentException(s"Invalid raw json string: $jsonString.")
+      case Right(json) => json.spaces2
```
mproch, line 11: if we parse it, let's have Json inside... :)

lciolecki: yep, I know it. Right now I've left it as string, because it was easier.

api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/GraphProcess.scala
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    new GraphProcess(json)
+  }
+
+}
+
+final case class GraphProcess private(json: Json) extends AnyVal {
+  override def toString: String = jsonString
+  def jsonString: String = json.spaces2
```
mproch, line 27: don't think it's needed here, we should use it rarely - only when serializing stuff during deployment..

api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/GraphProcess.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object GraphProcess {
+
+  implicit val encoder: Encoder[GraphProcess] = Encoder.encodeJson.contramap(_.json)
+  implicit val decoder: Decoder[GraphProcess] = Decoder.decodeJson.map(GraphProcess(_))
+
+  val empty: GraphProcess = GraphProcess("{}")
+
+  //TODO: Add json validation here?
+  def apply(jsonString: String): GraphProcess = {
```
mproch, line 14: You can just use CirceUtil.decode here if you defined codec...

lciolecki: I saw decoding in another place - try to decode json to Canonical form.. But maybe in next step?

api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/GraphProcess.scala
lciolecki: I disagree, earlier we had a lot of usage: space2 / noSpaces on different places. Beacuse of that in db we have sometimes json with spacees or json without spaces. I think better way is normalize it and decide how present json in string in only one way.

lciolecki: I've removed jsonString, right nowe is only toString method.

arkadius:
please move changelog entry to 1.3 and add migration guide entry about changes in deployment manager api

mproch (approved):

PR #2745, opened 2022-01-13
K8s DM: scaling options

arkadius:


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -72,11 +73,12 @@ class K8sDeploymentManager(modelData: ModelData, config: K8sDeploymentManagerCon
   override def deploy(processVersion: ProcessVersion, deploymentData: DeploymentData,
                       processDeploymentData: ProcessDeploymentData,
                       savepointPath: Option[String]): Future[Option[ExternalDeploymentId]] = {
+    val (scalingOptions, scenarioWithCorrectedParallelism) = determineScalingOptions(processDeploymentData)
     for {
-      configMap <- k8sUtils.createOrUpdate(k8s, configMapForData(processVersion, processDeploymentData))
+      configMap <- k8sUtils.createOrUpdate(k8s, configMapForData(processVersion, scenarioWithCorrectedParallelism))
```
mproch, line 78: I wonder if it's the best way to achieve this. Maybe it would be better to pass a parameter to the image, saying how parallelism maps to threads?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sScalingOptionsDeterminer.scala
```diff
@@ -0,0 +1,73 @@
...
+    K8sScalingOptions(config.fixedReplicasCount, noOfTasksInReplica)
+  }
+
+}
+
+class DividingParallelismK8sScalingOptionsDeterminer(config: DividingParallelismConfig) extends K8sScalingOptionsDeterminer {
+
+  override def determine(parallelism: Int): K8sScalingOptions = {
+    val replicasCount = Math.ceil(parallelism.toDouble / config.tasksPerReplica).toInt
+    val noOfTasksInReplica = Math.ceil(parallelism.toDouble / replicasCount).toInt
```
mproch, line 44: I wonder, can it be sth different than config.tasksPerReplica? And if so - do we want it?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
arkadius: let's do this in next change

mproch (approved):

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -75,7 +83,11 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
     }
   }
 
-  private def prepareScenarioInterpreter(runtimeConfig: Config): KafkaTransactionalScenarioInterpreter = {
+  private def parseDeploymentData(path: Path): LiteKafkaJobData = {
```
mproch, line 86: I think this is not a good name for this config, since we already have DeploymentData class (which we don't parse properly yet...)

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/K8sScalingOptionsDeterminer.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def determine(parallelism: Int): K8sScalingOptions
+
+}
+
+case class K8sScalingOptions(replicasCount: Int, noOfTasksInReplica: Int)
+
+object K8sScalingOptionsDeterminer {
+
+  // 4 because it is quite normal number of cpus reserved for one container
```
mproch, line 19: I wonder, probably we could connect it to configured resources somehow? For our use case - scenario per pod - 4 seems quite large number?

dswiecki (approved):

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: Adding prefix `Lite` will be ok for you?

mproch: deploymentConfig?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/K8sScalingOptionsDeterminer.scala
arkadius: > connect it to configured resources somehow
I don't know what you have in mind by "configured resources"?
If we will need some better scaling options determiner, we can always change this strategy or add new. `K8sScalingOptions` can return more info, and `determine(...)` can take more parameters. But let's talk about some concrete target that we want to achieve.
Also I don't like to shuffle logic of scaling with "low level" `k8sDeploymentConfig` - it will become too hard to understand.  
> scenario per pod - 4 seems quite large number
why? why it depends on use case? It shouldn't depend only on available resources?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: why config? configs are something that are configured by user, not code

arkadius: It's are data/parameters passed to invocation of runtime, and it is passed similar as `DeploymentData` in Flink, so IMO name should be similar. It will be strange, if someone will ask you, how it is called in flink, and you will say that it is some totally other name

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/K8sScalingOptionsDeterminer.scala
mproch: 1. in the future we will probably set some default resource limits (e.g. on CPU), if they will be sth like "100m", then 4 threads is too much
2. for our target performance "10k/s per thread" 4 threads seems quite much :) 

you can ignore this comment, just wanted to discuss a bit :)

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
mproch: but DeploymentData is in deployment-api... for me it's a bit confusing to have two of them. Maybe we should additionalDeploymentData field or sth?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/K8sScalingOptionsDeterminer.scala
arkadius: probable you've right, but currently I don't have enough knowledge to do tuning of defaults. My thinking was - If our jobs are "fat" we should have lower number of them instead of higher. But I can be wrong - let's do tuning after performance tests

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: After rethinking - it can be deployment config. will change

PR #2758, opened 2022-01-19
jdk11 check

arkadius:


utils/util/src/main/java/pl/touk/nussknacker/engine/util/JavaClassVersionChecker.java
```diff
@@ -0,0 +1,10 @@
+package pl.touk.nussknacker.engine.util;
+
+// Scala 2.12 does not support target > 8 so we want to fail fast otherwise - be invoking this class at the beginning of all our Apps
+// Firstly loading of this class should fail because we pass -release to javac
+// Secondary we check if we are in java 11 invoking method that is only available there
+public class JavaClassVersionChecker {
+    public static void check() {
+        "".isBlank();
```
mproch, line 8: I wonder, maybe we should catch exception and rethrow sth meaningful?

nussknacker-dist/src/universal/bin/run.sh
```diff
@@ -44,6 +48,10 @@ if [[ "${RUN_IN_BACKGROUND}" == "true" ]]; then
   echo $! > $PID_FILE
   echo "Nussknacker up and running"
 else
+  echo "JVM:"
```
mproch, line 51: maybe sth like ```echo "JVM: `java --version`"``` (--version is to stdout, -version to stderr :) )

engine/lite/kafka/runtime/src/universal/bin/run.sh
```diff
@@ -33,6 +37,10 @@ if [[ "${RUN_IN_BACKGROUND}" == "true" ]]; then
   echo $! > $PID_FILE
   echo "Nussknacker Kafka Runtime up and running"
 else
+  echo "JVM:"
+  java -version
```
mproch, line 41: maybe it's time we support JAVA_HOME? 
like in: https://github.com/apache/flink/blob/master/flink-dist/src/main/flink-bin/bin/config.sh#L214?

utils/util/src/main/java/pl/touk/nussknacker/engine/util/JavaClassVersionChecker.java
arkadius: ok

nussknacker-dist/src/universal/bin/run.sh
arkadius: ok

engine/lite/kafka/runtime/src/universal/bin/run.sh
arkadius: I'd rather don't want to spend more time on this issues - let's wait with this until it will be needed

mproch (approved):

PR #2773, opened 2022-01-24
Improvement: Using VersionId instead of Long

lciolecki:


api/src/main/scala/pl/touk/nussknacker/engine/api/process/VersionId.scala
```diff
@@ -2,10 +2,9 @@ package pl.touk.nussknacker.engine.api.process
 
 import io.circe.{Decoder, Encoder}
 
-final case class VersionId(value: Long) extends AnyVal
+final case class VersionId(value: Long)
```
lciolecki, line 5: Because of `CONTRIBUTIN.MD`:
- Avoid using AnyVal if the class is in API that will be used from Java

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessActivityResource.scala
```diff
@@ -73,8 +73,8 @@ class AttachmentResources(attachmentService: ProcessAttachmentService,
           }
         }
       }
-    } ~ path("processes" / Segment / LongNumber / "activity" / "attachments" / LongNumber) { (processName, versionId, attachmentId) =>
-      (get & processId(processName)) { processId =>
+    } ~ path("processes" / Segment / VersionIdSegment / "activity" / "attachments" / LongNumber) { (processName, versionId, attachmentId) =>
+      (get & processId(processName)) { _ =>
```
mproch, line 77: I wonder, what do we need processName/versionId for?? it's also a security problem - you can pass attachmendId from scenario you don't have access to, processName/version you have access to, and.. 
let's add FIXME for that...

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessAttachmentService.scala
```diff
@@ -17,14 +17,14 @@ import scala.util.control.NonFatal
...
-  def saveAttachment(processId: ProcessId, processVersionId: Long, originalFileName: String, byteSource: Source[ByteString, Any])
+  def saveAttachment(processId: ProcessId, processVersionId: VersionId, originalFileName: String, byteSource: Source[ByteString, Any])
                     (implicit ec: ExecutionContext, loggedUser: LoggedUser, mat: Materializer): Future[Unit] = {
     val relativeFilePath = s"${processId.value}/${s"${System.currentTimeMillis()}-$originalFileName"}"
     val attachmentFile = getAttachmentFile(relativeFilePath)
     attachmentFile.getParentFile.mkdirs()
     val fileSink = FileIO.toPath(attachmentFile.toPath)
     byteSource.runWith(fileSink).flatMap { _ =>
-      val attachmentToAdd = AttachmentToAdd(processId.value, processVersionId, originalFileName, relativeFilePath)
+      val attachmentToAdd = AttachmentToAdd(processId.value, processVersionId.value, originalFileName, relativeFilePath)
```
mproch: why AttachmentToAdd has Longs?

mproch (commented):
Have you considered: 
https://scala-slick.org/doc/3.0.0/userdefined.html#using-custom-scalar-types-in-queries
(probably not in this change though ;) )

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessActivityResource.scala
lciolecki: Yep, a have looked on it. But if we want use `Mapped To [Long]`, we provide another dependency to api package. I'm not sure do we want it?

ui/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessAttachmentService.scala
lciolecki: It's old version ;)

lciolecki:
> Have you considered: https://scala-slick.org/doc/3.0.0/userdefined.html#using-custom-scalar-types-in-queries (probably not in this change though ;) )

Yep, a have looked on it. But if we want use Mapped To [Long], we provide another dependency to api package. I'm not sure do we want it?

We already use it at PeriodicProcessId. But periodic model assumes to use db?

arkadius:
> > Have you considered: https://scala-slick.org/doc/3.0.0/userdefined.html#using-custom-scalar-types-in-queries (probably not in this change though ;) )
> 
> Yep, I have looked on it. But if we want use Mapped To [Long], we provide another dependency to api package. I'm not sure do we want it?

I agree that we don't want to add another dependency to api...

arkadius (approved):
LGTM, please fix tests

mproch (approved):

PR #2775, opened 2022-01-24
Fix for: kafka-registry-typed-json not handle logical types in runtime

arkadius:
kafka-registry-typed-json source was recognizing logical types during typing but during evaluation were used raw, underlying types

### DIFF
--- a/utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
+++ b/utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
@@ -6,14 +6,15 @@ import org.apache.avro.Conversions.UUIDConversion
 import org.apache.avro.data.TimeConversions
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
-import org.apache.avro.{LogicalTypes, Schema}
+import org.apache.avro.{AvroRuntimeException, LogicalTypes, Schema}
 import pl.touk.nussknacker.engine.avro.schemaregistry.confluent.serialization.jsonpayload.JsonPayloadToAvroConverter._
 import tech.allegro.schema.json2avro.converter.types.AvroTypeConverter
-import tech.allegro.schema.json2avro.converter.{CompositeJsonToAvroReader, JsonAvroConverter}
+import tech.allegro.schema.json2avro.converter.{CompositeJsonToAvroReader, JsonAvroConverter, PathsPrinter}
 
 import java.math.RoundingMode
 import java.time.{Instant, LocalDate, LocalTime}
 import java.util
+import scala.util.Try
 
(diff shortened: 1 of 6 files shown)


utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
```diff
@@ -0,0 +1,146 @@
...
+    private val conversion = new TimeConversions.DateConversion
+
+    override def canManage(schema: Schema, path: util.Deque[String]): Boolean =
+      schema.getType == Schema.Type.INT && schema.getLogicalType == LogicalTypes.date()
+
+    override def convert(field: Schema.Field, schema: Schema, jsonValue: AnyRef, path: util.Deque[String], silently: Boolean): AnyRef = {
+      jsonValue match {
+        case number: Number => conversion.fromInt(number.intValue(), schema, schema.getLogicalType)
+        case str: String => Decoder[LocalDate].decodeJson(fromString(str)).fold(throw _, identity)
+        case other => other
```
mproch, line 46: do you think it's sensible to pass anything here? Wouldn't it be better to throw exception for sth unexpected?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
```diff
@@ -0,0 +1,146 @@
...
+    override def convert(field: Schema.Field, schema: Schema, jsonValue: AnyRef, path: util.Deque[String], silently: Boolean): AnyRef = {
+      jsonValue match {
+        case number: Number => conversion.fromLong(number.intValue(), schema, schema.getLogicalType)
+        case str: String => Decoder[Instant].decodeJson(fromString(str)).fold(throw _, identity)
+        case other => other
+      }
+    }
+  }
+
+  object TimestampMicrosConverter extends AvroTypeConverter {
```
mproch, line 96: all those methods look quite similar (in terms of convert method)? Shouldn't we extract sth common?

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
arkadius: I wrongly understand concept of fallbacking to other options in this converter. it looks that I should handle `silently` parameter and return INCOMPATIBLE or throw exception. Changed and added test for union showing where is a problem.

utils/avro-util/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/serialization/jsonpayload/JsonPayloadToAvroConverter.scala
arkadius: I've extracted something - we can't do too much with this

PR #2781, opened 2022-01-25
Rename lite split component to avoid clash with base component name

kdutkowski:

### DIFF
--- a/engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
+++ b/engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
@@ -14,19 +14,19 @@ import scala.language.higherKinds
 object ForEachTransformer extends CustomStreamTransformer {
 
   @MethodToInvoke(returnType = classOf[Object])
-  def invoke(@ParamName("list") list: LazyParameter[java.util.Collection[Any]],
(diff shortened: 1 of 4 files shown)


engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
```diff
@@ -11,22 +11,22 @@ import pl.touk.nussknacker.engine.lite.api.utils.transformers.SingleElementCompo
 import scala.collection.JavaConverters._
 import scala.language.higherKinds
 
-object ProcessSplitter extends CustomStreamTransformer {
+object ForEachTransformer extends CustomStreamTransformer {
 
   @MethodToInvoke(returnType = classOf[Object])
-  def invoke(@ParamName("parts") parts: LazyParameter[java.util.Collection[Any]],
+  def invoke(@ParamName("list") list: LazyParameter[java.util.Collection[Any]],
```
mproch, line 17: I wonder if `elements` would be more meaningful?

mproch (approved):
Added small comment, but overall +1

engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
kdutkowski: makes sense, I'll change it to `elements`

PR #2803, opened 2022-01-27
Set runtime image strictly

dswiecki:

### DIFF
--- a/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/DeploymentPreparer.scala
+++ b/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/DeploymentPreparer.scala
@@ -81,7 +81,8 @@ class DeploymentPreparer(config: K8sDeploymentManagerConfig) extends LazyLogging
       val containerLens = GenLens[Container](_.env).modify(_ ++ runtimeContainer.env) andThen
         GenLens[Container](_.volumeMounts).modify(_ ++ runtimeContainer.volumeMounts) andThen
         GenLens[Container](_.readinessProbe).modify(_.orElse(runtimeContainer.readinessProbe)) andThen
-        GenLens[Container](_.livenessProbe).modify(_.orElse(runtimeContainer.livenessProbe))
+        GenLens[Container](_.livenessProbe).modify(_.orElse(runtimeContainer.livenessProbe)) andThen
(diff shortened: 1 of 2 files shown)


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/deployment/DeploymentPreparer.scala
```diff
@@ -81,7 +81,8 @@ class DeploymentPreparer(config: K8sDeploymentManagerConfig) extends LazyLogging
       val containerLens = GenLens[Container](_.env).modify(_ ++ runtimeContainer.env) andThen
         GenLens[Container](_.volumeMounts).modify(_ ++ runtimeContainer.volumeMounts) andThen
         GenLens[Container](_.readinessProbe).modify(_.orElse(runtimeContainer.readinessProbe)) andThen
-        GenLens[Container](_.livenessProbe).modify(_.orElse(runtimeContainer.livenessProbe))
+        GenLens[Container](_.livenessProbe).modify(_.orElse(runtimeContainer.livenessProbe)) andThen
+        GenLens[Container](_.image).set(runtimeContainer.image)
```
arkadius, line 85: If someone will use other image in config, he/she can be confused that it was overridden. Maybe we can change it to:
```
.modify { configuredImage =>
  if (configuredImage != runtiemContainer.image) logger.warn...
  runtiemContainer.image
}
```
WDYT?

mproch: or even throw exception if image is set :)

dswiecki: @mproch @arkadius would do we prefer ? log.warn and throwing exception looks as if two alternative approaches ? @arkadius in case one do you think we should also log such a warning for other configs as label selector etc ?

arkadius: @dswiecki which one you prefer? I agree that throwing is also fine, probable even better. Regards to other places where `set` was used IMO we can behave the same. If we find out cases where it disallow some configuration, we will think how to handle it.

arkadius (approved):

PR #2825, opened 2022-02-04
scenario-api extracted - part 1

arkadius:


scenario-api/src/main/scala/pl/touk/nussknacker/engine/marshall/ProcessMarshaller.scala
```diff
@@ -100,10 +100,14 @@ object ProcessMarshaller {
...
+  def toJson(canonical: CanonicalProcess): Json =
+    Encoder[CanonicalProcess].apply(canonical)
+
+  def fromJson(jsonString: String): Validated[String, CanonicalProcess] =
+    Validated.fromEither(CirceUtil.decodeJson[Json](jsonString)).leftMap(_.getMessage)
+      .andThen(fromJson)
 
-  def fromGraphProcess(graphProcess: GraphProcess): Validated[ProcessJsonDecodeError, CanonicalProcess] =
-    Validated.fromEither(Decoder[CanonicalProcess].decodeJson(graphProcess.json)).leftMap(_.getMessage).leftMap(ProcessJsonDecodeError)
+  def fromJson(json: Json): Validated[String, CanonicalProcess] =
```
kdutkowski, line 110: Should this change `graphProcess` -> `json` be mentioned in changelog? or are we treating this part as internal API?

arkadius: To be fun, in 1.2 it was `fromJson` :D @lciolecki changed it to fromGraphProcess without mentioning about it in migration guide and now I'm reverting it because it won't has sense after removing GraphProcess (I've done it in second PR)

kdutkowski (approved):

PR #2830, opened 2022-02-08
Introduce more specific values for RunMode and rename it to ComponentUseCase

gadomsky:


api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
```diff
@@ -3,16 +3,21 @@ package pl.touk.nussknacker.engine.api.process
...
+  * In this mode real implementation of node should be used.</li>
   * <li>Test - process is run in test mode to collect results for test data. Can be used to stub real implementation.</li>
   * </ul>
   */
 sealed trait RunMode
 
 object RunMode {
 
-  case object Normal extends RunMode
+  sealed trait Normal extends RunMode
```
arkadius, line 15: Do u see some advantage in grouping four modes below to "normal" mode other than for backward compatibility? IMO it is not straightforward why e.g. TestDataGeneration is normal mode and test is test mode

api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
```diff
@@ -3,16 +3,21 @@ package pl.touk.nussknacker.engine.api.process
...
+
+  case object Engine extends Normal
+  case object Validation extends Normal
```
arkadius, line 18: Validation doesn't cause that node will be "run". What case do you see that user will need to handle it in component?

api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
```diff
@@ -3,16 +3,21 @@ package pl.touk.nussknacker.engine.api.process
...
+  case object ServiceQuery extends Normal
+  case object TestDataGeneration extends Normal
```
arkadius, line 20: The same as above . I think that exposing too many modes will cause that consumer of those modes will be confused how they should be handled. Also when we change implementation details, some of those modes will go away and we will be hard to migrate to new approach. Maybe it should be extracted only one new mode ServiceQuery which indeed is in context of running component and validation / testdatageneration should be hidden as a normal mode as a implementation detail? (also removed from documentation)

api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
gadomsky: Actually I see that from diffrent angle - I care only about `Engine` mode, to be able to distinguish that my code is run via engine vs designer. 
What mode do you think should I use while performing validation?

api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
gadomsky: No, it was only for sake of backwards compatibility.

docs/MigrationGuide.md
```diff
@@ -9,7 +9,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   * Replace scenario jsonString by GraphProcess at DeploymentManager, ProcessConfigEnricherInputData
 * [#2773](https://github.com/TouK/nussknacker/pull/2773) Using VersionId / ProcessId / ProcessName instead of Long or String:
   * `PullProcessRepository` API was changed, right now we use VersionId instead of Long
-  
+* [#2830](https://github.com/TouK/nussknacker/pull/2830) `RunMode.Normal` becomes sealed trait and has specific values: Engine, Validation, ServiceQuery, TestDataGeneration
```
dswiecki, line 12: copy this to changelog

api/src/main/scala/pl/touk/nussknacker/engine/api/process/RunMode.scala
arkadius: We agree that the real usage is not only for run part but also for definition part and runMode is passed that way to both places. So the only problem was Run in name of class. Let's rename it to ComponentUsage or similar

arkadius (approved):

docs/MigrationGuide.md
gadomsky: done

PR #2847, opened 2022-02-10
Add release name config option

dswiecki:


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -199,10 +201,12 @@ object K8sDeploymentManager {
       (other way to mitigate this would be to generate some hash, but it's a bit more complex...)
     - ensure some level of readability - only id would be hard to match name to scenario
    */
-  private[manager] def objectNameForScenario(processVersion: ProcessVersion, hashInput: Option[String]): String = {
+  private[manager] def objectNameForScenario(processVersion: ProcessVersion, hashInput: Option[String], releaseName: Option[String] = None): String = {
```
arkadius, line 204: 1. can you remove default value? I could cause that in some places it will be accidentally omitted
2. can it be at the beginning of parameter list like it is rendered at the end?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -52,7 +52,9 @@ case class K8sDeploymentManagerConfig(dockerImageName: String = "touk/nussknacke
                                       dockerImageTag: String = BuildInfo.version,
                                       scalingConfig: Option[K8sScalingConfig] = None,
                                       configExecutionOverrides: Config = ConfigFactory.empty(),
-                                      k8sDeploymentConfig: Config = ConfigFactory.empty())
+                                      k8sDeploymentConfig: Config = ConfigFactory.empty(),
+                                      releaseName: Option[String] = None
```
arkadius, line 56: release name is helm specific word. I wonder if it is a good option to see here technical detail of fact that nussknacker is bundled by helm. What will happen if we replace helm by sth else? Maybe it should be renamed to `objectNamePrefix`? WDYT? Also maybe it should be at the beginning. It looks like first thing that should be configured

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sUtils.scala
```diff
@@ -34,7 +34,7 @@ object K8sUtils {
   }
 
   //TODO: generate better correct name for 'strange' scenario names?
-  private[manager] def sanitizeName(base: String, canHaveUnderscore: Boolean, append: String = ""): String = {
+      private[manager] def sanitizeName(base: String, canHaveUnderscore: Boolean, append: String = ""): String = {
```
arkadius, line 37: It looks like it was better formatted before the change. Wasn't it?

PR #2850, opened 2022-02-10
Add nussknacker instance name label to runtime pods (#2849)

dswiecki:


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -158,6 +160,8 @@ object K8sDeploymentManager {
 
   val defaultParallelism = 1
 
+  val nussknackerInstanceNameLabel: String = "nussknacker.io/nussknackerInstanceName"
```
arkadius, line 163: It should be `app.kubernetes.io/instance` take a look at: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

arkadius (approved):

PR #2868, opened 2022-02-15
api splitted to components-api and extensions-api

mproch:


arkadius:
5. Maybe we should move `SpelConversionsProvider` into `extensions-api`? `NamedServiceProvider` shouldn't be also there?

arkadius (approved):

PR #2871, opened 2022-02-16
Introduce ExceptionExtractorProvider

gskrobisz:


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/exception/FlinkExceptionHandler.scala
```diff
@@ -26,31 +26,35 @@ object FlinkExceptionHandler {
 
   val exceptionHandlerConfigPath = "exceptionHandler"
   val typeConfigPath = "type"
-  val withRateMeterPath = "withRateMeter"
+  val withRateMeterConfigPath = "withRateMeter"
+  val extractorConfigPath = "exceptionExtractor"
```
arkadius, line 30: It will be hidden feature or maybe we should extends docs here: https://nussknacker.io/documentation/docs/installation_configuration_guide/model/Flink#configuring-exception-handling

arkadius (approved):

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/exception/FlinkExceptionHandler.scala
gskrobisz: Good point. I've updated the docs.

PR #2878, opened 2022-02-20
Extract utils-internal, utils cleanup

mproch:


arkadius:
Good job,
I see a few things that IMO would be great to done to make utils convention clear enough:
1. httpUtil - can u change to http-utils like it is for other cases?
2. util in name of directories - can u change to plural form like in artifact names ?
3. IMO it will be helpful to extract `component-utils` from `utils`. I know that you already added `utils-internal` which is :+1: but still I see problem, that current utils have many things that are component specifici like `EagerServiceWithStaticParameters`, `SharedService`. I would love to have tiny `utils` model with shared things. WDYT?

arkadius (approved):

PR #2879, opened 2022-02-21
Influxdb optional. In case of missing INFLUXDB_URL, reporting of metrics and counts mechanism will be disabled

arkadius:


mproch (approved):

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/metrics/dropwizard/influxdb/LiteEngineInfluxDbReporter.scala
```diff
@@ -6,22 +6,29 @@ import io.dropwizard.metrics5.{MetricName, MetricRegistry}
...
 import java.util.concurrent.TimeUnit
+import scala.util.Try
+import scala.util.control.NonFatal
 
 object LiteEngineInfluxDbReporter extends LazyLogging {
 
   import net.ceedubs.ficus.Ficus._
   import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 
   def createAndRunReporterIfConfigured(metricRegistry: MetricRegistry, prefix: MetricName, config: Config): Option[InfluxDbReporter] = {
```
dswiecki, line 17: maybe we could have some unit tests for this method

arkadius: added

PR #2883, opened 2022-02-22
Ability to change DEFAULT_SCENARIO_TYPE + some conf cleanups + stream…

arkadius:


nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -18,25 +19,50 @@ scenarioTypes {
...
 
 flinkDeploymentConfig {
-  jobManagerTimeout: 1m
   type: "flinkStreaming"
+  jobManagerTimeout: 1m
   restUrl: ${?FLINK_REST_URL}
   queryableStateProxyUrl: ${?FLINK_QUERYABLE_STATE_PROXY_URL}
 }
 
+streamingLiteDeploymentConfig {
```
mproch, line 47: Is there a reason to put it here?

arkadius: Just to be consistent with `flinkDeploymentConfig`. Should I inline both? WDYT?

mproch: I'd inline both, there is some reason to extract it in dev conf, as we share config between many configs, but here it's just noise IMO

nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -18,25 +19,50 @@ scenarioTypes {
...
-          kafkaProperties {
-            "schema.registry.url": ${?SCHEMA_REGISTRY_URL}
-          }
-        }
+        config: ${kafkaConfig}
+      }
+    }
+  }
+  "streaming-lite-embedded": {
+    deploymentConfig: ${streamingLiteDeploymentConfig}
```
dswiecki, line 27: maybe streamingLiteDeploymentConfig -> embeddedDeploymentConfig

arkadius: good point

mproch (approved):
looks cool :+1:

PR #2889, opened 2022-02-24
Add container spec override tests

kdutkowski:


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -94,6 +94,8 @@ class K8sDeploymentManager(modelData: ModelData, config: K8sDeploymentManagerCon
       ))
     } yield {
       logger.info(s"Deployed ${processVersion.processName.value}, with deployment: ${deployment.name}, configmap: ${configMap.name}")
+      logger.trace(s"K8s deployment name: ${deployment.name}: $deployment")
```
dswiecki, line 97: maybe it would be more natural if config is logged first but I not persist

arkadius (approved):

PR #2900, opened 2022-03-02
Use one ScenarioBuilder

dswiecki:

### DIFF
--- a/scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
+++ b/scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
@@ -47,6 +47,20 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
   }
 }
 
+object ScenarioBuilder {
+
+  def streaming(id: String) =
+    new ProcessMetaDataBuilder(MetaData(id, StreamMetaData()))
+
+  def streamingLite(id: String) =
+    new ProcessMetaDataBuilder(MetaData(id, LiteStreamMetaData()))
+
+  def requestResponse(id: String) =
+    new ProcessMetaDataBuilder(MetaData(id, RequestResponseMetaData(None)))
+
+}
+
+@deprecated("use ScenarioBuilder streaming method", "1.3")
 object EspProcessBuilder {
 
   def id(id: String) =
@@ -54,6 +68,7 @@ object EspProcessBuilder {
 
 }
 
+@deprecated("use ScenarioBuilder streamingLite method", "1.3")
 object StreamingLiteScenarioBuilder {
 
   def id(id: String) =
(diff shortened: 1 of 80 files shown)


scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -47,23 +46,15 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
...
     new ProcessMetaDataBuilder(MetaData(id, LiteStreamMetaData()))
 
-}
-
-object RequestResponseScenarioBuilder {
-
-  def id(id: ProcessName) =
-    new ProcessMetaDataBuilder(MetaData(id.value, RequestResponseMetaData(None)))
+  def requestResponse(id: String) =
+    new ProcessMetaDataBuilder(MetaData(id, RequestResponseMetaData(None)))
```
gadomsky, line 58: should'nt these 3 method return difrrent subtypes of `ProcessMetaDataBuilder` ? 

Now it is possible to invoke e.g.`ScenarioBuilder.requestResponse("x").parallelizm(4)` which will throw an exception. It would be more straightforward if api wouldn't allow such invocation

dswiecki: Probably API should be designed that way but after this change nothing is different to the previous approach - there still would be an exception becouse you could invoke ```RequestResponseScenarioBuilder
        .id(ProcessName("sample_scenario"))
        .parallelism(3)```

mproch: maybe we can use requestResponse/streaming methods with default parameters?

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -47,23 +47,42 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
...
+
+@deprecated("use ScenarioBuilder streaming method", "1.3")
 object EspProcessBuilder {
 
   def id(id: String) =
     new ProcessMetaDataBuilder(MetaData(id, StreamMetaData()))
 
 }
 
+@deprecated("use ScenarioBuilder streamingLite method", "1.3")
```
arkadius, line 71: IMO we can just remove those two builders - lite isn't used anywhere except our code

arkadius (approved):
Nice change

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
dswiecki: yeah, you're right but to be consistent let's remove it in 1.4 version

PR #2914, opened 2022-03-04
Rename branch to union in scenario builder

dswiecki:

### DIFF
--- a/scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/GraphBuilder.scala
+++ b/scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/GraphBuilder.scala
@@ -1,5 +1,6 @@
 package pl.touk.nussknacker.engine.build
 
+import com.github.ghik.silencer.silent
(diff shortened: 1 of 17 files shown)


mproch (approved):

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/GraphBuilder.scala
```diff
@@ -75,6 +79,21 @@ trait GraphBuilder[R] {
   def to(node: SubsequentNode): R =
     creator(node)
 
+  @silent("deprecated")
```
arkadius, line 82: branch should use join - not otherwise

arkadius (changes requested):
please fix one thing

arkadius (approved):

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/GraphBuilder.scala
dswiecki: done

PR #2915, opened 2022-03-04
Use ScenarioBuilder where possible

dswiecki:


mproch (approved):

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -47,23 +54,15 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
   }
 }
 
-object EspProcessBuilder {
+object ScenarioBuilder {
 
-  def id(id: String) =
+  def streaming(id: String = RandomStringUtils.randomAlphabetic(10)) =
```
arkadius, line 59: Let's talk about this

arkadius (changes requested):
Let's talk about random names

PR #2924, opened 2022-03-07
Extracted kafka-components-utils

arkadius:


mproch (commented):
Overall +1, but I think we need some migrationguide here?

build.sbt
```diff
@@ -686,16 +686,29 @@ lazy val kafkaUtils = (project in utils("kafka-utils")).
...
+    libraryDependencies ++= {
+      Seq(
+        "org.apache.kafka" % "kafka-clients" % kafkaV
+      )
+    }
+  // Depends on componentsApi because of dependency to NuExceptionInfo and NonTransientException -
+  // lite kafka engine handles component exceptions in runtime part
+  ).dependsOn(commonUtils % Provided, componentsApi % Provided)
+
+lazy val kafkaComponentsUtils = (project in utils("kafka-components-utils")).
```
dswiecki, line 698: please add changelog and migration guide that such a new modules is available and has classes from nussknacker-kafka-utils

arkadius: Added

PR #2957, opened 2022-03-21
Reschedule failed periodic flag

pjagielski:
Add `executionConfig.rescheduleOnFailure` flag to control whether failed deployment should be rescheduled for a next run.
Made `FailedOnDeploy` status final after retries, and introduced `RetryingDeploy` for retries.


jedrz (approved):
LGTM :+1: 
I think we can add an entry in changelog?

PR #2967, opened 2022-03-24
Create json-util module

Filemon279:


build.sbt
```diff
@@ -843,6 +843,12 @@ lazy val testUtils = (project in utils("test-utils")).
     }
   )
 
+lazy val jsonUtils = (project in utils("json-utils")).
```
mproch, line 846: can we have everit-json-schema dependency here, instead of components-utils?

Filemon279: We even should, thanks :)

mproch (approved):

PR #2968, opened 2022-03-25
Enable request-response sink with editor

mproch:


engine/lite/request-response/app/src/test/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpAppSpec.scala
```diff
@@ -100,7 +107,7 @@ class RequestResponseHttpAppSpec extends RequestResponseHttpTest {
 
       Post(s"/${procId.value}", toEntity(Request("a", "b"))) ~> processesRoute ~> check {
         status shouldBe StatusCodes.OK
-        responseAs[String] shouldBe "[\"b\"]"
+        responseAs[String] shouldBe """[{"Value":"b"}]"""
```
Filemon279, line 110: maybe for consistency we could use here `SinkValueParamName` ?

Filemon279 (approved):

PR #2981, opened 2022-03-30
Improve avro serialization docs

dswiecki:

### DIFF
--- a/docs/operations_guide/Common.md
+++ b/docs/operations_guide/Common.md
@@ -207,7 +207,7 @@ To see the exact error you can look at (depending on the configuration of the Ex
 }
 ```
 #### Avro serialization problem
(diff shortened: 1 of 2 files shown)


docs/operations_guide/Common.md
```diff
@@ -206,7 +206,13 @@ To see the exact error you can look at (depending on the configuration of the Ex
   "additionalData": {}
 }
 ```
-
+#### Avro serialization problem
+Common problem is when your AVRO data scheme not correspond with the one Nussknacker expects. You will find then `Unknown magic byte!` error indicating that data you produced to the topic are not serialized properly.
```
mproch, line 210: I would write "Avro data encoding" or sth similar, as it's not about schema itself?

mproch (approved):

PR #2983, opened 2022-03-30
Improvements: Extract Security Api

lciolecki:
`pl.touk.nussknacker.ui.security.api.Permission` (security module) moved to `pl.touk.nussknacker.security.Permission` (extension-api module)

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -2,6 +2,7 @@
 
 1.4.0 (Not released yet)
 ------------------------
+* [#2983](https://github.com/TouK/nussknacker/pull/2983) Extract Permission to extensions-api
 * [#2996](https://github.com/TouK/nussknacker/pull/2996) Flink job registration optimization - do not prepare interpretation function for ending node
 * [#3010](https://github.com/TouK/nussknacker/pull/3010) Feature: Docker Java Debug Option
 
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -8,6 +8,11 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 * `security.rolesClaim` changed to `security.rolesClaims`, type changed to list of strings 
 
+### Code API changes
+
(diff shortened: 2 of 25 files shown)


docs/MigrationGuide.md
```diff
@@ -2,6 +2,14 @@
 
 To see the biggest differences please consult the [changelog](Changelog.md).
 
+## In version 1.4.0 (Not released yet)
+
+### Code API changes
+
+* [#2983](https://github.com/TouK/nussknacker/pull/2983) Extract security api module (Permission, GlobalPermission)
+  * `pl.touk.nussknacker.ui.security.api.Permission` (security module) moved to `pl.touk.nussknacker.security.api.Permission` (extension-api module)
```
dswiecki, line 10: we've created APIs mostly without changing packages, if we change package here maybe lets put it in right package, do we indroduce the extension package @mproch wdyt ?

mproch: I think security is not currently considered as "api", so it may be good time to move stuff around. Or try to be backward compatible via depracted class/types in current package?

docs/Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+1.4.0 (Not released yet)
+------------------------
+* [#2983](https://github.com/TouK/nussknacker/pull/2983) Extract security api module (Permission, GlobalPermission)
```
mproch, line 5: it's not a module. 
`Extract Permission to extensions-api`?

docs/MigrationGuide.md
```diff
@@ -2,6 +2,13 @@
 
 To see the biggest differences please consult the [changelog](Changelog.md).
 
+## In version 1.4.0 (Not released yet)
+
+### Code API changes
+
+* [#2983](https://github.com/TouK/nussknacker/pull/2983) Extract security api module (Permission, GlobalPermission)
```
mproch, line 9: only Permission was extracted

PR #3006, opened 2022-04-13
fixes to 'Restarting' status

gadomsky:


mproch (approved):
Let's add sth to the changelog?

PR #3007, opened 2022-04-13
Fix: ProcessService should return only not archived scenarios

lciolecki:


JulianWielga (changes requested):
I think it's FE problem

lciolecki:
> I think it's FE problem

Why do you think that?

mproch (approved):
@JulianWielga this should be handled on BE... we should just ignore archived scenarios...

JulianWielga (approved):

PR #3008, opened 2022-04-14
fixed components archived usages styles & filter

JulianWielga:


lciolecki (changes requested):
I think it's BE problem :)

PR #3029, opened 2022-04-22
SchemaRegistry caching config from config instead of hardcoded

arkadius:


engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaComponentProvider.scala
```diff
@@ -38,6 +38,10 @@ class LiteKafkaComponentProvider extends ComponentProvider {
     val schemaRegistryTypedJson = "DataSourcesAndSinks#schema-registry--json-serialization"
     val noTypeInfo = "DataSourcesAndSinks#no-type-information--json-serialization"
 
+    val cacheConfig = KafkaConfig.parseConfigOpt(dependencies.config).map(_.schemaRegistryCacheConfig).getOrElse(SchemaRegistryCacheConfig())
```
mproch, line 41: let's use config, not dependencies.config, dependencies.config should be removed in the future probably

docs/installation_configuration_guide/ModelConfiguration.md
```diff
@@ -150,17 +150,20 @@ Important thing to remember is that Kafka server addresses/schema registry addre
...
+| topicsExistenceValidationConfig.validatorConfig.topicsFetchCacheTtl           | Low        | duration         | 30 seconds       | TTL for caching list of existing topics                                                                                                                                                                                                                      |
+| topicsExistenceValidationConfig.validatorConfig.adminClientTimeout            | Low        | duration         | 500 milliseconds | Timeout for communicating with Kafka cluster                                                                                                                                                                                                                 |
+| schemaRegistryCacheConfig.availableSchemasExpirationTime                      | Low        | duration         | 10 seconds       | How often available schemas cache will be invalidated. This determine how long you will wait after adding new schema or new schema version until it will be available in Designer                                                                            |
```
mproch, line 164: This determines the maximum time you'll have to wait after ...

benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/serialization/avro/AvroBenchmark.scala
```diff
@@ -41,7 +38,7 @@ class AvroBenchmark {
       val parsedSchema = ConfluentUtils.convertToAvroSchema(AvroSamples.sampleSchema, Some(1))
       schemaRegistryMockClient.register("foo-value", parsedSchema, 1, AvroSamples.sampleSchemaId)
       val factory: CachedConfluentSchemaRegistryClientFactory =
-        new CachedConfluentSchemaRegistryClientFactory(CacheConfig.defaultMaximumSize, None, None) {
+        new CachedConfluentSchemaRegistryClientFactory(SchemaRegistryCacheConfig.noExpire) {
```
mproch, line 41: I think, maybe we need an entry in changelog? Or also in MigrationGuide, as those are -utils?

mproch (commented):
Nice :+1: added minor comments

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaComponentProvider.scala
arkadius: Be aware that in lite we use kafka inside `modelConfig`. You probable have in mind `FlinkKafkaComponentProvider` where we have `TemporaryKafkaConfigMapping` and indeed we can use component's `config` but ... it will be `config.getConfig("config")` because we have one extra level of nesting. Because of this fact I've decided to leave it consistent with other places where kafka config is used - let's fix all places when we get rid of `TemporaryKafkaConfigMapping`. Maybe we decide to remove this nested level. WDYT?

benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/serialization/avro/AvroBenchmark.scala
arkadius: Good point, added both

mproch (approved):
Cool, thx :+1:

PR #3040, opened 2022-04-28
Fix missing metrics for end nodes

jedrz:


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/EndNodeInterpretationFunction.scala
```diff
@@ -0,0 +1,28 @@
+package pl.touk.nussknacker.engine.process.registrar
+
+import org.apache.flink.api.common.functions.{RichMapFunction, RuntimeContext}
+import org.apache.flink.configuration.Configuration
+import pl.touk.nussknacker.engine.api.{Context, MetaData, ProcessListener}
+
+private[registrar] class EndNodeInterpretationFunction(prepareListeners: RuntimeContext => Seq[ProcessListener],
```
mproch, line 7: I think it suffices to use ProcessPartFunction? It opens only needed services. Or at least let's make sth similar to ExceptionHandlerFunction?

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/ProcessListener.scala
```diff
@@ -20,6 +22,7 @@ trait ProcessListener extends Lifecycle {
                      params: Map[String, Any],
                      result: Try[Any]): Unit
 
+  // TODO: deprecated? remove?
```
mproch, line 25: Let's add deprecation for now

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/ProcessListener.scala
```diff
@@ -8,6 +8,8 @@ trait ProcessListener extends Lifecycle {
 
   def nodeEntered(nodeId: String, context: Context, processMetaData: MetaData): Unit
 
+  def endEncountered(lastNodeId: String, context: Context, processMetaData: MetaData): Unit
```
mproch, line 11: We need entry in migration guide

mproch (approved):
:+1:

PR #3044, opened 2022-04-28
fix docs after moving attachments to db

gadomsky:


docs/operations_guide/Flink.md
```diff
@@ -320,7 +320,6 @@ What’s more, increasing parallelism above the count of partitions of the Kafka
 * Flink Savepoints/checkpoints path should have some retention mechanism configured (e.g. S3 has some retention configuration options) - savepoints are not deleted automatically, while checkpoints are not deleted in case of e.g. jobmanager failures
 * Logs retention
 * Nussknacker stores following data on filesystem (in /opt/nussknacker/storage for default docker configuration) - make sure to configure appropriate backups/persistent volumes
-  * Configured location of scenario attachments
```
arkadius, line 323: We have only one case so maybe more gramatically correct would be if you write just "Nussknacker stores database file when using HSQL on filesystem ... ). WDYT?

arkadius (approved):

PR #3049, opened 2022-05-09
Add collector node for lite engine

Filemon279:


engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
```diff
@@ -29,8 +29,9 @@ class ForEachTransformerComponent(elements: LazyParameter[java.util.Collection[A
     val interpreter = context.interpreter.syncInterpretationFunction(elements)
     (ctx: Context) => {
       val partsToRun = interpreter(ctx)
-      val partsToInterpret = partsToRun.asScala.toList.map { partToRun =>
-        ctx.withVariable(outputVariable, partToRun)
+      val partsToInterpret = partsToRun.asScala.toList.zipWithIndex.map { case (partToRun, index) =>
+        val parentContextId = ctx.parentContext.fold("")(c => s"${c.id}_")
```
arkadius, line 33: Why this parent context part is important in new context id?

engine/lite/components/base/src/main/scala/pl/touk/nussknacker/engine/lite/components/ForEachTransformer.scala
```diff
@@ -29,8 +29,9 @@ class ForEachTransformerComponent(elements: LazyParameter[java.util.Collection[A
...
+        ctx.withVariable(outputVariable, partToRun).copy(id=s"${ctx.id}_$parentContextId$index")
```
arkadius, line 34: I've taken a look into `ContextIdGenerator` and generated context id is dash saparated. Maybe would be better to separate index also with dash? WDYT?

engine/lite/request-response/runtime/src/test/scala/pl/touk/nussknacker/engine/requestresponse/test/RunnableEspScenario.scala
```diff
@@ -0,0 +1,81 @@
...
+import pl.touk.nussknacker.engine.resultcollector.ProductionServiceInvocationCollector
+import pl.touk.nussknacker.engine.testing.LocalModelData
+import pl.touk.nussknacker.engine.testmode.TestProcess.TestResults
+import pl.touk.nussknacker.test.PatientScalaFutures
+
+import java.nio.charset.StandardCharsets
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Using
+
+trait RunnableEspScenario extends Matchers with PatientScalaFutures {
```
arkadius, line 28: Have u taken a look into `LiteTestScenarioRunnerSpec` and `RequestResponseTestScenarioRunner` in our brand new `lite-components-testkit` module? We have plans to have one, standarized, well documented way to test components by running scenarios. We planned to make it quite runtime agnostic - no matter if you run your test on flink, streaming-lite or req-resp. We wanted to make it such easy as writing tests for akka http resources, spring boot integration test and so on. We've just started a work on that in previous release but maybe you could take a look into it and eventually extend it if you see sth worth adding?

Filemon279: That is nice, looks like something I wanted to created here but it is more generic - I haven't seen that. Do you plan to keep tests for component like `for-each` or `collector` inside `lite-components-testkit` or use it as test dependency? I am asking since `liteComponentsTestkit` depends on `requestResponseRuntime` so project `requestResponseRuntime` cannot depend on dependency that depends on it - or am I missing something?

arkadius: The plan is that testkit is a dependency that will be added to projects with coponents. testkit depnends on runtime because it need runtime to execute scenario :) All components like `fore-each` etc. are in separate modules which are not dependants of runtime so I guess that everything should work as expected?

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -157,6 +157,22 @@ UnionMemo merges multiple branches into one stream. For each incoming branch two
 }
 ```
 
+## Collector
+
+![for_each](img/collector.png)
```
arkadius, line 162: collector

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -157,6 +157,22 @@ UnionMemo merges multiple branches into one stream. For each incoming branch two
...
+
+`collector` collects values from nodes that executed multiple times (e.g. for-each subsequent nodes) and store them in a list.
+
+Collector takes one argument:
+- Input expression - expression which will be collected from all nodes invocations. 
+
+For example:
+- We use `for-each` component on list `{"one", "two", "three"}`
+- Attach bellow `for-each` some node which do `.size` on each element and returns `#elementSize`
```
arkadius, line 171: 1. Attach -> Connect/Add ? 
2. which do `#element.size` on each string maybe? To show more precisely final expression and to clarify that it meant to be size of those string elements - not list of strings

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -157,6 +157,22 @@ UnionMemo merges multiple branches into one stream. For each incoming branch two
 }
 ```
 
+## Collector
+
+![collector](img/collector.png)
+
+`collector` collects values from nodes that executed multiple times (e.g. for-each subsequent nodes) and store them in a list.
```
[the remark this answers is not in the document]
Filemon279: You are right, it is not working as we would expect by looking at a request-response. I have added a comment about that in a docs. Hope that it should be sufficient (?)

arkadius (approved):

PR #3063, opened 2022-05-17
Add PrometheusJmxExporter to docker images

gadomsky:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,9 @@
 import com.typesafe.sbt.packager.SettingsHelper
 import com.typesafe.sbt.packager.docker.DockerPlugin.autoImport.dockerUsername
+import com.typesafe.sbt.packager.docker.ExecCmd
 import pl.project13.scala.sbt.JmhPlugin
 import pl.project13.scala.sbt.JmhPlugin._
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -295,11 +297,22 @@ val sttpV = "2.2.9"
...
 val monocleV = "2.1.0"
 
-lazy val commonDockerSettings = {
+lazy val commonDockerSettings = (workingDir: String) => {
   Seq(
     dockerBaseImage := "openjdk:11-jre-slim",
     dockerUsername := dockerUserName,
     dockerUpdateLatest := dockerUpLatestFromProp.getOrElse(!isSnapshot.value),
+    dockerCommands := {
+      val (before, after) = dockerCommands.value.splitAt(dockerCommands.value.size - 3) // 3rd from the end command is setting USER, and we want to do the following yet as root
```
mproch, line 306: Let's do it via dist, so it doesn't clutter docker and is available also in tar distribution. E.g via https://github.com/sbt/sbt-javaagent ?

mproch (approved):
:+1:  Let's start with opinionated version :)

PR #3071, opened 2022-05-20
Improvements: Avro Sink Validation

lciolecki:
* Fix optional schema validation
* Remove `ValidationMode.strict`
* More strict Avro Schema validation vs TypingResult


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+    val schemaAsTypedResult = AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
+
+    (schemaAsTypedResult, typingResult) match {
+      case (TypedClass(schemaClass, _), TypedClass(typeClass, _)) if typeClass.equals(schemaClass) => valid
+      case (_, TypedClass(typeClass, _)) if allowedClasses.contains(typeClass) => valid
+      case _ => invalid(typingResult, schema, path) //we don't use canBeSubclassOf here, because it can return true eg. Integer (Number) vs BigDecimal but Avro doesn't allow for that...
+    }
+  }
+
+  private def canBeSubclassOf(objTypingResult: TypingResult, schema: Schema, path: Option[String])(implicit isNullable: Boolean = false): ValidatedNel[OutputValidatorError, Unit] = {
```
mproch, line 223: let's not pass boolean as implicit with default value...

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+      case (array@TypedClass(cl, _), Type.ARRAY) if classOf[java.util.List[_]].isAssignableFrom(cl) =>
+        validateArraySchema(array, schema, path)
+// Right now we can't handle null value.. Typing for null is Unknown and canBeSubclassOf returns for it always true..
+//      case (Unknown, _) if !schema.isNullable => //null is converted to Unknown, Unknown is base type and canBeSubclassOf of any Type
+//        invalid(typingResult, schema, path)
+//      case (Unknown, _) => //situation when we pass null as value and schema is null or can be nullable
+//        valid
+      case (union: TypedUnion, _) if union.isEmptyUnion && schema.isNullable => //situation when we pass #input witch null schema as value
+        valid
+      case (TypedClass(cl, _), Type.ENUM) if classOf[java.lang.String].isAssignableFrom(cl) =>
```
mproch, line 87: why do we need this if here, and in FIXED there isn't one?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+      validateTypingResult(value, schema.getValueType, fieldPath)(isNullable = false)
+    }.foldLeft[ValidatedNel[OutputValidatorError, Unit]](().validNel)((a, b) => a combine b)
+
+    schemaFieldsValidation
+  }
+
+  private def validateArraySchema(array: TypedClass, schema: Schema, path: Option[String])(implicit isNullable: Boolean) = {
+    def validateListElements(params: List[TypingResult]) = {
+      params
+        .distinct
```
mproch, line 158: why distinct? I think this list should never have > 1 element?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+      results.foldLeft[ValidatedNel[OutputValidatorError, Unit]](().validNel)((a, b) => a combine b)
+    }
+  }
+
+  private def validateClass[T:ClassTag](typingResult: TypingResult, schema: Schema, path: Option[String]): Validated[NonEmptyList[OutputValidatorTypeError], Unit] = {
+    val clazz: Class[_] = implicitly[ClassTag[T]].runtimeClass
+    validateClass(typingResult, schema, List(clazz), path)
+  }
+
+  private def validateClass(typingResult: TypingResult, schema: Schema, allowedClasses: List[Class[_]], path: Option[String]): Validated[NonEmptyList[OutputValidatorTypeError], Unit] = {
```
mproch, line 213: do we ever use it with > 1 class?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+    val schemaAsTypedResult = AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
+
+    (schemaAsTypedResult, typingResult) match {
```
mproch, line 216: it would be easier for me to understand if it was just
(schemaAsTypedResult :: additionalAllowedClasses).contains(typedClass)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/OutputValidatorErrorsConverter.scala
```diff
@@ -0,0 +1,72 @@
...
+
+}
+
+class OutputValidatorErrorsConverter(schemaParamName: String) {
+
+  import OutputValidatorErrorsMessageFormatter._
+
+  final def convertValidationErrors(errors: NonEmptyList[OutputValidatorError])(implicit nodeId: NodeId): CustomNodeError = {
+    val accumulator = (List.empty[OutputValidatorMissingFieldsError], List.empty[OutputValidatorRedundantFieldsError], List.empty[OutputValidatorTypeError])
+    val (missingFieldsError, redundantFieldsError, typeFieldsError) = errors.toList.foldRight(accumulator) {
```
mproch, line 25: I wonder is it worthy to write this whole complex grouping logic instead of some simpler list...

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
```diff
@@ -94,4 +52,473 @@ class LiteKafkaAvroFunctionalTest extends FunSuite with Matchers with ValidatedV
 
     new LiteKafkaTestScenarioRunner(mockSchemaRegistryClient, mockedComponents, config)
   }
+
+  test("should test end to end kafka avro record data at sink / source") {
```
mproch, line 56: Do we have some sanity check for "for all avro schemas if we do #input => #input with same schemas on input and output, it will work"?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
```diff
@@ -1,17 +1,15 @@
 package pl.touk.nussknacker.engine.avro.encode
 
-case class ValidationMode private(name: String, label: String, acceptUnfilledOptional: Boolean, acceptRedundant: Boolean)
+case class ValidationMode private(name: String, label: String, acceptRedundant: Boolean)
 
 object ValidationMode {
 
-  val strict: ValidationMode = ValidationMode("strict", "Strict mode", acceptUnfilledOptional = false, acceptRedundant = false)
+  val allowOptional: ValidationMode = ValidationMode("allowOptional", "Allow missing optional parameters", acceptRedundant = false)
```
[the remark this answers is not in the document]
lciolecki: For me strict is ok, but earlier I discussed with @arkadius about this and we decided to use allowOptional. @arkadius WDYT?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidatorPrinter.scala
```diff
@@ -0,0 +1,82 @@
+package pl.touk.nussknacker.engine.avro.encode
+
+import org.apache.avro.{LogicalType, LogicalTypes, Schema}
+import pl.touk.nussknacker.engine.api.process.{OutputValidatorErrorsConverter, OutputValidatorErrorsMessageFormatter}
+import pl.touk.nussknacker.engine.avro.schema.AvroStringSettings
+import pl.touk.nussknacker.engine.avro.typed.AvroSchemaTypeDefinitionExtractor
+
+import scala.collection.JavaConverters
+
+object AvroSchemaOutputValidatorPrinter {
```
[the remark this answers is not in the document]
lciolecki: But.. Does our TypingResult display in good way information about avro type? TypingResult can't display information about nullability..

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/OutputValidatorErrorsConverter.scala
```diff
@@ -0,0 +1,72 @@
+package pl.touk.nussknacker.engine.api.process
+
+import cats.data.NonEmptyList
+import pl.touk.nussknacker.engine.api.NodeId
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.CustomNodeError
+import pl.touk.nussknacker.engine.api.typed.typing.{TypedUnion, TypingResult}
+
+object OutputValidatorErrorsConverter {
```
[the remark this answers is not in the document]
lciolecki: Because in future we could use it also to convert errors from: json / other schema validation result - that was a idea.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ClassExtractionSettings.scala
```diff
@@ -221,7 +221,10 @@ object ClassExtractionSettings {
         Pattern.compile(s"clockSequence|randomUUID|fromString|getLeastSignificantBits|getMostSignificantBits|node|timestamp|$ToStringMethod|variant|version")),
       ClassMemberPatternPredicate(
         SuperClassPredicate(ExactClassPredicate(classOf[Traversable[_]], classOf[Option[_]])),
-        Pattern.compile(s"apply|applyOrElse|contains|get|getOrDefault|indexOf|isDefined|isEmpty|size|values|keys|diff"))
+        Pattern.compile(s"apply|applyOrElse|contains|get|getOrDefault|indexOf|isDefined|isEmpty|size|values|keys|diff")),
+      ClassMemberPatternPredicate(
+        ClassPatternPredicate(Pattern.compile("scala\\.concurrent\\.duration\\..*")),
```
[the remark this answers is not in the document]
lciolecki: Ups... My mistake, I've tested adding java.nio.ByteBuffer / other stuff to SpEL context and I forget remove it.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/AvroUtils.scala
```diff
@@ -108,28 +108,43 @@ object AvroUtils extends LazyLogging {
...
+      case (any, Schema.Type.UNION) =>
+        schema.getTypes.asScala.filterNot(_.isNullable).map(createValue(any, _)).headOption.orNull
       case (_, _) => value
     }
 
+    val fields = schema.getFields.asScala.map(_.name()).toSet
     val builder = new LogicalTypesGenericRecordBuilder(schema)
-    data.foreach{ case (key, value) =>
+    
+    data
```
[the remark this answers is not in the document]
lciolecki: Yep, it's just helper -> if we put there redundant fields, we just ignore them on record building process.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/OutputValidatorErrorsConverter.scala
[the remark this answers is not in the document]
mproch: still, let's have it in component-utils?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
mproch: I thought more about additionalAllowedClasses: TypingResult etc,

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidatorPrinter.scala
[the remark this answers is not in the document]
mproch: at the moment - probably not, but who knows the future... I know we can display nullability this way, just wonder if it's good to introduce 100LOC and new notation just for that...

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,236 @@
...
+  )
+
+  /**
+    * see {@link pl.touk.nussknacker.engine.avro.encode.BestEffortAvroEncoder} for underlying avro types
+    */
+  def validateTypingResultToSchema(typingResult: TypingResult, parentSchema: Schema)(implicit nodeId: NodeId): ValidatedNel[OutputValidatorError, Unit] =
+    validateTypingResult(typingResult, parentSchema, None)(isNullable = false)
+
+  /**
+    * isNullable is kind of workaround... In nullable nested schemas ([null, Record] / [null, Map] / [null, List]) we
```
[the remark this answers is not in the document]
lciolecki: I assume we want to present information about schema in best way.. So, please imagine situation when you have union: [null, List[Any]] and when someone put there just Integer or other type then will be nice so information about that field type is null | List[Any]. 

This is use only when we have nullable union, because when we do comparing inside object then we want to see information only about comparing elements, we don't want to see information about general nullability of root object -> it's difficult to explain, but maybe we can just talk about it?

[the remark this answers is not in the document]
lciolecki: After discussion we decided to remove it.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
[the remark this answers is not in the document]
lciolecki: I've changed it.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidatorPrinter.scala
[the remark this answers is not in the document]
lciolecki: This object display proper all allowed types for schema.. In my opinion it's necessary to use it if we want to show proper type for avro schema - we will not achieve it using TypingResulg.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/AvroUtils.scala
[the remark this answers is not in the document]
lciolecki: I replaced this tool by `BestEffortAvroEncoder`

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
lciolecki: I've added two property based test for #input -> #input, and some SpEL as output.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/AvroUtils.scala
```diff
@@ -107,16 +107,16 @@ object AvroUtils extends LazyLogging {
...
+      case (map: collection.Map[String@unchecked, _], Schema.Type.RECORD) =>
+        createRecord(schema, map)
+      case (collection: Traversable[_], Schema.Type.ARRAY) =>
+        collection.map(createValue(_, schema)).toList.asJava
+      case (map: collection.Map[String@unchecked, _], Schema.Type.MAP) =>
         map.mapValues(createValue(_, schema)).asJava
-      case (str: String, Schema.Type.ENUM) =>
-        new GenericData.EnumSymbol(schema, str)
+      case (any, Schema.Type.UNION) if schema.isNullable && any != null =>
+        schema.getTypes.asScala.filterNot(_.isNullable).map(createValue(any, _)).toList.head
```
arkadius, line 119: It will blow for union of `["array", "record"]` and record on input

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/SinkOutputSpELConverter.scala
```diff
@@ -0,0 +1,46 @@
...
+
+trait SinkOutputSpELConverter {
+
+  import collection.JavaConverters._
+
+  val Input: String = "#input"
+  val EmptyRecord: String = "{:}"
+  val EmptyList: String = "{}"
+
+  def convert(data: Any, isField: Boolean = false): String = {
```
mproch, line 13: maybe we can add .toSpel implicit?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/SinkOutputSpELConverter.scala
```diff
@@ -0,0 +1,46 @@
...
+      case collection: Traversable[_] =>
+        val elements = collection.toList.map(v => convert(v, isField = true))
+        convertCollection(elements)
+      case collection: util.Collection[_] =>
+        convert(collection.asScala)
+      case str: String if str == Input => str
+      case spel: String if spel.startsWith("T(") => spel
+      case spel: String if spel.startsWith("{") => spel
+      case str: String if isField => s""""$str""""
+      case str: String if !isField => s"'$str'"
```
mproch, line 35: don't think it's needed...

lciolecki: Hmm, you're right ;)

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -10,37 +10,37 @@ import java.nio.ByteBuffer
...
-          .filterNot(field => skipOptionalFields && field.hasDefaultValue)
           .map(field => field.name() -> typeDefinition(field.schema(), possibleTypes))
           .toList
 
         Typed(possibleTypes.map(pt => TypedObjectTypingResult(fields, pt)))
       }
-      case Schema.Type.ENUM =>  //It's should by Union, because output can store map with string for ENUM
-        Typed(Set(Typed.typedClass[EnumSymbol], AvroStringSettings.stringTypingResult))
+      case Schema.Type.ENUM =>
+        Typed.typedClass[EnumSymbol]
```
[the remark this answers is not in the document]
lciolecki: It depends.. If you need string you have to do `.toString`. Before this change type enum hadn't worked properly: it was represented as Union[EnumSymbol, String], but when you did some comparing enum to string the result always was false, because in runtime there was always EnumType. Right now we are more strict, we allow to put string when avro contains enum but we don't allow comparing, use enum where string is required.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,290 @@
...
+    }.foldLeft[ValidatedNel[OutputValidatorError, Unit]](().validNel)((a, b) => a combine b)
+
+    schemaFieldsValidationResult
+  }
+
+  private def validateArraySchema(array: TypedClass, schema: Schema, path: Option[String]) = {
+     val valuesValidationResult = array
+        .params
+        .zipWithIndex
+        .map { case (el, index) =>
```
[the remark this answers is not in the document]
lciolecki: I don't know? What does mean situation when `param > 1` ? In fact when there will be more than one element then we will check all elements and I think it's okey - but maybe I don't know something about Typing?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,290 @@
...
+    validateWithValue[java.lang.String](isValueValid, typingResult: TypingResult, schema: Schema, path: Option[String])
+  }
+
+  private def validateWithValue[T:ClassTag](isValueValid: (TypedObjectWithValue, Schema) => Boolean, typingResult: TypingResult, schema: Schema, path: Option[String]) = {
+    val typeResult = validateClass[T](typingResult, schema, path)
+
+    typingResult match {
+      case obj: TypedObjectWithValue if typeResult.isValid && isValueValid(obj, schema) => valid
+      case _: TypedObjectWithValue => invalid(typingResult, schema, path)
+      case _ if typeResult.isValid => valid
```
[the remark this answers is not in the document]
lciolecki: Ups. I think it's rubbish - thanks.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,290 @@
...
+      }
+    }
+
+    validateWithValue[java.lang.String](isValueValid, typingResult: TypingResult, schema: Schema, path: Option[String])
+  }
+
+  private def validateWithValue[T:ClassTag](isValueValid: (TypedObjectWithValue, Schema) => Boolean, typingResult: TypingResult, schema: Schema, path: Option[String]) = {
+    val typeResult = validateClass[T](typingResult, schema, path)
+
+    typingResult match {
```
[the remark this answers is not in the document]
lciolecki: Hmm, If you want so ;)

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/AvroSchemaCreator.scala
```diff
@@ -0,0 +1,59 @@
+package pl.touk.nussknacker.engine.avro
+
+import org.apache.avro.{LogicalType, LogicalTypes, Schema}
+import org.apache.avro.Schema.{Field, Type}
+
+import scala.util.Random
+
+object AvroSchemaCreator {
```
arkadius, line 8: It shouldn't be in avro-component-utils. Why not just move to test classes?

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
```diff
@@ -179,14 +169,12 @@ class AvroSchemaSpelExpressionSpec extends FunSpec with Matchers {
     val schema = wrapWithRecordSchema(
       """[
         |  { "name": "stringField", "type": "string" },
-        |  { "name": "mapField", "type": { "type": "map", "values": "string" } },
-        |  { "name": "enumField", "type": { "type": "enum", "name": "sampleEnum", "symbols": ["One", "Two"] } }
+        |  { "name": "mapField", "type": { "type": "map", "values": "string" } }
```
arkadius, line 172: Why you removed enum from this test? It was to hard to prepare good test data after changes in enum handling?

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
```diff
@@ -138,26 +138,16 @@ class AvroSchemaSpelExpressionSpec extends FunSpec with Matchers {
 
   it("should recognize record with enum") {
     val ctx = ValidationContext.empty.withVariable("input", AvroSchemaTypeDefinitionExtractor.typeDefinition(PaymentV1.schema), paramName = None).toOption.get
-
-    parse[CharSequence]("#input.currency", ctx) should be ('valid)
```
arkadius, line 142: Why it is out?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  private def validateClass[T:ClassTag](typingResult: TypingResult, schema: Schema, path: Option[String]): Validated[NonEmptyList[OutputValidatorTypeError], Unit] = {
+    val schemaAsTypedResult = AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
+    val clazz: Class[_] = implicitly[ClassTag[T]].runtimeClass
+
+    (schemaAsTypedResult, typingResult) match {
+      case (TypedClass(schemaClass, _), TypedClass(typeClass, _)) if typeClass.equals(schemaClass) => valid
+      case (_, TypedClass(typeClass, _)) if clazz.equals(typeClass) => valid
+      case (_, TypedObjectWithValue(underlying, _)) if underlying.klass.equals(clazz) => valid
```
arkadius, line 267: Can u change the order to be the same as above to be visible that it is the same case? I mean clazz == underlying.klass

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+      case _ => invalid(typingResult, schema, path) //we don't use canBeSubclassOf here, because it can return true eg. Integer (Number) vs BigDecimal but Avro doesn't allow for that...
+    }
+  }
+
+  /**
+    * TODO: Consider verification class instead of using .canBeSubclassOf from Typing - at avro we want to avoid:
+    * * Unknown.canBeSubclassOf(Any) => true
+    * * Long.canBeSubclassOf(Integer) => true
```
arkadius, line 275: I can't see tests for that. What happen if someone pass record with long on input and try to save it as a int and vice versa? Validation will accept that and what about runtime?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+    def createUnionValidationResults(withoutNullability: Boolean): List[ValidatedNel[OutputValidatorError, Unit]] = {
+      val schemas = if(withoutNullability) schema.getTypes.asScala.filterNot(_.isNullable) else schema.getTypes.asScala
+      schemas.map(validateTypingResult(union, _, path)).toList
+    }
+
+    val results = schema match {
+      case UnionNullableSchemaType(_) => //Nullability: union[null, any] it's most common situation, so we want to handle that in special way
+        val notNullableResults = createUnionValidationResults(withoutNullability = true)
+        notNullableResults.toValidated match {
+          case Invalid(errors) if singleObjectTypingError(errors) => //when single typing error is true, we have to validate again including nullability
```
arkadius, line 193: 1. Why UnionNullableSchemaType is special case?
2. Why u use extractor if you don't use extracted schema?
3. Why we have to "validate again including nullability"?
4. double negation is hard - I mean by "including nullability" you do withoutNullability = false

scenario-api/src/main/scala/pl/touk/nussknacker/engine/spel/Implicits.scala
```diff
@@ -1,9 +1,53 @@
...
+      case map: util.Map[String@unchecked, _] =>
+        toSpEL(map.asScala)
+      case collection: Traversable[_] if collection.isEmpty =>
+        EmptyList
+      case collection: Traversable[_] =>
+        val elements = collection.toList.map(v => toSpEL(v))
+        convertCollection(elements)
+      case collection: util.Collection[_] =>
+        toSpEL(collection.asScala)
+      case str: String if str == Input => str
```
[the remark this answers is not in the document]
lciolecki: Uuuu - of corse you're right ;) I'll change it - thanks :)

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/AvroSchemaCreator.scala
lciolecki: I thought this object will be helpfully in other tests, we don't have any avro-test-utils, so I decided to put it here :) But if you insist?

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
lciolecki: Because enum isn't string. I've changed it from union [EnumSymbol, String] to just EnumSymbol - before it was done only for allow to pass enum in sink where field is string. Right now validation can handle it and we don't have to do hacks with union.

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
lciolecki: EnumSymbol isn't String.

lciolecki: if you want it so badly, I can rollback schema to previous version and do something like this:

`parse[CharSequence]("#input.currency.toString", ctx) should be ('valid)` 

but does it have sens?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+      val schemaAsTypedResult = AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
+      condNel(objTypingResult.canBeSubclassOf(schemaAsTypedResult), (),
+        OutputValidatorTypeError(path.getOrElse(SimpleAvroPath), objTypingResult, AvroSchemaExpected(schema))
+      )
+  }
+
+  private def invalid(typingResult: TypingResult, schema: Schema, path: Option[String]): ValidatedNel[OutputValidatorTypeError, Nothing] =
+    Validated.invalidNel(OutputValidatorTypeError(path.getOrElse(SimpleAvroPath), typingResult, AvroSchemaExpected(schema)))
+
+  private def buildPath(key: String, path: Option[String], isGeneric: Boolean = false) = Some(
```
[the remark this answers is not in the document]
lciolecki: Generic mean generic / nested objects. But I'll change it.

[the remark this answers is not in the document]
lciolecki: Hmm, but right now when I'm looking at it.. `useIndexer` is definitely better :)

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+   requiredFieldsValidation combine schemaFieldsValidation combine redundantFieldsValidation
+  }
+
+  private def validateTypedClassToMapSchema(map: TypedClass, schema: Schema, path: Option[String]) = {
+    map.params match {
+      case _ :: value :: Nil =>
+        validateTypingResult(value, schema.getValueType, buildPath("*", path, isGeneric = true))
+      case _ => //TODO: will be there more then two parameters?
```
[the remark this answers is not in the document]
lciolecki: I will return:
` case _ =>
        invalid(map, schema, path)
        `

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
lciolecki: Ok there were some small mistake, but lets see first tests:
```
      (sConfig(sampleInteger, longSchema, integerSchema, Input), valid(sampleInteger)), //Long -> Int?
      (sConfig(sampleLong, longSchema, integerSchema, Input), valid(sampleLong.toInt)), //Long -> Int?
      (sConfig(sampleBoolean, booleanSchema, integerSchema, sampleLong), valid(sampleLong.toInt)), //Long -> Int?
      ```

File: utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala [hunk 30]
```diff
@@ -0,0 +1,292 @@
...
+  def validateTypingResultToSchema(typingResult: TypingResult, parentSchema: Schema)(implicit nodeId: NodeId): ValidatedNel[OutputValidatorError, Unit] =
+    validateTypingResult(typingResult, parentSchema, None)
+
+  final private def validateTypingResult(typingResult: TypingResult, schema: Schema, path: Option[String]): ValidatedNel[OutputValidatorError, Unit] = {
+    (typingResult, schema.getType) match {
+      case (tc@TypedClass(cl, _), _) if AvroUtils.isSpecificRecord(cl) =>
+        validateSpecificRecord(tc, schema, path)
+      case (record: TypedObjectTypingResult, Type.RECORD) =>
+        validateRecordSchema(record, schema, path)
+      case (map: TypedObjectTypingResult, Type.MAP) =>
```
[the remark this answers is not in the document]
lciolecki: Changed as we have spoken.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,290 @@
...
+      condNel(validationMode.acceptRedundant || redundantFields.isEmpty, (), OutputValidatorRedundantFieldsError(prepareFields(redundantFields)))
+    }
+
+   requiredFieldsValidation combine schemaFieldsValidation combine redundantFieldsValidation
+  }
+
+  private def validateTypedClassToMapSchema(map: TypedClass, schema: Schema, path: Option[String]) = {
+    map.params match {
+      case _ :: value :: Nil =>
+        validateTypingResult(value, schema.getValueType, buildPath("*", path, isGeneric = true))
```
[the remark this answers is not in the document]
lciolecki: Yep, you right - I've added

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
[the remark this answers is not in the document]
lciolecki: I've added some changes here - Right now I check array legth.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/spel/Implicits.scala
```diff
@@ -1,9 +1,53 @@
...
+    data match {
+      case map: collection.Map[String@unchecked, _] if map.isEmpty =>
+        EmptyRecord
+      case map: collection.Map[String@unchecked, _] =>
+        val elements = map.map{case (key, value) => s""""$key": ${toSpEL(value)}"""}
+        convertCollection(elements.toList)
+      case map: util.Map[String@unchecked, _] =>
+        toSpEL(map.asScala)
+      case collection: Traversable[_] if collection.isEmpty =>
+        EmptyList
```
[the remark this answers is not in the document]
arkadius: It's not the point of that :) Take a look - you have two cases for `collection: Traversable`. For maps it was needed because empty map is represented as a special literal `{:}` in spel. But for lists - it is just like for one element but without this element ;) Remove this case and you'll see that everything will work as expected.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/spel/Implicits.scala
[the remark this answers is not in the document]
arkadius: But this comment wasn't only for that case - you can handle the same way also those cases:
```
      case spel: String if spel.startsWith("T(") => spel
      case spel: String if spel.startsWith("{") => spel
```
You should have only one case for strings

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
arkadius: Yes it was the point :) Why it doesn't have sense?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,292 @@
...
+import scala.language.implicitConversions
+import scala.reflect.ClassTag
+import scala.util.Try
+
+private[encode] case class AvroSchemaExpected(schema: Schema) extends OutputValidatorExpected {
+  override def expected: String = AvroSchemaOutputValidatorPrinter.print(schema)
+}
+
+object AvroSchemaOutputValidator {
+  private[encode] val SimpleAvroPath = "Data"
```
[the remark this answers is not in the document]
arkadius: Ok

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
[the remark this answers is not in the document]
arkadius: I'm not sure if spoken about this. IMO if you send record with nested field ot record type in place of map, it will blow up in runtime and we should check if objType is a map

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
arkadius: 1. I get it
2. Now is fine
3. I get it
4. Now is better, but IMO we still can work a bit on this name to make this hard part more understable. You don't check nullability because you already know that this schema is nullable. You check if provided type matches at lease one element of union of schemas. So maybe it should be "includeNullTypeInChecking" or just "checkNullSchemaType"?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -0,0 +1,285 @@
...
+      schemas.map(validateTypingResult(typingResult, _, path)).toList
+    }
+
+    def isNullableSchema(schema: Schema): Boolean = schema.getTypes.size() == 2 && schema.isNullable
+
+    def asSingleValidatedResults(results: List[ValidatedNel[OutputValidatorError, Unit]]) =
+      if (results.exists(_.isValid)) valid else results.sequence.map(_=> ())
+
+    val unionValidationResults = schema match {
+      case sch if isNullableSchema(sch) =>
```
arkadius, line 179: It is not exactly nullable schema. To be precise it is union of null and one more type. I wonder if won't be better to just inline this expression here and add a little comment that we handle it that way to not overwhelm messages with X is not null

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
```diff
@@ -1,14 +1,14 @@
...
-case class ValidationMode private(name: String, label: String, acceptRedundant: Boolean)
+case class ValidationMode private(name: String, label: String, strict: Boolean)
 
 object ValidationMode {
 
-  val strict: ValidationMode = ValidationMode("strict", "Strict mode", acceptRedundant = false)
+  val strict: ValidationMode = ValidationMode("strict", "Strict mode", strict = true)
 
-  val allowRedundant: ValidationMode =  ValidationMode("allowRedundant", "Allow redundant parameters", acceptRedundant = true)
+  val loose: ValidationMode = ValidationMode("flexible", "Loose mode", strict = false)
```
arkadius, line 9: Change name to 'loose' as well.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
```diff
@@ -1,14 +1,14 @@
 package pl.touk.nussknacker.engine.avro.encode
 
-case class ValidationMode private(name: String, label: String, acceptRedundant: Boolean)
+case class ValidationMode private(name: String, label: String, strict: Boolean)
```
arkadius, line 3: We we still have this strict field if you can just check if `mode == ValidationMode.strict`?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/LiteralSpEL.scala
```diff
@@ -0,0 +1,54 @@
...
+  val EmptyRecord = "{:}"
+  val EmptyList: String = "{}"
+}
+
+trait LiteralSpEL {
+
+  import collection.JavaConverters._
+
+  protected def toSpELLiteral(data: Any): String = {
+    def convertCollection(data: List[String]) = s"""{${data.mkString(",")}}"""
```
arkadius, line 17: Can u add space after colon?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -81,26 +81,7 @@ class AvroSchemaTypeDefinitionExtractor(skipOptionalFields: Boolean) {
       case Schema.Type.BOOLEAN =>
         Typed[Boolean]
       case Schema.Type.NULL =>
-        Typed.empty
+        TypedNull
```
arkadius, line 84: I don't think that it is a good idea. This class show what type will be returned in source and it is the only purpose of it. (maybe we should change the name of class to avoid confusions?). For simple schema null | other type we will present union type which 1) is verbose and it doesn't help user 2) doesn't work correctly for some edge cases - don't remember which once.

arkadius: My bad, It's good - we filtering out null from unions

scenario-api/src/main/scala/pl/touk/nussknacker/engine/spel/Implicits.scala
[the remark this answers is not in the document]
lciolecki: Ahh, ok! Chagned.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/spel/Implicits.scala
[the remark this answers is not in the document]
lciolecki: Changed!

engine/flink/avro-components-utils/src/test/scala/pl/touk/nussknacker/engine/avro/AvroSchemaSpelExpressionSpec.scala
lciolecki: In same way you can compare int / other objects.. .toString always will return string?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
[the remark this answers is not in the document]
lciolecki: You're right.. I've fixed it and added some test:
```
(rConfig(Map("price" -> sampleDouble.toString), fakeMapRecordWithStringPrice, recordMapOfStringsSchema, Input), invalidTypes("path 'field' actual: '{price: String}' expected: 'Map[String, Null | String]'")),
```

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
lciolecki: Inlined.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
lciolecki: ups.. thnaks!

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/ValidationMode.scala
lciolecki: Hmm, I've removed field strinct and right now operate on const: `strict` as you said.

mproch (approved):
I think we're good to merge...

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/LiteralSpEL.scala
lciolecki: Done.

gadomsky (approved):

PR #3077, opened 2022-05-24
Change scenarios tab to use new UI by default

gadomsky:


dswiecki (approved):
lets just check it e.g on staging

PR #3099, opened 2022-05-30
Added checking for input nodes named key in UnionMemo.

dawidsula26:
Added check for input nodes named "key" in union-memo. Without this check, it is possible to create situation where one field in union-memo's output is used to store both key of elements and value from branch.


mproch (commented):
Can we have a test for that?

dawidsula26:
Error message, when two input nodes have identical names after sanitation, might need improving. I didn't have any better ideas for it.

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionWithMemoTransformer.scala
```diff
@@ -69,7 +71,20 @@ class UnionWithMemoTransformer(timestampAssigner: Option[TimestampWatermarkHandl
...
+    val validatedIdentical = ids.toList.groupBy(ContextTransformation.sanitizeBranchName).map {
+      case (_, values) if values.size >= 2 =>
+        val namesList = values.map("\"" + _ + "\"").mkString(", ")
+        ProcessCompilationError.CustomNodeError(s"Nodes $namesList have too similar names", None).invalidNel
+      case _ =>
+        Validated.validNel(Unit)
+    }.toList.sequence_
+
+    val validatedBranches = Validated.condNel(!ids.map(ContextTransformation.sanitizeBranchName).contains(KeyField), (),
+      ProcessCompilationError.CustomNodeError(s"""Input node can not be named \"$KeyField\"""", None))
```
mproch, line 85: With """ you don't need to escape "

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionWithMemoTransformer.scala
```diff
@@ -69,7 +71,20 @@ class UnionWithMemoTransformer(timestampAssigner: Option[TimestampWatermarkHandl
   def transformContextsDefinition(valueByBranchId: Map[String, LazyParameter[AnyRef]], variableName: String)
                                  (inputContexts: Map[String, ValidationContext])
                                  (implicit nodeId: NodeId): ValidatedNel[ProcessCompilationError, ValidationContext] = {
-    ContextTransformation.findUniqueParentContext(inputContexts).map { parent =>
+    val ids = valueByBranchId.keySet
+
+    val validatedIdentical = ids.toList.groupBy(ContextTransformation.sanitizeBranchName).map {
```
mproch, line 76: let's extract method "validateBranchNames"

engine/flink/components/base/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionWithMemoTransformerSpec.scala
```diff
@@ -92,6 +91,77 @@ class UnionWithMemoTransformerSpec extends FunSuite with FlinkSpec with Matchers
...
+    ))
+
+    val sourceFoo = BlockingQueueSource.create[OneRecord](_.timestamp, Duration.ofHours(1))
+    val sourceBar = BlockingQueueSource.create[OneRecord](_.timestamp, Duration.ofHours(1))
+
+    val collectingListener = ResultsCollectingListenerHolder.registerRun(identity)
+
+    val model = LocalModelData(ConfigFactory.empty(), new UnionWithMemoTransformerSpec.Creator(sourceFoo, sourceBar, collectingListener))
+    val validationResult = model.validator.validate(process).result
+    assert(validationResult.isInvalid)
```
mproch, line 126: maybe some assertion on error? e.g. using
https://www.scalatest.org/user_guide/using_matchers#matchingAPattern

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ContextTransformation.scala
```diff
@@ -56,6 +56,23 @@ object ContextTransformation {
...
+    } else {
+      List()
+    }
+  }
+
+  def checkSanitizedBranchNames(branchIds: Set[String])(implicit nodeId: NodeId): List[ProcessCompilationError] =
+    branchIds.toList.groupBy(sanitizeBranchName).flatMap{
+      case (_, values) if values.size >= 2 =>
+        val namesList = values.map("\"" + _ + "\"").mkString(", ")
+        List(ProcessCompilationError.CustomNodeError(s"Nodes $namesList have too similar names", None))
```
arkadius, line 71: Maybe to be more precise "conflicting names"?

arkadius (approved):

PR #3102, opened 2022-05-30
Fixed selecting sum as parameter

dawidsula26:
Change of one line to fix bug where selecting first or sum aggregator works incorrectly.


mproch (approved):
thx

PR #3114, opened 2022-06-02
add flush method to SharedKafkaProducer

gemramarcin:


mproch (approved):
Maybe let's add entry to changelog?

PR #3126, opened 2022-06-07
Fix/fetch remote scenario version

gskrobisz:
Right now NU raises an exception while comparing scenario with remote env version. Version id in request path seems to be wrong:
```
GET /api/processes/SOME_SCENARIO_ID/VersionId(53)
```
It should be:
```
GET /api/processes/SOME_SCENARIO_ID/53
```


mproch (approved):
Let's move it to 1.4 branch...

PR #3136, opened 2022-06-08
Improvements: Lite Kafka testkit

lciolecki:


engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
```diff
@@ -0,0 +1,50 @@
...
+  private val inputTopic = "input"
+  private val outputTopic = "output"
+
+  test("should test end to end kafka avro sink / source") {
+    //Given
+    val scenario = ScenarioBuilder.streamingLite("check avro serialization")
+      .source("my-source", SourceName, TopicParamName -> s"'$inputTopic'", SchemaVersionParamName -> s"'${SchemaVersionOption.LatestOptionName}'")
+      .emptySink("my-sink", SinkName, TopicParamName -> s"'$outputTopic'",  SchemaVersionParamName -> s"'${SchemaVersionOption.LatestOptionName}'", SinkKeyParamName -> "", SinkValueParamName -> s"#input", SinkValidationModeParameterName -> s"'${ValidationMode.strict.name}'")
+
+    val runtime = LiteKafkaAvroTestScenarioRunner(new LiteKafkaSourceImplFactory, LiteKafkaAvroSinkImplFactory)
```
arkadius, line 39: We someone who want to test his/her kafka avro scenario have to provide this KafkaSourceImpl and KafkaAvroSinkImpl? Why not to make it default?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaAvroTestScenarioRunner.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  private val schemasCache = new DefaultCache[String, Schema](cacheConfig = CacheConfig())
+  private val dependencies = ProcessObjectDependencies(config, DefaultNamespacedObjectNaming)
+  private val schemaRegistryMockClient: MockSchemaRegistryClient = new MockSchemaRegistryClient
+  private val mockedSchemaProvider = ConfluentSchemaRegistryProvider.avroPayload(new MockConfluentSchemaRegistryClientFactory(schemaRegistryMockClient))
+  private val avroEncoder: BestEffortAvroEncoder = BestEffortAvroEncoder(ValidationMode.strict)
+
+  private lazy val delegate = {
+    val sourceComponent = ComponentDefinition(SourceName, new KafkaAvroSourceFactory(mockedSchemaProvider, dependencies, sourceAvroImpl))
+    val sinkComponent = ComponentDefinition(SinkName, new KafkaAvroSinkFactory(mockedSchemaProvider, dependencies, sinkAvroImpl))
```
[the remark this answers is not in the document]
lciolecki: Good point, I will improve it.

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  implicit class ConsumerRecordWihValue(record: ConsumerRecord[String, Any]) {
+    def withValue(value: Any) = new ConsumerRecord(record.topic, record.partition, record.offset, record.key, value)
+  }
+
+  def createConfig(config: Config): Config =
+    config.withValue("kafka.kafkaAddress", ConfigValueFactory.fromAnyRef("kafka:666"))
+}
+
+class KafkaTestScenarioRunner(components: List[ComponentDefinition], config: Config,
```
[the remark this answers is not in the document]
arkadius: This question is related with comment above about making all `LiteKafkaComponentProvider` components accessible for tests - if we merge both classes, it will much easier to make it possible. Also we have a plan to merge all sources/sinks to one kafka source and one kafka sink in next Nussknacker release.

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
lciolecki: Because of dependencies.. toolkit doesn't have access to Implementation. If add at toolkit dependency to light implementation then we can't do functional tests ad light implementation because of looping dependencies?

arkadius: I see now, It looks like we need to have another module kafka-functional-tests depending on kafka-testkit and than kafka-testkit might depend on on kafka-components.
After that change people could write kafka functional tests without troubles. Can you add at least TODO that this API should be simplified?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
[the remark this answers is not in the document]
arkadius: ok, ignore this comment, let's fucus on others, resolving it

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
```diff
@@ -0,0 +1,68 @@
...
+      .source("my-source", KafkaAvroName, TopicParamName -> s"'$inputTopic'", SchemaVersionParamName -> s"'${SchemaVersionOption.LatestOptionName}'")
+      .emptySink("my-sink", KafkaSinkRawAvroName, TopicParamName -> s"'$outputTopic'",  SchemaVersionParamName -> s"'${SchemaVersionOption.LatestOptionName}'", SinkKeyParamName -> "", SinkValueParamName -> s"#input", SinkValidationModeParameterName -> s"'${ValidationMode.strict.name}'")
+
+    val (mockedComponents, mockSchemaRegistryClient) = prepareRunnerConfig
+    val runtime = LiteKafkaTestScenarioRunner(mockSchemaRegistryClient, mockedComponents)
+    runtime.registerSchemaAvro(inputTopic, recordSchema)
+    runtime.registerSchemaAvro(outputTopic, recordSchema)
+
+    //When
+    val record = AvroUtils.createRecord(recordSchema,
```
arkadius, line 50: Can u use LogicalTypesGenericRecordBuilder to reduce "magic" in generation of input data?

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
```diff
@@ -0,0 +1,68 @@
...
+      Map("first" -> "Jan", "last" -> "Kowalski", "age" -> 18, "address" -> Map("city" -> "Warsaw"))
+    )
+
+    val input = KafkaConsumerRecord(inputTopic, record)
+    val result = runtime.runWithAvroData(scenario, List(input))
+
+    //Then
+    result.map(_.value()) shouldBe List(input.value())
```
arkadius, line 58: Do you really want to compare raw bytes? What you'll do when bytes will be different?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,99 @@
...
+import scala.reflect.ClassTag
+
+object LiteKafkaTestScenarioRunner {
+  val DefaultKafkaConfig: Config =
+    ConfigFactory
+      .empty()
+      .withValue("kafka.kafkaAddress", ConfigValueFactory.fromAnyRef("kafka:666"))
+      .withValue("kafka.kafkaProperties.\"schema.registry.url\"", fromAnyRef("schema-registry:666"))
+      // we disable default kafka components to replace them by mocked
+      .withValue("components.kafka.disabled", ConfigValueFactory.fromAnyRef(true))
```
arkadius, line 28: The trick of mocking schemas is now divided into two places: creation of mocked components + disabling not mocked here. IMO config should be created in the place where we create mocked components. Here only faked `kafka.kafkaAddress` has sense

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,99 @@
...
+    delegate
+      .runWithData[SerializedInput, SerializedOutput](scenario, serializedData)
+      .map(output => {
+        val schema = getSchemaData(output.topic()).schema.asInstanceOf[AvroSchema]
+        val (_, value) = ConfluentUtils.deserializeSchemaIdAndRecord(output.value(), schema.rawSchema())
+        new ProducerRecord(output.topic(), output.partition(), output.timestamp(), output.key(), value)
+      })
+  }
+
+  private def serialize(input: AvroInput ): SerializedInput = {
```
arkadius, line 64: I don't like having this code here. Now you have responsibility of creation of avro record in two places - in test and here. How you are sure that you should pick latest schema? Why not to just use schema id that is returned during registration of schema?

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
lciolecki: I don't compare bytes.. Please look how is implemented `runWithAvroData`.

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
lciolecki: In first version I divided config to 2 parts.. Maybe you are right.

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
lciolecki: If you want to use `LogicalTypesGenericRecordBuilder` you still can do it. If someone wants se how is implemented creation record then he can go inside method. For me it's more convenient way to create GenericRecord.

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
arkadius: I saw somewhere in code `ConsumerRecord[String, Array[Byte]]` but I see now that this method takes other `ConsumerRecord`

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
arkadius: Doing such things you can have false positives in tests. Verification code should be as stupid as it is possible - your verification code has magic of conversion between map and generic record.

lciolecki: Test will not pass because record will be not compatibility with schema. But ok, I will change it. But to be honest with you, I I would rather write test an use magic conversion then use building record step by step. The same think I can say about: 
`ConfluentUtils.serializeRecordToBytesArray` what if confluent will change their serialization way?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,99 @@
...
+
+class LiteKafkaTestScenarioRunner(schemaRegistryClient: SchemaRegistryClient, components: List[ComponentDefinition], config: Config) extends TestScenarioRunner {
+
+  override type Input = ConsumerRecord[String, Any]
+  override type Output = ProducerRecord[String, Any]
+
+  type SerializedInput = ConsumerRecord[String, Array[Byte]]
+  type SerializedOutput = ProducerRecord[String, Array[Byte]]
+
+  type AvroInput = ConsumerRecord[String, GenericRecord]
```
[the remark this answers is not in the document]
lciolecki: You can still use `runWithData` where you can pass serialized data. `ConfluentUtils` can't serialize primitive so, the only way to test it is use standard method.

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
lciolecki: Okey, I've changed it. Right now on input we have `KafkaAvroValue(record: GenericRecord, schemaId: Int)` and we do serialization based on registered schemaId. Deserialization is based on schemaId fetched from payload.

engine/lite/components/kafka/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaAvroFunctionalTest.scala
arkadius: > record will be not compatibility with schema
Will be if field will be optional - it is one of scenarios that u will assume in tests. Won't you?
> what if confluent will change their serialization way
I don't get this comparison

To be honest, at the beginning I was thinking that you are using `BestEffortAvroEncoder` for that, but after that I've realized that you wrote almost no-magic map to record translation. It is ok for me - I was scary about  `BestEffortAvroEncoder` magic

lciolecki: Ah.. Okey, so I will rollback it! :)

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+import pl.touk.nussknacker.engine.util.test.TestScenarioRunner
+
+import scala.reflect.ClassTag
+
+object LiteKafkaTestScenarioRunner {
+  val DefaultKafkaConfig: Config =
+    ConfigFactory
+      .empty()
+      .withValue("kafka.kafkaAddress", ConfigValueFactory.fromAnyRef("kafka:666"))
+      .withValue("kafka.kafkaProperties.\"schema.registry.url\"", fromAnyRef("schema-registry:666"))
```
arkadius, line 25: Can you move this line to test? I think that it is also a part of schema mocking logic

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+    val value = ConfluentUtils.serializeRecordToBytesArray(input.value().record, input.value().schemaId)
+    new ConsumerRecord(input.topic, input.partition, input.offset, input.key, value)
+  }
+
+  def registerSchemaAvro(topic: String, schema: Schema): Int = schemaRegistryClient.register(
+    ConfluentUtils.topicSubject(topic, false),
+    ConfluentUtils.convertToAvroSchema(schema)
+  )
+
+  def registerSchemaAvro(topic: String, strSchema: String): Int =
```
arkadius, line 71: It is unused? Also I think that it is better to force keeping schemas in value objects (`Schema`) instead of passing `Stiring`. It reduce possibility of mistake and invalid schema is detected at the very beginning

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+  def apply(schemaRegistryClient: SchemaRegistryClient, components: List[ComponentDefinition]): LiteKafkaTestScenarioRunner =
+    new LiteKafkaTestScenarioRunner(schemaRegistryClient, components, DefaultKafkaConfig)
+}
+
+class LiteKafkaTestScenarioRunner(schemaRegistryClient: SchemaRegistryClient, components: List[ComponentDefinition], config: Config) extends TestScenarioRunner {
+
+  override type Input = ConsumerRecord[String, Any]
+  override type Output = ProducerRecord[String, Any]
+
+  type SerializedInput = ConsumerRecord[String, Array[Byte]]
```
arkadius, line 36: And what if we want to test e.g. consumer record with avro record in key? We have long term plans to rewrite all tests in flink/avro-component-utils too such mechanism. I'm not saying that all combinations should be provided in this cllass but maybe fallback raw version should be `ConsumerRecord[Array[Byte], Array[Byte]]`. WDYT?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+    ConfigFactory
+      .empty()
+      .withValue("kafka.kafkaAddress", ConfigValueFactory.fromAnyRef("kafka:666"))
+      .withValue("kafka.kafkaProperties.\"schema.registry.url\"", fromAnyRef("schema-registry:666"))
+
+  def apply(schemaRegistryClient: SchemaRegistryClient, components: List[ComponentDefinition]): LiteKafkaTestScenarioRunner =
+    new LiteKafkaTestScenarioRunner(schemaRegistryClient, components, DefaultKafkaConfig)
+}
+
+class LiteKafkaTestScenarioRunner(schemaRegistryClient: SchemaRegistryClient, components: List[ComponentDefinition], config: Config) extends TestScenarioRunner {
```
arkadius, line 31: Why you extends `TestScenarioRunner`? Because of that you need to have Input and Output implemented. What it gives to you?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+        new ProducerRecord(output.topic(), output.partition(), output.timestamp(), output.key(), value)
+      })
+  }
+
+  private def serialize(input: AvroInputRecord ): SerializedInput = {
+    val value = ConfluentUtils.serializeRecordToBytesArray(input.value().record, input.value().schemaId)
+    new ConsumerRecord(input.topic, input.partition, input.offset, input.key, value)
+  }
+
+  def registerSchemaAvro(topic: String, schema: Schema): Int = schemaRegistryClient.register(
```
arkadius, line 66: `registerAvroSchema` is more grammatically correct

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
lciolecki: I can try to deal with this.

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
lciolecki: So maybe question is: do we really need `TestScenarioRunner`?

arkadius: IMO as long as it gave something to us it was good to have it. One advantage that I saw was common method of interacting with test mechanism. But when it started to be problematic and disadvantages (unnecessary complexity) exceeded advantages, we should decide to get rid of this.

build.sbt
```diff
@@ -823,7 +823,7 @@ lazy val liteComponentsTestkit = (project in utils("lite-components-testkit")).
   settings(commonSettings).
   settings(
     name := "nussknacker-lite-components-testkit",
-  ).dependsOn(componentsTestkit, requestResponseRuntime, liteEngineRuntime)
+  ).dependsOn(componentsTestkit, requestResponseRuntime, liteEngineRuntime, avroComponentsUtils)
```
dswiecki, line 826: imo avroComponentsUtils should be dependency of componentsTestkit

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
dswiecki: One of main purposes of TestScenarioRunner is that runners are easily interchangeable so that you could run test either on lite or flink without deep test refactoring, thus I'm not 100% sure should we hide schemaRegistryClient inside LiteKafkaTestScenarioRunner

build.sbt
lciolecki: Can you talk about it with @arkadius ?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentUtils.scala
```diff
@@ -60,14 +62,37 @@ object ConfluentUtils extends LazyLogging {
...
-    writeSchemaId(schemaId, bos)
-    val encoder = EncoderFactory.get().binaryEncoder(bos, null)
-    val writer = new GenericDatumWriter[GenericRecord](record.getSchema, AvroUtils.genericData)
-    writer.write(record, encoder)
-    encoder.flush()
-    bos.toByteArray
+  /**
+    * Based on serializeImpl from [[io.confluent.kafka.serializers.AbstractKafkaAvroSerializer]]
+    */
+  def serializeDataToBytesArray(data: Any, schemaId: Int): Array[Byte] = {
```
arkadius, line 68: :+1:

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
```diff
@@ -0,0 +1,93 @@
...
+    new ConsumerRecord(topic, DefaultPartition, DefaultOffset, key, value)
+}
+
+case class KafkaAvroElement[T](data: T, schemaId: Int)
+
+object KafkaAvroConsumerRecord {
+  def apply[V](topic: String, value: V, schemaId: Int): ConsumerRecord[KafkaAvroElement[String], KafkaAvroElement[V]] =
+    KafkaConsumerRecord(topic, KafkaAvroElement(value, schemaId))
+
+  def apply[K, V](topic: String, key: K, keySchemaId: Int, value: V, valueSchemaId: Int): ConsumerRecord[KafkaAvroElement[K], KafkaAvroElement[V]] =
```
arkadius, line 91: :+1:

build.sbt
arkadius: At the end we want to have kafka avro tests for both lite and flink engine, but at the beginning we could start with test only based on lite engine. If we need it for flink, we will extract it to upper level. WDYT @dswiecki ?

utils/lite-components-testkit/src/main/scala/pl/touk/nussknacker/engine/lite/util/test/LiteKafkaTestScenarioRunner.scala
arkadius: @dswiecki after longer conversation with @mproch and @lciolecki we've decided to keep this `TestScenarioRunner` trait as a "marking" trait for all test runners and have a little bit different methods in kafka implemenation. We will try to keep all implementations in common convention but with slight differences. Do you agree with that approach?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/ConfluentUtils.scala
arkadius, line 68: Please add migration guide about this change. I think that It could be used somewhere in client code

arkadius (approved):

PR #3149, opened 2022-06-13
fix: escape characters for spel inside sql

### LINKED ISSUE #2943: SQL editor parses spel expression incorrectly
Screenshot from 2022-03-15 16-16-99

The editor parses brackets incorrectly as seen above and sends expression:
`SELECT
      1 as STH
      FROM SOME_TABLE
      WHERE "1" IN #{ {"1", "2"} }`
as 
`'SELECT
      1 as STH
      FROM SOME_TABLE
      WHERE "1" IN '+{"1", "2"+' }'`

Version: 1.3.0-staging-2022-03-11-7007-4ae648b585467ede279e719388c53e12332ed97d-SNAPSHOT

JulianWielga:
fixes #2943


mproch (commented):
Let's add entry in changelog, did you think about {{ ddd }} like in yaml? :)

JulianWielga:
> Let's add entry in changelog, did you think about {{ ddd }} like in yaml? :)

same problem witch `{{{"a"}}}`. must be sth not from sql and not from spel. for ace syntax it's easy (but you know how much work is done there by ace) to support almost anything, but take a look at our escape/unescape, it's regex only and i think we should keep it simple as possible. `{{ {"a"} }}` (with spaces) is an option but i think it's easier to teach user escape seq instead of when to use space ;) ...and trim simplifies a lot.

dswiecki (approved):

PR #3151, opened 2022-06-13
cypress tests update

JulianWielga:


mproch (approved):
Why on second image there

PR #3165, opened 2022-06-15
Introduce 'enableConfigEndpoint' configuration with default value fal…

gadomsky:
…se, which enables 'GET /api/app/config/' endpoint.


docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -658,6 +658,7 @@ Currently, you can only configure secondary environment if it uses BASIC authent
 | intervalTimeSettings.processes   | Low        | int     | 20000         | How often frontend reloads scenario list                                                      |
 | intervalTimeSettings.healthCheck | Low        | int     | 30000         | How often frontend reloads checks scenarios states                                            |
 | developmentMode                  | Medium     | boolean | false         | For development mode we disable some security features like CORS. **Don't** use in production |
+| enableConfigEndpoint             | Medium     | boolean | false         | Expose config over http (GET /api/app/config/) - requires admin permission                    |
```
mproch, line 661: maybe let's add notice that it can expose passwords etc. ?

gadomsky: done

mproch (approved):

PR #3169, opened 2022-06-15
Exposing simple health check endpoint publicly

arkadius:


arkadius:
@dswiecki please add info if this change is go/nogo.

dswiecki:
@arkadius lets add changelogs entry and migration guide so that user knows that now its not secured and react accordingly

dswiecki (approved):

PR #3191, opened 2022-06-22
fixed bug showing wrong row value after removing a row in MapVariable 

maciej-brzezinski:


maciej-brzezinski:
> @maciej-brzezinski please check is there open issue for this bug and add info about solving in this pr.

I didn't find an open issue regarding this bug

dswiecki (approved):

PR #3198, opened 2022-06-27
Fix: request response metrics

lciolecki:


mproch (approved):
:+1: pls add entry to changelog

PR #3201, opened 2022-06-28
Add types that hold information about variable's value

dawidsula26:
Current state:
 - only handles String, Long and Boolean. Serialization must be changed to allow more types
 - operators like ==, <, ? :, do not work correctly when types have value in them
 - spel tests fail, because of problems with operators, and, because some expected results have not been changed yet
 - everything seem to be working ok when used from frontend, it works with sample data and can be deployed


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -67,12 +67,20 @@ object typing {
...
+  }
 
-    override def display: String = s"${underlying.display} @ $tag"
+  case class TypedTaggedValue(underlying: SingleTypingResult, tag: String) extends TypedObjectWithData {
+    override def data: AdditionalDataValue = tag
+    override def display: String = s"${underlying.display} @ ${data.display}"
+  }
 
+  case class TypedObjectWithValue(underlying: SingleTypingResult, data: AdditionalDataValue) extends TypedObjectWithData {
+    override def display: String = s"${underlying.display}"
```
mproch, line 83: maybe we do want to display value here? Like "${underlying.display} of value ... } or sth? Or is it just to make tests pass ATM?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -204,8 +218,10 @@ object typing {
         case javaList: java.util.List[_] =>
           typedClass(obj.getClass, List(unionOfElementTypes(javaList.asScala.toList)))
         case typeFromInstance: TypedFromInstance => typeFromInstance.typingResult
-        case other =>
-          Typed(other.getClass)
+        case string: String => Typed.typedValue(string)
```
mproch, line 221: I think we need to do it a bit differently. StringValue/LongValue were needed, as we don't know what additional data will be. Here the situation is different - we know the type, so we can serialize/deserialize more wisely.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
arkadius: IMO we could have `s"${underlying.display} @ ${data.display}"` on `TypedObjectWithData` level. It will looks pretty good for both tags and values

dawidsula26: Yes, we should display value here. This is temporary so that I don't have to correct all test cases, that depend on this method, yet. But `TypedObjectWithValue` cannot have display format like `s"${underlying.display} @ ${data.display}" ` because it will not look right for strings like `"", " "`. I was thinking about `s${underlying.display}{${data.display}}`.
I was also thinking about displaying types only when they are neccessary (eg. when expecting `Int{17}`, in union type `String{a} | String{b}`), but it would require changing code in a lot of places.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -77,8 +77,17 @@ trait CanBeSubclassDeterminer {
         case _ => ().validNel
       }
     }
+    val dataValueRestriction = (_: Unit) => {
+      (givenType, superclassCandidate) match {
+        case (TypedObjectWithValue(_, givenValue), TypedObjectWithValue(_, candidateValue)) if givenValue == candidateValue =>
+          ().validNel
+        case (TypedObjectWithValue(_, givenValue), TypedObjectWithValue(_, candidateValue)) =>
+          s"Types with value have different values: $givenValue and $candidateValue".invalidNel
+        case _ => ().validNel
```
arkadius, line 86: Am I right that this means that `Typed[String].canBeSubclassOf(TypedObjectWithValue(Typed[String], "foo"))`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
```diff
@@ -50,6 +50,10 @@ object TypeEncoders {
       val objTypeEncoded = encodeTypingResult(underlying)
       val tagEncoded = "tag" -> fromString(tag)
       objTypeEncoded.+:(tagEncoded)
+    case TypedObjectWithValue(underlying, data) =>
+      val objTypeEncoded = encodeTypingResult(underlying)
+      val dataEncoded = "data" -> SimpleObjectEncoder.encode(underlying, data).toOption.get
```
arkadius, line 55: Please don't use `Option.get`. It can cause scala's NPE (NoSuchElement Option.get). Better is be more explicit like `.getOrElse(illegal => throw new IllegalStateException(s"Not supported data type: $illegal")` or sth like this

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -67,12 +69,29 @@ object typing {
...
+  }
 
+  case class TypedTaggedValue(underlying: SingleTypingResult, tag: String) extends TypedObjectWithData {
+    override def data: Any = tag
     override def display: String = s"${underlying.display} @ $tag"
+  }
 
+  case class TypedObjectWithValue(underlying: TypedClass, data: Any) extends TypedObjectWithData {
+    val maxDataDisplaySize = 15
+    val maxDataDisplaySizeWithDots = 12
```
arkadius, line 86: Can we have only one constant and compute e.g. this by `maxDataDisplaySize - "...".length`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
arkadius: After discussion it is ok - please just leave a comment that we can't compute values in all situations so for method arguments we should be lax in values checking or sth similar

arkadius:
Please remember about adding changelog entry

mproch (changes requested):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -188,6 +207,9 @@ object typing {
 
     def tagged(typ: SingleTypingResult, tag: String): TypedTaggedValue = TypedTaggedValue(typ, tag)
 
+    def typedValue[T: ClassTag](data: T): TypedObjectWithValue =
```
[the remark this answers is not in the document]
arkadius: By:
> we really need this method
you mean that we already have `fromInstance`? I aggree

PR #3208, opened 2022-06-29
Fix set maxAge

dswiecki:

### DIFF
--- a/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactory.scala
+++ b/security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactory.scala
@@ -61,8 +61,10 @@ object ExampleOAuth2ServiceFactory {
 
   val testRules: List[AuthenticationConfiguration.ConfigRule] = AuthenticationConfiguration.getRules(testConfig.usersFile)
 
-  @ConfiguredJsonCodec case class TestAccessTokenResponse(@JsonKey("access_token") accessToken: String, @JsonKey("token_type") tokenType: String) extends OAuth2AuthorizationData {
-    val expirationPeriod: Option[FiniteDuration] = None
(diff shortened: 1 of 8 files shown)


utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/cache/Cache.scala
```diff
@@ -12,6 +12,7 @@ trait Cache[K, V] {
 
 trait AsyncCache[K, V] {
   def getOrCreate(key: K)(value: => Future[V]): Future[V]
+  def get(key: K): Option[V]
```
mproch, line 15: is it for tests only? If so, let's add comment

security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactory.scala
```diff
@@ -61,8 +61,10 @@ object ExampleOAuth2ServiceFactory {
 
   val testRules: List[AuthenticationConfiguration.ConfigRule] = AuthenticationConfiguration.getRules(testConfig.usersFile)
 
-  @ConfiguredJsonCodec case class TestAccessTokenResponse(@JsonKey("access_token") accessToken: String, @JsonKey("token_type") tokenType: String) extends OAuth2AuthorizationData {
-    val expirationPeriod: Option[FiniteDuration] = None
+  @ConfiguredJsonCodec case class TestAccessTokenResponse(@JsonKey("access_token") accessToken: String,
+                                                          @JsonKey("token_type") tokenType: String,
+                                                          @JsonKey("expires_in") expirationPeriod: Option[FiniteDuration],
```
mproch, line 66: is it for tests/debug?

dswiecki: this is basicly solely a test class

PR #3210, opened 2022-06-29
JMX reporter for other metrics

mproch:
Hi,
Please add changelog entry in docs/Changelog.md.
After that update description of this PR.
Thanks.


build.sbt
```diff
@@ -242,6 +242,9 @@ lazy val commonSettings =
         "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonV,
         "com.fasterxml.jackson.core" % "jackson-core" % jacksonV,
         "com.fasterxml.jackson.core" % "jackson-databind" % jacksonV,
+
+        "io.dropwizard.metrics5" % "metrics-core" % dropWizardV,
```
dswiecki, line 246: why do we use 5.0.0-rc3 exactly ? there's 5.0.0-rc11 already available ? https://search.maven.org/artifact/io.dropwizard.metrics5/metrics-core/5.0.0-rc11/bundle

build.sbt
```diff
@@ -1414,6 +1418,10 @@ lazy val ui = (project in file("ui/server"))
 
         "com.dimafeng" %% "testcontainers-scala-scalatest" % testcontainersScalaV % "test",
         "com.dimafeng" %% "testcontainers-scala-postgresql" % testcontainersScalaV % "test",
+
+        "io.dropwizard.metrics5" % "metrics-core" % dropWizardV,
```
dswiecki, line 1422: don't we already have this in common ?

mproch: nope. It's only in lite-runtime

build.sbt
mproch: good point, I'll upgrade...

gadomsky (approved):
LGTM

PR #3213, opened 2022-06-30
fix: redirect to archived scenarios after archive

JulianWielga:


ui/client/containers/paths.ts
```diff
@@ -11,4 +11,5 @@ export const SignalsPath = `/signals`
 export const ServicesPath = `/services`
 export const CustomTabBasePath = `/customtabs`
 export const ScenariosBasePath = `/scenarios`
+export const ArchivedPath = `${ScenariosBasePath}/?SHOW_ARCHIVED=true&HIDE_ACTIVE=true`
```
dswiecki, line 14: wouldn't it work in both cases, I mean in old and new one ?

arkadius: Let's add comment that it don't work with old scenario list

arkadius (approved):

PR #3217, opened 2022-07-04
Fix NullPointerException in QueryResult Encoder

gemramarcin:


mproch (commented):
Can we add a test for that?

PR #3224, opened 2022-07-05
Switch component renamed to Choice

arkadius:


ui/client/assets/json/nodeAttributes.json
```diff
@@ -35,7 +35,7 @@
     }
   },
   "Switch": {
```
lciolecki, line 37: What about this?

arkadius: Switch component type (in scenario json) stay with us. We change only displayed labels on FE.

lciolecki: Ehh this will be misleading..

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -10,6 +10,32 @@ Most of the nodes, with source and sink nodes being notable exceptions, have inp
 
 Sinks and filters can be disabled by selecting `Disable` checkbox. 
 
+## Choice
```
dswiecki, line 13: I'll put it after variable and mapVariable, just as it was previously

dswiecki (approved):
LGTM

PR #3240, opened 2022-07-07
Create error topic for lite if doesn't exist

mproch:


utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaErrorTopicInitializer.scala
```diff
@@ -0,0 +1,35 @@
...
+      (topicExists, exceptionHandlerConfig.createTopicIfNotExists) match {
+        case (true, _) =>
+          logger.debug("Topic exists, skipping")
+        case (false, true) =>
+          logger.info(s"Creating error topic: $errorTopic with default configs, please check if the values are correct")
+          val errorTopicConfig = new NewTopic(errorTopic, Optional.empty[Integer](), Optional.empty[lang.Short]())
+          admin.createTopics(util.Arrays.asList(errorTopicConfig)).all().get(timeoutSeconds, TimeUnit.SECONDS)
+        case (false, false) =>
+          logger.error(s"Topic $errorTopic does not exist, createTopicIfNotExists == false")
+          throw new IllegalArgumentException(s"Errors topic: $errorTopic does not exist, 'createTopicIfNotExists' is 'false', please create the topic")
```
gadomsky, line 30: doesn't `IllegalStateException` suits more?

docs/installation_configuration_guide/ModelConfiguration.md
```diff
@@ -189,14 +189,15 @@ Errors can be sent to specified Kafka topic in following json format (see below
...
-| additionalParams      | {}            | Map of fixed parameters that can be added to Kafka message                                                                                                                                 |
+| Name                   | Default value | Description                                                                                                                                                                                |
+|------------------------|---------------| ------------                                                                                                                                                                               |
+| topic                  | -             | Topic where errors will be sent.                                                     |
+| createTopicIfNotExists | true          | Should error topic be created if it does not exist. Please note that it will be created using default configs - for production deployments make sure the config is ok. Set to false if e.g. Kafka user for Nussknacker does not have permissions to create (in that case the topic has to be configured manually) |
```
gadomsky, line 195: maybe let's add notice, that in such case Nussknacker will start normally and accept scenario deployments, but those are gonna fail.

arkadius, line 195: IMO we don't need this configuration. If someone want to creates its on his own, he can do it or recreate it. 
We could have only a notice next to topic, that if it doesn't exist, it will be created using default topics configuration defined on server. But if you want it can stay as it is.

docs/installation_configuration_guide/model/Lite.md
```diff
@@ -37,7 +37,9 @@ Errors are sent to Kafka, to a dedicated topic:
 modelConfig {
   exceptionHandlingConfig: {
     topic: "errors"
+    createTopicIfNotExists: "true"
```
arkadius, line 40: Why it is here? It is default value

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaErrorTopicInitializer.scala
mproch: :+1:

docs/installation_configuration_guide/ModelConfiguration.md
mproch: You're right, doesn't add too much value. I removed, changed error descriptions a bit and added notice.

docs/installation_configuration_guide/model/Lite.md
mproch: removed...

arkadius (approved):
Good job

gadomsky (approved):

PR #3249, opened 2022-07-11
confluent 5.5->7.2, avro 1.9->1.11 bumps

arkadius:


lciolecki:
Avro 1.11 provides new LogicalType: https://avro.apache.org/docs/current/spec.html#Duration. I think we should handle it?

PR #3268, opened 2022-07-14
preserve typing info after failed validation

JulianWielga:

### DIFF
--- a/ui/client/reducers/graph/reducer.ts
+++ b/ui/client/reducers/graph/reducer.ts
@@ -243,6 +243,7 @@ const graphReducer: Reducer<GraphState> = (state = emptyGraphState, action) => {
           ...state.processToDisplay,
           validationResult: {
             ...action.validationResult,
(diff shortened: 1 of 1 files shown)


ui/client/reducers/graph/reducer.ts
```diff
@@ -241,7 +241,13 @@ const graphReducer: Reducer<GraphState> = (state = emptyGraphState, action) => {
         ...state,
         processToDisplay: {
           ...state.processToDisplay,
-          validationResult: action.validationResult,
+          validationResult: {
+            ...action.validationResult,
+            nodeResults: {
+              ...state.processToDisplay.validationResult.nodeResults,
+              ...action.validationResult.nodeResults,
```
mproch, line 248: This happens when SaveNotAllowed errors happen, can we add comment for that?

mproch (approved):

mproch (approved):

PR #3270, opened 2022-07-14
Added Null literal

dawidsula26:


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -24,6 +24,8 @@ trait CanBeSubclassDeterminer {
     (givenType, superclassCandidate) match {
       case (_, Unknown) => ().validNel
       case (Unknown, _) => ().validNel
+      case (TypedNull, _) => ().validNel
+      case (_, TypedNull) => "Type cannot be subclass of Null".invalidNel
```
arkadius, line 28: maybe `TypedNull.display` instead of hardcoded `Null`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
```diff
@@ -153,7 +157,7 @@ object TypingType extends Enumeration {
 
   type TypingType = Value
 
-  val TypedUnion, TypedDict, TypedObjectTypingResult, TypedTaggedValue, TypedClass, TypedObjectWithValue, Unknown = Value
+  val TypedUnion, TypedDict, TypedObjectTypingResult, TypedTaggedValue, TypedClass, TypedObjectWithValue, NullLiteral, Unknown = Value
```
arkadius, line 160: It should be TypedNull here

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -96,6 +96,10 @@ object typing {
     }
   }
 
+  case object TypedNull extends TypingResult {
```
arkadius, line 99: Maybe it could be helpful to have another trait `WithValue { def value: Any }` implemented by both `TypedNull` and `TypedObjectWithValue` and value will be just `null` here? WDYT?

arkadius (commented):
Please add some tests

arkadius (commented):
Please add changelog entry about new behavior

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -24,6 +24,8 @@ trait CanBeSubclassDeterminer {
     (givenType, superclassCandidate) match {
       case (_, Unknown) => ().validNel
       case (Unknown, _) => ().validNel
+      case (TypedNull, _) => ().validNel
```
mproch, line 27: this is where we differ with Avro ;)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
```diff
@@ -22,10 +22,13 @@ object TypeEncoders {
   private val encodeUnknown = JsonObject("refClazzName" -> fromString(classOf[Object].getName),
     "params" -> fromValues(Nil))
 
+  private val encodeNull = encodeUnknown
```
mproch, line 25: I wonder if it's ok. Will show up in FE as Unknown? Maybe we'd better write sth specific?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/supertype/CommonSupertypeFinder.scala
```diff
@@ -20,6 +20,9 @@ class CommonSupertypeFinder(classResolutionStrategy: SupertypeClassResolutionStr
     (left, right) match {
       case (Unknown, _) => Unknown // can't be sure intention of user - union is more secure than intersection
       case (_, Unknown) => Unknown
+      case (TypedNull, TypedObjectWithValue(r, _)) => r
```
mproch, line 23: :+1: Let's add comment also here (that null cannot be subclass of 5)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -24,6 +24,8 @@ trait CanBeSubclassDeterminer {
...
+      case (_, TypedNull) => s"No type can be subclass of ${TypedNull.display}".invalidNel
```
mproch, line 28: What about  union with Null?

mproch (commented):
:+1: Cool, I added some minor comments

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
dawidsula26: Union must be made out of types that we know something about and we do not know anything about type of Null. If I allowed Null in union that there would be a problem with things like method suggestions on FE.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
dawidsula26: Deserialization and showing this on FE will work fine thanks to `typeField` and `display` fields added in `encodeTypingResult`. I am not sure what these fields are used for, but Object seems like good choice since Null can represent any type. I will add a comment.

dawidsula26:
Current version allows Null to be subclass of Map(a -> Integer{5}, b -> Integer{3}), but it would require more changes to have it handle this case, so I think it can stay like that for now.

mproch (approved):

PR #3276, opened 2022-07-14
SchemaBasedMessagesSerdeProvider abstracts from ParsedSchema

gadomsky:

### DIFF
--- a/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaWithMetadata.scala
+++ b/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaWithMetadata.scala
@@ -1,8 +1,6 @@
 package pl.touk.nussknacker.engine.avro.schemaregistry
 
 import io.confluent.kafka.schemaregistry.ParsedSchema
-import io.confluent.kafka.schemaregistry.avro.AvroSchema
-import io.confluent.kafka.schemaregistry.client.SchemaMetadata
 
 /**
  * This class holds information that are stored next to schema in registry.
@@ -10,12 +8,3 @@ import io.confluent.kafka.schemaregistry.client.SchemaMetadata
  * is not coupled with concrete schema registry implementation.
  */
 case class SchemaWithMetadata(schema: ParsedSchema, id: Int)
-
-object SchemaWithMetadata{
-  def apply(schemaMetadata: SchemaMetadata): SchemaWithMetadata = {
-    schemaMetadata.getSchemaType match {
-      case "AVRO" => SchemaWithMetadata(new AvroSchema(schemaMetadata.getSchema), schemaMetadata.getId)
-      case other => throw new IllegalArgumentException(s"Not supported schema type: $other")
-    }
-  }
-}
--- a/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/source/UniversalKafkaSourceFactory.scala
+++ b/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/source/UniversalKafkaSourceFactory.scala
@@ -2,6 +2,7 @@ package pl.touk.nussknacker.engine.avro.source
 
 import cats.data.Validated
 import cats.data.Validated.Valid
+import io.confluent.kafka.schemaregistry.ParsedSchema
 import io.confluent.kafka.schemaregistry.avro.AvroSchema
 import org.apache.kafka.clients.consumer.ConsumerRecord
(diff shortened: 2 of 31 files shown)


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaWithMetadata.scala
```diff
@@ -14,7 +15,8 @@ case class SchemaWithMetadata(schema: ParsedSchema, id: Int)
 object SchemaWithMetadata{
   def apply(schemaMetadata: SchemaMetadata): SchemaWithMetadata = {
     schemaMetadata.getSchemaType match {
-      case "AVRO" => SchemaWithMetadata(new AvroSchema(schemaMetadata.getSchema), schemaMetadata.getId)
+      case AvroSchema.TYPE => SchemaWithMetadata(new AvroSchema(schemaMetadata.getSchema), schemaMetadata.getId)
```
mproch, line 18: I wonder if we can do it somehow "automatically"... e.g. using SchemaRegistryClient.parseSchema?

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/source/UniversalKafkaSourceFactory.scala
```diff
@@ -27,7 +28,7 @@ import scala.reflect.ClassTag
   * This is universal kafka source - it will handle both avro and json
   * TODO: Move it to some other module when json schema handling will be available
   */
-class UniversalKafkaSourceFactory[K: ClassTag, V: ClassTag](val schemaRegistryProvider: SchemaRegistryProvider,
+class UniversalKafkaSourceFactory[K: ClassTag, V: ClassTag](val schemaRegistryProvider: SchemaRegistryProvider[AvroSchema],
```
mproch, line 31: this is temporary, right?

mproch (commented):
Looks good, let's start to write migration guide for this set of changes ;)

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaRegistryProvider.scala
```diff
@@ -5,13 +5,14 @@ import io.confluent.kafka.schemaregistry.ParsedSchema
 import pl.touk.nussknacker.engine.avro.serialization.{KafkaAvroDeserializationSchemaFactory, KafkaAvroSerializationSchemaFactory}
 import pl.touk.nussknacker.engine.kafka.RecordFormatterFactory
 
-trait SchemaRegistryProvider extends Serializable with BaseSchemaRegistryProvider {
+//todo: rename and separate from BaseSchemaRegistryProvider
+trait SchemaRegistryProvider[T<:ParsedSchema] extends Serializable with BaseSchemaRegistryProvider {
```
arkadius, line 9: IMO The genesis of `SchemaRegistryProvider` name was "provider of implementation of mechanisms serializing/deserializing avro records".  You can imagine situations like:
- you have own, file based/configuration based schema registries (two our clients wanted such feature)
- you have other method of avro serailization - e.g. in confluent it is magic byte + 4bytes for schema id + payload, hermes uses schemaid in header, payload in body, I don't know how other schema registries works (e.g. hortonworks schema registry)

We can change the name to more suitable. Maybe: `SchemaBasedMessagesSerdeProvider`?
Good point that it should be separated from `BaseSchemaRegistryProvider` after adding this generic parameter.

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/SchemaWithMetadata.scala
arkadius: Yeah, we had a conversation about it with @gadomsky - Confluent do exactly the same as we do - they even have their own `nonRestrictiveParseSchema` `(validateDefaults = false)`

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/source/UniversalKafkaSourceFactory.scala
arkadius: IMO we shouldn't have `SchemaRegistryProvider[ParsedSchema]`, we should rather pass here sth like `fun[T<: ParsedSchema](schema: T): SchemaRegistryProvider[T]`. Maybe this PR can show a little bit further direction? Until now it isn't too much code.

PR #3310, opened 2022-07-21
Fix for: Typed for Arrays of objects was generated with empty generic…

arkadius:
… parameters list


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -161,19 +161,20 @@ object typing {
...
-        TypedClass(klass, parameters)
-      } else {
-        parameters match {
-          //it may happen that parameter will be decoded via other means, we have to to sanity check if they match
-          case Nil | `decodedComponentType` :: Nil =>
-            Typed.typedClass(classOf[Array[Object]], List(decodedComponentType))
-          case _: List[TypingResult] =>
-            throw new IllegalArgumentException(s"Array parameter passed twice, klass component type: ${klass.getComponentType}, type passed from parameters: ${parameters.head.display}")
-        }
+      parameters match {
```
dawidsula26, line 169: Maybe you should change type of parameters to `Option[TypingResult]`. This method fail every time it gets a list that has more than one element.

arkadius: I'll refactor it a bit in next PR

dawidsula26 (approved):

PR #3328, opened 2022-07-27
Schema type aware serialization of `NkSerializableParsedSchema`

gadomsky:


utils/avro-components-utils/src/main/java/org/apache/flink/formats/avro/typeutils/NkSerializableParsedSchema.java
```diff
@@ -36,42 +37,58 @@
...
-			this.schema = null;
-		}
-	}
+    @SuppressWarnings("unchecked")
+    private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
+        if (ois.readBoolean()) {
+            byte schemaType = ois.readByte();
+            switch (schemaType) {
+                case avroSchemaType:
+                    String avroSchemaStr = ois.readUTF();
```
lciolecki, line 80: This `ois.readUTF();` is commont stuff and you can move it before the switch.

gadomsky: not really. I do not want to read this if schemaType is unsupported

dswiecki (approved):

PR #3331, opened 2022-07-27
Improvements: more strict avro base type validation

lciolecki:


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -242,32 +248,28 @@ class AvroSchemaOutputValidator(validationMode: ValidationMode) extends LazyLogg
...
-      condNel(typingResult.canBeSubclassOf(schemaAsTypedResult), (),
-        OutputValidatorTypeError(path.getOrElse(SimpleAvroPath), typingResult, AvroSchemaExpected(schema))
-      )
+    val schemaAsTypedResult = AvroSchemaTypeDefinitionExtractor.typeDefinition(schema)
+    val additionalTypes = additionalTypesMapping.getOrElse(schema.getType, Set.empty)
+
+    (schemaAsTypedResult, typingResult) match {
+      case (schemaType: SingleTypingResult, typing: SingleTypingResult) if schemaType.objType.klass == typing.objType.klass => valid
+      case (_, typing: SingleTypingResult) if additionalTypes.contains(typing.objType.klass) => valid
+      case (_, Unknown) => valid //Unknown is proper for each of type
```
lciolecki, line 270: I'm not sure about `Unknown`..  Right now I think, we should remove this line. WDYT @arkadius ?

arkadius: For method invocations (`#UTILS.method()`) you can pass Unknown as an argument and I assume that it can fail in the same way as in avro thanks to this line. So it looks like we are consistent in this. I would leave that with comment that we probably shouldn't accept Unknown here and in other places where some value is passed (like method invocations). But before that we must be sure that user can live without Unknowns. After introducing generic functions we will be one step closer to that...

arkadius (approved):
Good job!

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
arkadius: If you decide to leave this line, I think it should be moved from here to first line of `validateTypingResult` - it should work the same for all other cases like records, maps etc.

PR #3334, opened 2022-07-28
Improvements: Allow to import file with different id

lciolecki:


ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -252,6 +258,18 @@ class DBProcessService(managerActor: ActorRef,
       }).toSet)
   }
 
+  def importProcess(processId: ProcessIdWithName, jsonString: String)(implicit user: LoggedUser): Future[XError[ValidatedDisplayableProcess]] = {
+    withNotArchivedProcess(processId, "Import is not allowed for archived process.") { process =>
+      val result = for {
+        jsonCanonical <- ProcessMarshaller.fromJson(jsonString).leftMap(UnmarshallError).toEither
+        canonical = jsonCanonical.withProcessId(processId.name)
+        displayable = ProcessConverter.toDisplayable(canonical, process.processingType)
+      } yield new ValidatedDisplayableProcess(displayable, processValidation.validate(displayable))
```
mproch, line 267: what happens if we use processResolving.validateBeforeUiResolving(...) - like in process update?

lciolecki: Good catch, we can't remove processValidation and use processResolving instead of it.

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -252,6 +257,19 @@ class DBProcessService(managerActor: ActorRef,
...
+  def importProcess(processId: ProcessIdWithName, jsonString: String)(implicit user: LoggedUser): Future[XError[ValidatedDisplayableProcess]] = {
+    withNotArchivedProcess(processId, "Import is not allowed for archived process.") { process =>
+      val result = for {
+        jsonCanonical <- ProcessMarshaller.fromJson(jsonString).leftMap(UnmarshallError).toEither
+        canonical = jsonCanonical.withProcessId(processId.name)
+        displayable = ProcessConverter.toDisplayable(canonical, process.processingType)
+        validationResult = processResolving.validateBeforeUiResolving(displayable)
+      } yield new ValidatedDisplayableProcess(displayable, validationResult)
+
+      Future.apply(result)
```
mproch, line 269: It's better to use Future.successful for sync stuff

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -252,6 +257,19 @@ class DBProcessService(managerActor: ActorRef,
       }).toSet)
   }
 
+  def importProcess(processId: ProcessIdWithName, jsonString: String)(implicit user: LoggedUser): Future[XError[ValidatedDisplayableProcess]] = {
+    withNotArchivedProcess(processId, "Import is not allowed for archived process.") { process =>
+      val result = for {
+        jsonCanonical <- ProcessMarshaller.fromJson(jsonString).leftMap(UnmarshallError).toEither
+        canonical = jsonCanonical.withProcessId(processId.name)
+        displayable = ProcessConverter.toDisplayable(canonical, process.processingType)
+        validationResult = processResolving.validateBeforeUiResolving(displayable)
```
mproch, line 266: I think we should use validateBeforeUiReverseResolving on canonical process

ui/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -252,6 +257,19 @@ class DBProcessService(managerActor: ActorRef,
       }).toSet)
   }
 
+  def importProcess(processId: ProcessIdWithName, jsonString: String)(implicit user: LoggedUser): Future[XError[ValidatedDisplayableProcess]] = {
+    withNotArchivedProcess(processId, "Import is not allowed for archived process.") { process =>
+      val result = for {
+        jsonCanonical <- ProcessMarshaller.fromJson(jsonString).leftMap(UnmarshallError).toEither
```
mproch, line 263: maybe it'd be clearer to just use map instead of for here?

mproch (approved):
Cool :+1:, this is what our users were waiting for for years :1st_place_medal:

PR #3337, opened 2022-07-28
initial version of BestEffortJsonSchemaEncoder

gadomsky:


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/UniversalSchemaSupport.scala
```diff
@@ -48,32 +49,28 @@ object UniversalSchemaSupport {
       case Some(schema: AvroSchema) => ConfluentKafkaAvroSerializer(kafkaConfig, schemaRegistryClient, Some(schema), isKey = isKey)
       case Some(schema: AvroSchemaWithJsonPayload) => new JsonPayloadKafkaSerializer(kafkaConfig, schemaRegistryClient, new DefaultAvroSchemaEvolution, Some(schema.avroSchema), isKey = isKey)
       case Some(schema: JsonSchema) =>
-        val circeSerializer = new CirceJsonSerializer(schema.rawSchema())
+        lazy val circeSerializer = new CirceJsonSerializer(schema.rawSchema())
```
mproch, line 52: Why lazy? CirceJsonSerializer - do we need it's behaviour now? It's inefficient - serializes and parses once more just to validate schema. If we'll do it in BestEffortJsonSchemaEncoder we won't need it's logic anymore

gadomsky: `CirceJsonSerializer` had `todo` comment on top of it, but as it is right now - this class is unnecessary. Deleted

dswiecki (approved):

PR #3344, opened 2022-07-29
Test type method reference

dawidsula26:
Added tests for type method reference, because before it was possible to make any use of overloaded functions cause errors and still pass all tests. Also made handling typing errors caused by overloaded functions more deterministic.

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
@@ -93,13 +93,19 @@ class TypeMethodReference(methodName: String,
     }
   }
 
+  private def combineArgumentTypeErrors(left: ArgumentTypeError, right: ArgumentTypeError): ArgumentTypeError = {
+    if (!left.found.equals(right.found))
+      throw new IllegalArgumentException("Cannot combine ArgumentTypeErrors where found signatures differ.")
+    new ArgumentTypeError(left.found, left.possibleSignatures ::: right.possibleSignatures)
+  }
+
   // We try to combine ArgumentTypeErrors into one error. If we fail
   // then we only one first error. All regular functions return
   // only ArgumentTypeError, so we will lose information only when
   // there is more than one generic function.
   private def combineErrors(errors: NonEmptyList[(MethodInfo, ExpressionParseError)]): ExpressionParseError = errors match {
     case xs if xs.forall(_._2.isInstanceOf[ArgumentTypeError]) =>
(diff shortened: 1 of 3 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -91,13 +94,15 @@ class TypeMethodReference(methodName: String,
   }
 
   // We try to combine ArgumentTypeErrors into one error. If we fail
-  // then we return only first error. All regular functions return
+  // then we only one first error. All regular functions return
   // only ArgumentTypeError, so we will lose information only when
   // there are two
```
mproch, line 99: only when there are two what?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
```diff
@@ -91,13 +94,15 @@ class TypeMethodReference(methodName: String,
...
-    case xs if xs.forall(_.isInstanceOf[ArgumentTypeError]) =>
-      xs.map(_.asInstanceOf[ArgumentTypeError]).toList.reduce(_.combine(_))
-    case NonEmptyList(head, _) =>
-      head
+  private def combineErrors(errors: NonEmptyList[(MethodInfo, ExpressionParseError)]): ExpressionParseError = errors match {
+    case xs if xs.forall(_._2.isInstanceOf[ArgumentTypeError]) =>
+      xs.map(_._2.asInstanceOf[ArgumentTypeError]).toList.reduce(_.combine(_))
+    case list =>
+      // We return error caused by method with the most parameters, because
+      // this is the method the would be displayed in suggestions on FE.
```
mproch, line 105: I wonder, isn't this logic a bit quirky (it's a bit of an edge case anyway) - if there is overloaded method:
toInvoke(arg: String)
toInvoke(arg1: Long, arg2: Int) and I wanted to invoke the string one, I'll get only error from the second one? We can leave it, but maybe let's move ArgMethodError.combine here -> since it's quite specific for this use case...

mproch (approved):
Overall ok, added minor comments

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/typer/TypeMethodReference.scala
dawidsula26: Yes, as long as at least one of them does not return ArgumentTypeError. I tried to display information about all errors, but with current way of displaying them, it quickly becomes illegible.

PR #3349, opened 2022-08-01
refactor UniversalSchemaSupport

gadomsky:
* `UnsupportedSchemaType` is thrown once in one place
* When introducing new schema type, implementation of all "support methods" is now forced by interface

### DIFF
--- a/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/UniversalSchemaSupport.scala
+++ b/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/UniversalSchemaSupport.scala
@@ -18,35 +18,23 @@ import pl.touk.nussknacker.engine.util.output.OutputValidatorError
 import pl.touk.nussknacker.engine.util.sinkvalue.SinkValueData.SinkValueParameter
 
 object UniversalSchemaSupport {
-
-  implicit class RichParsedSchema(parsedSchema: ParsedSchema) extends UniversalSchemaSupport {
-    private lazy val support: ParsedSchemaSupport[ParsedSchema] = parsedSchema match {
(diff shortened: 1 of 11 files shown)


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/UniversalSchemaSupport.scala
```diff
@@ -6,107 +6,47 @@ import io.confluent.kafka.schemaregistry.ParsedSchema
...
-  def rawOutputValidator(schema: ParsedSchema)(implicit nodeId: NodeId): (TypingResult, ValidationMode) => ValidatedNel[OutputValidatorError, Unit] = (t: TypingResult, validationMode: ValidationMode) =>
-    schema match {
-      case schema: AvroSchema => new AvroSchemaOutputValidator(validationMode).validateTypingResultToSchema(t, schema.rawSchema())
-      case AvroSchemaWithJsonPayload(avroSchema) => new AvroSchemaOutputValidator(validationMode).validateTypingResultToSchema(t, avroSchema.rawSchema())
-      case schema: JsonSchema => new JsonSchemaOutputValidator(validationMode).validateTypingResultToSchema(t, schema.rawSchema())
-      case _ => throw new UnsupportedSchemaType(schema)
-    }
+trait UniversalSchemaSupport {
+  val payloadDeserializer: UniversalSchemaPayloadDeserializer
+  def serializerFactory[T]: (ConfluentSchemaRegistryClient, KafkaConfig, Boolean) => Serializer[T]
```
arkadius, line 43: 1. what does boolean mean here? maybe we should have some case class for those params?
2. why this method return function instead of just `def createSerializer(client, config, ...)`?

arkadius (approved):
LGTM

dswiecki (approved):

PR #3354, opened 2022-08-02
UniversalKakfaSource optimization - if message without schemaId, using cache when getting one

gadomsky:
* SchemaRegistryClient - introduced `getLatestSchemaId` operation
* ConfluentUniversalKafkaDeserializer fix - latest schemaId taken from cache

### DIFF
--- a/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClient.scala
+++ b/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClient.scala
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.engine.avro.schemaregistry.confluent.client
 
 import cats.data.Validated
+import cats.data.Validated.valid
 import com.typesafe.scalalogging.LazyLogging
 import io.confluent.kafka.schemaregistry.SchemaProvider
 import io.confluent.kafka.schemaregistry.avro.AvroSchemaProvider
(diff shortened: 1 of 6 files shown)


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/schemaregistry/confluent/client/CachedConfluentSchemaRegistryClient.scala
```diff
@@ -65,6 +66,14 @@ class CachedConfluentSchemaRegistryClient(val client: CSchemaRegistryClient, cac
     val rawSchema = client.getSchemaById(id)
     SchemaWithMetadata(new SchemaMetadata(id, unknownVersion, rawSchema.schemaType(), rawSchema.references(), rawSchema.canonicalString()), config)
   }
+
+  override def getLatestSchemaId(topic: String, isKey: Boolean): Validated[SchemaRegistryError, Int] = {
```
mproch, line 70: isn't it enough to invoke getAllVersions(...).max?

gadomsky: I think it would return versionId, not schemaId, but i like the idea - I'll see what can be done

gadomsky: Curently there is no cache'able method which that transforms pair (`subject`, `schemaVersion`) to `SchemaMetadata` or `SchemaID` 
It would require futher develoment.

mproch: :+1: let's leave it... I forgot it's version and not schema id :)

mproch (approved):

PR #3355, opened 2022-08-02
Modified Signature

dawidsula26:


components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
```diff
@@ -16,23 +16,18 @@ class GenericFunctionError(messageInner: String) extends ExpressionParseError {
...
-class NoVarArgSignature(val name: String, val types: List[TypingResult]) extends Signature {
-  def display = s"$name(${typesToString(types)})"
+  def display: String = varArg match {
+    case Some(x) => s"$name(${typesToString(noVarArgs :+ x)}...)"
+    case None => s"$name(${typesToString(noVarArgs)})"
+  }
 
   override def equals(obj: Any): Boolean = obj match {
-    case x: NoVarArgSignature => name == x.name && types == x.types
+    case x: Signature => name == x.name && noVarArgs == x.noVarArgs && varArg == x.varArg
```
arkadius, line 29: Why it is not just a case class?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
```diff
@@ -16,23 +16,18 @@ class GenericFunctionError(messageInner: String) extends ExpressionParseError {
 }
 
 
-sealed abstract class Signature {
-  def display: String
-
-  protected def typesToString(types: List[TypingResult]): String =
+final class Signature(val name: String, val noVarArgs: List[TypingResult], val varArg: Option[TypingResult]) {
```
arkadius, line 19: Can't we do the same with `VarArgsMethodInfo` and `NoVarArgsMethodInfo`?

mproch: :+1:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
mproch: :+1:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
dawidsula26: I have already done it, but its with adding support for varArgs in generic functions

arkadius (approved):

PR #3358, opened 2022-08-03
Merged varArg and noVarArg versions of StaticMethodInfo

dawidsula26:


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -21,112 +21,103 @@ object TypeInfos {
...
   sealed trait MethodInfo {
     def computeResultType(arguments: List[TypingResult]): ValidatedNel[ExpressionParseError, TypingResult]
 
-    def staticParameters: List[Parameter]
+    def staticParameters: List[Parameter] =
+      staticNoVarArgParameters ::: staticVarArgParameter.map{ case Parameter(name, refClazz) =>
+        Parameter(name, Typed.genericTypeClass(classOf[Array[Object]], List(refClazz)))
+      }.toList
+
+    def staticParametersWithSimpleVarArg: List[Parameter] = staticNoVarArgParameters ::: staticVarArgParameter.toList
```
mproch, line 32: So ATM we presenting var args in UI as single parameter? Can we add comment on that?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -21,112 +21,103 @@ object TypeInfos {
...
-      case (false, _) =>
-        NoVarArgsMethodInfo(parameters, refClazz, name, description)
-    }
-  }
-
   sealed trait MethodInfo {
     def computeResultType(arguments: List[TypingResult]): ValidatedNel[ExpressionParseError, TypingResult]
 
-    def staticParameters: List[Parameter]
+    def staticParameters: List[Parameter] =
```
mproch, line 27: when do we use this method? For runtime checks?

mproch (approved):
:+1: Added minor comments

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
arkadius: yep, please add comment - it looks like some kind of hack that we probably need for now ;-), the same with `staticParametersWithSimpleVarArg`

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -21,112 +21,103 @@ object TypeInfos {
...
+    def staticNoVarArgParameters: List[Parameter]
+
+    def staticVarArgParameter: Option[Parameter]
 
     def staticResult: TypingResult
 
     def description: Option[String]
 
-    def varArgs: Boolean
+    def varArgs: Boolean = staticVarArgParameter.isDefined
```
arkadius, line 42: can u do "non interface" methods final? it will be easier to analyze interface when I'll see less methods that can be overridden

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
arkadius: If it's true, what @mproch mentioned, maybe we can replace "simple" with some more meaningful word. staticParametersWithVarArgAsASingleParameter or similar.

arkadius (approved):

PR #3359, opened 2022-08-03
migrate flink kafka integration tests to universal components

gadomsky:
also
* fix `NkSerializableParsedSchema`
* fix generating test data for messages without schemaId (but only when latest schema expected)


utils/avro-components-utils/src/main/java/org/apache/flink/formats/avro/typeutils/NkSerializableParsedSchema.java
```diff
@@ -65,7 +67,11 @@ private void writeObject(ObjectOutputStream oos) throws IOException {
             } else if (schema instanceof JsonSchema) {
                 oos.writeByte(jsonSchemaType);
                 oos.writeUTF(((JsonSchema) schema).rawSchema().toString());
-            } else {
+            } else if (schema instanceof AvroSchemaWithJsonPayload) {
```
dswiecki, line 70: yeah I was thinking that it could fails in place on flink 👍

arkadius (approved):
LGTM

PR #3369, opened 2022-08-04
Fixed calling varArg functions with one argument

dawidsula26:


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/internal/DefaultSpelConversionsProvider.scala
```diff
@@ -23,10 +23,8 @@ class DefaultSpelConversionsProvider extends SpelConversionsProvider {
 
   override def getConversionService: GenericConversionService = {
     val service = new GenericConversionService
-    // We only need ObjectToArrayConverter, but it has package
-    // visibility so we add all collection converters.
-    DefaultConversionService.addCollectionConverters(service)
     service.addConverterFactory(new NumberToNumberConverterFactory())
+    service.addConverter(new ObjectToArrayConverter(service))
```
arkadius, line 27: Please add notice in comment that it is for varargs purpose

dawidsula26: What exactly should I test? Converting from object to array is something that is neccessary for other things to work, not a desired behaviour of ConversionProvider.

arkadius: Test that you already have :) spel expression with single parameter in vararg - for example usage of String.format

arkadius (approved):

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/internal/DefaultSpelConversionsProvider.scala
arkadius: Intresting. I missed spel epxression tests. I see that it is included in this PR. Sorry for hasle

PR #3379, opened 2022-08-05
Fix/avro union validation abr

arkadius:

### DIFF
--- a/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
+++ b/utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
@@ -197,8 +197,12 @@ class AvroSchemaOutputValidator(validationMode: ValidationMode) extends LazyLogg
       asSingleValidatedResults(unionValidationResults)
     }
 
+    def verifyUnion(union: TypedUnion): Boolean =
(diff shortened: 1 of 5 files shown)


utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
```diff
@@ -88,6 +90,15 @@ class AvroSchemaOutputValidator(validationMode: ValidationMode) extends LazyLogg
     }
   }
 
+  private def validateUnionInput(union: TypedUnion, schema: Schema, path: Option[String]) = {
+    if (validationMode == ValidationMode.strict && !union.possibleTypes.forall(validateTypingResult(_, schema, path).isValid))
```
lciolecki, line 94: Maybe better change it to positive result? I mean: 
```
if (validationMode == ValidationMode.strict && union.possibleTypes.forall(validateTypingResult(_, schema, path).isValid)) => valid
```
below the same and at the end return invalid?

lciolecki (approved):
Looks really grate!

utils/avro-components-utils/src/main/scala/pl/touk/nussknacker/engine/avro/encode/AvroSchemaOutputValidator.scala
lciolecki, line 94: Maybe better is check success? 
```
union.possibleTypes.forall(validateTypingResult(_, schema, path).isValid) => valid
```

PR #3388, opened 2022-08-09
New errors for generic functions

dawidsula26:


components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/GenericFunctionTypingError.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.engine.api.generics
+
+import cats.data.NonEmptyList
+
+sealed trait GenericFunctionTypingError
+
+object GenericFunctionTypingError {
+  case class ArgumentTypeError() extends GenericFunctionTypingError
```
arkadius, line 8: use case object instead

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/GenericFunctionTypingError.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  case class ArgumentTypeErrorWithSignatures(signatures: NonEmptyList[Signature]) extends GenericFunctionTypingError
```
arkadius, line 10: Developer shouldn't specify method name

arkadius (approved):

PR #3393, opened 2022-08-10
Fix choice docsUrl

dswiecki:


defaultModel/src/main/resources/defaultModelConfig.conf
```diff
@@ -13,7 +13,7 @@
     mapVariable: {
       docsUrl: ${baseDocsUrl}"BasicNodes#mapvariable"
     }
-    choice: {
+    switch: {
```
mproch, line 16: maybe some comment that it should match node name not label?

mproch (approved):

PR #3399, opened 2022-08-10
Turn off old kafka components

dswiecki:


utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -48,7 +48,7 @@ object KafkaConfig {
   import net.ceedubs.ficus.readers.EnumerationReader._
   import TopicsExistenceValidationConfig._
 
-  val lowLevelComponentsEnabled: Boolean = true
+  val lowLevelComponentsEnabled: Boolean = false
```
mproch, line 51: Let's change changelog/migrationguide, as currently it describes this flag as true by default?

dswiecki: yes yes, I remember

PR #3409, opened 2022-08-11
Static parameters change

dawidsula26:


engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/source/flink/KafkaSourceFactoryDefinitionExtractorSpec.scala
```diff
@@ -16,14 +16,14 @@ class KafkaSourceFactoryDefinitionExtractorSpec extends KafkaSourceFactoryProces
     // and they must not be returned by other services.
     extractedTypes should contain allOf (
       ClazzDefinition(Typed.genericTypeClass(classOf[SampleKey],Nil), Map(
-        "partOne" -> List(StaticMethodInfo(ParameterList(Nil, None), Typed[String], "partOne", None)),
-        "partTwo" -> List(StaticMethodInfo(ParameterList(Nil, None), Typed[Long], "partTwo", None)),
-        "toString" -> List(StaticMethodInfo(ParameterList(Nil, None), Typed[String], "toString", None))
+        "partOne" -> List(StaticMethodInfo(MethodTypeInfo(Nil, None, Typed[String]), "partOne", None)),
```
arkadius, line 19: maybe we should add factory methods like `MethodTypeInfo.noArg`, `MethodTypeInfo.withoutVarargs()`? This is is a part of API so it would be nice to have some helpers to speed up devs work

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -87,13 +97,12 @@ object TypeInfos {
         typeFunction(arguments).leftMap(_.map(SpelExpressionParseErrorConverter(this, arguments).convert(_)))
         }
         .map{res =>
-          if (!res.canBeSubclassOf(staticResult)) throw new AssertionError("Generic function returned type that does not match static parameters.")
-            res
+          if (!staticInfo.signatures.map(_.result).exists(res.canBeSubclassOf))
```
arkadius, line 100: IMO better would be to check against return type returned by `staticInfo.computeResultType`. Won't be?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -23,54 +24,63 @@ object TypeInfos {
...
   }
 
-  case class StaticMethodInfo(staticParameters: ParameterList,
-                              staticResult: TypingResult,
+  object StaticMethodInfo {
+    def apply(signatures: MethodTypeInfo, name: String, description: Option[String]): StaticMethodInfo =
+      StaticMethodInfo(NonEmptyList.one(signatures), name, description)
+  }
+
+  case class StaticMethodInfo(signatures: NonEmptyList[MethodTypeInfo],
```
arkadius, line 47: IMO would be better to replace make computing mainSignatures eager. I mean here pass one signature instead of Nel of signatures, remove mainSignature and explicitly named method like `StaticMethodInfo.flatten(signatures: MethodTypeInfo)` flatten/generalized/widened - I don't like "main" word in this context. Thanks to this approach this class will look like simple DTO - serialize/deserialize will create the same object.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -74,17 +74,16 @@ object TypeInfos {
...
-            throw new AssertionError("Generic function returned type that does not match static parameters.")
-          res
-        }
+    override def computeResultType(arguments: List[TypingResult]): ValidatedNel[ExpressionParseError, TypingResult] = {
+      val errorConverter = SpelExpressionParseErrorConverter(this, arguments)
+      val typeFromStaticInfo = staticInfo.computeResultType(arguments)
+      val typeCalculated = typeFromStaticInfo.andThen(_ => typeFunction(arguments).leftMap(_.map(errorConverter.convert)))
+      typeFromStaticInfo.toOption.zip(typeCalculated.toOption).foreach{ case (fromStatic, calculated) =>
+        if (!calculated.canBeSubclassOf(fromStatic))
+          throw new AssertionError("Generic function returned type that does not match static parameters.")
```
arkadius, line 83: Can u add info about:
1. which generic function
2. what static signature? 
I mean: Generic function: ${this} ... does not match : ${...}

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -1,103 +1,85 @@
...
 
-    def staticParameters: ParameterList
-
-    final def staticNoVarArgParameters: List[Parameter] = staticParameters.noVarArgs
-
-    final def staticVarArgParameter: Option[Parameter] = staticParameters.varArg
+    def signatures: NonEmptyList[MethodTypeInfo]
 
-    def staticResult: TypingResult
+    val mainSignature: MethodTypeInfo = signatures.toList.maxBy(_.parametersToList.length)
```
arkadius, line 16: This shouldn't be computed on this level. For StaticMethodInfo it is just signature, no computation is needed + add comment / rename as we discussed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/TypingFunction.scala
```diff
@@ -4,35 +4,24 @@ import cats.data.ValidatedNel
...
-   * Defaults to type extracted from methods signature if it is
-   * not specified.
+   * List of possible combinations of parameters and result types that
+   * this function can accept. Must be at least as specific as types that
+   * can be derived from associated method and less specific than
+   * computeResultType.
+   * Defaults to type of associated method if empty list is provided.
    */
-  def staticResult: Option[TypingResult] = None
+  def signatures: List[MethodTypeInfo] = Nil
```
arkadius, line 24: WDYT about Option[Nel]? will be more obvious that for None, it will be generated.

arkadius (approved):
Good job

PR #3412, opened 2022-08-12
Corrected method filter for generic types

dawidsula26:
Fixed bug where method filter in EspTypeUtils did not check for types like List[UndesiredType]

### DIFF
--- a/engine/flink/tests/src/test/resources/extractedTypes/defaultModel.json
+++ b/engine/flink/tests/src/test/resources/extractedTypes/defaultModel.json
@@ -3915,23 +3915,6 @@
           "varArgs": false
         }
       ],
-      "getUnits": [
-        {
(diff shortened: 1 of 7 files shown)


arkadius (approved):

engine/flink/tests/src/test/resources/extractedTypes/defaultModel.json
```diff
@@ -3915,23 +3915,6 @@
           "varArgs": false
         }
       ],
-      "getUnits": [
```
arkadius, line 3918: You have to do the same changes for DevClassExtractionTest

arkadius:
After fixing tests, please add changelog/migration guide as well. Someone can be surprised why he/she method disappear - even If this method wasn't used

PR #3424, opened 2022-08-17
Better typing in numeric

dawidsula26:


dawidsula26:
Continued as [Better types in operators and helper functions](https://github.com/TouK/nussknacker/pull/3436/files#top)

PR #3425, opened 2022-08-17
LiteRuntime: Healthcheck on the same port as req-resp

gadomsky:


engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -45,25 +51,35 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
-      _ <- healthCheckServer.start()
-      _ <- scenarioInterpreter.run()
-    } yield (), Duration.Inf)
+    val healthCheckProvider = new HealthCheckRoutesProvider(system, scenarioInterpreter)
+
+    val httpServer = Http().newServerAt(interface = httpConfig.interface, port = httpConfig.port)
 
+    val run = scenarioInterpreter.run()
+    val healthCheckRoutes = healthCheckProvider.routes()
+    val routes = scenarioInterpreter.routes().map(_ ~ healthCheckRoutes).getOrElse(healthCheckRoutes)
```
arkadius, line 73: Use `Directives.concat` instead

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -45,25 +51,35 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
+    httpServer.bind(routes).foreach { b =>
```
arkadius, line 74: Please sequence `run` with `bind` - otherwise we won't see binding failures in await

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -45,25 +51,35 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
-    val healthCheckServer = new HealthCheckServerRunner(system, scenarioInterpreter)
-    Await.result(for {
-      _ <- healthCheckServer.start()
-      _ <- scenarioInterpreter.run()
-    } yield (), Duration.Inf)
+    val healthCheckProvider = new HealthCheckRoutesProvider(system, scenarioInterpreter)
+
+    val httpServer = Http().newServerAt(interface = httpConfig.interface, port = httpConfig.port)
 
+    val run = scenarioInterpreter.run()
```
arkadius, line 71: runFuture

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
```diff
@@ -45,25 +51,35 @@ object NuKafkaRuntimeApp extends App with LazyLogging {
...
   System.exit(exitCode)
 
+  private val akkaHttpCloseTimeout = 10 seconds
+
   private def runAfterActorSystemCreation(): Unit = {
     val scenarioInterpreter = RunnableScenarioInterpreterFactory.prepareScenarioInterpreter(scenario, runtimeConfig, liteKafkaJobData, system)
     Runtime.getRuntime.addShutdownHook(new Thread() {
       override def run(): Unit = {
+        logger.info("Closing http server")
+        if (server != null) Await.result(server.terminate(akkaHttpCloseTimeout), akkaHttpCloseTimeout)
```
arkadius, line 61: Why you moved it from runAfterActorSystemCreation to here (and extracted volataile variable because of that)? Is it some reason that it should be executed in this order?

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/requestresponse/RequestResponseRunnableScenarioInterpreter.scala
```diff
@@ -29,28 +26,16 @@ class RequestResponseRunnableScenarioInterpreter(jobData: JobData,
...
-    implicit val materializer: Materializer = Materializer(actorSystem)
-    logger.info(s"Binding scenario route into ${requestResponseConfig.interface}:${requestResponseConfig.port}")
-    Http().newServerAt(
-      interface = requestResponseConfig.interface,
-      port = requestResponseConfig.port
-    ).bind(route.route).map { binding =>
-      logger.info("Scenario route bound")
-      server = binding
+    i.open()
+    interpreter = i
```
arkadius, line 37: I wonder if it would be better to move this all interpreter initialization to class body and make routes just a val (here - not in interface). Thanks to that, this class will be very easy, there won't be any constraint between order of execution of run and routes, status would just return TaskStatus.Running. WDYT?

engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/engine/embedded/requestresponse/RequestResponseDeploymentStrategy.scala
```diff
@@ -31,12 +31,12 @@ object RequestResponseDeploymentStrategy {
   import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 
   def apply(config: Config)(implicit as: ActorSystem, ec: ExecutionContext): RequestResponseDeploymentStrategy = {
-    new RequestResponseDeploymentStrategy(config.rootAs[RequestResponseConfig])
+    new RequestResponseDeploymentStrategy(config.rootAs[HttpConfig], config.rootAs[RequestResponseConfig])
```
arkadius, line 34: Maybe we should make the same layout of config as we have in docker? Am I right that it is now:
```
http { interface,port }
request-response { definitionMetadata }
```
?

engine/lite/kafka/integration-test/src/it/scala/pl/touk/nussknacker/engine/lite/kafka/BaseNuKafkaRuntimeDockerTest.scala
```diff
@@ -53,7 +53,7 @@ trait BaseNuKafkaRuntimeDockerTest extends ForAllTestContainer with BeforeAndAft
 
   protected lazy val kafkaClient = new KafkaClient(kafkaBoostrapServer, suiteName)
 
-  protected def mappedRuntimeManagementPort: Int = runtimeContainer.mappedPort(runtimeManagementPort)
+  protected def mappedRuntimeManagementPort: Int = runtimeContainer.mappedPort(runtimeApiPort)
```
arkadius, line 56: It is now mappedRuntimeApiPort

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/requestresponse/RequestResponseRunnableScenarioInterpreter.scala
gadomsky: I tought about it, but it `ScenarioInterpreterFactory.createInterpreter` compiles scenario, runs validation etc.. 

But let's try.

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: I think that you can remove `volatile private var server` field now. Can't you?

engine/lite/embeddedDeploymentManager/src/main/scala/pl/touk/nussknacker/engine/embedded/requestresponse/RequestResponseDeploymentStrategy.scala
arkadius: Unfortunately it requires entry in migration guide

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/requestresponse/RequestResponseRunnableScenarioInterpreter.scala
arkadius: :+1: It almost ok. Originally I was thinking about moving interpreter.open() to class body. After that it won't be any problem with returning TaskStatus.Running in status. I see two options:
1. We move interpreter.open() to class body
2. We add isOpened in `Lifecycle` or in `RequestResponseRunnableScenarioInterpreter` and we have `if (interpreter.isOpened) TaskStatus.Running else TaskStatus.DuringDeploy`
WDYT?

gadomsky: Right.
Lets go with option 1.

engine/lite/kafka/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/kafka/NuKafkaRuntimeApp.scala
arkadius: ... I've added one more slight detail - wrapped await with try-finally to properly handle edge case like bound exception

arkadius (approved):

PR #3436, opened 2022-08-19
Better types in operators and helper functions

dawidsula26:


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionParseError.scala
```diff
@@ -179,6 +179,28 @@ object SpelExpressionParseError {
...
+    case class DivisionByZeroError private(expression: String) extends OperatorError {
+      override def message: String = s"Division by zero: $expression"
+    }
+
+    object DivisionByZeroError {
+      def apply(expression: String): DivisionByZeroError =
+        new DivisionByZeroError(stripOperatorAST(expression))
+    }
+
+    case class TakingModuloZeroError private(expression: String) extends OperatorError {
```
arkadius, line 192: Why "taking"? Is it commonly used with modulo? Shouldn't be just ComputingModuloZeroError? Or just ModuloZeroError?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -104,6 +105,19 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
       case _ => PartTypeError.invalidNel
     }
 
+    def withTwoChildrenOfType[A: universe.TypeTag, R: universe.TypeTag](op: (A, A) => R) = {
+      val expectedType = Typed.fromDetailedType[A]
+      val resultType = Typed.fromDetailedType[R]
+      withTypedChildren {
+        case lst@left :: right :: Nil if lst.forall(_.typingResult.canBeSubclassOf(expectedType)) =>
+          val res = left.typingResult.value
```
arkadius, line 113: Can u rewrite it using for-comprehension? It will look better

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -104,6 +105,19 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
+            .flatMap(l => right.typingResult.value.map((l, _)))
+            .map{ case (x, y) => op(x.asInstanceOf[A], y.asInstanceOf[A]) }
```
arkadius, line 115: I think that we could create some generic extractor which could ensure that if `typingResult.canBeSubclassOf` some type then `typingResult.value` will return `Option[T]`. WDYT?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -210,44 +224,62 @@ private[spel] class Typer(classLoader: ClassLoader, commonSupertypeFinder: Commo
...
 
-      case e: OpAnd => withChildrenOfType[Boolean](TypingResultWithContext(Typed[Boolean]))
-      case e: OpOr => withChildrenOfType[Boolean](TypingResultWithContext(Typed[Boolean]))
-      case e: OpGE => withChildrenOfType[Number](TypingResultWithContext(Typed[Boolean]))
-      case e: OpGT => withChildrenOfType[Number](TypingResultWithContext(Typed[Boolean]))
-      case e: OpLE => withChildrenOfType[Number](TypingResultWithContext(Typed[Boolean]))
-      case e: OpLT => withChildrenOfType[Number](TypingResultWithContext(Typed[Boolean]))
+      case e: OpAnd => withTwoChildrenOfType[Boolean, Boolean](_ && _)
+      case e: OpOr => withTwoChildrenOfType[Boolean, Boolean](_ || _)
+      case e: OpGE => withTwoChildrenOfType[Number, Boolean](MathUtils.greaterOrEqual)
```
arkadius, line 229: Number and Boolean can't be inferred from MathUtils.greaterOrEqual signature by scala? :thinking:

interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -869,6 +871,37 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
    result shouldBe 'valid
   }
 
+  private def checkExpressionWithKnownResult(expr: String): Unit = {
```
arkadius, line 874: nice

interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -650,8 +652,8 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
   }
 
   test("be able to type string concatenation") {
-    parse[Any]("12 + ''", ctx).toOption.get.returnType shouldBe Typed[String]
-    parse[Any]("'' + 12", ctx).toOption.get.returnType shouldBe Typed[String]
+    parse[Any]("12 + ''", ctx).toOption.get.returnType shouldBe Typed.fromInstance("12")
```
arkadius, line 655: This tests double logic of `should calculate values of operators on strings`. Shouldn't we remove them?

utils/math-utils/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
```diff
@@ -49,17 +49,102 @@ trait MathUtils {
...
-        override def onBytes(n1: Byte, n2: Byte): Byte = throw new IllegalStateException("Bytes should be promoted to Ints before addition")
-        override def onShorts(n1: Short, n2: Short): Short = throw new IllegalStateException("Shorts should be promoted to Ints before addition")
-        override def onInts(n1: Int, n2: Int): Int = n1 + n2
-        override def onLongs(n1: Long, n2: Long): Long = n1 + n2
-        override def onBigIntegers(n1: BigInteger, n2: BigInteger): BigInteger = n1.add(n2)
-        override def onFloats(n1: Float, n2: Float): Float = n1 + n2
-        override def onDoubles(n1: Double, n2: Double): Double = n1 + n2
-        override def onBigDecimals(n1: java.math.BigDecimal, n2: java.math.BigDecimal): java.math.BigDecimal = n1.add(n2)
+      withValuesWithTheSameType(n1, n2)(new SameNumericTypeHandler[Number] {
+        override def onBytes(n1: Byte, n2: Byte): Number = n1 + n2
```
arkadius, line 140: Why u removed this assertion?

utils/math-utils/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
```diff
@@ -105,17 +190,16 @@ trait MathUtils {
...
-  protected trait SameNumericTypeHandler {
-    def onBytes(n1: Byte, n2: Byte): Byte
-    def onShorts(n1: Short, n2: Short): Short
-    def onInts(n1: Int, n2: Int): Int
-    def onLongs(n1: Long, n2: Long): Long
-    def onBigIntegers(n1: java.math.BigInteger, n2: java.math.BigInteger): java.math.BigInteger
-    def onFloats(n1: Float, n2: Float): Float
-    def onDoubles(n1: Double, n2: Double): Double
-    def onBigDecimals(n1: java.math.BigDecimal, n2: java.math.BigDecimal): java.math.BigDecimal
+  protected trait SameNumericTypeHandler[R] {
```
arkadius, line 193: why this R is need?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -85,16 +92,20 @@ object typing {
...
 
+    override def value: Option[Any] = underlying.value
+
     override def withoutValue: TypedTaggedValue = TypedTaggedValue(underlying.withoutValue, tag)
 
     override def display: String = s"${underlying.display} @ $tag"
   }
 
-  case class TypedObjectWithValue private[typing](underlying: TypedClass, value: Any) extends TypedObjectWithData {
+  case class TypedObjectWithValue private[typing](underlying: TypedClass, justValue: Any) extends TypedObjectWithData {
```
arkadius, line 102: I think that we rather use "opt"/"optional". Also value should be rather `valueOpt` instead of changing of this field. Be aware that this is part of our core API and each change will need migration guide entry

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -57,6 +59,9 @@ object typing {
   case class TypedObjectTypingResult(fields: ListMap[String, TypingResult],
                                      objType: TypedClass,
                                      additionalInfo: Map[String, AdditionalDataValue] = Map.empty) extends SingleTypingResult {
+    override def value: Option[Map[String, Any]] =
```
arkadius, line 62: Can we preserve order of fields (ListMap). I'm not sure if someone will use it, but It was such requirement for types, for values could be the same

utils/math-utils/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
dawidsula26: We want to handle any possible return type without using asInstanceOf. In most cases R is still Number, but in compare it must be Int.

arkadius: Concrete return type for each method had advantage that someone won't introduce e.g. overflow bug - types will protect you from doing sth like xxx.toLong everywhere. All math operations should promote to wider type before doing operation that gives wider return type. IMO we don't need to handle compare in the same way. IMO you can assume that all numbers are Comparable[_] as well, so the only thing that you should do is to cast to Comparable and invoke compare method on them

utils/math-utils/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
arkadius: One important think is (maybe not written anywhere in comment) that those utils are in the "spel" spirit. In spel, short and bytes are promoted to int before summing which is quite reasonable for such concise languages

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
dawidsula26: Do you have some reasonable situation where `typingResult.value.map(_.asInstanceOf[T])` fails, but `typingResult.valueAsType[T]` works correctly? I think that, every time .value is used, we already are sure what type we expect there, and it would be better to fail than try to handle such case.

utils/math-utils/src/main/scala/pl/touk/nussknacker/engine/util/MathUtils.scala
```diff
@@ -191,15 +197,34 @@ trait MathUtils {
...
+    override def onShorts(n1: lang.Short, n2: lang.Short): java.lang.Short
+    override def onInts(n1: java.lang.Integer, n2: java.lang.Integer): java.lang.Integer
+    override def onLongs(n1: java.lang.Long, n2: java.lang.Long): java.lang.Long
+    override def onBigIntegers(n1: java.math.BigInteger, n2: java.math.BigInteger): java.math.BigInteger
+    override def onFloats(n1: java.lang.Float, n2: java.lang.Float): java.lang.Float
+    override def onDoubles(n1: java.lang.Double, n2: java.lang.Double): java.lang.Double
+    override def onBigDecimals(n1: java.math.BigDecimal, n2: java.math.BigDecimal): java.math.BigDecimal
+  }
+
+  protected trait SameNumericTypeHandlerForMathOp extends SameNumericTypeHandlerReturningNumber {
```
arkadius, line 221: Maybe `SameNumericTypeHandlerForPromotingMathOp`? Because it is not good choice for all math ops.

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
arkadius: I'm not sure if I get you correct. By my comment I mean sth like this: https://github.com/TouK/nussknacker/pull/3442/files  . Do you think it gives some value? The problem I've seen was that between `canBeSubclassOf` and `asInstanceOf` can happens many things which cause ClassCastException at the end. Thanks to extractors, it is done in one step so we are safe that nothing danger happen. It is similar approach to: https://www.scala-exercises.org/shapeless/type_safe_cast

dawidsula26: I thought you ment something different. What you proposed is good, I have merged it.

arkadius (approved):

PR #3441, opened 2022-08-22
Upgrade Flink to 1.15

dawidsula26:


dawidsula26:
Tests from file may not work even when all tests pass. It must be manualy tested before merging.

mproch (approved):
:+1: 
Maybe let's add some more info in MigrationGuide? Like: 
- Some of the Flink artifacts no longer have Scala version
- Be careful to set checkpointInterval to low value in custom component tests, as Flink performs additional checkpoint after source finishes.

PR #3451, opened 2022-08-24
Remove JsonToTypedMapConverter

### LINKED ISSUE #3226: `application/json` not fully supported in OpenAPI enrichments
When using OpenAPI components with a service providing `application/json` request body not all JSON values are supported.
The only supported value is a JSON object.
When an OpenAPI service specifies a request as a JSON array, string, number or boolean, an empty request is always sent to the service with the `Content-type` header set to `text/plain`. Although the designer has a form field to describe the request body.

For instance, the following is a valid request definition, which is not handled properly:
```
openapi: 3.0.0
[...]
paths:
  /some:
    post:
      [...]
      requestBody:
        content:
          'application/json':
            schema:
              type: string
[...]
```

dswiecki:
Fixes #3226


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/serde/CirceJsonDeserializer.scala
```diff
@@ -22,8 +24,28 @@ class CirceJsonDeserializer(jsonSchema: Schema) {
...
+    //after validate jsonObject has set default field values
+    catchValidationError(jsonSchema.validate(jsonObject))
+    Validated.fromEither(CirceUtil.decodeJson[Json](jsonObject.toString).map(JsonToObject.apply(_, swaggerTyped)))
+  }
+
+  protected def prepareValidationErrorMessage(exception: Throwable): String = {
+    exception match {
+      case ve: ValidationException => ve.getAllMessages.asScala.mkString("\n\n")
+      case je: JSONException => s"Invalid JSON: ${je.getMessage}"
+      case _ => "unknown error message"
```
lciolecki, line 40: Can we start message from big letter? And also I think it would be nice to see exception message?

dswiecki: It's moved code from JsonSchemaRequestResponseSource, i don't want to change de behaviour

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/serde/CirceJsonDeserializer.scala
```diff
@@ -22,8 +24,28 @@ class CirceJsonDeserializer(jsonSchema: Schema) {
   }
 
   def deserialize(string: String): Validated[circe.Error, Any] = {
+    //we basically do parsing twice:
+    //1. for schema validation
+    //2. for typing purposes which is based on Circe
     val jsonObject = new JSONTokener(string).nextValue()
-    jsonSchema.validate(jsonObject)
-    Validated.fromEither(CirceUtil.decodeJson[Json](string).map(JsonToObject.apply(_, swaggerTyped)))
+    //after validate jsonObject has set default field values
```
arkadius, line 31: nice comments, thanks

arkadius (approved):

PR #3460, opened 2022-08-25
Add config flag to skip redirect after archive. reload page after unarchive

Filemon279:


JulianWielga (approved):

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -680,6 +680,7 @@ Currently, you can only configure secondary environment if it uses BASIC authent
 | intervalTimeSettings.healthCheck | Low        | int     | 30000         | How often frontend reloads checks scenarios states                                                                                                                                                                          |
 | developmentMode                  | Medium     | boolean | false         | For development mode we disable some security features like CORS. **Don't** use in production                                                                                                                               |
 | enableConfigEndpoint             | Medium     | boolean | false         | Expose config over http (GET /api/app/config/) - requires admin permission. Please mind, that config often contains password or other confidential data - this feature is meant to be used only on 'non-prod' envrionments. |
+| skipArchiveRedirect              | Low        | boolean |               | Preventing from redirect to scenarios list after doing archive oprtation                                                                                                                                                    |
```
arkadius, line 683: can make it positive instead of negative? I mean redirectAfterArchive or sth like that

ui/server/src/main/scala/pl/touk/nussknacker/ui/config/FeatureTogglesConfig.scala
```diff
@@ -42,6 +43,7 @@ object FeatureTogglesConfig extends LazyLogging{
     val tabs = parseOptionalConfig[List[TopTab]](config, "tabs")
     val intervalTimeSettings = config.as[IntervalTimeSettings]("intervalTimeSettings")
     val testDataSettings = config.as[TestDataSettings]("testDataSettings")
+    val skipArchiveRedirect = config.getAs[Boolean]("skipArchiveRedirect")
```
dswiecki, line 46: can we make it backward compatible so that if no skipArchiveRedirect in configuration everything works fine ?

Filemon279: I think it is right now - `getAs[A]` works like `as[Option[A]]` so if config path is not defined we will end up with `None`

docs/installation_configuration_guide/DesignerConfiguration.md
Filemon279: Yes, we can - please find new version with `redirectAfterArchive` and default value set to true when no configuration is set.

ui/server/src/main/scala/pl/touk/nussknacker/ui/config/FeatureTogglesConfig.scala
dswiecki: 👍

dswiecki (approved):

arkadius (approved):

PR #3464, opened 2022-08-26
Universal kafka sink key: using null (empty) expression instead expli…

arkadius:
…cit null literal

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/editor/ParameterTypeEditorDeterminer.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/editor/ParameterTypeEditorDeterminer.scala
@@ -19,7 +19,7 @@ class ParameterTypeEditorDeterminer(val typ: TypingResult) extends ParameterEdit
       case klazz if classOf[java.lang.CharSequence].isAssignableFrom(klazz) =>
         DualParameterEditor(
           simpleEditor = StringParameterEditor,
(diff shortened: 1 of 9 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/parameter/editor/ParameterTypeEditorDeterminer.scala
```diff
@@ -13,14 +13,13 @@ class ParameterTypeEditorDeterminer(val typ: TypingResult) extends ParameterEdit
...
-        // We can pick here simple editor, but for compatibility reasons we choose dual editor instead
-        // - to be able to provide some other expression
         DualParameterEditor(FixedValuesParameterEditor(
           possibleValues = klazz.getEnumConstants.toList.map(ParameterTypeEditorDeterminer.extractEnumValue(klazz))
         ), DualEditorMode.SIMPLE)
-      case klazz if klazz == classOf[java.lang.String] =>
+      case klazz if classOf[java.lang.CharSequence].isAssignableFrom(klazz) =>
         DualParameterEditor(
           simpleEditor = StringParameterEditor,
+          // Ror compatibility reasons we choose raw editor - otherwise users could write expressions in simple mode
```
gadomsky, line 22:
```suggestion
          // For compatibility reasons we choose raw editor - otherwise users could write expressions in simple mode
```

gadomsky (approved):
LGTM

PR #3475, opened 2022-08-30
hide parameters & errors blink

JulianWielga:


mproch:
We fix it on the backend with https://github.com/TouK/nussknacker/pull/3497

PR #3484, opened 2022-08-30
Remove cross compile job

dswiecki:


dswiecki:
we'll have two version back again soon

PR #3485, opened 2022-08-31
Kafka doc refactor

mproch:

### DIFF
--- a/docs/integration/KafkaIntegration.md
+++ b/docs/integration/KafkaIntegration.md
@@ -45,3 +45,149 @@ At runtime Nussknacker determines the schema version of a message value and key 
 1. it checks `value.schemaId` and `key.schemaId` headers;
 2. if there are no such headers, it looks for a magic byte and schema version in the message, [in a format used by Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format);
 3. if the magic byte is not found, it assumes the schema version chosen by the user in the scenario.
+
+
+## Common Kafka configuration
+
+### Kafka connection configuration
+
+Both engines share common Kafka configuration, see [Streaming-Lite](model/Lite#kafka-configuration) or [Streaming-Flink](model/Flink#kafka-configuration) docs for details on configuring sources/sinks.
+
+Important thing to remember is that Kafka server addresses/schema registry addresses have to be resolvable from:
+- Nussknacker Designer host (to enable schema discovery and scenario testing)
+- Streaming-Lite runtime - to be able to run job
+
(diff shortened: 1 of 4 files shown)


docs/integration/KafkaIntegration.md
```diff
@@ -44,4 +44,136 @@ In the Designer the user can choose, for each source and sink, which schema vers
 At runtime Nussknacker determines the schema version of a message value and key in the following way:
 1. it checks `value.schemaId` and `key.schemaId` headers;
 2. if there are no such headers, it looks for a magic byte and schema version in the message, [in a format used by Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format);
-3. if the magic byte is not found, it assumes the schema version chosen by the user in the scenario.
\ No newline at end of file
+3. if the magic byte is not found, it assumes the schema version chosen by the user in the scenario.
+
+  
+# Configuration
```
ZbyszekMM, line 50: A major question and difficulty I always had with configuration was trying to figure out where exactly in this tree of configuration settings the given setting should be? What is convention used in this documentation to answer this question? 

The answer is _I do not know, but may be I omitted something_. So, if this convention is not obvious, I would consider addition something like this:
_Unless stated otherwise all settings here are relative scenarioTypes.ScenarioTypeName.modelConfig key. You can find the high level structure of the configuration file [here](https://nussknacker.io/documentation/docs/installation_configuration_guide/#configuration-areas).

docs/integration/KafkaIntegration.md
```diff
@@ -44,4 +44,136 @@ In the Designer the user can choose, for each source and sink, which schema vers
...
+
+## Common part
```
ZbyszekMM, line 52: The key relative which Kafka is configured in Streaming-Flink and Streaming-Lite is (most likely) different. I think it should be mentioned here ..

docs/integration/KafkaIntegration.md
ZbyszekMM, line 50: I would also mention that keys under which Kafka is configured for Lite and Flink are different, the reason being that in Flink case it is "just a" (?) component <there could be a different input(?)>, in Lite case it is "at the core"  of the specific, hm, version.

mproch: :+1:

mproch: Mentioned in respective sections

docs/integration/KafkaIntegration.md
mproch: Added mention

ZbyszekMM (approved):

PR #3494, opened 2022-09-01
GZIP content + Cache-Control header

mproch:

### DIFF
--- a/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
+++ b/ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
@@ -20,23 +20,21 @@ class WebResources(publicPath: String) extends Directives with LazyLogging {
     tempMainContentFile
   }
 
-  val route: Route =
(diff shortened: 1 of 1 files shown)


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/WebResources.scala
```diff
@@ -20,23 +22,23 @@ class WebResources(publicPath: String) extends Directives with LazyLogging {
...
+  }
+
+  private def handleAssets(webSubfolder: String) = {
+    pathPrefix(webSubfolder) {
       get {
-        extractRequest { matched =>
-          logger.debug(s"Try to get static data:  ${matched.uri.path}.")
-          getFromResourceDirectory("web/static")
+        encodeResponse {
+          respondWithHeader(`Cache-Control`(List(CacheDirectives.public, CacheDirectives.`max-age`(0)))) {
```
gadomsky, line 34: there's a lot of javascripts there. why do you not want to cache it?

gadomsky (approved):

PR #3501, opened 2022-09-02
fixed window content after history back action

JulianWielga:


ui/client/components/graph/node-modal/node/NodeDetails.tsx
```diff
@@ -105,6 +105,11 @@ export function NodeDetails(props: NodeDetailsProps): JSX.Element {
     return {HeaderTitle}
   }, [node])
 
+  //no process? no nodes? no window contents! no errors for whole tree!
```
mproch, line 108: what about empty scenario?

JulianWielga: nodeDetails on empty scenario?

JulianWielga: I see... properties... but it's not this case. empty scenario has empty array in nodes.

mproch (approved):
Cool, thx :+1: Please remember to backport it to release/1.5 branch

PR #3502, opened 2022-09-05
refactor kafka properties

gadomsky:


utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -25,10 +24,16 @@ case class KafkaConfig(kafkaAddress: String,
...
                        useStringForKey: Boolean = true,
                        schemaRegistryCacheConfig: SchemaRegistryCacheConfig = SchemaRegistryCacheConfig(),
-                       avroAsJsonSerialization: Option[Boolean] = None
-                      ) {
+                       avroAsJsonSerialization: Option[Boolean] = None,
+                       kafkaAddress: Option[String] = None) {
 
-  def schemaRegistryClientKafkaConfig = SchemaRegistryClientKafkaConfig(kafkaProperties.getOrElse(Map.empty), schemaRegistryCacheConfig, avroAsJsonSerialization)
+  // delete this method, make `kafkaProperties` required and use it instead, when `kafkaAddress` is removed
+  def definedKafkaProperties = kafkaAddress.map(bs => Map("bootstrap.servers" -> bs)).getOrElse(Map.empty) ++ kafkaProperties.getOrElse(Map.empty)
```
mproch, line 31: maybe we can skip adding this method, and add address if needed only in KafkaUtils.withPropertiesFromConfig?

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -25,10 +24,16 @@ case class KafkaConfig(kafkaAddress: String,
...
+
+  //is that check needed? Sometimes we do not need this property eg. to create SchemaRegistryClient
+  require(definedKafkaProperties.contains("bootstrap.servers"), "Missing 'bootstrap.servers' property in kafkaProperties")
```
mproch, line 34: I'm not sure if it's needed - in almost all settings except for simple development/demo you have to set other properties anyway...

engine/lite/k8sDeploymentManager/src/test/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerKafkaTest.scala
```diff
@@ -355,11 +355,11 @@ class K8sDeploymentManagerKafkaTest extends BaseK8sDeploymentManagerTest
   }
 
   private val kafkaDeployConfig: Config = baseDeployConfig("streaming")
-    .withValue("configExecutionOverrides.modelConfig.kafka.kafkaAddress", fromAnyRef(s"${KafkaK8sSupport.kafkaServiceName}:9092"))
+    .withValue("configExecutionOverrides.modelConfig.kafka.kafkaProperties.\"bootstrap.servers\"", fromAnyRef(s"${KafkaK8sSupport.kafkaServiceName}:9092"))
     .withValue("configExecutionOverrides.modelConfig.kafka.kafkaProperties.\"schema.registry.url\"", fromAnyRef(s"http://${KafkaK8sSupport.srServiceName}:8081"))
   private val modelData: LocalModelData = LocalModelData(ConfigFactory.empty
     //e.g. when we want to run Designer locally with some proxy?
-    .withValue("kafka.kafkaAddress", fromAnyRef("localhost:19092"))
+    .withValue("kafka.kafkaProperties.\"bootstrap.servers\"", fromAnyRef("localhost:19092"))
```
mproch, line 362: I wonder, maybe some helper would be nice? Though we also use other properties quite often...

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
gadomsky: but purpose of this method is also to get config as `Map[String,String]`, now both `kafkaAddress` and `kafkaProperties` are Options. 

If I was to delete this method, I would have to do in several places:  `kafkaProperties.getOrElse(Map.empty)` or change api in non-compatible way - that is change to: `kafkaProperties: Map[String,String]`

gadomsky: Deleted, please see the last commit

mproch: thx

mproch (approved):
Cool, thx

PR #3503, opened 2022-09-06
Check max surge when validating K8s deployment

mproch:


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -115,10 +115,12 @@ class K8sDeploymentManager(modelData: BaseModelData, config: K8sDeploymentManage
 
   override def validate(processVersion: ProcessVersion, deploymentData: DeploymentData, canonicalProcess: CanonicalProcess): Future[Unit] = {
     val scalingOptions = determineScalingOptions(canonicalProcess)
+    val deploymentStrategy = deploymentPreparer.deploymentStrategy(canonicalProcess.metaData.typeSpecificData)
```
gadomsky, line 118: does `deploymentPreparer.deploymentStrategy` not return only default deployment strategy? I think, it could be overriden by `userConfigurationBasedDeployment` in `DeploymentPreparer.prepare`

mproch: eeee... you're fully right, thanks...

gadomsky (approved):
LGTM

PR #3507, opened 2022-09-07
Modules cleanup after runtime-app

arkadius:


build.sbt
```diff
@@ -476,8 +476,17 @@ lazy val requestResponseRuntime = (project in lite("request-response/runtime")).
...
+    libraryDependencies ++= {
+      Seq(
+        "com.typesafe.akka" %% "akka-http" % akkaHttpV,
+        "com.typesafe.akka" %% "akka-stream" % akkaV,
+        "com.typesafe.akka" %% "akka-testkit" % akkaV % "test",
+        "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpV % "test"
+      )
+    }
   ).
+  // TODO: remove deploymentManagerApi from deps after removing requestResponseApp
```
mproch, line 489: :+1:

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/HttpConfig.scala
```diff
@@ -0,0 +1,3 @@
+package pl.touk.nussknacker.engine.lite
```
mproch, line 1: why is it in this module? I'd expect it as private to App/embedded?

mproch (commented):
Cool, added one minor comment

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/HttpConfig.scala
arkadius: Because we reuse it in embedded DM. BTW it looks like some common http binding settings. Maybe I should move it to some utils-internal and rename to HttpBindingConfig? or maybe I should duplicate it?

mproch: TBH I'd prefer duplicating. But utils-internal is also okeeish

arkadius: For me duplication of this class is fine as well :+1: - duplicated

mproch (approved):

PR #3511, opened 2022-09-08
Trigger sample components + flink-compat builds

arkadius:


mproch (approved):
I'm wondering if using matrix for all this triggering would make it nicer?

PR #3513, opened 2022-09-09
Limit use of EspProcess to compiler internals

mproch:


scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/CanonicalProcess.scala
```diff
@@ -2,18 +2,22 @@ package pl.touk.nussknacker.engine.canonicalgraph
...
 import pl.touk.nussknacker.engine.graph.node._
 import pl.touk.nussknacker.engine.marshall.ProcessMarshaller
 
+import scala.language.implicitConversions
+
 sealed trait CanonicalTreeNode
 
 object CanonicalProcess {
 
+  //TODO: replace EspProcess use with CanonicalProcess where needed
```
arkadius, line 18: To removal?

arkadius (approved):
Good job, one comment added

scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/CanonicalProcess.scala
mproch: :+1:

PR #3540, opened 2022-09-16
Dev setup improvements

arkadius:

### DIFF
--- a/ui/buildServer.sh
+++ b/ui/buildServer.sh
@@ -2,7 +2,5 @@
 
 set -e
 
-cd ..
-export addDevArtifacts=true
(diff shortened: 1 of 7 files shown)


ui/buildServer.sh
```diff
@@ -2,7 +2,5 @@
 
 set -e
 
-cd ..
-export addDevArtifacts=true
-./sbtwrapper "set ThisBuild / packageDoc / publishArtifact := false; set Compile / doc / sources := Seq.empty; dist/Universal/stage"
-cd -
+cd "$(dirname $0)"/..
```
piotrp, line 5: `$0` should be quoted, or this will break when path contains space character. There are also a few more shell scripts where this was omitted.

piotrp (changes requested):

piotrp (approved):

PR #3542, opened 2022-09-19
New feature: properties additional info

bartektartanus:
New feature, similar to `NodeAdditionalInfo`


ui/server/src/main/scala/pl/touk/nussknacker/ui/api/PropertiesResources.scala
```diff
@@ -0,0 +1,116 @@
...
+import pl.touk.nussknacker.restmodel.process.ProcessingType
+import pl.touk.nussknacker.ui.process.subprocess.SubprocessRepository
+import pl.touk.nussknacker.ui.validation.AdditionalPropertiesValidator
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * This class should contain operations invoked for each node (e.g. node validation, retrieving additional data etc.)
+  */
+class PropertiesResources(val processRepository: FetchingProcessRepository[Future],
```
mproch, line 36: I wonder if it's necessary to create new resource for that? Maybe we can merge it with NodeResources, they do the same operations, it'll be easier to keep them in sync?

ui/client/actions/nk/nodeDetails.ts
```diff
@@ -44,19 +45,21 @@ export function nodeValidationDataClear(nodeId: string): NodeValidationClear {
 //TODO: use sth better, how long should be timeout?
 const validate = debounce(async (processId: string, validationRequestData: ValidationRequest, callback: (data: ValidationData, nodeId: NodeId) => void) => {
   const nodeId = validationRequestData.nodeData.id
-  const {data} = await HttpService.validateNode(processId, validationRequestData)
-  callback(data, nodeId)
+  const nodeType = NodeUtils.nodeType(validationRequestData.nodeData);
+  if (nodeType === "Properties") {
```
mproch, line 49: Do we have constant/method for this check?

bartektartanus: Yes, we do. 
Done.

PR #3551, opened 2022-09-22
Flink test runner allows to test sources and sinks

jedrz:
For now unbounded sources, like Kafka are still not supported.


utils/components-testkit/src/main/scala/pl/touk/nussknacker/engine/util/test/TestScenarioRunner.scala
```diff
@@ -53,25 +54,31 @@ trait TestScenarioRunnerBuilder[R <: TestScenarioRunner, B <: TestScenarioRunner
...
+  def successes[T](data: List[T]): RunListResult[T] =
+    RunListResult(Nil, data)
 
-  def errors[T](errors: List[NuExceptionInfo[_]]): RunResult[T] =
-    RunResult[T](errors, List())
+}
 
+sealed trait RunResult[T] {
+  def errors: List[NuExceptionInfo[_]]
+  def successes: T
```
arkadius, line 72: Let's name this success - because on this level it is still singular

utils/flink-components-testkit/src/main/scala/pl/touk/nussknacker/engine/flink/util/test/FlinkTestScenarioRunner.scala
```diff
@@ -30,18 +30,43 @@ object testComponents {
...
+    val componentsWithTestComponents = this.components ++ componentsWithData
+    run(scenario, new TestServiceInvocationCollector(testComponentHolder.runId), Some(testComponentHolder.runId), componentsWithTestComponents).map { _ =>
+      collectResults()
+    }
+  }
+
+  /**
+   * Can be used to test Flink bounded sources - we wait for the scenario to finish.
+   */
+  def runAndCollectResults[R](scenario: CanonicalProcess): RunnerListResult[R] = {
```
arkadius, line 49: 1. I'm thinking a while if this name is consistent with others. We have runWithData and no CollectingResults suffix. To be consistent it probable should be named just `run` or other methods should be renamed to `runWithDataCollectingResults`. On the other hand just `run` looks a little bit weird and `runWithDataCollectingResults` looks too much verbose. WDYT?
2. `testComponents.withDataList` registers `noopSource` as well. Can you add this component to be consistent with other run methods? It is useful in some cases

utils/flink-components-testkit/src/main/scala/pl/touk/nussknacker/engine/flink/util/test/FlinkTestScenarioRunner.scala
```diff
@@ -30,18 +30,43 @@ object testComponents {
...
 
+  /**
+   * Can be used to test Flink based sinks.
+   */
+  def runWithDataIgnoringResults[I: ClassTag](scenario: CanonicalProcess, data: List[I]): RunnerResult[Unit] = {
     implicit val typeInf: TypeInformation[I] = TypeInformation.of(implicitly[ClassTag[I]].runtimeClass.asInstanceOf[Class[I]])
-    val modelData = LocalModelData(config, new EmptyProcessConfigCreator)
     val componentsWithData = testComponents.withDataList(data)
     val componentsWithTestComponents = this.components ++ componentsWithData
+    run(scenario, ProductionServiceInvocationCollector, None, componentsWithTestComponents)
```
arkadius, line 62: You register here `invocationCollector` component, because it is done implicitly in  `run` method. Can you refactor this code a bit, so preparing all test components (source, noop and this invocationCollector) will be on the same level? It will be easier to analyze how all those methods are different from each other.

arkadius (approved):
LGTM

PR #3553, opened 2022-09-22
Check slug uniqueness

mproch:


arkadius (approved):

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -30,7 +30,7 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
 
 
   //TODO: exception when non-request-response process?
-  def path(p: Option[String]) =
+  def slug(p: Option[String]) =
```
arkadius, line 33: We need a migration guide entry I guess

mproch: :+1:

PR #3558, opened 2022-09-23
Cleanup K8s service after slug change

mproch:

### DIFF
--- a/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
+++ b/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
@@ -144,6 +144,7 @@ class K8sDeploymentManager(modelData: BaseModelData, config: K8sDeploymentManage
                       canonicalProcess: CanonicalProcess,
                       savepointPath: Option[String]): Future[Option[ExternalDeploymentId]] = {
     val scalingOptions = determineScalingOptions(canonicalProcess)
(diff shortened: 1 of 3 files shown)


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -167,6 +170,11 @@ class K8sDeploymentManager(modelData: BaseModelData, config: K8sDeploymentManage
         requirementForName(processVersion.processName),
         secretIdLabel isNot secret.name
       ))
+      //cleaning up after after possible slug change
+      _ <- k8s.deleteAllSelected[ListResource[Service]](LabelSelector(
+        requirementForName(processVersion.processName),
```
arkadius, line 175: I wonder if for cleaning won't be better to use requirementForId in case if someone change the name. For cancel and findJobStatus we have TODO - I know. WDYT? Will it work correctly for rename case?

arkadius (commented):
LGTM, I've left one thing to consider

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
mproch: :+1: should be working ATM, as we don't allow redeploy of scenarios that are not running. Nevertheless, I changed to requirementForId in all places where we have access

arkadius (approved):

PR #3559, opened 2022-09-23
Fix: Cypress fragment outputs

lciolecki:
Temprary rollback to old version -> without changing output name.


ui/client/cypress/integration/fragment.ts
```diff
@@ -44,7 +44,7 @@ describe("Fragment", () => {
     cy.wait(750)
     cy.get("[data-testid=window]").toMatchImageSnapshot()
 
-    cy.get("[data-testid=window]").contains("testOutput").parent().find("input").type("fragmentResult")
+    cy.get("[data-testid=window]").contains("testOutput").parent().find("input").clear().type("fragmentResult")
```
arkadius, line 47: shouldn't be `fragmentResult <dot>` ?

lciolecki: Nope, here we put dedicated name for output. Below is `fragmentResult` with <dot>.

arkadius (approved):

PR #3566, opened 2022-09-26
OpenApi invocation path relative to definition path + servers configuration as a list instead of single entry

arkadius:


engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/openapi/RequestResponseOpenApiGenerator.scala
```diff
@@ -1,47 +1,94 @@
...
-import pl.touk.nussknacker.engine.requestresponse.RequestResponseInterpreter.RequestResponseScenarioInterpreter
 import pl.touk.nussknacker.engine.requestresponse.api.openapi.RequestResponseOpenApiSettings.OPEN_API_VERSION
 import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
 
-import scala.language.higherKinds
+import scala.collection.JavaConverters._
 
-object RequestResponseOpenApiGenerator {
+class RequestResponseOpenApiGenerator(oApiInfo: OApiInfo, serversDescription: Option[OApiServer], additionalInvocationPaths: List[String],
+                                      invocationPathFromDefinitionPath: List[String] => List[String]) extends LazyLogging {
```
mproch, line 13: why is it defined in ScenarioRoute? When would we want to provide different function?
Also, shouldn't we have same logic in line 24?

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/openapi/RequestResponseOpenApiGenerator.scala
```diff
@@ -1,47 +1,94 @@
...
+    generateOpenApiForPathList(List(toInvocationPath(request) -> pathDefinitionGenerator) ++ additionalInvocationPaths.map(_ -> pathDefinitionGenerator))
+  }
 
-  def generateOpenApi[Effect[_]](pathWithInterpreter: List[(String, RequestResponseScenarioInterpreter[Effect])], oApiInfo: OApiInfo, serversDescription: Option[OApiServer]): String = {
-    val scenarioDefinitions: Json = generateScenarioDefinitions(pathWithInterpreter)
-    OApiDocumentation(OPEN_API_VERSION, oApiInfo, serversDescription.map(List(_)), scenarioDefinitions).asJson.spaces2
+  private def toInvocationPath(request: HttpRequest): String = {
+    val originalPathSegmentsOpt = request.headers.collectFirst {
+      // It can be configured using https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-add-original-uri-header
+      case HttpHeader("x-original-uri", originalPath) =>
```
mproch, line 22: Is it standard by any means? e.g. https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-original-path
etc. Maybe it should be also configurable?

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/openapi/RequestResponseOpenApiGenerator.scala
arkadius: Ad1. Because this "one level up" logic is related with how we organize routes in ScenarioRoute.
Ad2. invocationPathFromDefinitionPath is invoked on both originalPathSegments and request.getUri() segments - there is getOrElse before invocation of this method

arkadius:
I changed implementation - now I uses relative server url instead of default /, paths are not changed

mproch (approved):

PR #3568, opened 2022-09-27
UI: Order test cases by context id

rafa-minimal:
The list of test cases is in random order (as returned from backend) and might differ from one node to another. The goal of this change is to have a consistent ordering of test cases (ordered by context id).


ui/client/common/TestResultUtils.ts
```diff
@@ -91,7 +91,7 @@ class TestResultUtils {
   }
 
   private _nodeResults(results: TestResults, nodeId: NodeId): NodeResult[] {
-    return results?.nodeResults?.[nodeId] || []
+    return sortBy(results?.nodeResults?.[nodeId] || [], (x) => x.context.id)
```
arkadius, line 94: shouldn't it be sorted on BE?

rafa-minimal: It's the same question I asked myself :D
Ok, let me change it

rafa-minimal: Moved sorting to the backend (sort each list by context id for consistency)

arkadius (approved):

PR #3574, opened 2022-09-29
Nu custom logo

bartektartanus:


ui/client/components/MenuBar.tsx
```diff
@@ -60,6 +60,10 @@ type Props = {
 
 const Spacer = () => <Flex flex={1}/>
 
+function NussknackerLogo() {
+  return <img src={absoluteBePath("/assets/img/nussknacker-logo.svg")} className={"navbar-brand-logo"}/>;
```
arkadius, line 64: We use sth like `UrlUtils.absoluteBePath` which is easy to test on dev setup - because this asset is taken from BE instead of FE. Also you should move this asset on BE side to /web/static/assets.

arkadius (approved):

arkadius (approved):

PR #3582, opened 2022-09-30
Move slug validation to DeploymentManager

mproch:

### DIFF
--- a/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
+++ b/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
@@ -44,7 +44,7 @@ class K8sDeploymentManagerProvider extends DeploymentManagerProvider {
 
   override def additionalValidators(config: Config): List[CustomProcessValidator] = forMode(config)(
     _ => Nil,
-    config => List(new LiteScenarioValidator(config.nussknackerInstanceName))
+    config => List(new RequestResponseScenarioValidator(config.nussknackerInstanceName))
   )
 
   override def name: String = "lite-k8s"
--- a/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
+++ b/engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
@@ -1,0 +1,62 @@
+package pl.touk.nussknacker.k8s.manager
+
+import akka.actor.ActorSystem
+import com.typesafe.config.Config
+import pl.touk.nussknacker.engine.ModelData.BaseModelDataExt
+import pl.touk.nussknacker.engine.api.component.AdditionalPropertyConfig
+import pl.touk.nussknacker.engine.api.deployment.{DeploymentManager, ProcessingTypeDeploymentService}
+import pl.touk.nussknacker.engine.api.process.ProcessName
+import pl.touk.nussknacker.engine.api.{LiteStreamMetaData, RequestResponseMetaData}
+import pl.touk.nussknacker.engine.requestresponse.api.openapi.RequestResponseOpenApiSettings
+import pl.touk.nussknacker.engine.{BaseModelData, CustomProcessValidator, DeploymentManagerProvider, TypeSpecificInitialData}
+import pl.touk.nussknacker.k8s.manager.RequestResponseSlugUtils.defaultSlug
+import sttp.client.{NothingT, SttpBackend}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/*
+  Each scenario is deployed as Deployment+ConfigMap
+  ConfigMap contains: model config with overrides for execution and scenario
+ */
+class K8sDeploymentManagerProvider extends DeploymentManagerProvider {
+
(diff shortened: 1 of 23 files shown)


engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  override def additionalPropertiesConfig(config: Config): Map[String, AdditionalPropertyConfig] = forMode(config)(
+    _ => Map.empty,
+    _ => RequestResponseOpenApiSettings.additionalPropertiesConfig
+  )
+
+  override def additionalValidators(config: Config): List[CustomProcessValidator] = forMode(config)(
+    _ => Nil,
+    config => List(new LiteScenarioValidator(config.nussknackerInstanceName))
```
arkadius, line 47: Do you see some problems in always returning this validator? I would like to keep `K8sDeploymentManager` as dynamic as possible - to distinguish between rr and streaming at last possible moment - when scenario is passed

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/LiteScenarioValidator.scala
```diff
@@ -20,12 +23,13 @@ class LiteScenarioValidator(nussknackerInstanceName: Option[String]) {
...
   }
 
-  private[manager] def validateRequestResponse(scenarioName: ProcessName, rrMetaData: RequestResponseMetaData): Validated[Throwable, Unit] = {
+  private[manager] def validateRequestResponse(scenarioName: ProcessName, rrMetaData: RequestResponseMetaData): ValidatedNel[ProcessCompilationError, Unit] = {
     val slug = RequestResponseSlugUtils.determineSlug(scenarioName, rrMetaData, nussknackerInstanceName)
     // We don't sanitize / validate against url because k8s object names are more restrictively validated than urls, see https://datatracker.ietf.org/doc/html/rfc3986
     val withoutSanitization = ServicePreparer.serviceNameWithoutSanitization(nussknackerInstanceName, slug)
     val withSanitization = ServicePreparer.serviceName(nussknackerInstanceName, slug)
-    Validated.cond(withSanitization == withoutSanitization, (), IllegalRequestResponseSlug(nussknackerInstanceName, slug))
+    Validated.cond(withSanitization == withoutSanitization, (), NonEmptyList.of(SpecificDataValidationError("slug", "Allowed characters include lowercase letters, digits, hyphen, " +
```
arkadius, line 31: What about name length restriction?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
arkadius, line 47: What about validatePreparedService? IMO  we should run both validations:
- slug acceptable characters and length restrictions
- slug uniqueness
at once dynamically during validation of scenario.

mproch: We check metadata type anyway, so this `forMode` is in fact not really needed. I think we have bigger problem of this kind with additionalPropertiesConfig and typeSpecificInitialData already...

mproch: I would rather validate uniqueness only during deploy. We need access to K8s then, and it's more along the lines "transient errors (like duplication)" are done during deploy...

arkadius: On the other hand it is very easy to duplicate slug - let's say someone do export -> import of scenario. We validate scenario name, things with slug looks the same for me. Maybe we could do validation only on designer side by some query on DB :thinking: or some short living cache for this info from k8s. WDYT?

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/LiteScenarioValidator.scala
mproch: :+1: added

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
mproch: We could... but I'd leave it for separate PR not to block the API

arkadius: ok

arkadius (approved):

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManagerProvider.scala
mproch: replaced generic validator with RequestResponseScenarioValidator

PR #3584, opened 2022-09-30
RR authorization

dswiecki:


engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/RequestResponseConfig.scala
```diff
@@ -13,4 +13,7 @@ object OpenApiDefinitionConfig {
 
 }
 
-case class RequestResponseConfig(definitionMetadata: OpenApiDefinitionConfig = OpenApiDefinitionConfig())
+case class BasicAuthConfig(user: String = "publisher", password: String)
```
mproch, line 16: Don't think publisher is a good name for req-res :) Let's leave it blank

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/ScenarioRoute.scala
```diff
@@ -15,14 +14,34 @@ import pl.touk.nussknacker.engine.api.exception.NuExceptionInfo
...
   // We use this relative reference instead of default '/' because runtime can by server reverse proxies (i.e. ingress controller) that can rewrite this url
   private val defaultServerUrl = "./"
 
+  val securityDirectiveOpt: Option[AuthenticationDirective[String]] = {
+
+    val password = config.basicAuthConfig.map(_.password)
+
+    def rrAuthenticator(credentials: Credentials): Option[String] =
+      credentials match {
+        case p@Credentials.Provided(id) if id == "publisher" && p.verify(password.get) => Some(id)
```
mproch, line 32: Let's check name properly ;)

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/RequestResponseConfig.scala
dswiecki: changed

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/ScenarioRoute.scala
dswiecki: changed

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/ScenarioRoute.scala
```diff
@@ -15,14 +14,35 @@ import pl.touk.nussknacker.engine.api.exception.NuExceptionInfo
...
   // Regarding https://spec.openapis.org/oas/v3.1.0#serverObject server url accept urls "relative to the location where the OpenAPI document is being served"
   // We use this relative reference instead of default '/' because runtime can by server reverse proxies (i.e. ingress controller) that can rewrite this url
   private val defaultServerUrl = "./"
 
+  val securityDirectiveOpt: Option[AuthenticationDirective[String]] = {
+
+    val password = config.basicAuthConfig.map(_.password)
+    val user = config.basicAuthConfig.map(_.user)
+
+    def rrAuthenticator(credentials: Credentials): Option[String] =
```
mproch, line 31: this _.get doesn't look too nice... I know they won't fail, but I looks a bit dangerous. 
Let's extract method:
```
prepareAuthenticator(basicAuthConfig): Credentials => Option[String]
```
and invoke it in line 39 (and then you won't have `flatMap { _ ` which also looks a bit suspicious

docs/installation_configuration_guide/DeploymentManagerConfiguration.md
```diff
@@ -197,13 +197,15 @@ Just like in [Designer installation](./Installation.md#Basic environment variabl
...
+| http.port                                                 | int    | 8181            | Port on which REST API of scenarios will be exposed                                                                              | 
+| request-response.definitionMetadata.servers               | string | [{"url": "./"}] | Configuration of exposed servers in scenario's OpenApi definition. When not configured, will be used server with ./ relative url | 
+| request-response.definitionMetadata.servers[].url         | string |                 | Url of server in scenario's OpenApi definition                                                                                   | 
+| request-response.definitionMetadata.servers[].description | string |                 | (Optional) description of server in scenario's OpenApi definition                                                                | 
+| request-response.basicAuthConfig.user                     | string |                 | (Optional) Basic auth user                                                                                                       |
```
gadomsky, line 207: What do you think about nesting in under `security` (so it will be `request-response.security.basicAuth.user`
), therefore when we introduce another type of auth in the future, we will not have to change basic-auth api?

dswiecki: done

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/ScenarioRoute.scala
dswiecki: refactored

mproch (approved):
:+1: I wonder if it would be better to mention somewhere that changes to config will require redeploy of scenarios? (or do we already have it?)

PR #3602, opened 2022-10-06
Fix testing from file with DisplayJsonWithEncoder

mproch:

### DIFF
--- a/engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSpec.scala
+++ b/engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSpec.scala
@@ -1,19 +1,24 @@
 package pl.touk.nussknacker.engine.kafka.consumerrecord
 
+import com.typesafe.config.ConfigFactory
 import org.apache.flink.api.common.typeinfo.TypeInformation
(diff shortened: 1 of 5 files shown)


engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSpec.scala
```diff
@@ -23,8 +28,10 @@ class ConsumerRecordDeserializationSpec extends AnyFunSuite with Matchers with K
     val typeInformation: TypeInformation[TestConsumerRecord] = TypeInformation.of(classOf[TestConsumerRecord])
 
     intercept[Exception] {
+      Serializers.registerSerializers(emptyModel, executionConfigWithoutKryo)
```
gadomsky, line 31: this invocation could be moved outside of `intercept`? so we are 100% sure that exception is thrown by `getSerializeRoundTrip`

gadomsky (approved):
LGTM

engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/consumerrecord/ConsumerRecordDeserializationSpec.scala
mproch: :+1:

PR #3610, opened 2022-10-10
Cleanup: remove deprecations

dswiecki:


mproch (approved):
Let's add MigrationGuid entry

PR #3621, opened 2022-10-13
Fix fragment output

mproch:

### DIFF
--- a/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/SubprocessResolverSpec.scala
+++ b/interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/SubprocessResolverSpec.scala
@@ -243,14 +243,12 @@ class SubprocessResolverSpec extends AnyFunSuite with Matchers with Inside{
       .source("source", "source1")
       .subprocessEnd("sub", "subProcess1", "ala" -> "'makota'")
 
-    val subprocess = CanonicalProcess(MetaData("subProcess1", FragmentSpecificData()),
-      List(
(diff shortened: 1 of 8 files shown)


interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/SubprocessResolverSpec.scala
```diff
@@ -297,6 +295,26 @@ class SubprocessResolverSpec extends AnyFunSuite with Matchers with Inside{
     }
   }
 
+  test("handle subprocess with empty outputs") {
+    val fragment = ScenarioBuilder
+      .fragment("fragment1")
```
arkadius, line 300: It looks a little bit different than in scenarios. In scenarios each node is visible as a separate builder invocation, so there is .source. Here .fragment contains this input node inside invocation. Can it be consistent with scenarios or there are some troubles with that?

arkadius (approved):

interpreter/src/test/scala/pl/touk/nussknacker/engine/compile/SubprocessResolverSpec.scala
mproch: no, it can be done with some code ;) . For fragments the separation just doesn't add any value. I can change it in next PR if you like consistency better

arkadius: It's fine

PR #3640, opened 2022-10-20
Remove output from InterpretationResult

mproch:

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
@@ -92,13 +92,14 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
       case EndingProcessor(id, ref, false) =>
         listeners.foreach(_.endEncountered(id, ref.id, ctx, metaData))
         invoke(ref, ctx).map {
-          case Left(ValueWithContext(output, newCtx)) =>
-            List(Left(InterpretationResult(EndReference(id), output, newCtx)))
+          //for processor Result should be null
+          case Left(ValueWithContext(_, newCtx)) =>
+            List(Left(InterpretationResult(EndReference(id), newCtx)))
           case r@Right(exInfo) => List(Right(exInfo))
         }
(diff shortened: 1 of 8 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -92,13 +92,14 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
...
         invoke(ref, ctx).map {
-          case Left(ValueWithContext(output, newCtx)) =>
-            List(Left(InterpretationResult(EndReference(id), output, newCtx)))
+          //for processor Result should be null
+          case Left(ValueWithContext(_, newCtx)) =>
+            List(Left(InterpretationResult(EndReference(id), newCtx)))
           case r@Right(exInfo) => List(Right(exInfo))
         }
       case EndingProcessor(id, _, true) =>
         //FIXME: null??
```
gadomsky, line 101: this FIXME is still needed?

gadomsky (approved):

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
mproch: :+1: true

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -92,13 +92,14 @@ private class InterpreterInternal[F[_]](listeners: Seq[ProcessListener],
       case EndingProcessor(id, ref, false) =>
         listeners.foreach(_.endEncountered(id, ref.id, ctx, metaData))
         invoke(ref, ctx).map {
-          case Left(ValueWithContext(output, newCtx)) =>
-            List(Left(InterpretationResult(EndReference(id), output, newCtx)))
+          //for processor Result should be null
```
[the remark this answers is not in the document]
mproch: It turns out its null/BoxedUnit/Void etc - changed comment

PR #3645, opened 2022-10-20
Fix TypingResultAware typing for fixed values

mproch:

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/TypingResultAwareTypeInformationDetection.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/TypingResultAwareTypeInformationDetection.scala
@@ -59,16 +59,8 @@ class TypingResultAwareTypeInformationDetection(customisation:
     (typingResult match {
       case a if additionalTypeInfoDeterminer.isDefinedAt(a) =>
         additionalTypeInfoDeterminer.apply(a)
-      case a: TypedTaggedValue => forType(a.underlying)
-      case a: TypedDict => forType(a.objType)
-      case a: TypedClass if a.params.isEmpty =>
-        //TODO: scala case classes are not handled nicely here... CaseClassTypeInfo is created only via macro, here Kryo is used
-        registeredTypeInfos.find(_.getTypeClass == a.klass).getOrElse(TypeInformation.of(a.klass))
-
-      case a: TypedClass if a.klass == classOf[java.util.List[_]] => new ListTypeInfo[AnyRef](forType(a.params.head))
-
-      case a: TypedClass if a.klass == classOf[java.util.Map[_, _]] => new MapTypeInfo[AnyRef, AnyRef](forType(a.params.head), forType(a.params.last))
-
+      case a: TypedClass if a.klass == classOf[java.util.List[_]] && a.params.size == 1 => new ListTypeInfo[AnyRef](forType(a.params.head))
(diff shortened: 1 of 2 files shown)


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/TypingResultAwareTypeInformationDetection.scala
```diff
@@ -59,23 +59,19 @@ class TypingResultAwareTypeInformationDetection(customisation:
...
-      case a: TypedClass if a.params.isEmpty =>
-        //TODO: scala case classes are not handled nicely here... CaseClassTypeInfo is created only via macro, here Kryo is used
-        registeredTypeInfos.find(_.getTypeClass == a.klass).getOrElse(TypeInformation.of(a.klass))
-
-      case a: TypedClass if a.klass == classOf[java.util.List[_]] => new ListTypeInfo[AnyRef](forType(a.params.head))
-
-      case a: TypedClass if a.klass == classOf[java.util.Map[_, _]] => new MapTypeInfo[AnyRef, AnyRef](forType(a.params.head), forType(a.params.last))
-
+      case a: TypedClass if a.klass == classOf[java.util.List[_]] && a.params.size == 1 => new ListTypeInfo[AnyRef](forType(a.params.head))
+      case a: TypedClass if a.klass == classOf[java.util.Map[_, _]] && a.params.size == 2 => new MapTypeInfo[AnyRef, AnyRef](forType(a.params.head), forType(a.params.last))
```
arkadius, line 63: I thought that thanks to those line:
>         genericTypeClass(runtimeClass, typ.dealias.typeArgs.map(fromType))
and
>           TypedClass(klass, klass.getTypeParameters.map(_ => Unknown).toList)
in typing.scala it is rather imposssible to have other number of type parameters than declared in class. The only situation that I see if someone would invoke very low level version `Typed.genericTypeClass` and by mistake pass invalid number of type parameters. Those checks are only a form of defensive programming?

mproch: yes, I don't expect them to be needed, more defensive programming/sanity check

arkadius: :+1:

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/TypingResultAwareTypeInformationDetection.scala
```diff
@@ -59,23 +59,19 @@ class TypingResultAwareTypeInformationDetection(customisation:
...
         TypedScalaMapTypeInformation(a.fields.mapValuesNow(forType))
       case a: TypedObjectTypingResult if a.objType.klass == classOf[TypedMap] =>
         TypedMapTypeInformation(a.fields.mapValuesNow(forType))
       //TODO: better handle specific map implementations - other than HashMap?
       case a: TypedObjectTypingResult if classOf[java.util.Map[String, _]].isAssignableFrom(a.objType.klass) =>
         TypedJavaMapTypeInformation(a.fields.mapValuesNow(forType))
+      case a: SingleTypingResult if a.objType.params.isEmpty =>
+        val klass = a.objType.klass
+        //TODO: scala case classes are not handled nicely here... CaseClassTypeInfo is created only via macro, here Kryo is used
+        registeredTypeInfos.find(_.getTypeClass == klass).getOrElse(TypeInformation.of(klass))
```
arkadius, line 74: maybe we should do this find in case expression so for else will work fallback as below?

arkadius (approved):
I added one thing to consider

PR #3657, opened 2022-10-26
Fix schema validation when using additionalProperties without required

Filemon279:
```
{
  "properties": {
    "msisdn": {
      "type": "string"
    }
  },
  "additionalProperties": false,
} 
```
Before fix above example didn't allowed to pass request as `{"msisdn": "test"}` - It was required to add `"required": ["msisdn"]`.

### DIFF
--- a/engine/lite/request-response/app/src/test/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpJsonSchemaSpec.scala
+++ b/engine/lite/request-response/app/src/test/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpJsonSchemaSpec.scala
@@ -96,6 +96,24 @@ class RequestResponseHttpJsonSchemaSpec extends RequestResponseHttpTest {
     }
   }
 
+  it should "should handle additionalProperties without required" in {
+    assertProcessNotRunning(procId)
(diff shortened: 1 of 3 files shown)


engine/lite/request-response/app/src/test/scala/pl/touk/nussknacker/engine/requestresponse/http/RequestResponseHttpJsonSchemaSpec.scala
```diff
@@ -96,6 +96,24 @@ class RequestResponseHttpJsonSchemaSpec extends RequestResponseHttpTest {
     }
   }
 
+  it should "should handle additionalProperties without required" in {
```
lciolecki, line 99: You don't have to do full request to test determiner. I think we should write unit test only for JsonSchemaSubclassDeterminer.

Filemon279: That's a great idea, we should do it in the next MR :)

dswiecki (approved):

dswiecki:
maybe also add some changelog entry about this bug fix

PR #3668, opened 2022-10-28
TestScenarioRunner.requestResponseBased() api enhancements: returning scenario compilation errors as a ValidatedNel

arkadius:

### DIFF
--- a/engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/LiteRequestResponseFunctionalTest.scala
+++ b/engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/LiteRequestResponseFunctionalTest.scala
@@ -1,9 +1,13 @@
 package pl.touk.nussknacker.engine.lite.components.requestresponse
 
 import akka.http.scaladsl.model.{HttpMethods, HttpRequest}
+import cats.data.NonEmptyList
+import cats.data.Validated.Invalid
 import io.circe.syntax.EncoderOps
+import org.scalatest.Inside.inside
 import org.scalatest.funsuite.AnyFunSuite
 import org.scalatest.matchers.should.Matchers
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.CustomNodeError
 import pl.touk.nussknacker.engine.build.ScenarioBuilder
 import pl.touk.nussknacker.engine.lite.util.test.RequestResponseTestScenarioRunner
 import pl.touk.nussknacker.engine.lite.util.test.RequestResponseTestScenarioRunner._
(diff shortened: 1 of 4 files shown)


gadomsky (approved):

engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/LiteRequestResponseFunctionalTest.scala
```diff
@@ -23,16 +28,23 @@ class LiteRequestResponseFunctionalTest extends AnyFunSuite with Matchers {
...
       .requestResponse("test")
       .additionalFields(properties = sampleSchemas)
       .source("input", "request")
-      .emptySink("output", "response", "Raw editor" -> "true", "Value" -> """{field1: #input.field1, invalid: 1}""")
+      .emptySink("output", "response", "Raw editor" -> "true", "Value" -> outputValueExpression)
 
-    val thrown = the[AssertionError] thrownBy runner.runWithRequests(scenario) { invoker =>
-      invoker(HttpRequest(HttpMethods.POST, entity = Map("field1" -> "value").asJson.spaces2))
+    val redundantField = """invalid"""
+    inside(runner.runWithRequests(scenario(s"""{field1: #input.field1, $redundantField: 1}""")) { _ => }) {
```
dswiecki, line 38: for me this test is less readable than previous one :( can `{ _ => }` be the default ?

arkadius: But you know that this code in closure (```invoker =>
      invoker(HttpRequest(HttpMethods.POST, entity = Map("field1" -> "value").asJson.spaces2)}```) was a dead code (never invoked)?

I don't think that  making ```{ _ => }``` the default is a good idea. It will be easy to forgot passing block of checks, but I don't have any better idea to make it more readable. Maybe instead of Invoker => Unit, it hould be ValidationNel[Invoker] => Unit? It won't be tricky?

dswiecki: Only now I see passing block of checks is via `run` function. I think we should
- keep consistent API - checks should be done on the result of `runner.runWithRequests` just like in flink and lite - then we'll use valid/invalidValue
- regarding default value - it won't be the case since checks will be separated

dswiecki: no strange `)) { _ => }) {`, no inside and PF, just val result = runner.runWithRequests

arkadius: But the idea of original author (@mproch ) was to be able to do many invocations on running scenario. Runner  have lifecycle so to be sure that close() will be invoked at the end, it must be some kind of  closure passed to method. Regarding `inside` and `PF` it is normal way to do verfiications. I want to check some elements of returned value, I don't want to exactly check message (to not make test cocreted) and because of that I've done it. I don't want to force someone to change thousand of tests after some slight change in message formatting - like adding colon, dot or space somewhere.

dswiecki (changes requested):
we should unify RR api

dswiecki (approved):
lgtm 👍

PR #3669, opened 2022-10-31
Fix contextId assignment for the output of ForEachTransformer

gskrobisz:
Fix for incorrect contextId assignment in `ForEachTransformer`.

Environment: 1.7.0-staging-2022-10-27-10064, e.g. https://staging.nussknacker.io/visualization/fmc_c1
Steps to reproduce: 
1. run "Test from file" on scenario with for-each component
2. see the values assigned to the output of for-each
3. on "Test case" list it is impossible to switch between individual cases within the same input collection, because all elements have the same Test case id = contenxtId (see attached image)

The "lite" version ForEachTransformer assigns contextId correctly.

obraz

### DIFF
--- a/engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
+++ b/engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
@@ -23,8 +23,10 @@ object ForEachTransformer extends CustomStreamTransformer {
       stream
         .flatMap(ctx.lazyParameterHelper.lazyMapFunction(elements))
         .flatMap((valueWithContext: ValueWithContext[util.Collection[AnyRef]], c: Collector[ValueWithContext[AnyRef]]) => {
-          valueWithContext.value.asScala
-            .map(new ValueWithContext[AnyRef](_, valueWithContext.context))
+          valueWithContext.value.asScala.toList.zipWithIndex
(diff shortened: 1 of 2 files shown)


engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
```diff
@@ -23,8 +23,10 @@ object ForEachTransformer extends CustomStreamTransformer {
       stream
         .flatMap(ctx.lazyParameterHelper.lazyMapFunction(elements))
         .flatMap((valueWithContext: ValueWithContext[util.Collection[AnyRef]], c: Collector[ValueWithContext[AnyRef]]) => {
-          valueWithContext.value.asScala
-            .map(new ValueWithContext[AnyRef](_, valueWithContext.context))
+          valueWithContext.value.asScala.toList.zipWithIndex
```
mproch, line 26: is toList needed?

mproch (commented):
Added small comment + can you add test for that?

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
gskrobisz: Fixed.

dswiecki (approved):

PR #3671, opened 2022-11-02
Fix: do not show extra scrollbar on scenario screen when panel to lar…

dswiecki:
…ge (#3642)


arkadius (approved):
Please add changelog entry

PR #3672, opened 2022-11-02
Fix/for each transformer context id for element 1.6

gskrobisz:


arkadius (approved):
Please add changelog entry

PR #3695, opened 2022-11-09
Fix delaying JSON records by field in universal source

jedrz:


engine/flink/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/UniversalTimestampFieldAssigner.scala
```diff
@@ -3,15 +3,17 @@ package pl.touk.nussknacker.engine.schemedkafka.source.flink
...
   def apply[K, V](fieldName: String): SerializableTimestampAssigner[ConsumerRecord[K, V]] = new SerializableTimestampAssigner[ConsumerRecord[K, V]] {
-    // TODO: Handle exceptions thrown within sources (now the whole process fails)
     override def extractTimestamp(element: ConsumerRecord[K, V], recordTimestamp: Long): Long =
-      Option(element.value().asInstanceOf[GenericRecord].get(fieldName))
-        .map(_.asInstanceOf[Long])
-        .getOrElse(0L) // explicit null to 0L conversion (instead of implicit unboxing)
+      element.value() match {
+        case null => 0
+        case genericRecord: GenericRecord => genericRecord.get(fieldName).asInstanceOf[Long]
+        case typedMap: TypedMap => typedMap.get(fieldName).asInstanceOf[Long]
```
gadomsky, line 15: maybe also a case for `_` when schema is not of type `object` ?
e.g.
```
{
    "type": "string"
}
```

mproch: also, maybe let's  `case map: java.util.Map`?
(probably we should just have expression instead of name, since we don't handle nested fields etc, but let's leave it for now...)

gadomsky (approved):
LGTM, 
maybe we can also add position in changelog?

PR #3696, opened 2022-11-09
Fix: json schema deserialization part I

lciolecki:


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBuilder.scala
```diff
@@ -1,27 +0,0 @@
...
-  def parseSchema(rawJsonSchema: String): Schema =
-    parseSchema(rawJsonSchema, useDefaults = true, nullableSupport = true)
-
-  def parseSchema(rawJsonSchema: String, useDefaults: Boolean, nullableSupport: Boolean): Schema = {
-    val rawSchema: JSONObject = new JSONObject(rawJsonSchema)
-
-    SchemaLoader
-      .builder()
-      .useDefaults(useDefaults)
-      .nullableSupport(nullableSupport)
```
gadomsky, line 18: @lciolecki  maybe, until we figure out what to do with this PR, we can just delete this line? 

I think it would fix most of our knonw issues with RR, if not all of them.

PR #3707, opened 2022-11-17
Full OpenAPI apiKey support

mproch:


components/openapi/src/main/scala/pl/touk/nussknacker/openapi/SwaggerSecurity.scala
```diff
@@ -18,4 +18,12 @@ sealed trait SecurityInHeader extends SwaggerSecurity {
 
 case class ApiKeyInHeader(name: String, key: String) extends SecurityInHeader {
   def value: String = key
+}
```
gadomsky, line 21: do we need `SecurityInHeader` trait? Is it possible that, there will be different implementations?

mproch: Authorization header with Bearer token will be such case I think

gadomsky (approved):
LGTM

PR #3709, opened 2022-11-17
support for typing Maps using 'aditionalProperties' in JsonSchema

gadomsky:


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
```diff
@@ -59,7 +59,7 @@ object JsonToNuStruct {
         extract[JsonNumber](_.asNumber, _.toBigDecimal.map(_.bigDecimal).orNull)
       case SwaggerArray(elementType) =>
         extract[Vector[Json]](_.asArray, _.zipWithIndex.map { case (el, idx) => JsonToNuStruct(el, elementType, s"$path[$idx]") }.asJava)
-      case SwaggerObject(elementType) =>
+      case SwaggerObject(elementType, _) => //todo
```
mproch, line 62: Guess we'll also need to handle it in encoding?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
```diff
@@ -59,8 +59,8 @@ object JsonToNuStruct {
         extract[JsonNumber](_.asNumber, _.toBigDecimal.map(_.bigDecimal).orNull)
       case SwaggerArray(elementType) =>
         extract[Vector[Json]](_.asArray, _.zipWithIndex.map { case (el, idx) => JsonToNuStruct(el, elementType, s"$path[$idx]") }.asJava)
```
lciolecki, line 61: Above, exactly at 32 line you have to change the conditions:
```
case (key, value) if elementType.contains(key) 
```
Because right now struct will cut additional fields..

gadomsky: it is expected acctualy, please see https://github.com/TouK/nussknacker/pull/3709/files#diff-bc73474afc18bd51d0dd81754c63dd2b308b8e457ed1d5bac55a98d11b834777R127

we handle `addionalFields` only when used without `properties` -> and in that case such schema is typed to Map, not Object.

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/json/EitherCodecs.scala
```diff
@@ -0,0 +1,18 @@
+package pl.touk.nussknacker.engine.util.json
+
+import io.circe._
+
+trait EitherCodecs {
+  implicit def eitherDecoder[A, B](implicit a: Decoder[A], b: Decoder[B]): Decoder[Either[A, B]] = {
+    val left: Decoder[Either[A, B]] = a.map(Left.apply)
+    val right: Decoder[Either[A, B]] = b.map(Right.apply)
+    left or right
```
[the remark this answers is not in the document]
gadomsky: nice catch,

I will make issue in GH and resolve this, as it does not concern this PR now right now.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
lciolecki: Right, great! :)

mproch (approved):
Cool :)

PR #3715, opened 2022-11-18
Refactor of the R-R docs - multiple responses

ZbyszekMM:
The original motivation behind this PR was to describe how to handle multiple scenario branches in R-R processing mode. I ended up adding considerations how scenarios are processed in R-R mode and changed description of the Collect node.

### DIFF
--- a/docs/integration/KafkaIntegration.md
+++ b/docs/integration/KafkaIntegration.md
@@ -2,12 +2,12 @@
 sidebar_position: 1
 ---
 
-# Kafka
+# Kafka | Streaming only
 
 ## Prerequisites
 
 To fully understand how Nussknacker works with Kafka topics, it's best to read the following first:
-* [Role of Schema Registry](/about/TypicalImplementation)
+* [Role of Schema Registry](/about/TypicalImplementation/)
 * [Confluent Schema Registry](https://docs.confluent.io/platform/current/schema-registry/index.html)
 * [Kafka topic key](https://kafka.apache.org/intro)
 
@@ -22,7 +22,7 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
 
 Nussknacker integrates with a schema registry. It is the source of topics available to choose from in sources and sinks. It also allows Nussknacker to provide syntax suggestions and validation. Nussknacker assumes that for the topic `topic-name` a schema `topic-name-value` and optionally `topic-name-key` (for the Kafka topic key) will be defined in the schema registry.
 
-Schemas are stored and managed by Confluent Schema Registry; it is [bundled with Nussknacker](/about/TypicalImplementation) in all deployment versions. Schemas can be registered in Schema Registry by means of REST API based CLI or using AKHQ, an open source GUI for Apache Kafka and Confluent Schema Registry. AKHQ is bundled with Nussknacker in all deployment versions.
+Schemas are stored and managed by Confluent Schema Registry; it is [bundled with Nussknacker](/about/TypicalImplementation/) in all deployment versions. Schemas can be registered in Schema Registry by means of REST API based CLI or using AKHQ, an open source GUI for Apache Kafka and Confluent Schema Registry. AKHQ is bundled with Nussknacker in all deployment versions.
 
 Nussknacker supports both JSON and AVRO schemas, and JSON and AVRO topic payloads. Detailed information on how AVRO data should be serialized/deserialized can be found in [Confluent Wire Documentation](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format).
 
--- a/docs/scenarios_authoring/AggregatesInTimeWindows.md
+++ b/docs/scenarios_authoring/AggregatesInTimeWindows.md
@@ -2,7 +2,7 @@
 sidebar_position: 5
 ---
 
-# Aggregates in time windows
+# Aggregates in time windows | Streaming only
 
 **(Flink engine only)**
 
--- a/docs/scenarios_authoring/RRDataSourcesAndSinks.md
+++ b/docs/scenarios_authoring/RRDataSourcesAndSinks.md
@@ -2,42 +2,73 @@
 sidebar_position: 4
 ---
 
-# Request-Response sources and sinks
+# Request-Response specific components
 
-## Http request as input and output to Nussknacker scenarios
+## Concepts
 
-### Concepts
-
-In Request-Response mode you interact with Nussknacker Lite runtime engine in a similar way as with any "server" in the
+### General
+In Request-Response mode you interact with Nussknacker Lite runtime engine in the same way as with any "server" in the
 Internet. You send request which contains request parameters and you get response from the server. The protocol for this
-communication is HTTP - the same protocol is used by browsers when they communicate with web servers. One of the
-differences is that the content of the communication with the Nussknacker Lite engine ("the server") is encoded as JSON,
-not HTML. Not very surprisingly, in Request-Response processing mode the only possible Source component is Request and
-the only Sink component is Response.
+communication is HTTP - the same protocol is used by browsers when they communicate with web servers. When the  browser is used to browse internet pages, the web server responds with content encoded as HTML. In Nussknacker case, the Nussknacker Lite engine ("the server") respondes with content encoded as JSON. You may envoke Nussknacker Lite engine directly from the browser and watch the JSON response in the browser window. 
 
-## Request-Response schema
+Not very surprisingly, in the Request-Response processing mode the only possible source component is request and the only sink component is response.
 
-Each scenario needs to define two schemas in [JSON Schema](https://json-schema.org/) format:
+### Schemas
+Unlike in the streaming processing mode, there is no Schema Registry which store data schemas - request and response schemas in our case. Still, the data schema of the request and response ("input" and "output") need to be declared to Nussknacker; they are defined alongside the scenario (as the scenario property). The "language" used to define schemas is called [JSON Schema](https://json-schema.org/). Nussknacker uses those schemas to understand how the request and response should look like, assist you with hints and validate expressions you write. 
 
-- Input schema
-- Output schema
+### Is Nussknacker request response scenario a decision tree?
+The way in which the scenario diagram should be interpreted in the Request - Response processing mode requires some attention. The request - response scenario diagram is not a [decision tree](https://en.wikipedia.org/wiki/Decision_tree). During a decision tree "processing" one and only one "branch" of a tree is "activated"; "processing" ends up in one of the end leaves. As in a decision tree each leaf represents a decision (or result) - one decision (result) is generated.
 
-Nussknacker uses those schema to detect how the request and response should look like.
+In Request-Response processing mode, the diagram is more like a flow chart. The data record processed by the scenario (#input at the very start of the processing) "flows" through the scenario. If there are [splits](./BasicNodes.md#split), the data records start to "flow" in parallel through many branches. If there are [for-each](./BasicNodes.md#foreach) nodes, multiple records are produced as the result. Because the Request-Response scenario can return only one response, a question arises what exactly will be returned in such cases and how to - if neccessary - return all data records produced in a scenario. 
+
+### Scenario response in scenarios with split and for-each nodes
+If parallel branches end with response nodes and more than one response is generated (e.g there is no filtering after split), the only response which is returned is the one which was chronologically generated first. If no response is generated (e.g. after filtering), an error is returned (in the future this behaviour may change).
+
+If you want to to return data from multiple parallel branches, use [union](./BasicNodes.md#union) node to merge branches and [collect](./BasicNodes.md#collect) node to collect data records into a list. Your scenario should have only one response node.
+
+If you use for-each node and need to collect all the results into a list of values, use [collect](#collect) node.
+
+## Components
+### Common configuration - Request-Response schema
+
+See [schemas](#schemas) subsection of the Concepts section for the rationale behind schemas. 
+
+Click Properties icon on the right panel to define input and output schemas.
 
 ![RR schema](img/rrProperties.png "RR properties")
 
-## Request-Response source
+### Request-Response source
 
-Thanks to the [Input schema](#request-response-schema) Nussknacker knows what data to expect in the request, so it will
-hint field names and will perform additional validations on SpEL expressions involving data from the selected topic.
+The request node is very simple - it does not require any configuration. Just drop it on the diagram and proceed with authoring the scenario.
 
-![RR source](img/rrSource.png "RR source")
+### Request-Response sink
 
-## Request-Response sink
-
-The `response` sink configuration form will show a list of fields defined in [Input schema](#request-response-schema).
+The `response` sink configuration form will show a list of fields defined in the [Output schema](#common-configuration---request-response-schema).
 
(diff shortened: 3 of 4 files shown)


docs/integration/KafkaIntegration.md
```diff
@@ -2,12 +2,12 @@
 sidebar_position: 1
 ---
 
-# Kafka
+# Kafka | Streaming only
```
dswiecki, line 5: I'm not sure wheather we can have this note here, node docs links may stop working

docs/integration/KafkaIntegration.md
```diff
@@ -2,12 +2,12 @@
...
 
 ## Prerequisites
 
 To fully understand how Nussknacker works with Kafka topics, it's best to read the following first:
-* [Role of Schema Registry](/about/TypicalImplementation)
+* [Role of Schema Registry](/about/TypicalImplementation/)
```
mproch, line 10: should be https://docs.nussknacker.io/documentation/about/TypicalImplementationStreaming

docs/integration/KafkaIntegration.md
```diff
@@ -22,7 +22,7 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
 
 Nussknacker integrates with a schema registry. It is the source of topics available to choose from in sources and sinks. It also allows Nussknacker to provide syntax suggestions and validation. Nussknacker assumes that for the topic `topic-name` a schema `topic-name-value` and optionally `topic-name-key` (for the Kafka topic key) will be defined in the schema registry.
 
-Schemas are stored and managed by Confluent Schema Registry; it is [bundled with Nussknacker](/about/TypicalImplementation) in all deployment versions. Schemas can be registered in Schema Registry by means of REST API based CLI or using AKHQ, an open source GUI for Apache Kafka and Confluent Schema Registry. AKHQ is bundled with Nussknacker in all deployment versions.
+Schemas are stored and managed by Confluent Schema Registry; it is [bundled with Nussknacker](/about/TypicalImplementation/) in all deployment versions. Schemas can be registered in Schema Registry by means of REST API based CLI or using AKHQ, an open source GUI for Apache Kafka and Confluent Schema Registry. AKHQ is bundled with Nussknacker in all deployment versions.
```
mproch, line 25: https://docs.nussknacker.io/documentation/about/TypicalImplementationStreaming

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
-In Request-Response mode you interact with Nussknacker Lite runtime engine in a similar way as with any "server" in the
+### General
+In Request-Response mode you interact with Nussknacker Lite runtime engine in the same way as with any "server" in the
 Internet. You send request which contains request parameters and you get response from the server. The protocol for this
-communication is HTTP - the same protocol is used by browsers when they communicate with web servers. One of the
-differences is that the content of the communication with the Nussknacker Lite engine ("the server") is encoded as JSON,
-not HTML. Not very surprisingly, in Request-Response processing mode the only possible Source component is Request and
-the only Sink component is Response.
+communication is HTTP - the same protocol is used by browsers when they communicate with web servers. When the  browser is used to browse internet pages, the web server responds with content encoded as HTML. In Nussknacker case, the Nussknacker Lite engine ("the server") respondes with content encoded as JSON. You may envoke Nussknacker Lite engine directly from the browser and watch the JSON response in the browser window.
```
mproch, line 12: Currently our requests are POSTs mostly, so I don't think it's so easy without custom plugins...
The idea of web server returning content as HTML is also rather outdated, since more and more pages are invoking REST APIs similar to Nussknacker's :)

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
 
-## Request-Response schema
+Not very surprisingly, in the Request-Response processing mode the only possible source component is request and the only sink component is response.
 
-Each scenario needs to define two schemas in [JSON Schema](https://json-schema.org/) format:
+### Schemas
+Unlike in the streaming processing mode, there is no Schema Registry which store data schemas - request and response schemas in our case. Still, the data schema of the request and response ("input" and "output") need to be declared to Nussknacker; they are defined alongside the scenario (as the scenario property). The "language" used to define schemas is called [JSON Schema](https://json-schema.org/). Nussknacker uses those schemas to understand how the request and response should look like, assist you with hints and validate expressions you write.
```
mproch, line 17: Will it be understandable for those who didn't read streaming description?

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
+### Schemas
+Unlike in the streaming processing mode, there is no Schema Registry which store data schemas - request and response schemas in our case. Still, the data schema of the request and response ("input" and "output") need to be declared to Nussknacker; they are defined alongside the scenario (as the scenario property). The "language" used to define schemas is called [JSON Schema](https://json-schema.org/). Nussknacker uses those schemas to understand how the request and response should look like, assist you with hints and validate expressions you write. 
 
-- Input schema
-- Output schema
+### Is Nussknacker request response scenario a decision tree?
+The way in which the scenario diagram should be interpreted in the Request - Response processing mode requires some attention. The request - response scenario diagram is not a [decision tree](https://en.wikipedia.org/wiki/Decision_tree). During a decision tree "processing" one and only one "branch" of a tree is "activated"; "processing" ends up in one of the end leaves. As in a decision tree each leaf represents a decision (or result) - one decision (result) is generated.
```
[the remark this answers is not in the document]
Filemon279: A lot of quotes in one sentance `During a decision tree "processing" one and only one "branch" of a tree is "activated"` Maybe we should simplify this to `During a decision tree only one branch is processed at once` ?

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
+Click Properties icon on the right panel to define input and output schemas.
+
+![RR schema](img/rrProperties.png "RR properties")
 
-Thanks to the [Input schema](#request-response-schema) Nussknacker knows what data to expect in the request, so it will
-hint field names and will perform additional validations on SpEL expressions involving data from the selected topic.
+### Request-Response source
 
-![RR source](img/rrSource.png "RR source")
+The request node is very simple - it does not require any configuration. Just drop it on the diagram and proceed with authoring the scenario.
```
Filemon279, line 42: It is, but maybe we add something about input schema and how we can later use defined in schema properties in scenario? ex. `#input.name` etc.

docs/integration/KafkaIntegration.md
ZbyszekMM: My Visual Code interpreted this link correctly, not sure about Node / Docusaurs.
Do you mean "_existing incoming links will stop working_" or do you mean "_this may not work with Docusaurus"_?

docs/integration/KafkaIntegration.md
ZbyszekMM: updated

docs/integration/KafkaIntegration.md
ZbyszekMM: updated

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
 Internet. You send request which contains request parameters and you get response from the server. The protocol for this
-communication is HTTP - the same protocol is used by browsers when they communicate with web servers. One of the
-differences is that the content of the communication with the Nussknacker Lite engine ("the server") is encoded as JSON,
-not HTML. Not very surprisingly, in Request-Response processing mode the only possible Source component is Request and
-the only Sink component is Response.
+communication is HTTP - the same protocol is used by browsers when they communicate with web servers. When the  browser is used to browse internet pages, the web server responds with content encoded as HTML. In Nussknacker case, the Nussknacker Lite engine ("the server") respondes with content encoded as JSON. You may envoke Nussknacker Lite engine directly from the browser and watch the JSON response in the browser window. 
 
-## Request-Response schema
+Not very surprisingly, in the Request-Response processing mode the only possible source component is request and the only sink component is response.
```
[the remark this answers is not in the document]
ZbyszekMM: applied this suggestion

docs/scenarios_authoring/RRDataSourcesAndSinks.md
ZbyszekMM: The idea here is only to highlight the fact that there is a difference against streaming processing mode. Understanding this information does not require any prior or additional understanding of the streaming mode, it is just a fact which is easy to accept.

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
+
+In such cases, the `collect` node provides a convenient way of collecting all these records into a list. If data records are in parallel branches, a [union](./BasicNodes.md/#union) node should be used  to merge the branches first.
+
+Collect node takes two arguments:
+- Input expression - this expression will be evaluated for all data "records' which "flow" through the scenario and will be collected into a list. 
+- Output variable name - name of the variable which will store the above mentioned list.
+
+Example:
+- For-each node executes on a list `{"one", "two", "three"}`, the output variable is `Element`.
+- The subsequent variable node defines `elementSize` variable; `#Element.length` (returns length of the string) is an expression defining this variable. Because this node is "after" the for-each node, it will execute as many times as there are elements in the list on which for-each node executed, producing 3 #elementSize records.
```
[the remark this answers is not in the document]
ZbyszekMM: applied suggestion

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
+The way in which the scenario diagram should be interpreted in the Request - Response processing mode requires some attention. The request - response scenario diagram is not a [decision tree](https://en.wikipedia.org/wiki/Decision_tree). During a decision tree "processing" one and only one "branch" of a tree is "activated"; "processing" ends up in one of the end leaves. As in a decision tree each leaf represents a decision (or result) - one decision (result) is generated.
 
-Nussknacker uses those schema to detect how the request and response should look like.
+In Request-Response processing mode, the diagram is more like a flow chart. The data record processed by the scenario (#input at the very start of the processing) "flows" through the scenario. If there are [splits](./BasicNodes.md#split), the data records start to "flow" in parallel through many branches. If there are [for-each](./BasicNodes.md#foreach) nodes, multiple records are produced as the result. Because the Request-Response scenario can return only one response, a question arises what exactly will be returned in such cases and how to - if neccessary - return all data records produced in a scenario.
```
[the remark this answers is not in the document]
ZbyszekMM: The issue is that I was also not certain what are the implications of _the nodes are executed multiple times_. The fact that the node is executed multiple times does not necessarily mean that multiple some-things (how to call it?) will be generated. In fact, there could be a list of valued generated - why not? Generating a list of values would be more natural than generating multiple some-things (just thing how hard it is to deal with it in the description). So my point is that we have to state it black and white: _multiple some-things will be generated.Unless you do something about it these some-things will "flow" further down the graph._  

BTW, why do not we produce a list as the result of the for-each node?

docs/scenarios_authoring/RRDataSourcesAndSinks.md
ZbyszekMM: The section before - "one sentence away" is: _Click Properties icon on the right panel to define input and output schemas_

docs/scenarios_authoring/RRDataSourcesAndSinks.md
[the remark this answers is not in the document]
ZbyszekMM: Reply to the first comment: we use 3 terms for roughly the same idea: request, source, input - see the screenshot.  I wonder whether we need term "input" and "output" at all; for the time being I was trying to create a link in this way between all those 3 terms we use. May be it is a better idea to remove input and output at all? 
WDYT?

Re the second comment - I was trying to cover briefly the role of schemas in the Schemas section. Maybe too briefly. If expand considerations on schemas - I would prefer in the "Schemas" section.  What would you add there?

docs/scenarios_authoring/RRDataSourcesAndSinks.md
```diff
@@ -2,42 +2,73 @@
...
-## Request-Response schema
+Not very surprisingly, in the Request-Response processing mode the only possible source component is request and the only sink component is response.
 
-Each scenario needs to define two schemas in [JSON Schema](https://json-schema.org/) format:
+### Schemas
+Unlike in the streaming processing mode, there is no Schema Registry which store data schemas - request and response schemas in our case. Still, the data schema of the request and response ("input" and "output") need to be declared to Nussknacker; they are defined alongside the scenario (as the scenario property). The "language" used to define schemas is called [JSON Schema](https://json-schema.org/). Nussknacker uses those schemas to understand how the request and response should look like, assist you with hints and validate expressions you write. 
 
-- Input schema
-- Output schema
+### Is Nussknacker request response scenario a decision tree?
```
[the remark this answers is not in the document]
ZbyszekMM: Made it consistently Request-Response

docs/scenarios_authoring/RRDataSourcesAndSinks.md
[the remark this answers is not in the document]
Filemon279: Exactly, the thing is that "for-each" do not generate same amount of data as it has at the begging. 

For example to `for-each` you can pass list `{1,2,3,4,5}` and then add filter `#outputVar % 2 == 0` which is true only for {2, 4} and all other cases will be dropped on filter. Then when you collect your for-each invocations with for ex. `collect` you will get only 2 element (and u stared with 5 of them). 

I think of `for-each` like this:

After placing for-each node all of the next scenario nodes will be executed for each element in list passed as a argument, it does not have to produce any 'output'. At same point we can collect some data from x contexts and put it in a list by using 'collect' node.

docs/scenarios_authoring/RRDataSourcesAndSinks.md
Filemon279: Hmm, you are right - it should be enough :)

docs/scenarios_authoring/AggregatesInTimeWindows.md
```diff
@@ -2,7 +2,7 @@
 sidebar_position: 5
 ---
 
-# Aggregates in time windows
+# Aggregates in time windows | Streaming only
```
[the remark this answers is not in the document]
dswiecki: Right now we have stateful processing only in Flink engine - not in Lite engine - that's why I think `Steaming only` is too wide

docs/integration/KafkaIntegration.md
dswiecki: everything's good 👍

docs/scenarios_authoring/RRDataSourcesAndSinks.md
ZbyszekMM: I simplified this bit and shortend it.

ZbyszekMM:
Changes within this PR were merged within a [broader PR](https://github.com/TouK/nussknacker/pull/3725).

PR #3727, opened 2022-11-21
 Improvements: Change RR Sink validator from JsonSchemaSubclassDeterminer to JsonSchemaOutputValidator

lciolecki:
* Removed `JsonSchemaSubclassDeterminer` instead of please use `JsonSchemaOutputValidator`
  * Added param `Value validation mode` at RR response component
  * We no longer support `nullable` param from Everit schema. Nullable schema are supported by union with null e.g. `["null", "string"]


docs/Changelog.md
```diff
@@ -28,6 +28,7 @@
 * [#3709](https://github.com/TouK/nussknacker/pull/3709) Fix BestEffortJsonSchemaEncoder - fix encoding only strings when `"type": String`
 * [#3707](https://github.com/TouK/nussknacker/pull/3707) Handling apikey in OpenAPI
 * [#3716](https://github.com/TouK/nussknacker/pull/3716) Fix: JsonSchema sinks allow to add additional fields also in `strict validation mode`, if schema permits them.
+* [#3727](https://github.com/TouK/nussknacker/pull/3727) Improvements: Change RR Sink validator from JsonSchemaSubclassDeterminer to JsonSchemaOutputValidator
```
arkadius, line 31: From the user's perspective, it is a technical detail, maybe we can write some short human-friendly explanation?

arkadius (approved):
LGTM, please add migration for validation mode parameter

docs/Changelog.md
```diff
@@ -28,6 +28,9 @@
 * [#3709](https://github.com/TouK/nussknacker/pull/3709) Fix BestEffortJsonSchemaEncoder - fix encoding only strings when `"type": String`
 * [#3707](https://github.com/TouK/nussknacker/pull/3707) Handling apikey in OpenAPI
 * [#3716](https://github.com/TouK/nussknacker/pull/3716) Fix: JsonSchema sinks allow to add additional fields also in `strict validation mode`, if schema permits them.
+* [#3727](https://github.com/TouK/nussknacker/pull/3727) Improvements: Change RR Sink validation way:
+  * Added param `Value validation mode` at RR response component
+  * We no longer support `nullable` param from Everit schema. Nullable schema are supported by union with null e.g. `["null", "string"]
```
gadomsky, line 33: please add this point to MigrationGuide as well

lciolecki: Is there any sens to duplicate these information?

gadomsky: I think so. I would even say - MigrationGuide is more imporant one.

MigrationGuide should contain **all of the infromation** to upgrade NU installation consciously and safely.

Filemon279 (approved):
LGTM :+1:

PR #3730, opened 2022-11-22
Fix: Trimming additional fields at JsonToNuStruct when additionalProperties is enabled

lciolecki:


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -39,7 +39,7 @@ case class SwaggerEnum(values: List[String]) extends SwaggerTyped
 
 case class SwaggerArray(elementType: SwaggerTyped) extends SwaggerTyped
 
-case class SwaggerObject(elementType: Map[PropertyName, SwaggerTyped]) extends SwaggerTyped
+case class SwaggerObject(elementType: Map[PropertyName, SwaggerTyped], additionalPropertiesOn: Boolean, additionalPropertiesType: Option[SwaggerTyped]) extends SwaggerTyped
```
gadomsky, line 42: can we model this like that:
```
case class SwaggerObject(elementType: Map[PropertyName, SwaggerTyped], additionalProperties: Either[Boolean,SwaggerTyped]) extends SwaggerTyped
```

because now you can set `additionalPropertiesOn = false` and ant the same time `additionalPropertiesType = Some(String)` which is confusing

lciolecki: Changed

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
```diff
@@ -26,13 +26,17 @@ object JsonToNuStruct {
...
         jo => TypedMap(
           jo.toMap.collect {
-            case (key, value) if elementType.contains(key) =>
-              key -> JsonToNuStruct(value, elementType(key), addPath(key))
+            case (key, value) if obj.elementType.contains(key) =>
+              key -> JsonToNuStruct(value, obj.elementType(key), addPath(key))
+            case (key, value) if obj.additionalProperties.isInstanceOf[AdditionalPropertiesSwaggerTyped] =>
+              key -> JsonToNuStruct(value, obj.additionalProperties.asInstanceOf[AdditionalPropertiesSwaggerTyped].value, addPath(key))
+            case (key, value) if obj.additionalProperties.asInstanceOf[AdditionalPropertiesBoolean].value =>
+              key -> jsonToAny(value)
```
gadomsky, line 39: do you handle here situation when `addionalProperties == false`?
and also can we get rid of `isInstanceOf` nad `asInstanceOf` since we have seald trait?

lciolecki:
```
  test("should trim useless fields") {
    val definition = SwaggerObject(elementType = Map("field3" -> SwaggerLong), AdditionalPropertiesDisabled)
    extractor.JsonToNuStruct(json, definition) shouldBe TypedMap(Map.empty)
  }``

File: utils/json-utils/src/test/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStructTest.scala [hunk 3]
```diff
@@ -76,13 +78,41 @@ class JsonToNuStructTest extends AnyFunSuite with Matchers {
...
+        "field1" -> "value",
+        "field2"-> 1
+      ))
+    }
+  }
+
   test("should parse union") {
     val definition =
-      SwaggerObject(elementType = Map("field2" -> SwaggerUnion(List(SwaggerString, SwaggerLong))))
+      SwaggerObject(elementType = Map("field2" -> SwaggerUnion(List(SwaggerString, SwaggerLong))), AdditionalPropertiesWithoutType)
```
gadomsky, line 115: maybe let's have default value `additionalProperties = AdditionalPropertiesBoolean(true)` as it is in JsonSchema specification?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
```diff
@@ -26,13 +24,19 @@ object JsonToNuStruct {
...
+    def extractObject(obj: SwaggerObject): AnyRef =
       extract[JsonObject](
         _.asObject,
         jo => TypedMap(
           jo.toMap.collect {
-            case (key, value) if elementType.contains(key) =>
-              key -> JsonToNuStruct(value, elementType(key), addPath(key))
+            case (key, value) if obj.elementType.contains(key) =>
+              key -> JsonToNuStruct(value, obj.elementType(key), addPath(key))
+            case (key, value) => obj.additionalProperties match {
```
gadomsky, line 34: shouldnt' it be 
```
case (key, value) if obj.additionalProperties != AdditionalPropertiesBoolean(false) =>
```
so we do not have match error inside?
?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/serde/CirceJsonDeserializer.scala
```diff
@@ -28,7 +28,8 @@ class CirceJsonDeserializer(jsonSchema: Schema) {
     val jsonObject = new JSONTokener(string).nextValue()
     //after validate jsonObject has set default field values
     catchValidationError(jsonSchema.validate(jsonObject))
-    JsonToNuStruct(toCirce(jsonObject), swaggerTyped)
+    val struct = JsonToNuStruct(toCirce(jsonObject), swaggerTyped)
+    struct
```
gadomsky, line 32: let's leave it as it was before

lciolecki: Hmm, I would rather leave it.. But if you think so..

gadomsky: No it's ok. I tought that it was by mistake.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStruct.scala
lciolecki: Hmm, right!

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -132,7 +127,9 @@ object SwaggerArray {
 }
 
 object SwaggerObject {
-  import AdditionalProperties._
+  def apply(elementType: Map[PropertyName, SwaggerTyped]): SwaggerObject =
+    SwaggerObject(elementType, AdditionalPropertiesWithoutType)
```
gadomsky, line 131: Why not default directly in `SwaggerObject` definition?

utils/json-utils/src/test/scala/pl/touk/nussknacker/engine/json/swagger/extractor/JsonToNuStructTest.scala
```diff
@@ -76,13 +78,41 @@ class JsonToNuStructTest extends AnyFunSuite with Matchers {
...
+  test("should throw exception on trying convert string to integer") {
+    val json = Json.obj("field1" -> fromString("value"), "field2" -> Json.fromInt(1))
+    val definition = SwaggerObject(elementType = Map("field3" -> SwaggerLong), AdditionalPropertiesSwaggerTyped(SwaggerLong))
+
+    intercept[JsonToObjectError] {
+      extractor.JsonToNuStruct(json, definition) shouldBe TypedMap(Map(
+        "field1" -> "value",
+        "field2"-> 1
+      ))
+    }
```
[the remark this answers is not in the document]
gadomsky: still there is 
```
shouldBe TypedMap(Map(
        "field1" -> "value",
        "field2"-> 1
      ))
```
which is unnecessery

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
lciolecki: Because it's case class, and we shouldn't do default value here? But I understand you want to by consistent with json schema?

gadomsky (approved):
Nice, thanks :)

PR #3740, opened 2022-11-24
Improvements: Refactor Json Schema streaming functional tests

lciolecki:


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBuilder.scala
```diff
@@ -6,11 +6,11 @@ import org.json.JSONObject
 
 object JsonSchemaBuilder {
 
-  def parseSchema(rawJsonSchema: String): Schema =
+  def parseSchema[T <: Schema](rawJsonSchema: String): T =
```
gadomsky, line 9: I can't see any invocation different than `parseSchema[Schema]` so why do we need this generic here?

mproch: yeah, maybe it'd be better to have two methods, if we use `parseSchema[Schema]` in > 90% cases and it's just casting anyway ;)

lciolecki: I've changed it a little bit.

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/utils/JsonTestData.scala
```diff
@@ -59,143 +61,142 @@ object JsonTestData {
...
+
+  val objOutputAsInputFieldSpEL: Map[String, SpecialSpELElement] = Map(ObjectFieldName -> OutputField)
+
+  val sampleObjStrSpEL: Map[String, String] = Map(ObjectFieldName -> sampleStr)
+
+  val sampleObjMapIntSpEL: Map[String, Map[String, Integer]] = Map(ObjectFieldName -> sampleMapIntSpEL)
+
+  val sampleObjMapAnySpEL: Map[String, Map[String, Any]] = Map(ObjectFieldName -> sampleMapAnySpEL)
+
+  val sampleObjMapPersonSpEL: Map[String, Map[String, Any]] = Map(ObjectFieldName -> sampleMapPersonSpEL)
```
gadomsky, line 121: Why those variable are named ..SpEL?

gadomsky (approved):

PR #3742, opened 2022-11-25
Improvements: Json Schema more strict encoding - matching to schema validate

lciolecki:
* Remove `ValidationMode` from `BestEffortJsonSchemaEncoder`
* `BestEffortJsonSchemaEncoder` at the end of encoding runs json schema `validate` to proper verification data
* `JsonToNuStruct` convert json number additional property without schema to java.math.BigDecimal instead of scala.math.BigDecimal
* More functional tests for RR Lite & Streaming Lite 
* Consistent way of encoding data to json on RR Sink and Streaming Sink


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -107,6 +108,19 @@ class BestEffortJsonSchemaEncoder(validationMode: ValidationMode) {
...
+        schema.validate(json) //validate can change json object.. e.g. fill default values
+        JsonSchemaUtils.jsonToCirce(json)
+      }
+
+      Validated
+        .fromTry(tryToValid)
+        .leftMap(e => NonEmptyList.of(e.getMessage))
+    }
+
   def encode(value: Any, schema: Schema, fieldName: Option[String] = None): EncodeOutput = {
```
gadomsky, line 124: this method is public and allows encoding without validation

lciolecki: This method is public because we use it at `BestEffortJsonSchemaEncoderTest`. We can't make it private because package of `BestEffortJsonSchemaEncoderTest` is different then `BestEffortJsonSchemaEncoder` but maybe I will change it..  It's intentional to make  `encodeWithJsonValidation` as separate method, the end result will be tested on functional tests. I think there is no need to test whole cases here because we will have to test them again on functional tests.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -1,59 +1,52 @@
...
-          case Some(additionalPropertySchema) => encode(value, additionalPropertySchema).map(fieldName -> _)
-          case None =>  Valid(jsonEncoder.encode(value)).map(fieldName -> _)
-        }
+    fields.keys.toList.union(parentSchema.getPropertySchemas.keySet.asScala.toList).distinct.map{ key =>
+      val schema = Option(parentSchema.getPropertySchemas.get(key))
+      val value = fields.getOrElse(key, null)
+      (key, value, fields.contains(key), schema)
+    }.collect {
+      case (fieldName, null, true, Some(schema)) if schema.isNullableSchema => Valid((fieldName, Json.Null))
+      case (fieldName, null, false, Some(_)) if parentSchema.getRequiredProperties.contains(fieldName) =>
```
mproch, line 39: do we need to handle nulls here at all? The first case will be handled by `encodeBasedOnSchema`, the second will be handled by validation

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -1,59 +1,52 @@
...
+        error(s"Missing property: $fieldName for schema: $parentSchema.")
+      case (fieldName, value, true, Some(schema)) if parentSchema.getPropertySchemas.keySet().contains(fieldName) =>
```
mproch, line 41: I think this if is redundant, if schema is not None, it means the key is there?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -1,59 +1,52 @@
...
-import pl.touk.nussknacker.engine.util.json.{BestEffortJsonEncoder, EncodeInput, EncodeOutput, ToJsonBasedOnSchemaEncoder}
+import pl.touk.nussknacker.engine.util.json._
 
 import java.time.{LocalDate, OffsetDateTime, OffsetTime, ZonedDateTime}
 import java.util.ServiceLoader
 import scala.collection.convert.ImplicitConversions.`map AsScala`
 import scala.jdk.CollectionConverters.iterableAsScalaIterableConverter
 
+
+//TODO: Do we need validation mode here??
```
mproch, line 17: No, we don't. Let's just remove it

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/util/json/JsonSchemaUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.json.{JSONArray, JSONObject, JSONTokener}
+import io.circe.Json._
+
+object JsonSchemaUtils {
+
+  import collection.JavaConverters._
+
+  private val CirceJsonFolder: Json.Folder[Object] = new Json.Folder[Object] {
+    def onNull: Object = JSONObject.NULL
+    def onBoolean(value: Boolean): Object = Predef.boolean2Boolean(value)
```
mproch, line 13: I think it will be handled automatically?

lciolecki: If I set ` def onBoolean(value: Boolean): Object = value` then IDEA throws:
`the result type of an implicit conversion must be more specific than Object
    def onBoolean(value: Boolean): Object = value
`

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/util/json/JsonSchemaUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+    def onString(value: String): Object = value
+    def onNumber(value: JsonNumber): Object =
+      value.toInt match {
+        case Some(int) => Predef.int2Integer(int)
+        case None => new JSONTokener(value.toString).nextValue
```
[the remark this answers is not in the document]
lciolecki: Hmm, you're right. I took shortcuts and I used: https://github.com/circe/circe-json-schema/blob/master/schema/src/main/scala/io/circe/schema/Schema.scala#L36 :) I will leave:  new JSONTokener(value.toString).nextValue

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/util/json/JsonSchemaUtils.scala
mproch: I think we usually use `value : java.lang.Boolean` or `def onBoolean(value: Boolean): java.lang.Boolean = value` in such cases

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
lciolecki: We do here collect, so we need to handle this cases if we want to go to `encodeBasedOnSchema`. In fact we can remove whole collect cases and replace it by doing `encode` on all  map fields and at the end run schema validation.. 

As we had spoken before, we want here do some own pre-validation with humanfriedly message.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
lciolecki: Ahh, your're right.. This if is done in patter matching..  I'm silly ;)

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
mproch: let's add a comment that this is for better error messages

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -105,9 +95,18 @@ class BestEffortJsonSchemaEncoder(validationMode: ValidationMode) {
...
+    encode(value, schema, fieldName).andThen { result =>
+      val json = JsonSchemaUtils.circeToJson(result)
+      schema
+        .validateData(json)
+        .leftMap(errorMsg => NonEmptyList.of(errorMsg))
+        .map(_ => result) //we return base json (e.g. without felt defaults - there is no need to put these data as output)
+    }
 
-  def encode(value: Any, schema: Schema, fieldName: Option[String] = None): EncodeOutput = {
+  private[encode] def encode(value: Any, schema: Schema, fieldName: Option[String] = None): EncodeOutput = {
```
gadomsky, line 109: Can we make this method private, and change unit tests to invoke this one `private[encode] def encodeWithJsonValidation` instead?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -22,38 +24,27 @@ class BestEffortJsonSchemaEncoder(validationMode: ValidationMode) {
...
-        case ((fieldName, value), propertySchema) if (propertySchema != null) && notNullOrRequired(fieldName, value, parentSchema) => encode(value, propertySchema, Some(fieldName)).map(fieldName -> _)
-        case ((fieldName, _), null) if !parentSchema.permitsAdditionalProperties() => error(s"Not expected field with name: ${fieldName} for schema: $parentSchema and policy $validationMode does not allow redundant")
-        case ((fieldName, value), null) => Option(parentSchema.getSchemaOfAdditionalProperties) match {
-          case Some(additionalPropertySchema) => encode(value, additionalPropertySchema).map(fieldName -> _)
-          case None =>  Valid(jsonEncoder.encode(value)).map(fieldName -> _)
-        }
+    fields.keys.toList.union(parentSchema.getPropertySchemas.keySet.asScala.toList).distinct.map{ key =>
+      val schema = Option(parentSchema.getPropertySchemas.get(key))
+      val value = fields.getOrElse(key, null)
+      (key, value, fields.contains(key), schema)
```
gadomsky, line 34: `value, fields.contains(key),` -> these two can be changed to single Option.

Then we will have three fields in pattern-matching (which is still a lot) and match for `None` or `Some(value)`

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -105,9 +95,18 @@ class BestEffortJsonSchemaEncoder(validationMode: ValidationMode) {
...
-    Invalid(NonEmptyList.of(s"Not expected type: ${runtimeType} for field${fieldName.map(f => s": '$f'").getOrElse("")} with schema: $schema"))
+    Invalid(NonEmptyList.of(s"Not expected type: $runtimeType for field${fieldName.map(f => s": '$f'").getOrElse("")} with schema: $schema."))
+
+  private[encode] def encodeWithJsonValidation(value: Any, schema: Schema, fieldName: Option[String] = None): EncodeOutput =
+    encode(value, schema, fieldName).andThen { result =>
+      val json = JsonSchemaUtils.circeToJson(result)
+      schema
+        .validateData(json)
+        .leftMap(errorMsg => NonEmptyList.of(errorMsg))
+        .map(_ => result) //we return base json (e.g. without felt defaults - there is no need to put these data as output)
```
gadomsky, line 106: in [deserialization](https://github.com/TouK/nussknacker/pull/3742/files#diff-e512d01d8979f3d4c273889b9675a818ade23b7488403e051c1961340d35fa8bR36)   we take validated object, and here not, why?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
lciolecki: We can't.. Because some can explicit send `null` here :)

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
lciolecki: Because when you do deserialization you want to have all data in HashMap -> e.g. after source you can use field with default and this field should be felt. At sink we want to have optimized object (there is no need to fill by defaults etc. because this object is still valid)

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
lciolecki: Hmmm, Yes I can ;)

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
gadomsky: `Some(null)` won't be possible ?

lciolecki: `Option(null)` is None? :)

lciolecki: I cane remove 3th param and replace it by explicit ` fields.contains(fieldName)`

lciolecki: You was right, fields.get(key) for null returns Some(null). I though it would return None.. :)

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
```diff
@@ -1,19 +1,20 @@
...
+import org.everit.json.schema._
+import pl.touk.nussknacker.engine.util.json._
 
 import java.time.{LocalDate, OffsetDateTime, OffsetTime, ZonedDateTime}
 import java.util.ServiceLoader
 import scala.collection.convert.ImplicitConversions.`map AsScala`
 import scala.jdk.CollectionConverters.iterableAsScalaIterableConverter
 
-class BestEffortJsonSchemaEncoder(validationMode: ValidationMode) {
+object BestEffortJsonSchemaEncoder {
```
mproch, line 15: TBH I wonder if it should be object, as we may want to customize `optionalEncoders` and/or `highPriority` in the future?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
gadomsky: cool, thanks! looks a lot better now!

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/encode/BestEffortJsonSchemaEncoder.scala
mproch: :+1: can be...

mproch (approved):

PR #3751, opened 2022-11-29
Configuration docs cognitive aspects polishing

ZbyszekMM:
Rewrote the "Configuring replicas count" in R-R processing mode.
+
This PR focuses on the cognitive aspects of the configuration documentation. The purpose is to reduce the cognitive distress of the reader - there is so much to understand and remember that we need to give the reader a hand in the journey through the documentation. 

The typical questions asked when you install Nu and try to understand its configuration:
- will I need to configure anything in this particular area (Designer, Deployment Manager, Model) or is everything set for me?
- If there are incremental changes to be applied, how to apply them in the Docker case and how to apply them in the K8s case?

These questions need to be answered for all three configuration areas.

### DIFF
--- a/docs/installation_configuration_guide/DeploymentManagerConfiguration.md
+++ b/docs/installation_configuration_guide/DeploymentManagerConfiguration.md
@@ -10,12 +10,13 @@ Below you can find a snippet of Deployment Manager configuration.
 deploymentConfig {     
   type: "flinkStreaming"
   restUrl: "http://localhost:8081"
+  
   # additional configuration goes here
 }
 ```
-`type` parameter determines engine to which the scenario is deployed. Currently there are only two values allowed: `flinkStreaming` and `lite-k8s`. The `type` parameter is set in both the minimal working configuration file (Docker image and binary distribution) and Helm chart - you will need to set it on your won. 
+`type` parameter determines engine to which the scenario is deployed.The `type` parameter is set in both the minimal working configuration file (Docker image and binary distribution) and Helm chart - you will not need to set it on your won. 
 
-
+&nbsp;
 ## Kubernetes native Lite engine configuration
                                                                                 
 Please note, that K8s Deployment Manager has to be run with properly configured K8s access. If you install the Designer
@@ -38,7 +39,8 @@ The table below contains configuration options for the Lite engine. If you insta
 | logbackConfigPath         | string                                              | {}                                  | see [below](#configuring-runtime-logging)                                                |
 | commonConfigMapForLogback | string                                              | {}                                  | see [below](#configuring-runtime-logging)                                                |
 | servicePort               | int                                                 | 80                                  | Port of service exposed in request-response processing mode                              |
-                                                 
+
+&nbsp;                                                 
 ### Customizing K8s deployment resource definition
 
 By default, each scenario creates K8s Deployment. By default, Nussknacker will create following deployment:
@@ -143,7 +145,8 @@ spec {
 ```
 This config will be merged into the final K8s deployment resource definition. 
 Please note that you cannot override names or labels configured by Nussknacker. 
-                              
+
+&nbsp;                              
 ### Overriding configuration passed to runtime. 
 
 In most cases, the model configuration values passed to the Lite Engine runtime are the ones from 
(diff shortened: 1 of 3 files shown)


docs/installation_configuration_guide/DeploymentManagerConfiguration.md
```diff
@@ -3,21 +3,22 @@ sidebar_position: 3
...
+
+Below you can find a snippet of Deployment Manager configuration. 
 ```
 deploymentConfig {     
   type: "flinkStreaming"
   restUrl: "http://localhost:8081"
   # additional configuration goes here
 }
 ```
+`type` parameter determines engine to which the scenario is deployed. Currently there are only two values allowed: `flinkStreaming` and `lite-k8s`. The `type` parameter is set in both the minimal working configuration file (Docker image and binary distribution) and Helm chart - you will need to set it on your won.
```
mproch, line 16: There are also embedded types: `request-response-embedded`, `streaming-lite-embedded`, though they're not heavy-docummented. What's more, they can be customizable with code

ZbyszekMM: Removed the fragment with allowed values, corrected the meaning ("not" was missing) -> _you will NOT need to ..._

mproch (approved):
:+1:

docs/installation_configuration_guide/DeploymentManagerConfiguration.md
```diff
@@ -165,39 +169,42 @@ modelConfig {
 }
 ```
 
+&nbsp;
 ### Configuring replicas count
 
-Each scenario has its own, configured parallelism. It describes how many worker threads,
```
ZbyszekMM, line 170: Decideded to rewrite this section as the fragment on replicas count in R-R sounded like simply bolted onto the previous content without thinking about the overall structure. 

Also, decided to remove terms  _worker thread_ and _workers_ . The setting is called _tasksPerReplica_, so referring to it later as to _worker threads_ is misleading. Is it something different from these tasks?

docs/installation_configuration_guide/DeploymentManagerConfiguration.md
```diff
@@ -165,39 +169,42 @@ modelConfig {
...
-- ```{ tasksPerReplica: y }```
-- by default `tasksPerReplica: 4`
+In the request-response processing mode, the only way to affect parallelism is to set the count of scenario pods (replicas). This is achieved with the `fixedReplicasCount` configuration key; its default setting is 2:
+
+`{ fixedReplicasCount: x }`.
+
+​In the  streaming processing mode you can affect scenario parallelism in the following ways:
+- Set scenario parallelism in the scenario properties - this will determine how many parallel tasks are used to process the scenario. The number of replicas in the K8s deployment will be set to the `ceiling(scenarioParallelism / tasksPerReplica)`. In this case, the default value of 4 will be used for `tasksPerReplica`.
+- Modify the default value of `tasksPerReplica` by setting:
+  - `{ tasksPerReplica: t }`
```
Filemon279, line 182:
```suggestion
  - `{ tasksPerReplica: y }`
```

Filemon279 (approved):
LGTM :+1:

docs/installation_configuration_guide/ModelConfiguration.md
```diff
@@ -3,14 +3,14 @@ sidebar_position: 5
...
+Model defines how to configure [components](https://nussknacker.io/documentation/about/GLOSSARY#component) and certain runtime behavior (e.g. error handling) for a given scenario type. Model configuration is processed not only at the Designer but also passed to the execution engine (e.g. Flink), that’s why it’s parsed and processed a bit differently: 
 
-Look at [configuration areas](./#configuration-areas) to understand where Model configuration should be placed in Nussknacker configuration.
-                  
+* Some Components can use a special mechanism which resolves and adds additional configuration during deployment, which is then passed to the execution engine. Such configuration is read and resolved only at the Designer. Example: OpenAPI enrichers need to read its definition from external sites - so e.g. Flink cluster does not have to have access to the site with the definition.
```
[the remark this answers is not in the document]
ZbyszekMM: applied

docs/installation_configuration_guide/DeploymentManagerConfiguration.md
ZbyszekMM: applied

PR #3754, opened 2022-11-30
New scenario migration fixup

### LINKED ISSUE #3700: Migration of a new scenario throws an error "No scenario X found"
**Describe the bug**
Migration of a new scenario throws an error "No scenario X found", when scenario with the same name doesn't exist on the secondary env.

**To Reproduce**
1. Create a scenario & save
2. Try to migrate the scenario to the secondary env

**Technical context**
During migration before updating/creating scenario on the remote env we first validate the scenario against the remote env.
Error is issued because endpoint used for validation on remote env now requires the scenario to already exists in db (it is used to discover scenario category which is not included in validation request).
To deal with it we probably should add category to validation request in some way.

**Environment (please complete the following information):**
 - Nussknacker version 1.6.1-release_1.6-2022-11-07-10140-a15598adc26583e9d260d77b478350861e006cb5-SNAPSHOT

raphaelsolarski:
fixes #3700

### DIFF
--- a/designer/client/types/process.ts
+++ b/designer/client/types/process.ts
@@ -14,6 +14,7 @@ export type Process = {
   properties: PropertiesType,
   validationResult: ValidationResult,
   processingType?: ProcessingType,
+  category?: string, // optional - see the comment for a field with the same name in DisplayableProcess.scala
 }
 
 export type ProcessId = Process["id"]
--- a/designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
+++ b/designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
@@ -158,13 +158,13 @@ trait EspItTest extends LazyLogging with WithHsqlDbTesting with TestPermissions 
 
   protected def createDeploymentManager(): MockDeploymentManager = new MockDeploymentManager
 
-  protected def saveProcessAndAssertSuccess(processId: String, process: CanonicalProcess): Assertion =
-    saveProcess(ProcessName(processId), process) {
+  protected def saveProcessAndAssertSuccess(processId: String, process: CanonicalProcess, category: String = TestCat): Assertion =
+    saveProcess(ProcessName(processId), process, category) {
       status shouldEqual StatusCodes.OK
     }
 
-  protected def saveProcess(processName: ProcessName, process: CanonicalProcess)(testCode: => Assertion): Assertion =
-    createProcessRequest(processName) { _ =>
+  protected def saveProcess(processName: ProcessName, process: CanonicalProcess, category: String)(testCode: => Assertion): Assertion =
+    createProcessRequest(processName, category) { _ =>
       val json = parser.decode[Json](responseAs[String]).right.get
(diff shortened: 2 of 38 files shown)


designer/client/types/process.ts
```diff
@@ -14,6 +14,7 @@ export type Process = {
   properties: PropertiesType,
   validationResult: ValidationResult,
   processingType?: ProcessingType,
+  category?: string, //// optional - see the comment for a field with the same name in DisplayableProcess.scala
```
mproch, line 17: double comment? :)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
```diff
@@ -173,24 +173,24 @@ trait EspItTest extends LazyLogging with WithHsqlDbTesting with TestPermissions
       updateProcess(processName, process)(testCode)
     }
 
-  protected def saveProcess(process: DisplayableProcess)(testCode: => Assertion): Assertion =
-    createProcessRequest(ProcessName(process.id)) { code =>
+  protected def saveProcess(process: DisplayableProcess, category: ProcessingType)(testCode: => Assertion): Assertion =
```
mproch, line 176: I think it would be simpler to have process.category.get here and assume that in tests we have category in displayable

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
```diff
@@ -173,24 +173,24 @@ trait EspItTest extends LazyLogging with WithHsqlDbTesting with TestPermissions
...
-    Post(s"/processes/${processName.value}/$TestCat?isSubprocess=false") ~> processesRouteWithAllPermissions ~> check {
+  protected def createProcessRequest(processName: ProcessName, category: String = TestCat)(callback: StatusCode => Assertion): Assertion =
+    Post(s"/processes/${processName.value}/$category?isSubprocess=false") ~> processesRouteWithAllPermissions ~> check {
       callback(status)
     }
 
-  protected def saveSubProcess(process: CanonicalProcess)(testCode: => Assertion): Assertion = {
-    val displayable = ProcessConverter.toDisplayable(process, TestProcessingTypes.Streaming)
+  protected def saveSubProcess(process: CanonicalProcess, category: String)(testCode: => Assertion): Assertion = {
+    val displayable = ProcessConverter.toDisplayable(process, TestProcessingTypes.Streaming, category)
```
mproch, line 188: I wonder, we assume fixed ProcessingType here, maybe we can do so also for category?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
raphaelsolarski: ok

designer/client/types/process.ts
raphaelsolarski: ok

mproch (approved):

PR #3757, opened 2022-12-01
[FIX] export and test work correctly with unclean FE state

mproch:


designer/client/actions/nk/importExport.ts
```diff
@@ -0,0 +1,43 @@
...
+    )
+    return {
+      type: "IMPORT_FILES",
+    }
+  }
+}
+
+export function exportProcessToJSON(process: Process, versionId: ProcessVersionId): ThunkAction {
+  return (dispatch) => {
+    let noEmptyEdges = withoutEmptyEdges(process)
```
arkadius, line 22: Can we have this cleaning in one class used in both cases?

mproch: added comment, changed name to suggest the need of better solution :)

PR #3770, opened 2022-12-05
Finer typing of maps in JSON Schemas

mproch:


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSinkValueParameter.scala
```diff
@@ -24,7 +24,7 @@ object JsonSinkValueParameter {
   private def toSinkValueParameter(schema: Schema, paramName: Option[String], defaultParamName: String, defaultValue: Option[Expression], isRequired: Option[Boolean])
                                   (implicit nodeId: NodeId): ValidatedNel[ProcessCompilationError, SinkValueParameter] = {
     schema match {
-      case objectSchema: ObjectSchema =>
+      case objectSchema: ObjectSchema if !(objectSchema.getPropertySchemas.isEmpty && objectSchema.permitsAdditionalProperties()) =>
```
gadomsky, line 27: can we extract `objectSchema.getPropertySchemas.isEmpty && objectSchema.permitsAdditionalProperties()` to a method?
my point is to have descriptive name, cause this expressions means, that schema represent Map, right?

gadomsky (approved):
thanks, I should have thought about this case

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSinkValueParameter.scala
gadomsky: actually we have it already. `JsonSchemaOutputValidator` provides with implicit class with this method

mproch: :+1: I moved `representsMap` to `JsonSchemaImplicits`

engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/LiteRequestResponseFunctionalTest.scala
```diff
@@ -76,6 +77,61 @@ class LiteRequestResponseFunctionalTest extends AnyFunSuite with Matchers with E
...
+        "inputSchema" -> input,
+        "outputSchema" -> output
+      ))
+      .source("input", "request")
+      .emptySink(sinkName, "response",
+        (SinkRawEditorParamName -> ("false": Expression)) :: params: _*
+      )
+
+    val result = runner.runWithRequests(scenario) { invoker =>
+      invoker(HttpRequest(HttpMethods.POST, entity = "{}"))
```
dswiecki, line 127: is that call needed ?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/util/json/JsonSchemaImplicits.scala
```diff
@@ -12,6 +12,11 @@ object JsonSchemaImplicits {
 
   implicit class ExtendedSchema(schema: Schema) {
 
+    def representsMap: Boolean = schema match {
```
lciolecki, line 15: isMapSchema? :)

engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/LiteRequestResponseFunctionalTest.scala
arkadius: exactly, it is dead code ;)

mproch: :+1:

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/util/json/JsonSchemaImplicits.scala
gadomsky: and what's wrong with `representsMap`? The point of this method is to check whether a schema should be translated to Map type in NU. Maybe in the future this implementation will evolve and we will check another conditions, not only `addionoalProperties` - that's why I don't like `hasAdditionalPropertiesOnly`

gadomsky: ok. But having a name `representsMap` has allowed me to not write comments why I handle differently this schema when typing. Now it might be not as understandable to a reader. But ok. let's leave it and move on.

arkadius: For me it is the metter where this method is, it is part of "our" logic or "json schema logic". Do we added some extensions for easier handling json schemas that are related with json schema or it is our specific extensions. If it is our specific extensions we should name class otherwise, place it closer to place of usage etc. If its are "neutral" schema extensions, we shouldn't introduce some specific logic. TL;DR for me it could be fine to call it `representsMap` but rename class / move class then.

raphaelsolarski (approved):

arkadius (approved):

PR #3775, opened 2022-12-06
Allow to pass custom context id to service invocation

Filemon279:


engine/lite/request-response/components-api/src/main/scala/pl/touk/nussknacker/engine/requestresponse/api/RequestResponseSourceFactory.scala
```diff
@@ -31,6 +32,11 @@ trait RequestResponseSource[T] extends BaseLiteSource[Any] {
   def openApiDefinition: Option[OpenApiSourceDefinition] = None
 
   override def transform(record: Any): Context = {
-    new BasicContextInitializingFunction[Any](contextIdGenerator, VariableConstants.InputVariableName)(record)
+    record match {
```
arkadius, line 35: Let's pass to RR sources additional http request fields like headers instead. Thanks to that, contextId can be extracted from headers

arkadius (changes requested):
.

Filemon279:
It turned out in the mid time that my change was solving a scenario which is not common and only happen in my project. I will close it then. Thanks for your time :)

PR #3776, opened 2022-12-06
Handle recursive schema in OpenAPI with fallback

mproch:

### DIFF
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
@@ -53,8 +53,10 @@ case object SwaggerRecursiveSchemaFallback extends SwaggerTyped
 
 object SwaggerTyped {
 
+  def apply(schema: Schema[_], swaggerRefSchemas: SwaggerRefSchemas): SwaggerTyped = apply(schema, swaggerRefSchemas, Set.empty)
+
   @tailrec
-  def apply(schema: Schema[_], swaggerRefSchemas: SwaggerRefSchemas, usedSchemas: Set[String] = Set.empty): SwaggerTyped = schema match {
+  private[swagger] def apply(schema: Schema[_], swaggerRefSchemas: SwaggerRefSchemas, usedSchemas: Set[String]): SwaggerTyped = schema match {
     case objectSchema: ObjectSchema => SwaggerObject(objectSchema, swaggerRefSchemas, usedSchemas)
     case mapSchema: MapSchema => SwaggerObject(mapSchema, swaggerRefSchemas, usedSchemas)
     case arraySchema: ArraySchema => SwaggerArray(arraySchema, swaggerRefSchemas, usedSchemas)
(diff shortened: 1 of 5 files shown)


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -48,23 +48,29 @@ case class SwaggerObject(elementType: Map[PropertyName, SwaggerTyped], additiona
...
+case object SwaggerRecursiveSchemaFallback extends SwaggerTyped
+
 object SwaggerTyped {
 
   @tailrec
-  def apply(schema: Schema[_], swaggerRefSchemas: SwaggerRefSchemas): SwaggerTyped = schema match {
-    case objectSchema: ObjectSchema => SwaggerObject(objectSchema, swaggerRefSchemas)
-    case mapSchema: MapSchema => SwaggerObject(mapSchema, swaggerRefSchemas)
-    case arraySchema: ArraySchema => SwaggerArray(arraySchema, swaggerRefSchemas)
+  def apply(schema: Schema[_], swaggerRefSchemas: SwaggerRefSchemas, usedSchemas: Set[String] = Set.empty): SwaggerTyped = schema match {
```
arkadius, line 57: Can we have two methods - one public, without usedSchemas and second private? Do you see some case when we would like to pass it from the upper level?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -83,7 +89,7 @@ object SwaggerTyped {
     }
   }
 
-  private def swaggerUnion(schemas: java.util.List[Schema[_]], swaggerRefSchemas: SwaggerRefSchemas) = SwaggerUnion(schemas.asScala.map(SwaggerTyped(_, swaggerRefSchemas)).toList)
+  private def swaggerUnion(schemas: java.util.List[Schema[_]], swaggerRefSchemas: SwaggerRefSchemas) = SwaggerUnion(schemas.asScala.map(SwaggerTyped(_, swaggerRefSchemas, Set.empty)).toList)
```
arkadius, line 92: Why empty?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -48,23 +48,29 @@ case class SwaggerObject(elementType: Map[PropertyName, SwaggerTyped], additiona
...
+    case objectSchema: ObjectSchema => SwaggerObject(objectSchema, swaggerRefSchemas, usedSchemas)
+    case mapSchema: MapSchema => SwaggerObject(mapSchema, swaggerRefSchemas, usedSchemas)
+    case arraySchema: ArraySchema => SwaggerArray(arraySchema, swaggerRefSchemas, usedSchemas)
     case _ => Option(schema.get$ref()) match {
+      //handle recursive schemas better
+      case Some(ref) if usedSchemas.contains(ref) =>
+        SwaggerRecursiveSchemaFallback
       case Some(ref) =>
-        SwaggerTyped(swaggerRefSchemas(ref), swaggerRefSchemas)
+        SwaggerTyped(swaggerRefSchemas(ref), swaggerRefSchemas, usedSchemas = usedSchemas + ref)
```
arkadius, line 66: Does it mean that we will find recursion only on "second" level?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
mproch: :+1: add second method

arkadius (approved):

PR #3781, opened 2022-12-07
Deployment manager stats + source field

arkadius:

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsHtmlSnippet.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsHtmlSnippet.scala
@@ -25,19 +25,36 @@ object UsageStatisticsHtmlSnippet {
     }
   }
 
-  private def prepareQueryParams(config: UsageStatisticsReportsConfig,
-                                 processingTypeStatisticsMap: Map[ProcessingType, ProcessingTypeUsageStatistics]): ListMap[ProcessingType, ProcessingType] = {
+  private[statistics] def prepareQueryParams(config: UsageStatisticsReportsConfig,
+                                             processingTypeStatisticsMap: Map[ProcessingType, ProcessingTypeUsageStatistics]): ListMap[String, String] = {
+    val deploymentManagerTypes = processingTypeStatisticsMap.values.map(_.deploymentManagerType)
+    val dmParams = prepareValuesParams(deploymentManagerTypes, "dm")
+    val processingModes = processingTypeStatisticsMap.values.collect {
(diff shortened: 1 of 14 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsHtmlSnippet.scala
```diff
@@ -12,25 +14,68 @@ case class UsageStatisticsHtmlSnippet(value: String)
...
+                                             processingTypeStatisticsMap: Map[ProcessingType, ProcessingTypeUsageStatistics]): ListMap[String, String] = {
+    val deploymentManagerTypes = processingTypeStatisticsMap.values.map(_.deploymentManagerType).map {
+      case dm if knownDeploymentManagerTypes.contains(dm) => dm
+      case _ => aggregateForCustomValues
+    }
+    val dmParams = prepareValuesParams(deploymentManagerTypes, "dm")
+
+    val processingModes = processingTypeStatisticsMap.values.collect {
+      case ProcessingTypeUsageStatistics(_, Some(mode)) if knownProcessingModes.contains(mode) => mode
+      case ProcessingTypeUsageStatistics(_, Some(_)) => aggregateForCustomValues
```
mproch, line 47: I think this should be default at the end, regardless of mode?

nussknacker-dist/src/universal/bin/nussknacker-entrypoint.sh
```diff
@@ -35,6 +35,10 @@ else
   JAVA_PROMETHEUS_OPTS="-javaagent:$agentPath=$PROMETHEUS_METRICS_PORT:$PROMETHEUS_AGENT_CONFIG_FILE"
 fi
 
+if [ "$USAGE_REPORTS_SOURCE" == "" ]; then
```
mproch, line 38: I wonder if we need to allow overrinding it? Or it's just in case?

arkadius: It is intensional. "docker" is too much general. I want to override it by "quickstart-docker-compose", "helmchart"

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsHtmlSnippet.scala
arkadius: :+1: changed

nussknacker-dist/src/universal/bin/nussknacker-entrypoint.sh
mproch: :+1:

mproch (approved):
:+1:

PR #3792, opened 2022-12-08
Fixup to NotBlankParameterValidator

raphaelsolarski:
Fixup to NotBlankParameterValidator:
* not empty expressions containing somewhere '' were treated as errors by FE validator
* spel string literals created using "" weren't checked
* test on be/fe added

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -50,7 +50,7 @@ case object MandatoryParameterValidator extends ParameterValidator {
 
 case object NotBlankParameterValidator extends ParameterValidator {
 
-  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("'\\s*'")
+  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("['\"]\\s*['\"]")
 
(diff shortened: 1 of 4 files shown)


mproch (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -50,7 +50,7 @@ case object MandatoryParameterValidator extends ParameterValidator {
 
 case object NotBlankParameterValidator extends ParameterValidator {
 
-  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("'\\s*'")
+  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("['\"]\\s*['\"]")
```
arkadius, line 53: Shouldn't we check that the closing quote is the same as the opening one?

bartektartanus: Totally agree. Regex should look like this:
```
"(['\"])\\s*\\1"
```

arkadius (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
raphaelsolarski: regex stays as is 😄 we have common agreement that we validate only expressions that are possible to compile

PR #3793, opened 2022-12-09
kafka & sr docker images to latest versions

gadomsky:

### DIFF
--- a/e2e-tests/docker-compose.yml
+++ b/e2e-tests/docker-compose.yml
@@ -8,6 +8,7 @@ services:
       DEFAULT_SCENARIO_TYPE: streaming-lite-embedded
       JDK_JAVA_OPTIONS: -Xmx512M
       KAFKA_ADDRESS: kafka:9092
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -256,7 +256,8 @@ lazy val commonSettings =
 val flinkV = "1.16.0"
 val avroV = "1.11.0"
 //we should use max(version used by confluent, version acceptable by flink), https://docs.confluent.io/platform/current/installation/versions-interoperability.html - confluent version reference
-val kafkaV = "3.2.3"
+val kafkaClientV = "3.2.3"
+val kafkaServerV = "3.3.1"
```
mproch, line 260: There is a dependency on `kafka-clients` in kafka server, so I don't think it's a good idea to mix versions...

gadomsky: Yes, I noticed this when the PR failed.

dswiecki (approved):
lgtm

PR #3809, opened 2022-12-13
Fix support for enum type in JsonSchema

gadomsky:


arkadius (approved):

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -105,8 +105,8 @@ object SwaggerTyped {
       TypedObjectTypingResult(elementType.mapValuesNow(typingResult).toList.sortBy(_._1))
     case SwaggerArray(ofType) =>
       Typed.genericTypeClass(classOf[java.util.List[_]], List(typingResult(ofType)))
-    case SwaggerEnum(_) =>
-      Typed.typedClass[String]
```
lciolecki, line 109: What about validation?  Before we assumed that an enum is represented by a string right now it's a set. Are you sure? In Avro, it works in a different way..

PR #3810, opened 2022-12-13
Usage statistcs reporting trggered by FE client

gadomsky:

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
@@ -6,12 +6,14 @@ import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport
 import io.circe.{Decoder, Encoder}
 import io.circe.generic.JsonCodec
 import pl.touk.nussknacker.ui.config.{AnalyticsConfig, FeatureTogglesConfig}
+import pl.touk.nussknacker.ui.statistics.UsageStatisticsReportsSettings
 
 import scala.concurrent.ExecutionContext
 
 class SettingsResources(config: FeatureTogglesConfig,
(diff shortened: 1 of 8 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
```diff
@@ -95,7 +98,8 @@ object TopTabType extends Enumeration {
                                             tabs: Option[List[TopTab]],
                                             intervalTimeSettings: IntervalTimeSettings,
                                             testDataSettings: TestDataSettings,
-                                            redirectAfterArchive: Boolean
+                                            redirectAfterArchive: Boolean,
+                                            usageStatisticsReports: UsageStatisticsReportsSettings,
```
arkadius, line 102: Do you know what is the rule for choosing whether we should add sth directly to `UISettings` or to `ToggleFeaturesOptions`. `AnalyticsConfig` we have on top level.

arkadius (approved):
LGTM, one confusion pointed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
gadomsky: Unfortunetly I don't know.. maybe `analytics: Option[AnalyticsSettings]` should be moved to `ToggleFeaturesOptions` ?

gadomsky: but it's not relatetd to this change, so can you e.g. create issue for later and resolve this comment?

arkadius: Or everything should be flatten :D Let's leave it as it is ;)

PR #3819, opened 2022-12-14
Refs JSON Schemas in sinks

mproch:

### DIFF
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/SwaggerBasedJsonSchemaTypeDefinitionExtractor.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/SwaggerBasedJsonSchemaTypeDefinitionExtractor.scala
@@ -22,7 +22,7 @@ object SwaggerBasedJsonSchemaTypeDefinitionExtractor {
   // Extensions are all redundant elements in schema. This mechanism will work onl for limited usages,
   // some constructions described here: http://json-schema.org/understanding-json-schema/structuring.html
   // like anchors, recursive schemas, nested relative schemas won't work.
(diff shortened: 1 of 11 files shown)


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/SwaggerBasedJsonSchemaTypeDefinitionExtractor.scala
```diff
@@ -13,21 +13,17 @@ object SwaggerBasedJsonSchemaTypeDefinitionExtractor {
...
-  def swaggerType(schema: Schema): SwaggerTyped = {
+  def swaggerType(schema: Schema, parentSchema: Option[Schema] = None): SwaggerTyped = {
     val deserializedSchema: media.Schema[_] = OpenAPISchemaParser.parseSchema(schema.toString)
-    swaggerType(deserializedSchema)
-  }
-
-  def swaggerType(schema: media.Schema[_]): SwaggerTyped = {
-    val refSchemas = collectSchemaDefs(schema)
-    SwaggerTyped(schema, refSchemas)
+    SwaggerTyped(deserializedSchema, collectSchemaDefs(parentSchema.getOrElse(schema)))
```
gadomsky, line 18: if `parentSchema` is defined why we don't want to `collectSchemaDefs` also from `schema` and then merge them?

mproch: well, the idea is that the parent schema contains also refs from schema, but as it costs almost nothing I added also refs from schema as you suggest :+1:

gadomsky (approved):

PR #3824, opened 2022-12-14
Fix for Flink serialization of large schemas

mproch:


utils/schemed-kafka-components-utils/src/main/java/org/apache/flink/formats/avro/typeutils/NkSerializableParsedSchema.java
```diff
@@ -63,13 +63,13 @@ private void writeObject(ObjectOutputStream oos) throws IOException {
             oos.writeBoolean(true);
             if (schema instanceof AvroSchema) {
                 oos.writeByte(avroSchemaType);
-                oos.writeUTF(((AvroSchema) schema).rawSchema().toString(false));
+                oos.writeObject(((AvroSchema) schema).rawSchema().toString(false));
```
arkadius, line 66: Is it backward compatible or we should write something in MigrationGuide?

mproch: probably not, I'll added entry...

arkadius (approved):

PR #3826, opened 2022-12-19
The essence of the sliding window(docs)

ZbyszekMM:
The description of the aggregates misses the point when it comes to getting key differences of Nu aggregate components. 
The disappearance of the #input variable is presented a bit as a _side effect_; the fact that sliding window in its default configuration is an enricher is not mentioned.


docs/scenarios_authoring/AggregatesInTimeWindows.md
```diff
@@ -77,15 +70,15 @@ The result of the `groupBy` expression must be of type String.
...
+## Common behavior
+
+The sliding, tumbling and session window components differ among themselves not only in the way the time window is defined. A second key aspect which differentiates these components is how they `transform' input events into aggregate. This transformation can be either creation of a new (aggregate) event or an enrichment of incoming events with the value of the aggregate.  
+
+When the time window is closed, tumbling and session window nodes generate a new event containing an aggregate. Its timestamp is equal to the time of the timer that generated it, not system time of the moment when it happened. In other words the timestamp of the newly generated event which contains the aggregate will continue to use the notion of time used by events which the aggregate window saw. These nodes 'terminate' the events which entered them; consequently the #input and #inputMeta variables will not be available downstream. The sliding-window behaves in the same way in its **non-default** configuration when parameter `emitWhenEventLeft` is set to `true`.
```
mproch, line 77:
```suggestion
When the time window is closed, tumbling and session window nodes generate a new event containing an aggregate. Its timestamp is equal to the time of the timer that generated it, not system time of the moment when it happened. In other words the timestamp of the newly generated event which contains the aggregate will continue to use the notion of time used by events which the aggregate window saw. These nodes 'terminate' the events which entered them; consequently all the variables defined before the node (including `#input` and #inputMeta`) will not be available downstream. The sliding-window behaves in the same way in its **non-default** configuration when parameter `emitWhenEventLeft` is set to `true`. 
```

docs/scenarios_authoring/AggregatesInTimeWindows.md
```diff
@@ -77,15 +70,15 @@ The result of the `groupBy` expression must be of type String.
...
+
+When the time window is closed, tumbling and session window nodes generate a new event containing an aggregate. Its timestamp is equal to the time of the timer that generated it, not system time of the moment when it happened. In other words the timestamp of the newly generated event which contains the aggregate will continue to use the notion of time used by events which the aggregate window saw. These nodes 'terminate' the events which entered them; consequently the #input and #inputMeta variables will not be available downstream. The sliding-window behaves in the same way in its **non-default** configuration when parameter `emitWhenEventLeft` is set to `true`. 
+
+A sliding-window in its default configuration (`emitWhenEventLeft` is set to `false`) enriches the incoming event with the value of the aggregate. Because the original events are still available downstream, the #input and #inputMeta variables will be available 'downstream'.
```
mproch, line 79:
```suggestion
A sliding-window in its default configuration (`emitWhenEventLeft` is set to `false`) enriches the incoming event with the value of the aggregate. Because the original events are still available downstream, the `#input` and `#inputMeta` variables will be available 'downstream'.
```

mproch: I would move this paragraph above the one about tumbling/session. This mode (sliding, without emitWhenEventLeft) is used often and is IMHO simpler conceptually

docs/scenarios_authoring/AggregatesInTimeWindows.md
```diff
@@ -77,15 +70,15 @@ The result of the `groupBy` expression must be of type String.
...
 
-| aggregationType   | emitWhenEventLeft | variables                |
-|-------------------|-------------------|--------------------------|
-| sliding           | false             | #input, #inputMeta, #key |
-| sliding           | true              | #key                     |
-| tumbling, session | not configurable  | #key                     |
+A new #key variable will be available 'downstream' in all cases.
```
mproch, line 81: Maybe also mention about new output variable?

mproch, line 81:
```suggestion
A new `#key` variable will be available 'downstream' in all cases.
```

mproch (approved):

PR #3829, opened 2022-12-20
Terminate ActorSystem if initialization fails

mproch:


gadomsky (approved):
lgtm

designer/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerApp.scala
```diff
@@ -273,6 +272,8 @@ class NussknackerAppInitializer(baseUnresolvedConfig: Config) extends LazyLoggin
   def init(router: NusskanckerAppRouter): (Route, Iterable[AutoCloseable]) = {
     JavaClassVersionChecker.check()
     SLF4JBridgeHandlerRegistrar.register()
+    //we prepare temporary ExceptionHandler to shutdown actorSystem in case of exception during creation
+    prepareUncaughtExceptionHandler(Nil)
```
arkadius, line 276: Shouldn't be an actor system passed here? How does it work?

PR #3877, opened 2023-01-05
Bump Kafka to 3.3.1

mproch:


dswiecki (approved):
lgtm, which project do we need to update after that change ?

PR #3886, opened 2023-01-10
cross publish docker images

gadomsky:


build.sbt
```diff
@@ -341,6 +344,7 @@ lazy val commonDockerSettings = {
 
       List(dockerVersion, updateLatest, latestBranch, dockerTagName)
         .flatten
+        .flatMap(v => List(v, s"${v}_scala-${CrossVersion.binaryScalaVersion(scalaVersion.value)}"))
```
mproch, line 347: I think we should be more explicit on which scala version is the default. Currently e.g. scala version in `latest`, `staging-latest` etc. is determined by build order?

gadomsky: good point. now only build against `defaultScalaV` publishes images without `scalaV` suffix in tag.

mproch (approved):
:+1: Cool

PR #3893, opened 2023-01-11
Added diagnostics for flaky K8sDeploymentManagerReqRespTest + avoid i…

arkadius:
…dle connections to make sure that nginx doesn't talk with terminating instances


gadomsky (approved):
lgtm

PR #3914, opened 2023-01-16
Azure schema registry support

arkadius:

### DIFF
--- a/docs/integration/KafkaIntegration.md
+++ b/docs/integration/KafkaIntegration.md
@@ -24,23 +24,23 @@ Nussknacker integrates with Schema Registries. It is the source of knowledge abo
 - Which topics are available in Kafka sources and sinks
 - What schema versions are available for topic
 - How code completions and validations on types described by schemas should work
-- How messages will be deserialized and serialized
+- How messages will be serialized and deserialized
 
 Currently, Nussknacker supports two implementations of schema registries: based on *Confluent Schema Registry* and based on *Azure Schema Registry*. 
-During runtime, it deserializes and serializes messages in the way that is compatible with standard Kafka serializers and deserializers 
+During runtime, it serializes and deserializes messages in the way that is compatible with standard Kafka serializers and deserializers 
 delivered by those schema registries providers. Thanks to that you should be able to send messages to Nussknacker and read messages 
 produced by Nussknacker using standard tooling available around those Schema Registries.
 
-Given implementation is picked based on `schema.registry.url` property. By default, it is used Confluent-based implementation. 
-For urls ended with `.servicebus.windows.net` it is used Azure-based implementation.
+Given implementation is picked based on `schema.registry.url` property. By default, Confluent-based implementation is used. 
+For urls ended with `.servicebus.windows.net` Azure-based implementation is used.
 
-### Schema with topic association
+### Association between schema with topic
 
 To properly present information about topics and version and to recognize which schema is assigned to version, Nussknacker follow conventions:
 - For Confluent-based implementation it uses [TopicNameStrategy for subject names](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#subject-name-strategy).
   It means that it looks for schemas available at topic-name-(key or value). For example for topic transactions, it looks for schemas at transactions-key for key and transactions-value subject for value
 - In Azure Schema Registry, subjects concept doesn't exist - schemas are grouped by the same schema name. Because of that, Nussknacker introduces
-  own convention for schema with topic association: schema name should be in format: CamelCasedTopicNameKey for keys and CamelCasedTopicNameValue for values.
+  own convention for association between schema and topic: schema name should be in format: CamelCasedTopicNameKey for keys and CamelCasedTopicNameValue for values.
   For example for `input-events` topic, schema name should be named `InputEventsKey` for key or `InputEventsValue` for value. Be aware that it may require change of schema name
   not only in Azure portal but also inside schema content - those names should be the same to make serialization works correctly
   
@@ -60,7 +60,7 @@ Nussknacker supports schema evolution.
 For sources and sinks, you can choose which schema version should be used for syntax [suggestions and validation](/docs/integration/DataTypingAndSchemasHandling.md).
 
 At runtime Nussknacker determines the schema version of a message value and key in the following way:
-1. It checks in  `key.schemaId`, `value.schemaId` and Azure-specific `content-type` header;
+1. It checks in `key.schemaId`, `value.schemaId` and Azure-specific `content-type` headers;
 2. If there are no such headers, it looks for the magic byte (0x00) and a schema id in the message, [in a format used by Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format);
 3. If the magic byte is not found, it assumes the schema version chosen by the user in the scenario.
 
@@ -68,12 +68,12 @@ At runtime Nussknacker determines the schema version of a message value and key 
 
 Below you can find a quick comparison of how given schema registry types are handled:
 
-| Schema registry type | What is used for schema with topic association | Convention                                           | The way how schema id is passed in message                            |
-|----------------------|------------------------------------------------|------------------------------------------------------|-----------------------------------------------------------------------|
-| Confluent            | Subjects                                       | Subject = ${topic name}-(key or value)               | For Avro: in payload in format: 0x00, 4 bytes schema id, Avro payload |
-|                      |                                                |                                                      | For JSON: in  `key.schemaId` or `value.schemaId` headers              |
-| Azure                | Schema names                                   | Schema name = ${CamelCased topic name)(Key or Value) | For Avro: In header: content-type: avro/binary+schemaId               |
-|                      |                                                |                                                      | For JSON: not supported yet                                           |
+| Schema registry type | What is used for association between schema with topic | Convention                                        | The way how schema id is passed in message                            |
+|----------------------|--------------------------------------------------------|---------------------------------------------------|-----------------------------------------------------------------------|
+| Confluent            | Subjects                                               | Subject = <topic-name>-(key or value)             | For Avro: in payload in format: 0x00, 4 bytes schema id, Avro payload |
+|                      |                                                        |                                                   | For JSON: in  `key.schemaId` or `value.schemaId` headers              |
+| Azure                | Schema names                                           | Schema name = <CamelCasedTopicName>(Key or Value) | For Avro: In header: content-type: avro/binary+schemaId               |
+|                      |                                                        |                                                   | For JSON: not supported yet                                           |
 
 ## Configuration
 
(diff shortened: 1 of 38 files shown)


utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/azure/AzureSchemaRegistryClientFactory.scala
```diff
@@ -0,0 +1,166 @@
...
+  // We need to create our own schemas service because some operations like schema listing are not exposed by default client
+  // or even its Schemas inner class. Others like listing of versions are implemented incorrectly (it has wrong json field name in model)
+  private val enhancedSchemasService = new EnhancedSchemasImpl(
+    fullyQualifiedNamespace,
+    SchemaRegistryVersion.getLatest.getVersion,
+    httpPipeline,
+    SchemaRegistryJsonSerializer)
+
+  override def getSchemaById(id: SchemaId): SchemaWithMetadata = {
+    toSchemaWithMetada(schemaRegistryClient.getSchema(id.asString))
```
gadomsky, line 55: I think we need also `AvroUtils.adjustParsedSchema`  here

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaBasedSerdeProvider.scala
```diff
@@ -20,15 +22,18 @@ object UniversalSchemaBasedSerdeProvider {
...
+    createSchemaIdFromMessageExtractor(isConfluent, isAzure)
   }
 
   // SchemaId can be obtain in several ways. Precedent:
   // * from nu kafka headers - it is our own, Nussknacker headers standard format: key.schemaId and value.schemaId headers
+  // * from azure header - content-type: avro/binary+schemaId (only value schema ids are supported)
   // * from payload serialized in 'Confluent way' ([magicbyte][schemaid][payload])
   // * (fallback) from source editor version param - this is just an assumption we make (when processing no-schemed-data, everything can happen)
-  private[schemaregistry] def createSchemaIdFromMessageExtractor(isConfluent: Boolean): ChainedSchemaIdFromMessageExtractor = {
+  private[schemaregistry] def createSchemaIdFromMessageExtractor(isConfluent: Boolean, isAzure: Boolean): ChainedSchemaIdFromMessageExtractor = {
```
gadomsky, line 34: WDYT about one parameter -`providerType` which is enum or something, instead of multiple Boolean flags?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaRegistryClientFactory.scala
```diff
@@ -10,11 +11,12 @@ class UniversalSchemaRegistryClientFactory extends SchemaRegistryClientFactory {
 
   override type SchemaRegistryClientT = SchemaRegistryClient
 
-  override def create(config: SchemaRegistryClientKafkaConfig): SchemaRegistryClientT =
-    createConfluentFactory.create(config)
-
-  private def createConfluentFactory: SchemaRegistryClientFactory = {
-    CachedConfluentSchemaRegistryClientFactory
+  override def create(config: SchemaRegistryClientKafkaConfig): SchemaRegistryClientT = {
+    if (config.kafkaProperties.get("schema.registry.url").exists(_.endsWith(".servicebus.windows.net"))) {
```
gadomsky, line 15: now, one can't have SR behind proxy - maybe it's worth to write this in the docs?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/azure/AzureSchemaRegistryClientFactory.scala
arkadius: I would rather like remove this at all: https://github.com/TouK/nussknacker/pull/3968 WDYT? 70 LOC less, everything works like expected, no synthetic entries, no worry to remember to adjust schema everywhere, I see only benefits

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaBasedSerdeProvider.scala
arkadius: I like current approach because it handles 3rd option that we haven't got yet - not confluent format, not azure format - only our headers. For example we could easily add support of json schemas (or some other) kept on file system (we already have one question from our clients to handle such situation). Current approach will cause changes In only needed places - no need to think how to handle it - simple it would work - no confluent logic, no azure logic will happen, only our headers. But I understand why you don't like that. If you are very convicted to do such a change, I can do it, because it is rather small thing to change anytime

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaRegistryClientFactory.scala
arkadius: Good point, I wrote about that. If smb will have such case, we will introduce option to force specific schema registry implementation

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaBasedSerdeProvider.scala
gadomsky: No it's ok. I see a value in what you said

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
+- How code completions and validations on types described by schemas should work
+- How messages will be deserialized and serialized
+
+Currently, Nussknacker supports two implementations of schema registries: based on *Confluent Schema Registry* and based on *Azure Schema Registry*. 
+During runtime, it deserializes and serializes messages in the way that is compatible with standard Kafka serializers and deserializers 
+delivered by those schema registries providers. Thanks to that you should be able to send messages to Nussknacker and read messages 
+produced by Nussknacker using standard tooling available around those Schema Registries.
+
+Given implementation is picked based on `schema.registry.url` property. By default, it is used Confluent-based implementation. 
+For urls ended with `.servicebus.windows.net` it is used Azure-based implementation.
```
gadomsky, line 35:
```
By default, Confluent-based implementation is used. 
For urls ended with `.servicebus.windows.net` Azure-based implementation is used.
```
?

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
-
-Schemas are stored and managed by Confluent Schema Registry; it is [bundled with Nussknacker](/about/TypicalImplementation) in all deployment versions. Schemas can be registered in Schema Registry by means of REST API based CLI or using AKHQ, an open source GUI for Apache Kafka and Confluent Schema Registry. AKHQ is bundled with Nussknacker in all deployment versions.
-
-Nussknacker supports both JSON and Avro schemas, and JSON and Avro topic payloads. Detailed information on how Avro data should be serialized/deserialized can be found in [Confluent Wire Documentation](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format).
-
+Nussknacker integrates with Schema Registries. It is the source of knowledge about:
+- Which topics are available in Kafka sources and sinks
+- What schema versions are available for topic
+- How code completions and validations on types described by schemas should work
+- How messages will be deserialized and serialized
```
gadomsky, line 27: for me `serialized and deserialized ` sounds better, but it's a trifle, we can leave it as it is now.

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
+produced by Nussknacker using standard tooling available around those Schema Registries.
+
+Given implementation is picked based on `schema.registry.url` property. By default, it is used Confluent-based implementation. 
+For urls ended with `.servicebus.windows.net` it is used Azure-based implementation.
+
+### Schema with topic association
+
+To properly present information about topics and version and to recognize which schema is assigned to version, Nussknacker follow conventions:
+- For Confluent-based implementation it uses [TopicNameStrategy for subject names](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#subject-name-strategy).
+  It means that it looks for schemas available at topic-name-(key or value). For example for topic transactions, it looks for schemas at transactions-key for key and transactions-value subject for value
```
gadomsky, line 41: `<topic-name>-(key or value)` maybe we can mark somewhow, that topic-name it's a placeholder?

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
+- In Azure Schema Registry, subjects concept doesn't exist - schemas are grouped by the same schema name. Because of that, Nussknacker introduces
```
gadomsky, line 42: is that breakline in the middle of the sentence intended?

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
+  own convention for schema with topic association: schema name should be in format: CamelCasedTopicNameKey for keys and CamelCasedTopicNameValue for values.
```
gadomsky, line 43: schema and topic association?

docs/integration/KafkaIntegration.md
```diff
@@ -20,12 +20,30 @@ Kafka topics are native streaming data input to Nussknacker and the native outpu
...
+  For example for `input-events` topic, schema name should be named `InputEventsKey` for key or `InputEventsValue` for value. Be aware that it may require change of schema name
+  not only in Azure portal but also inside schema content - those names should be the same to make serialization works correctly
```
gadomsky, line 45: this sentence has also breakline in it

docs/integration/KafkaIntegration.md
```diff
@@ -42,11 +60,21 @@ Nussknacker supports schema evolution.
 For sources and sinks, you can choose which schema version should be used for syntax [suggestions and validation](/docs/integration/DataTypingAndSchemasHandling.md).
 
 At runtime Nussknacker determines the schema version of a message value and key in the following way:
-1. it checks `value.schemaId` and `key.schemaId` headers;
-2. if there are no such headers, it looks for a magic byte and schema version in the message, [in a format used by Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format);
-3. if the magic byte is not found, it assumes the schema version chosen by the user in the scenario.
+1. It checks in  `key.schemaId`, `value.schemaId` and Azure-specific `content-type` header;
```
gadomsky, line 63: * are those two spaces before `key.schemaId` intended?
* header -> headers?

gadomsky (approved):

docs/integration/KafkaIntegration.md
arkadius: placeholders in <> looks better :+1: . Regarding the placeholders description, I don't know how to do this in a way that it won't look too much complex.

docs/integration/KafkaIntegration.md
arkadius: I changed everywhere to "association between schema and topic" - I see that it is more commonly used

docs/integration/KafkaIntegration.md
gadomsky: For me it's understandable, since you provided an example. Altough I would add topic/subject names between ` ` : 

For example for topic `transactions`, it looks for schemas at `transactions-key` for key and `transactions-value` subject for value

PR #3916, opened 2023-01-17
Survey update

JulianWielga:


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
```diff
@@ -53,7 +53,7 @@ class SettingsResources(config: FeatureTogglesConfig,
 
 @JsonCodec case class RemoteEnvironmentConfig(targetEnvironmentId: String)
 
-@JsonCodec case class EnvironmentAlert(content: String, cssClass: String)
+@JsonCodec case class EnvironmentAlert(content: String, color: String)
```
dswiecki, line 56: I think we should add info about this in the changelog, it's incompatible change in case of the configuration

designer/client/containers/EnvironmentTag.tsx
```diff
@@ -0,0 +1,65 @@
...
+    backgroundColor: color,
+  })
+)
+
+export function EnvironmentTag() {
+  const {environmentAlert: {content, color} = {}} = useSelector(getFeatureSettings)
+  const background = useMemo(
+    () => {
+      switch (color) {
+        case EnvironmentTagColor.green:
```
arkadius, line 41: IMO user should provide e.g. "green" instead of "indicator-green". Setting is named "color" so as a user I would expect color, not CSS class

dswiecki (approved):

PR #3939, opened 2023-01-24
Update testing docs

dswiecki:

### DIFF
--- a/docs/integration/OpenAPI.md
+++ b/docs/integration/OpenAPI.md
@@ -72,6 +72,10 @@ components {
 
 ## Operations 
 
-You can enable logging invocations in the runtime by setting logger
+You can enable enricher level runtime logging by
 `pl.touk.nussknacker.openapi.enrichers.[enricher name]` to `DEBUG`. In particular, setting `pl.touk.nussknacker.openapi.enrichers` 
-level to `DEBUG` will turn on logging on all enrichers. Please remember to configure logging on appropriate component (e.g. Lite runtime or Flink TaskManager).
+level to `DEBUG` will turn on logging on all enrichers.
+
+Enricher level logging can be enabled:
+- in Flink TaskManager [configuration](https://github.com/TouK/nussknacker-quickstart/blob/main/docker/streaming/flink/log4j-console.properties)
+- in Lite runtime [configuration](../installation_configuration_guide/DeploymentManagerConfiguration#configuring-runtime-logging)
--- a/docs/integration/OpenAPI.md
+++ b/docs/integration/OpenAPI.md
@@ -22,17 +22,17 @@ of HTTP method, path and parameters (to provide some level of uniqueness)
 
 Table below describes data types that OpenAPI integration handles:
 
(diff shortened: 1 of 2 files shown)


gadomsky (approved):

arkadius (approved):

docs/integration/OpenAPI.md
```diff
@@ -74,4 +74,4 @@ components {
 
 You can enable logging invocations in the runtime by setting logger
 `pl.touk.nussknacker.openapi.enrichers.[enricher name]` to `DEBUG`. In particular, setting `pl.touk.nussknacker.openapi.enrichers` 
-level to `DEBUG` will turn on logging on all enrichers. Please remember to configure logging on appropriate component (e.g. Lite runtime or Flink TaskManager).
\ No newline at end of file
+level to `DEBUG` will turn on logging on all enrichers. Please remember to configure logging on appropriate component (e.g. Lite runtime or Flink TaskManager).
```
ZbyszekMM, line 77: Term "component" has a special meaning in Nu case, so I would avoid using it here. Do we mean engine? Additionally, instead of "pls remember", I would prefer some explanation which would enforce mental model of what is going on here. Example:

_For enriched level logging to work, engine  level logging has to be enabled_  (links to documentation or page titles or ... ) . 

I would also have this sentence at the beginning of this section (?), because it is a _prerequisite_, so adding this information at the end is not optimal to me.

docs/integration/OpenAPI.md
```diff
@@ -74,4 +74,4 @@ components {
 
 You can enable logging invocations in the runtime by setting logger
```
ZbyszekMM, line 75: Do we need this term _invocations_ here? 

Why not just _you can enable enricher level runtime logging by ...._  ?

docs/scenarios_authoring/TestingAndDebugging.md
```diff
@@ -4,18 +4,26 @@ sidebar_position: 8
 
 # Testing and Debugging
 
-There are several features in Nussknacker designed specifically to ease testing and debugging; this page provides a brief explanation how to use them. This functionality is grouped in the Test panel docked in  the right side of the Designer canvas; if you did not rearrange panels it is in the right part of the Designer canvas. 
+There are several features in Nussknacker designed specifically to ease testing and debugging; this page provides a brief explanation how to use them. This functionality is grouped in the Test panel docked on the right side of the Designer canvas; if you did not rearrange panels it is in the right part of the Designer canvas.
```
ZbyszekMM, line 7: last sentence contains repeated content; can be shortened (?) to:

This functionality is grouped in the Test panel;  if you did not rearrange panels it is in the right part of the Designer canvas.

docs/integration/OpenAPI.md
dswiecki: done

docs/integration/OpenAPI.md
dswiecki: done

PR #3945, opened 2023-01-25
Add endpoint to get category with processingType

Filemon279:

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
@@ -5,14 +5,11 @@ import akka.http.scaladsl.server.{Directives, Route}
 import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 import pl.touk.nussknacker.engine.{ModelData, ProcessingTypeData}
-import pl.touk.nussknacker.restmodel.process.ProcessingType
-import pl.touk.nussknacker.security.Permission.Read
 import pl.touk.nussknacker.ui.definition.UIProcessObjectsFactory
-import pl.touk.nussknacker.ui.process.ProcessCategoryService.Category
 import pl.touk.nussknacker.ui.process.processingtypedata.ProcessingTypeDataProvider
 import pl.touk.nussknacker.ui.process.{ProcessCategoryService, ProcessObjectsFinder}
 import pl.touk.nussknacker.ui.process.subprocess.SubprocessRepository
-import pl.touk.nussknacker.ui.security.api.{AdminUser, CommonUser, LoggedUser}
+import pl.touk.nussknacker.ui.security.api.LoggedUser
 import pl.touk.nussknacker.ui.util.EspPathMatchers
 
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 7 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -37,6 +40,18 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
           modelDataProvider.mapValues(_.processDefinition.services.mapValuesNow(UIProcessObjectsFactory.createUIObjectDefinition(_, processCategoryService))).all
         }
       }
+    } ~ path("processDefinitionData" / "categoriesWithProcessingType") {
+      get {
+        complete {
+          val categories = user match {
+            case user: CommonUser => processCategoryService.getAllCategories.filter(user.can(_, Read))
```
gadomsky, line 47: why not `processCategoryService.getUserCategories` ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -37,6 +40,18 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
...
+            case _: AdminUser => processCategoryService.getAllCategories
+          }
+          categories
+            .map(category => category -> processCategoryService.getTypeForCategory(category))
+            .toMap[Category, Option[ProcessingType]]
```
gadomsky, line 52: is it possible, that there is no ProcessingType (None) for a category?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
Filemon279: Didn't noticed that so I reinvented the wheel .. Thanks, changed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
Filemon279: API defines `getTypeForCategory` with `Option[ProcessingType]` so I didn't try to go deep into it. As I look at it now I think that in my case it always have processingType but I can see cases when this `getTypeForCategory` can return None. 

We could drop Option but it wouldn't make any difference. WDYT?

gadomsky: ok lets leave it

gadomsky (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -37,6 +40,14 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
           modelDataProvider.mapValues(_.processDefinition.services.mapValuesNow(UIProcessObjectsFactory.createUIObjectDefinition(_, processCategoryService))).all
         }
       }
+    } ~ path("processDefinitionData" / "categoriesWithProcessingType") {
+      get {
+        complete {
+            processCategoryService.getUserCategories(user)
```
lciolecki, line 46: Hmm, maybe better way is to create one method for userCategoriesType?

Filemon279: No problem'o - I've added `getUserCategoriesWithType` method and it helped me to drop `Option` from return type without any handlers. :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -37,6 +40,12 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
           modelDataProvider.mapValues(_.processDefinition.services.mapValuesNow(UIProcessObjectsFactory.createUIObjectDefinition(_, processCategoryService))).all
         }
       }
+    } ~ path("processDefinitionData" / "categoriesWithProcessingType") {
```
lciolecki, line 43: One thing more. This is not processDefinitionData, but just settings / config?

Filemon279: I've moved it next to the config endpoint.

lciolecki: Ofc, It will stay here forever :)

PR #3948, opened 2023-01-25
Kafka universal unconfluentization

arkadius:
- Using confluent abstractions only when needed
- Non-confluent classes renamed and moved to desired packages
- Removed some duplication of code in: universal record fromatter implementation and AvroSchemaEvolution
- Extracted new class: SchemaIdFromMessageExtractor to make Confluent logic explicit and moved to top level
- Extracted SchemaValidator to make Confluent logic explicit and be able to compose
- Some renames: ConsumerRecordUtils -> KafkaRecordUtils
- RecordDeserializer -> AvroRecordDeserializer (also inheritance replaced by composition)
- (De)SerializerFactory - easier abstractions
- Performance fix: UniversalKafkaSource now produce GenericRecordWithSchemaId
- AvroSerializerRegistrar fixes: wasn't checked if component is a kafka component, bad path was used (config.kafka instead of kafka)
- ConfluentSchemaRegistryFactory is not necessary now - removed
- ConfluentSchemaBasedSerdeProvider.universal replaced by UniversalSchemaBasedSerdeProvider.create


utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/confluent/schemaid/SchemaIdFromNuHeadersAndPotentiallyConfluentPayload.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.schemedkafka.schemaregistry.confluent.schemaid
+
+import org.apache.kafka.common.header.Headers
+import pl.touk.nussknacker.engine.schemedkafka.schemaregistry.{GetSchemaIdArgs, SchemaId, SchemaIdFromMessageExtractor, SchemaIdWithPositionedBuffer}
+import pl.touk.nussknacker.engine.schemedkafka.schemaregistry.confluent.ConfluentUtils
+
+import java.nio.ByteBuffer
+import scala.util.Try
+
+object SchemaIdFromNuHeadersAndPotentiallyConfluentPayload extends SchemaIdFromMessageExtractor with Serializable {
```
gadomsky, line 10: Maybe let's name it to `SchemaIdFromNuHeaders` and leave a comment, that even though we take schemaId from headers it can be also in payload - so bytearray must be shifted?


If you do not want to rename, I think a comment would be still welcomed, cause it's not so obvious.

arkadius: I'd rather like to keep confluent in name to make this Confluent logic more explicit but:
- I changed the name to be visible so that it only shifts payload - not extract schema id from it
- I added verbose comment why it is done this way
WDYT?

gadomsky: this name is a lot better, and a comment explains everything, thanks

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -31,10 +33,11 @@ sealed trait ParsedSchemaSupport[+S <: ParsedSchema] extends UniversalSchemaSupp
 }
 
 object AvroSchemaSupport extends ParsedSchemaSupport[AvroSchema] {
-  override val payloadDeserializer: UniversalSchemaPayloadDeserializer = ConfluentAvroPayloadDeserializer.default
+  override def payloadDeserializer(kafkaConfig: KafkaConfig): UniversalSchemaPayloadDeserializer = AvroPayloadDeserializer(kafkaConfig)
 
-  override def serializer[T](schema: ParsedSchema, client: ConfluentSchemaRegistryClient, kafkaConfig: KafkaConfig, isKey: Boolean): Serializer[T] =
-    ConfluentKafkaAvroSerializer(kafkaConfig, client, Some(schema.cast()), isKey = isKey).asInstanceOf[Serializer[T]]
+  override def serializer(schemaOpt: Option[ParsedSchema], client: SchemaRegistryClient, kafkaConfig: KafkaConfig, isKey: Boolean): Serializer[Any] = {
+    ConfluentKafkaAvroSerializer(kafkaConfig, client, schemaOpt.map(_.cast()), isKey = isKey)
```
gadomsky, line 39: shouldn't be confluent-agnostic (universal) serializer here?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -72,16 +82,17 @@ object JsonSchemaSupport extends ParsedSchemaSupport[OpenAPIJsonSchema] {
...
 
 object AvroSchemaWithJsonPayloadSupport extends ParsedSchemaSupport[AvroSchemaWithJsonPayload] {
 
-  override val payloadDeserializer: UniversalSchemaPayloadDeserializer = ConfluentJsonPayloadDeserializer
+  override def payloadDeserializer(k: KafkaConfig): UniversalSchemaPayloadDeserializer = JsonPayloadDeserializer
 
-  override def serializer[T](schema: ParsedSchema, client: ConfluentSchemaRegistryClient, kafkaConfig: KafkaConfig, isKey: Boolean): Serializer[T] =
-    new JsonPayloadKafkaSerializer(kafkaConfig, client, new DefaultAvroSchemaEvolution, Some(schema.cast().avroSchema), isKey = isKey).asInstanceOf[Serializer[T]]
+  override def serializer(schemaOpt: Option[ParsedSchema], client: SchemaRegistryClient, kafkaConfig: KafkaConfig, isKey: Boolean): Serializer[Any] =
+    new ConfluentJsonPayloadKafkaSerializer(kafkaConfig, client, new DefaultAvroSchemaEvolution, schemaOpt.map(_.cast().avroSchema), isKey = isKey)
```
gadomsky, line 95: and here also?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
arkadius: done

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
arkadius: done

engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/StateCompatibilityTest.scala
```diff
@@ -133,7 +133,7 @@ class StateCompatibilityTest extends FlinkWithKafkaSuite with Eventually with La
     })
   }
 
-  test("should restore from snapshot") {
+  ignore("should restore from snapshot") {
```
gadomsky, line 136: let's go back to as it was before

arkadius: woops

gadomsky (approved):
LGTM

PR #3958, opened 2023-01-26
Add content-type to openapi enrichers

dswiecki:

### DIFF
--- a/components/openapi/src/main/scala/pl/touk/nussknacker/openapi/parser/ParseToSwaggerService.scala
+++ b/components/openapi/src/main/scala/pl/touk/nussknacker/openapi/parser/ParseToSwaggerService.scala
@@ -47,7 +47,7 @@ private[parser] object ParseToSwaggerService {
 }
 
 private[parser] class ParseToSwaggerService(openapi: OpenAPI, openAPIsConfig: OpenAPIServicesConfig)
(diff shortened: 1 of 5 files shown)


components/openapi/src/main/scala/pl/touk/nussknacker/openapi/parser/ParseToSwaggerService.scala
```diff
@@ -138,7 +133,15 @@ private[parser] class ParseToSwaggerService(openapi: OpenAPI, openAPIsConfig: Op
       .map(_._2)
   }
 
-  private def requestParameter(request: RequestBody): ValidationResult[Option[SwaggerParameter]] = {
+  private def findContentType(content: Content): Option[String] = {
+    val scalaVersion = content.asScala
+    scalaVersion
+      .find(_._1.startsWith("application/json"))
+      .orElse(scalaVersion.find(_._1.contains("*/*")))
```
gadomsky, line 140: can we extract common part from this method and `findMediaType`?

dswiecki: done

gadomsky (approved):
LGTM

Issue #3954, opened 2023-01-26
Build Error: while running C:\NusskNacker\staging\nussknacker-staging\designer\runServer.sh


pokkireddy:
Build Error: while running C:\NusskNacker\staging\nussknacker-staging\designer\runServer.sh

Attached the log file for your reference. please let me know what is missed. 

**Environment (please fill in the following information):**
 - OS:  Ubuntu 18.04.6 LTS
 - Nussknacker version [staging]
 - Flink version
 - Java version:
 -openjdk 17.0.6 2023-01-17
  OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
  OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)
 - Scala version
 - Method of running of components [e.g. in Docker, lite, flink]
[](url)
[runServer.sh.log](https://github.com/TouK/nussknacker/files/10504850/runServer.sh.log)

gadomsky:
Hi @pokkireddy, thanks for reporting a issue. Looks like you're build is missing devModel (`<repo dir>/nussknacker-dist/target/universal/stage/managers/development-tests-manager.jar`)

If you build source using our scripts as described in Readme https://github.com/TouK/nussknacker/blob/staging/CONTRIBUTING.md#running-full-version-of-designer-from-command-line
 everything should work.
 
So in summary, the order of operations is as follows:
 * cloning fresh repo from repository
 * cd designer
 * ./buildClient.sh
 * ./buildServer.sh (please notice, that this script sets flag `addDevArtifacts` to `true` - I think, this is what you were missing) 
 * ./runServer.sh
 
 Please try to execute this instructions, and let us know if it had worked for you :)

pokkireddy:
[runServer_Error.txt](https://github.com/TouK/nussknacker/files/10510495/runServer_Error.txt)

I have followed the same steps, still getting error. 
I can able to see this jar ( (<repo dir>/nussknacker-dist/target/universal/stage/managers/development-tests-manager.jar) attached in the log at the end.

Please find the attached log.

My use case is I want to change some UI text for ex: Instead of scenarios we would to like to change something else
and build a docker image from this. Please let me know the steps for that.

pokkireddy:
Based on the fix, I can able to runServer.sh without any error.

My use case is I want to change some UI text for ex: Instead of scenarios we would to like to change something else
[...]

PR #3959, opened 2023-01-26
fix buildServer.sh script

gadomsky:


dswiecki:
fixes #3954

arkadius (approved):

arkadius:
> After deeper thinking it doesn't look like the way to go...
Our dist target already build all artifacts: `developmentTestsDeploymentManager` , `devModelArtifacts`, `componentArtifacts`

PR #3963, opened 2023-01-27
Filter services based on user permissions

Filemon279:

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
@@ -5,7 +5,7 @@ import akka.http.scaladsl.server.{Directives, Route}
 import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 import pl.touk.nussknacker.engine.{ModelData, ProcessingTypeData}
-import pl.touk.nussknacker.security.Permission.Read
+import pl.touk.nussknacker.security.Permission
 import pl.touk.nussknacker.ui.definition.UIProcessObjectsFactory
 import pl.touk.nussknacker.ui.process.processingtypedata.ProcessingTypeDataProvider
 import pl.touk.nussknacker.ui.process.{ProcessCategoryService, ProcessObjectsFinder}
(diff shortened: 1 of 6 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -34,7 +35,11 @@ class DefinitionResources(modelDataProvider: ProcessingTypeDataProvider[ModelDat
     } ~ path("processDefinitionData" / "services") {
       get {
         complete {
-          modelDataProvider.mapValues(_.processDefinition.services.mapValuesNow(UIProcessObjectsFactory.createUIObjectDefinition(_, processCategoryService))).all
+          modelDataProvider.mapValues(
+            _.processDefinition.services.filter {
+              case (_, definition) => definition.categories.forall(_.exists(user.can(_, Read)))
```
lciolecki, line 40: Why for all? One isn't enough to run the service?

lciolecki: Ane where are the tests?! :)

Filemon279: `forall` does not work like loop "for all elements". In this case, on `Option[List[String]]` type, it returns `true` if Option is `None` - otherwise it goes till it meets first `true` in given predicate. Take a look on REPL example

```
@ val optList = Option(List(1,2,3,4,5)).forall(_.exists(a => {println(a); a>1})) 
1
2
optList: Boolean = true
```

lciolecki: Ahh I thought in categories is List[String] not Option[List]] ;) It's allright! But still filtering should be done earlier :)

Filemon279: IMO if category is not set for service then we should return it - if somebody wants to hide some component, then this component should contain category for possibility to set correct access rights.

Filemon279: And I think that you thought same way, since you've done exactly same assumptions :) 
```
def availableForCategory(category: String): Boolean = categories.isEmpty || categories.exists(_.contains(category))
```

Filemon279: I've also added test case

lciolecki (approved):
LGTM

gadomsky (approved):
LGTM

PR #3976, opened 2023-02-06
fix 'prepareDev' sbt task

gadomsky:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1609,7 +1609,7 @@ lazy val prepareDev = taskKey[Unit]("Prepare components and model for running fr
 prepareDev := {
   val workTarget = (designer / baseDirectory).value / "work"
   val artifacts = componentArtifacts.value ++ devModelArtifacts.value ++ developmentTestsDeployManagerArtifacts.value ++
-    (if (addManagerArtifacts) managerArtifacts.value else Nil)
+    Def.taskIf(if (addManagerArtifacts) managerArtifacts.value else Nil).value
   IO.copy(artifacts.map { case (source, target) => (source, workTarget / target) })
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -1609,7 +1609,7 @@ lazy val prepareDev = taskKey[Unit]("Prepare components and model for running fr
 prepareDev := {
   val workTarget = (designer / baseDirectory).value / "work"
   val artifacts = componentArtifacts.value ++ devModelArtifacts.value ++ developmentTestsDeployManagerArtifacts.value ++
-    (if (addManagerArtifacts) managerArtifacts.value else Nil)
+    Def.taskIf(if (addManagerArtifacts) managerArtifacts.value else Nil).value
```
arkadius, line 1612: In the examples I've seen, additionally the whole task is wrapped with `Def.taskDyn`. Shouldn't be?

arkadius: Yeah I'm pretty sure, that it should be wrapped. Taka a look at: https://www.scala-sbt.org/1.x/docs/Howto-Dynamic-Task.html , Also it looks like `Def.taskIf` is some old name, and now it should be `Def.task`. At the end it is described: http://eed3si9n.com/selective-functor-in-sbt/.

gadomsky: Ok, I will look into this, but I have tested it before push, and as it is now, it evealuates task only when it's needed

gadomsky: Ok, please see the code now. Since `taskIf` is not documented (but it's also not deprecated, from what I see, at least in ours sbt version) let's use `taskDyn`

arkadius (approved):

PR #3984, opened 2023-02-08
Cleanup after scala 2.11

lciolecki:


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/TypingResultAwareTypeInformationDetection.scala
```diff
@@ -86,7 +86,7 @@ class TypingResultAwareTypeInformationDetection(customisation:
   }
 
   //we have def here, as Scala 2.11 has problems with serialization of PartialFunctions...
-  private def additionalTypeInfoDeterminer = customisation.customise(this)
+  private def additionalTypeInfoDeterminer: PartialFunction[TypingResult, TypeInformation[_]] = customisation.customise(this)
```
gadomsky, line 89: could we have `lazy val` now?

gadomsky (approved):
Cool, thanks

PR #3986, opened 2023-02-08
Upgrade sttp

piotrp:


components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
```diff
@@ -3,34 +3,31 @@ package pl.touk.nussknacker.openapi.http.backend
...
+class LoggingAndCollectingSttpBackend[+P](delegate: SttpBackend[Future, P], baseLoggerName: String)
+                                     (implicit ec: ExecutionContext, serviceInvocationCollector: ServiceInvocationCollector) extends DelegateSttpBackend[Future, P](delegate) {
 
   import LoggingAndCollectingSttpBackend._
 
   private val requestLogger = Logger(LoggerFactory.getLogger(s"$baseLoggerName.request"))
   private val responseLogger = Logger(LoggerFactory.getLogger(s"$baseLoggerName.response"))
 
-  override def send[T](request: Request[T, Nothing]): Future[Response[T]] = {
+  override def send[T, R >: P with Effect[Future]](request: Request[T, R]): Future[Response[T]] = {
```
arkadius, line 30: Wow. Can't we fix this R generic type in class declaration? It looks like our usages of this class are simple - we don't need to change effect type for each invocation of send

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
```diff
@@ -3,34 +3,31 @@ package pl.touk.nussknacker.openapi.http.backend
...
-
   private case class WithStringResponseBody[T](original: T, stringResponseBody: () => Option[String])
 
   private val transmissionNames = TransmissionNames("request", "response")
 
 }
 
-class LoggingAndCollectingSttpBackend(delegate: HttpBackend, baseLoggerName: String)
-                                     (implicit ec: ExecutionContext, serviceInvocationCollector: ServiceInvocationCollector) extends HttpBackend {
+class LoggingAndCollectingSttpBackend[+P](delegate: SttpBackend[Future, P], baseLoggerName: String)
```
arkadius, line 22: indent! ;)

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/SwaggerOpenApiDefinitionDiscovery.scala
```diff
@@ -44,7 +44,7 @@ class SwaggerOpenApiDefinitionDiscovery(implicit val httpBackend: SttpBackend[Fu
       FileUtils.readFileToString(new File(discoveryUrl.getPath), StandardCharsets.UTF_8)
     } else {
       Await
-        .result(basicRequest.get(Uri(discoveryUrl.toURI)).send(), 20 seconds)
+        .result(basicRequest.get(Uri(discoveryUrl.toURI)).send(httpBackend), 20 seconds)
```
arkadius, line 47: I see that now, backend must be passed explicitly. If their pick this approach, IMO we should remove implicit variables modifier. Otherwise it can cause confusions how this backend is used - why it is used implicitly and ath the same time, passed explicitly.

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
piotrp: Then we wouldn't be able to use `DelegateSttpBackend`, which implements `close` and `responseMonad` for us

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/SwaggerOpenApiDefinitionDiscovery.scala
piotrp: Yes, implicit SttpBackend passing is deprecated, but I'm not sure about passing it explicitly in our code - from what I see we do pass various contexts as implicits, so that would make Sttp special for us.

Also, I wanted this change to be as small as possible, it already touches enough files.

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
piotrp: Good catch :D

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
arkadius: I see now, o let's stay that way

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
arkadius: It was general notice, can you check other places? For example `handleResponse` in the same file... :P

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/SwaggerOpenApiDefinitionDiscovery.scala
arkadius: Ok, let it be that way

arkadius (approved):
LGTM, please fix indentions in other, not pointed places

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/http/backend/LoggingAndCollectingSttpBackend.scala
piotrp: Checked and fixed, looks like this was the last one

PR #3989, opened 2023-02-09
Use legacy scenarios submodule 

JulianWielga:


arkadius (approved):
For me this change contains a little bit to much refactors besides the main goal of extracting legacy scenarios tab. Because of that  I can't see what's going on, but let's merge it and let's see what's happen next :)

PR #3997, opened 2023-02-14
Remove subprocessVersions

gskrobisz:
Removal of deprecated field, see `//TODO: field subprocessVersions is deprecate - to remove`

### DIFF
--- a/common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
+++ b/common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
@@ -14,8 +14,7 @@ import scala.concurrent.duration.Duration
 // todo: MetaData should hold ProcessName as id
 @ConfiguredJsonCodec case class MetaData(id: String,
                                          typeSpecificData: TypeSpecificData,
-                                         additionalFields: Option[ProcessAdditionalFields] = None,
-                                         subprocessVersions: Map[String, Long] = Map.empty) { //TODO: field subprocessVersions is deprecate - to remove
+                                         additionalFields: Option[ProcessAdditionalFields] = None) {
   val isSubprocess: Boolean = typeSpecificData.isSubprocess
 }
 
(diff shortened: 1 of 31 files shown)


scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -33,9 +33,6 @@ class ProcessMetaDataBuilder private[build](metaData: MetaData) {
   def slug(p: Option[String]) =
     new ProcessMetaDataBuilder(metaData.copy(typeSpecificData = RequestResponseMetaData(p)))
 
-  def subprocessVersions(subprocessVersions: Map[String, Long]) =
```
arkadius, line 36: I think that we need a migration guide entry that this API method is unavailable anymore

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -14,8 +14,7 @@ import scala.concurrent.duration.Duration
 // todo: MetaData should hold ProcessName as id
 @ConfiguredJsonCodec case class MetaData(id: String,
                                          typeSpecificData: TypeSpecificData,
-                                         additionalFields: Option[ProcessAdditionalFields] = None,
-                                         subprocessVersions: Map[String, Long] = Map.empty) { //TODO: field subprocessVersions is deprecate - to remove
```
arkadius, line 18: This is an api module so migration guide entry is necessary

arkadius (commented):
1. Have you tested, that scenario jsons witch obsolete `subprocessVersions` field will be accepted by import mechanism or we check there if json doesn't contain redundant fields?
2. Similar question regarding database. migration is not need?

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
gskrobisz: Fixed.

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
gskrobisz: Fixed.

gskrobisz:
> 1. Have you tested, that scenario jsons witch obsolete `subprocessVersions` field will be accepted by import mechanism or we check there if json doesn't contain redundant fields?
Yes, json deserailzation ignores unknown fields (confirmed in test).

>     2. Similar question regarding database. migration is not need?
Database migration is not required. Obsolete fields are ignored when json is retrieved from repository.

arkadius (approved):
LGTM but let's product team add one more +1

PR #4022, opened 2023-02-22
Forward port:  Fix for: generate test data mechanism didn't work for json messages with defined schema id

arkadius:


gadomsky (approved):
LGTM

PR #4046, opened 2023-02-27
ProcessStateService merged into DeploymentService

arkadius:


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentManagerDispatcher.scala
```diff
@@ -16,4 +17,8 @@ class DeploymentManagerDispatcher(managers: ProcessingTypeDataProvider[Deploymen
     processRepository.fetchProcessingType(processId).map(managers.forTypeUnsafe)
```
gadomsky, line 17: maybe `.map(deploymentManager)` so logic is not duplicated?

gadomsky (approved):
one minor comment, but overall LGTM :)

PR #4065, opened 2023-03-03
TypingResult.primitiveClass for easier checking if numeric is assignable to other numeric (simpler logic of AvroSchemaOutputValidator)

arkadius:


gadomsky (approved):
LGTM

PR #4066, opened 2023-03-03
Add route returning state definitions aggregated for all processing types

gskrobisz:


extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -51,6 +52,7 @@ object ProcessActionType extends Enumeration {
 }
 
 object StateStatus {
+  type StateId = String
```
arkadius, line 55: Maybe Status, not state. We use "state" for naming status + additional information like allowed actions, deployed id, and version. Also if we don't change the method `name` name, maybe we should also use "name" here to avoid confusion.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStatusDefinitionService.scala
```diff
@@ -0,0 +1,43 @@
...
+
+class ProcessStatusDefinitionService(typeToConfig: ProcessingTypeDataProvider[ProcessingTypeData],
+                                     categoryService: ProcessCategoryService) {
+
+  def fetchStateDefinitions(): List[StateDefinition] = {
+    //gather state definitions for all processing types
+    val allProcessingTypeStateDefinitions = typeToConfig.all.toList.flatMap { case (processingType, processingTypeData) =>
+      val processingTypeCategories = categoryService.getProcessingTypeCategories(processingType).toSet
+      val stateDefinitionManager = processingTypeData.deploymentManager.processStateDefinitionManager
+      stateDefinitionManager.statusIds().map(name =>
```
arkadius, line 18: IMO would be nice to have one public method: `stateDefinitions()` inside  `ProcessStateDefinitionManager` and other methods like `statusDisplayableName` and `statusIcon` would be only a helper methods. The only problem I can see is that you would have a tuple (statedefinition, category) but maybe clean API, with less methods is more important?

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
gskrobisz: Fixed (but this is a bit confusing, I could have introduced not intentionally some more misleading names).

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStatusDefinitionService.scala
gskrobisz: Good point.
Added `stateDefinitions` to provide default set of states and their default properties. This "reversed" current behaviour of `ProcessStateDefinitionManager`: 

1. stateDefinitions - method returns default "fixed" set of states, used to display filtering options and used as defaults for `ProcessState` properties (which are displayed with status icon)
2. statusTooltip, statusIcon and statusDescription - by default those methods return values from "stateDefinitions", but you can override them to customize "dynamic" properties of state (which happens in e.g. `PeriodicProcessStateDefinitionManager` and other managers, that mechanism of "embelishments" does not change)

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessStateDefinitionManager.scala
```diff
@@ -8,14 +8,42 @@ import pl.touk.nussknacker.engine.deployment.ExternalDeploymentId
...
+/**
+  * Provides state definitions, state transitions and status property transformations for UI visualization.
+  */
 trait ProcessStateDefinitionManager {
+
+  /**
+    * Default set of state definitions that provides state defaults
+    * and allows filtering by fixed set of status IDs (see [[ProcessState]] and [[StateStatus.name]]).
+    */
+  def stateDefinitions(): Set[StateDefinition]
```
gskrobisz, line 20: Maybe it is better to keep definitions as `Map[StatusName, StateDefinition]` to enforce key-by-name uniquness.

designer/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerApp.scala
```diff
@@ -191,7 +193,9 @@ trait NusskanckerDefaultAppRouter extends NusskanckerAppRouter {
           processResolving = processResolving,
           processAuthorizer = processAuthorizer,
           processChangeListener = processChangeListener,
-          typeToConfig = typeToConfig
+          typeToConfig = typeToConfig,
+          categoryService = processCategoryService,
+          stateDefinitionService = statusDefinitionService
```
gskrobisz, line 198: @arkadius state? status?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStateDefinitionService.scala
```diff
@@ -0,0 +1,59 @@
...
+    val namesWithNonUniqueDefinitions = countDefinitionsForName
+      .collect { case (statusName, size) if size > 1 => statusName}
+
+    if (namesWithNonUniqueDefinitions.nonEmpty) {
+      throw new IllegalStateException(s"State definitions are not unique for states: ${namesWithNonUniqueDefinitions.mkString(", ")}")
+    }
+  }
+
+  def fetchStateDefinitions(): Set[UIStateDefinition] = {
+    validate()
```
gskrobisz, line 35: This validation seems to be in wrong place. @arkadius  should I check it on NussknackerApp start?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessStateDefinitionManager.scala
arkadius: I aggree

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
```diff
@@ -15,4 +18,116 @@ object SimpleStateStatus {
...
+    ProcessActionType.UnArchive -> SimpleStateStatus.NotDeployed
+  )
+
+  val statusActionsMap: Map[StateStatus, List[ProcessActionType]] = Map(
+    SimpleStateStatus.Unknown -> List(ProcessActionType.Deploy),
+    SimpleStateStatus.NotDeployed -> List(ProcessActionType.Deploy, ProcessActionType.Archive),
+    SimpleStateStatus.DuringDeploy -> List(ProcessActionType.Deploy, ProcessActionType.Cancel), // Deploy? see FlinkStateStatus
+    SimpleStateStatus.Running -> List(ProcessActionType.Cancel, ProcessActionType.Pause, ProcessActionType.Deploy),
+    SimpleStateStatus.Canceled -> List(ProcessActionType.Deploy, ProcessActionType.Archive),
+    SimpleStateStatus.Restarting -> List(ProcessActionType.Deploy, ProcessActionType.Cancel), // Deploy? see FlinkStateStatus
```
gskrobisz, line 37: Is it ok to allow Deploy action in Restarting or DuringDeploy status? 
It is not allowed by FlinkStateStatus and that seems to be a correct behaviour.
@arkadius wdyt?

designer/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerApp.scala
arkadius: You ask me if it should be stateDefinitionService or statusDefinitionService? I'm not sure, pick your best choice.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
arkadius: Regarding "Restarting" it means that something not ok happens on engine side. Ideally would be the best to allow override this "not ok" state of engine, but AFAIK it is problematic on Flink so user have to cancel the previous job and after that deploy next job. Summing up it looks like Flink's quirk that deploy is not allowed in the restarting state

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStatusDefinitionService.scala
arkadius: Uhh I see now that we hadn't understand correctly each other. By "one public method" I mean one new method instead of three new methods. IMO we shouldn't generalize this status definition so much. We need to have only name, displayableName and icon in this definition, putting tooltip and description there will be clumsy for "dynamic" statuses. What description you want to provide to `Scheduled(on Friday)` state with status.name =  "SCHEDULED"?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessStateDefinitionManager.scala
```diff
@@ -3,19 +3,48 @@ package pl.touk.nussknacker.engine.api.deployment
...
+/**
+  * Provides state definitions, state transitions and status property transformations for UI visualization.
+  */
 trait ProcessStateDefinitionManager {
+
+  /**
+    * Default set of state definitions that provides state defaults
+    * and allows filtering by fixed set of status IDs (see [[ProcessState]] and [[StateStatus.name]]).
+    */
+  def stateDefinitions(): Map[StatusName, StateDefinition]
```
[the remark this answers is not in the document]
gskrobisz: Fixed braces and comment.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStateDefinitionService.scala
gskrobisz: Moved validation to `ProcessingTypeDataReader.loadProcessingTypeData`. Also added some comments with explanation.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessStatusDefinitionService.scala
gskrobisz: After discussion: 1) we do not add another generalization but reusing current one 2) this approach eliminates several mistakes in state definitions 3) part of `StateDefinitionDetails` properties are used to build filtering options (displayableName, icon, tooltip) and some are used as default ProcessState properties (icon, tooltip, description) - hence misunderstanding.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessState.scala
```diff
@@ -65,31 +67,36 @@ object StateStatus {
...
 }
 
 // This status class is a walk around for fact that StateStatus is encoded and decoded. It causes that there is no easy option
 // to add own status with some specific fields without passing Encoders and Decoders to many places in application.
 // TODO: we should find places where StateStatuses are encoded and decoded and replace them with some DTOs for this purpose
-class CustomStateStatus(val name: String) extends StateStatus
+class CustomStateStatus(val name: StatusName) extends StateStatus
+
+@JsonCodec case class StateDefinition(displayableName: String,
+                                      icon: Option[URI],
```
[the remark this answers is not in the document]
gskrobisz: After discussion: we keep Options to preserve current approach that allows "undefined" icons and/or tooltips in UI.

arkadius (approved):
Good job, please extract FE part to separate change

PR #4077, opened 2023-03-08
fix database query invoker to be async

gadomsky:

### DIFF
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseEnricherInvoker.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseEnricherInvoker.scala
@@ -30,16 +30,15 @@ class DatabaseEnricherInvoker(query: String,
   override def invokeService(params: Map[String, Any])
                             (implicit ec: ExecutionContext, collector: ServiceInvocationCollector, contextId: ContextId, componentUseCase: ComponentUseCase): Future[queryExecutor.QueryResult] =
     getTimeMeasurement().measuring {
-      Future.successful {
-        queryDatabase(queryArgumentsExtractor(argsCount, params))
-      }
+      queryDatabase(queryArgumentsExtractor(argsCount, params))
     }
 
-  protected def queryDatabase(queryArguments: QueryArguments): queryExecutor.QueryResult =
+  protected def queryDatabase(queryArguments: QueryArguments)(implicit ec: ExecutionContext): Future[queryExecutor.QueryResult] = Future {
     withConnection(query) { statement =>
       setQueryArguments(statement, queryArguments)
       queryExecutor.execute(statement)
     }
+  }
 
   protected def setQueryArguments(statement: PreparedStatement, queryArguments: QueryArguments): Unit =
     queryArguments.value.foreach { arg =>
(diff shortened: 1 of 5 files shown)


components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseEnricherInvoker.scala
```diff
@@ -30,16 +30,15 @@ class DatabaseEnricherInvoker(query: String,
...
                             (implicit ec: ExecutionContext, collector: ServiceInvocationCollector, contextId: ContextId, componentUseCase: ComponentUseCase): Future[queryExecutor.QueryResult] =
     getTimeMeasurement().measuring {
-      Future.successful {
-        queryDatabase(queryArgumentsExtractor(argsCount, params))
-      }
+      queryDatabase(queryArgumentsExtractor(argsCount, params))
     }
 
-  protected def queryDatabase(queryArguments: QueryArguments): queryExecutor.QueryResult =
+  protected def queryDatabase(queryArguments: QueryArguments)(implicit ec: ExecutionContext): Future[queryExecutor.QueryResult] = Future {
```
arkadius, line 36: Can you move this `Future {` to separate line? I don't know if only I have that, but for me, this Future vanish on the end of line.

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseEnricherInvokerWithCache.scala
```diff
@@ -33,19 +33,20 @@ class DatabaseEnricherInvokerWithCache(query: String,
   // TODO: cache size
   private val cache = Caffeine.newBuilder()
     .expireAfterWrite(cacheTTL)
-    .build[CacheKey, CacheEntry[queryExecutor.QueryResult]]
+    .buildAsync[CacheKey, CacheEntry[queryExecutor.QueryResult]]
+
+  import scala.jdk.FutureConverters._
```
arkadius, line 38: You need to add some compat lib for that - it won't work on 2.12

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseEnricherInvokerWithCache.scala
```diff
@@ -33,19 +33,20 @@ class DatabaseEnricherInvokerWithCache(query: String,
   // TODO: cache size
   private val cache = Caffeine.newBuilder()
     .expireAfterWrite(cacheTTL)
-    .build[CacheKey, CacheEntry[queryExecutor.QueryResult]]
+    .buildAsync[CacheKey, CacheEntry[queryExecutor.QueryResult]]
```
dswiecki, line 36: maybe lets extend some Tests just to be sure that caching works when is should

gadomsky: it's actually testing a library.. but I've added test proving that it works

gadomsky: + it was already tested though - we do update on DB, and checking that enricher returns old value

gadomsky: or maybe you mean testing other behaviour? if so, which one?

arkadius (approved):
LGTM

PR #4080, opened 2023-03-08
Notifications using db

arkadius:


designer/server/src/main/resources/db/migration/hsql/V1_036__ProcessActionState.sql
```diff
@@ -8,6 +8,7 @@ ALTER TABLE "process_actions" ALTER COLUMN "process_version_id" SET NULL;
 ALTER TABLE "process_actions" ADD COLUMN "id" UUID;
 ALTER TABLE "process_actions" ADD COLUMN "state" VARCHAR(16);
 ALTER TABLE "process_actions" ADD COLUMN "created_at" TIMESTAMP;
+ALTER TABLE "process_actions" ADD COLUMN "failure" VARCHAR(1022);
```
lciolecki, line 11: failure_message? And I think this field shouldn't be empty when state is FAILD?

lciolecki (approved):
LGTM, nice job!

designer/server/src/main/resources/db/migration/hsql/V1_036__ProcessActionState.sql
arkadius: failure_message -> ok
regarding the second point, I see know that I handle in the wrong way situation when getMessage return null - it will be NPE now. I'll change to save empty string than. But I think that I can't add constraint that this message is not null for failures because some DBs treat empty strings as nulls. Am I right?

lciolecki: Hmm, okey -> if getMessage can return null, we should allow put here null as well. But without any cropping?

arkadius: Ok, I'll fix it that way

PR #4087, opened 2023-03-09
fixed filter/query loop on screnarios

JulianWielga:

### DIFF
--- a/designer/submodules/packages/components/src/components/rootRoutes.tsx
+++ b/designer/submodules/packages/components/src/components/rootRoutes.tsx
@@ -9,7 +9,8 @@ export function RootRoutes(): JSX.Element {
         <Routes>
             <Route index element={<Navigate to="/scenarios" />} />
             <Route path="scenarios" element={<ScenariosView />} />
-            <Route path="components" element={<ComponentsRoutes />} />
+            <Route path="scenarios/table" element={<ScenariosView table />} />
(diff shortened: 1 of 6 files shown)


designer/submodules/packages/components/src/scenarios/scenariosView.tsx
```diff
@@ -82,10 +81,12 @@ function Scenarios({ children, table }: PropsWithChildren<{ table?: boolean }>):
     );
 }
 
-export type ScenariosViewProps = ActionsProps;
+export interface ScenariosViewProps extends ActionsProps {
+    table?: boolean;
```
SpOOnman, line 85: This is fine

designer/submodules/packages/components/src/components/rootRoutes.tsx
```diff
@@ -9,7 +9,8 @@ export function RootRoutes(): JSX.Element {
         <Routes>
             <Route index element={<Navigate to="/scenarios" />} />
             <Route path="scenarios" element={<ScenariosView />} />
-            <Route path="components" element={<ComponentsRoutes />} />
+            <Route path="scenarios/table" element={<ScenariosView table />} />
```
SpOOnman, line 12: I think scenarios with table should stay hidden by default? Was it only for testing?

SpOOnman (changes requested):

SpOOnman (approved):

PR #4091, opened 2023-03-10
A bit clearer structure of the security config

ZbyszekMM:

### DIFF
--- a/docs/installation_configuration_guide/DesignerConfiguration.md
+++ b/docs/installation_configuration_guide/DesignerConfiguration.md
@@ -3,7 +3,7 @@ sidebar_position: 4
 ---
 # Designer configuration
 
-Designer configuration area contains configs for web application interface, security, various UI settings, database and more. Check [configuration areas](./#configuration-areas) to understand where Designer configuration should be
+Designer configuration area contains configs for web application interface, security, various UI settings, database and more. Check [configuration areas](./Common.md#configuration-areas)) to understand where Designer configuration should be
 placed in the Nussknacker configuration.
 
 The default Designer configuration can be found in [defaultDesignerConfig.conf](https://github.com/TouK/nussknacker/blob/staging/designer/server/src/main/resources/defaultDesignerConfig.conf).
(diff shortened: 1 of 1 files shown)


docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -111,13 +114,15 @@ See [development configuration](https://github.com/TouK/nussknacker/blob/staging
 
 ## Security
 
-Nussknacker has pluggable security architecture - by default we support two type of authentication: BasicAuth and
-OAuth2. You can either use default authentication provider, based on Basic authentication and static user configuration
```
ZbyszekMM, line 115: This second sentence (_you can either use default authentication provider, based on Basic Auth ..._ does not make sense in the context of having 2 (3) providers. It is probably from the old times when there was only BasicAuth supported. Alternatively, I did not understand something, hence this comment.

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -147,6 +154,10 @@ Currently supported permissions:
 
 #### Users' file format:
 
+The association of the users to the roles is in the users' configuration file; in the case of OIDC it can additionally be supplemented by  the list of roles provided by the OpenId provider.
```
dswiecki, line 157: by  the list > by the list

dswiecki (approved):

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -3,7 +3,10 @@ sidebar_position: 4
 ---
 # Designer configuration
 
-The default Designer configuration is defined in [defaultDesignerConfig.conf](https://github.com/TouK/nussknacker/blob/staging/designer/server/src/main/resources/defaultDesignerConfig.conf).
+Designer configuration area contains configs for web application interface, security, various UI settings, database and more. Check [configuration areas](./#configuration-areas) to understand where Designer configuration should be
```
gadomsky, line 6: `(./#configuration-areas)` is it correct link?

ZbyszekMM: corrected to (./Common.md#configuration-area) as  it is [here](https://github.com/TouK/nussknacker/blob/staging/docs/installation_configuration_guide/DeploymentManagerConfiguration.md?plain=1#L9). 

Originally I used stale commit ....

PR #4096, opened 2023-03-10
Fix for NPE during handling int schemas without ranges on scala 2.12

arkadius:

### DIFF
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
@@ -199,8 +199,8 @@ object SwaggerTyped {
 
   private def inferredIntType(minValue: java.math.BigDecimal, exclusiveMinValue: java.math.BigDecimal, maxValue: java.math.BigDecimal, exclusiveMaxValue: java.math.BigDecimal): SwaggerTyped = {
 
-    val lowerBoundary: Option[BigDecimal] = List(Option(exclusiveMinValue).map(_.add(java.math.BigDecimal.valueOf(1))), Option(minValue)).flatten.map(bd => bd: BigDecimal).sorted(Ordering.BigDecimal.reverse).headOption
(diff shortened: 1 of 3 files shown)


vder (commented):
za dużo o jedną '1' mozna po prostu dać java.math.BigDecimal.ONE

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -197,10 +197,10 @@ object SwaggerTyped {
     }
   }
 
-  private def inferredIntType(minValue: BigDecimal, exclusiveMinValue: BigDecimal, maxValue: BigDecimal, exclusiveMaxValue: BigDecimal): SwaggerTyped = {
+  private def inferredIntType(minValue: java.math.BigDecimal, exclusiveMinValue: java.math.BigDecimal, maxValue: java.math.BigDecimal, exclusiveMaxValue: java.math.BigDecimal): SwaggerTyped = {
 
-    val lowerBoundary: Option[BigDecimal] = List(Option(exclusiveMinValue).map(_ + 1), Option(minValue)).flatten.sorted(Ordering.BigDecimal.reverse).headOption
-    val upperBoundary: Option[BigDecimal] = List(Option(exclusiveMaxValue).map(_ - 1), Option(maxValue)).flatten.sorted(Ordering.BigDecimal).headOption
+    val lowerBoundary: Option[BigDecimal] = List(Option(exclusiveMinValue).map(_.add(java.math.BigDecimal.valueOf(1))), Option(minValue)).flatten.map(bd => bd: BigDecimal).sorted(Ordering.BigDecimal.reverse).headOption
+    val upperBoundary: Option[BigDecimal] = List(Option(exclusiveMaxValue).map(_.add(java.math.BigDecimal.valueOf(-11))), Option(maxValue)).flatten.map(bd => bd: BigDecimal).sorted(Ordering.BigDecimal).headOption
```
vder, line 203: too many ones here -  probably its better to use java.math.BigDecimal.One instead

arkadius: valueOf is optimized for <0, 10> but you've right - it will be more visible that there is no unexpected overhead here. I also changed add -1 to subtract 1 - I couldn't find "minus" method in initial change :)

PR #4099, opened 2023-03-10
Release workflow

SpOOnman:


.github/workflows/release.yml
```diff
@@ -12,12 +12,91 @@ on:
...
+
+      - name: "Build"
+        env:
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USER }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+        run: sbt 'release ${{ env.SBT_RELEASE_NEXT_VERSION }} skip-tests'
+
+      - name: "Push to master"
+        if: ${{ env.NU_DOCKER_UPDATE_LATEST == true }}
+        run: echo "TBD git push origin HEAD:master -f && ./dockerhub/pulishReadme.sh"
```
arkadius, line 94: What is the plan for this?

.github/workflows/release.yml
```diff
@@ -12,12 +12,91 @@ on:
...
 
 jobs:
-  setup:
+  build:
     runs-on: ubuntu-latest
     env:
+      SBT_RELEASE_NEXT_VERSION: ${{ (inputs.release_version && format('{0} {1} with-defaults', 'release-version', inputs.release_version)) || 'with-defaults' }}
       NU_DOCKER_UPDATE_LATEST: ${{ inputs.backport_release == false && inputs.release_candidate == false }}
+      JAVA_OPTS: "-Xss6M -Xms4G -Xmx4G -XX:ReservedCodeCacheSize=256M -XX:MaxMetaspaceSize=2500M -DdockerUpLatest=false"
+      JVM_OPTS: "-Xss6M -Xms4G -Xmx4G -XX:ReservedCodeCacheSize=256M -XX:MaxMetaspaceSize=2500M -DdockerUpLatest=false"
```
arkadius, line 27: Do we need both env variables?

SpOOnman: You're right, one should be enough.

.github/workflows/release.yml
SpOOnman: This was taken from manual to update master branch. I'm not sure it should be automatic. What do you think?

build.sbt
```diff
@@ -95,10 +95,11 @@ lazy val publishSettings = Seq(
   },
   organization := "pl.touk.nussknacker",
   homepage := Some(url(s"https://github.com/touk/nussknacker")),
-  credentials := nexusHostFromProps.map(host => Credentials("Sonatype Nexus Repository Manager",
-    host, propOrEnv("nexusUser", "touk"), propOrEnv("nexusPassword", null))
-    // otherwise ~/.sbt/1.0/sonatype.sbt will be used
-  ).toSeq
+  // This didn't work with GitHub Actions release. Without this credentials are taken from env properly.
```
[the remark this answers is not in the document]
SpOOnman: Ok, thanks!

.github/workflows/release.yml
arkadius: I think, that we should automate this as well - because master is changed only by this workflow. We shouldn't be afraid of pushing to repo, we already do this with release branch and tag.

.github/workflows/release.yml
```diff
@@ -12,12 +12,120 @@ on:
...
+        type: string
 
 jobs:
-  setup:
+  build:
     runs-on: ubuntu-latest
     env:
+      SBT_RELEASE_NEXT_VERSION: ${{ (inputs.release_version && format('{0} {1} with-defaults', 'release-version', inputs.release_version)) || 'with-defaults' }}
       NU_DOCKER_UPDATE_LATEST: ${{ inputs.backport_release == false && inputs.release_candidate == false }}
+      JVM_OPTS: "-Xss6M -Xms4G -Xmx4G -XX:ReservedCodeCacheSize=256M -XX:MaxMetaspaceSize=2500M -DdockerUpLatest=false"
```
gadomsky, line 26: 1. Can we have this variable only on relese a.k. `name: "Build"` job? 
2. why `-DdockerUpLatest` is hardcoded?  shouldn't be passed `NU_DOCKER_UPDATE_LATEST` there?

.github/workflows/release.yml
```diff
@@ -12,12 +12,120 @@ on:
...
+      - name: "Push to master"
+        if: env.NU_DOCKER_UPDATE_LATEST == true
+        run: git push origin HEAD:master -f
+
+      - name: "Read nussknacker short description from file"
+        if: env.NU_DOCKER_UPDATE_LATEST == true
+        id: short_nu
+        run: echo "::set-output name=short::$(cat dockerhub/nussknacker/short.txt)"
+
+      - name: "Dockerhub publish readme nussknacker XD"
```
gadomsky, line 100: `XD`? XD

.github/workflows/release.yml
arkadius: good catch

.github/workflows/release.yml
SpOOnman: Sorry, that was for testing XD

version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "1.9.0-SNAPSHOT"
+version in ThisBuild := "1.8.2-SNAPSHOT"
```
gadomsky, line 1: let's bring back previous version

SpOOnman: Done

SpOOnman:
Git history is broken here. Continues in #4139

PR #4100, opened 2023-03-10
Presenting in progress actions on scenario list

arkadius:
+ refactor: now only DeploymentService is responsible for scenario state computation. Before the change, this logic was in resource class

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -4,10 +4,12 @@ import akka.http.scaladsl.marshalling.ToResponseMarshallable
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.server._
 import akka.stream.Materializer
+import cats.instances.future._
+import cats.instances.list._
+import cats.syntax.traverse._
 import com.typesafe.scalalogging.LazyLogging
 import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport
-import pl.touk.nussknacker.engine.ProcessingTypeData
-import pl.touk.nussknacker.engine.api.deployment.{DeploymentManager, ProcessState}
+import pl.touk.nussknacker.engine.api.deployment.ProcessState
 import pl.touk.nussknacker.engine.api.process.{ProcessId, ProcessName, VersionId}
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
 import pl.touk.nussknacker.engine.util.Implicits._
@@ -26,7 +28,6 @@ import pl.touk.nussknacker.ui.process.ProcessService.{CreateProcessCommand, Upda
 import pl.touk.nussknacker.ui.process._
 import pl.touk.nussknacker.ui.process.deployment.DeploymentService
 import pl.touk.nussknacker.ui.process.marshall.ProcessConverter
-import pl.touk.nussknacker.ui.process.processingtypedata.ProcessingTypeDataProvider
 import pl.touk.nussknacker.ui.process.repository.FetchingProcessRepository
 import pl.touk.nussknacker.ui.process.repository.FetchingProcessRepository.FetchProcessesDetailsQuery
 import pl.touk.nussknacker.ui.security.api.LoggedUser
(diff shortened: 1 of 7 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -31,6 +32,12 @@ trait ProcessStateService {
   def getProcessState(processIdWithName: ProcessIdWithName)
                      (implicit user: LoggedUser, ec: ExecutionContext): Future[ProcessState]
 
+  // This method in contrary to getProcessState doesn't invoke target DeploymentManager - it only compute state
+  // based on information available in DB
+  // TODO: add caching of state returned by DeploymentManager
+  def getDbProcessState(processDetails: BaseProcessDetails[_])
```
lciolecki, line 38: Why there is Db in the method name? I think it's just technical details?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -260,22 +260,14 @@ class ProcessesResources(
...
-  //This is temporary function to enriching process state data
-  //TODO: Remove it when we will support cache for state
-  private def enrichDetailsWithProcessState[PS: ProcessShapeFetchStrategy](process: BaseProcessDetails[PS]): BaseProcessDetails[PS] =
-    process.copy(state = deploymentManager(process.processingType).map(m => m.processStateDefinitionManager.processState(
-      m.processStateDefinitionManager.mapActionToStatus(process.lastAction.map(_.action))
-    )))
-
-  private def deploymentManager(processingType: ProcessingType): Option[DeploymentManager] =
-    typeToConfig.forType(processingType).map(_.deploymentManager)
+  private def enrichDetailsWithProcessState[PS: ProcessShapeFetchStrategy](process: BaseProcessDetails[PS])
```
lciolecki, line 267: Maybe it's time to move this method to the DeploymentService / some state service?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
arkadius: I was looking for some better word but I couldn't find it. Maybe you have some better candidate? Currently, we have:
1. "real state" based on remote state on engine + local 
2. only local state

local state - not local for designer, but local in our database. I wanted to distinguish between them so I call the second one "Db" + add this comment. So in fact, all we do here is about technical, performance details :)

 I want to introduce caching in next PR, so this name probably will be changed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
arkadius: I think that now it can be just inlined. I don't want to put the logic of "orchestration" of fetching process from db + enriching into DeploymentService. Let it be responsible only for executing actions and providing states. Maybe when this orchestration logic will be too big, we can introduce some new service

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
lciolecki: `getInitialProcessState` - as I understood, it is state before we call the manager?

lciolecki (approved):
LGTM, it's a good direction.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
arkadius: nope it is totally different state, it is not "initial" state. "initial" is rather "not deployed". It is local db state, without knowledge from remote engine (without calling DM)

gadomsky: `getInternalProcessState`? But i am not sure also.. For me `getDbProcessState` might stay

gadomsky (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
lciolecki: I prefer Internal rather than DB... But still, we compute here the initial state from our storage (in this case it's the DB).

PR #4102, opened 2023-03-10
Flink - deploy waitForDuringDeployFinished.

arkadius:
Thanks to that comments, activities and state are refreshed "at once"


docs/installation_configuration_guide/DeploymentManagerConfiguration.md
```diff
@@ -292,9 +292,11 @@ Deployment Manager of type `lite-embedded` has the following configuration optio
...
+| shouldVerifyBeforeDeploy            | boolean  | true          | By default, before redeployment of scenario with state from savepoint, verification of savepoint compatibility is performed. There are some cases when it can be too time consuming or not possible. Use this flag to disable it. |
+| shouldCheckAvailableSlots           | boolean  | true          | When set to true, Nussknacker checks if there are free slots to run new job. This check should be disabled on Flink Kubernetes Native deployments, where Taskmanager is started on demand.                                        |
+| waitForDuringDeployFinish.maxChecks | boolean  | 60            | After Flink job execution, we check if tasks were successfully started on TaskMangers. This parameter describe how many times we should check it before notifying about deployment failure.                                       |
+| waitForDuringDeployFinish.delay     | boolean  | 1 second      | After Flink job execution, we check if tasks were successfully started on TaskMangers. This parameter describe how long should be delay between checks.                                                                           |
```
lciolecki, line 302: Isn't it too small? Maybe 5/10s between each check?

lciolecki (approved):
Nice! :)

docs/installation_configuration_guide/DeploymentManagerConfiguration.md
arkadius: I'm not sure about that number - I would like to provide to user the quickest response. Maybe we can check how it will work and increase it if it will be not ok. WDYT?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkConfig.scala
```diff
@@ -14,4 +14,8 @@ import scala.concurrent.duration.DurationInt
 case class FlinkConfig(restUrl: String,
                        jobManagerTimeout: FiniteDuration = 1 minute,
                        shouldVerifyBeforeDeploy: Boolean = true,
-                       shouldCheckAvailableSlots: Boolean = true)
+                       shouldCheckAvailableSlots: Boolean = true,
+                       waitForDuringDeployFinish: Option[FlinkWaitForDuringDeployFinishedConfig] =
+                       Some(FlinkWaitForDuringDeployFinishedConfig(60, 1 second)))
```
gadomsky, line 19: Now to turn it off in config, one has to write:
```
waitForDuringDeployFinish {} 
```
?
It looks a bit weird.
Maybe it's better to introduce `waitForDuringDeployFinish.enabled` flag or get rid of `Option` ? So this checking behaviour is required.

arkadius: You're right - it can be inconvenient. I added enabled option

gadomsky (approved):
LGTM, thanks!

PR #4112, opened 2023-03-15
unique name for test components providers

gadomsky:
work around to bug in intelliJ which causes that wrong classpath is provided when runining tests from intelliJ


arkadius (approved):
It's ok but I'd change the description. It is not a solution but only a work around. The real solution would cause that only one mock kafka components implementation should be available on classpath

PR #3031, opened 2022-04-25
Store attachments in DB instead of on disk

gadomsky:

PR #4116, opened 2023-03-15
Obsolete attachmentsPath configuration property removal

arkadius:


gadomsky (approved):
Do we need MigrationGuide for that?
Anyway, LGTM

arkadius:
> Do we need MigrationGuide for that?
No, it was only a garbage (dead, unused configuration part). MigrationGuide should be provided in the change that was moving attachments from disk to DB: https://github.com/TouK/nussknacker/pull/3031 It is only the final step that was missing in the previous change

PR #4133, opened 2023-03-20
short git sha in NU version

gadomsky:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -58,7 +58,7 @@ jobs:
         run: |
           GIT_SOURCE_BRANCH=`([ "${GITHUB_HEAD_REF}" != "" ] && echo "${GITHUB_HEAD_REF}" || echo "${GITHUB_REF}") | sed 's/refs\/heads\///g'`
           SANITIZED_BRANCH=`echo ${GIT_SOURCE_BRANCH} | sed 's/[^a-zA-Z0-9._-]/\_/g' | awk '{print tolower($0)}'`
-          VERSION_SUFFIX="-$SANITIZED_BRANCH-$(date -I)-$GITHUB_RUN_NUMBER-$GITHUB_SHA"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: |
           GIT_SOURCE_BRANCH=`([ "${GITHUB_HEAD_REF}" != "" ] && echo "${GITHUB_HEAD_REF}" || echo "${GITHUB_REF}") | sed 's/refs\/heads\///g'`
           SANITIZED_BRANCH=`echo ${GIT_SOURCE_BRANCH} | sed 's/[^a-zA-Z0-9._-]/\_/g' | awk '{print tolower($0)}'`
-          VERSION_SUFFIX="-$SANITIZED_BRANCH-$(date -I)-$GITHUB_RUN_NUMBER-$GITHUB_SHA"
+          VERSION_SUFFIX="-$SANITIZED_BRANCH-$(date -I)-$GITHUB_RUN_NUMBER-${GITHUB_SHA::7}"
```
arkadius, line 61: `git rev-parse --short HEAD` returns 8 chars instead of 7. Let's stick to this standard convention.

gadomsky: on my machine it returns 9...
I gues its computed based on repository objects count..
https://www.git-scm.com/docs/git-config/2.12.5#Documentation/git-config.txt-coreabbrev

gadomsky: i changed to 9 but now thinking about setting it to 7 again, since github 
image
shows only 7 chars

arkadius: Yeah, you've right - it is 9, I incorrectly checked it. I forgot about github convention. IMO will be better to take max from both conventions ;)

arkadius (approved):

.github/workflows/pr.yml
gadomsky: ok, let's leave 9, but as repository grows it might be 10

PR #4135, opened 2023-03-20
Limit process state fetching to reasonable timeout

arkadius:

### DIFF
--- a/utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
+++ b/utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
@@ -10,9 +10,9 @@ import scala.concurrent.duration._
 trait BasePatientScalaFutures extends ScalaFutures with Eventually {
 
   implicit class FutureOps[T](future: Future[T]) {
-    // Sometimes we know how long should take time after which some Future will be completed. In this case
-    // we want to wait a little big longer than this time, to make sure that in case of Failure, inner Failure won't
-    // be suppressed by exception "A timeout occurred waiting for a future to complete" and extract this inner exception
+    // Sometimes we know the maximum time after which a particular Future will complete. In this case,
+    // we want to wait a little longer, to make sure that in case of a Failure, the inner Failure won't
+    // be suppressed by exception "A timeout occurred waiting for a future to complete"
     def futureValueEnsuringInnerException(innerOperationDuration: FiniteDuration): T = {
       try {
         future.futureValue(Timeout(innerOperationDuration.plus(patienceConfig.timeout)))


utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.scalatest.concurrent.{Eventually, ScalaFutures}
+import org.scalatest.exceptions.TestFailedException
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+trait BasePatientScalaFutures extends ScalaFutures with Eventually {
+
+  implicit class FutureOps[T](future: Future[T]) {
+    // Sometimes we know how long should take time after which some Future will be completed. In this case
```
gadomsky, line 13:
```suggestion
    // Sometimes we know how long we should wait after a Future will be completed. In this case
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -266,15 +269,34 @@ class DeploymentServiceImpl(dispatcher: DeploymentManagerDispatcher,
...
     }
 
+    scenarioStateTimeout.map { timeout =>
+      Future.firstCompletedOf(Seq(
+        stateFromEngineFuture,
+        akka.pattern.after(timeout) {
+          if (!statusFetchCompleted) {
+            logger.warn(s"Timeout: $timeout occurred during waiting for response from engine for ${processIdWithName.name}")
+          }
+          Future.successful(failedToGetResult)
```
gadomsky, line 295:
```scala
if (!statusFetchCompleted) {
  logger.warn(s"Timeout: $timeout occurred during waiting for response from engine for ${processIdWithName.name}")
  Future.successful(failedToGetResult)
} else {
  stateFromEngineFuture
}
```

if `statusFetchCompleted` is true ,`stateFromEngineFuture` should be completed by now, so we may return it. WDYT?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
arkadius: I can't agree with this suggestion. We should wait a little bit longer than this assumed time. The time that we should wait is a derivate of time that we assume that future will be completed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -266,15 +269,34 @@ class DeploymentServiceImpl(dispatcher: DeploymentManagerDispatcher,
...
   private def getStateFromEngine(deploymentManager: DeploymentManager, processIdWithName: ProcessIdWithName)
-                                (implicit ec: ExecutionContext, freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[Option[ProcessState]]] =
-    deploymentManager.getProcessState(processIdWithName.name).recover {
+                                (implicit ec: ExecutionContext, freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[Option[ProcessState]]] = {
+    lazy val failedToGetResult =
+      WithDataFreshnessStatus(
+        Option(SimpleProcessStateDefinitionManager.errorFailedToGet),
+        cached = false)
+
+    @volatile var statusFetchCompleted = false
```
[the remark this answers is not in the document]
arkadius: I removed this local var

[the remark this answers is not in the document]
arkadius: BTW On this page: https://stackoverflow.com/a/50108482/1370301 is clarified that sometimes it makes sense to mark local variable as a `@volatile`. It depends on if this variable escape the local scope or not. I think that in this case this variable escaped from the local scope - but I'm not 100% sure. Regarding java, there is limitation that local variable that may be accessed by anonymous class, should be `final` - I think that it is done to make sure that it stay on stack. In scala, this limitation is turned off.

arkadius:
I additionally added a little improvement that is not related with this change - better handling situation when process type is not configured for some scenario - now list of scenarios will be loaded. I broke it in one of previous changes. I don't want to open a new PR for that because I'll have conflicts

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
gadomsky: Hmmm, but this first sentence is just an introduction to this definition, I think my suggestions has the same meaning (or not?) as it was before.
The wole comment would look like:
```scala
// Sometimes we know how long we should wait until a Future is completed. In this case
// we want to wait a little big longer than this time, to make sure that in case of Failure, inner Failure won't
// be suppressed by exception "A timeout occurred waiting for a future to complete" and extract this inner exception
```
isn't it what you said?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gadomsky: looks a lot better now :)

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/BasePatientScalaFutures.scala
arkadius: I rephrased this comment a bit to make it less complicated.

gadomsky: thanks

gadomsky (approved):

PR #4161, opened 2023-03-23
Bump dependencies

piotrp:
Mostly safe upgrades, except:
* some commons dependencies were outdated when they were originally added to Nu, but there was no comment explaining that this was required - so I updated them
* netty is upgraded from 4.1.48.Final to 4.1.90.Final - we're already running model with 4.1.60, and AsyncHttpClient seems to work fine with 4.1.87: https://github.com/AsyncHttpClient/async-http-client/commits/main/pom.xml

Not updated:
* cats-effect left at 2.x
* testcontainers-keycloak - newer version fails in CI due to some JUnit dependency mismatch
* ignite - newer version fails in CI on some data type conversion error, I couldn't quickly find out why

Also, I added version overrides for various netty and Jackson libraries pulled as transitive dependencies - we were missing some of them, resulting in a mix of various versions.

### DIFF
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
@@ -230,11 +230,8 @@ object SwaggerObject {
 
     val additionalProperties = schema.getAdditionalProperties match {
       case null => AdditionalPropertiesEnabled(SwaggerAny)
-      case additionalPropertyEnabled if additionalPropertyEnabled == true => AdditionalPropertiesEnabled(SwaggerAny)
-      case additionalPropertyEnabled if additionalPropertyEnabled == false => AdditionalPropertiesDisabled
-      case additionalPropertySchema: Schema[_] =>
-        val typed = AdditionalPropertiesEnabled(SwaggerTyped(additionalPropertySchema, swaggerRefSchemas, usedRefs))
-        typed
+      case schema: Schema[_] if schema.getBooleanSchemaValue == false => AdditionalPropertiesDisabled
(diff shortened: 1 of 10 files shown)


utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/swagger/SwaggerTyped.scala
```diff
@@ -230,11 +230,8 @@ object SwaggerObject {
 
     val additionalProperties = schema.getAdditionalProperties match {
       case null => AdditionalPropertiesEnabled(SwaggerAny)
-      case additionalPropertyEnabled if additionalPropertyEnabled == true => AdditionalPropertiesEnabled(SwaggerAny)
-      case additionalPropertyEnabled if additionalPropertyEnabled == false => AdditionalPropertiesDisabled
-      case additionalPropertySchema: Schema[_] =>
-        val typed = AdditionalPropertiesEnabled(SwaggerTyped(additionalPropertySchema, swaggerRefSchemas, usedRefs))
-        typed
+      case schema: Schema[_] if schema.getBooleanSchemaValue == false => AdditionalPropertiesDisabled
+      case schema: Schema[_] => AdditionalPropertiesEnabled(SwaggerTyped(schema, swaggerRefSchemas, usedRefs))
```
gadomsky, line 234: do we change behaviour here?

piotrp: Tests:
https://github.com/TouK/nussknacker/blob/73400320cd4d7dbc89110ef2a0f0d7ca0fe619c6/utils/json-utils/src/test/scala/pl/touk/nussknacker/engine/json/SwaggerBasedJsonSchemaTypeDefinitionExtractorTest.scala#L15:L35

gadomsky (approved):
nice, thanks!

arkadius (approved):
LGTM, let's see what will happen

PR #4172, opened 2023-03-24
workaround missing status icon

JulianWielga:


JulianWielga:
same from BE (better) side #4174

PR #4186, opened 2023-03-27
Simpler status definition without StateStatus

gskrobisz:
- [x]  Serialize StateStatus as String (StatusName)
- [ ] Use status identifiers instead of StateStatus.isXYZ
- [x] Remove sealed trait
- [ ] In ProcessState replace `status: StateStatus` with `statusName: StateName`

Out of scope changes / to discuss:
- ProcessState.tooltip: Option[String] -> String => https://github.com/TouK/nussknacker/pull/4189
- ProcessState.description: Option[String] -> String => https://github.com/TouK/nussknacker/pull/4189
- Do we need status ProblemStateStatus.processWithoutAction?
- Add another unsafe check in ProcessStateDefinitionService: check if every status definition has non-empty list of allowed actions


designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/StateDefinitionDetails.scala
```diff
@@ -0,0 +1,24 @@
...
+  * When a status has dynamic properties use ProcessStateDefinitionManager to handle them.
+  *
+  * @see default values of a status in [[ProcessStateDefinitionManager]]
+  * @see filtering options in [[UIStateDefinition]]
+  * @see overriding state definitions in [[OverridingProcessStateDefinitionManager]]
+  */
+case class StateDefinitionDetails(displayableName: String,
+                                  icon: URI,
+                                  tooltip: Option[String],
+                                  description: Option[String])
```
lciolecki, line 20: Hmm, why not add here allowedActions and remove ProblemStateStatus? After that change we won't need `statusActionsPF`

arkadius: Good point, but we still will need it in some cases because this definition is only for static cases - see the docs above

arkadius:
Closing it. It is too much desynchronized and it looks like it will be impossible to rebase this branch

PR #4191, opened 2023-03-28
Refactor: Periodic - merge Flink's process state with deployments state in db done in Service

arkadius:


arkadius:
@gadomsky Thank you for the review, I've added one commit moving this new `ProcessStateInputData` class into `ProcessState.scala` file and renaming to `StatusDetails`. We have plans to reuse it in the wider cleanup that @gskrobisz is working on.

PR #4210, opened 2023-04-03
Fix scenario properties form types

mslabek:
Typing a non-integer value in the parallelism or checkpoint interval properties resulted in an ugly looking backend unmarshalling error since backend expects an integer/long in these fields. There doesn't seem to be a way to get types and default values of type specific scenario properties without modifying the backend api. The simplest solution is to prevent the user from typing non-integers.

The form with new placeholders looks like this:
properties-flink-defaults

There is no feedback to the user when he tries to type a non-integer, but that is still preferable than throwing an ugly error.

### DIFF
--- a/designer/client/components/graph/node-modal/editors/field/Field.tsx
+++ b/designer/client/components/graph/node-modal/editors/field/Field.tsx
@@ -1,4 +1,4 @@
-import React, {PropsWithChildren} from "react"
+import React, {PropsWithChildren, useCallback} from "react"
 import Checkbox from "./Checkbox"
 import Input from "./Input"
 import LabeledInput from "./LabeledInput"
@@ -16,6 +16,7 @@ export enum FieldType {
 interface FieldProps {
   isMarked: boolean,
   readOnly: boolean,
+  placeholder?: string,
   showValidation: boolean,
   autoFocus: boolean,
   className: string,
@@ -32,6 +33,7 @@ export default function Field({type, children, ...props}: PropsWithChildren<Fiel
         <LabeledInput
           {...props}
           value={props.value?.toString() || ""}
+          placeholder={props.placeholder}
           onChange={({target}) => props.onChange(target.value)}
         >
           {children}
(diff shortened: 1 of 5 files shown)


designer/client/components/graph/node-modal/editors/field/Field.tsx
```diff
@@ -5,24 +5,36 @@ import LabeledInput from "./LabeledInput"
...
+  options?: Option[],
   value: string | boolean,
-  onChange: (value: string | boolean) => void,
+  onChange: (value: string | boolean | Option) => void,
+}
+
+function getCurrentBooleanOption(options: Option[], value: boolean | string): Option {
+  return options.find(x => {
+    const optionIsTrueFalse = x.value === value
+    const optionIsNull = x.value === null && value === undefined
```
JulianWielga, line 35: ?

lciolecki:
But what about validating it on the backend side? I think we should also validate it on the backend and return a nice message.

designer/client/components/graph/node-modal/properties.tsx
```diff
@@ -94,10 +97,15 @@ export function Properties({
                 node={node}
                 renderFieldLabel={renderFieldLabel}
                 setProperty={setProperty}
-                fieldType={FieldType.checkbox}
-                fieldLabel={"Spill state to disk"}
-                fieldProperty={"typeSpecificProperties.spillStateToDisk"}
-                validators={[errorValidator(fieldErrors || [], "spillStateToDisk")]}
+                fieldType={FieldType.select}
+                fieldLabel={"I/O mode"}
```
arkadius, line 101: It shouldn't be done on FE, we should move this to BE, we already has options to do that. Let's talk

arkadius (changes requested):
This change explores the direction of static scenario definitions on FE side, that we are trying to get rid of

mslabek:
Thanks for feedback. I'm taking a new approach to this problem in line with your suggestions. The WIP PR: https://github.com/TouK/nussknacker/pull/4225

PR #4216, opened 2023-04-05
Fix in fragment definition extraction: properly handling empty fragments

arkadius:

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/SubprocessComponentDefinitionExtractor.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/SubprocessComponentDefinitionExtractor.scala
@@ -1,8 +1,8 @@
 package pl.touk.nussknacker.engine.definition
 
 import cats.Id
-import cats.data.Validated.{invalid, valid}
-import cats.data.{NonEmptyList, ValidatedNel, WriterT}
+import cats.data.Validated.{Invalid, Valid, invalid, valid}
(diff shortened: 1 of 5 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/SubprocessComponentDefinitionExtractor.scala
```diff
@@ -46,29 +46,34 @@ abstract class SubprocessDefinitionExtractor {
 
 }
 
+sealed trait FragmentDefinitionError
+
+case object EmptyFragmentError extends FragmentDefinitionError
```
gadomsky, line 51: what message is propaged to UI when we have this kind of error?

gadomsky: Ok, I guess I've found it, it is mapped to `InvalidSubprocess` so we have description: `Node $nodeId uses fragment $id which is invalid` 

What do you think about a change, so we have:
`InvalidSubprocess(id: String, nodeId: String, cause: FragmentDefinitionError)` and have dedicated error description for each kind of `cause`?

arkadius: IMO exact errors should be presented when smb is in fragment. When fragment is used, user should rather don't think about exact problems - after all, fragments are for decomposition

PR #4217, opened 2023-04-05
Draft: Lock cancel button when there are unsaved changes

mslabek:
While a scenario is deployed, when a user makes changes and cancels the deployment he loses the unsaved changes without any previous prompt. This change locks the cancel button when there are unsaved changes (same as deploy button). The user gets feedback in the form of a tooltip.

cancel-blocked


arkadius:
Why just not lost this unsaved changes instead?

lciolecki:
I agree with @arkadius this solution is a "palette knife" :)

mslabek:
Thanks for your notes guys. Will start working on a non-temporary, less "pallete-knifey" solution :)

PR #4225, opened 2023-04-12
Draft: Make type specific properties generic

mslabek:


designer/client/components/graph/node-modal/properties.tsx
```diff
@@ -45,109 +45,29 @@ export function Properties({
...
-          <NodeField
-            isEditMode={isEditMode}
+      {typeSpecificPropertiesConfig.map(([propName, propConfig]) => (
+        <AdditionalProperty
+            key={propName}
+            showSwitch={showSwitch}
             showValidation={showValidation}
-            node={node}
+            propertyName={propName}
+            propertyPathPrefix={"typeSpecificProperties.properties"}
```
gadomsky, line 55: and in such case, here we should have 
```
            propertyPathPrefix={"typeSpecificProperties"}
```

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
```diff
@@ -30,6 +32,43 @@ class FlinkStreamingDeploymentManagerProvider extends DeploymentManagerProvider
 
   override def typeSpecificInitialData(config: Config): TypeSpecificInitialData = TypeSpecificInitialData(StreamMetaData(Some(1)))
 
+  override def typeSpecificPropertiesConfig(config: Config): Map[String, AdditionalPropertyConfig] = {
+    Map(parallelismConfig, spillStateConfig, asyncInterpretationConfig, checkpointIntervalConfig)
+  }
+
+  private val parallelismConfig: (String, AdditionalPropertyConfig) = "parallelism" ->
+    AdditionalPropertyConfig(
```
gadomsky, line 40: and `AdditionalPropertyConfig` could be extendend with new field `type` (for now we could support only basic types - String, Boolean, Int, Long, Double), and make FE forms handle those types.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/DisplayableProcess.scala
```diff
@@ -50,21 +53,68 @@ import pl.touk.nussknacker.engine.graph.NodeDataCodec._
...
-    Encoder.forProduct3("typeSpecificProperties", "isSubprocess", "additionalFields") { p =>
-    (p.typeSpecificProperties, p.isSubprocess, p.additionalFields)
+
+  def apply(typeSpecificProperties: TypeSpecificData,
+            additionalFields: Option[ProcessAdditionalFields] = None): ProcessProperties = {
+    val genericTypeSpecificData = typeSpecificProperties.toGenericMap
+    val additionalProps = additionalFields.map(a => a.properties).getOrElse(None)
+    val description = additionalFields.flatMap(a => a.description)
+
+    val props = ProcessAdditionalFields(description, genericTypeSpecificData ++ additionalProps)
```
gadomsky, line 81: what happens when we have duplicate propertyNames? (I mean one in additonal and one in typespecific)

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
```diff
@@ -3,14 +3,17 @@ package pl.touk.nussknacker.engine.management
...
 import pl.touk.nussknacker.engine.api.StreamMetaData
+import pl.touk.nussknacker.engine.api.component.AdditionalPropertyConfig
+import pl.touk.nussknacker.engine.api.definition.{BoolParameterEditor, FixedExpressionValue, FixedValuesParameterEditor, FixedValuesValidator, LiteralIntegerValidator, MandatoryParameterValidator, MinimalNumberValidator, StringParameterEditor}
 import pl.touk.nussknacker.engine.api.deployment.cache.CachingProcessStateDeploymentManager
 import pl.touk.nussknacker.engine.api.deployment.{DeploymentManager, ProcessingTypeDeploymentService}
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class FlinkStreamingDeploymentManagerProvider extends DeploymentManagerProvider {
+class FlinkStreamingDeploymentManagerProvider extends DeploymentManagerProvider with LazyLogging {
```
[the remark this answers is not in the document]
mslabek: No, its a leftover after debugging, accidentally commited. Will check the commit for more.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
```diff
@@ -41,4 +46,44 @@ object FlinkStreamingDeploymentManagerProvider {
...
+object FlinkStreamingPropertiesConfig {
+
+  lazy val properties: Map[String, AdditionalPropertyConfig] =
+    Map(parallelismConfig, spillStateConfig, asyncInterpretationConfig, checkpointIntervalConfig)
+
+  private val parallelismConfig: (String, AdditionalPropertyConfig) = "parallelism" ->
+    AdditionalPropertyConfig(
+      defaultValue = None,
+      editor = Some(StringParameterEditor),
+      validators = Some(List(MandatoryParameterValidator, LiteralIntegerValidator, MinimalNumberValidator(1))),
```
arkadius, line 60: IMO we shouldn't add `MandatoryParameterValidator` - we should pick the default when user will left this not defined

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
```diff
@@ -41,4 +46,44 @@ object FlinkStreamingDeploymentManagerProvider {
...
+    AdditionalPropertyConfig(
+      defaultValue = None,
+      editor = Some(BoolParameterEditor),
+      validators = None,
+      label = Some("Spill state to disk"))
+
+  private val asyncPossibleValues = List(
+    FixedExpressionValue("false", "Synchronous"),
+    FixedExpressionValue("true", "Asynchronous"),
+    FixedExpressionValue("", "Server default"))
```
arkadius, line 73: The first option on the list is considered the default. I think that we should have "" as first

engine/lite/deploymentManager/src/main/scala/pl/touk/nussknacker/lite/manager/LiteDeploymentManagerProvider.scala
```diff
@@ -2,30 +2,48 @@ package pl.touk.nussknacker.lite.manager
...
+      streamingInitialMetaData,
       (scenarioName: ProcessName, _: String) => RequestResponseMetaData(Some(defaultRequestResponseSlug(scenarioName, config)))
     )
   }
 
+  private val parallelismConfig: (String, AdditionalPropertyConfig) = "parallelism" ->
+    AdditionalPropertyConfig(
+      defaultValue = None,
+      editor = Some(StringParameterEditor),
+      validators = Some(List(MandatoryParameterValidator, LiteralIntegerValidator, MinimalNumberValidator(1))),
```
arkadius, line 26: Like for flink - we rather should make all properties optional and use designer default values in that cases

build.sbt
```diff
@@ -1197,6 +1197,7 @@ lazy val developmentTestsDeploymentManager = (project in development("deployment
     name := "nussknacker-development-tests-manager",
   ).dependsOn(
   deploymentManagerApi % "provided",
+  flinkDeploymentManager % "provided",
```
gadomsky, line 1200: `provided` will be enough? Then it will require to use it allways with  `flinkDeploymentManager` in runtime

mslabek: When I declared it with test scope, the DevelopmentDeploymentManagerProvider, which is in main (not test) folder could not access it. But I'm thinking of reverting these introduced dependencies for tests so that testing deployment managers don't have to rely on the flink deployment manager.

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -18,44 +15,41 @@ import scala.concurrent.duration.Duration
...
-// TODO: rename to FlinkStreamMetaData
-case class StreamMetaData(parallelism: Option[Int] = None,
-                          //we assume it's safer to spill state to disk and fix performance than to fix heap problems...
-                          spillStateToDisk: Option[Boolean] = Some(true),
-                          useAsyncInterpretation: Option[Boolean] = None,
-                          checkpointIntervalInSeconds: Option[Long] = None) extends ScenarioSpecificData {
+    val otherProps = properties.filterNot(k => typeSpecificPropNamesList.contains(k._1))
+    val fields = otherProps match {
+      case m if m.isEmpty && description.isEmpty => None
+      case e => Some(ProcessAdditionalFields(description, e))
```
gadomsky, line 35: why `e`?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
```diff
@@ -0,0 +1,133 @@
...
+object StreamMetaData {
+  private val parallelismFlinkName = "parallelism"
+  private val spillStateToDiskName = "spillStateToDisk"
+  private val useAsyncInterpretationName = "useAsyncInterpretation"
+  private val checkpointIntervalInSecondsName = "checkpointIntervalInSeconds"
+
+  def apply(properties: Map[String, String]): StreamMetaData = {
+    StreamMetaData(
+      parallelism = properties.get(parallelismFlinkName).flatMap(convertPropertyWithLog(_, _.toInt, parallelismFlinkName)),
+      spillStateToDisk = properties.get(spillStateToDiskName).flatMap(convertPropertyWithLog(_, _.toBoolean, spillStateToDiskName)),
```
gadomsky, line 80: let's have fallback to `Some(true)` here

gadomsky (approved):

gadomsky (approved):

common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
```diff
@@ -0,0 +1,133 @@
...
+  override val toProperties: Map[String, String] = toStringMap(List(
+    parallelism.map(parallelismFlinkName -> _),
+    spillStateToDisk.map(spillStateToDiskName -> _),
+    useAsyncInterpretation.map(useAsyncInterpretationName -> _),
+    checkpointIntervalInSeconds.map(checkpointIntervalInSecondsName -> _)
+  ))
+}
+
+object StreamMetaData {
+  private val parallelismFlinkName = "parallelism"
```
arkadius, line 72: why Flink name?  Everywhere else is without flink

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/displayedgraph/DisplayableProcess.scala
```diff
@@ -49,22 +50,60 @@ import pl.touk.nussknacker.engine.graph.NodeDataCodec._
...
+    // a property with the same name was written as a generic property and in type specific properties. This shouldn't
+    // happen, but if it does, the data will heal on next save:
+    //  - if the value has a type that can be stored in TypeSpecificData it will be stored there and removed from
+    //    generic properties
+    //  - if the value has different type than in TypeSpecificData it will be stored in generic properties and in
+    //    TypeSpecificData it will be set as null
+    // This is a temporary solution until we get rid of TypeSpecificData completely.
+    typeSpecificData.toProperties
+      .filter(p => fields.properties.contains(p._1))
+      .foreach(p => logger.warn(s"Duplicate properties with the same name '${p._1}'. Overwriting additional property " +
```
arkadius, line 98: I assume that this will happen every time and we will have this warn during normal usage?  I would add this warn only when a value is different. WDYT?

mslabek: The 'toProperties' method returns a map of non-null properties. So if we have an invalid property (which in TypeSpcificData will be written as null), it won't be placed in the returned map. I will add this explanation to the comment - your assumption is definitely warranted.

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
mslabek: No reason, will write something less misleading

common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
mslabek: Thanks, working on it now

common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
mslabek: I assumed the StreamMetaData wasn't renamed to FlinkStreamMetaData was because of some database migration, but I'm totally unsure. I named all new variables as flink since its more informative. Would it be better to remain consistent and name it 'parallelismStreamName'?

PR #4246, opened 2023-04-24
Store components usages along with scenario JSON

jedrz:
* Store components usages for a single scenario in database along with scenario JSON to avoid querying and traversing all scenario JSONs.
* Component service combines usages across all scenarios and all processing types (computes unique component id based on component type, name and processing type).
* In the first step of Java-based migrations for the process versions table, the query only retrieves all process IDs and version IDs to prevent excessive memory usage. Afterwards, the scenario JSON is queried for each row separately.
* Tests for Flyway migrations. Combined both Flyway and process migrations tests for different databases into one test class. 
* Last reverted commit has changed component usages signature from `Map[ComponentIdParts, List[NodeId]]` to `List[ComponentUsages]` where `ComponentUsages` is `(ComponentName, ComponentType, List[NodeIds])` but decided to leave the map version. Nevertheless, it can be easily changed to list based signature.


designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentIdProvider.scala
```diff
@@ -13,6 +13,7 @@ import pl.touk.nussknacker.ui.process.ProcessCategoryService
 //TODO: It is work around for components duplication across multiple scenario types, until we figure how to do deduplication.
 trait ComponentIdProvider {
   def createComponentId(processingType: ProcessingType, name: String, componentType: ComponentType): ComponentId
+  def createComponentId(processingType: ProcessingType, name: Option[String], componentType: ComponentType): ComponentId
```
arkadius, line 16: Let's have one method and pass Some() in places where it is defined

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
```diff
@@ -330,7 +332,14 @@ trait EspItTest extends LazyLogging with WithHsqlDbTesting with TestPermissions
...
   private def saveAndGetId(process: CanonicalProcess, category: String, isSubprocess: Boolean, processingType: ProcessingType = Streaming): Future[ProcessId] = {
     val processName = ProcessName(process.id)
-    val action = CreateProcessAction(processName, category, process, processingType, isSubprocess, None)
+    val action = CreateProcessAction(
+      processName,
+      category,
+      process,
+      processingType,
+      isSubprocess,
+      ComponentsUsageHelper.computeUsagesForScenario(process),
```
arkadius, line 341: 1. WDYT about moving this `computeUsagesForScenario` into repository level? Usages are just an index, if we have json indexes in DB, we won't need it
a) If you don't like this concept, at least extract `Create/UpdateProcessAction.apply` factory methods computing usages based on process - it will be haderd to desynchronize it
2. WDYT about moving `computeUsagesForScenario` logic into `CanonicalProcess` and having just a `CanonicalProcess.componentUsages`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessVersionEntityFactory.scala
```diff
@@ -62,13 +95,41 @@ trait ProcessVersionEntityFactory extends BaseEntityFactory {
...
                                     json: Option[CanonicalProcess],
                                     createDate: Timestamp,
                                     user: String,
-                                    modelVersion: Option[Int])
+                                    modelVersion: Option[Int],
+                                    componentsUsages: Option[ScenarioComponentsUsages],
+                                   )
+
+@JsonCodec
+case class ComponentUsages(componentName: Option[String], componentType: ComponentType, nodeIds: List[NodeId])
```
arkadius, line 119: It can't be private?

arkadius (commented):
Nice changel, minor comments added

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/EspItTest.scala
jedrz: Good point! I chose option 1 because moving the logic into `CanonicalProcess` would unnecessarily expose component usages in the public API.

arkadius (approved):

Issue #2676, opened 2021-12-22
Uniqueness of input/output parameters in fragments isn't validated


piotrp:
Names of input/output parameters in fragments aren't validated, making both configurations below valid:

![obraz](https://user-images.githubusercontent.com/1409807/147104357-c9c64980-4dea-4c32-b034-1bf2273118ce.png)
![obraz](https://user-images.githubusercontent.com/1409807/147104382-a3d16eb4-a140-4cf9-bf1e-472e0d913609.png)

Version: 1.2.0-staging-2021-12-22-6130-62963fe50b10eccc585edb4408a3adae9932e523-SNAPSHOT

PR #4257, opened 2023-04-26
Fe map fields name validation

JulianWielga:


JulianWielga:
partial fix for #2676

gadomsky (approved):

PR #4264, opened 2023-04-27
Add Any/Unknown to fragment's input types

vder:

### DIFF
--- a/engine/flink/management/dev-model/src/test/resources/extractedTypes/devCreator.json
+++ b/engine/flink/management/dev-model/src/test/resources/extractedTypes/devCreator.json
@@ -25,6 +25,23 @@
     "staticMethods": {}
   },
   {
+    "clazzName": {
+      "type" : "Unknown"
+    },
+    "methods": {
+      "toString": [
+        {
+          "name": "toString",
+          "signature": {
+            "noVarArgs": [],
+            "result": {"refClazzName": "java.lang.String"}
+          }
+        }
+      ]
+    },
+    "staticMethods": {}
+  },
+  {
     "clazzName": {"refClazzName": "com.cronutils.model.Cron"},
     "methods": {
       "asString": [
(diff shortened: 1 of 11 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -75,12 +75,23 @@ object TypeInfos {
...
-
-  case class ClazzDefinition(clazzName: TypedClass,
+  case class ClazzDefinition(clazzName: TypingResult,
                              methods: Map[String, List[MethodInfo]],
                              staticMethods: Map[String, List[MethodInfo]]) {
+
+    def getClazz: Class[_] = this.clazzName match {
+      case TypedClass(klass, _) => klass
+      case Unknown => AnyClass
+      case _ => ???
```
arkadius, line 85: Let's replace it with throwing `IllegalAccessException` with some meaningful message that we currently support only classes and unknown in fragments parameter types

engine/flink/management/dev-model/src/test/resources/extractedTypes/devCreator.json
```diff
@@ -24,6 +24,23 @@
     },
     "staticMethods": {}
   },
+  {
+    "clazzName": {
+      "type" : "Unknown"
```
arkadius, line 29: Shoulndn't something change near `refClazzName` usages on FE side? I see that it isn't optional

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/TypeInfos.scala
```diff
@@ -75,12 +75,23 @@ object TypeInfos {
     }
   }
 
-
-  case class ClazzDefinition(clazzName: TypedClass,
+  case class ClazzDefinition(clazzName: TypingResult,
```
arkadius, line 78: Let's add tests in `InterpreterSpec` showing that it is possible to pass anything into `Unknown` and it is possible to use it inside the fragment

PR #4265, opened 2023-04-27
Remove implicit helper methods in Spel (sum, today, now, distinct)

rafa-minimal:
These methods would be always added to the Spel context. I guess we don't need them and any helper should be explicitly added in ProcessConfigCreator.


gadomsky (approved):
LGTM

PR #4267, opened 2023-04-28
Removing -target jvm opts from scala 2.13 build because is depreated …

arkadius:
…in java17 (we use -release instead)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -199,8 +199,7 @@ lazy val commonSettings =
           "-target:jvm-1.8",
         ),
         (2, 13) -> Seq(
-          "-Ymacro-annotations",
-          "-target:jvm-11",
+          "-Ymacro-annotations"
         )
       ),
       javacOptions := Seq(


piotrp:
Official Scala documentation at https://docs.scala-lang.org/overviews/compiler-options/index.html mentions that `11` is supported for `jvm:target` (which is deprecated) and isn't for `-release`.

But test with real `scalac` proves that you are right :D
```sh
$ docker run --rm -i sbtscala/scala-sbt:eclipse-temurin-jammy-17.0.5_8_1.8.2_2.13.10 scalac | grep -E 'target|release'
  -release:<release>           Compile for a version of the Java API and target class file. (8,9,10,11,12,13,14,15,16,[17])
  -target:<target>             Target platform for object files. ([8],9,10,11,12,13,14,15,16,17)
                                 deprecated: Use -release instead to compile against the correct platform API.
```

arkadius:
So it looks like official docs are outdated. Tests pass - they are run at `temurin:1.11.0.17`, also we have `JavaClassVersionChecker` which just in case check if methods from java 11 API are available (in case if was used -target instead of -release which is done for scala 2.12)

piotrp (approved):
Oops, I didn't give you approval with my first comment because I didn't remember that I could... But that was me being blind👍

PR #4275, opened 2023-05-05
Add collection helpers to default model (e.g. sum, shuffle, distinct, sorted...)

rafa-minimal:
* hide toString on helper objects

### DIFF
--- a/docs/scenarios_authoring/Spel.md
+++ b/docs/scenarios_authoring/Spel.md
@@ -256,6 +256,8 @@ listOfPersons = {person1, person2}
 | `#listOfPersons.![#this.age]`  | {42, 24}         | List[Integer] |
 | `#listOfPersons.![7]`          | {7, 7}           | List[Integer] |
 
+For other operations on lists, please see the `#COLLECTION` [helper](#built-in-helpers).
+
 ## Safe navigation
 
 When you access nested structure, you have to take care of null fields, otherwise you'll end up with 
@@ -318,11 +320,11 @@ You can also use explicit conversions that are available in utility classes and 
 ## Built-in helpers 
 
 | Helper        | Functions                                      |
-| -------       | -----------                                    |
-| `GEO`         | Simple distance measurements                   |
-| `NUMERIC`     | Number parsing                                 |
+|---------------|------------------------------------------------|
+| `COLLECTION`  | Operations on collections                      |
 | `CONV`        | General conversion functions                   |
 | `DATE`        | Date operations (conversions, useful helpers)  |
 | `DATE_FORMAT` | Date formatting/parsing operations             |
+| `GEO`         | Simple distance measurements                   |
+| `NUMERIC`     | Number parsing                                 |
 | `UTIL`        | Various utilities (e.g. identifier generation) |
-| `AGG`         | Aggregator functions                           |
--- a/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
+++ b/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
@@ -1,0 +1,236 @@
+package pl.touk.nussknacker.engine.util.functions
+
+import cats.data.{Validated, ValidatedNel}
+import cats.implicits.catsSyntaxValidatedId
+import org.springframework.util.{NumberUtils => SpringNumberUtils}
+import pl.touk.nussknacker.engine.api.generics.{GenericFunctionTypingError, GenericType, TypingFunction}
+import pl.touk.nussknacker.engine.api.typed.supertype.NumberTypesPromotionStrategy.ForLargeNumbersOperation
+import pl.touk.nussknacker.engine.api.typed.typing
+import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypedClass, TypedObjectTypingResult, Unknown}
+import pl.touk.nussknacker.engine.api.{Documentation, HideToString, ParamName}
+
+import java.util.{Collections, Objects}
+import scala.collection.immutable.ListMap
+import scala.jdk.CollectionConverters._
+import scala.reflect.ClassTag
+
+object collection extends HideToString {
+
+  @Documentation(description = "Concatenates two lists")
+  @GenericType(typingFunction = classOf[ListAdditionTyping])
+  def concat[T](@ParamName("list1") list1: java.util.List[T], @ParamName("list2") list2: java.util.List[T]): java.util.List[T] =
+    (list1.asScala.toList ++ list2.asScala).asJava
+
+  @Documentation(description = "Merges two maps. Values in the first map will be overwritten with values from the second map if keys are the same")
+  @GenericType(typingFunction = classOf[MapMergeTyping])
+  def merge[K, V](@ParamName("map1") map1: java.util.Map[K, V], @ParamName("map2") map2: java.util.Map[K, V]): java.util.Map[K, V] = {
+    val merged = new java.util.LinkedHashMap[K, V](map1)
+    merged.putAll(map2)
+    merged
+  }
+
+  @Documentation(description = "Returns the smallest element (elements must be comparable)")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def min[T <: Comparable[T]](@ParamName("list") list: java.util.Collection[T]): T = Collections.min[T](list)
+
+  @Documentation(description = "Returns the largest element (elements must be comparable)")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def max[T <: Comparable[T]](@ParamName("list") list: java.util.Collection[T]): T = Collections.max[T](list)
+
+  @Documentation(description = "Returns a slice of the list starting with start index (inclusive) and ending at stop index (exclusive)")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def slice[T](@ParamName("list") list: java.util.Collection[T], @ParamName("start") start: Int, @ParamName("stop") stop: Int): java.util.List[T] =
+    list.asScala.slice(start, stop).toList.asJava
+
+  // TODO: This method currently has the following limitations:
+  // * it won't return the expected type for empty list
+  // * it works slower than it could with statically typed result
+  // To overcome this limitations we should supply returnType to this method
+  @Documentation(description = "Returns a sum of all elements")
+  @GenericType(typingFunction = classOf[ListElementTypingForSum])
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): T = {
+    val types = list.asScala.map(Typed.fromInstance(_)).toList
+    val targetType = types match {
+      // We are not able to determine the expected type for an empty list
+      case Nil => return 0.0.asInstanceOf[T]
+      case t::Nil => ForLargeNumbersOperation.promoteSingle(t)
+      case l => l.reduce((a, b) => ForLargeNumbersOperation.promote(a, b))
+    }
+
+    if (targetType == Typed[java.lang.Long]) {
+      list.asScala.map(_.longValue()).sum.asInstanceOf[T]
+    } else if (targetType == Typed[java.lang.Double]) {
+      list.asScala.map(_.doubleValue()).sum.asInstanceOf[T]
+    } else if (targetType == Typed[java.math.BigInteger]) {
+      list.asScala.map(SpringNumberUtils.convertNumberToTargetClass(_, classOf[java.math.BigInteger]))
+        .reduce((a, b) => a.add(b)).asInstanceOf[T]
+    } else {
+      list.asScala.map(SpringNumberUtils.convertNumberToTargetClass(_, classOf[java.math.BigDecimal]))
+        .reduce((a, b) => a.add(b)).asInstanceOf[T]
+    }
+  }
+
+  @Documentation(description = "Returns a list of all elements sorted in ascending order (elements must be comparable)")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def sortedAsc[T <: Comparable[T]](@ParamName("list") list: java.util.Collection[T]): java.util.List[T] = sorted(list, desc = false)
+
+  @Documentation(description = "Returns a list of all elements sorted in descending order (elements must be comparable)")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def sortedDesc[T <: Comparable[T]](@ParamName("list") list: java.util.Collection[T]): java.util.List[T] = sorted(list, desc = true)
+
+  private def sorted[T <: Comparable[T]](list: java.util.Collection[T], desc: Boolean): java.util.List[T] = {
+    checkIfComparable(list)
+    val values = new java.util.ArrayList[T](list)
+    Collections.sort[T](values)
+    if (desc) {
+      Collections.reverse(values)
+    }
+    values
+  }
+
+  @Documentation(description = "Returns a list made of first n elements of the given list")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def take[T](@ParamName("list") list: java.util.List[T], @ParamName("max") max: Int): java.util.List[T]
+  = list.asScala.take(max).asJava
+
+  @Documentation(description = "Returns a list made of last n elements of the given list")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def takeLast[T](@ParamName("list") list: java.util.List[T], @ParamName("max") max: Int): java.util.List[T]
+  = list.asScala.takeRight(max).asJava
+
+  @Documentation(description = "Creates a string made of all elements of the list separated with the given separator")
+  def join[T](@ParamName("list") list: java.util.List[T], @ParamName("separator") separator: String): String
+  = String.join(separator, list.asScala.map(Objects.toString).asJava)
+
+  @Documentation(description = "Cross joins two lists of maps: eg. product({{a: 'a'},{b: 'b'}}, {{c: 'c'},{d: 'd'}}) => {{a: 'a',c: 'c'},{b: 'b',c: 'c'},{a: 'a',d: 'd'},{b: 'b',d: 'd'}}")
+  def product[K, V](list1: java.util.List[java.util.Map[K, V]], list2: java.util.List[java.util.Map[K, V]]): java.util.List[java.util.Map[K, V]] = {
+    val l1 = list1.asScala.map(_.asScala)
+    val l2 = list2.asScala.map(_.asScala)
+    val res = for {
+      m1 <- l1
+      m2 <- l2
+    } yield m1 ++ m2
+    res.map(_.asJava).asJava
+  }
+
+  @Documentation(description = "Returns a list that contains all elements contained in list1, that don't appear in list2")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def diff[T](@ParamName("list1") list1: java.util.List[T], @ParamName("list2") list2: java.util.List[T]): java.util.List[T] =
+    list1.asScala.filterNot(list2.asScala.toSet).asJava
+
+  @Documentation(description = "Returns a list that contains all unique elements that are contained by both list1 and list2")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def intersect[T](@ParamName("list1") list1: java.util.List[T], @ParamName("list2") list2: java.util.List[T]): java.util.List[T] =
+    (list1.asScala.toSet intersect list2.asScala.toSet).toList.asJava
+
+  @Documentation(description = "Returns a list that contains unique elements from the given list")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def distinct[T](@ParamName("list") list: java.util.List[T]): java.util.List[T] =
+    new java.util.ArrayList(new java.util.HashSet[T](list))
+
+  @Documentation(description = "Returns a copy of the list with its elements shuffled")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def shuffle[T](@ParamName("list") list: java.util.Collection[T]): java.util.List[T] = {
+    val values = new java.util.ArrayList[T](list)
+    Collections.shuffle(values)
+    values
+  }
+
+  @Documentation(description = "Returns a list of all elements from all lists in the given list")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def flatten[T](@ParamName("list") list: java.util.Collection[java.util.Collection[T]]): java.util.List[T] =
+    list.asScala.flatMap(_.asScala).toList.asJava
+
+  private def checkIfComparable(list: Iterable[_]): Unit =
+    list.foreach(checkIfComparable)
+
+  private def checkIfComparable(list: java.util.Collection[_]): Unit =
+    list.asScala.foreach(checkIfComparable)
+
+  private def checkIfComparable(element: Any): Unit =
+    if (!element.isInstanceOf[Comparable[_]]) {
+      throw new java.lang.ClassCastException("Provided value is not comparable: " + element)
+    }
+
+  private val unknownMapType = Typed.fromDetailedType[java.util.Map[Any, Any]]
+  private val numberType = Typed.fromDetailedType[java.lang.Number]
+
+  class CollectionTyping[F[_]](implicit classTag: ClassTag[F[_]]) extends TypingFunction {
+    private val fClass: Class[F[_]] = classTag.runtimeClass.asInstanceOf[Class[F[_]]]
+
+    override def computeResultType(arguments: List[typing.TypingResult]): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+      case (f@TypedClass(`fClass`, element :: Nil)) :: _ => f.copy(params = element.withoutValue :: Nil).validNel
+      case firstArgument :: _ => firstArgument.validNel
+      case _ => GenericFunctionTypingError.ArgumentTypeError.invalidNel
+    }
+  }
+
+  class CollectionElementTyping[F[_]](implicit classTag: ClassTag[F[_]]) extends TypingFunction {
+    private val fClass: Class[F[_]] = classTag.runtimeClass.asInstanceOf[Class[F[_]]]
+
+    override def computeResultType(arguments: List[typing.TypingResult]): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+      case TypedClass(`fClass`, componentType :: Nil) :: _ => componentType.withoutValue.validNel
+      case firstArgument :: _ => firstArgument.withoutValue.validNel
+      case _ => GenericFunctionTypingError.ArgumentTypeError.invalidNel
+    }
+  }
+
+  class CollectionElementTypingForSum[F[_]](implicit classTag: ClassTag[F[_]]) extends CollectionElementTyping[F] {
+    override def computeResultType(arguments: List[typing.TypingResult]): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = {
+      super.computeResultType(arguments).map { elementType =>
+        if (elementType == Typed[java.lang.Number]) {
+          // If it is a Number, leave it as is, we will check exact types in runtime
+          elementType
+        } else {
+          ForLargeNumbersOperation.promoteSingle(elementType)
+        }
+      }
+    }
+  }
+
+  class CollectionMergeTyping[F[_]](implicit classTag: ClassTag[F[_]]) extends TypingFunction {
+    private val fClass: Class[F[_]] = classTag.runtimeClass.asInstanceOf[Class[F[_]]]
+
+    private def commonFieldHasTheSameType(fields1: ListMap[String, typing.TypingResult], fields2: ListMap[String, typing.TypingResult]) = {
+      val commonFields = fields1.keys.toSet intersect fields2.keys.toSet
+      fields1.filter { case (key, _) => commonFields.contains(key) }.view.mapValues { value => value.withoutValue }.toMap ==
+        fields2.filter { case (key, _) => commonFields.contains(key) }.view.mapValues { value => value.withoutValue }.toMap
+    }
+
+    override def computeResultType(arguments: List[typing.TypingResult]): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+      case (listType@TypedClass(`fClass`, firstComponentType :: Nil)) :: TypedClass(`fClass`, secondComponentType :: Nil) :: Nil =>
+        (firstComponentType, secondComponentType) match {
+          case (TypedObjectTypingResult(x, _, infoX), TypedObjectTypingResult(y, _, infoY)) if commonFieldHasTheSameType(x, y) =>
+            listType.copy(params = TypedObjectTypingResult(ListMap.empty ++ x.view.mapValues { value => value.withoutValue } ++ y.view.mapValues { value => value.withoutValue }.toMap, Typed.typedClass[java.util.HashMap[_, _]], infoX ++ infoY) :: Nil).validNel
+          case (_: TypedObjectTypingResult, _: TypedObjectTypingResult) =>
+            listType.copy(params = Unknown :: Nil).validNel
+          case (`unknownMapType`, _: TypedObjectTypingResult) |
+               (_: TypedObjectTypingResult, `unknownMapType`) |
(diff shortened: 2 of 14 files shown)


arkadius (commented):
Good job. Some comments added. Please add new helper here in docs: https://nussknacker.io/documentation/docs/scenarios_authoring/Spel/#built-in-helpers. Also I would add some short notice after this section: https://nussknacker.io/documentation/docs/scenarios_authoring/Spel/#mapping-lists that for all other operations on lists, you can use the `#COLLECTION` helper

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Documentation(description = "Returns a sum of all elements")
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): java.lang.Number = {
+    list.asScala.map(x => new java.math.BigDecimal(x.toString)).reduce((a, b) => a.add(b))
+  }
+
+  @Documentation(description = "Returns a list of all elements sorted in ascending order (elements must be comparable)")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def sorted[T<:Comparable[T]](@ParamName("list") list: java.util.Collection[T]): java.util.List[T] = {
```
arkadius, line 51: WDYT aboud adding `sortedAsc` and `sortedDesc` variants like for maps?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  // To address this problem we expect `java.util.List[java.util.List[T]]` instead of `java.util.Collection[java.util.Collection[T]]` to avoid auto conversion to set.
+  // This scenario is tested in [[org.p4.esp.process.services.FlattenCollectionHelperSpec]]
+  @Documentation(description = "Returns a list of all elements from all lists in the given list")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListElementTyping])
+  def flatten[T](@ParamName("list") list: java.util.List[java.util.List[T]]): java.util.List[T] = {
+    list.asScala.flatMap(_.asScala).toList.asJava
+  }
+
+  @Documentation(description = "Splits given text into a list of Strings using given regular expression")
+  def split(@ParamName("text") text: String, @ParamName("regexp") regexp: String): java.util.List[String] = {
```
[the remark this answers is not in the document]
arkadius: Also I'm wondering if it is a good place for this method. It doesn't operate on collection. IMO we should move it to e.g. `util`

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  = ListMap(map.asScala.toSeq.sortBy(_._2):_*).asJava
+
+  @GenericType(typingFunction = classOf[TypingFunctions.MapTyping])
+  @Documentation(description = "Returns a map with elements sorted by value in descending order")
+  def sortedMapByValueDesc[K, V<:Comparable[V]](map: java.util.Map[K, V]): java.util.Map[K, V]
+  = ListMap(map.asScala.toSeq.sortBy(_._2)(implicitly[Ordering[V]].reverse):_*).asJava
+
+  @Documentation(description = "Returns a list made of first n elements of the given list")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def take[T](@ParamName("list") list: java.util.List[T], @ParamName("max") max: Int): java.util.List[T]
```
arkadius, line 70: Maybe we could have takeLeft, takeRight, dropLeft, dropRight. WDYT?

raphaelsolarski (approved):

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.{Collections, Objects}
+import scala.collection.immutable.ListMap
+import scala.jdk.CollectionConverters.{CollectionHasAsScala, _}
+
+object collection extends HideToString {
+
+  @Documentation(description = "Concatenates two lists")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListAdditionTyping])
+  def concat[T](@ParamName("list1") list1: java.util.List[T], @ParamName("list2") list2: java.util.List[T]): java.util.List[T] = (list1.asScala.toList ++ list2.asScala).asJava
```
piotrp, line 14: I know that we are writing in Scala, but I believe that library-level routines should be efficient - how about pre-allocating an ArrayList of proper size and adding both lists into that?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
piotrp: Would a non-programmer understand what are "left" and "right" items in a collection? IMO "first"/"last" are better, and I'm not sure "first" is even needed. IMO take/takeLast and drop/dropLast reads better than takeLeft/takeRight and dropLeft/dropRight.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  @Documentation(description = "Returns a slice of the list starting with start index (inclusive) and ending at stop index (exclusive)")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def slice[T](@ParamName("list") list: java.util.Collection[T], @ParamName("start") start: Int, @ParamName("stop") stop: Int): java.util.List[T] = {
+    list.asScala.slice(start, stop).toList.asJava
+  }
+
+  @Documentation(description = "Returns a sum of all elements")
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): java.lang.Number = {
+    list.asScala.map(x => new java.math.BigDecimal(x.toString)).reduce((a, b) => a.add(b))
```
piotrp, line 46: Do we really need to cast all collection elements to strings?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
piotrp: (also - we don't enforce any coding style, but this line is 176 characters long, can you wrap it? or buy me 4k screen? 😉)

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
+package pl.touk.nussknacker.engine.util.functions
+
+import pl.touk.nussknacker.engine.api.generics.GenericType
+import pl.touk.nussknacker.engine.api.{Documentation, HideToString, ParamName}
+
+import java.util.{Collections, Objects}
+import scala.collection.immutable.ListMap
+import scala.jdk.CollectionConverters.{CollectionHasAsScala, _}
```
piotrp, line 8: `import scala.jdk.CollectionConverters._`

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  @Documentation(description = "Returns a copy of the list with its elements shuffled")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def shuffle[T](@ParamName("list") list: java.util.Collection[T]): java.util.List[T] = {
+    val values = new java.util.ArrayList[T](list)
+    Collections.shuffle(values)
+    values
+  }
+
+  // spel *does* auto conversion between object and collection (`a` -> {`a`}).
+  // Unfortunately, sometimes spel converts object to *set* what causes unexpected application distinct to collection.
```
piotrp, line 115: That's... not English :D

And I don't see this `org.p4.esp.process.services.FlattenCollectionHelperSpec` test.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  def merge[K, V](@ParamName("map1") map1: java.util.Map[K, V], @ParamName("map2") map2: java.util.Map[K, V]): java.util.Map[K, V] = {
+    val merged = new java.util.LinkedHashMap[K, V](map1)
+    merged.putAll(map2)
+    merged
+  }
+
+  @Documentation(description = "Returns the smallest element (elements must be comparable)")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListElementTyping])
+  def min[T<:Comparable[T]](@ParamName("list") list: java.util.Collection[T]): T = {
+    checkIfComparable(list)
```
piotrp, line 27: Can you add a test that verifies what happens when elements aren't comparable?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  def sortedMapByValueDesc[K, V<:Comparable[V]](map: java.util.Map[K, V]): java.util.Map[K, V]
+  = ListMap(map.asScala.toSeq.sortBy(_._2)(implicitly[Ordering[V]].reverse):_*).asJava
+
+  @Documentation(description = "Returns a list made of first n elements of the given list")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def take[T](@ParamName("list") list: java.util.List[T], @ParamName("max") max: Int): java.util.List[T]
+  = list.subList(0, Math.min(max, list.size()))
+
+  @Documentation(description = "Creates a string made of all elements of the list separated with the given separator")
+  def join[T](@ParamName("separator") separator: String,  @ParamName("list") list: java.util.List[T]): String
```
piotrp, line 74: Shouldn't argument order be reversed here? We're operating in the context of lists, so I'd expect it to be passed first (e.g. like in `take`)

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  @GenericType(typingFunction = classOf[TypingFunctions.MapTyping])
+  @Documentation(description = "Returns a map with elements sorted by value in descending order")
+  def sortedMapByValueDesc[K, V<:Comparable[V]](map: java.util.Map[K, V]): java.util.Map[K, V]
+  = ListMap(map.asScala.toSeq.sortBy(_._2)(implicitly[Ordering[V]].reverse):_*).asJava
+
+  @Documentation(description = "Returns a list made of first n elements of the given list")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def take[T](@ParamName("list") list: java.util.List[T], @ParamName("max") max: Int): java.util.List[T]
+  = list.subList(0, Math.min(max, list.size()))
```
piotrp, line 71: This will throw an exception if `max` is negative. `take` can accept invalid input, so this should do the same

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+    val l1 = list1.asScala.map(_.asScala)
+    val l2 = list2.asScala.map(_.asScala)
+    val res = for {
+      m1 <- l1
+      m2 <- l2
+    } yield m1 ++ m2
+    res.map(_.asJava).asJava
+  }
+
+  @Documentation(description = "Returns a list that contains all elements contained in list1, that doesn't appear in list2")
```
piotrp, line 88: doesn't -> don't

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  @Documentation(description = "Returns a list that contains all unique elements that are contained by both list1 and list2")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def intersect[T](@ParamName("list1") list1: java.util.List[T], @ParamName("list2") list2: java.util.List[T]): java.util.List[T] = {
+    (list1.asScala.toSet intersect list2.asScala.toSet).toList.asJava
+  }
+
+  @Documentation(description = "Returns a list that contains unique elements from the given list")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def distinct(@ParamName("list") list: java.util.List[Any]): java.util.List[Any] = {
```
piotrp, line 102: Why `Any` instead of a generic type?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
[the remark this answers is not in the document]
piotrp: We're exposing regex operations - shouldn't we cache compiled result?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/date.scala
```diff
@@ -1,18 +1,24 @@
...
 
   @Documentation(description = "Returns current time as an Instant")
   def now(): Instant = Instant.now(clock)
 
-  // We can't has one now() method with multiple overloaded options because ZoneOffset extends ZoneId and we can get
+  @Documentation(description = "Returns current date-time as a LocalDateTime (without a time-zone)")
+  def localDateTime(): LocalDateTime = LocalDateTime.now(clock)
+
+  @Documentation(description = "Returns current date as a LocalDate (without a time-zone)")
+  def today(): LocalDate = LocalDate.now(clock)
```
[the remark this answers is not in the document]
rafa-minimal: > So far we were trying to avoid having too many methods using default time zone
Makes sense to me, let's stick with that :) I've removed these new methods.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Good idea :+1:

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Makes sense. Let's stick to Scala/Kotlin naming convention

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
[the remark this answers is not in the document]
rafa-minimal: You are right, will move it to `util`. 

As for the optimization. I guess that a single character separator is the most common case and Java library has a short path for it: https://github.com/openjdk/jdk/blob/ad348a8cec50561d3e295b6289772530f541c6b1/src/java.base/share/classes/java/lang/String.java#L3327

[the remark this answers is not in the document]
rafa-minimal: I've also added `-1` limit parameter to disable the default behavior: "trailing empty strings will be discarded.". It seems more reasonable to me, especially when we deal with CSV row

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Documentation(description = "Returns a slice of the list starting with start index (inclusive) and ending at stop index (exclusive)")
+  @GenericType(typingFunction = classOf[TypingFunctions.ListTyping])
+  def slice[T](@ParamName("list") list: java.util.Collection[T], @ParamName("start") start: Int, @ParamName("stop") stop: Int): java.util.List[T] = {
+    list.asScala.slice(start, stop).toList.asJava
+  }
+
+  @Documentation(description = "Returns a sum of all elements")
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): java.lang.Number = {
```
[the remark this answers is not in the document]
rafa-minimal: Good idea. it now returns type T. But I think we can't leverage `NumberTypesPromotionStrategy` here. I mean:
* if the exact type of element is known at compilation type -> we return this type
* if it's `java.lang.Number` there's not much we can do.

At runtime on the other hand, all values will be coerced to the highest precision type that we find in the collection, assuming a simple ordering: `byte, short, int, long, BigInteger, float, double, BigDecimal`

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Absolutely, there is a better way :)
Now every element is coerced to the highest precision type that is found in the collection

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Thanks, fixed

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Seems like the whole comment no longer applies. I've changed the argument type to `java.util.Collection` and works as expected

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Makes sense to me :+1:

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Fixed

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: :+1:

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Good point

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+    classOf[java.lang.Byte] -> 2,
+    classOf[java.lang.Integer] -> 3,
+    classOf[java.lang.Long] -> 4,
+    classOf[java.math.BigInteger] -> 5,
+    classOf[java.lang.Float] -> 6,
+    classOf[java.lang.Double] -> 7,
+    classOf[java.math.BigDecimal] -> 8,
+  ).asInstanceOf[Map[Class[_], Int]]
+
+  private def getNumberPriority(number: Number): Int = numberPriorities
```
piotrp, line 55: `getNumberTypePriority` (and `numberPriorities`)?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): T = {
+    if (list.isEmpty) {
+      return 0.asInstanceOf[T]
+    }
+
+    val targetType = list.asScala.reduce { (a, b) =>
+      if (getNumberPriority(a) > getNumberPriority(b)) a else b
+    }
+
+    targetType match {
```
arkadius, line 97: We already have a similar thing in `MathUtils.withValuesWithTheSameType` but for two numbers. Would be nice to have this in one place. It uses `SpringNumberUtils.convertNumberToTargetClass` so no `toBigDecimal`, `toBigInteger` is needed.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+      case _: java.lang.Byte => list.asScala.map(_.byteValue()).sum.asInstanceOf[T]
+      case _: java.lang.Short => list.asScala.map(_.shortValue()).sum.asInstanceOf[T]
+      case _: Integer => list.asScala.map(_.intValue()).sum.asInstanceOf[T]
+      case _: java.lang.Long => list.asScala.map(_.longValue()).sum.asInstanceOf[T]
+      case _: java.math.BigInteger => list.asScala.map(toBigInteger(_)).reduce((a, b) => a.add(b)).asInstanceOf[T]
```
arkadius, line 102: You have a bug here and 3 lines below. Please add tests for summing one element lists for each type

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @Documentation(description = "Returns a sum of all elements")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): T = {
+    if (list.isEmpty) {
+      return 0.asInstanceOf[T]
+    }
+
+    val targetType = list.asScala.reduce { (a, b) =>
```
arkadius, line 93: 1. We could do this check only when `typingFunction` return `Number`. For statically known element type, we can reuse `ListElementTyping` logic. Thanks to that we will be sure that type in runtime is the same as during compilation. Also we would avoid one cycle of checking the list
2. You have a bug here - please add tests for one element lists
3. IMO we should behave similarly to `NumberTypesPromotionStrategy.ForLargeNumbersOperation` - I mean list of bytes should be promoted to `Long`. Otherwise, it will be easy to overflow types, because lists can contain many elements. WDYT?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+  @Documentation(description = "Returns the largest element (elements must be comparable)")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def max[T <: Comparable[T]](@ParamName("list") list: java.util.Collection[T]): T = Collections.max[T](list)
+
+  @Documentation(description = "Returns a slice of the list starting with start index (inclusive) and ending at stop index (exclusive)")
+  @GenericType(typingFunction = classOf[ListTyping])
+  def slice[T](@ParamName("list") list: java.util.Collection[T], @ParamName("start") start: Int, @ParamName("stop") stop: Int): java.util.List[T] =
+    list.asScala.slice(start, stop).toList.asJava
+
+  private val numberPriorities = Map(
```
arkadius, line 44: We already have this lists in `NumberTypesPromotionStrategy.DecimalNumbers`. Please just use `NumberTypesPromotionStrategy.promoteClasses`

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+      case x: java.math.BigInteger => x
+      case x => throw new IllegalArgumentException(s"Can't convert to BigInteger, unsupported number of type ${x.getClass.getName}: $x")
+    }
+  }
+
+  @Documentation(description = "Returns a sum of all elements")
+  @GenericType(typingFunction = classOf[ListElementTyping])
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): T = {
+    if (list.isEmpty) {
+      return 0.asInstanceOf[T]
```
piotrp, line 90: Will this work if we call this method with `List[BigDecimal]`?

docs/scenarios_authoring/Spel.md
```diff
@@ -318,11 +320,12 @@ You can also use explicit conversions that are available in utility classes and
 ## Built-in helpers 
 
 | Helper        | Functions                                      |
-| -------       | -----------                                    |
-| `GEO`         | Simple distance measurements                   |
-| `NUMERIC`     | Number parsing                                 |
+|---------------|------------------------------------------------|
+| `AGG`         | Aggregator functions                           |
```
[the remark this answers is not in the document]
arkadius: Yep, it is mistake. We have `AGG` available only as local variable injected into parameters in aggregating components.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+    Collections.sort[T](values)
+    if (desc) {
+      Collections.reverse(values)
+    }
+    values
+  }
+
+  @GenericType(typingFunction = classOf[MapTyping])
+  @Documentation(description = "Returns a map with elements sorted by value in ascending order")
+  def sortedMapByValueAsc[K, V <: Comparable[V]](@ParamName("map") map: java.util.Map[K, V]): java.util.Map[K, V] = {
```
piotrp, line 130: Do we need this method? It will work, but our UI can't display ordered maps (our JSON serializer will output it in "random" order, so users can't really test this method)

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
arkadius: Ad. 1. Let's add a TODO, that we should pass returnType to this method, so we could be sure that we return the same type as typed and info in which cases it won't work correctly (empty list with return type) or slower than necessary (statically known type)

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Makes, sense, but it was removed anyway :)

docs/scenarios_authoring/Spel.md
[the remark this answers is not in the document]
rafa-minimal: Removed `AGG`

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Thanks for pointing to the `SpringNumberUtils.convertNumberToTargetClass`. Fixed

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,278 @@
...
+    classOf[java.math.BigDecimal] -> 8,
+  ).asInstanceOf[Map[Class[_], Int]]
+
+  private def getNumberPriority(number: Number): Int = numberPriorities
+    .getOrElse(
+      number.getClass,
+      throw new IllegalArgumentException(s"Can't calculate the sum of elements. Unexpected element of type ${number.getClass.getName}: $number")
+    )
+
+  def toBigDecimal(number: Number): java.math.BigDecimal = {
```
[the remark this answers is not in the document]
rafa-minimal: Good point. But it's now removed anyway

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: I don't understand generics in Scala very well, but it seems that at runtime what the following does
```
return 0.0.asInstanceOf[T]
```
is casting `0.0` to `java.lang.Number`, no matter what the argument is.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
rafa-minimal: I;m totally convinced :) Removed

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/util.scala
```diff
@@ -1,12 +1,18 @@
...
+import scala.jdk.CollectionConverters._
 
-object util {
+object util extends HideToString {
 
   @Documentation(description = "Generate unique identifier (https://en.wikipedia.org/wiki/Universally_unique_identifier)")
   def uuid: String = UUID.randomUUID().toString
 
+  // todo: can be removed once we have a better support for Java arrays in SpEL expressions, then we could simply use `"".split("regex")`
+  @Documentation(description = "Splits given text into a list of Strings using given regular expression")
```
arkadius, line 14: WDYT about adding notice, that empty trailing empty strings won't be discarded as in standard, java String.split?
Without this I see dangerous that someone can think that it is jus "".split("x", 0)

arkadius (approved):
Please add a changelog entry about this powerful change

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/util.scala
rafa-minimal: Good idea. Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -0,0 +1,248 @@
...
+    list.asScala.slice(start, stop).toList.asJava
+
+  // TODO: This method currently has the following limitations:
+  // * it won't return the expected type for empty list
+  // * it works slower than it could with statically typed result
+  // To overcome this limitations we should supply returnType to this method
+  @Documentation(description = "Returns a sum of all elements")
+  @GenericType(typingFunction = classOf[ListElementTypingForSum])
+  def sum[T <: java.lang.Number](@ParamName("listOfNumbers") list: java.util.Collection[T]): T = {
+    val listType = Typed.fromInstance(list)
```
arkadius, line 53: It won't be a `List[Number]` or `List[Uknown]` anytime? If so, let's remove this and leave only:
```
    val targetType = 
      list.asScala.map(Typed.fromInstance(_)).reduce { (a, b) =>
        ForLargeNumbersOperation.promote(a, b)
      }
```
type computation method

rafa-minimal: `Typed.fromInstance(list)` would go through list elements, so `listType` be a more detailed type. On the other hand, this is exactly what we do here:
```
list.asScala.map(Typed.fromInstance(_)).reduce { (a, b) =>
  ForLargeNumbersOperation.promote(a, b)
}
```
So the first loop is redundant. You are right, I've left only the latter computation method.

PR #4283, opened 2023-05-10
fix oidc access token check

gadomsky:

### DIFF
--- a/docs/installation_configuration_guide/DesignerConfiguration.md
+++ b/docs/installation_configuration_guide/DesignerConfiguration.md
@@ -249,22 +249,23 @@ and provider discovery. The only supported flow is the authorization code flow w
 
 You can select this authentication method by setting the `authentication.method` parameter to `Oidc`
 
-| Parameter name                       | Importance  | Type           | Default value               | Description                                                                                                   |
-|--------------------------------------|-------------|----------------|-----------------------------|---------------------------------------------------------------------------------------------------------------|
-| authentication.issuer                | required    | url            |                             | OpenID Provider's location                                                                                    |
-| authentication.clientId              | required    | string         |                             | Client identifier valid at the authorization server                                                           |
-| authentication.clientSecret          | required    | string         |                             | Secret corresponding to the client identifier at the authorization server                                     |
-| authentication.audience              | recommended | string         |                             | Required `aud` claim value of an access token that is assumed to be a JWT.                                    |
+| Parameter name                       | Importance  | Type           | Default value               | Description                                                                                                                 |
+|--------------------------------------|-------------|----------------|-----------------------------|-----------------------------------------------------------------------------------------------------------------------------|
+| authentication.issuer                | required    | url            |                             | OpenID Provider's location                                                                                                  |
+| authentication.clientId              | required    | string         |                             | Client identifier valid at the authorization server                                                                         |
+| authentication.clientSecret          | required    | string         |                             | Secret corresponding to the client identifier at the authorization server                                                   |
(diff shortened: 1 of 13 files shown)


security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala
```diff
@@ -73,7 +73,11 @@ object OpenIdConnectProfile extends OAuth2Profile[OpenIdConnectUserInfo] {
     val userRoles =
       profile.roles ++
       getUserRoles(profile.subject.get, configuration)
-    val username = profile.preferredUsername.orElse(profile.nickname).orElse(profile.subject).get
+    val username =
+      getUserUsername(profile.subject.get, configuration)
```
arkadius, line 77: Can you add some method in `OpenIdConnectUserInfo` to avoid calling `subject.get` in various places and call it once in this place? Also let's add `.getOrElse(throw` at the end of this username determining chain

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -249,22 +249,23 @@ and provider discovery. The only supported flow is the authorization code flow w
...
+| authentication.jwksUri               | auxiliary   | url or path    | discovered                  | as above                                                                                                                    |
+| authentication.tokenCookie.name      | auxiliary   | string         |                             | name of cookie to store access token                                                                                        |
+| authentication.tokenCookie.path      | auxiliary   | string         |                             | path of access token cookie                                                                                                 |
+| authentication.tokenCookie.domain    | auxiliary   | string         |                             | domain of access token cookie                                                                                               |
+| authentication.accessTokenIsJwt      | optional    | boolean        | true                        | Access token might be of `JWT` format but it's not mandated by the `OIDC` spec                                              |
```
arkadius, line 268: I'm missing some part in this configuration option name or description. Access token might be either a `JWT` or ...? From looking into the code it looks like than it is a Bearer token that is used for obtaining the `JWT` by calling `userinfoEndpoint` . Am I right? Maybe we should call it `accessTokenPurpose` with options: `JWT | UserInfoBearerToken` or sth like this? Or at least we could improve the description

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2Profile.scala
```diff
@@ -13,4 +13,10 @@ object OAuth2Profile {
       .find(us => identity.equals(us.identity))
       .map(_.roles ++ defaults)
       .getOrElse(defaults)
+
+  def getUserUsername(identity: String, configuration: OAuth2Configuration): Option[String] =
+    configuration
+      .users
+      .find(us => identity.equals(us.identity))
```
arkadius, line 20: Why here and above we don't use `==`? :)

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -165,6 +165,11 @@ users: [
     identity: "admin"
     roles: ["Admin"]
   },
+  {
+    identity: "234a9sa98234@clientId",
```
arkadius, line 169: Do you added by mistake, or intentionally? If intentionally, I think we need to add some notice what this example is trying to tell users in this place. Without it, it looks magical. Why should I configure M2M tech account?

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2Profile.scala
arkadius: Ok, thx

arkadius (approved):

PR #4287, opened 2023-05-11
Improvements: bump cats effect

lciolecki:


piotrp (changes requested):

piotrp:
@lciolecki can you recheck performance after these changes?

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/SynchronousExecutionContextAndIORuntime.scala
```diff
@@ -0,0 +1,47 @@
...
+      scheduler = DummyScheduler,
+      shutdown = () => {},
+      config = IORuntimeConfig()
+    )
+  }
+
+  private object DummyScheduler extends Scheduler {
+
+    override def sleep(delay: FiniteDuration, task: Runnable): Runnable = {
+      throw new IllegalStateException("I'm dummy scheduled. I should not be used by the production code.")
```
raphaelsolarski, line 34: `scheduled` -> `scheduler`

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/SynchronousExecutionContextAndIORuntime.scala
```diff
@@ -0,0 +1,47 @@
...
+      blocking = syncEc,
+      scheduler = DummyScheduler,
+      shutdown = () => {},
+      config = IORuntimeConfig()
+    )
+  }
+
+  private object DummyScheduler extends Scheduler {
+
+    override def sleep(delay: FiniteDuration, task: Runnable): Runnable = {
```
raphaelsolarski, line 33: Can we leave some comment why this scheduler won't be used? 
If I understand correctly, this is because we only use flatMap/map methods on the effect in the interpreter and no time dependent callbacks.

raphaelsolarski (approved):
Good job!
I leaved a few minor comments.

Please, squash before merge 😄

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/AsyncExecutionContextPreparer.scala
```diff
@@ -2,14 +2,20 @@ package pl.touk.nussknacker.engine.api.process
 
 import scala.concurrent.ExecutionContext
 
+// TODO: rename me to ServiceExecutionContextPreparer in the future
```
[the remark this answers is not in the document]
coutoPL: yes. The `AsyncExecutionContextPreparer` trait is in the API module, so we don't want to break compatibility. But when we decide to do so, it'll be nice to rename it.

raphaelsolarski (approved):

PR #4291, opened 2023-05-12
Refactor: MethodDefinition moved from StandardObjectWithMethodDef into ComponentImplementationInvoker

arkadius:

### DIFF
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/GlobalVariableDefinitionExtractor.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/GlobalVariableDefinitionExtractor.scala
@@ -25,6 +25,8 @@ object GlobalVariableDefinitionExtractor {
       componentConfig = varWithCategories.componentConfig
     )
     StandardObjectWithMethodDef(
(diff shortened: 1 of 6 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/GlobalVariableDefinitionExtractor.scala
```diff
@@ -1,44 +1,35 @@
...
-    )
     val objectDef = ObjectDefinition(
       parameters = Nil,
       returnType = Some(returnType),
       categories = varWithCategories.categories,
       componentConfig = varWithCategories.componentConfig
     )
-    StandardObjectWithMethodDef(varWithCategories.value, methodDef, objectDef)
+    StandardObjectWithMethodDef(
+      ComponentImplementationInvoker.nullImplementationInvoker,
```
jedrz, line 28: Is it correct to pass null invoker here? Do variables ever get invoked?

arkadius: Variables are always accessed via `StandardObjectWithMethodDef.obj`, I added a comment about that

jedrz (approved):

PR #4293, opened 2023-05-15
Remove obsolete property accessors

arkadius:


arkadius:
I change my mind, we need to wait until we fully get rid of using Option wrappers, especially `EspTypeUtils.extractGenericReturnType`

PR #4294, opened 2023-05-16
Add user info to migrate scenario action

Filemon279:

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -30,6 +30,7 @@ import pl.touk.nussknacker.ui.process.deployment.DeploymentService
 import pl.touk.nussknacker.ui.process.marshall.ProcessConverter
 import pl.touk.nussknacker.ui.process.repository.FetchingProcessRepository
 import pl.touk.nussknacker.ui.process.repository.FetchingProcessRepository.FetchProcessesDetailsQuery
+import pl.touk.nussknacker.ui.process.repository.ProcessRepository.RemoteUserName
 import pl.touk.nussknacker.ui.security.api.LoggedUser
 import pl.touk.nussknacker.ui.uiresolving.UIProcessResolving
 import pl.touk.nussknacker.ui.util._
@@ -195,31 +196,18 @@ class ProcessesResources(
           }
         } ~ path("processes" / Segment / Segment) { (processName, category) =>
           authorize(user.can(category, Permission.Write)) {
-            parameter(Symbol("isSubprocess") ? false) { isSubprocess =>
-              post {
-                complete {
-                  processService
-                    .createProcess(CreateProcessCommand(ProcessName(processName), category, isSubprocess, None))
-                    .withSideEffect(response => sideEffectAction(response) { process =>
-                      OnSaved(process.id, process.versionId)
-                    })
-                    .map(toResponseEither[ProcessResponse](_, StatusCodes.Created))
-                }
-              }
-            }
-          }
-        } ~ path("migrateScenario") {
-          entity(as[CreateProcessCommand]) { command =>
-            authorize(user.can(command.category, Permission.Write)) {
-              canOverrideUsername(command.category, command.forwardedUserName)(user) {
-                post {
-                  complete {
-                    processService
-                      .createProcess(command)
-                      .withSideEffect(response => sideEffectAction(response) { process =>
-                        OnSaved(process.id, process.versionId)
-                      })
-                      .map(toResponseEither[ProcessResponse](_, StatusCodes.Created))
+            optionalHeaderValue(RemoteUserName.extractFromHeader) { remoteUserName =>
+              canOverrideUsername(category, remoteUserName)(user) {
+                parameter(Symbol("isSubprocess") ? false) { isSubprocess =>
+                  post {
+                    complete {
+                      processService
+                        .createProcess(CreateProcessCommand(ProcessName(processName), category, isSubprocess, remoteUserName))
+                        .withSideEffect(response => sideEffectAction(response) { process =>
+                          OnSaved(process.id, process.versionId)
+                        })
+                        .map(toResponseEither[ProcessResponse](_, StatusCodes.Created))
+                    }
                   }
                 }
               }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -146,14 +146,16 @@ class ProcessesResources(
               }
             } ~ (put & canWrite(processId)) {
               entity(as[UpdateProcessCommand]) { updateCommand =>
-                complete {
-                  processService
-                    .updateProcess(processId, updateCommand)
-                    .withSideEffect(response => sideEffectAction(response.toOption.flatMap(_.processResponse)) { resp =>
-                      OnSaved(resp.id, resp.versionId)
-                    })
-                    .map(_.map(_.validationResult))
-                    .map(toResponseEither[ValidationResult])
+                canOverrideUsername(processId.id, updateCommand.forwardedUserName)(ec, user) {
+                  complete {
+                    processService
+                      .updateProcess(processId, updateCommand)
+                      .withSideEffect(response => sideEffectAction(response.toOption.flatMap(_.processResponse)) { resp =>
+                        OnSaved(resp.id, resp.versionId)
+                      })
+                      .map(_.map(_.validationResult))
+                      .map(toResponseEither[ValidationResult])
+                  }
                 }
               }
             } ~ (get & skipValidateAndResolveParameter) { skipValidateAndResolve =>
@@ -197,7 +199,7 @@ class ProcessesResources(
               post {
                 complete {
                   processService
-                    .createProcess(CreateProcessCommand(ProcessName(processName), category, isSubprocess))
+                    .createProcess(CreateProcessCommand(ProcessName(processName), category, isSubprocess, None))
                     .withSideEffect(response => sideEffectAction(response) { process =>
                       OnSaved(process.id, process.versionId)
                     })
@@ -209,14 +211,16 @@ class ProcessesResources(
         } ~ path("migrateScenario") {
           entity(as[CreateProcessCommand]) { command =>
             authorize(user.can(command.category, Permission.Write)) {
-              post {
-                complete {
-                  processService
-                    .createProcess(command)
-                    .withSideEffect(response => sideEffectAction(response) { process =>
-                      OnSaved(process.id, process.versionId)
-                    })
-                    .map(toResponseEither[ProcessResponse](_, StatusCodes.Created))
+              canOverrideUsername(command.category, command.forwardedUserName)(user) {
+                post {
+                  complete {
+                    processService
+                      .createProcess(command)
+                      .withSideEffect(response => sideEffectAction(response) { process =>
+                        OnSaved(process.id, process.versionId)
+                      })
+                      .map(toResponseEither[ProcessResponse](_, StatusCodes.Created))
+                  }
                 }
               }
             }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
@@ -206,6 +206,21 @@ class ProcessesResources(
               }
             }
           }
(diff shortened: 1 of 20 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
```diff
@@ -56,7 +55,7 @@ class RemoteEnvironmentResources(remoteEnvironment: RemoteEnvironment,
           path(Segment / VersionIdSegment / "migrate") { (processName, version) =>
             (post & processId(processName)) { processId =>
               complete {
-                withProcess(processId.id, version, remoteEnvironment.migrate)
+                withProcess(processId.id, version, remoteEnvironment.migrateScenario)
```
Filemon279, line 58: That one is just for tests - should we create new endpoint (and new button on UI) for backward compatibility with old migration way?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -206,6 +206,21 @@ class ProcessesResources(
               }
             }
           }
+        } ~ path("migrateScenario") {
```
Filemon279, line 209: This could be just `createScenario`. If we could change `path("processes" / Segment / Segment)` API definition to just `path("processes")` with entity `CreateProcessCommand` we could remove this endpoint at all.

lciolecki:
Here we should allow a user to put a comment connected with the migration? Or just add some similar prefix about migration like in deployment: CANCEL / DEPLOY ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
Filemon279: Created flag to use whatever migration we want

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -36,9 +36,9 @@ import scala.language.higherKinds
 object ProcessService {
   type EmptyResponse = XError[Unit]
 
-  @JsonCodec case class CreateProcessCommand(processName: ProcessName, category: String, isSubprocess: Boolean)
+  @JsonCodec case class CreateProcessCommand(processName: ProcessName, category: String, isSubprocess: Boolean, forwardedUserName: Option[String] = None)
 
-  @JsonCodec case class UpdateProcessCommand(process: DisplayableProcess, comment: UpdateProcessComment)
+  @JsonCodec case class UpdateProcessCommand(process: DisplayableProcess, comment: UpdateProcessComment, forwardedUserName: Option[String] = None)
```
arkadius, line 41: So you are saying that any user will be able to send `UpdateProcessCommand` with any other `forwardedUserName`? :D

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -179,9 +212,9 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
     } yield validation
   }
 
-  private def saveProcess(process: DisplayableProcess, comment: UpdateProcessComment)(implicit ec: ExecutionContext): Future[Unit] = {
+  private def saveProcess(process: DisplayableProcess, comment: UpdateProcessComment, createdBy: Option[String] = None)(implicit ec: ExecutionContext): Future[Unit] = {
```
arkadius, line 215: 1. Please don't use default values. Default values are acceptable in tests, in dsl etc. In normal situation they can introduce hard to diagnose bug, that someone won't pass some value on some level. It's an opposition to statical type checking 
2. Let's keep naming consistent: let's stick with `forwardedUserName`. `forwardedUserName` became `createdBy` when it is wrapped by  `Remote[$name]`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -194,8 +227,8 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
     request(uri, method, requestEntity) flatMap f
   }
 
-  private def invokeForSuccess(method: HttpMethod, pathParts: List[String], query: Query = Query.Empty)(implicit ec: ExecutionContext): Future[XError[Unit]] =
-    invoke(method, pathParts, query) { response =>
+  private def invokeForSuccess(method: HttpMethod, pathParts: List[String], query: Query = Query.Empty, requestEntity: RequestEntity = HttpEntity.Empty)(implicit ec: ExecutionContext): Future[XError[Unit]] =
```
arkadius, line 230: The same as above

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -119,6 +127,31 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
     } yield ()).value
   }
 
+  override def migrateScenario(localProcess: DisplayableProcess, category: String)(implicit ec: ExecutionContext, loggedUser: LoggedUser): Future[Either[EspError, Unit]] = {
```
arkadius, line 130: Why not just create one method with forwardedUserName: Option ? and left TODO: make forwardedUserName required? Is this the only difference between those two methods?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -28,12 +28,20 @@ import scala.language.higherKinds
 
 object ProcessRepository {
 
+  private def forwardedNameWrapper(name: String) = s"Remote[$name]"
```
arkadius, line 31: Instead of having this method here, IMO better option would be to pass everywhere new value class `RemoteUserName` instead of `String` and in the last possible moment (https://github.com/TouK/nussknacker/pull/4294/files#diff-c41b00a7511457085374bebee61f84f69583d44160dc9fae32f44661148f58ffR78) do `action.forwardedUserName.map(_.display).getOrElse(loggedUser.username)`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
Filemon279: I've added new Permission to protect from overriding user name

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
Filemon279: I've refactored it a little so duplicated code was removed along with `migrateScenario`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
Filemon279: That is a good idea - I've added new value class :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -30,11 +31,13 @@ import scala.concurrent.{ExecutionContext, Future}
...
 trait RemoteEnvironment {
 
+  val passUsernameInMigration: Boolean = true
+
   def compare(localProcess: DisplayableProcess, remoteProcessVersion: Option[VersionId])(implicit ec: ExecutionContext) : Future[Either[EspError, Map[String, Difference]]]
 
   def processVersions(processName: ProcessName)(implicit ec: ExecutionContext) : Future[List[ProcessVersion]]
 
-  def migrate(localProcess: DisplayableProcess, category: String)(implicit ec: ExecutionContext, loggedUser: LoggedUser) : Future[Either[EspError, Unit]]
+  def migrate(localProcess: DisplayableProcess, category: String, passUsername: Boolean)(implicit ec: ExecutionContext, loggedUser: LoggedUser) : Future[Either[EspError, Unit]]
```
arkadius, line 40: Why do you have this in a parameter and as a field? (`passUsernameInMigration`)?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -112,17 +116,28 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
       _ <- processEither match {
         case Right(processDetails) if processDetails.isArchived => EitherT.leftT[Future, EspError](MigrationToArchivedError(processDetails.idWithName.name, environmentId))
         case Right(_) => EitherT.rightT[Future, EspError](())
-        case Left(RemoteEnvironmentCommunicationError(StatusCodes.NotFound, _)) => createProcessOnRemote(localProcess, category)
+        case Left(RemoteEnvironmentCommunicationError(StatusCodes.NotFound, _)) =>
+          if(passUsername) createScenarioOnRemote(localProcess, category, loggedUser) else createProcessOnRemote(localProcess, category)
```
arkadius, line 120: Add a space before bracket

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
arkadius: The name is wrong: this resource doesn't handle the whole migration process. We still do an update (`saveProcess`) on the triggering designer side in case if the scenario is missing.
Also I think that we need to have only one resource for scenario creation. WDYT about passing this user in header in the original endpoint? How bad it would be? :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -112,17 +116,28 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
         case Left(other) => EitherT.leftT[Future, EspError](other)
       }
-      _ <- EitherT.right[EspError](saveProcess(localProcess, UpdateProcessComment(s"Scenario migrated from $environmentId by ${loggedUser.username}")))
+      usernameToPass = if(passUsername) Some(RemoteUserName(loggedUser.username)) else None
+      _ <- EitherT { saveProcess(localProcess, UpdateProcessComment(s"Scenario migrated from $environmentId by ${loggedUser.username}"), usernameToPass) }
     } yield ()).value
   }
 
+  private def createScenarioOnRemote(localProcess: DisplayableProcess, category: String, loggedUser: LoggedUser)
```
arkadius, line 128: Let's have a one method (`createScenarioOnRemote`) instead of two with loggedUser as an Option

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
Filemon279: Without any reason, I've added one and forgot about a second one :facepalm: Removed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
Filemon279: :upside_down_face: Removed

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
Filemon279: When we removed `migrateScenario` it was unnecessary to have it - removed. :+1:

arkadius (approved):
Nice change

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -32,6 +33,14 @@ object ProcessRepository {
     def display: String = s"Remote[$name]"
   }
 
+  object RemoteUserName {
+    val headerName = "remoteUserNameHeader".toLowerCase
```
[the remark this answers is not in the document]
Filemon279: Changed to `Remote-User-Name` - thanks

PR #4295, opened 2023-05-18
DeploymentManager: Handling multiple deployments

arkadius:


engine/development/deploymentManager/src/main/scala/pl/touk/nussknacker/development/manager/DevelopmentDeploymentManagerProvider.scala
```diff
@@ -109,8 +109,8 @@ class DevelopmentDeploymentManager(actorSystem: ActorSystem)
 
   override def test[T](name: ProcessName, canonicalProcess: CanonicalProcess, scenarioTestData: ScenarioTestData, variableEncoder: Any => T): Future[TestProcess.TestResults[T]] = ???
 
-  override def getFreshProcessState(name: ProcessName): Future[Option[StatusDetails]] =
-    Future.successful(memory.get(name))
+  override protected def getFreshProcessStates(name: ProcessName): Future[List[StatusDetails]] =
```
lciolecki, line 112: Why right now it's protected?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -40,14 +40,12 @@ trait DeploymentManager extends AutoCloseable {
...
+  // TODO: add cancel for given deployment id
   def cancel(name: ProcessName, user: User): Future[Unit]
 
   def test[T](name: ProcessName, canonicalProcess: CanonicalProcess, scenarioTestData: ScenarioTestData, variableEncoder: Any => T): Future[TestResults[T]]
 
-  /**
-    * Gets status from the engine
-    */
-  def getProcessState(name: ProcessName)(implicit freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[Option[StatusDetails]]]
+  def getProcessStates(name: ProcessName)(implicit freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[List[StatusDetails]]]
```
lciolecki, line 48: I wonder do we really need here a list of the states? This list is only specific for the Flink.. Maybe better will be still return Option[StatusDetails] and in the Flink implementation use `InconsistentStateDetector.extractAtMostOneStatus`? In fact, at the end we have tun invoke `extractAtMostOneStatus`.

engine/development/deploymentManager/src/main/scala/pl/touk/nussknacker/development/manager/DevelopmentDeploymentManagerProvider.scala
arkadius: IDE automatically add protected when I override this method. Why it have to be public? `getFreshProcessStates` is not meant to be publically available, `getProcessStates` is for public usage

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
arkadius: As we discussed offline, List will stay here

Filemon279 (approved):
LGTM :+1:

PR #4305, opened 2023-05-19
Update flink-scala

piotrp:
Simplify Flink deployment: `scala-compiler` and `scala-reflect` are now included in `flink-scala` (https://github.com/TouK/flink-scala-2.13/pull/2)


gadomsky (approved):
thanks

PR #4313, opened 2023-05-22
Fix: sink json schema int / long to BigDecimal (#4289)

lciolecki:


arkadius (approved):

arkadius:
Let's add the changelog entry

PR #4326, opened 2023-05-24
Typer always typing the whole expression

arkadius:

### DIFF
--- a/engine/flink/executor/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
+++ b/engine/flink/executor/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
@@ -246,7 +246,7 @@ class CustomNodeProcessSpec extends AnyFunSuite with Matchers with ProcessTestHe
 
     val thrown = the [IllegalArgumentException] thrownBy processInvoker.invokeWithSampleData(process, data)
 
-    thrown.getMessage shouldBe "Compilation errors: ExpressionParserCompilationError(Unresolved reference 'input',proc2,Some(all),#input.id)"
+    thrown.getMessage should startWith ("Compilation errors: ExpressionParserCompilationError(Unresolved reference 'input',proc2,Some(all),#input.id)")
   }
 
   test("should validate types in custom node output variable") {
--- a/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
+++ b/interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
@@ -1,8 +1,7 @@
 package pl.touk.nussknacker.engine.spel
 
-import cats.data.NonEmptyList._
 import cats.data.Validated._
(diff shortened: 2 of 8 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -464,32 +492,32 @@ private[spel] class Typer(commonSupertypeFinder: CommonSupertypeFinder,
...
-          .map(Valid(_))
-          .getOrElse(NoPropertyError(t, e.getName).invalidNel)
+          .map(Writer(List.empty[ExpressionParseError], _))
+          .getOrElse(Writer(List(NoPropertyError(t, e.getName)), Unknown))
       case TypedObjectTypingResult(fields, objType, _) =>
         val typeBasedOnFields = fields.get(e.getName)
         typeBasedOnFields.orElse(propertyTypeBasedOnMethod(objType, e))
-          .map(Valid(_))
-          .getOrElse(NoPropertyError(t, e.getName).invalidNel)
+          .map(Writer(List.empty[ExpressionParseError], _))
```
gadomsky, line 519: maybe we can extract method `def valid[T](result: T): Writer[...` for creating Writer with empty list? 
I think it would be easier to read this code then.

And maybe also `def invalid[T](errors: List[ExpressionParseError], result: T): Writer[..`

engine/flink/executor/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
```diff
@@ -246,7 +246,7 @@ class CustomNodeProcessSpec extends AnyFunSuite with Matchers with ProcessTestHe
 
     val thrown = the [IllegalArgumentException] thrownBy processInvoker.invokeWithSampleData(process, data)
 
-    thrown.getMessage shouldBe "Compilation errors: ExpressionParserCompilationError(Unresolved reference 'input',proc2,Some(all),#input.id)"
+    thrown.getMessage should startWith ("Compilation errors: ExpressionParserCompilationError(Unresolved reference 'input',proc2,Some(all),#input.id)")
```
gadomsky, line 249: maybe we can assert with "shouldBe" to show that there are actually more then one error? Or have assertion for multiple errors in other test?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
arkadius: Good point, added both. Also, I removed unnecessary wrapping with `TypingResultWithContext` which was done in many places but could be done in only a few once where it is imporant.

engine/flink/executor/src/test/scala/pl/touk/nussknacker/engine/process/functional/CustomNodeProcessSpec.scala
arkadius: I added a new test for that

gadomsky (approved):
LGTM

PR #4333, opened 2023-05-26
Split cross-scala-version tests in seprate jobs

gadomsky:

### DIFF
--- a/ciRunSbt.sh
+++ b/ciRunSbt.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -e
 #if $CROSS_BUILD is true - we do cross build, otherwise we use default scala version
-if [[ $CROSS_BUILD == 'true' ]]
-then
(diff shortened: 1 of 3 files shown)


lciolecki (approved):
LGTM

ciRunSbt.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -e
 #if $CROSS_BUILD is true - we do cross build, otherwise we use default scala version
-if [[ $CROSS_BUILD == 'true' ]]
-then
+if [[ -n "$FORCE_SCALA_VERSION" ]]; then
```
arkadius, line 4: why it is "force"? Why not just "SCALA_VERSION"? because we already have defined `defaultScalaVersion`? Also we already have `NUSSKNACKER_SCALA_VERSION`, it become to be complicated

PR #4334, opened 2023-05-26
Rethinking components usage calculation

DeamonDev:
The aim of this PR is to change presentation layer of "Usages" functionality in "Components" tab. Currently when one look for all the usages for a particular component the corresponding outcome contains information about usages of this component in scenarios and fragments. Unfortunately when component is used in fragment inside scenario we don't see this usage in the corresponding scenario row (in usages tab).


designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -25,11 +27,43 @@ object ComponentsUsageHelper {
...
     }
 
-    processesDetails
+    @tailrec
+    def mergeFragments(fragmentsProcessesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]],
+                       acc: Map[ComponentId, List[(BaseProcessDetails[Unit], List[NodeId])]]): Map[ComponentId, List[(BaseProcessDetails[Unit], List[NodeId])]] =
+      fragmentsProcessesDetails match {
+        case Nil => acc
+        case fragmentProcessesDetails :: t => {
+          val componentIdUsages = toComponentIdUsages(fragmentProcessesDetails).toMap
```
arkadius, line 36: Can't you just compute it once before this method? It will be faster and will have only arguments that are changeable. Also let's rename it to sth with "fragment" in name

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -25,11 +27,43 @@ object ComponentsUsageHelper {
       }
     }
 
-    processesDetails
+    @tailrec
+    def mergeFragments(fragmentsProcessesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]],
+                       acc: Map[ComponentId, List[(BaseProcessDetails[Unit], List[NodeId])]]): Map[ComponentId, List[(BaseProcessDetails[Unit], List[NodeId])]] =
+      fragmentsProcessesDetails match {
```
arkadius, line 33: What gives you this pattern matching? It won't give a big performance improvement and instead obfuscates the code a bit

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -25,11 +27,43 @@ object ComponentsUsageHelper {
       }
     }
 
-    processesDetails
+    @tailrec
+    def mergeFragments(fragmentsProcessesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]],
```
arkadius, line 31: Why did you do this recursively? Currently we have only two levels (we don't support fragments inside fragments). Is it for the future when we get fragments nesting?

arkadius (commented):
Where is a test for that? :) Please do TDD

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
DeamonDev: It's for another reason. Firstly I compute all scenario component usages by splitting all processes to scenario and fragment processes. Then for each fragment process I merge its component usages into these already computed scenario usages.

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
DeamonDev: Note that I compute it per `fragment`, not `fragments`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
DeamonDev: I wanted to avoid `if-else` statement and did list `head-tail` destruct. What you suggest instead?

arkadius: Sorry, I misunderstand it. It is related with my another question about `componentIdUsages` preparation

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
arkadius: If so, it should be possible to have a semigroup for componentId -> List[NodeId], compute it for usages in scenarios, flatmap usages for fragments for each scenario and combine them. WDYT? IMO it would be easier to read, will be linear

DeamonDev: I think it is good idea :)

DeamonDev: @arkadius I refactored the code with a hope it is more readable now

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -15,21 +16,32 @@ object ComponentsUsageHelper {
...
+      processDetails <- processesDetails
+      componentIdNodesPair <- processDetails.json.value.toList
+      (ComponentIdParts(componentName, componentType), nodeIds) = componentIdNodesPair
+      componentId = componentIdProvider.createComponentId(processDetails.processingType, componentName, componentType)
+      nodeId <- nodeIds
+    } yield ScenarioComponentsUsage(componentId, componentType, componentName, processDetails, nodeId)
+
+    val scenariosComponentUsagesFlatten = flattenUsages(processesDetails.filter(_.isSubprocess == false))
+    val fragmentsComponentUsagesFlattenMap = flattenUsages(processesDetails.filter(_.isSubprocess == true))
+      .groupBy(_.processDetails.name).mapValuesNow(_.map(u => u.copy(nodeId = s"<<fragment>> ${u.nodeId}")))
```
lciolecki, line 30: Can you extract this prefix to const?

DeamonDev: It'll not live long enough (see my newest comment)

lciolecki:
Why not create an `@ConfiguredJsonCodec sealed trait` and once time return just `nodeId` another time `fragmentNodeId: String, nodeId: NodeId`? There can be a lot of nodes, so in that way you will reduce `fragmentNodeId: null`.

designer/server/src/test/scala/pl/touk/nussknacker/ui/component/DefaultComponentServiceSpec.scala
```diff
@@ -499,14 +499,13 @@ class DefaultComponentServiceSpec extends AnyFlatSpec with Matchers with Patient
       (admin, fraudCustomerDataEnricherComponentId, List((CanceledFraudProcessWith2Enrichers, List(DefaultCustomName, SecondCustomName)))),
       (admin, filterComponentId, List(
         (DeployedFraudProcessWith2Filters, List(DefaultFilterName, SecondFilterName)),
-        (FraudProcessWithSubprocess, List(SecondFilterName)),
-        (FraudSubprocess, List(SubprocessFilterName)),
+        (FraudProcessWithSubprocess, List(SecondFilterName, s"<<fragment>> $SubprocessFilterName"))
```
lciolecki, line 502: Why did you change it?

DeamonDev: That is because at this moment I show only component usages for scenarios. That's why I removed line 503.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/component/package.scala
```diff
@@ -16,6 +21,20 @@ package object component {
 
   type NodeId = String
 
+  @ConfiguredJsonCodec sealed trait NodeMetadata {
```
arkadius, line 24: Maybe let's call it `NodeUsageId`/`NodeUsageRef`/`NodeUsageData`
to be visible that it is related with usages mechanism. Node metadata looks like some generic metadata e.g. component reference, link documentation etc.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/component/package.scala
```diff
@@ -16,6 +21,20 @@ package object component {
...
+    def getNodeId: NodeId
```
arkadius, line 25: Regarding scala style, it should be just `def nodeId: Node`, to be easy to migrate between def/lazy val/val

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -15,21 +17,39 @@ object ComponentsUsageHelper {
...
-        componentId -> (processDetails.mapProcess(_ => ()), nodeIds)
-      }
+                             processesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]]): Map[ComponentId, List[(BaseProcessDetails[_], List[NodeMetadata])]] = {
+    def flattenUsages(processesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]]) = for {
+      processDetails <- processesDetails
+      componentIdNodesPair <- processDetails.json.value.toList
+      (ComponentIdParts(componentName, componentType), nodeIds) = componentIdNodesPair
+      componentId = componentIdProvider.createComponentId(processDetails.processingType, componentName, componentType)
+      nodeId <- nodeIds
+      nodeMetadata = if (processDetails.isSubprocess) FragmentNodeMetadata(processDetails.name, nodeId) else ScenarioNodeMetadata(nodeId)
```
arkadius, line 27: Something is not yes here :) Why you cheating and putting scenario name into place where should be nodeId (fragmentNodeId)? IMO we would have `nodeId: String` after this step. You should rewrite `String` into `NodeMetadata` in next step when you resolve usages of fragments. OFC here:
```
      case usageOfOtherComponentType =>
        List(usageOfOtherComponentType)
```
will be necessary to replace nodeId by `ScenarioNodeMetadata` as well.

Other option I could see is to have more generic purpose structure - `NodeIdPath` which at the beginning will be one element list, and during resolving of fragment usages become `fragmentNodeId :: nodeId :: Nil`. On FE would be some logic that if the path has > 1 element1, then has other color.  Thanks to this approach it will be easier to migrate into nesting of fragments in the future

DeamonDev: > IMO we would have nodeId: String after this step. You should rewrite String into...

And indeed it is rewritten here:

```
val scenarioUsagesWithResolvedFragments = scenariosComponentUsagesFlatten.flatMap {
   case fragmentUsage@ScenarioComponentsUsage(_, ComponentType.Fragments, Some(fragmentName), processDetails, ScenarioNodeMetadata(fragmentNodeId)) =>
    fragmentUsage :: fragmentsComponentUsagesFlattenMap.get(fragmentName).toList.flatten.collect {
       case u@ScenarioComponentsUsage(componentId, componentType, componentName, _, FragmentNodeMetadata(_, nodeId)) =>
        u.copy(processDetails = processDetails, nodeMetadata = FragmentNodeMetadata(fragmentNodeId, nodeId))
    }
```

And I'm cheating here because in previous approach I had `NodeMetadata` structure with `Option` field for its fragment node id. With that solution I set fragment node id to `None` and rewritten it to `Some(...)` in the code above.

designer/submodules/packages/components/src/components/usages/usagesTable.tsx
```diff
@@ -67,8 +67,8 @@ export function UsagesTable(props: TableViewData<ComponentUsageType>): JSX.Eleme
                 sortingOrder: ["desc", "asc", null],
             },
             {
-                field: "nodesId",
-                headerName: t("table.usages.title.NODES_ID", "Nodes"),
+                field: "nodesMetadata",
+                headerName: t("table.usages.title.NODES_METADATA", "Nodes"),
```
lciolecki, line 71: Why did you change `table.usages.title.NODES_ID`? This placeholder hasn't been changed.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/component/package.scala
```diff
@@ -16,6 +21,20 @@ package object component {
...
+  }
+  object NodeMetadata {
+    case class FragmentNodeMetadata(fragmentNodeId: String, nodeId: NodeId) extends NodeMetadata {
+      override def getNodeId: NodeId = this.nodeId
+    }
+
+    case class ScenarioNodeMetadata(nodeId: NodeId) extends NodeMetadata {
```
lciolecki, line 32: `override nodeId: NodeId` is enough. You havn''t add this:
```
override def getNodeId: NodeId = this.nodeId
```

designer/submodules/packages/components/src/components/usages/nodesCell.tsx
```diff
@@ -19,37 +20,52 @@ export const NodesCell = ({ filterText, ...props }: GridRenderCellParams & { fil
...
-            size="small"
-            component={ExternalLink}
-            href={nodeHref(rowId, node)}
-            tabIndex={0}
-            label={matched > 0 ? <Highlight value={node} filterText={filterText} /> : node}
-            color={matched !== 0 ? "primary" : "default"}
-            variant={matched > 0 ? "outlined" : "filled"}
-            icon={icon}
-        />
+        nodeType === "FragmentUsageData" ?
```
JulianWielga, line 49: diff with one or two props? duplicated 10 lines

designer/client/src/http/HttpService.ts
```diff
@@ -93,11 +93,17 @@ export type SourceWithParametersTest = {
...
+    nodeId: string;
+    type: string;
+};
+
 export type ComponentUsageType = {
     id: string;
     name: string;
     processId: string;
-    nodesId: string[];
+    nodesMetadata: NodeUsageData[];
```
arkadius, line 106: meta -> usage

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentService.scala
```diff
@@ -76,7 +76,7 @@ class DefaultComponentService private(componentLinksConfig: ComponentLinksConfig
 
         componentsUsage
           .get(componentId)
-          .map(data => Right(data.map { case (process, nodesId) => ComponentUsagesInScenario(process, nodesId) }.sortBy(_.id)))
+          .map(data => Right(data.map { case (process, nodesMetadata) => ComponentUsagesInScenario(process, nodesMetadata) }.sortBy(_.id)))
```
arkadius, line 79: meta -> usage

designer/server/src/test/scala/pl/touk/nussknacker/ui/component/DefaultComponentServiceSpec.scala
```diff
@@ -489,24 +491,25 @@ class DefaultComponentServiceSpec extends AnyFlatSpec with Matchers with Patient
...
+        (FraudProcessWithSubprocess, List(ScenarioUsageData(SecondFilterName), FragmentUsageData(FraudSubprocess.name, SubprocessFilterName))),
       )),
     )
 
-    forAll(testingData) { (user: LoggedUser, componentId: ComponentId, expected: List[(BaseProcessDetails[_], List[String])] ) =>
+    forAll(testingData) { (user: LoggedUser, componentId: ComponentId, expected: List[(BaseProcessDetails[_], List[NodeUsageData])] ) =>
       val result = defaultComponentService.getComponentUsages(componentId)(user).futureValue
-      val componentProcesses = expected.map{ case (process, nodesId) => ComponentUsagesInScenario(process, nodesId) }
+      val componentProcesses = expected.map {
+        case (process, nodesMetadata) => ComponentUsagesInScenario(process, nodesMetadata)
```
arkadius, line 511: meta -> usage

designer/submodules/packages/components/src/components/usages/nodesCell.tsx
```diff
@@ -19,35 +20,40 @@ export const NodesCell = ({ filterText, ...props }: GridRenderCellParams & { fil
         [filterSegments],
     );
 
-    const sorted = useMemo(() => value.map((node) => [countMatches(node), node]).sort(([a], [b]) => b - a), [countMatches, value]);
+    const sorted = useMemo(() => value.map((node: NodeUsageData) => [countMatches(node.nodeId), node.nodeId, node.fragmentNodeId, node.type]).sort(([a], [b]) => b - a), [countMatches, value]);
 
-    const elements = sorted.map(([match, node]) => (
-        <NodeChip key={node} icon={icon} node={node} filterText={filterText} rowId={id} matched={filterText ? match : -1} />
+    const elements = sorted.map(([match, nodeId, fragmentNodeId, nodeType]) => (
```
arkadius, line 25: This order is correct? We want to sort by nodeId / fragmentNodeId instead of fragmentNodeId / nodeId? We you have needed to add nodeType as a one more orderBy element?

designer/server/src/test/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelperTest.scala
```diff
@@ -168,7 +168,12 @@ class ComponentsUsageHelperTest extends AnyFunSuite with Matchers with TableDriv
     )
 
     forAll(table) { (processesDetails, expected) =>
+      import pl.touk.nussknacker.engine.util.Implicits._
+
       val result = ComponentsUsageHelper.computeComponentsUsage(defaultComponentIdProvider, withComponentsUsages(processesDetails))
+        .mapValuesNow(_.map {
+          case (baseProcessDetails, nodeIds) => (baseProcessDetails.mapProcess(_ => ()), nodeIds.map(_.nodeId))
```
arkadius, line 175: I think that we shouldn't do _.nodeId extraction. Instead we should change expectations to make sure that everything works as expected

designer/submodules/packages/components/src/components/usages/nodesCell.tsx
DeamonDev: I corrected that. Also I removed `nodeType` from it and replaced it with simple `null`-check. I don't think such a data pass into component is good design choice, because every new data is currently taken into consideration by sorting function.

designer/server/src/test/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelperTest.scala
DeamonDev: Done :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentsUsageHelper.scala
```diff
@@ -15,21 +17,39 @@ object ComponentsUsageHelper {
...
+                             processesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]]): Map[ComponentId, List[(BaseProcessDetails[_], List[NodeUsageData])]] = {
+    def flattenUsages(processesDetails: List[BaseProcessDetails[ScenarioComponentsUsages]]) = for {
+      processDetails <- processesDetails
+      componentIdNodesPair <- processDetails.json.value.toList
+      (ComponentIdParts(componentName, componentType), nodeIds) = componentIdNodesPair
+      componentId = componentIdProvider.createComponentId(processDetails.processingType, componentName, componentType)
+      nodeId <- nodeIds
+    } yield ScenarioComponentsUsage[NodeId](componentId, componentType, componentName, processDetails, nodeId)
+
+    val scenariosComponentUsagesFlatten = flattenUsages(processesDetails.filter(_.isSubprocess == false))
```
lciolecki, line 29: scenariosComponentUsagesFlatten -> scenariosComponentUsages
fragmentsComponentUsagesFlattenMap -> fragmentsComponentUsages

arkadius (approved):
LGTM

arkadius (approved):
LGTM

PR #4339, opened 2023-05-29
Improvements: Don't fetch state for archived/unarchived scenario

lciolecki:


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -222,33 +223,59 @@ class DeploymentServiceImpl(dispatcher: DeploymentManagerDispatcher,
...
         logger.debug(s"Status for: '${processDetails.name}' is: ${SimpleStateStatus.DuringCancel}")
         DBIOAction.successful(manager.processStateDefinitionManager.processState(SimpleStateStatus.DuringDeploy))
       } else if (inProgressActionTypes.contains(ProcessActionType.Cancel)) {
         logger.debug(s"Status for: '${processDetails.name}' is: ${SimpleStateStatus.DuringCancel}")
         DBIOAction.successful(manager.processStateDefinitionManager.processState(SimpleStateStatus.DuringCancel))
       } else {
-        checkStateInDeploymentManager(manager, processDetails)
+        processDetails.lastAction match {
+          case Some(action) if action.isUnArchived =>
+            getProcessStateForArchiveActions(processDetails, stateLastActionType)(manager)
```
arkadius, line 254: Why we have to handle this otherwise than in a normal situation?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessActionRepository.scala
```diff
@@ -146,6 +146,19 @@ extends Repository[F] with EspTables with CommentActions with ProcessActionRepos
     run(query.result.map(_.toSet))
   }
 
+  def getLastStateActionType(processId: ProcessId): F[Option[ProcessActionType]] = {
```
arkadius, line 149: Let's have a more general method `getLastFinishedActionTypes(ProcessId, Set[ActionType])` and move this upper to the place of usage, that:
1) will be clear why this set is important
2) will be the chance that we will use it in more places

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessAction.scala
```diff
@@ -29,7 +31,8 @@ object ProcessActionType extends Enumeration {
   val UnArchive: Value = Value("UNARCHIVE")
   val Pause: Value = Value("PAUSE") //TODO: To implement in future..
 
-  val defaultActions: List[ProcessActionType] = Nil
+  val DefaultActions: List[ProcessActionType] = Nil
+  val StateActions = List(Cancel, Deploy)
```
arkadius, line 35: I would inline this in the place of usage so it will be more clear why this set contains only cancel and deploy and also add a comment next to it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -259,10 +260,15 @@ class DeploymentServiceImpl(dispatcher: DeploymentManagerDispatcher,
 
   //We assume that checking the state for archived doesn't make sense, and we compute the state based on the last state action
   private def getArchivedProcessState(processDetails: BaseProcessDetails[_])(implicit manager: DeploymentManager) = {
-    processDetails.lastStateAction match {
-      case Some(_) => //We assume that cancellation is required before the archiving process
+    processDetails.lastStateAction.map(_.action) match {
+      case Some(Cancel) =>
         logger.debug(s"Status for: '${processDetails.name}' is: ${SimpleStateStatus.Canceled}")
         DBIOAction.successful(manager.processStateDefinitionManager.processState(SimpleStateStatus.Canceled))
+      case Some(Deploy) =>
```
arkadius, line 267: IMO this Some(Deploy) and Some(_) is the same case. We have some inconsistency and we should decide how to handle it. One way is thow an exception and let's user inform admin to fix this. Other way is to return `Problem` and log error. Pick one of them.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessAction.scala
arkadius: After rethinking, we have to move this from `extensions-api` to `designer`. Now I see that it is in an API module. It can't be here:
- Nobody writing and extensions suppose to use it, we don't have such a use case
- Each change in an API module should be described in MigrationGuide. We don't want to do this.
Please move it to `DBFetchingProcessRepository` where it is only used.
 
BTW change above should be described in MigrationGuide as well.

lciolecki: I disagree, when I move `StateActions` then we don't change API, so there is no need to change MigrationGuide? But I've added some info about the change.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -237,27 +238,50 @@ class DeploymentServiceImpl(dispatcher: DeploymentManagerDispatcher,
...
   }
 
+  //We assume that checking the state for archived doesn't make sense, and we compute the state based on the last state action
+  private def getArchivedProcessState(processDetails: BaseProcessDetails[_])(implicit manager: DeploymentManager) = {
+    processDetails.lastStateAction.map(_.action) match {
+      case Some(Cancel) =>
+        logger.debug(s"Status for: '${processDetails.name}' is: ${SimpleStateStatus.Canceled}")
+        DBIOAction.successful(manager.processStateDefinitionManager.processState(SimpleStateStatus.Canceled))
+      case Some(_) =>
+        logger.debug(s"Status for: '${processDetails.name}' is: ${ProblemStateStatus.archivedShouldBeCanceled}")
```
arkadius, line 268: change to warn level

arkadius (approved):

PR #4343, opened 2023-05-31
Update Flink to 1.16.2

piotrp:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -281,7 +281,7 @@ lazy val commonSettings =
       )
     )
 
-val flinkV = "1.16.1"
+val flinkV = "1.16.2"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -281,7 +281,7 @@ lazy val commonSettings =
       )
     )
 
-val flinkV = "1.16.1"
+val flinkV = "1.16.2"
```
gadomsky, line 284: In tests we also have Dockerfile which is based on Flink docker image (1.16.1)

lciolecki (approved):

build.sbt
piotrp: Thanks, fixed

PR #1532, opened 2021-04-16
TypedObjectTypingResult keeps fields order

nadberezny:

PR #4352, opened 2023-06-06
TypedObjectTypingResult - use regular Map in place of ListMap, fields order no longer matters

rafa-minimal:
Let's have a consistent representation of TypedObject with keys sorted.

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
@@ -7,7 +7,6 @@ import pl.touk.nussknacker.engine.api.typed.TypingType.{TypingType, decoder}
 import pl.touk.nussknacker.engine.api.typed.typing._
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 
-import scala.collection.immutable.ListMap
 import scala.util.{Failure, Success, Try}
 
 //TODO: refactor way of encoding to easier handle decoding.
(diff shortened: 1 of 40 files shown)


bartektartanus (approved):

arkadius (changes requested):
Why? We have defined order, fields are in ListMap - taka a look at warning above the class

rafa-minimal:
I didn't really understand the reasons behind it. Ok, let's make things simple. Merge will return a type definition with keys sorted. Seems the best option for me. WDYT?

arkadius (approved):
I think, that it LGTM :)

arkadius (changes requested):
I have to revert my  :+1: I see that you changed how the result is typed, but you didn't change the implementation. If you change the implementation, the merge will change complexity from O(n+m) into O((n+m)log(n+m)). Do you think that it is worth it :thinking: (I'm not trying to state an opinion about that)

rafa-minimal:
All right, it required a little bit of analysis, but we've come to a solution.

The reason `TypedObjectTypingResult` was using `ListMap` is that the frontend form was generated based on the `TypingResult` and we wanted to keep fields in order (see #1532). This is no longer the case (see `AvroSchemaBasedParameter`), so we can put back a regular `Map` in place of `ListMap`.

This change reverts #1532.

Additionally let's sort `TypedObjectTypingResult` keys on display:
- to keep it consistent,
- to make it easier to compare two types or find a particular field in a large type definition.

engine/flink/tests/src/test/resources/extractedTypes/defaultModel.json
```diff
@@ -13128,7 +13128,7 @@
       ],
       "merge": [
         {
-          "description": "Merges two maps. Values in the first map will be overwritten with values from the second map if keys are the same",
+          "description": "Merges two maps. Values in the first map will be overwritten with values from the second map if keys are the same. The keys for the resulting type definition will be sorted.",
```
arkadius, line 13131: let's remove this pat of text about sorted keys

arkadius (approved):
LGTM

engine/flink/tests/src/test/resources/extractedTypes/defaultModel.json
rafa-minimal: Thanks!

PR #4354, opened 2023-06-06
Draft: Refactor: "subprocess" -> "fragment"

Filemon279:


designer/client/src/windowManager/WindowKind.tsx
```diff
@@ -1,18 +1,18 @@
...
-  customAction,
-  genericAction,
-  survey,
+    default,
+    confirm,
+    inform,
+    viewNode,
+    editNode,
+    addProcess,
+    addfragment,
```
JulianWielga, line 8: case lost

JulianWielga (approved):

designer/client/src/windowManager/WindowKind.tsx
Filemon279: Thanks, fixed

PR #4426, opened 2023-06-27
Show fragments in component usages

DeamonDev:
The aim of this PR is to show fragments in component usages tab. There will be added also "Other" filter to specify which entries (scenarios with fragments usages or sole fragments) should be listed.


gadomsky (approved):
LGTM

PR #4435, opened 2023-06-28
Fix: resolve references in Swagger 2.0

mslabek:
In SwaggerParser when we get a Swagger 2.0 definition which has references which we are using later (such as in parameters or response body), we are not able to properly parse it and convert to SwaggerService. This is because we are passing empty parser config to a dedicated Swagger 2.0 parser which has reference resolving turned off. This problem is not seen in OpenApi 3.0, because we are not passing any config, and it default to resolving refs.

I propose using a unified OpenAPIParser, since it is a higher level interface which parses Swagger 2.0 and Openapi definitions. It gets the same parser config with resolution turned on.


gadomsky (approved):
Nice! Can you please add entry in changelog?

PR #4450, opened 2023-07-03
Workaround for flaky fragments test: type expression without dot

arkadius:


arkadius:
Closing in favour of #4452

PR #4452, opened 2023-07-03
Ace editor bump 1.4.12 -> 1.23.1

arkadius:
What doesn't work:
- javadocs doesn't change live after method name change - ctrl+space hit is necessary to do that

Patch for react-ace was added because of https://github.com/securingsincity/react-ace/pull/1833


designer/client/package.json
```diff
@@ -68,7 +69,7 @@
         "prop-types": "15.8.1",
         "query-string": "7.1.0",
         "react": "17.0.2",
-        "react-ace": "8.0.0",
+        "react-ace": "10.1.0",
```
bartektartanus, line 72: Maybe add some "TODO: review patches when upgrading this dependency"?

bartektartanus (approved):
I've added one suggestion that you can take into consideration.

designer/client/package.json
arkadius: In json? :)

Dzuming (approved):

PR #4455, opened 2023-07-04
Improvement: Allow to run TestScenarioRunner in TestRuntime mode

lciolecki:

### DIFF
--- a/utils/components-testkit/src/main/scala/pl/touk/nussknacker/engine/util/test/TestScenarioRunner.scala
+++ b/utils/components-testkit/src/main/scala/pl/touk/nussknacker/engine/util/test/TestScenarioRunner.scala
@@ -4,6 +4,8 @@ import cats.data.ValidatedNel
 import pl.touk.nussknacker.engine.api.component.ComponentDefinition
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError
 import pl.touk.nussknacker.engine.api.exception.NuExceptionInfo
+import pl.touk.nussknacker.engine.api.process.ComponentUseCase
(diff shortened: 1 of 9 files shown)


utils/components-testkit/src/main/scala/pl/touk/nussknacker/engine/util/test/TestScenarioRunner.scala
```diff
@@ -45,8 +47,12 @@ trait TestScenarioRunner
 
 trait TestScenarioRunnerBuilder[R <: TestScenarioRunner, B <: TestScenarioRunnerBuilder[R, _]] {
 
+  def componentUseCase(testRuntimeMode: Boolean): ComponentUseCase = if (testRuntimeMode) TestRuntime else EngineRuntime
```
arkadius, line 50: Let's move it to some utils. It's make no sense to put here method that takes some argument that can be not related with `testRuntimeMode` set by `inTestRuntimeMode` method. IMO we should transform this boolean to `ComponentUseCase` in the last possible moment

utils/components-testkit/src/main/scala/pl/touk/nussknacker/engine/util/test/TestScenarioRunner.scala
```diff
@@ -45,8 +47,12 @@ trait TestScenarioRunner
...
+
   def withExtraComponents(components: List[ComponentDefinition]): B
 
+  def inTestRuntimeMode(): B
```
arkadius, line 54: Let's remove parentheses - it has no side effect. `build` has parentheses because the name is in imperative form. We can discuss removing them if it is confusing

arkadius (approved):

PR #4456, opened 2023-07-04
Fix Designer startup on Windows

piotrp:


arkadius (approved):
LGTM but don't know if @JulianWielga will be happy

PR #4463, opened 2023-07-05
SpelTemplate expression suggestions

bartektartanus:


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/parser/NuTemplateAwareExpressionParser.java
Filemon279, line 1: Shouldn't this be placed under `java/...` rather then `scala/` ?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
```diff
@@ -167,25 +170,36 @@ class SpelExpressionSuggester(expressionConfig: ExpressionDefinition[_], typeDef
...
+  private val parser = new NuTemplateAwareExpressionParser(new SpelParserConfiguration)
+
+  def parse(input: String, language: String, position: Int, variables: Map[String, TypingResult]): Try[Option[(NuSpelNode, Int)]] = {
+    val rawExpression = if (language == Expression.Language.Spel) {
+      Try(parser.parseExpression(input, null))
+    } else if (language == Expression.Language.SpelTemplate) {
+      Try(parser.parseExpression(input, new TemplateParserContext()))
+    } else {
+      Failure(new IllegalArgumentException(s"Language $language is not supported"))
+    }
```
Filemon279, line 182: That looks like a perfect place for `match` and `case` :)

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
```diff
@@ -167,25 +170,36 @@ class SpelExpressionSuggester(expressionConfig: ExpressionDefinition[_], typeDef
...
-
-  def parse(input: String, variables: Map[String, TypingResult]): Try[NuSpelNode] = {
-    val rawSpelExpression = Try(parser.parseRaw(input))
-    rawSpelExpression.map { parsedSpelExpression =>
-      val collectedTypingResult = typer.doTypeExpression(parsedSpelExpression, ValidationContext(variables))._2
-      val nuAst = new NuSpelNode(parsedSpelExpression.getAST, None, collectedTypingResult)
-      nuAst
+  private val parser = new NuTemplateAwareExpressionParser(new SpelParserConfiguration)
+
+  def parse(input: String, language: String, position: Int, variables: Map[String, TypingResult]): Try[Option[(NuSpelNode, Int)]] = {
```
Filemon279, line 175: It is not super important but since we have already 2 languages maybe `language` should not be `String` anymore? Maybe something that extends `sealed trait` so we know in patter matching that we are missing handlers for some languages?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/parser/NuTemplateAwareExpressionParser.java
```diff
@@ -0,0 +1,287 @@
...
+				String exprStripLeading = expr.stripLeading();
+				String exprStripTrailing = exprStripLeading.stripTrailing();
+				int leadingWhitespaces = expr.length() - exprStripLeading.length();
+				int trailingWhitespaces = exprStripLeading.length() - exprStripTrailing.length();
+				if (expr.isEmpty()) {
+					throw new ParseException(expressionString, prefixIndex,
+							"No expression defined within delimiter '" + prefix + suffix +
+							"' at character " + prefixIndex);
+				}
+				expressions.add(new ExpressionWithPosition(doParseExpression(expr),prefixIndex + prefix.length() + leadingWhitespaces, suffixIndex + trailingWhitespaces));
```
arkadius, line 127: 1. Why do you keep in the `ExpressionWithPosition` class parsed expression with whitespace but start index after whitespace stripping? Would've been a difference if you had passed `expr.trim()` into `doParseExpression` method?
2. Why `suffixIndex + trailingWhitespaces`? Shouldn't it be minus?

bartektartanus: 1. Good point, this works now because in test the expression is longer than spaces before. Added another case with more spaces and it fails. Fixed that.
2. Another good point, fixed.

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/parser/NuTemplateAwareExpressionParser.java
bartektartanus: Ok 👍

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
bartektartanus: Done! :)

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
bartektartanus: I feel you, but the Expression class that defines this values is used in over 100 places (as Intellij tells me), so it might bloat this MR... maybe next time? :D

arkadius (approved):
LGTM

Filemon279 (approved):

PR #4469, opened 2023-07-07
Faster loading of components tab: Component's usages fetched once instead of fetching for each processing type

arkadius:


designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentService.scala
```diff
@@ -155,6 +148,7 @@ class DefaultComponentService private(componentLinksConfig: ComponentLinksConfig
       .getOrElse(componentLinks)
   }
 
+  // TODO: We should log some errors when we find some differences between definition of components
```
jedrz, line 151: I think it's not needed because validation is already performed in `ComponentsValidator`.

jedrz (approved):
:+1:

raphaelsolarski (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/component/ComponentService.scala
arkadius: I forgot about it - I replaced this comment with comment about where it can be found

PR #4478, opened 2023-07-11
Update testing documentation with ad hoc definition

Filemon279:


docs/scenarios_authoring/TestingAndDebugging.md
```diff
@@ -20,13 +20,33 @@ Below you can see how such file looks like.
 ```
 Each line of this file represents the next ongoing event and specify which source it targets with `sourceId` field. The json representation of the event is in `record` field.
 
+You could also use `generated` button to perform `generate file` and `from file` testing at once - without downloading locally and then uploading data to designer.
+
+## Testing using ad hoc form
+
+If you are unable to create data based on recent scenario executions, you can manually provide data using the 'ad hoc' functionality.
```
arkadius, line 27: Why only if you are UNABLE? What about the situation when it is just more convenient for someone?

Filemon279: Fair, I was telling a story here :smiley: I've added a convenient part as well :+1:

arkadius (approved):

PR #4479, opened 2023-07-11
Preview/cleaning and upgrade packages

Dzuming:
- checking if pipeline success


Dzuming:
It was only for the testing purpose

PR #4490, opened 2023-07-13
K8s test: more precise logging of test failure cause

arkadius:

### DIFF
--- a/engine/lite/k8sDeploymentManager/src/test/scala/pl/touk/nussknacker/k8s/manager/BaseK8sDeploymentManagerTest.scala
+++ b/engine/lite/k8sDeploymentManager/src/test/scala/pl/touk/nussknacker/k8s/manager/BaseK8sDeploymentManagerTest.scala
@@ -118,6 +118,9 @@ class BaseK8sDeploymentManagerTest extends AnyFunSuite with Matchers with Extrem
       logger.info("events:\n" + k8s.list[ListResource[Event]]().futureValue.items.mkString("\n"))
       pods.foreach { p =>
         logger.info(s"Printing logs for pod: ${p.name}")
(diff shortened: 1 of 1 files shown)


engine/lite/k8sDeploymentManager/src/test/scala/pl/touk/nussknacker/k8s/manager/BaseK8sDeploymentManagerTest.scala
```diff
@@ -115,7 +118,7 @@ class BaseK8sDeploymentManagerTest extends AnyFunSuite with Matchers with Extrem
       logger.info("events:\n" + k8s.list[ListResource[Event]]().futureValue.items.mkString("\n"))
       pods.foreach { p =>
         logger.info(s"Printing logs for pod: ${p.name}")
-        k8s.getPodLogSource(p.name, LogQueryParams()).futureValue.runForeach(bs => println(bs.utf8String)).futureValue
+        Try(k8s.getPodLogSource(p.name, LogQueryParams()).futureValue.runForeach(bs => println(bs.utf8String)).futureValue)
```
bartektartanus, line 121: Silently ignoring the exception?

arkadius: yes, I guess that it is a common situation when this logging can take a longer time - probable for some situations it never ends. I can add a comment about that

arkadius: I added comment - take a look if it is fine for you

PR #4491, opened 2023-07-13
Allow to override canTestWithForm for sources

Filemon279:


Filemon279:
Closing, we will do it different way

PR #4492, opened 2023-07-13
Add ad hoc testing support to fragments

Filemon279:
.

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/processreport/ProcessCounter.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/processreport/ProcessCounter.scala
@@ -2,14 +2,10 @@ package pl.touk.nussknacker.ui.processreport
 
 import cats.data.NonEmptyList
 import io.circe.generic.JsonCodec
-import pl.touk.nussknacker.engine.api.ProcessAdditionalFields
-import pl.touk.nussknacker.engine.api.process.VersionId
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
 import pl.touk.nussknacker.engine.canonicalgraph.canonicalnode._
 import pl.touk.nussknacker.engine.graph.node.{BranchEndData, FragmentInputDefinition}
-import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 import pl.touk.nussknacker.ui.process.fragment.FragmentRepository
-import shapeless.syntax.typeable._
 
 class ProcessCounter(fragmentRepository: FragmentRepository) {
 
@@ -35,8 +31,9 @@ class ProcessCounter(fragmentRepository: FragmentRepository) {
 
       nodes.flatMap {
         //TODO: this is a bit of a hack. Metric for fragment input is counted in node with fragment occurrence id...
-        case FlatNode(FragmentInputDefinition(id, _, _)) => Map(id -> nodeCountOption(None))
-        //BranchEndData is kind of artifical entity
+        //We want to count it though while testing fragments
+        case FlatNode(FragmentInputDefinition(id, _, _)) if !canonicalProcess.metaData.isFragment => Map(id -> nodeCountOption(None))
+        //BranchEndData is kind of artificial entity
         case FlatNode(BranchEndData(_)) => Map.empty[String, NodeCount]
         case FlatNode(node) => Map(node.id -> nodeCount(node.id))
         case FilterNode(node, nextFalse) => computeCountsSamePrefixes(nextFalse) + (node.id -> nodeCount(node.id))
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/StubbedFlinkProcessCompiler.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/StubbedFlinkProcessCompiler.scala
@@ -11,7 +11,7 @@ import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
 import pl.touk.nussknacker.engine.definition.DefinitionExtractor.{ComponentImplementationInvoker, ObjectWithMethodDef}
 import pl.touk.nussknacker.engine.definition.ProcessDefinitionExtractor.ModelDefinitionWithTypes
-import pl.touk.nussknacker.engine.graph.node.Source
+import pl.touk.nussknacker.engine.graph.node.{FragmentInputDefinition, Source}
 import shapeless.syntax.typeable._
 
 abstract class StubbedFlinkProcessCompiler(process: CanonicalProcess,
@@ -42,12 +42,20 @@ abstract class StubbedFlinkProcessCompiler(process: CanonicalProcess,
         sourceType -> stubbedDefinition
       }
 
+    def sourceDefForFragment(frag: FragmentInputDefinition): ObjectWithMethodDef = {
+      new StubbedFragmentInputDefinitionSource(processConfig, userCodeClassLoader).createSourceDefinition(frag)
+    }
+
+    val stubbedSourceForFragment: Seq[(String, ObjectWithMethodDef)] = process.allStartNodes.map(_.head.data).collect {
+      case frag: FragmentInputDefinition => frag.id -> prepareSourceFactory(sourceDefForFragment(frag), context)
+    }
+
     val stubbedServices = originalDefinition.services.mapValuesNow(prepareService(_, context))
 
     (ModelDefinitionWithTypes(
       originalDefinition
         .copy(
-          sourceFactories = originalDefinition.sourceFactories ++ stubbedSources,
+          sourceFactories = originalDefinition.sourceFactories ++ stubbedSources ++ stubbedSourceForFragment,
           services = stubbedServices)), originalDictRegistry)
   }
 
@@ -60,6 +68,7 @@ abstract class StubbedFlinkProcessCompiler(process: CanonicalProcess,
 case class ComponentDefinitionContext(userCodeClassLoader: ClassLoader,
                                       originalDefinitionWithTypes: ModelDefinitionWithTypes,
                                       originalDictRegistry: EngineDictRegistry)
+
 abstract class StubbedComponentImplementationInvoker(original: ComponentImplementationInvoker,
                                                      originalReturnType: Option[TypingResult]) extends ComponentImplementationInvoker {
   def this(componentDefinitionWithImpl: ObjectWithMethodDef) =
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
@@ -247,7 +247,7 @@ class FlinkProcessRegistrar(compileProcess: (CanonicalProcess, ProcessVersion, R
       })
       val asyncExecutionContextPreparer = processWithDeps.asyncExecutionContextPreparer
       val metaData = processWithDeps.metaData
-      val streamMetaData = MetaDataExtractor.extractTypeSpecificDataOrFail[StreamMetaData](metaData)
+      val streamMetaData = if(metaData.isFragment) StreamMetaData() else MetaDataExtractor.extractTypeSpecificDataOrFail[StreamMetaData](metaData)
 
       val configParameters = globalParameters.flatMap(_.configParameters)
       val useIOMonad = configParameters.flatMap(_.useIOMonadInInterpreter).getOrElse(true)
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
@@ -247,7 +247,7 @@ class FlinkProcessRegistrar(compileProcess: (CanonicalProcess, ProcessVersion, R
       })
       val asyncExecutionContextPreparer = processWithDeps.asyncExecutionContextPreparer
       val metaData = processWithDeps.metaData
-      val streamMetaData = if(metaData.isFragment) StreamMetaData() else MetaDataExtractor.extractTypeSpecificDataOrFail[StreamMetaData](metaData)
+      val streamMetaData = MetaDataExtractor.extractTypeSpecificDataOrDefault[StreamMetaData](metaData, StreamMetaData())
 
       val configParameters = globalParameters.flatMap(_.configParameters)
       val useIOMonad = configParameters.flatMap(_.useIOMonadInInterpreter).getOrElse(true)
--- a/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
+++ b/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
@@ -79,8 +79,8 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
       .emptySink("end", "sinkForInputMeta", SingleValueParamName -> "#input.city + '-' + #input.street")
 
     val parameterExpressions: Map[String, Expression] = Map(
-      "city" -> Expression("spel", "'Lublin'"),
-      "street" -> Expression("spel", "'Lipowa'"),
+      "city" -> Expression.spel("'Lublin'"),
+      "street" -> Expression.spel("'Lipowa'"),
     )
     val scenarioTestData = ScenarioTestData("start", parameterExpressions)
 
@@ -96,7 +96,7 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
       .fragmentOutput("fragmentEnd", "output", "out" -> "#in")
 
     val parameterExpressions: Map[String, Expression] = Map(
-      "in" -> Expression("spel", "'some-text-id'")
+      "in" -> Expression.spel("'some-text-id'")
     )
     val scenarioTestData = ScenarioTestData("fragment1", parameterExpressions)
     val results = run(fragment, scenarioTestData)
--- a/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
+++ b/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
@@ -89,6 +89,26 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
 
   }
 
+  test("should handle fragment test parameters in test") {
+    val fragment = ScenarioBuilder
+      .fragment("fragment1", "in" -> classOf[String])
+      .filter("filter", "#in != 'stop'")
+      .fragmentOutput("fragmentEnd", "output", "out" -> "#in")
+
+    val parameterExpressions: Map[String, Expression] = Map(
+      "in" -> Expression("spel", "'some-text-id'")
+    )
+    val scenarioTestData = ScenarioTestData("fragment1", parameterExpressions)
+    val results = run(fragment, scenarioTestData)
+
+    results.nodeResults("fragment1") shouldBe List(
+      NodeResult(ResultContext("fragment1-fragment1-0-0", Map("in" -> "some-text-id")))
+    )
+
+   // results.externalInvocationResults("fragmentEnd") shouldBe List(ExternalInvocationResult("fragment1", "fragmentEnd", "some-text-id"))
+    results.exceptions shouldBe empty
+  }
+
   private def registerSchema(topic: String) = {
     val subject = ConfluentUtils.topicSubject(topic, isKey = false)
     val parsedSchema = ConfluentUtils.convertToAvroSchema(Address.schema)
--- a/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
+++ b/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
@@ -105,7 +105,7 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
       NodeResult(ResultContext("fragment1-fragment1-0-0", Map("in" -> "some-text-id")))
     )
 
-   // results.externalInvocationResults("fragmentEnd") shouldBe List(ExternalInvocationResult("fragment1", "fragmentEnd", "some-text-id"))
+    results.invocationResults("fragmentEnd") shouldBe List(ExpressionInvocationResult("fragment1-fragment1-0-0", "out", "some-text-id"))
     results.exceptions shouldBe empty
   }
 
--- a/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
+++ b/engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
@@ -101,10 +101,8 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
     val scenarioTestData = ScenarioTestData("fragment1", parameterExpressions)
     val results = run(fragment, scenarioTestData)
 
-    results.nodeResults("fragment1") shouldBe List(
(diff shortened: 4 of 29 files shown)


engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
```diff
@@ -102,4 +102,24 @@ class InterpreterTestRunnerTest extends AnyFunSuite with Matchers {
...
+    val parameterExpressions: Map[String, Expression] = Map(
+      "in" -> Expression("spel", "'some-text-id'")
+    )
+    val scenarioTestData = ScenarioTestData("fragment1", parameterExpressions)
+    val results = sample.test(fragment, scenarioTestData)
+
+    results.nodeResults("fragment1") shouldBe List(
+      NodeResult(ResultContext("fragment1", Map("in" -> "some-text-id")))
+    )
+    //TODO - why FragmentOutputDefinition is not handled well in tests? :(
```
arkadius, line 120: I suppose that it may happen that way because, `PartSubGraphCompiler` handles `FragmentOutputDefinition` as a `Sink` (see line 120 and 129), and `ProcessSplitter` split `Sink` as a separate part. I'm not sure but I remember that there were some tricks around that. Have you tried to follow the debugger in our `InterpreterInternal` and check why debugger doesn't enter `ProcessListener.nodeEntered`? AFAIK this method is responsible for collecting testing results.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProvider.scala
```diff
@@ -100,8 +102,8 @@ class ModelDataTestInfoProvider(modelData: ModelData) extends TestInfoProvider w
     }.sequence.map(scenarioTestRecords => ScenarioTestData(scenarioTestRecords))
   }
 
-  private def collectAllSources(scenario: CanonicalProcess): List[Source] = {
-    scenario.collectAllNodes.flatMap(asSource)
+  private def collectAllSources(scenario: CanonicalProcess): List[SourceNodeData] = {
```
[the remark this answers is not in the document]
Filemon279: I meant it to be the opposite of complicating :sweat_smile: By adding `FragmentInputDefinition` to `allSources` we get "testing capabilities" and "test parameters" handled nicely without any effort or case/if statements.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -98,8 +98,32 @@ class NodeCompiler(definitions: ProcessDefinition[ObjectWithMethodDef],
           val defaultCtx = contextWithOnlyGlobalVariables.withVariable(VariableConstants.InputVariableName, Unknown, paramName = None)
           NodeCompilationResult(Map.empty, None, defaultCtx, error)
       }
-    case FragmentInputDefinition(_, params, _) =>
-      NodeCompilationResult(Map.empty, None, Valid(contextWithOnlyGlobalVariables.copy(localVariables = params.map(p => p.name -> loadFromParameter(p)).toMap)), Valid(new Source {}))
+    case frag@FragmentInputDefinition(id, params, _) =>
+      definitions.sourceFactories.get(id) match {
+        case Some(definition) =>
+          val parameters = fragmentDefinitionExtractor.extractParametersDefinition(frag).value
+          val variables: Map[String, TypingResult] = parameters.map(a => a.name -> a.typ).toMap
```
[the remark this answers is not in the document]
Filemon279: Well that is a place where a few words may be necessary:
- This whole fragment with `case Some(definition)` was created to handle fragment testing at flink engine. For RR it worked nicely with just a small change (extend it by `TestWithParametersSupport`) but for flink it was blocked by `def registerSourcePart` -> `val source = part.obj.asInstanceOf[FlinkSource]`. I've got to the point where I've just handled it as a all other sources in flink engine -> I've added simple source definition for `FragmentInputDefinition` to source definitions and used that to compile source as for all other flink sources with `compileObjectWithTransformation[Source]`.
- To use `compileObjectWithTransformation[Source]` I had to somehow go from `FragmentParameter` to `evaluatedparam.Parameter` and that is why I've used that maping. 

How it worked before? I dont think it was used in a way I am using it now. - `FragmentInputDefinition` was hidden in fragments only - in a scenario there is no `FragmentInputDefinition` - I think it is processed as `FragmentInput`.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/FragmentComponentDefinitionExtractor.scala
```diff
@@ -67,6 +67,12 @@ class FragmentComponentDefinitionExtractor(componentConfig: String => Option[Sin
       .mapWritten(_.map(data => FragmentParamClassLoadError(data.fieldName, data.refClazzName, nodeId.id)))
   }
 
+  def extractParametersDefinition(fragmentInputDefinition: FragmentInputDefinition): Writer[List[PartSubGraphCompilationError], List[Parameter]] = {
```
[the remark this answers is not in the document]
Filemon279: I've extracted common part to separate method :+1:

engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
Filemon279: I was able to make it work, thanks for a hint - it helped :)

interpreter/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/node.scala
```diff
@@ -43,6 +41,8 @@ object node {
 
   case class EndingCustomNode(id: String, ref: String) extends Node
 
+  case class EndingFragmentDefinition(id: String, fieldsWithExpression: Map[String, TypedExpression], isDisabled: Boolean) extends Node
```
arkadius, line 44: `EndingProcessor` and `EndingCustomNode` are special cases of `Processor` and `CustomNode` at the end of scenario :) fragment output definition can be in the middle of valid scenario?

designer/server/src/main/scala/pl/touk/nussknacker/ui/processreport/ProcessCounter.scala
```diff
@@ -35,8 +31,9 @@ class ProcessCounter(fragmentRepository: FragmentRepository) {
 
       nodes.flatMap {
         //TODO: this is a bit of a hack. Metric for fragment input is counted in node with fragment occurrence id...
-        case FlatNode(FragmentInputDefinition(id, _, _)) => Map(id -> nodeCountOption(None))
-        //BranchEndData is kind of artifical entity
+        //We want to count it though while testing fragments
```
arkadius, line 34: But when you use fragments inside sceanarios you don't have input definition but input usage. Do you?

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
```diff
@@ -266,6 +267,8 @@ object ScenarioInterpreterFactory {
         case er: EndReference =>
           //FIXME: do we need it at all
           monad.pure[ResultType[PartResult]](Writer.value(irs.map(ir => EndPartResult(er.nodeId, ir.finalContext, null.asInstanceOf[Res]))))
+        case fer: FragmentEndReference =>
+          monad.pure[ResultType[PartResult]](Writer.value(irs.map(ir => EndPartResult(fer.nodeId, ir.finalContext, fer.declaredVariables.asInstanceOf[Res]))))
```
arkadius, line 271: I'm just curious, why do you need return this result? For req-resp it is necessary to return it via HTTP response. For tests I think that you need only to notify listener. Don't you?

engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
```diff
@@ -102,4 +102,23 @@ class InterpreterTestRunnerTest extends AnyFunSuite with Matchers {
...
+    val parameterExpressions: Map[String, Expression] = Map(
+      "in" -> Expression("spel", "'some-text-id'")
+    )
+    val scenarioTestData = ScenarioTestData("fragment1", parameterExpressions)
+    val results = sample.test(fragment, scenarioTestData)
+
+    results.nodeResults("fragment1") shouldBe List(
+      NodeResult(ResultContext("fragment1", Map("in" -> "some-text-id")))
+    )
+    results.externalInvocationResults("fragmentEnd") shouldBe List(ExternalInvocationResult("fragment1", "fragmentEnd", Map("out" -> "some-text-id")))
```
arkadius, line 120: Why is it external invocation? I thought that external invocations are for services invocations tests. Why not nodeResults?

engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
```diff
@@ -89,6 +89,26 @@ class TestWithTestDataSpec extends AnyFunSuite with Matchers with LazyLogging {
 
   }
 
+  test("should handle fragment test parameters in test") {
```
arkadius, line 92: We have to add tests also using our `TestScenarioRunner` mechanism - for both Flink and Lite

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
arkadius: I see that it can be helpful for our scenario testing mechanism (TestScenarioRunner)

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/StubbedFlinkProcessCompiler.scala
```diff
@@ -42,12 +42,20 @@ abstract class StubbedFlinkProcessCompiler(process: CanonicalProcess,
         sourceType -> stubbedDefinition
       }
 
+    def sourceDefForFragment(frag: FragmentInputDefinition): ObjectWithMethodDef = {
+      new StubbedFragmentInputDefinitionSource(processConfig, userCodeClassLoader).createSourceDefinition(frag)
+    }
+
+    val stubbedSourceForFragment: Seq[(String, ObjectWithMethodDef)] = process.allStartNodes.map(_.head.data).collect {
```
arkadius, line 49: Is it possible to have this pattern matching inside `TestFlinkProcessCompiler.prepareSourceFactory` ?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProvider.scala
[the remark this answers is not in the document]
arkadius: I see. I forgot that it is not only for testing capabilities purpose but also for test parameter definition :+1:

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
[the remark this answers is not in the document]
arkadius: `definitions.sourceFactories`  AFAIK contains only sources defined in model (not by user). So why this `Some(definition)` is needed?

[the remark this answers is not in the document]
Filemon279: It is needed for fragment testing in Flink - while in Lite mode we can create `Source with TestWithParametersSupport` that in Flink we need to create `FlinkSource` and we cannot do it here. I've done it in `StubbedFlinkProcessCompiler` by adding prepared for Flink source to `definitions` and I'm extracting it here using `Some(definition)`.

interpreter/src/test/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProviderSpec.scala
```diff
@@ -95,6 +95,11 @@ class ModelDataTestInfoProviderSpec extends AnyFunSuite with Matchers with Optio
     capabilities shouldBe TestingCapabilities(canBeTested = true, canGenerateTestData = false, canTestWithForm = true)
   }
 
+  test("should detect capabilities for fragment with valid input") {
```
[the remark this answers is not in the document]
Filemon279: Nope, I will write some soon :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/processreport/ProcessCounter.scala
Filemon279: Hmm, I didn't challenge it but I think it should work like you said. In test `"compute counts for scenario with fragment"` it looks like `FragmentInputDefinition` and `FragmentOutputDefinition` are passed explicitly - I am not sure if that happens in NU testing cases (?). 

What I've done here is just a simple if statement to handle `FlatNode(FragmentInputDefinition)` a little later in `case FlatNode(node)` few lines bellow it - in scenario testing this node is skipped in favor of node that represents fragment which is not present in fragment testing.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/api/PartReference.scala
```diff
@@ -7,7 +7,11 @@ sealed trait EndingReference extends PartReference {
...
 
 case class NextPartReference(id: String) extends PartReference
+
 case class JoinReference(id: String, branchId: String, joinId: String) extends PartReference
+
 case class DeadEndReference(nodeId: String) extends EndingReference
+
 case class EndReference(nodeId: String) extends EndingReference
 
+case class FragmentEndReference(nodeId: String, declaredVariables: Map[String, Any]) extends EndingReference
```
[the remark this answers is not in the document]
Filemon279: Changed to `outputFields` :+1:

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -247,7 +247,7 @@ class FlinkProcessRegistrar(compileProcess: (CanonicalProcess, ProcessVersion, R
       })
       val asyncExecutionContextPreparer = processWithDeps.asyncExecutionContextPreparer
       val metaData = processWithDeps.metaData
-      val streamMetaData = MetaDataExtractor.extractTypeSpecificDataOrFail[StreamMetaData](metaData)
+      val streamMetaData = if(metaData.isFragment) StreamMetaData() else MetaDataExtractor.extractTypeSpecificDataOrFail[StreamMetaData](metaData)
```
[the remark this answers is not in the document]
Filemon279: That is a good idea - changed

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProvider.scala
```diff
@@ -73,13 +75,12 @@ class ModelDataTestInfoProvider(modelData: ModelData) extends TestInfoProvider w
   private def prepareTestDataGenerators(scenario: CanonicalProcess): List[(NodeId, TestDataGenerator)] = {
     for {
       source <- collectAllSources(scenario)
-      sourceObj <- prepareSourceObj(source)(scenario.metaData)
+      sourceObj <- prepareSourceObj(source)(scenario.metaData, NodeId(scenario.id))
```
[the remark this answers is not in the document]
Filemon279: That is a nasty typo - thanks for that :+1:

engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
Filemon279: Doesn't `TestScenarioRunner` test scenarios invocation? We are never running fragment "as scenario" - we can only run fragment by test mechanism. There is no scenario in which we can call fragment as we would do with scenario. I might be wrong but isn't it the case? :thinking:

interpreter/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/node.scala
Filemon279: Not to confuse those names I've changed `EndingFragmentDefinition` to `FragmentOutput`

interpreter/src/test/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProviderSpec.scala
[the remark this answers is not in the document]
Filemon279: I've added new tests

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
arkadius: WDYT by "I couldn't see what we are getting at the end of fragment."? To see where? In testing mechanism. Test results for this ending node are reported before you create `FragmentEndReference`, here: https://github.com/TouK/nussknacker/pull/4492/files/915fa585694829a0c7909f54d9638514244a5ff9#diff-7dcb7c349f9082c18d9b7b2cfd5524059175576524e1e715d4cb6ee52994193bR111

engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
arkadius: Do you have an idea why it was done this way? I suspect that it was because someone wanted to see output of sink in unit tests - I mean you have sink with parameter `a: XYZ` which creates complex result where this a is somewhere used (e.g. nested). Am I right that we still also have those output fields as nodeResults (variables)? These externalInvocationResults are also visible in GUI?

engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
arkadius: Thanks to `TestScenarioRunner` our users can test various scenario constructions in unit test. I think, that they also should be able to tests fragments as a decomposed problem that will be composed in scenario. I know that it can look a little bit "out of scope" but  IMO thanks to that we will see if this implementation is good or not. So far everything that is possible in test in GUI mechanism is also possible in our "testkit" framework

designer/server/src/main/scala/pl/touk/nussknacker/ui/processreport/ProcessCounter.scala
arkadius: I see, probably this `FragmentInputDefinition` is somewhere in model even after fragment resolution

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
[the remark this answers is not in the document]
arkadius: Ok, I know understand the idea. Thank  you for clarification.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/StubbedFlinkProcessCompiler.scala
arkadius: It is impossible, forgot about this

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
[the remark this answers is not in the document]
arkadius: Please replace `inputParameters` with `Nil` here: https://github.com/TouK/nussknacker/pull/4492/files/915fa585694829a0c7909f54d9638514244a5ff9#diff-5df69bb691ef491cb7ba2617e59f09b66fb99c705cbec26251d051f9fc31c9a7R30 . Thanks to that you won't need to create this list of `evaluatedparam.Parameter` here

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -98,8 +98,22 @@ class NodeCompiler(definitions: ProcessDefinition[ObjectWithMethodDef],
...
+          val defaultContextTransformation =
```
arkadius, line 106: Why default?

Filemon279: Just followed similar naming (from a little bit above) but as you noticed it can be more meaningful - changed to simple `validationContext` :+1:

arkadius: Yep, but in the case above, it was a function that was used as a fallback :P

engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
Filemon279: Ok, I tried to figure it out and it looks like in lite we have result in `nodeResults` `invocationResults` and `externalInvocationResults`. `externalInvocationResults` in this case is a result of invoking `SinkInvocationCollector` on sinks in `lite` mode. It is not used in sinks in flink (or at least not in this case) since in flink mode we have result only in `nodeResults` `invocationResults`.

I see it that way: 
- `nodeResults` - is a good place for it to be visible since we invoke noteEntered on node enter. 
- `invocationResults` - is also valid since its an effect on using `ExpressionEvaluator`
- `externalInvocationResults` - I am not sure but it can be omitted for sources in test. I've tried to remove it on two tests and I was able to replace it with check on `invocationResults`. I am nut sure yet if in all cases it will act the same but if you want me to I can try to remove this listener from sinks in test.

I've also updated tests to check both `nodeResults` `invocationResults`.

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
Filemon279: It is strongly correlated with conversation we are having bellow. Since I've used under the hood mechanism to report sink output to `externalInvocationResults` two things happend:
- I've looked up `externalInvocationResults` in tests and what I've got in `EndPartResult` I have also got in `externalInvocationResults` result. That is why I've passed this result here. 
- `externalInvocationResults` are also sent to FE, so passing null here crashes NU FE. (`TypeError: Cannot read properties of null (reading 'pretty')`) 

What I think I should do is not to pass result to `EndPartResult` and in the same time do not report sink invocation to `externalInvocationResults`. To do that I would have to remove `collectSinkResults` from `TestRunner.scala`. WDYT?

engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/TestWithTestDataSpec.scala
Filemon279: The idea is great but it sounds like some refactor is needed to handle fragments in such a form. I wonder if that isn't something we should do separately from this MR?

arkadius: It might be in separate MR

arkadius (approved):

engine/lite/runtime/src/test/scala/pl/touk/nussknacker/engine/lite/InterpreterTestRunnerTest.scala
Filemon279: Ok, I know now that `externalInvocationResults` is in lite tests for a reason - to collect output from sink after some transformation `LazyParamSink -> prepareResponse`. In tests it can be found here `pl.touk.nussknacker.engine.requestresponse.ParameterResponseSinkFactory.ParameterResponseSink`. In my case that does not matter but matters in different places. I've added it to my TODO list to try to figure it out and do some refactor. I will come back to it in another PR if that is ok with you.

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
Filemon279: I will close it, since it is connected with conversation we are having bellow.

PR #4498, opened 2023-07-13
Fix for: PeriodicDeploymentManager was presenting finished schedule a…

arkadius:
…s a problem - works when caching enabled too


lciolecki (approved):

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/cache/CachingProcessStateDeploymentManager.scala
```diff
@@ -85,7 +85,15 @@ object CachingProcessStateDeploymentManager extends LazyLogging {
       val cacheTTL = cachingConfig.cacheTTL
         .getOrElse(throw new IllegalArgumentException(s"Invalid config: $this. If you want to enable processStateCaching, you have to define cacheTTL"))
       logger.debug(s"Wrapping DeploymentManager: $delegate with caching mechanism with TTL: $cacheTTL")
-      new CachingProcessStateDeploymentManager(delegate, cacheTTL)
+      delegate match {
```
lciolecki, line 88: Can we add any test for it?

PR #4506, opened 2023-07-18
Node errors i18n with templates

mslabek:


designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -43,15 +43,14 @@ export default function ValidationLabels(props: Props) {
             const { errorCode, templateValues, message } = validationError;
 
             const canRenderI18nTemplate = errorCode != null && templateValues != null;
```
Dzuming, line 45: We can simplify it to `const canRenderI18nTemplate = Boolean(errorCode && templateValues)`

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -22,22 +25,33 @@ const LabelsContainer = styled.div({
...
         validationErrors.map((validationError) => {
+            const { errorCode, templateValues, message } = validationError;
+
+            const canRenderI18nTemplate = errorCode != null && templateValues != null;
+
             // we don't pass description as tooltip message until we make changes on the backend
             return (
-                <LimitedValidationLabel key={validationError.message} title={validationError.message} type="ERROR">
-                    {validationError.message}
+                <LimitedValidationLabel key={message} type="ERROR" title={message}>
```
Dzuming, line 49: 1. The function of this component is to display a tooltip, right? Can we rename it to the <ValidationTooltip />?
2. Do we have a generic way to display tooltips in the application? If not, maybe we should make this component generic and even consider using https://mui.com/material-ui/react-tooltip/ for tooltips. (but for now, making it a shareable component, will be enough IMO)

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -22,22 +25,33 @@ const LabelsContainer = styled.div({
...
+                    {canRenderI18nTemplate ? (
+                        <I18nTemplateLabel errorCode={errorCode} templateValues={validationError.templateValues} />
```
Dzuming, line 51: How many template variants we will have? I checked `designer/client/src/types/definition.ts` but I'm not able to determine it, based on these types. Maybe it will be easier, if we try to write down, what variants of validation errors we have in our system

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -54,3 +68,64 @@ export default function ValidationLabels(props: Props) {
...
+
+const TemplatedType = styled.span({
+    fontWeight: "bold",
+    textDecoration: "underline",
+    "&:hover": {
+        color: "white",
+    },
+});
+
+function I18nTemplateLabel(props: TemplateLabelProps) {
```
Dzuming, line 85: Maybe we can hide the implementation details and make the component name more generic?

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -54,3 +68,64 @@ export default function ValidationLabels(props: Props) {
...
+    return canRender ? (
+        <Trans
+            i18nKey={`nodeErrorCodes:${errorCode}.messageTemplate`}
+            values={displayableTemplateValues}
+            components={templateComponents}
+        ></Trans>
+    ) : null;
+}
+
+function validateTemplateInterpolation(template: string, valueNames: string[], componentNames: string[]): boolean {
```
Dzuming, line 127: Do we need explicitly say that the return value is `boolean`? Ts is smart enough to calculate it bases on return value :)

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -54,3 +68,64 @@ export default function ValidationLabels(props: Props) {
...
+
+    const canRender =
+        i18n.exists(`nodeErrorCodes:${errorCode}.messageTemplate`) &&
+        validateTemplateInterpolation(
+            i18n.t(`nodeErrorCodes:${errorCode}.messageTemplate`),
+            Object.keys(displayableTemplateValues),
+            Object.keys(templateComponents),
+        );
+
+    return canRender ? (
```
Dzuming, line 118: IMO, This check should only warn us somehow (for example console warn). We should also have e2e tests to check if we are able to render specific template variants (not sure how many variants we will have). Thanks to it we will be able to verify if someone not broke the contract.

designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -13,6 +14,8 @@ type Props = {
 type ValidationError = {
```
Dzuming, line 14: To me, the type should be like

type Params = OperationMismatchParams | SomethingDifferentParams;    
type ErrorCode =  'OPERATOR_MISMATCH' | SomethingDifferent

type ValidationError = {
description: string;     
params: Params
errorCode: ErrorCode;
}

designer/client/src/components/modals/ValidationLabels.tsx
mslabek: Right now it's a mess :/ - I'm still trying to figure out how many are needed and right now I think we should have 2: a simple string value and typingResult value. The string could be interpolated without having a dedicated component, but the typingResult requires a custom span component. 

I am working on having a union type with a template and non-templated error - I'm hoping this will be a valid solution.

designer/client/src/components/modals/ValidationLabels.tsx
mslabek: Agree, with proper testing and contract this will not be needed.

PR #4514, opened 2023-07-21
Replace AkkaHttpBackend with AsyncHttpClientFutureBackend

gskrobisz:
In https://github.com/TouK/nussknacker/commit/5205163169f5e11c7170d5df91ea8587cbe6e063  we added `readTimeout` to `HttpFlinkClient.findJobsByName`.
Akka does not support such setting and sttp uses idleTimeout as readTimeout, see sttp.client3.akkahttp.AkkaHttpBackend :
```
  private def connectionSettings(r: Request[_, _]): ConnectionPoolSettings = {
    //...
    connectionPoolSettingsWithProxy
    .withUpdatedConnectionSettings(_.withIdleTimeout(r.options.readTimeout))
  } 
```
And that idle/readTimeout means that after timeout connection is killed via RST (see also https://github.com/akka/akka-http/issues/3346).
Therefore, to avoid broken sessions of findJobsByName, when HttpFlinkClient is created we should replace use AsyncHttpClientFutureBackend instead of AkkaHttpBackend.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1387,6 +1387,9 @@ lazy val httpUtils = (project in utils("http-utils")).
         "com.softwaremill.sttp.client3" %% "core" % sttpV,
         "com.softwaremill.sttp.client3" %% "json-common" % sttpV,
         "com.softwaremill.sttp.client3" %% "circe" % sttpV,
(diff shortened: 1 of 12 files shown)


arkadius:
Please remember to change `scenarioStateIdleTimeout` parameter name and update docs related with this

arkadius (approved):

PR #4497, opened 2023-07-13
component usages filters

JulianWielga:
upgraded usages display
added filter for usage type
fixed text filter
cleanup

PR #4521, opened 2023-07-25
Fix for flaky cypress test: Components list should filter usages

arkadius:
Waiting for table render befor submit filtering


Dzuming:
@arkadius I think it's not a problem with waiting for a table render. Based on a Cypress video, the filter is not correct, because it expects that, it will be 8 in the component name `8 xxx` but, I assume that the name of the component contains a timestamp, so, it can be a situation that, `8` won't be found. 

https://github.com/TouK/nussknacker/assets/9945753/43661416-910c-47b9-84bd-ca56ba43c421

arkadius:
It is resolved correctly in #4497

PR #4554, opened 2023-08-04
Improvement/simplify result of invokeCustomActions

deikare:


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
```diff
@@ -243,18 +244,20 @@ class ManagementResources(val processAuthorizer: AuthorizeProcess,
               val params = req.params.getOrElse(Map.empty)
               complete {
                 customActionInvokerService.invokeCustomAction(req.actionName, process, params)
-                  .flatMap {
-                    case res@Right(_) =>
+                  .transformWith {
```
vder, line 247: I dont like this part. Why dont' you use  flatMap + recoverWith ?  It  allows one to separate business logic from error handing and I believe you dont have to care about this generic case `case Failure(err) =>` at all.

lciolecki: You can do it even better / easier. Specific errors should implement specific interfaces and error code mapping should be done by EspErrorToHttp.espErrorToHttp -> I'm not sure, but I think instead of returning Future.failed, you can just throw an exception as well.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/CustomActionInvokerService.scala
```diff
@@ -53,17 +53,13 @@ class CustomActionInvokerServiceImpl(processRepository: FetchingProcessRepositor
               if (customAction.allowedStateStatusNames.contains(status.status.name)) {
                 manager.invokeCustomAction(actionReq, process.json)
               } else
-                actionError(CustomActionInvalidStatus(actionReq, status.status.name))
+                Future.failed(CustomActionInvalidStatus(actionReq, status.status.name))
```
lciolecki, line 56: If throwing exception doesn't work, extract some error message for Future.failed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
arkadius: @deikare the point of this change was to remove this piece of code and use `EspErrorToHttp` instead :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/Error.scala
```diff
@@ -0,0 +1,31 @@
...
+import akka.http.scaladsl.model.{StatusCode, StatusCodes}
+
+
+object Error {
+
+  type XError[A] = Either[Error, A]
+
+}
+
+trait Error {
```
Filemon279, line 12: If that is the only place where we extend `trait Error` then we should make it sealed

designer/server/src/main/scala/pl/touk/nussknacker/ui/Error.scala
```diff
@@ -0,0 +1,31 @@
...
+  def getMessage: String
+  val statusCode: Option[StatusCode] = None
```
Filemon279, line 14: Is `Option` really required here? All errors we declared have `statusCode` assigned. Or maybe we want to allow creating new errors in new places? If so forget about my 2 comments :)

deikare: We want to allow creating new errors in new places

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/CustomAction.scala
```diff
@@ -34,26 +34,31 @@ case class CustomActionRequest(name: String,
 
 case class CustomActionResult(req: CustomActionRequest, msg: String)
 
-sealed trait CustomActionError extends Exception {
+abstract class CustomActionError extends Exception with Error {
```
arkadius, line 37: Why we need both `CustomActionError` and `Error`. What benefits do you see in this approach?

engine/development/deploymentManager/src/main/scala/pl/touk/nussknacker/development/manager/DevelopmentDeploymentManagerProvider.scala
```diff
@@ -121,28 +121,27 @@ class DevelopmentDeploymentManager(actorSystem: ActorSystem)
...
+        val statusDetails = memory.getOrElse(processName, createAndSaveProcessState(NotDeployed, actionRequest.processVersion))
+
+        if (customAction.allowedStateStatusNames.contains(statusDetails.status.name)) {
+          customActionStatusMapping
+            .get(customAction)
+            .map { status =>
+              asyncChangeState(processName, status)
+              Future.successful(CustomActionResult(actionRequest, s"Done ${actionRequest.name}"))
+            }
+            .getOrElse(Future.failed(CustomActionInvalidStatus(actionRequest, statusDetails.status.name)))
```
arkadius, line 138: IMO this example need refactor. I don't understand why someone would want to check `customAction.allowedStateStatusNames.contains(statusDetails.status.name)` by himself inside DeploymnetManager. It is already checked inside `CustomActionInvokerService`. I think that we should remove all these `CustomActionError` implementations from `extensions-api` module and make them internal (`designer` module)

arkadius:
Closing it in favour of #4687

PR #4556, opened 2023-08-07
Handle isFragment / isSubprocess in Decoder / Encoder

Filemon279:

### DIFF
--- a/designer/client/cypress/e2e/components.cy.ts
+++ b/designer/client/cypress/e2e/components.cy.ts
@@ -196,7 +196,7 @@ describe("Components list", () => {
             .click();
         cy.wait(500); //ensure "loading" mask is hidden
         cy.get("#app-container>main").matchImage({
-            maxDiffThreshold: 0.01,
+            maxDiffThreshold: 0.02,
         });
 
         cy.get("@options")
@@ -219,7 +219,7 @@ describe("Components list", () => {
         cy.viewport(1600, 500);
         cy.wait(500); //ensure "loading" mask is hidden
         cy.get("#app-container>main").matchImage({
-            maxDiffThreshold: 0.01,
+            maxDiffThreshold: 0.02,
         });
     });
 
--- a/designer/client/cypress/e2e/components.cy.ts
+++ b/designer/client/cypress/e2e/components.cy.ts
@@ -207,7 +207,7 @@ describe("Components list", () => {
             .click();
         cy.wait(500); //ensure "loading" mask is hidden
         cy.get("#app-container>main").matchImage({
-            maxDiffThreshold: 0.01,
+            maxDiffThreshold: 0.02,
         });
 
         cy.get("@options")
--- a/designer/restmodel/src/test/scala/pl/touk/nussknacker/restmodel/ProcessDetailsCodecSpec.scala
+++ b/designer/restmodel/src/test/scala/pl/touk/nussknacker/restmodel/ProcessDetailsCodecSpec.scala
@@ -1,5 +1,6 @@
 package pl.touk.nussknacker.restmodel
 
+import io.circe.Json
 import io.circe.parser.parse
 import io.circe.syntax._
 import org.scalatest.funsuite.AnyFunSuite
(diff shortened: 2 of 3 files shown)


lciolecki:
Wher are tests for it? :)

Filemon279:
@lciolecki I've added tests for encoding to JSON with `isSubprocess` field added and decoding JSON that contains that field :+1:

designer/restmodel/src/test/scala/pl/touk/nussknacker/restmodel/ProcessDetailsCodecSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+      |  "history" : [],
+      |  "isSubprocess" : $isFragment
+      |}""".stripMargin
+
+  test("decode BaseProcessDetails with isSubprocess") {
+    val validJson = parse(baseProcessDetailsWithSubprocess).toOption.get
+    val baseProcessDetails = validJson.as[BaseProcessDetails[DisplayableProcess]].toOption
+    baseProcessDetails shouldBe Symbol("defined")
+    baseProcessDetails.get.isFragment shouldBe isFragment
+  }
```
DeamonDev, line 64: Maybe it is good idea to add second check: 

` baseProcessDetails.get.isSubprocess shouldBe isSubprocess (=true)`

And also the second test for process without subprocess?

Filemon279: But `BaseProcessDetails` does not have `isSubprocess` field - it can accept that field in json format and translate it to `isFragment` field - we cannot call `.isSubprocess` explicitly.

I've added extra test for a case when we are using `isFragment` field in JSON - not `isSubprocess` if that what you had in mind?

DeamonDev: Ahh okay, then ignore my _add-line_ suggestion. 

>I've added extra test for a case when we are using isFragment field in JSON - not isSubprocess if that what you had in mind?

Exactly! :)

designer/client/cypress/e2e/components.cy.ts
```diff
@@ -196,7 +196,7 @@ describe("Components list", () => {
             .click();
         cy.wait(500); //ensure "loading" mask is hidden
         cy.get("#app-container>main").matchImage({
-            maxDiffThreshold: 0.01,
+            maxDiffThreshold: 0.02,
```
lciolecki, line 199: I'm not sure about it..

designer/restmodel/src/test/scala/pl/touk/nussknacker/restmodel/ProcessDetailsCodecSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+      |  "isArchived" : false,
+      |  "processingType" : "streaming",
+      |  "processCategory" : "Category1",
+      |  "modificationDate" : "2023-08-07T10:57:33.986223Z",
+      |  "modifiedAt" : "2023-08-07T10:57:33.986227Z",
+      |  "modifiedBy" : "user1",
+      |  "createdAt" : "2023-08-07T10:57:33.986230Z",
+      |  "createdBy" : "user1",
+      |  "tags" : [],
+      |  "json" : {
```
lciolecki, line 36: You can do tests without this field, but let's leave it.

lciolecki (approved):

designer/client/cypress/e2e/components.cy.ts
Filemon279: Yeah, I'm not a fan of that either - without it tests cant complete (because scenarios uses timestamp in their names). I've started conversation about it elsewhere - hope to get it resolved soon.

Filemon279: If there is a possibility like that and name does not matter in any way in that test then it can be an option. I was rather thinking about creating a name in more predictable way so it wont change on every test.

PR #4561, opened 2023-08-08
Improvement: Users are not requrired at OAuth2 config files

lciolecki:


security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationConfiguration.scala
```diff
@@ -28,7 +28,7 @@ object AuthenticationConfiguration {
   val rulesConfigurationPath = "rules"
 
   def getUsers(config: Config): List[ConfigUser] = config.as[List[ConfigUser]](usersConfigurationPath)
-
+  def getUsersOrEmptyList(config: Config): List[ConfigUser] = config.as[Option[List[ConfigUser]]](usersConfigurationPath).getOrElse(Nil)
```
arkadius, line 31: Why not to have one methods instead?

lciolecki: Because for BasicAuth we require the users field on file - so we want to throw an exception when the field is missed.

lciolecki: Hmm, after thinking we can do one method -> check it right now.

arkadius (approved):

PR #4562, opened 2023-08-08
Some links for better interconnection of concepts

ZbyszekMM:
I propose to add several links to interconnect concepts which are described in sometimes pretty 'remote' branches of documentation and as a result it is not easy or obvious how to find the given contents.


arkadius (approved):
Good job

PR #4563, opened 2023-08-08
Custom decoder for ProcessAction

DeamonDev:


extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessAction.scala
```diff
@@ -25,6 +28,47 @@ import java.util.UUID
                                     comment: Option[String],
                                     buildInfo: Map[String, String])
 
+object ProcessAction {
+  //custom decoder for compatibility reasons
+  implicit val decodeProcessAction: Decoder[ProcessAction] = new Decoder[ProcessAction] {
+    override def apply(c: HCursor): Result[ProcessAction] =
+      for {
+        id               <- c.downField("id").as[ProcessActionId] match {
+                              case Left(_) => Right(ProcessActionId(UUID.randomUUID()))
```
raphaelsolarski, line 37: if it is only for migrations/compare to a remote env then please:
* add some comment that these values are ignored in this specific features and are only dummy values
* maybe use some constants ("0000-0000-0000-0000, 0, ...) instead of Random.nextLong(), UUID.randomUUID() and Instant.now().

gskrobisz (approved):
lgtm

Filemon279 (approved):

PR #4571, opened 2023-08-11
Allow nonparametrized queries in sql enricher

mslabek:


arkadius (approved):

components/sql/src/test/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricherValidationTest.scala
```diff
@@ -42,6 +42,25 @@ class DatabaseQueryEnricherValidationTest extends BaseHsqlQueryEnricherTest {
     )
   }
 
+  test("should handle non-parametrized queries") {
```
arkadius, line 45: IMO a little bit better would be to write test on higher level (using scenario api), but I see that there already are tests written that way so it is also fine for me

mslabek: Will do in the future

PR #4578, opened 2023-08-16
Nu 987 undo redo unwanted state changes fixed

JulianWielga:
- moved process validation from action creators to middleware
- fixed node selection highlight


designer/client/src/reducers/graph/reducer.ts
```diff
@@ -297,17 +298,29 @@ const graphReducer: Reducer<GraphState> = (state = emptyGraphState, action) => {
...
     },
 });
 
-const undoableReducer = undoable(reducer, {
+const pick = <T extends NonNullable<unknown>>(object: T, props: NestedKeyOf<T>[]) => _pick(object, props);
+const omit = <T extends NonNullable<unknown>>(object: T, props: NestedKeyOf<T>[]) => _omit(object, props);
+
+const pickKeys: NestedKeyOf<GraphState>[] = ["fetchedProcessDetails", "processToDisplay", "unsavedNewName", "layout", "selectionState"];
+const omitKeys: NestedKeyOf<GraphState>[] = [
+    "processToDisplay.validationResult",
```
Dzuming, line 312: I understand this concept and the problem we want to resolve here, but now, the validation error won't be clean after undo

To reproduce:
Remove the connection between nodes
Click undo button
The scenario is valid, but I can still see an error

Dzuming (changes requested):

designer/client/src/reducers/graph/reducer.ts
Dzuming: I was able to reproduce the situation, that after redo, validation is missing

https://github.com/TouK/nussknacker/assets/9945753/2df705c2-53eb-4a8d-9af1-b3bd51139034

Dzuming (changes requested):
I added one comment about redo https://github.com/TouK/nussknacker/pull/4578#discussion_r1299704988

designer/client/src/reducers/graph/reducer.ts
JulianWielga: should work now

Dzuming: Another visual thing, before process validation, nodes that had validation errors, were marked as invalid. It's not a big blocker, so if it's easy to fix, I would say, let's do it.


https://github.com/TouK/nussknacker/assets/9945753/02ad7cee-c93a-4a42-9e89-46bfbceae2aa

JulianWielga: better now?

Dzuming: LGTM, thank you!

Dzuming (approved):

PR #4582, opened 2023-08-18
[NU-891] app start refactoring

coutoPL:
In this PR:
* Fixed: Releasing app resources on the designer close
* refactoring of the Nu boot code
* minor readme fixes


designer/server/src/test/scala/pl/touk/nussknacker/ui/notifications/NotificationServiceTest.scala
```diff
@@ -41,13 +41,13 @@ import scala.util.{Failure, Success, Try}
 class NotificationServiceTest extends AnyFunSuite with Matchers with PatientScalaFutures with MockitoSugar with WithHsqlDbTesting with EitherValuesDetailedMessage with OptionValues with DBIOActionValues {
 
   private implicit val system: ActorSystem = ActorSystem(getClass.getSimpleName)
-  override protected val dbioRunner: DBIOActionRunner = DBIOActionRunner(db)
+  override protected val dbioRunner: DBIOActionRunner = DBIOActionRunner(dbConfig)
```
arkadius, line 44: If you already touched `DbConfig` class, WDYT about changing its name? IMO it is misleading now. It suggests that it contains a bean representation of some static, text configuration but in fact, it is a database client with a profile allowing access to specific query constructions.  Maybe `DbSetup` or sth similar?

build.sbt
```diff
@@ -948,6 +948,9 @@ lazy val testUtils = (project in utils("test-utils")).
         "ch.qos.logback" % "logback-classic" % logbackV,
         "commons-io" % "commons-io" % flinkCommonsIOV,
         "org.scala-lang.modules" %% "scala-collection-compat" % scalaCollectionsCompatV,
+        "com.softwaremill.sttp.client3" %% "slf4j-backend" % sttpV,
+        "org.typelevel" %% "cats-effect" % "2.5.5",
```
arkadius, line 952: Let's extract the version to a separate variable. BTW @lciolecki opened this PR: https://github.com/TouK/nussknacker/pull/4287

designer/server/src/main/scala/pl/touk/nussknacker/ui/NussknackerBoot.scala
```diff
@@ -0,0 +1,14 @@
+package pl.touk.nussknacker.ui
+
+import cats.effect.{ExitCode, IO, IOApp}
+import pl.touk.nussknacker.ui.factory.NussknackerApp
+
+object NussknackerBoot extends IOApp {
```
arkadius, line 6: We have fixed IDEA configurations in `.run/*.xml`, in scripts with entrypoints and also could be used in some dependant projects. IMO would be better to keep the old name and instead change extracted by you class to e.g. `NussknackerAppFactory`. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerApp.scala
```diff
@@ -0,0 +1,113 @@
...
+        .make(
+          acquire = IO(JdbcBackend.Database.forConfig("db", resolvedConfig))
+        )(
+          release = db => IO(db.close())
+        )
+      _ <- Resource.eval(IO(DatabaseInitializer.initDatabase("db", resolvedConfig)))
+    } yield DbConfig(db, chooseDbProfile(resolvedConfig))
+  }
+
+  private def chooseDbProfile(config: Config): JdbcProfile = {
```
arkadius, line 104: IMO we should move this to `DbConfig` factory method and keep this class as concise as possible

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/NuResourcesTest.scala
```diff
@@ -34,52 +38,50 @@ import pl.touk.nussknacker.ui.config.FeatureTogglesConfig
...
-import _root_.sttp.client3.SttpBackend
-import _root_.sttp.client3.akkahttp.AkkaHttpBackend
-import pl.touk.nussknacker.engine.api.deployment.simple.SimpleStateStatus
-import pl.touk.nussknacker.engine.definition.test.{ModelDataTestInfoProvider, TestInfoProvider}
 
 import java.net.URI
 import scala.concurrent.{ExecutionContext, Future}
 
-trait EspItTest extends LazyLogging with WithHsqlDbTesting with TestPermissions with BeforeAndAfterEach {
+@deprecated("Consider using NuItTest with NuScenarioConfigurationHelper instead. Will be removed in the future.")
```
arkadius, line 55: We rather don't add deprecations for internal purposes. Instead, we add `TODO`. Thanks to we have deprecation warnings silenced in only those places that are necessary and see new deprecations that we have to fix. WTDY about this approach?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -181,6 +181,8 @@ class FlinkProcessRegistrar(compileProcess: (CanonicalProcess, ProcessVersion, R
         case part@SinkPart(sink: FlinkSink, _, contextBefore, _) =>
           registerSinkPark(start, part, sink, contextBefore)
         case part: SinkPart =>
+          // todo: fixme "part.obj" is not stringified well
```
arkadius, line 184: We rather use upper case TODO / FIXME marks

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/NuResourcesTest.scala
coutoPL: I prefer to use @deprecation because IMO it's much more useful. Eg. when the deprecated class/method is used in a newly created code, the author of the code has a chance to notice that he should not use the deprecated thing. It doesn't apply to the comment-based approach - the coder can easily miss the info that the code is deprecated and use it in a new code. 

But I'll do it in the way you propose because of the cross-compilation I cannot use @nowarn annotation (yeah, silencer could be the solution)

arkadius (approved):
Good job

PR #4586, opened 2023-08-20
NU-976 remove dependency to Popover from react-bootstrap

JulianWielga:

### DIFF
--- a/designer/client/src/components/Process/ProcessStateIcon.tsx
+++ b/designer/client/src/components/Process/ProcessStateIcon.tsx
@@ -1,13 +1,9 @@
-import React, { PropsWithChildren, useMemo } from "react";
+import React from "react";
 import { useTranslation } from "react-i18next";
 import { ProcessStateType, ProcessType } from "./types";
-import { Popover } from "react-bootstrap";
-import { OverlayTrigger } from "react-bootstrap/lib";
 import ProcessStateUtils from "./ProcessStateUtils";
-import { css } from "@emotion/css";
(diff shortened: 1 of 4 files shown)


designer/client/src/components/Process/ProcessStateIcon.tsx
```diff
@@ -33,45 +29,38 @@ function Errors({ state }: { state: ProcessStateType }) {
...
     const tooltip = ProcessStateUtils.getStatusTooltip(process, processState);
 
+    const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(null);
+
     return (
-        <StateIconPopover processName={process.name} processState={processState} tooltip={tooltip}>
-            <UrlIcon src={icon} title={tooltip} />
-        </StateIconPopover>
+        <>
+            <UrlIcon src={icon} title={tooltip} onClick={(event) => setAnchorEl(event.currentTarget)} />
```
Dzuming, line 40: It's not the regression because I'm able to reproduce it on the Staging, but on Safari (16.4) The icon is missing. As I see, also metrics icon is missing. Can you take a look at it in this pr, or maybe we should create another task for it? 

image

JulianWielga: will fix it here

JulianWielga: no, i'll do separate pr. it's to big to mix with this.

Dzuming (approved):

PR #333, opened 2019-11-08
Improve multiple UI elements

PPiedel:
This fixes #315
This fixes #332 
This fixes #305 
This fixes #316

PR #4588, opened 2023-08-21
Customize search and replace ace editor styles

DeamonDev:
Currently text input color inside Search & Replace Ace Editor's  is very unfriendly. Here I propose new colours 

====== BEFORE ======
image

====== AFTER ======
Screenshot Capture - 2023-08-21 - 15-27-58

### DIFF
--- a/designer/client/src/brace/theme/nussknacker.css
+++ b/designer/client/src/brace/theme/nussknacker.css
@@ -16,7 +16,7 @@
 }
 
 .ace-nussknacker .ace_search_field {
-    color: #a8ce91;
+    color: #ccc;
 }
 
 .ace-nussknacker {
--- a/designer/client/src/brace/theme/nussknacker.css
+++ b/designer/client/src/brace/theme/nussknacker.css
@@ -10,6 +10,15 @@
     width: 1px;
     background: #555651
 }
+
+.ace-nussknacker .ace_search {
+    background: #818181;
+}
+
+.ace-nussknacker .ace_search_field {
+    color: #a8ce91;
+}
+
 .ace-nussknacker {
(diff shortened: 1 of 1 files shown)


designer/client/src/brace/theme/nussknacker.css
```diff
@@ -10,6 +10,15 @@
     width: 1px;
     background: #555651
 }
+
+.ace-nussknacker .ace_search {
+    background: #818181;
+}
+
+.ace-nussknacker .ace_search_field {
+    color: #a8ce91;
```
Dzuming, line 19: In the case of edit node fields, we use `#ccc` color, maybe it's worth it to be consistent with existing fields?

designer/client/src/brace/theme/nussknacker.css
```diff
@@ -10,6 +10,15 @@
     width: 1px;
     background: #555651
 }
+
+.ace-nussknacker .ace_search {
+    background: #818181;
```
Dzuming, line 15: In the case of edit node fields, we use `#333` color, maybe it's worth it to be consistent with existing fields?

DeamonDev: I  don't think it is good idea since `#333` is the same as input background.

Dzuming: Ohh, I thought that it was the color of a background of the input :P

Dzuming (approved):
LGTM

designer/client/src/brace/theme/nussknacker.css
DeamonDev: Ok, changed.

PR #4601, opened 2023-08-23
[bugfix] interpreters fixes

coutoPL:
Fixed issues:

- `IOShape`: use `IO.apply` instead of `IO.pure` 
- `SyncInterpretationFunction`: in case of IO monad usage, the code should be symmetric to the Future-based implementation (added timeout to the sync run)
- `AsyncInterpretationFunction`: in case of the Future-based implementation - the callback was not invoked
- `interpreter.scala`: converting from Future to Effect (IO or Future) should be done using the sync execution context (now the intention is more clear than before)

IMO the issues above should be solved before bumping cats-effect lib. Ref: https://github.com/TouK/nussknacker/pull/4287

--- 

Benchmarks:

Before changes:
```
[info] Benchmark                                           Mode  Cnt        Score        Error  Units
[info] ManyParamsInterpreterBenchmark.benchmarkAsync      thrpt    8   256944.447 ±  14652.777  ops/s
[info] ManyParamsInterpreterBenchmark.benchmarkSync       thrpt    8   519288.756 ±   4219.996  ops/s
[info] OneParamInterpreterBenchmark.benchmarkAsyncIO      thrpt    8   174094.575 ±   6846.656  ops/s
[info] OneParamInterpreterBenchmark.benchmarkFutureAsync  thrpt    8    93019.418 ±   8049.856  ops/s
[info] OneParamInterpreterBenchmark.benchmarkFutureSync   thrpt    8  2543090.275 ± 135655.268  ops/s
[info] OneParamInterpreterBenchmark.benchmarkSyncIO       thrpt    8  1791769.229 ±  89304.281  ops/s
```

After changes:
```
[info] Benchmark                                           Mode  Cnt        Score       Error  Units
[info] ManyParamsInterpreterBenchmark.benchmarkAsync      thrpt    8   231363.777 ±  1324.431  ops/s
[info] ManyParamsInterpreterBenchmark.benchmarkSync       thrpt    8   422723.163 ± 17596.191  ops/s
[info] OneParamInterpreterBenchmark.benchmarkAsyncIO      thrpt    8   201109.627 ± 23130.016  ops/s
[info] OneParamInterpreterBenchmark.benchmarkFutureAsync  thrpt    8    76491.629 ± 12645.501  ops/s
[info] OneParamInterpreterBenchmark.benchmarkFutureSync   thrpt    8  2571591.618 ± 34357.808  ops/s
[info] OneParamInterpreterBenchmark.benchmarkSyncIO       thrpt    8  1517608.097 ± 33637.059  ops/s
```

### DIFF
--- a/benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
+++ b/benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
@@ -26,13 +26,13 @@ class ManyParamsInterpreterBenchmark {
     .enricher("e1", "out", "service", (1 to 20).map(i => s"p$i" -> ("''": Expression)): _*)
     .emptySink("sink", "sink")
 
-  private def prepareInterpreter(executionContext: ExecutionContext) = {
+  private def prepareIoInterpreter(executionContext: ExecutionContext) = {
     val setup = new InterpreterSetup[String].sourceInterpretation[IO](process, Map("service" -> new ManyParamsService(executionContext)), Nil)
     (ctx: Context) => setup(ctx, executionContext)
   }
 
-  private val interpreterSyncIO = prepareInterpreter(SynchronousExecutionContext.ctx)
-  private val interpreterAsyncIO = prepareInterpreter(ExecutionContext.Implicits.global)
+  private val interpreterSyncIO = prepareIoInterpreter(SynchronousExecutionContext.ctx)
+  private val interpreterAsyncIO = prepareIoInterpreter(ExecutionContext.Implicits.global)
 
   @Benchmark
   @BenchmarkMode(Array(Mode.Throughput))
--- a/benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
+++ b/benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
@@ -31,14 +31,14 @@ class ManyParamsInterpreterBenchmark {
     (ctx: Context) => setup(ctx, executionContext)
   }
 
-  private val interpreterSync = prepareInterpreter(SynchronousExecutionContext.create())
-  private val interpreterAsync = prepareInterpreter(ExecutionContext.Implicits.global)
+  private val interpreterSyncIO = prepareInterpreter(SynchronousExecutionContext.ctx)
+  private val interpreterAsyncIO = prepareInterpreter(ExecutionContext.Implicits.global)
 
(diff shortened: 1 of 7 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
```diff
@@ -262,15 +258,19 @@ object Interpreter {
...
+    }
   }
 
   class FutureShape(implicit ec: ExecutionContext) extends InterpreterShape[Future] {
 
     override def monad: Monad[Future] = cats.instances.future.catsStdInstancesForFuture(ec)
 
-    override def fromFuture[T](implicit ec: ExecutionContext): Future[T] => Future[Either[T, Throwable]] = transform(_)
+    override def fromFuture[T]: Future[T] => Future[Either[T, Throwable]] = {
+      transform(_)(SynchronousExecutionContext.ctx)
```
lciolecki, line 272: Let's be consistent and do `implicit val ctx = SynchronousExecutionContext.ctx`?

benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
```diff
@@ -31,22 +31,22 @@ class ManyParamsInterpreterBenchmark {
     (ctx: Context) => setup(ctx, executionContext)
   }
 
-  private val interpreterSync = prepareInterpreter(SynchronousExecutionContext.create())
-  private val interpreterAsync = prepareInterpreter(ExecutionContext.Implicits.global)
+  private val interpreterSyncIO = prepareInterpreter(SynchronousExecutionContext.ctx)
```
lciolecki, line 34: prepareIOInterpreter?

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
coutoPL: nope, there is already `ec` implicit in scope

coutoPL:
ref: https://github.com/TouK/nussknacker/pull/4287

interpreter/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala
lciolecki: So, we we pass this `SynchronousExecutionContext.ctx`?

coutoPL: as you can see, in the previous implementation, two execution contexts were passed:
1. in the constructor - it was used to create a monad instance
2. in the `fromFuture` method - it was used to do the `transform`. 

In this change, I don't want to remove the constructor execution context, because it's not related to any of the issues I found.

arkadius (approved):
LGTM

PR #4602, opened 2023-08-24
Remove isSubprocess backward compatibility

Filemon279:


mslabek (approved):
Can you also add info for clarity that this change also removes backward compatibility for ProcessAction? Changing the PR name would be ok for me. Other than that, LGTM :)

PR #4603, opened 2023-08-24
Fix migration by adding missing field in SubprocessInput migration

Filemon279:

### DIFF
--- a/designer/server/src/main/scala/db/migration/V1_043__RenameSubprocessToFragmentDefinition.scala
+++ b/designer/server/src/main/scala/db/migration/V1_043__RenameSubprocessToFragmentDefinition.scala
@@ -1,62 +1,11 @@
 package db.migration
 
-import io.circe.Json._
-import io.circe._
+import io.circe.Json
 import pl.touk.nussknacker.ui.db.migration.ProcessJsonMigration
 
 trait V1_043__RenameSubprocessToFragmentDefinition extends ProcessJsonMigration {
 
-  override def updateProcessJson(jsonProcess: Json): Option[Json] =
-    V1_043__RenameSubprocessToFragmentDefinition.updateProcessJson(jsonProcess)
-
-}
-
-object V1_043__RenameSubprocessToFragmentDefinition {
-
-  private val legacyProperty = "subprocessParams"
-  private val newProperty = "fragmentParams"
-
-  private[migration] def updateProcessJson(jsonProcess: Json): Option[Json] = {
-    val updatedNodes = updateField(jsonProcess, "nodes", updateCanonicalNodes)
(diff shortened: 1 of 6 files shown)


designer/server/src/main/scala/db/migration/V1_046__RenameSubprocessToFragmentDefinition.scala
```diff
@@ -0,0 +1,65 @@
...
+    (obj.hcursor downField field).success.flatMap(_.withFocus(update).top).getOrElse(obj)
+  }
+
+  private def updateNodes(array: Json, fun: Json => Json) = fromValues(array.asArray.getOrElse(List()).map(fun))
+
+  private def updateCanonicalNode(node: Json): Json = {
+    node.hcursor.downField("type").focus.flatMap(_.asString).getOrElse("") match {
+      case "SubprocessInput" =>
+        val updateType = updateField(node, "type", _ => Json.fromString("FragmentInput"))
+        updateField(updateType, "outputs", outputs => outputs.mapObject(obj => obj.mapValues(updateCanonicalNodes)))
```
Filemon279, line 39: This is a field that was missing

designer/server/src/main/scala/db/migration/V1_046__RenameSubprocessToFragmentDefinition.scala
```diff
@@ -0,0 +1,65 @@
...
+      case "FragmentInput" => updateField(node, "outputs", outputs => outputs.mapObject(obj => obj.mapValues(updateCanonicalNodes)))
```
Filemon279, line 40: And this is a line that we need to add for this to work after first migration

designer/server/src/main/scala/db/migration/V1_043__RenameSubprocessToFragmentDefinition.scala
```diff
@@ -1,62 +1,11 @@
...
-        case obj if obj.contains(legacyProperty) =>
-          Json.fromJsonObject(obj
-            .add(newProperty, obj(legacyProperty).get)
-            .filterKeys(_ != legacyProperty))
-        case _ => json
-      }.getOrElse(json)
-    }).top.getOrElse(node)
-  }
+  //This migration was missing "outputs" field in SubprocessInput migration so it was replaced with empty migration and fixed in another one.
+  override def updateProcessJson(jsonProcess: Json): Option[Json] = Some(jsonProcess)
```
piotrp, line 9: You can return None:
```scala
* <li>if we want to continue and fall back to previous json - the implementation should return None</li>
```

piotrp (approved):

piotrp:
Add changelog entry for 1.11.1

Filemon279:
changelog added, thanks

PR #4606, opened 2023-08-25
NU-999 fixed toolbar problems

JulianWielga:
fixed tab indexes
fixed button styles
fixed safari problems
fixed some performance problems


designer/client/src/components/graph/GraphContext.tsx
```diff
@@ -1,14 +1,25 @@
...
+    graph: _graphGetter,
+    children,
+}: PropsWithChildren<{
+    graph: GraphContextType;
+}>) => {
+    const graph = _graphGetter();
+
+    // force new value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const graphGetter = useMemo(() => () => _graphGetter(), [_graphGetter, graph]);
```
Dzuming, line 19: If i understand correctly, every component rerender, the new value is assigned, so IMO this useMemo has no sense or i missed something?

designer/client/src/reducers/selectors/graph.ts
```diff
@@ -8,7 +8,7 @@ import { ProcessCounts } from "../graph";
 import { RootState } from "../index";
 import { getProcessState } from "./scenarioState";
 
-export const getGraph = (state: RootState): RootState["graphReducer"] => state.graphReducer;
+export const getGraph = (state: RootState) => state.graphReducer.history.present;
```
Dzuming, line 11: Tests failed here (Nice feature btw)
image

Dzuming (approved):

designer/client/src/components/graph/GraphContext.tsx
JulianWielga: new value is provided only when `graph` changes not every time `GraphProvider` renders.

it's sth like this
https://codesandbox.io/s/admiring-curie-dkms53?file=/src/App.js

PR #4612, opened 2023-08-29
[NU-981] app resources integration tests refactor

coutoPL:
- `AppResourcesSpec` refactoring - now, the spec is Akka HTTP  independent 
- introduced `MockableDeploymentManager` in tests - it allows to easily mock the scenario execution statuses in integration tests


utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureMatchers.scala
```diff
@@ -0,0 +1,170 @@
...
+    private def compareArrays(a1: Arr, a2: Arr, parent: JsonValueParent) = {
+      if (a1 == a2) {
+        Validated.Valid(())
+      } else if (a1.value.size == a2.value.size) {
+        a1.value.zip(a2.value).zipWithIndex
+          .foldLeft(List.empty[Validated[String, Unit]]) {
+            case (validationResults, ((elemA1, elemA2), idx)) =>
+              compareJsons(elemA1, elemA2, JsonValueParent.Array(parent, idx)) :: validationResults
+          }
+          .reverse
```
arkadius, line 103: Can't be replaced by just `sequence` from traverse syntax?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureMatchers.scala
```diff
@@ -0,0 +1,170 @@
...
+        case _ => false
+      }
+    }
+
+    override def describeTo(description: Description): Unit = {
+      description.appendValue(expectedJsonString)
+    }
+  }
+
+  private class MatchJsonWithRegexValues(expectedJsonWithRegexValuesString: String)
```
arkadius, line 41: Nothing ready is available for all of this things? I see a lot of potential code that can has bugs. Is it possible to at least use some ready json traverse and use monocle to check only nested regexes?

designer/server/src/test/resources/designer-api-test.conf
```diff
@@ -0,0 +1,191 @@
...
+  password: ""
+  connectionTimeout: 30000
+}
+
+authentication:  {
+  method: "BasicAuth"
+  usersFile: "designer/server/src/test/resources/basicauth-users.conf"
+}
+
+http {
```
arkadius, line 14: Why we need another config? Can't we use just designer.conf? Variety of config make changes in config structure hard

designer/server/src/test/resources/logback-test.xml
```diff
@@ -7,10 +7,12 @@
         </encoder>
     </appender>
 
-    <root level="WARN">
+    <root level="INFO">
```
arkadius, line 10: At some point in time, we had a problem that we were hitting limit of logs in GH actions. I wonder if this problem won't back after this change

designer/server/src/test/resources/designer-api-test.conf
coutoPL: I'd not say "hard". Rather "laborious". However, having multiple conf files prepared to be used in some contexts is more readable. 

But ok, I'm not aware yet of all specs contexts in Nu, so I refactor it to use the existing designer.conf (I'm changing part of it in the `WithMockableDeploymentManager` trait.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/AppResourcesSpec.scala
```diff
@@ -1,198 +1,197 @@
...
+        ProcessName("id1") -> SimpleStateStatus.Running
       )
     )
 
-    val result = Get("/app/healthCheck/process/deployment") ~> withPermissions(resources, testPermissionRead)
-
-    result ~> check {
-      status shouldBe StatusCodes.OK
-    }
+    given()
```
lciolecki, line 123: I think you cane move whole code to a function with response code and responseBody?

coutoPL: what will be the purpose of the change you propose?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureMatchers.scala
arkadius: Too complicated to trus it works - I :heart: it ;)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/AppResourcesSpec.scala
arkadius: I think that @lciolecki would like to optimize the amount of code, and BDD is standing in opposition to drastic repetition reductions. We should pick one of two: DRY as much as you can vs BDD with a little dose of repetition. I like the second approach, what about you @lciolecki ?

arkadius (approved):
LGTM

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/AppResourcesSpec.scala
coutoPL: IMO in tests the approach DRY is not always right. One of the most important things is the readability of tests. Currently, thanks to rest-assured DSL, we have a nice given-when-then approach, actually forced by the lib. 

So, I agree with @arkadius and I'd stay with the current approach.

PR #4629, opened 2023-08-31
Fix resource close

piotrp:
Fixes bug introduced by #4514: reloading model using `/api/app/processingtype/reload` API causes application's shared `SttpBackend` to close.

Old backend implementation used [`AkkaHttpBackend`](https://github.com/softwaremill/sttp/blob/v3/akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala), which was created with `terminateActorSystemOnClose = false`, i.e. it did nothing when `close()` was called. Our new backend, `AsyncHttpClientFutureBackend`, defaults to closing all of its resources (`closeClient = true`).


designer/processReports/src/main/scala/pl/touk/nussknacker/processCounts/influxdb/InfluxGenerator.scala
```diff
@@ -62,8 +62,6 @@ private[influxdb] class InfluxGenerator[F[_]](config: InfluxConfig, env: String)
   private def parseInfluxDate(date:String) : Instant =
     ZonedDateTime.parse(date, DateTimeFormatter.ISO_ZONED_DATE_TIME).toInstant
 
-  def close(): F[Unit] = influxClient.close()
```
coutoPL, line 65: `SimpleInfluxClient` has a close method so it looks like we need to remove it too. Especially, the method implementation is wrong - it closes the implicitly passed `backend` (it's not an owner of the resource, so it should not bother stopping it).

coutoPL: ok. I didn't notice changes in SimpleInfluxClient

PR #4631, opened 2023-08-31
latest deployments in counts

JulianWielga:


designer/client/src/i18n.ts
```diff
@@ -15,7 +15,7 @@ i18n.init({
     defaultNS: "main",
     fallbackLng: "en",
     backend: {
-        loadPath: urlJoin(BACKEND_STATIC_URL, "assets/locales/{{lng}}/{{ns}}.json"),
+        loadPath: urlJoin(__DEV__ ? "/" : BACKEND_STATIC_URL, "assets/locales/{{lng}}/{{ns}}.json"),
```
Dzuming, line 18: Why do we need this change?

Dzuming (approved):

designer/client/src/i18n.ts
JulianWielga: to force use of newly edited value in dev

PR #4640, opened 2023-09-05
Expand timestamp support to handle more types/formats

wrzontek:


engine/flink/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/flink/UniversalTimestampFieldAssigner.scala
```diff
@@ -3,17 +3,25 @@ package pl.touk.nussknacker.engine.schemedkafka.source.flink
...
 
 object UniversalTimestampFieldAssigner {
 
   def apply[K, V](fieldName: String): SerializableTimestampAssigner[ConsumerRecord[K, V]] = new SerializableTimestampAssigner[ConsumerRecord[K, V]] {
     override def extractTimestamp(element: ConsumerRecord[K, V], recordTimestamp: Long): Long = {
-      val timestampOpt = Option(element.value() match {
-        case genericRecord: GenericRecord => genericRecord.get(fieldName)
-        case typedMap: java.util.Map[String, Any]@unchecked => typedMap.get(fieldName)
+      val timestampOpt: Option[Long] = Option(element.value() match {
+        case genericRecord: GenericRecord => supportedTypeToMillis(
```
arkadius, line 13: Let's do `supportedTypeToMillis` once in this place.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
```diff
@@ -82,7 +83,7 @@ object DelayedKafkaSourceFactory {
   def validateTimestampField(field: String, typingResult: TypingResult)(implicit nodeId: NodeId): List[ProcessCompilationError] = {
     typingResult match {
       case TypedObjectTypingResult(fields, _, _) => fields.get(field) match {
-        case Some(fieldTypingResult) if List(Typed[java.lang.Long], Typed[Long]).contains(fieldTypingResult) => List.empty
+        case Some(fieldTypingResult) if List(Typed[java.lang.Long], Typed[Long], Typed[Int], Typed[Instant], Typed[LocalDate], Typed[ZonedDateTime]).contains(fieldTypingResult) => List.empty
```
arkadius, line 86: It would be nice to have this list somewhere near runtime (`supportedTypeToMillis`) so it will be harder to desynchronize this code

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+    } else { // l is nearly certainly given in seconds
+      l * 1000
+    }
+
+  def supportedTypeToMillis(value: Any, errorMsg: String): Long = value match {
+    case v: Long => normalizeTimestampToMillis(v)
+    case v: Int => normalizeTimestampToMillis(v)
+    case v: Instant => v.toEpochMilli
+    case v: LocalDate => Timestamp.valueOf(v.atStartOfDay()).getTime
+    case v: ZonedDateTime => Timestamp.valueOf(v.toLocalDateTime).getTime
```
arkadius, line 21: Please no. It is not the correct way to handle it. What about dates that are duplicated during a year because of DST? Please use `.toInstant`. `LocalDateTime` is in most cases bad choice for programmable usage. You should very rarely use it.

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
```diff
@@ -0,0 +1,29 @@
+package pl.touk.nussknacker.engine.util
+
+import java.sql.Timestamp
```
arkadius, line 3: What sql do you have in your code so you have to use SQL types?

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+  def supportedTypeToMillis(value: Any, errorMsg: String): Long = value match {
+    case v: Long => normalizeTimestampToMillis(v)
+    case v: Int => normalizeTimestampToMillis(v)
+    case v: Instant => v.toEpochMilli
+    case v: LocalDate => Timestamp.valueOf(v.atStartOfDay()).getTime
+    case v: ZonedDateTime => Timestamp.valueOf(v.toLocalDateTime).getTime
+    case v: OffsetDateTime => Timestamp.valueOf(v.toLocalDateTime).getTime
+    case v => normalizeTimestampToMillis(Try(v.asInstanceOf[Long]).getOrElse {
```
arkadius, line 23: Why do you do this? You already have a Long case and a Int case in pattern matching

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
wrzontek: it's just an IntelliJ miss-import, my bad

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
wrzontek: is    
 `case v: LocalDate => v.atStartOfDay(ZoneId.systemDefault()).toInstant.toEpochMilli`
ok?

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
wrzontek: I misshandled `null`
handling `Any`, especially gotten from java code, can be a bit confusing

wrzontek:
I've removed the normalizing heuristic part, to make this PR focus only on supporting more timestamp formats.

utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/TimestampUtils.scala
arkadius: This case led us to the place that LocalDate is tricky. Lets imagine situation when you have scenario with join with LocalDate on one branch and Instant on second. He think about LocalDate in Autralia because he know that these events are from Australia. How you will know this time zone?

wrzontek: it's not possible to sensibly handle without additional info about time zone, and for example, avro docs specifically mention that `date` doesn't carry information about time-zone

I think it's more trouble than it's worth, I doubt timestamp in date (without time) format are used with any kind of popularity, i'll just remove support for LocalDate

wrzontek:
i think the PR, in this reduced scope, is ready to be undrafted

arkadius:
It's even ready to merge! :)

PR #4643, opened 2023-09-05
Heuristically normalize Int/Long timestamps to milliseconds

wrzontek:


arkadius:
I propose to close this MR, we can reopen it when we decide that it becomes more important

PR #4649, opened 2023-09-06
Added the checklist to PR tempalate

arkadius:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,13 @@
-<!--
-Thank you for sending the PR! We appreciate you spending the time to work on these changes.
+## Describe your changes
 
-Please make sure that you added appropriate entry in docs/Changelog.md and docs/MigrationGuide.md
-
-You can learn more about contributing to Nussknacker here: https://github.com/TouK/nussknacker/blob/staging/CONTRIBUTING.md
-
-Happy contributing!
-
--->
+## Checklist before merge
+- [ ] Code is cleaned from temporary changes and commented lines 
+- [ ] Code formatting checked
+- [ ] Parts of the code that are not easy to understand are documented in the code
+- [ ] Automatic tests added
+- [ ] Showcase in dev-application.conf added if necessary 
+- [ ] Documentation added or updated
+- [ ] Review done                              
+- [ ] All comments resolved
+- [ ] Changelog.md entry added if necessary
+- [ ] MigrationGuide.md entry added in the corresponding subcategory if necessary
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,10 +1,13 @@
...
+## Describe your changes
+
+## Checklist before merge
+- [ ] Code is cleaned from temporary changes and commented lines 
+- [ ] Code formatting checked
+- [ ] Parts of the code that are not easy to understand are documented in the code
+- [ ] Automatic tests added
+- [ ] Showcase in dev-application.conf added if necessary 
+- [ ] Documentation added or updated
+- [ ] Review done
```
Dzuming, line 10: Isn't it redundant? There is no way to merge when there is no approval.

.github/pull_request_template.md
```diff
@@ -1,10 +1,13 @@
...
+
+## Checklist before merge
+- [ ] Code is cleaned from temporary changes and commented lines 
+- [ ] Code formatting checked
+- [ ] Parts of the code that are not easy to understand are documented in the code
+- [ ] Automatic tests added
+- [ ] Showcase in dev-application.conf added if necessary 
+- [ ] Documentation added or updated
+- [ ] Review done                              
+- [ ] All comments resolved
```
Dzuming, line 11: Not sure how it works now, but maybe we can set restrictions to merge when there are unresolved comments, instead of adding this point?

.github/pull_request_template.md
```diff
@@ -1,10 +1,13 @@
...
+- [ ] Changelog.md entry added if necessary
```
Dzuming, line 12: Do we need this `if necessary`? Maybe we should introduce an optional "label" to increase readability? So something like

`Changelog.md entry added if necessary (Optional)`

.github/pull_request_template.md
```diff
@@ -1,10 +1,13 @@
...
-Please make sure that you added appropriate entry in docs/Changelog.md and docs/MigrationGuide.md
-
-You can learn more about contributing to Nussknacker here: https://github.com/TouK/nussknacker/blob/staging/CONTRIBUTING.md
-
-Happy contributing!
-
--->
+## Describe your changes
+
+## Checklist before merge
```
coutoPL, line 3: WDYT about "adding jira number in PR title" point? If some jira exists, obviously.

arkadius: I added it. However, take a look if it is fine for you. I tried to add it in the form that will be suitable both for internal PRs and external contributions

.github/pull_request_template.md
Dzuming: Looks good 👌

Dzuming (approved):

coutoPL (approved):
LGTM

PR #4656, opened 2023-09-08
NU-1044 remove stylus files from common folder

bohdanprog:
Removing stylus CSS from `designer/client/src/components/common` updates components that used stylus to using styled


designer/client/src/stylesheets/variables.ts
```diff
@@ -0,0 +1,4 @@
+export const variables = {
```
Dzuming, line 1: Based on the name of this component, it sounds like an MUI list divider, we can use it here. https://mui.com/material-ui/react-divider/

designer/client/src/components/common/ListSeparator.tsx
```diff
@@ -1,7 +1,14 @@
...
+import { variables } from "../../stylesheets/variables";
+
+const StyledSeparator = styled.hr(({ dark }: { dark?: boolean }) => ({
+    border: "none",
+    borderBottom: "1px solid",
+    margin: "10px 0",
+    color: dark ? variables.panelHeaderBackground : variables.panelTitleTextColor,
+}));
 
 export function ListSeparator(props: { dark?: boolean }): JSX.Element {
```
Dzuming, line 12: Based on the name of this component, it sounds like an MUI list divider, we can use it in our codebase instead of this component?. https://mui.com/material-ui/react-divider/

Dzuming (approved):

PR #4659, opened 2023-09-08
[NU-1028] toolbar collapse fix

JulianWielga:
hiding version info on touchscreens fixed


Dzuming:
@JulianWielga I checked it on ipad 6 mini, but the left/right expand/collapse toolbar icon is not visible

Simulator Screenshot - iPad mini (6th generation) - 2023-09-11 at 08 17 31

Dzuming (approved):

PR #4666, opened 2023-09-12
Additional fragment parameter parameters

wrzontek:
## Checklist before merge
- [ ] Code formatting checked

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
@@ -89,7 +89,10 @@ case class Parameter(name: String,
                      branchParam: Boolean,
                      isLazyParameter: Boolean,
                      scalaOptionParameter: Boolean,
-                     javaOptionalParameter: Boolean) extends NodeDependency {
+                     javaOptionalParameter: Boolean,
+                     validationErrorMessage: Option[String] = None,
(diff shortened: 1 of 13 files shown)


wrzontek:
OK i see that my attempts at the SPEL validators are highly amateurish, I'll read more into (and use) the classes in and around `SpelExpression.scala`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -89,7 +89,10 @@ case class Parameter(name: String,
                      branchParam: Boolean,
                      isLazyParameter: Boolean,
                      scalaOptionParameter: Boolean,
-                     javaOptionalParameter: Boolean) extends NodeDependency {
+                     javaOptionalParameter: Boolean,
+                     validationErrorMessage: Option[String] = None,
```
arkadius, line 93: It makes no sense. We can have many validators and each validator has it's own message.

wrzontek: that's the user-defined one for the user-defined validator, I guess it could be named clearer

wrzontek: or even bundled together with the CustomExpressionParameterValidator class

PR #4693, opened 2023-09-19
Change suggestions endpoint to work with processingType rather than p…

Filemon279:
…rocessId

## Checklist before merge
- [ ] Code formatting checked


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesResources.scala
```diff
@@ -114,21 +113,20 @@ class NodesResources(val processRepository: FetchingProcessRepository[Future],
...
-                expressionSuggestionRequest.expression,
-                expressionSuggestionRequest.caretPosition2d,
-                expressionSuggestionRequest.variables,
-              )
-            }
-          }
+        }
+      }
+    } ~ pathPrefix("parameters" / Segment / "suggestions") { processingType =>
+      val modelData = typeToConfig.forTypeUnsafe(processingType)
```
piotrp, line 119: What happens when we pass invalid `processingType`? We should return some sane error message

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesResources.scala
```diff
@@ -114,21 +113,20 @@ class NodesResources(val processRepository: FetchingProcessRepository[Future],
...
-              expressionSuggester.expressionSuggestions(
-                expressionSuggestionRequest.expression,
-                expressionSuggestionRequest.caretPosition2d,
-                expressionSuggestionRequest.variables,
-              )
-            }
-          }
+        }
+      }
+    } ~ pathPrefix("parameters" / Segment / "suggestions") { processingType =>
```
arkadius, line 118: The idea of the current approach was to make it as easy as possible to merge this feature with a validation mechanism. In the bright future, we don't want to call BE twice: for validations and suggestions purpose. Instead of this we would like to call it once and have consistent validations with suggestions. Why current API is harmful? Also, I'm wondering how does currently work `#meta` suggestions. Does it work at all?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesResources.scala
Filemon279: Thanks, I've added handler for that but I've choose Not Found (since there is no processingType found) - WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesResources.scala
arkadius: I can't imagine such a tool, looks like a brilliant idea! :) I see one problem in the code that was left that made this change so easy. Currenltly /parameters/$processName/validate endpoint takes all variables but global variables as a parameters (so `meta` variable is not passed). On the other hand suggestions takes all variables including global variables. Maybe you have an idea to unify it in one of two possible ways so it will be easier to merge them in the future?

Filemon279:
Closing in favor of : https://github.com/TouK/nussknacker/pull/4697

PR #4698, opened 2023-09-20
[NU-1041] Adapted nussknacker core to work with ververica deployment manager

pielas:
## Checklist before merge
- [ ] Code formatting checked


build.sbt
```diff
@@ -698,7 +698,7 @@ lazy val flinkExecutor = (project in flink("executor")).
         "org.apache.flink" % "flink-streaming-java" % flinkV % "provided",
         "org.apache.flink" % "flink-runtime" % flinkV % "provided",
         "org.apache.flink" % "flink-statebackend-rocksdb" % flinkV % "provided",
-        "org.apache.flink" % "flink-metrics-dropwizard" % flinkV,
```
arkadius, line 701: Are you sure about this change? Will it work correctly with test from file mechanism which run embedded Flink on the designer side?

pielas: it looks like it doesn't. I've read in comment is build.sbt that we use bundled metrics on flink. I assumed it meant that they should be provided, so we don't have to bring our own library in general as it is provided on ververica.

pielas: I've recently been able to fix all tests

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/ClassPaths.scala
```diff
@@ -7,7 +7,9 @@ object ClassPaths {
   val commonClasspath = List(
     s"./engine/flink/components/base/target/scala-${ScalaMajorVersionConfig.scalaMajorVersion}/flinkBase.jar",
     s"./engine/flink/components/kafka/target/scala-${ScalaMajorVersionConfig.scalaMajorVersion}/flinkKafka.jar",
-    s"./engine/flink/executor/target/scala-${ScalaMajorVersionConfig.scalaMajorVersion}/flinkExecutor.jar"
+    s"./engine/flink/executor/target/scala-${ScalaMajorVersionConfig.scalaMajorVersion}/flinkExecutor.jar",
+    "./engine/flink/executor/it-libs/flink-additional-deps/flink-metrics-dropwizard.jar",
```
arkadius, line 11: Shouldn't we put these libs into some target? Doesn't git try to track them?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkModelJar.scala
```diff
@@ -34,14 +36,29 @@ class FlinkModelJar extends LazyLogging {
     tempFile
   }
 
+  private def additionalArtifactsToIncludeInJar(): List[URL] = {
```
arkadius, line 39: I thought about an idea that we explicitly add `flink-additional-deps` into `modelConfig/classPath` in our `application.conf` and `dev-application.conf` in our distribution and here have a `if classpath doesn't contain flink-additional-deps then add them` code. Thanks to that this code will be only needed in the transitive period and after some time, we could remove it and stay with the explicit approach. WDYT?

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/ClassPaths.scala
pielas: ah, I forgot to commit .gitignore file. Do you think it would be better to place them under target directory?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkModelJar.scala
pielas: sounds like a good idea, I will make it work like that

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/ClassPaths.scala
arkadius: They are kind of products of our build script so IMO it is fine to keep them in the `target` directory. Maybe someone could want to invalidate them and wanted to do it via `sbt clean`? Also, it is nice to have as few directories as possible in modules directories.

arkadius (approved):

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkModelJar.scala
```diff
@@ -34,14 +36,33 @@ class FlinkModelJar extends LazyLogging {
...
+      .map(filename => new File(filename))
+      .filter(_.exists())
+      .map(_.toURI.toURL)
+    additionalJarUrls
+  }
+
   private def prepareModelFile(modelData: BaseModelData): File = {
     val tempFile = generateModelFileName()
-    modelData.modelClassLoaderUrls match {
+    val implicitlyIncludedArtifacts = additionalArtifactsToIncludeInJar().toSet -- modelData.modelClassLoaderUrls.toSet
```
arkadius, line 55: How you will override this in ververica DM? shouldn't be this if extracted into some protected methods or to some method's argument?

pielas: 1. I copied whole `FlinkModelJar` into ververica module. At first, I did it because of some custom adaptations to load scala 2.12 jars. It's not really needed now. However, I think it makes some sense to keep it separated from nussknacker core codebase.
2. It would be probably good to add such flag and add `dropwizard` word to directory name. It could help in case someone wanted to disable implicit loading. However, IMO there is no way somebody would use their own metrics library. 
3. ok

arkadius (approved):

PR #4707, opened 2023-09-22
[NU-1078] OIDC: Support for overrideFrontendAuthenticationStrategy configuration parameter

arkadius:
## Checklist before merge
- [x] Code formatting checked

### DIFF
--- a/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
+++ b/security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
@@ -4,7 +4,7 @@ import com.typesafe.config.Config
 import net.ceedubs.ficus.readers.ValueReader
 import pl.touk.nussknacker.engine.util.config.FicusReaders.forDecoder
 import pl.touk.nussknacker.engine.util.config.URIExtensions
-import pl.touk.nussknacker.ui.security.api.FrontendStrategySettings
+import pl.touk.nussknacker.ui.security.api.{AuthenticationConfiguration, FrontendStrategySettings}
 import pl.touk.nussknacker.ui.security.oauth2.ProfileFormat.OIDC
 import pl.touk.nussknacker.ui.security.oauth2.UsernameClaim.UsernameClaim
 import pl.touk.nussknacker.ui.security.oauth2.{JwtConfiguration, OAuth2Configuration, TokenCookieConfig}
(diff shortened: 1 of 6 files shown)


security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
```diff
@@ -90,6 +95,7 @@ object OidcAuthenticationConfiguration {
   import net.ceedubs.ficus.readers.EnumerationReader._
   import pl.touk.nussknacker.engine.util.config.CustomFicusInstances._
   import pl.touk.nussknacker.ui.security.api.AuthenticationConfiguration.authenticationConfigPath
+  private implicit val valueReader: ValueReader[FrontendStrategySettings] = forDecoder
```
coutoPL, line 98: WDYT about moving the implicit val to the `create` method?

security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2AuthenticationResourcesSpec.scala
```diff
@@ -102,4 +103,26 @@ class OAuth2AuthenticationResourcesSpec
       header[`Set-Cookie`] shouldBe Some(`Set-Cookie`(HttpCookie(name = cookieConfig.name, value = accessToken, httpOnly = true, secure = true, path = cookieConfig.path)))
     }
   }
+
+  it("should return authentication settings") {
+    Get(s"/authentication/oauth2/settings") ~> routes(authenticationResources()) ~> check {
+      status shouldBe StatusCodes.OK
+      val response = responseAs[Json]
+      response.hcursor.downField("authorizeUrl").as[String].value shouldBe defaultConfig.authorizeUri.toString + "?client_id=" + defaultConfig.clientId
```
coutoPL, line 111: maybe `response.hcursor.downField("authorizeUrl").as[String].value shouldBe s"${defaultConfig.authorizeUri.toString}?client_id=${defaultConfig.clientId}"` ?

security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationConfiguration.scala
arkadius: And what about implicit imports? should be in the `create` as well? Looks like it is the only usage of them.

coutoPL: IMO it'd be nice to move them too. E.g. when we remove the `create` method in the future, it'd be easier to keep the code clean.

coutoPL (approved):
LGTM

PR #4709, opened 2023-09-25
[NU-981] generating open api yaml

coutoPL:
## Describe your changes

- App API OpenAPI document was enhanced with examples and schemas are well-defined now
- generation of static OpenAPI document of Nu Designer API
- SBT task for generating the static document
- improvements that allow to maintain up to date state with the dynamically exposed OpenAPI and the static document mentioned above

## Checklist before merge
- [x] Code formatting checked


designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NuDesignerApiAvailableToExposeYamlSpec.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.scalatest.matchers.should.Matchers
+import pl.touk.nussknacker.ui.services.NuDesignerApiAvailableToExpose
+
+// if the test fails it probably means that you should regenerate the Nu Designer OpenAPI document
+// you can do it but running manually the object `GenerateDesignerOpenApiYamlFile` with main method or
+// using SBT's task: `sbt generateDesignerOpenApi`
+class NuDesignerApiAvailableToExposeYamlSpec extends AnyFunSuite with Matchers {
+
+  test("Nu Designer OpenAPI document with all available to expose endpoints has to be up to date") {
+    val currentNuDesignerOpenApiYamlContent = (pwd / "docs" / "api" / "internal" / "nu-designer-openapi.yaml").contentAsString
```
arkadius, line 14: Please don't user working directory. It can behave differently depending on many external things. Why not to use resources instead?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/AppApiEndpoints.scala
```diff
@@ -1,150 +1,267 @@
...
+    }
 
-      implicit val buildInfoDtoCodec: CirceCodec[BuildInfoDto] = {
+    @derive(schema)
+    final case class BuildInfoDto(name: String,
+                                  gitCommit: String,
+                                  buildTime: String,
+                                  version: String,
+                                  processingType: Map[String, Map[String, String]],
+                                  globalBuildInfo: Option[Map[String, String]] = None)
```
arkadius, line 261: Why it was better to extract these fields into a nested object instead of keeping them in the root level? Am I right that we don't have an option to say to tapir that we have json schema's `additionalProperties`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/NuDesignerApiAvailableToExpose.scala
```diff
@@ -0,0 +1,73 @@
...
+      title = NuDesignerApiAvailableToExpose.name,
+      version = ""
+    )
+
+    docs.toYaml
+  }
+
+  private def findApiEndpointsClasses() = {
+    val baseEndpointDefinitionsClass = classOf[BaseEndpointDefinitions]
+    val reflections = new Reflections(new ConfigurationBuilder().forPackages(baseEndpointDefinitionsClass.getPackageName))
```
arkadius, line 36: Can't we just have a list of definitions to be exposed in the same place as they exposed (`AppApiHttpService`) instead of using the "magic" of reflection? What if we want to bind two interfaces and expose one part of the API on one port and another on the second?

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/NuDesignerApiAvailableToExpose.scala
```diff
@@ -0,0 +1,73 @@
...
+      .getSubTypesOf(baseEndpointDefinitionsClass).asScala
+      .toList
+      .sortBy(_.getName)
+  }
+
+  private def findEndpointsInClass(clazz: Class[_ <: BaseEndpointDefinitions]) = {
+    val endpointDefinitions = createInstanceOf(clazz)
+    clazz
+      .getDeclaredMethods.toList
+      .filter(isEndpointMethod)
```
arkadius, line 47: Maybe we should keep these values as an enum entries and use https://github.com/lloydmeta/enumeratum for the discovery purpose. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
```diff
@@ -0,0 +1,21 @@
+package pl.touk.nussknacker.ui.util
+
+import better.files.Dsl._
+import cats.effect.{ExitCode, IO, IOApp}
+import com.typesafe.scalalogging.StrictLogging
+import pl.touk.nussknacker.ui.services.NuDesignerApiAvailableToExpose
+
+object GenerateDesignerOpenApiYamlFile extends IOApp with StrictLogging {
```
arkadius, line 8: I thought about having this generate logic in the same place as the test is. WDYT? We have similar approach in `ClassExtractionBaseTest` which is a base class for two test cases (`DefaultClassExtractionTest` and `DevClassExtractionTest`) We check there how changes in configuration of "model" and in definition of our component have influence on classes and methods that are available in spel expressions.

docs/api/internal/nu-designer-openapi.yaml
```diff
@@ -0,0 +1,361 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BuildInfoDto'
+              examples:
+                Example:
+                  summary: Application build info response
+                  value:
+                    name: nussknacker
+                    version: 1.12.0-SNAPSHOT
```
arkadius, line 42: maybe we should return some syntetic version so user won't think that he/she works on some old snapshot. e.g. `1.234.0` or similar

arkadius (commented):
Nice, some minor comments added

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  override def run(args: List[String]): IO[ExitCode] = for {
+    _ <- IO.delay(logger.info("Generating Nu Designer OpenAPI document ..."))
+    _ <- generateAndSave()
+    _ <- IO.delay(logger.info("DONE!"))
+  } yield ExitCode.Success
+
+  private def generateAndSave() = IO {
+    (pwd / "docs" / "api" / "internal" / "nu-designer-openapi.yaml")
```
mslabek, line 17: The docs directory is treated specially - I think all files there are used in building public docs. Do we want to integrate this into our public docs? Our pipeline also labels changes that modify these files and treats them differently. Have you considered this?

mslabek (commented):
Left a small comment on using docs directory. Other than that and arkadius' comments LGTM :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
nu-steward: nope. Didn't know about it. I'm not sure if we should treat it as external docs currently. Probably not. 
WDYT about moving it to some `internal-docs` folder instead? Because it's kind of docs.

mslabek: Yes, I think moving it to some other folder is the best solution for now. Maybe use resources folder as arkadius suggested?

nu-steward: it's not a resource. It's a doc. So, I'll go with the new folder idea

arkadius: It is docs for developers, so don't worry - we will find it no matter where you put it ;) But `internal-docs` is fine for me. Only, please use the resources mechanism to determine the path of this file in tests. You can find an example e.g. in `BaseNuRuntimeBinTestMixin:32`

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/AppApiHttpService.scala
```diff
@@ -99,7 +98,7 @@ class AppApiHttpService(config: Config,
       .serverLogicSuccess { _ =>
         Future {
           import net.ceedubs.ficus.Ficus._
-          val configuredBuildInfo = config.getAs[Map[String, String]]("globalBuildInfo").getOrElse(Map())
+          val configuredBuildInfo = config.getAs[Map[String, String]]("globalBuildInfo").flatMap(Some.apply)
```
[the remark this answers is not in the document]
coutoPL: changed

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
coutoPL: This file has two purposes:
* it's being run by the `GenerateDesignerOpenApiPlugin`
* it can be run by dev manually

I agree that the class should not be a part of the final jar. But at the same, it's not a test source file. This class is for a developer. They should take advantage of it to generate OpenAPI that can be easily reviewed in terms of backward compatibility in the review process (IMO it's easier to find it by reading the OpenAPI file than looking at Tapir's code). At the same time, the file is helpful when you want to check what API designer has (eg. for the frontend team). All of these are development-related activities. 

I think it'd be nice to have a module for development tools. Like this one. So, this is the direction I'd like to go if the current approach is not acceptable for now.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
coutoPL: because it's not a resource I think I cannot use Java's Resource. But I can use the `user.dir` system property. WDYT about this approach?

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/NuDesignerApiAvailableToExpose.scala
coutoPL: I don't know how to use `enumeratum` lib in this case. 
Yes, we want to get all endpoint definitions from a given class, but the class cannot converted to an object (we need to pass `Auth` obj) and each of the endpoints has a different returned type. All of these make `enumeratum` lib hard to be used in this particular case.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NuDesignerApiAvailableToExposeYamlSpec.scala
coutoPL: ref: https://github.com/TouK/nussknacker/pull/4709#discussion_r1342640241

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
coutoPL: I got your intention now. It's fixed

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/AppApiEndpoints.scala
arkadius: Having this schema more static IMO is not the goal. The previous approach with inlined additional properties was more idiomatic. It is pretty natural to have json witj some static fields and some other dynamic. The only thing that you have to do is to write `additionalProperties: true` in your schema. Unfortunately, I see that there is no way to modify the generated schema to make it work this way. You can only change some minor details: https://tapir.softwaremill.com/en/latest/endpoint/schemas.html#using-implicits. Here is a similar issue: https://github.com/softwaremill/tapir/issues/1699 . In this situation, it looks like this approach is a reasonable trade-off.

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/NuDesignerApiAvailableToExpose.scala
arkadius: I think that it is our near future. We want to have an external API for things like batch invocations. But let's start with this approach and see what's happened then.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
arkadius: > it's being run by the GenerateDesignerOpenApiPlugin
This plugin will be run by whom? Not a developer?

Writing tests is not a development-related activity? What about other places where we already have "some" approach like this `DefaultClassExtractionTest`? Don't you think that there is some value in having things consistent?

coutoPL: Obviously, it's. Writing and reading tests are a part of the developer's job. It's a special place for tests in a project (it's a `test` folder obviously). But there are some other things that a developer is supposed to do. One of the things is a review. Sometimes, there are some things in the project that are useful for other devs but are not production code or tests. This is the thing. 

So in this particular case, we have:
* `NuDesignerApiAvailableToExpose` - this is the object that has all the necessary logic to generate OpenAPI yaml for all endpoints defined in the `BaseEndpointDefinitions` class
* `GenerateDesignerOpenApiYamlFile` - is an object with the main method to generate the file (it's using the above one)
* `NuDesignerApiAvailableToExposeYamlSpec` - it a test if the generated OpenAPI yaml is a fresh one
* `GenerateDesignerOpenApiPlugin` - it's a plugin that runs the main method of the GenerateDesignerOpenApiYamlFile

The last one has to stay where it is now. The 3rd one is in tests - the place it is supposed to be. 
The first two, are in the "main" code. I agree it's not a proper place for these files. 
I could obviously move them to the test code but I'm not sure if this is the right approach. I proposed a new module (with dev tools).

I (as you probably know, because I used to mention that) see value in having things consistent, but I also like to improve things. IMO, this is sth we can easily improve. 

--- 
> We check there how changes in configuration of "model" and in definition of our component have influence on classes and methods that are available in spel expressions.

As I described above we don't want to only test sth in this case. We want to give developers a tool to generate static OpenAPI file. I'm not sure I see that the case you show me is the same. If I'm wrong, please explain (because I found the tests are not self-explanatory in this matter).

arkadius: >  I'm not sure I see that the case you show me is the same. If I'm wrong, please explain

It is exactly the same as with OpenAPI. We have some JSON format that describes the API that we expose to our users in SPeL. We have our own mechanism of automatic derivation of this API based on components, helpers (utils like `#COLLECTION` variable) and configuration (e.g. dev-application.conf). 

I appreciate your spirit to improve things but I was worried that we will have two similar approaches which two other concepts of solving this trade-off between a number of modules, having things separated etc. Let's extract this model and at least add a TODO into these two classes (`DefaultClassExtractionTest` and `DevClassExtractionTest`) that we want to follow this approach in this place as well.

mslabek (approved):
LGTM :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/GenerateDesignerOpenApiYamlFile.scala
coutoPL: I've moved the 1st and the 2nd class to the test sources, because I want to close this task as soon as I can. I'll create a jira for the improvement. 

Please take a look if it can be done like this, for now.

arkadius (approved):
nice

PR #4711, opened 2023-09-25
create AdditionalComponentsUIConfigProviderFactory, refactor UIProcessObjectsFactory

wrzontek:
## Describe your changes

create the option to provide additional components' UI configuration - without the need for model reload

## Checklist before merge
- [ ] Code formatting checked


designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -51,10 +51,12 @@ object UIProcessObjectsFactory {
 
     val dynamicComponentsConfig = uiProcessDefinition.allDefinitions.mapValuesNow(_.componentConfig)
 
+    val additionalComponentsUIConfig = modelDataForType.additionalComponentsUIConfigProvider.getAllForCategory(processingType)
+
     val fragmentsComponentsConfig = fragmentInputs.mapValuesNow(_.objectDefinition.componentConfig)
     //we append fixedComponentsConfig, because configuration of default components (filters, switches) etc. will not be present in dynamicComponentsConfig...
     //maybe we can put them also in uiProcessDefinition.allDefinitions?
-    val finalComponentsConfig = ComponentDefinitionPreparer.combineComponentsConfig(fragmentsComponentsConfig, fixedComponentsUiConfig, dynamicComponentsConfig)
+    val finalComponentsConfig = ComponentDefinitionPreparer.combineComponentsConfig(fragmentsComponentsConfig, additionalComponentsUIConfig, fixedComponentsUiConfig, dynamicComponentsConfig) // TODO consider the ordering
```
wrzontek, line 59: NOTE:
this goes into `UIProcessObjects` as `componentsConfig`, but I'm not sure if FE relies on that field for what it displays in the parameters' form

I count at least 4 sources of parameter information in the `UIProcessObjects` alone:
1. componentGroups -> *group* -> components -> *name* -> node -> service/ref/?-> parameters
2. processDefinition -> *group* -> *name* -> parameters
3. processDefinition -> *group* -> *name* -> componentConfig -> params
4. componentsConfig -> *name* -> params

This draft uses the 4th option, as it was most convenient in BE code.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/DefaultAdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,7 @@
+package pl.touk.nussknacker.engine.api.component
+
+class DefaultAdditionalComponentsUIConfigProvider extends AdditionalComponentsUIConfigProvider {
+
+  override def getAllForCategory(category: String): Map[String, SingleComponentConfig] = Map.empty
```
wrzontek, line 5: we could get rid of the `category` argument if the class instance is bound to one already, for example via `modelDataForType`, which is currently the only usage

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/AdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,7 @@
+package pl.touk.nussknacker.engine.api.component
+
+trait AdditionalComponentsUIConfigProvider {
+
+  def getAllForCategory(category: String): Map[String, SingleComponentConfig]
```
arkadius, line 5: We need the:
```
getAll: Map[ComponentId, SingleComponentConfigWithoutComponentId]
```
, otherwise how to deal with components deduplicated across multiple processing types?

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -28,6 +30,35 @@ object UIProcessObjectsFactory {
 
   import net.ceedubs.ficus.Ficus._
 
+  private def getComponentIdToNameMap(componentIdProvider: ComponentIdProvider,
```
arkadius, line 33: I don't like this method. It looks like you lost some information (`componentType` I guess) and you have to recover this here. I'd rather have a more suitable input structure (`ProcessDefinition`). Maybe we could add to `ProcessDefinition` a method that would take `ComponentIdProvider` and return `Map[ComponentIdWithOriginName, T]`? Also, is it ok that pass `UIProcessDefinition` and `ComponentsUiConfig` into `ComponentDefinitionPreparer.prepareComponentsGroupList`? Maybe we should have some smarter structure that uses ComponentId and things will be easier than?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/AdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,9 @@
+package pl.touk.nussknacker.engine.api.component
+
+trait AdditionalComponentsUIConfigProvider extends Serializable {
+
+  def getAllForProcessingType(processingType: String): Map[ComponentId, SingleComponentUIConfig]
+
+  def getAll: Map[ComponentId, SingleComponentUIConfig]
```
arkadius, line 7: Is it used anywhere?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/DefaultAdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,8 @@
+package pl.touk.nussknacker.engine.api.component
+
+class DefaultAdditionalComponentsUIConfigProvider extends AdditionalComponentsUIConfigProvider {
```
arkadius, line 3: Let's call it `Empty...`. I'd rather expect Map-based implementation as some Default. Or maybe we could have this Map-based implementation and have a `AdditionalComponentsUIConfigProvider.empty` method that would created this class with empty map in the input parameter?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
```diff
@@ -14,6 +14,14 @@ import pl.touk.nussknacker.engine.api.definition.{ParameterEditor, ParameterVali
   def paramConfig(name: String): ParameterConfig = params.flatMap(_.get(name)).getOrElse(ParameterConfig.empty)
 }
 
+@JsonCodec case class SingleComponentUIConfig(params: Option[Map[String, ParameterConfig]], icon: Option[String], docsUrl: Option[String], disabled: Boolean = false) {
```
arkadius, line 17: I thought that `SingleComponentConfig` is "UI". `componentGroup` is for the presentation purpose. Also, we shouldn't expose `componentGroup` in the `AdditionalComponentsUIConfigProvider` interface? I think that we need all but `componentId`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -28,6 +30,35 @@ object UIProcessObjectsFactory {
...
+                                      uiProcessDefinition: UIProcessDefinition,
+                                      isFragment: Boolean,
+                                      processingType: ProcessingType
+                                     ): Map[ComponentId, ProcessingType] = {
```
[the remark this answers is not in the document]
wrzontek: my bad, though it highlights that using "types" that are just aliases for "String" doesn't do much for type-safety, only somewhat helps readability

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
wrzontek: i don't like it either :thinking:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/AdditionalComponentsUIConfigProvider.scala
wrzontek: no

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
wrzontek: >Also, is it ok that pass UIProcessDefinition and ComponentsUiConfig into ComponentDefinitionPreparer.prepareComponentsGroupList?
I don't see how it could break anything, but it also seems unnecessary to use `UIObjectDefiniton` instead of `ObjectDefinition` in  `prepareComponentsGroupList`, i'll cook up a PR to change it

wrzontek: on second thoughts, `prepareComponentsGroupList` is an UI object, we don't want it to de-synchronize from UIProcessDefinition

still, here's the draft PR if you're interested: https://github.com/TouK/nussknacker/pull/4759

arkadius: Looks like a good idea. It is easier to think about this process or that we have a few steps that are bases on the same abstractions instead of using some intermediate structures.

wrzontek: hm, if you think so I'll undraft and assing your review

arkadius: Yep, but also rebase this PR so I could see how this code affect this change - if it already solve the problem or we'll need some more steps

wrzontek: it doesn't really change this PR at all, they can be reviewed independently

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/helpers/ProcessTestData.scala
```diff
@@ -60,6 +62,13 @@ object ProcessTestData {
...
   val optionalEndingStreamTransformer          = "optionalEndingTransformer"
 
+  class SimpleTestComponentIdProvider extends ComponentIdProvider {
+    def createComponentId(processingType: String, name: Option[String], componentType: ComponentType): ComponentId =
+      ComponentId.default(processingType, name.getOrElse(""), componentType)
+
+    def nodeToComponentId(processingType: String, node: NodeData): Option[ComponentId] = ???
+  }
+
   val processDefinition: ProcessDefinitionExtractor.ProcessDefinition[DefinitionExtractor.ObjectDefinition] =
```
arkadius, line 72: It is not used anymore, you can make it `private`

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ComponentIdProvider.scala
```diff
@@ -1,47 +1,22 @@
...
-
-}
-
-class DefaultComponentIdProvider(configs: Map[ProcessingType, ComponentsUiConfig]) extends ComponentIdProvider {
+class DefaultComponentIdProvider(configs: Map[String, ComponentsUiConfig]) extends ComponentIdProvider {
 
   override def createComponentId(
-      processingType: ProcessingType,
-      name: Option[ProcessingType],
+      processingType: String,
```
arkadius, line 18: Why do you replace this alias with String?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
```diff
@@ -21,6 +21,29 @@ import pl.touk.nussknacker.engine.api.definition.{ParameterEditor, ParameterVali
   def paramConfig(name: String): ParameterConfig = params.flatMap(_.get(name)).getOrElse(ParameterConfig.empty)
 }
 
+@JsonCodec case class SingleComponentUIConfig(
```
arkadius, line 24: It isn't serialized to json am I right? I'd rather move this class next to `AdditionalComponentsUIConfigProvider` and treat it as a part of this interface. Also let's rename it to something better, "UI" looks like it is a DTO returned to FE. Mabye `ComponentConfigWithoutId`? or `BaseComponentConfig`

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ComponentIdProvider.scala
wrzontek: when moving the the class to a different module, the typedef went out of scope, I didn't choose to duplicate/move it

i'm not a fan of `types` that are just aliases for String, in my opinion it creates more confusion than value
also there are plenty of cases in the code where processingType is just String already

moving the alias definition higher (to common-api for example) generates changes in 50+ files, I can do it in a separate PR

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
arkadius: 1. Please do not resolve reviewer's comments - I want to check of my own if this comment is to resolve
2. I also wrote about "I'd rather move this class next to AdditionalComponentsUIConfigProvider" and you didn't do that.

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -42,67 +62,119 @@ object UIProcessObjectsFactory {
...
+      defaultAsyncInterpretation = getDefaultAsyncInterpretation(modelDataForType.processConfig)
     )
   }
 
+  private def toComponentsUiConfig(
+      processDefinition: ProcessDefinitionWithComponentIds[ObjectDefinition]
+  ): ComponentsUiConfig =
+    processDefinition.allDefinitions.map { case (idWithName, value) => idWithName.name -> value.componentConfig }.toMap
+
+  private def finalizeProcessDefinition(
```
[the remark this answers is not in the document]
arkadius: The idea of `ObjectDefinition.componentConfig` was to hold a `componentConfig` defined programmatically (in user's `ProcessConfigCreator`) whereas this mentioned `Map` holds a result of combination configs defined programmatically, defined in the static config and now - using this plugin mechanism. But it is minor thing so for me we can stick with this approach 

but ... ;) I see another problem! Using the current approach it is impossible to use `AdditionalComponentsUIConfigProvider` mechanism to provide configuration for fragments and "base" components and it is not acceptable :( I think that we don't need to prepare finalizedProcessDefinition and after that extract config from them but we should think about the general concept, how we want to merge this additional components config provided by `AdditionalComponentsUIConfigProvider` into the existing configuration based on mentioned two sources of components config.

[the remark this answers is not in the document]
wrzontek: you're right that it doesn't allow for configuration of basic components and fragments - I thought this would be ok, especially with fragments getting their own (with same or very similar capabilities) way of editing parameter configs, but perhaps it's an issue

i'll think about a solution

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/ComponentIdProvider.scala
arkadius: I see now, I didn't realize that you moved `ComponentIdProvider` into the interpreter. It's fine, I have the same feeling about type aliases. I would like to replace this type alias with just a value class like we do e.g. with `ProcessName` and similar - we can do this in the separate PR.

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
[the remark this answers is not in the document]
arkadius: The merging logic would be a little bit tricky.  I think that for now, we can leave a comment next to the `AdditionalComponentsUIConfigProvider` that it can provide a configuration only for a standard component. In the future IMO we should return to FE components keyed by `ComponentId` and move the name (label) into `SingleComponentConfig`. You can leave a TODO as well.

arkadius (approved):
LGTM

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/AdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,37 @@
+package pl.touk.nussknacker.engine.api.component
+
+import pl.touk.nussknacker.engine.api.component.AdditionalComponentsUIConfigProvider.SingleComponentConfigWithoutId
+
+trait AdditionalComponentsUIConfigProvider extends Serializable {
```
arkadius, line 5: Please add an changelog entry about new API and add the API to the list here: https://nussknacker.io/documentation/docs/developers_guide/Basics/#other-spis-for-nussknacker-customization-documentation-will-follow-soon
Please also add a short documentation above disc class describing the purpose of this interface

docs/Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 1.12.0 (Not released yet)
 -------------------------
+* [#4711](https://github.com/TouK/nussknacker/pull/4711) Added AdditionalComponentsUIConfigProvider API that allows changing components' configs without model reload
```
arkadius, line 5: please move this to the new 1.13.0 section, we already cutted off the release branch

arkadius (approved):

arkadius (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/AdditionalComponentsUIConfigProvider.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def getAllForProcessingType(processingType: String): Map[ComponentId, SingleComponentConfigWithoutId]
+
+}
+
+object AdditionalComponentsUIConfigProvider {
+  val empty = new DefaultAdditionalComponentsUIConfigProvider(Map.empty)
+
+  case class SingleComponentConfigWithoutId(
+      params: Option[Map[String, ParameterConfig]],
```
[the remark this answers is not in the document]
wrzontek: there isn't a difference, you're right

it was copied from SingleComponentConfig, where the Option[Map...] makes sense, as it makes the JsonCode simpler (you can just not provide `params` if you don't want to change them)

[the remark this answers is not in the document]
arkadius: Let's change it in the further PRs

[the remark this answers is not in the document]
wrzontek: yeah, i've already changed it in https://github.com/TouK/nussknacker/pull/4862

PR #4738, opened 2023-09-29
[NU-1093] prepare for auto-formatting of Scala code

coutoPL:
## Describe your changes

- added: `formatStagedScalaFiles` sbt's task - formats staged scala-related files 
- added: `configureGitHooksPath` sbt's auto plugin to configure git's hook path
- improving `.husky/pre-commit` script (it's going to call the `formatStagedScalaFiles` task)
- adding scalafmt and scalafix custom configurations

## Checklist before merge
- [x] Code formatting checked


designer/formatStagedFrontendFiles.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
```
piotrp, line 1: This script doesn't change current directory and uses paths relative to directory root, so it will fail when called as `./formatStagedFrontendFiles.sh`. Is this ok?

project/FormatStagedScalaFilesPlugin.scala
```diff
@@ -0,0 +1,76 @@
...
+      .call()
+      .out
+      .lines()
+      .filter(f => f.endsWith(".scala") || f.endsWith(".sbt"))
+      .toList
+  }
+
+  private def callFormatFiles(files: List[String]) = {
+    val userDir       = getUserDir()
+    val fullPathFiles = files.map(f => s"${(userDir / f).path}")
```
piotrp, line 55: Do we need absolute paths here? If so, can we get root path from `baseDirectory.value`, or we need to use Java's system property?

project/FormatStagedScalaFilesPlugin.scala
```diff
@@ -0,0 +1,76 @@
...
+    for {
+      _ <- Step.task {
+        streams.map(_.log.info("Formatting backend files ..."))
+      }
+      _ <- Step.task {
+        (Compile / ScalafmtPlugin.autoImport.scalafmtOnly).toTask(s" ${fullPathFiles.mkString(" ")}")
```
piotrp, line 61: This will fail if project is checked out eg. at `/home/My Projects` (or any directory containing space) - paths should be quoted and escaped.

project/NuGitHooksPathConfiguratorPlugin.scala
```diff
@@ -0,0 +1,44 @@
...
+import sbt.Keys.{aggregate, onLoad, streams}
+import sbt.internal.util.ManagedLogger
+import sbt.{Def, Global, Level, Setting, State, taskKey}
+
+import scala.util.Using
+
+object NuGitHooksPathConfiguratorPlugin extends sbt.AutoPlugin {
+  override def trigger = allRequirements
+
+  lazy val configureGitHooksPath                       = taskKey[Unit]("Configure NU Git Hooks path")
```
piotrp, line 12: "NU" - shouldn't it be  "Nu"?

project/FormatStagedScalaFilesPlugin.scala
piotrp: ... or is it sbt-scalafmt limitation? Maybe let's throw an error if we see that `userDir` contains space character?

.husky/pre-commit
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
 
-npm run commit --if-present --prefix designer/client
-npm run commit --if-present --prefix designer/submodules
+if [ "$(git diff --cached --name-only --diff-filter=ACM | grep -E "^designer/client|designer/submodules" | wc -l)" -gt 0 ]; then
```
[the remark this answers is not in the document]
coutoPL: thanks for the hint

project/FormatStagedScalaFilesPlugin.scala
coutoPL: I've added the plugin explicitly to the root module of the project and now `baseDirectory` works as expected

piotrp: Then shouldn't you also remove absolute path resolution? `scalafmtOnly` does it for you

project/FormatStagedScalaFilesPlugin.scala
coutoPL: I've tested and it works now.

project/FormatStagedScalaFilesPlugin.scala
coutoPL: yes, yes. The review fixes were not ready yet. They are now.

piotrp (approved):

PR #4740, opened 2023-10-02
[NU-1100] fix issue with dragging nodes from a toolbar in touch display…

Dzuming:
## Describe your changes
- fix the issue with dragging nodes from a toolbar in touch displays and mouse events
- Fix issue with node connections on hybrid devices
- Fix the issue with node selecting on hybrid devices
- Fix issue with Pan on hybrid devices
## Checklist before merge
- [x] Code formatting checked


designer/client/src/components/graph/graphTheme.styl
```diff
@@ -5,14 +5,16 @@ linkColor = white
...
 
-.joint-theme-default.touch
-    .joint-link
-        .link-tool
-        &:hover .connection-wrap
-            display none
-        .marker-arrowhead
-            opacity 0
-            transform scale(.3)
+@media (hover: none)
```
Dzuming, line 8: We need to detect touch events on the CSS side, Passing custom className, in this case, broke the hybrid device node connection, and also events like panning, work "strange". That's because this custom class was added based on device type, not based on the current event

designer/client/src/components/graph/utils/graphUtils.ts
```diff
@@ -121,3 +123,11 @@ function addEdge(process: Process, fromId: NodeId, toId: NodeId): Process {
         edges: process.edges.concat({ from: fromId, to: toId }),
     };
 }
+
+export const handleGraphEvent =
```
JulianWielga, line 127: it's only for cell view events.  i'd like to see in name when it's in separate module

JulianWielga (approved):

designer/client/src/components/graph/utils/graphUtils.ts
Dzuming: I made it more generic, so it will be possible to pass any View based type

PR #4744, opened 2023-10-02
 [NU-1096] OIDC: Fix for support for M2M usage of JWT

arkadius:
## Checklist before merge
- [ ] Code formatting checked


security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CachingOAuth2Service.scala
```diff
@@ -60,5 +60,15 @@ class CachingOAuth2Service[
     userInfo.map { case (userInfo, expiration) => (userInfo, Some(expiration)) }
   }
 
+  override def introspectAccessToken(accessToken: String): Future[IntrospectedAccessTokenData] =
+    throw new IllegalAccessError(
```
coutoPL, line 64: For me, it's sth suspicious here. It's a decorator and it doesn't allow us to call some public methods on it. Why cannot we just delegate calls to the underlying object (the `delegate`)?

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/JwtOAuth2Service.scala
```diff
@@ -47,16 +46,15 @@ class JwtOAuth2Service[
...
-      Future(accessToken)
-        .flatMap(accessToken => introspectJwtToken[AccessTokenClaims](accessToken))
-        .flatMap(claims =>
-          if (verifyAccessTokenAudience(claims)) {
-            Future.successful(claims.expirationTime)
-          } else
-            Future.failed(OAuth2CompoundException(one(OAuth2AccessTokenRejection("Invalid audience claim"))))
-        )
+      introspectJwtToken[AccessTokenClaims](accessToken).map { claims =>
+        if (verifyAccessTokenAudience(claims)) {
```
coutoPL, line 52: can we move it to a private method? I mean, instead of `verifyAccessTokenAudience` there could be a method that checks the configuration and creates introspected data or throws the exception. The exception message could have a richer description of what is wrong (the human-readable version of the condition). I wanted to give a note here only about the message in the exception, but the condition is "too far" from the message to be sure that in the future when someone alters it, the message and condition won't fall apart.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
```diff
@@ -25,7 +26,24 @@ trait OAuth2Service[+UserInfoData, +AuthorizationData <: OAuth2AuthorizationData
...
+}
+
+final case class IntrospectedAccessTokenData(
+    subject: Option[String],
+    expirationTime: Option[Instant],
+    roles: Set[String]
+)
+
+object IntrospectedAccessTokenData {
+  val empty: IntrospectedAccessTokenData = IntrospectedAccessTokenData(None, None, Set.empty)
```
coutoPL, line 46: I don't know if I get why we need to do it like this instead using `Option[IntrospectedAccessTokenData]` or `sealed trait` hierarchy like this:

```scala
sealed trait IntrospectedAccessTokenData
object IntrospectedAccessTokenData {
   final case class Instance(subject ...) extends IntrospectedAccessTokenData
   case object Empty extends IntrospectedAccessTokenData
}
```
```

File: security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OpenIdConnectProfile.scala [hunk 4]
```diff
@@ -65,30 +66,62 @@ object OpenIdConnectUserInfo extends EitherCodecs with EpochSecondsCodecs {
...
+        .getOrElse(throw new IllegalStateException("Missing user identity"))
+      val userRoles = profile.roles ++ getUserRoles(userIdentity, configuration)
 
-    val username =
-      usernameBasedOnUsersConfiguration(userIdentity, configuration)
-        .orElse(usernameBasedOnConfigurationClaim)
-        .orElse(profile.preferredUsername) // backward compatibility
-        .orElse(profile.nickname)          // backward compatibility
-        .getOrElse(userIdentity)
+      val usernameBasedOnConfigurationClaim = configuration.usernameClaim match {
```
coutoPL, line 103: `lazy val`?

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/UserMappingOAuth2Service.scala
```diff
@@ -4,25 +4,41 @@ import io.circe.Decoder
 import pl.touk.nussknacker.ui.security.api.AuthenticatedUser
 import sttp.client3.SttpBackend
 
-import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
 
 class UserMappingOAuth2Service[UserInfoData: Decoder, AuthorizationData <: OAuth2AuthorizationData: Decoder](
     delegate: OAuth2Service[UserInfoData, AuthorizationData],
-    loggedUserFunction: UserInfoData => AuthenticatedUser
+    loggedUserFunction: LoggedUserFunctionParameters[UserInfoData] => Future[AuthenticatedUser]
```
coutoPL, line 11: WDYT about changing the param name to `authenticatedUserFrom`?

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/UserMappingOAuth2Service.scala
```diff
@@ -4,25 +4,41 @@ import io.circe.Decoder
...
-    }
-
-  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(AuthenticatedUser, Option[Instant])] =
-    delegate.checkAuthorizationAndObtainUserinfo(accessToken).map { case (userInfo, expiration) =>
-      (loggedUserFunction(userInfo), expiration)
-    }
+  ): Future[(AuthorizationData, AuthenticatedUser)] = {
+    for {
+      (authorization, userInfo) <- delegate.obtainAuthorizationAndUserInfo(authorizationCode, redirectUri)
+      loggedUser <- loggedUserFunction(
```
coutoPL, line 21: `authenticatedUser <- ...` ?

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/UserMappingOAuth2Service.scala
```diff
@@ -4,25 +4,41 @@ import io.circe.Decoder
...
+      accessTokenData: IntrospectedAccessTokenData
+  ): Future[AuthenticatedUser] = {
+    loggedUserFunction(
+      LoggedUserFunctionParameters(accessTokenData, () => delegate.obtainUserInfo(accessToken, accessTokenData))
+    )
+  }
+
 }
+
+final case class LoggedUserFunctionParameters[UserInfoData](
```
coutoPL, line 41: 1. we have a function in it, so maybe just `final class` (with adding `val`s obviously)?
2. BTW, the class doesn't look "natural". Maybe we don't need it? ;) Maybe we won't need it if we have a trait instead of `loggedUserFunction: LoggedUserFunctionParameters[UserInfoData] => Future[AuthenticatedUser]`? WDYT?

coutoPL (commented):
according to the failed test. It seems that the regex
```
"^\\\\d{4}-\\\\d{2}-\\\\d{2}T\\\\d{2}:\\\\d{2}:\\\\d{2}\\\\.\\\\d{6}$\"
```
was not matched (the current value was `"2023-10-03T22:54:20.577"` 

Currently, it won't be deterministic because of this:
```
The output will be one of the following ISO-8601 formats:
uuuu-MM-dd'T'HH:mm
uuuu-MM-dd'T'HH:mm:ss
uuuu-MM-dd'T'HH:mm:ss.SSS
uuuu-MM-dd'T'HH:mm:ss.SSSSSS
uuuu-MM-dd'T'HH:mm:ss.SSSSSSSSS
```

I will provide a fix. Don't bother. Just rerun.

coutoPL:
> I will provide a fix. Don't bother. Just rerun. 

ref: https://github.com/TouK/nussknacker/pull/4790

docs/Changelog.md
```diff
@@ -21,6 +21,9 @@
 * [#4685](https://github.com/TouK/nussknacker/pull/4685) App API OpenAPI-based documentation (e.g. `https://demo.nussknacker.io/api/docs`)
 * [#4707](https://github.com/TouK/nussknacker/pull/4707) Support for `overrideFrontendAuthenticationStrategy` configuration parameter in OIDC security model - works the same as in OAuth2 case.
 * [#4739](https://github.com/TouK/nussknacker/pull/4739) Add configuration parameter for sending additional headers to InfluxDB (`countsSettings.additionalHeaders`)
+* [#4744](https://github.com/TouK/nussknacker/pull/4744) Fix for OIDC: support for simultaneously Machine-2-Machine and Basic flow handling: 
+  * Skip user profile call based on the fact that access token is JWT with subject and user has username configured. 
+  * `accessTokenIsJwt` Oidc configuration introduced in [#4283](https://github.com/TouK/nussknacker/pull/4283) is removed: `audience` configuration specifies that accessToken is a JWT as it was before this change
```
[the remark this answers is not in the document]
arkadius: fixed

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
```diff
@@ -25,7 +26,24 @@ trait OAuth2Service[+UserInfoData, +AuthorizationData <: OAuth2AuthorizationData
...
       redirectUri: String
   ): Future[(AuthorizationData, UserInfoData)]
-  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(UserInfoData, Option[Instant])]
+  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(UserInfoData, Option[Instant])] = {
+    for {
+      accessTokenData <- introspectAccessToken(accessToken)
+      userInfo        <- obtainUserInfo(accessToken, accessTokenData)
+    } yield (userInfo, accessTokenData.expirationTime)
+  }
+  def introspectAccessToken(accessToken: String): Future[IntrospectedAccessTokenData]
```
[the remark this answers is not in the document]
arkadius: ok

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
```diff
@@ -25,7 +26,24 @@ trait OAuth2Service[+UserInfoData, +AuthorizationData <: OAuth2AuthorizationData
       authorizationCode: String,
       redirectUri: String
   ): Future[(AuthorizationData, UserInfoData)]
-  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(UserInfoData, Option[Instant])]
+  def checkAuthorizationAndObtainUserinfo(accessToken: String): Future[(UserInfoData, Option[Instant])] = {
```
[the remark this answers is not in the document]
arkadius: I added new lines and I'll add this fmt configuration tunning in a new PR

security/src/test/scala/pl/touk/nussknacker/ui/security/oauth2/ExampleOAuth2ServiceFactory.scala
```diff
@@ -26,17 +25,21 @@ class ExampleOAuth2Service(clientApi: OAuth2ClientApi[TestProfileResponse, TestA
...
+  override def introspectAccessToken(accessToken: String): Future[IntrospectedAccessTokenData] =
+    Future.successful(IntrospectedAccessTokenData.empty)
+
+  override def obtainUserInfo(
+      accessToken: String,
+      accessTokenData: IntrospectedAccessTokenData
+  ): Future[AuthenticatedUser] =
+    clientApi.profileRequest(accessToken).map { prf: TestProfileResponse =>
+      AuthenticatedUser(
+        prf.uid,
```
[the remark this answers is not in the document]
arkadius: ok

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CachingOAuth2Service.scala
arkadius: I did it because someone can think that one of these methods cache access token as well and this class is designed to be used only via `checkAuthorizationAndObtainUserinfo`. But I changed it as you proposed. Let me know if you change your mind after reading this comment.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/UserMappingOAuth2Service.scala
arkadius: Yeah I saw many readability problems around this code. One of the hardest for me was the inheritance structure: `OidcService <: GenericOidcService <: JwtOAuth2Service <: BaseOAuth2Service <: OAuth2Service`! Or methods like `obtainAuthorizationAndUserInfo` vs `checkAuthorizationAndObtainUserinfo`  I think that I could spend at least two days more making it more understandable but I didn't want to extend the scope of this task. But let's do some cleanups as you proposed - changed it

arkadius: I also changed all occurrences of `obtainUserInfo`. The need at the root is that we want to authenticate the user - obtaining the user info is just one step to do that. BTW this step is not always needed, sometimes all information is inside the access token or in id token so we don't obtain anything - only extract it.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
arkadius: This is because our hierarchy structure of service classes is prepared for dozens of methods of authentication. For example we can handle access token that is not a JWT token (which is allowed by OAuth2 specification) and during access token introspection we can extract only some of the data (e.g. only `expirationTime`). And even if the access token is JWT, according to the specification, every claim is optional so we can have `subject` but no `expiration`. You might ask if we already use all these possibilities. I can't say it for 100% per cent. I know that we at least use authentication with OIDC (keycloak, Auth0), with GH which has some custom username claims and no option to define role claims and some custom implementation in at least one deployment of our product.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/UserMappingOAuth2Service.scala
arkadius: :+1:  In fact, we already had this trait - it was `OAuth2Profile`. I renamed it to `AuthenticationStrategy` and extracted `OAuth2Configuration` parameter to constructors which allowed me to remove this class.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/OAuth2ServiceFactory.scala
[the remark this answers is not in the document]
arkadius: New PR: #4793

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/CachingOAuth2Service.scala
coutoPL: IMO it's OK now. We could think about changing the name of the class to suggest what is cached, but at the moment, the two methods have a `private [oauth2]` scope, so I don't think we need to do it.

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/AuthenticationStrategy.scala
```diff
@@ -0,0 +1,31 @@
...
+  def authenticateUser(
+      accessTokenData: IntrospectedAccessTokenData,
+      getProfile: => Future[ProfileResponse]
+  )(implicit ec: ExecutionContext): Future[AuthenticatedUser]
+
+}
+
+object AuthenticationStrategy {
+
+  def getUserRoles(
```
coutoPL, line 18: these two functions look like great candidates for the extension methods of the `OAuth2Configuration` class.

security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationResources.scala
```diff
@@ -12,10 +12,9 @@ class OidcAuthenticationResources(name: String, realm: String, configuration: Oi
       name = name,
       realm = realm,
       service = new CachingOAuth2Service(
-        new UserMappingOAuth2Service(
+        new UserMappingOAuth2Service[OpenIdConnectUserInfo, DefaultOidcAuthorizationData](
           new OidcService(configuration),
-          (userInfo: OpenIdConnectUserInfo) =>
-            OpenIdConnectProfile.getAuthenticatedUser(userInfo, configuration.oAuth2Configuration)
+          new OpenIdConnectProfileAuthentication(configuration.oAuth2Configuration)
```
coutoPL, line 17: shouldn't we use the `Oidc` abbreviation in the case of this class too? To be consistent. So, maybe `OidcProfileAuthentication`?

coutoPL (approved):
Two more small comments. But in general LGTM!

security/src/main/scala/pl/touk/nussknacker/ui/security/oauth2/AuthenticationStrategy.scala
arkadius: I converted the first one to a normal method in `OAuth2Configuration` and inlined the usage of the second one because it was used only in one place.

security/src/main/scala/pl/touk/nussknacker/ui/security/oidc/OidcAuthenticationResources.scala
arkadius: Yeah, I don't know what concept someone has about having two prefixes for oidc. I used `Oidc` abbreviation in all places where `OpenIdConnect` was and also moved all `oidc*` classes to `oidc` package

coutoPL (approved):

PR #4745, opened 2023-10-02
[NU-983] Add BE node and scenario name validation

mslabek:
## Describe your changes
Moves from FE to BE:
- node name validation to backend node validation endpoint,
- scenario/fragment name validation to backend properties validation endpoint.

Adds new validations - blank, leading spaces, trailing spaces to node id and scenario id.

## Checklist before merge
- [x] Code formatting checked


mslabek:
One problem related to this change is that there is still **unique node id** validation left on the FE. If we remove this validation from FE we will still have this validation on apply (when saving node) - that error will prevent saving the node and will be displayed as a red toast message.

We can replicate this feature on the BE by extending the request by passing all node ids. IMO the problem with this approach is that we add more context to a request which doesn't do whole validation anyways.

designer/client/src/store/nodeValidationMiddleware.ts
```diff
@@ -4,13 +4,13 @@ import { RootState } from "../reducers";
...
 import HttpService from "../http/HttpService";
-import { getProcessToDisplay } from "../reducers/selectors/graph";
+import { getProcessToDisplayWithUnsavedName } from "../reducers/selectors/graph";
 
 type ActionType = Action["type"];
 
 const debouncedValidate = debounce(
     (dispatch: ThunkDispatch, getState: () => RootState) =>
-        HttpService.validateProcess(getProcessToDisplay(getState())).then(({ data }) =>
+        HttpService.validateProcess(getProcessToDisplayWithUnsavedName(getState())).then(({ data }) =>
```
mslabek, line 13: Its not pretty but its necessary for consistency of limited context validation of scenario properties (where unsaved name is passed differently) with full context validation

mslabek:
Below is a simpified (and probably not fully correct) view of components used in validation of different requests. This is to provide some explanation on weird logic in some places:
1. Deduplication of errors - Validation of  DisplayableProcess and CanonicalProcess are done differently - if we want to have id validation before resolving into CanonicalProcess (for example with loose nodes) then we need to deduplicate errors in ProcessValidation
2. Multiple validation flows with different error results - There are multiple flows used in validation, sometimes this may result in non-complete validation. Thats why there is a need for functions that validate not only whole CanonicalProcess but nodeIds and scenario/fragment id separately
3. Possible inconsistencies - This is not visible on the graph but from a glance it seems that process import does not have validation on level of DisplayableProcess which allows for e.g. importing of scenarios with duplicated nodes which has display errors. Fortunately, this is still impossible to save due to the full context validation on save.

validation drawio (1)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -29,7 +29,7 @@ object ProcessCompilationError {
   val ValidatedNelApplicative: Applicative[ValidatedNelCompilationError] =
     Applicative[ValidatedNelCompilationError]
 
-  trait InASingleNode {
+  sealed trait InASingleNode {
```
mslabek, line 32: Is there a reason why the traits are not sealed and case classes not final here?

arkadius: Lately, we decided to make final, all case classes so you can add a final modifier always when you see that is missing. I thought that nobody was so crazy about extending case class so we could rely on the convention, but our new team members said that they saw cases of case class extension :) Regarding sealed, I'm a fan of the rule that you add this modifier when you want to do a pattern matching on all subclasses. I don't like the approach "let's add it because for now, it compiles". I think it is an unnecessary limitation. In the beginning, we had `StateStatus` sealed and it caused pain in the ass when we wanted to have some new status in our extension. We can talk more about props and cons if you don't feel convinced.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
```diff
@@ -164,18 +167,31 @@ class ProcessValidation(
     ValidationResult.warnings(disabledNodesWarnings)
   }
 
-  private def validateIds(displayable: DisplayableProcess): ValidationResult = {
-    val invalidCharsRegexp = "[\"'\\.]".r
+  private def validateScenarioId(displayable: DisplayableProcess): ValidationResult = {
+    val fakeCanonicalProcess =
```
arkadius, line 171: Why do you need to create a fake scenario here? Why don't you pass an id instead?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
arkadius: I see now that we already use these classes in pattern matching so it is fine to add sealed :+1:

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
```diff
@@ -42,27 +37,82 @@ object PrettyValidationErrors {
...
-        node(
-          "Nodes cannot have empty id",
-          "Nodes cannot have empty id",
-          errorType = NodeValidationErrorType.RenderNotAllowed
-        )
+      case propertiesError: ScenarioPropertiesError =>
+        propertiesError match {
+          case EmptyScenarioId(isFragment) =>
+            node(
+              s"${if (isFragment) "Fragment" else "Scenario"} name is mandatory and cannot be empty",
```
arkadius, line 51: Let's extract a method for this if-else

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
```diff
@@ -42,27 +37,82 @@ object PrettyValidationErrors {
...
+            node(
+              "Node name cannot have leading spaces",
+              "Leading spaces in node name",
+              fieldName = Some("$id")
+            )
+          case TrailingSpacesNodeId(_) =>
+            node(
+              "Node name cannot have trailing spaces",
+              "Trailing spaces in node name",
+              fieldName = Some("$id")
```
arkadius, line 113: Let's keep these special field names in some one place. We already do a similar thing with expressions in `NodeExpressionId`

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
```diff
@@ -0,0 +1,92 @@
...
+
+object IdValidator {
+
+  import cats.implicits._
+
+  private val nodeIdIllegalCharacters = Set('.', '"', '\'')
+
+  def validate(process: CanonicalProcess): ValidatedNel[ProcessCompilationError, Unit] = {
+    val scenarioIdValidationResult =
+      validateIdIsNotEmpty(process).andThen { _ =>
```
arkadius, line 18: You can accumulate all these errors using `.product` or `.mapN`

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
```diff
@@ -0,0 +1,92 @@
...
+        validateIdIsNotBlank(process).andThen { _ =>
+          (validateIdHasNoLeadingSpaces(process), validateIdHasNoTrailingSpaces(process)).mapN((_, _) => ())
+        }
+      }
+
+    val nodesIdValidationResult = process.nodes
+      .map(node => validateNodeId(node.data))
+      .foldLeft(Validated.validNel[ProcessCompilationError, Unit](())) { (acc, validation) =>
```
arkadius, line 26: I think that you can use shorter `sequence` instead.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
```diff
@@ -0,0 +1,92 @@
...
+      predicate: T => Boolean,
+      errorProducer: T => ProcessCompilationError
+  )(implicit process: T): ValidatedNel[ProcessCompilationError, T] = {
+    if (predicate(process))
+      invalid(NonEmptyList.one(errorProducer(process)))
+    else
+      valid(process)
+  }
+
+  private def validateIdIsNotEmpty(implicit process: CanonicalProcess) =
```
arkadius, line 55: Is there some reason why you do not just pass id to to these methods instead of extracting it in each method? Regarding fragment/scenario distinguishing in errors, we can pass some implicit parameter for that purpose.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
```diff
@@ -103,8 +106,8 @@ class ProcessValidation(
   }
 
   def uiValidation(displayable: DisplayableProcess): ValidationResult = {
-    validateIds(displayable)
-      .add(validateEmptyId(displayable))
+    validateScenarioId(displayable)
```
arkadius, line 109: Why do you do these validation here if you also do them on canonicalprocess (in ProcessCompiler)?

mslabek: Previously, FE validated empty node/scenario id independently so if you had errors that prevented converting DisplayableProcess to CanonicalProcess, the empty id errors still appeared. Now, if we do the validation only on the CanonicalProcess level, we lose this validation when we cant canonize the graph.

This generates some problems with the need to deduplicate these errors, but it's the only solution I've found for this validation to work as previously. This may not be worth because the gain is minimal for the complexity.

arkadius: Uh I see. I have no strong opinion on that. I see two options:
1. We can add a comment why it is duplicated
2. We can remove this step, lost this small benefit and revert it if we find out that it was crucial for someone

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
mslabek: Done, I've changed it to pass id and isFragment.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
mslabek: Done

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
mslabek: Good point, I think the best place for it would be in `scenario-api` in `CanonicalProcess` and `node` accordingly, added it there.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
mslabek: I'm using andThen to have more fine-grained control over the errors so that on empty id we dont get IsBlank error and when we have a blank value we don't get LeadingSpaces and TrailingSpaces errors. The result is that user gets more precise error message and FE in most cases gets only 1 error to display.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
mslabek: I've added a comment at https://github.com/TouK/nussknacker/pull/4745/files#diff-d4b08503fe91601c6e3b096ed946441b1a93faa59a8dc5467b6e64b2cbf47991R86

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
mslabek: No, there's no reason besides differentiating node/process and fragment/scenario errors. Now I've created a private class for scenario id validation `ScenarioIdValidationData(scenarioId: String, isFragment: Boolean)` not to pass entire CanonicalProcess. 

Or did you mean that the code would be cleaner with single methods per validation type for both scenario and node like `validateIdIsNotEmpty(implicit nodeId: String, isFragment: Option[Boolean])` and there we could differentiate between node/fragment/scenario based on the 3 isFragment states?

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
arkadius: Ok, it makes sense.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
arkadius: IMO it would be nice to have general purpose id validation mechanism to avoid duplication of this case of `validateIdIsNotEmpty andThen validateIdIsNotBlank ...`. It should be double if we figure out how to handle this error creation strategies depended on context. Maybe we should have generic errors e.g. `EmptyId` / `BlankId` and each of them should have something like `validatedIdName` and then map errors to '$validatedIdName is mandatory and can't be empty`. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ProcessValidation.scala
arkadius: It is nice, but I would also add a comment before `uiValidation` method that we duplicate some validation invocation because `uiValidation` is the first step that stops further validations and in case when we have some error we don't want to loose these validation errors.

mslabek: Good point, done.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
mslabek: Ok, I see now. I've changed it to a consistent validation strategy with mapping to specific errors internally in the IdValidator.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
```diff
@@ -21,80 +20,75 @@ object IdValidator {
...
-  private def applySingleErrorValidation[T](
-      isValid: Boolean,
-      error: ProcessCompilationError
-  )(implicit validated: T): ValidatedNel[ProcessCompilationError, T] = {
-    if (isValid)
-      invalid(NonEmptyList.one(error))
-    else
-      valid(validated)
-  }
+  private final case class EmptyIdError()                       extends IdValidationError
```
arkadius, line 88: Please use case object

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/IdValidator.scala
arkadius: It reduced the part of code that was the most important for me, so it is almost perfect. IMO we could go one step further and remove even these error mapping part. We could achieve that by like I wrote extracting `validatedIdName` (which could be e.g. `Fragment name` inside `EmptyIdError` etc. and use it in a message. But this code duplication isn't crucial so I'm not forcing you to do that - your choice.

designer/client/src/components/tips/error/NodeErrorLink.tsx
```diff
@@ -37,7 +37,8 @@ export const NodeErrorLink = (props: { onClick: MouseEventHandler<HTMLAnchorElem
         </span>
     ) : (
         <Link className={styles} to={`?nodeId=${nodeId}`} onClick={onClick}>
-            {nodeId}
+            {/* blank values don't render as links so this is a workaround */}
+            {nodeId.trim() === "" ? "blank" : nodeId}
```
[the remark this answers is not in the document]
Dzuming: What happens when I click on this blank link, is it an open node edit view?

designer/client/src/store/nodeValidationMiddleware.ts
Dzuming: Okay, so the previous behavior was, that name edit didn't update the id property during validation and now, we make a switch before the validation.  I'm wondering if this name is meaningful, because, now, it implies that, someone needs to have a knowledge, that name, is in fact id.

designer/client/src/components/tips/error/NodeErrorLink.tsx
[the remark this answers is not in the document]
mslabek: Yes, the router link component still gets the same `to` prop - this just changes what value the wrapped <a> tag has.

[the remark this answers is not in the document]
mslabek: The routing somehow works for blank nodes, the `?nodeId= ` prop  still contains the blank string so if you have 2 nodes with ids: " " and "   " it will still route correctly. The "blank" string is just for the link to be rendered  :smile: When they are repeated they just display the same value like this:
image Other solution would be to put quotaton marks around the whitespaces.

designer/client/src/store/nodeValidationMiddleware.ts
mslabek: The whole name/id is confusing - these concepts are mixed and in the FE app we mostly use "name" to refer to it (for example see graph selectors). In the BE api its similar. The main thing I'm trying to convey is that we are passing an object using some temporary unsaved fields. Would `getProcessToDisplayWithUnsavedData` be better?

designer/client/src/components/tips/error/NodeErrorLink.tsx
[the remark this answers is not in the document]
Dzuming: If it works, let's leave it ;).

docs/MigrationGuide.md
```diff
@@ -19,6 +19,12 @@ To see the biggest differences please consult the [changelog](Changelog.md).
     type aliases are not available anymore - you should probably use `ScenarioWithDetails` in these places
   * `pl.touk.nussknacker.restmodel.processdetails.ProcessVersion` was moved into `pl.touk.nussknacker.engine.api.process.ScenarioVersion`
   * `pl.touk.nussknacker.restmodel.processdetails.StateActionsTypes` was moved into `ProcessActionType.StateActionsTypes`
+* [#4959](https://github.com/TouK/nussknacker/pull/4959) `listener-api` module become decoupled from `restmodel` module.
```
arkadius, line 22: Looks like some glitch after rebase?

arkadius (approved):
BE side looks good

designer/client/src/store/nodeValidationMiddleware.ts
Dzuming: I want to just point out, that we switched the ID of the process to the unsaved name, and the user won't know about it without opening this function, but on the other hand, maybe it's just an implementation detail. 

IMO `getProcessToDisplayWithUnsavedName` sounds a little better than `getProcessToDisplayWithUnsavedData`,  so let's leave it as it is :).

Dzuming (approved):

PR #4769, opened 2023-10-03
Fix empty validation request on undo history clear

mslabek:
## Describe your changes
Clear history action triggered process validation, which made empty request which couldn't be decoded causing this error:
image
This change only adds this action to ignored list in the middleware.

## Checklist before merge
- [x] Code formatting checked

### DIFF
--- a/designer/client/src/store/nodeValidationMiddleware.ts
+++ b/designer/client/src/store/nodeValidationMiddleware.ts
@@ -20,7 +20,7 @@ export function nodeValidationMiddleware(
     validatedActions: ActionType[] = [],
     ignoredActions: ActionType[] = [],
 ): Middleware<void, RootState, ThunkDispatch> {
-    const ignore = ["VALIDATION_RESULT", ...ignoredActions];
+    const ignore = ["VALIDATION_RESULT", UndoActionTypes.CLEAR_HISTORY, ...ignoredActions];
     const validate = [...Object.values(UndoActionTypes), ...validatedActions];
     const shouldValidate = (action: ActionType) => !ignore.includes(action) && validate.includes(action);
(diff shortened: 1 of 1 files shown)


designer/client/src/store/nodeValidationMiddleware.ts
```diff
@@ -20,7 +20,7 @@ export function nodeValidationMiddleware(
     validatedActions: ActionType[] = [],
     ignoredActions: ActionType[] = [],
 ): Middleware<void, RootState, ThunkDispatch> {
-    const ignore = ["VALIDATION_RESULT", ...ignoredActions];
+    const ignore = ["VALIDATION_RESULT", UndoActionTypes.CLEAR_HISTORY, ...ignoredActions];
```
Dzuming, line 23: When this `UndoActionTypes.CLEAR_HISTORY` action is fired?

mslabek: From what I see, in 2 situations:
1. After saving - https://github.com/TouK/nussknacker/blob/d1665cfa2968adf59e7321791b692202f376e53c/designer/client/src/components/modals/SaveProcessDialog.tsx#L33
2. At unmounting of scenario visualization - https://github.com/TouK/nussknacker/blob/a0063b1633593fcab9afa31ca16b5b6de48ffdb9/designer/client/src/containers/Visualization.tsx#L41

When saving, the validation errors are passed from BE in the PUT request so re-validating doesn't seem to get different results. When testing locally it seems to work the same.

Dzuming: Ok, thanks for the analysis!

PR #4805, opened 2023-10-05
Formatting fixes

piotrp:
## Describe your changes
1. Allows pre-commit hook to work when `SBT_OPTS` environment variable is set (e.g. to limit sbt's memory usage).
2. Adds basic .editorconfig, to ensure that files have trailing newlines (will Idea pick it up and always add missing newline on save)

## Checklist before merge
- [x] Code formatting checked

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,5 +5,5 @@ if git diff --cached --name-only --diff-filter=ACM | grep --max-count=1 -E "^des
 fi
 
 if git diff --cached --name-only --diff-filter=ACM | grep -E --max-count=1 '\.scala$|\.sbt$' > /dev/null; then
-  ./sbtwrapper formatStagedScalaFiles
+  ./formatStagedBackendFiles.sh
 fi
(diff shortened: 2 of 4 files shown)


.husky/pre-commit
```diff
@@ -5,5 +5,10 @@ if git diff --cached --name-only --diff-filter=ACM | grep --max-count=1 -E "^des
 fi
 
 if git diff --cached --name-only --diff-filter=ACM | grep -E --max-count=1 '\.scala$|\.sbt$' > /dev/null; then
-  ./sbtwrapper formatStagedScalaFiles
-fi
\ No newline at end of file
+  # use real sbt if possible: https://github.com/dwijnand/sbt-extras/issues/377
+  if hash sbt 2>/dev/null; then
```
coutoPL, line 9: can we move it to a separate script file e.g. `formatStagedBackendFiles.sh`? To be consistent with the frontend approach done above

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
```
coutoPL, line 5: Why do we need it? As far as I see [scalafmt should handle that](https://github.com/scalameta/scalafmt/pull/1333). I guess it doesn't work?

The problem with this approach I see, is the fact that we cannot be sure it it will be taken into consideration. But maybe this is not a problem, because, as I mentioned somewhere before, the current mechanism can be skipped by adding `--no-verify` to `git commit` or editing the file though GH UI.

piotrp: scalafmt doesn't touch Markdown or bash files. This is a generic formatting rule for editors that understand EditorConfig (e.g. IDEA)

.husky/pre-commit
piotrp: done

coutoPL (approved):

PR #4846, opened 2023-10-12
[NU-765] fixed submodules dev proxy setup

JulianWielga:
fallback to BACKEND_DOMAIN if submodules not started locally - displays missing components in dev [NU-765]

## Checklist before merge

[NU-765]: https://touk-jira.atlassian.net/browse/NU-765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/designer/client/webpack.config.ts
+++ b/designer/client/webpack.config.ts
@@ -117,19 +117,25 @@ const config: Configuration = {
                 target: "http://localhost:5001",
                 changeOrigin: true,
                 pathRewrite: {
-                    "^/submodules/components": "/",
+                    "^/submodules/components": "",
+                },
+                onError: (err, req, res) => {
+                    const url = `${process.env.BACKEND_DOMAIN}/submodules/components${req.path}`;
+                    console.warn(`Submodules not available locally - falling back to ${url}`);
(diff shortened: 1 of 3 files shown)


designer/client/webpack.config.ts
```diff
@@ -117,19 +117,25 @@ const config: Configuration = {
                 target: "http://localhost:5001",
                 changeOrigin: true,
                 pathRewrite: {
-                    "^/submodules/components": "/",
+                    "^/submodules/components": "",
+                },
+                onError: (err, req, res) => {
```
arkadius, line 122: We should redirect in case of any error or only some 404? Maybe at least we should print this error as well to avoid some strange situation where some nasty error is silently hidden?

arkadius: After an offline discussion, @JulianWielga encouraged me, that this situation is already a narrow case and it won't give us too much. He don't see printing errors information as a valuable thing

arkadius (approved):
LGTM

PR #4887, opened 2023-10-17
ValidationExpressionParameterValidator

bartektartanus:
## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -273,7 +273,8 @@ case class ValidationExpressionParameterValidator(
   private def validateValue(paramName: String, value: Any)(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit] = {
-    val context = Context("validator", Map(paramName -> value), None)
+    // TODO: paramName should be used here, but a lot of parameters have names that are not valid variables (e.g. "Topic name")
+    val context = Context("validator", Map("value" -> value), None)
     Try(validationExpression.evaluate[Boolean](context, Map())).fold(
       e =>
         invalid(
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -13,6 +13,7 @@ import io.circe.parser._
 
 import scala.util.Try
 import pl.touk.nussknacker.engine.api.CirceUtil._
+import pl.touk.nussknacker.engine.api.definition.ValidationExpressionParameterValidator.variableName
 import pl.touk.nussknacker.engine.api.{Context, NodeId}
 import pl.touk.nussknacker.engine.api.expression.{Expression => ApiExpression}
 import pl.touk.nussknacker.engine.graph.expression.Expression
@@ -274,7 +275,7 @@ case class ValidationExpressionParameterValidator(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit] = {
     // TODO: paramName should be used here, but a lot of parameters have names that are not valid variables (e.g. "Topic name")
-    val context = Context("validator", Map("value" -> value), None)
+    val context = Context("validator", Map(variableName -> value), None)
     Try(validationExpression.evaluate[Boolean](context, Map())).fold(
       e =>
         invalid(
@@ -302,6 +303,8 @@ case class ValidationExpressionParameterValidator(
 
 object ValidationExpressionParameterValidator {
 
+  val variableName = "value"
+
   implicit val encoder: Encoder[ValidationExpressionParameterValidator] = deriveEncoder
   implicit val decoder: Decoder[ValidationExpressionParameterValidator] = deriveDecoder
 
@@ -310,17 +313,9 @@ object ValidationExpressionParameterValidator {
   }
 
   implicit val apiExpressionDecoder: Decoder[ApiExpression] = {
-    Decoder.forProduct2[ApiExpression, String, String]("language", "original")((l, o) => {
-      new ApiExpression() {
-        override def language: String = l
-
-        override def original: String = o
-
-        override def evaluate[T](ctx: Context, globals: Map[String, Any]): T = throw new RuntimeException(
-          "Cannot evaluate Expression in ValidationExpressionParameterValidator as loading from config file is not supported"
-        )
-      }
-    })
+    Decoder.failedWithMessage(
+      "Cannot evaluate Expression in ValidationExpressionParameterValidator as loading from config file is not supported"
+    )
   }
 
 }
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -4,13 +4,17 @@ import java.util.ServiceLoader
 import java.util.regex.Pattern
 import cats.data.Validated
 import cats.data.Validated.{invalid, valid}
+import io.circe.{Decoder, Encoder}
 import io.circe.generic.extras.ConfiguredJsonCodec
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
 import pl.touk.nussknacker.engine.api.context.PartSubGraphCompilationError
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError._
 import io.circe.parser._
-import org.apache.commons.lang3.StringUtils
+
+import scala.util.Try
 import pl.touk.nussknacker.engine.api.CirceUtil._
-import pl.touk.nussknacker.engine.api.NodeId
+import pl.touk.nussknacker.engine.api.{Context, NodeId}
+import pl.touk.nussknacker.engine.api.expression.{Expression => ApiExpression}
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
 import scala.collection.concurrent.TrieMap
@@ -35,6 +39,7 @@ trait Validator {
 @ConfiguredJsonCodec sealed trait ParameterValidator extends Validator
 
 //TODO: These validators should be moved to separated module
+
 case object MandatoryParameterValidator extends ParameterValidator {
 
   override def isValid(paramName: String, expression: Expression, value: Option[Any], label: Option[String])(
@@ -247,6 +252,75 @@ case object JsonValidator extends ParameterValidator {
       paramName,
       nodeId
     )
+
+}
+
+case class ValidationExpressionParameterValidator(
+    validationExpression: ApiExpression,
+    validationFailedMessage: Option[String]
+) extends ParameterValidator {
+
+  override def isValid(paramName: String, expression: Expression, value: Option[Any], label: Option[String])(
+      implicit nodeId: NodeId
+  ): Validated[PartSubGraphCompilationError, Unit] = {
+    value match {
+      case None       => valid(())
+      case Some(null) => valid(())
+      case Some(v)    => validateValue(paramName, v)
+    }
+  }
+
+  private def validateValue(paramName: String, value: Any)(
+      implicit nodeId: NodeId
+  ): Validated[PartSubGraphCompilationError, Unit] = {
+    val context = Context("validator", Map(paramName -> value), None)
+    Try(validationExpression.evaluate[Boolean](context, Map())).fold(
+      e =>
+        invalid(
+          CustomParameterValidationError(
+            s"Evaluation of validation expression '${validationExpression.original}' of language ${validationExpression.language} failed: ${e.getMessage}",
+            s"Please provide value that satisfies the validation expression '${validationExpression.original}'",
+            paramName,
+            nodeId.id
+          )
+        ),
+      result => if (result) valid(()) else invalid(error(paramName, nodeId.id))
+    )
+  }
+
+  private def error(paramName: String, nodeId: String): CustomParameterValidationError = CustomParameterValidationError(
+    validationFailedMessage.getOrElse(
+      s"This field has to satisfy the validation expression '${validationExpression.original}'"
+    ),
+    s"Please provide value that satisfies the validation expression '${validationExpression.original}'",
+    paramName,
+    nodeId
+  )
+
+}
+
+object ValidationExpressionParameterValidator {
+
+  implicit val encoder: Encoder[ValidationExpressionParameterValidator] = deriveEncoder
+  implicit val decoder: Decoder[ValidationExpressionParameterValidator] = deriveDecoder
+
+  implicit val apiExpressionEncoder: Encoder[ApiExpression] = {
+    Encoder.forProduct2("language", "original")(e => (e.language, e.original))
+  }
+
+  implicit val apiExpressionDecoder: Decoder[ApiExpression] = {
+    Decoder.forProduct2[ApiExpression, String, String]("language", "original")((l, o) => {
+      new ApiExpression() {
+        override def language: String = l
+
+        override def original: String = o
+
+        override def evaluate[T](ctx: Context, globals: Map[String, Any]): T = throw new RuntimeException(
+          "Cannot evaluate Expression in ValidationExpressionParameterValidator as loading from config file is not supported"
+        )
+      }
+    })
+  }
 
 }
 
--- a/components-api/src/test/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidatorSpec.scala
+++ b/components-api/src/test/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidatorSpec.scala
@@ -3,8 +3,11 @@ package pl.touk.nussknacker.engine.api.definition
 import org.scalatest.funsuite.AnyFunSuite
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.prop.TableDrivenPropertyChecks
-import pl.touk.nussknacker.engine.api.NodeId
+import org.springframework.expression.spel.standard.SpelExpressionParser
+import org.springframework.expression.spel.support.StandardEvaluationContext
+import pl.touk.nussknacker.engine.api.{Context, NodeId}
 import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.engine.api.expression.{Expression => ApiExpression}
 
 import scala.jdk.CollectionConverters._
 
@@ -221,6 +224,46 @@ class ParameterValidatorSpec extends AnyFunSuite with TableDrivenPropertyChecks 
     ) { (expression, value, validator, expected) =>
       validator.isValid("dummy", Expression.spel(expression), value, None).isValid shouldBe expected
     }
+  }
+
+  test("ValidationExpressionParameterValidator") {
+    forAll(
+      Table(
+        ("validationExpression", "paramName", "inputExpression", "value", "isValid"),
+        ("#param > 10", "param", "", Some(null), true),
+        ("#param > 10", "param", "#input.foo", None, true),
+        ("#param > 10", "param", "-14", Some(-14), false),
+        (
+          "#param.toLowerCase() == \"left\" || #param.toLowerCase() == \"right\"",
+          "param",
+          "'lEfT'",
+          Some("lEfT"),
+          true
+        ),
+        ("#param.toLowerCase() == \"left\" || #param.toLowerCase() == \"right\"", "param", "'up'", Some("up"), false),
+        ("#param", "param", "'up'", Some("up"), false),
+        ("#fooBar", "param", "'up'", Some("up"), false),
+        ("#list.size() == 2 && #list[0] == 'foo'", "list", "{'foo', 'bar'}", Some(List("foo", "bar").asJava), true)
+      )
+    ) { (validationExpression, paramName, inputExpression, value, isValid) =>
+      ValidationExpressionParameterValidator(new TestSpelExpression(validationExpression), None)
+        .isValid(paramName, Expression.spel(inputExpression), value, None)(nodeId)
+        .isValid shouldBe isValid
+    }
+  }
+
+}
+
+private class TestSpelExpression(expression: String) extends ApiExpression {
+
+  override def language: String = "spel"
+
+  override def original: String = expression
+
+  override def evaluate[T](ctx: Context, globals: Map[String, Any]): T = {
+    val context = new StandardEvaluationContext()
+    ctx.variables.foreach { case (param, value) => context.setVariable(param, value) }
+    new SpelExpressionParser().parseRaw(expression).getValue(context).asInstanceOf[T]
   }
 
 }
--- a/components-api/src/test/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidatorSpec.scala
+++ b/components-api/src/test/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidatorSpec.scala
@@ -230,20 +230,20 @@ class ParameterValidatorSpec extends AnyFunSuite with TableDrivenPropertyChecks 
     forAll(
       Table(
         ("validationExpression", "paramName", "inputExpression", "value", "isValid"),
-        ("#param > 10", "param", "", Some(null), true),
-        ("#param > 10", "param", "#input.foo", None, true),
-        ("#param > 10", "param", "-14", Some(-14), false),
+        ("#value > 10", "param", "", Some(null), true),
+        ("#value > 10", "param", "#input.foo", None, true),
+        ("#value > 10", "param", "-14", Some(-14), false),
         (
-          "#param.toLowerCase() == \"left\" || #param.toLowerCase() == \"right\"",
+          "#value.toLowerCase() == \"left\" || #value.toLowerCase() == \"right\"",
           "param",
           "'lEfT'",
           Some("lEfT"),
           true
         ),
-        ("#param.toLowerCase() == \"left\" || #param.toLowerCase() == \"right\"", "param", "'up'", Some("up"), false),
-        ("#param", "param", "'up'", Some("up"), false),
-        ("#fooBar", "param", "'up'", Some("up"), false),
-        ("#list.size() == 2 && #list[0] == 'foo'", "list", "{'foo', 'bar'}", Some(List("foo", "bar").asJava), true)
+        ("#value.toLowerCase() == \"left\" || #value.toLowerCase() == \"right\"", "param", "'up'", Some("up"), false),
+        ("#value", "param", "'up'", Some("up"), false),
+        ("#value", "param", "'up'", Some("up"), false),
+        ("#value.size() == 2 && #value[0] == 'foo'", "list", "{'foo', 'bar'}", Some(List("foo", "bar").asJava), true)
       )
     ) { (validationExpression, paramName, inputExpression, value, isValid) =>
       ValidationExpressionParameterValidator(new TestSpelExpression(validationExpression), None)
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -17,7 +17,7 @@
 * [#4919](https://github.com/TouK/nussknacker/pull/4919) Improvement: Support for handling runtime exceptions at FlinkTestScenarioRunner
 * [#4923](https://github.com/TouK/nussknacker/pull/4923) Fix non-unique test case ids when testing scenario with union
 * [#4745](https://github.com/TouK/nussknacker/pull/4745) Improvement: Stricter Node and scenario id validation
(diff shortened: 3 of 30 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -38,12 +42,48 @@ trait Validator {
...
+) extends ParameterValidator {
+
+  private val parser: ExpressionParser   = new SpelExpressionParser()
+  private val parsedValidationExpression = parser.parseExpression(validationExpression)
+
+  override def isValid(paramName: String, value: Any, label: Option[String])(
+      implicit nodeId: NodeId
+  ): Validated[PartSubGraphCompilationError, Unit] = {
+    val context: StandardEvaluationContext = new StandardEvaluationContext()
+    context.setVariable(paramName, value)
```
arkadius, line 57: What about the situation when `paramName` is not the correct variable name? Wouldn't be better to use just the `#value`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -38,12 +42,48 @@ trait Validator {
...
+    Try(parsedValidationExpression.getValue(context, classOf[Boolean])).fold(
+      e =>
+        invalid(
+          CustomParameterValidationError(
+            s"Expression evaluation failed: ${e.getMessage}",
```
arkadius, line 62: The user will see this message in the context when he/she uses the fragment. I think it won't be clear, which expression is evaluated with failure. Do you know how to change this message to make this more obvious?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -38,12 +42,48 @@ trait Validator {
...
+            paramName,
+            nodeId.id
+          )
+        ),
+      result => if (result) valid(()) else invalid(error(paramName, nodeId.id))
+    )
+  }
+
+  private def error(paramName: String, nodeId: String): CustomParameterValidationError = CustomParameterValidationError(
+    validationFailedMessage.getOrElse("This field has to satisfy the validation expression"),
```
arkadius, line 73: WDYT about displaying the expression in the fallback? I mean:
`Value '123' doesn't satisfy the defined validation expression '#value > 0'`. Also, shouldn't we have some placeholders in the validation failed message? E.g. `#value`? We can substitute it with just a `replaceAll`.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -56,13 +96,16 @@ case object MandatoryParameterValidator extends ParameterValidator {
 
 case object NotBlankParameterValidator extends ParameterValidator {
 
-  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("['\"]\\s*['\"]")
+  private final lazy val blankStringLiteralPattern: Pattern = Pattern.compile("^\\s*$")
 
   // TODO: for now we correctly detect only literal expression with blank string - on this level (not evaluated expression) it is the only thing that we can do
-  override def isValid(paramName: String, expression: String, label: Option[String])(
+  override def isValid(paramName: String, expression: Any, label: Option[String])(
```
arkadius, line 102: It should be `value` instead of `expression` now

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -77,15 +120,19 @@ case object NotBlankParameterValidator extends ParameterValidator {
...
+    // FIXME: we should properly evaluate `possibleValues`
     val values = possibleValues.map(possibleValue => possibleValue.expression)
 
-    if (StringUtils.isBlank(value) || values.contains(value))
-      valid(())
-    else
-      invalid(InvalidPropertyFixedValue(paramName, label, value, possibleValues.map(_.expression)))
+    value match {
+      case null                                  => valid(())
+      case s: String if values.contains(s"'$s'") => valid(())
```
arkadius, line 131: Fixes values are not only for strings. It was used with many other types - take a look at `AggregateHelper.SIMPLE_EDITOR`. Maybe we need to do a similar thing as we did in `SpelExpressionParameterValidator`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -96,29 +143,31 @@ case class LiteralRegExpParameterValidator(pattern: String, message: String, des
...
     def toResult(validated: Boolean) =
       if (validated) valid(()) else invalid(MismatchParameter(message, description, paramName, nodeId.id))
-
-    if (value.matches("^'.*'$")) { // workaround for SpEL comprehension
-      val trimmedSPeLValue = value.replaceAll("^'|'$", "")
-      toResult(regexpPattern.matcher(trimmedSPeLValue).matches())
-    } else {
-      toResult(StringUtils.isBlank(value) || regexpPattern.matcher(value).matches())
+    value match {
+      case s: String => toResult(StringUtils.isBlank(s) || regexpPattern.matcher(s).matches())
```
arkadius, line 152: this `StringUtils.isBlank` case was to handle `null` not the blank string. We don't want to always allow blanks - no matter which regex you will have. Please add a test case for that if it was missing :)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -180,13 +252,13 @@ case class MaximalNumberValidator(maximalNumber: BigDecimal) extends ParameterVa
 case object JsonValidator extends ParameterValidator {
 
   // Blank value should be not validate - we want to chain validators
-  override def isValid(paramName: String, value: String, label: Option[String])(
+  override def isValid(paramName: String, value: Any, label: Option[String])(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit] = {
-    val strippedValue = value.stripPrefix("'").stripSuffix("'").trim
-    val parsingResult = parse(strippedValue)
+    val trimmedValue  = value.toString.trim
```
arkadius, line 258: Why do you do `toString`? Let's handle non-string types explicitly.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -180,13 +252,13 @@ case class MaximalNumberValidator(maximalNumber: BigDecimal) extends ParameterVa
...
+    val parsingResult = parse(trimmedValue)
 
-    if (StringUtils.isBlank(value) || parsingResult.isRight)
+    if (StringUtils.isBlank(trimmedValue) || parsingResult.isRight)
```
arkadius, line 261: The same as above - `isBlank` was for nulls

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -77,15 +120,19 @@ case object NotBlankParameterValidator extends ParameterValidator {
...
 
-    if (StringUtils.isBlank(value) || values.contains(value))
-      valid(())
-    else
-      invalid(InvalidPropertyFixedValue(paramName, label, value, possibleValues.map(_.expression)))
+    value match {
+      case null                                  => valid(())
+      case s: String if values.contains(s"'$s'") => valid(())
+      case s: String => invalid(InvalidPropertyFixedValue(paramName, label, s"'$s'", possibleValues.map(_.expression)))
+      case x if values.contains(x.toString) => valid(())
```
arkadius, line 133: Please avoid implicit `toString` calls

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -126,7 +126,7 @@ object typing {
   case object TypedNull extends TypingResult {
     override def withoutValue: TypedNull.type = TypedNull
 
-    override val valueOpt: None.type = None
+    override val valueOpt: Some[Null] = Some(null)
```
arkadius, line 129: Please add a comment that we do `Some(null)` intentionally.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidatorSpec.scala
```diff
@@ -13,10 +17,11 @@ class ParameterValidatorSpec extends AnyFunSuite with TableDrivenPropertyChecks
     forAll(
       Table(
         ("inputExpression", "isValid"),
-        ("''", false),
-        (" '' ", false),
-        ("\"\"", false),
-        (" \"\" ", false),
+        ("", false),
```
arkadius, line 20: Maybe instead of changing the test input, we should change the scope of the test? I mean to type the expressions and after that run the `isValid` method on the result. It will make these tests more "functional" then "unit" which is IMO a good approach.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ScenarioPropertiesValidator.scala
```diff
@@ -60,7 +60,17 @@ class ScenarioPropertiesValidator(
 
     propertiesWithConfiguredValidator
       .collect { case (property, Some(config), validator: ParameterValidator) =>
-        validator.isValid(property._1, property._2, config.label).toValidatedNel
+        val expression = property._2
+        val value = if (expression.isBlank) {
+          null
+        } else {
+          // FIXME: scenario properties does not have TypingResult hence this little hack to convert them to proper values
```
arkadius, line 69: I assume that you plan to fix this in this PR?

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
```diff
@@ -87,7 +89,10 @@ class LiteKafkaUniversalAvroSchemaFunctionalTest
           sConfig(sampleBoolean, booleanSchema, integerSchema, sampleLong),
           invalidTypes(s"actual: '${typedLong.display}' expected: 'Integer'")
         ),
-        (sConfig(null, nullSchema, integerSchema, Input), invalidTypes("actual: 'Null' expected: 'Integer'")),
+        (
+          sConfig(null, nullSchema, integerSchema, Input),
+          emptyMandatoryParameter().findValid(invalidTypes("actual: 'Unknown' expected: 'Integer'"))
```
arkadius, line 94: Why didn't you just comment out this `invalidTypes` if this will always return the result of  `emptyMandatoryParameter`?

engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/RequestResponseTestWithParametersTest.scala
```diff
@@ -176,4 +172,28 @@ class RequestResponseTestWithParametersTest extends AnyFunSuite with Matchers {
     parameters.map(p => SimplifiedParam(p.name, p.typ, p.editor)) should contain theSameElementsAs expectedParameters
   }
 
+  test("should generate fragnent parameter with spel expression validator") {
```
arkadius, line 175: typo

scenario-api/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
```diff
@@ -298,7 +298,12 @@ object node {
   // shape of this data should probably change, currently we leave it for backward compatibility
   object FragmentInputDefinition {
 
-    @JsonCodec case class FragmentParameter(name: String, typ: FragmentClazzRef)
+    @JsonCodec case class FragmentParameter(
+        name: String,
+        typ: FragmentClazzRef,
+        validationExpression: Option[Expression] = None,
+        validationFailedMessage: Option[String] = None,
```
arkadius, line 305: failed message should be available only when expression is defined

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/GenericNodeTransformationValidator.scala
```diff
@@ -181,39 +181,26 @@ class GenericNodeTransformationValidator(
           )
           .sequence
         params
-          .andThen { branchParams =>
-            branchParams
-              .map { case (branchId, expression) =>
-                Validations.validate(
-                  parameter,
-                  evaluatedparam.Parameter(ParameterNaming.getNameForBranchParameter(parameter, branchId), expression)
```
arkadius, line 189: What happened with this code?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -38,12 +42,48 @@ trait Validator {
 
 //TODO: These validators should be moved to separated module
 
+case class SpelExpressionParameterValidator(
+    validationExpression: String,
+    validationFailedMessage: Option[String]
+) extends ParameterValidator {
+
+  private val parser: ExpressionParser   = new SpelExpressionParser()
```
arkadius, line 50: It is definitely not a good approach to use spell classes directly. We have some customizations of spel that we want to keep consistent across all spel expressions, we have global helpers as well.
Also, these things should be done outside of the constructor. The class should have objects that already can evaluate the expression.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: `#value` would be fine but using `paramName` allows us to validate parameters against each other. For example, enricher has two params: `hourFrom` and `hourTo`. You can write validation expression for the first one:
`#hourFrom >= 0 && #hourFrom < 24` 
and for second one:
`#hourTo >= 0 && #hourTo < 24 && #hourFrom < #hourTo`

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/GenericNodeTransformationValidator.scala
bartektartanus: Validation is now done after compilation :)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: It is an interesting idea and we can go in this direction, but it requires to do some variable name sanitization

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/GenericNodeTransformationValidator.scala
arkadius: Ok, but what about `getNameForBranchParameter`? Thanks to that, errors are shown in the exact place in the form. Branch parameters have two levels: one is parameter level and next one is expression for the given edge. You need to identify this two level's path to recognize where is the error.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/FragmentComponentDefinitionExtractor.scala
```diff
@@ -101,30 +103,45 @@ class FragmentComponentDefinitionExtractor(
...
+    val config          = componentConfig.params.flatMap(_.get(fragmentParameter.name)).getOrElse(ParameterConfig.empty)
     val parameterData   = ParameterData(typ, Nil)
     val extractedEditor = EditorExtractor.extract(parameterData, config)
+
+    val customExpressionValidator = fragmentParameter.validationExpression.flatMap(expr => {
+      expr match {
+        // TODO? log warning if expression is not (valid) Spel
+        case Expression(Language.Spel, e) =>
+          Some(SpelExpressionParameterValidator(e, fragmentParameter.validationFailedMessage))
+        case _ => None
```
arkadius, line 133: We should throw an exception instead of silently ignore this case.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: But this is just an editor, not a validator. Dual editor - to be precise, which allows user to write more complex aggregators, like `#AGG.map({last: #AGG.last, list: #AGG.list})`. The real (maybe even philosophical) question is - should we use **FixedValues**Validator for such complex objects? It was easy to compare them earlier - just check if expression strings are the same. Actually, aggregators can't even be compared:
image

Maybe validator API needs some further enhancement? Maybe we should also pass the raw expression?

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
arkadius: What an interesting side effect :) Why not to just combineAll if you know that both are invalid and you are not interested in valid case? :)

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/GenericNodeTransformationValidator.scala
arkadius: Nice, thx

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: > The class should have objects that already can evaluate the expression.

This little annotation:
`@ConfiguredJsonCodec sealed trait ParameterValidator extends Validator`
makes it very difficult to pass complex objects as constructor parameters. This issue can be by-passed if we implement `CustomParameterValidator` trait, but it still will be impossible to define this validator in config file.

arkadius: > will be impossible to define this validator in config file.

Validators config is IMO the feature that we should consider to remove eventually. After adding `AdditionalUIConfigProvider` there is an easier way to achieve configuratiblity programmatically. I don't know usages where an administrator did this in config manually.

Regards that fact, I see two options:
- `CustomParameterValidator` as you suggested + comment that we don't support it in config - only in `AdditionalUIConfigProvider`
- Write serializer from scratch instead of using generated - it would be difficult to pass some additional context of variables to them, but maybe circe has some options to do that

engine/lite/components/request-response-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/RequestResponseTestWithParametersTest.scala
bartektartanus: fixed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: CustomParameterValidator solution does not work, so I finally decided to write deserializer that throws an exception when trying to evaluate an expression.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/FragmentComponentDefinitionExtractor.scala
bartektartanus: Replaced SpelExpressionParser with ExpressionCompiler, so now the exception is not needed :) And furthermore, the validator is now generic - it supports any expression (`pl.touk.nussknacker.engine.api.expression.Expression`)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: done

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: Name sanitization should be done when defining fragment parameters, not when we are trying to fill them and validate, WDYT?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: It should be more obvious now :)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: As I added `expression` to `isValid` method, this validator now works the same as before.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
bartektartanus: Wrapped it up into another case class

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: I think that name validation should be probably done when defining fragment parameters. Here we are loosely coupled with fragments so we need to handle situations when these names are invalid variable names. Let's consider validation of some other parameters like `Topic name`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: It doesn't work the same as we discussed offline

docs/Changelog.md
```diff
@@ -17,6 +17,8 @@
 * [#4919](https://github.com/TouK/nussknacker/pull/4919) Improvement: Support for handling runtime exceptions at FlinkTestScenarioRunner
 * [#4923](https://github.com/TouK/nussknacker/pull/4923) Fix non-unique test case ids when testing scenario with union
 * [#4745](https://github.com/TouK/nussknacker/pull/4745) Improvement: Stricter Node and scenario id validation
+* [#4928](https://github.com/TouK/nussknacker/pull/4928) Braking change: `Validator.isValid` method now takes expression and evaluated value instead of just expression
```
arkadius, line 20: typo: breaking

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -258,6 +266,65 @@ case object LiteralParameterValidator {
...
+  implicit val encoder: Encoder[ValidationExpressionParameterValidator] = deriveEncoder
+  implicit val decoder: Decoder[ValidationExpressionParameterValidator] = deriveDecoder
+
+  implicit val apiExpressionEncoder: Encoder[ApiExpression] = {
+    Encoder.forProduct2("language", "original")(e => (e.language, e.original))
+  }
+
+  implicit val apiExpressionDecoder: Decoder[ApiExpression] = {
+    Decoder.forProduct2[ApiExpression, String, String]("language", "original")((l, o) => {
+      new ApiExpression() {
```
arkadius, line 314: Why not to throw expression immediately instead of creating failing ApiExpression?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/FragmentComponentDefinitionExtractor.scala
```diff
@@ -101,30 +104,52 @@ class FragmentComponentDefinitionExtractor(
...
+      componentConfig: SingleComponentConfig,
+      typ: typing.TypingResult,
+      fragmentParameter: FragmentParameter
+  )(implicit nodeId: NodeId): Parameter = {
+    val config          = componentConfig.params.flatMap(_.get(fragmentParameter.name)).getOrElse(ParameterConfig.empty)
     val parameterData   = ParameterData(typ, Nil)
     val extractedEditor = EditorExtractor.extract(parameterData, config)
+
+    val expressionConfig   = ProcessDefinitionBuilder.empty.expressionConfig
+    val expressionCompiler = ExpressionCompiler.withOptimization(classLoader, null, expressionConfig, null)
```
arkadius, line 132: we can use simple dict registry implementation, but `typeDefinitionSet` mus be real, from model. otherwise e.g. method calling won't work

bartektartanus: It wasn't easy, but it's done :)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: This way Nu app will still start, despite having bad config? Or exception will be catched anyway so there is no point of doing this?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: You're right. I've replaced that with `#value` and added "TODO".

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/FragmentComponentDefinitionExtractor.scala
```diff
@@ -101,30 +105,53 @@ class FragmentComponentDefinitionExtractor(
...
+  private def toParameter(
+      componentConfig: SingleComponentConfig,
+      typ: typing.TypingResult,
+      fragmentParameter: FragmentParameter
+  )(implicit nodeId: NodeId): Parameter = {
+    val config          = componentConfig.params.flatMap(_.get(fragmentParameter.name)).getOrElse(ParameterConfig.empty)
     val parameterData   = ParameterData(typ, Nil)
     val extractedEditor = EditorExtractor.extract(parameterData, config)
+    val expressionConfig = ProcessDefinitionBuilder.empty.expressionConfig
+    val expressionCompiler =
```
arkadius, line 132: Can we replace this with IoC/DI? I mean let's put `ExpressionCompiler` to constructor. Thanks to that
1. we will pass much fewer dependencies to `FragmentComponentDefinitionExtractor` 
2. I see that  `ExpressionCompiler` is already created on the same level as `FragmentComponentDefinitionExtractor` so probably we won't need to create it again
3.  This is potentially an expensive object, let's create it as fewer times as it is possible

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -250,6 +255,76 @@ case object JsonValidator extends ParameterValidator {
...
+      case Some(null) => valid(())
+      case Some(v)    => validateValue(paramName, v)
+    }
+  }
+
+  private def validateValue(paramName: String, value: Any)(
+      implicit nodeId: NodeId
+  ): Validated[PartSubGraphCompilationError, Unit] = {
+    // TODO: paramName should be used here, but a lot of parameters have names that are not valid variables (e.g. "Topic name")
+    val context = Context("validator", Map("value" -> value), None)
```
arkadius, line 277: Let's extract this "value" to some constant

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: Why it will start? Why expception will be catched? Who catch it? What is a point of creating fake object? Throwing an exception instead of this is less line of code.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: fixed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
bartektartanus: Done :)

arkadius (approved):
Good job

PR #4888, opened 2023-10-17
one-liner in readme/docs updated for mac users (--network host)

raphaelsolarski:


dockerhub/nussknacker-lite-runtime-app/README.md
```diff
@@ -22,7 +22,7 @@ where:
 ```
 tasksCount: 2
 ```
-- `--network host` - to be able to connect with Kafka and Schema Registry exposed on host machine
+- `--network host` - to be able to connect with Kafka and Schema Registry exposed on host machine. Note: It only works on linux and is used to connect to existing kafka/schema registry. In case of other OS you have to use different methods to make it accessible from Nussknacker container (e.g start Kafka/SR and Nussknacker in a single docker network)
```
arkadius, line 25: I think that `--network host` is a little bit obsolete approach. Have you tried to use `host.docker.internal` on your Mac? Regarding this post: https://medium.com/@TimvanBaarsen/how-to-connect-to-the-docker-host-from-inside-a-docker-container-112b4c71bc66 it should work on Mac and Windows and there is some workaround to emulate the same behaviour on Linux. IMO we could just replace localhost with it.

raphaelsolarski: Si, `host.docker.internal` will allow you to connect to some port bound on loopback host interface from container but such an approach would require from you to configure additional listener in kafka with "host.docker.internal" address on a separate port. SR will probably work as expected.

arkadius: Ah you are right.

arkadius (approved):

PR #4892, opened 2023-10-18
NU-1129 fix problem alert error message

bohdanprog:
## Describe your changes
fix problem alert error message
## Checklist before merge

### DIFF
--- a/designer/client/src/containers/Notifications.tsx
+++ b/designer/client/src/containers/Notifications.tsx
@@ -112,8 +112,6 @@ export function Notifications(): JSX.Element {
                 } else if (possibleServerNotAvailableHttpStatuses.some((status) => status === error?.response?.status)) {
                     handleChangeConnectionError("NO_BACKEND_ACCESS");
                 } else {
-                    const errorResponseData = error?.response?.data || error.message;
-                    console.error(`Error: ${errorResponseData}`);
                     dispatch(
                         NotificationActions.error(
                             i18next.t("notification.error.cannotFetchBackendNotifications", "Cannot fetch backend notification"),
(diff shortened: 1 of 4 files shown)


designer/client/src/containers/Notifications.tsx
```diff
@@ -109,16 +109,14 @@ export function Notifications(): JSX.Element {
...
+                } else if (possibleServerNotAvailableHttpStatuses.some((status) => status === error?.response?.status)) {
                     handleChangeConnectionError("NO_BACKEND_ACCESS");
                 } else {
                     const errorResponseData = error?.response?.data || error.message;
-
+                    console.error(`Error: ${errorResponseData}`);
                     dispatch(
                         NotificationActions.error(
                             i18next.t("notification.error.cannotFetchBackendNotifications", "Cannot fetch backend notification"),
-                            errorResponseData,
```
Dzuming, line 120: In this approach, we won't display all error details which is IMO bad, what we really want to achieve here, is not to show unhandled error messages like from a reverse proxy server, where there is no control over it, but still, we want to display handled errors send from our backend server. Maybe we need to change this condition?

```
const errorResponseData = error?.response?.data || error.message;
```

to

```
const errorResponseData = error?.response?.data;
```
?

designer/client/src/containers/Notifications.tsx
```diff
@@ -109,16 +109,14 @@ export function Notifications(): JSX.Element {
 
                 if (!isNetworkAccess) {
                     handleChangeConnectionError("NO_NETWORK_ACCESS");
-                } else if (possibleServerNotAvailableHttpStatuses.some((status) => status === error.response.status)) {
+                } else if (possibleServerNotAvailableHttpStatuses.some((status) => status === error?.response?.status)) {
                     handleChangeConnectionError("NO_BACKEND_ACCESS");
                 } else {
                     const errorResponseData = error?.response?.data || error.message;
-
+                    console.error(`Error: ${errorResponseData}`);
```
Dzuming, line 116: Maybe it's better to move this console.error inside of the `NotificationActions.error` method to make it callable in all display error notifications? Also, I'm wondering shouldn't log more info like https status or response origin (it's good to know if the error origin was reverse proxy or backend server)

Dzuming (approved):

PR #4902, opened 2023-10-19
Nu 1133/problem with counts

bohdanprog:
## Describe your changes
fix problem with counts
## Checklist before merge


designer/client/src/actions/nk/displayProcessCounts.ts
```diff
@@ -15,11 +16,34 @@ export function displayProcessCounts(processCounts: ProcessCounts): DisplayProce
     };
 }
 
+const checkActualCounts = (processCounts: ProcessCounts, processToDisplay: Process) => {
```
Dzuming, line 19: maybe something like checkPossibleCountsToCalculate will be meaningful? IIUC we handle the situation that, there are changes in the scenario  before the deploy, so the backend calculates old state of the counts, and until someone clicks deploy new version, there will be a difference between frontend and backend.

designer/client/src/actions/nk/displayProcessCounts.ts
```diff
@@ -15,11 +16,34 @@ export function displayProcessCounts(processCounts: ProcessCounts): DisplayProce
...
+    const processCountsName = Object.keys(processCounts).sort((a, b) => a.localeCompare(b));
+    const nodes = processToDisplay.nodes
```
Dzuming, line 21: what about rename it from`nodes` to `uncountableNodes`?

designer/client/src/actions/nk/displayProcessCounts.ts
```diff
@@ -15,11 +16,34 @@ export function displayProcessCounts(processCounts: ProcessCounts): DisplayProce
...
+        .sort((a, b) => a.id.localeCompare(b.id))
+        .filter((node, index) => node.id !== processCountsName[index]);
+    const newProcessCounts = { ...processCounts };
+    if (nodes.length !== 0 && processCountsName.length !== 0) {
```
Dzuming, line 25: Why do we need this check `processCountsName.length !== 0`?

designer/client/src/components/graph/EspNode/element.ts
```diff
@@ -69,14 +69,30 @@ export function getStringWidth(str = "", pxPerChar = 8, padding = 7): number {
     return toString(str).length * pxPerChar + 2 * padding;
 }
 
+function getTestCounts(hasCounts: boolean, shortCounts: boolean, count: NodeCounts): string {
+    if (hasCounts) {
```
Dzuming, line 73: Maybe it's my preferential but IMO it will be more readable

```
  if (!hasCounts) {
    return "";
  }

  if (shortCounts) {
    if (count && millify(count?.all)) {
      return count?.all?.toLocaleString();
    }
  } else if (count?.all?.toLocaleString()) {
    return count.all.toLocaleString() || "0";
  }
  
  return "?";
```

it's not blocking change, so resolve comment if you think we don't need it

designer/client/src/actions/nk/displayProcessCounts.ts
bohdanprog: at the moment when we have nodes, but processCounts is an empty object?

designer/client/src/actions/nk/displayProcessCounts.ts
bohdanprog: okay :)

designer/client/src/actions/nk/displayProcessCounts.ts
bohdanprog: okay, good point

designer/client/src/components/graph/EspNode/element.ts
bohdanprog: okay, maybe it's more readable :)

Dzuming (approved):

PR #4919, opened 2023-10-21
Improvement: Support for handling runtime exceptions at FlinkTestScenarioRunner

lciolecki:
## Checklist before merge


utils/flink-components-testkit/src/main/scala/pl/touk/nussknacker/engine/flink/util/test/FlinkTestExceptionHolder.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.flink.util.test
+
+import pl.touk.nussknacker.engine.api.exception.NuExceptionInfo
+import pl.touk.nussknacker.engine.testmode.TestRunId
+
+final class FlinkTestExceptionHolder(testRunId: TestRunId) extends Serializable with AutoCloseable {
```
arkadius, line 6: Why do you need this if we already have `ResultsCollectingListener.exceptionThrown`. I thought that we only need to use already stored exceptions

lciolecki: Oh, I didn't notice that class :) ResultsCollectingListener.exceptionThrown converts exception to `ExceptionResult[_]`. We can try to use it, but there will be an effort to reconvert these exceptions to NuExceptionInfo. I don't know which way is better. WDYT?

lciolecki: During the converting process from NuExceptionInfo to ExceptionResult we lose some information like from NodeComponentInfo.componentInfo..

extensions-api/src/main/scala/pl/touk/nussknacker/engine/testmode/TestProcess.scala
```diff
@@ -33,15 +33,8 @@ object TestProcess {
...
-    def updateExceptionResult(espExceptionInfo: NuExceptionInfo[_ <: Throwable]) = {
-      copy(exceptions =
-        exceptions :+ ExceptionResult(
-          toResult(espExceptionInfo.context),
-          espExceptionInfo.nodeComponentInfo.map(_.nodeId),
-          espExceptionInfo.throwable
-        )
-      )
-    }
+    def updateExceptionResult(espExceptionInfo: NuExceptionInfo[_ <: Throwable]) =
```
arkadius, line 36: Do we need this method? It is just a `copy` now

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
```diff
@@ -40,17 +42,55 @@ object ManagementResources {
 
   import pl.touk.nussknacker.engine.api.CirceUtil._
 
+  // TODO: Why not just use the BestEffortJsonEncoder?
```
arkadius, line 45: I see that we use it but we also wrap the result with object with pretty the `pretty` field.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/testmode/TestProcess.scala
```diff
@@ -9,7 +9,7 @@ object TestProcess {
       nodeResults: Map[String, List[NodeResult[T]]],
       invocationResults: Map[String, List[ExpressionInvocationResult[T]]],
       externalInvocationResults: Map[String, List[ExternalInvocationResult[T]]],
-      exceptions: List[ExceptionResult[T]],
+      exceptions: List[NuExceptionInfo[_ <: Throwable]],
```
arkadius, line 12: It is in extensions-API module so let's add a migration guide entry

arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
lciolecki: Yeah. I realized that after a while.

arkadius (approved):

extensions-api/src/main/scala/pl/touk/nussknacker/engine/testmode/TestProcess.scala
lciolecki: I will leave it. After the change there are some setbacks with type:
image
The effort to fix it doesn't have sens.

PR #4953, opened 2023-10-26
[NU-983] Add missing validations to base components

mslabek:
## Describe your changes
The goal is to make the validation consistent between FE and BE for base components by adding BE validations that are done only on the FE.

### Main goal - Bringing base components FE validations to BE
In base components these validations are:
1. [x] Switch - expression in edges - NotNullParameterValidator
2. [x] Filter - expression - NotNullParameterValidator
3. [x] Variable -  expression - MandatoryParameterValidator
4. [x] MapVariable:
    - [x] field key - unique value in relation to other field names in the node
    - [x] field value - MandatoryParameterValidator
   
Also, this removes the default Mandatory value validation on the FE if no validator was set. This is going to be removed anyways after removing the FE validators.

### Api changes
In MapVariable (and FragmentOuptut) and in Switch the validated parameters are not standard fields and they require their own solutions to be correctly displayed. BE can control under which field the error is displayed by the `fieldName` in the `NodeValidationError` object.
1. In switch nodes fieldName = name of the node to which the choice points to
2. In nodes using a user-controlled dynamic map of key-value pairs (MapVariable (and FragmentOuptut) the simplest solution was to rely on index of the field. This means that FE now has to display them in order of the index:
    - for keys fieldName = `$fields-${index}-$key`
    - for values fieldName = `$fields-${index}-$value`

### Problems
1. Drag'n'drop in MapVariable - this is problematic - after doing drag'n'drop the errors are displayed "as previously" until the next validation comes with the index after dropping. Solving this is problematic and imo out of scope of this PR.

## Checklist before merge


interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+        expectedType: TypingResult,
+        validator: ParameterValidator,
+        expression: Option[TypedExpression],
+        fieldName: String
+    )(implicit nodeId: NodeId): Validated[NonEmptyList[PartSubGraphCompilationError], Unit] = {
+      expression
+        .map { expr =>
+          Validations
+            // The parameter definition name does not mean anything here
+            .validate(Parameter("stub", expectedType, List(validator)), (TypedParameter(fieldName, expr), ()))
```
arkadius, line 294: Why not to just extract another method in `Validations` instead of creating fake parameter?

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+        .compileExpression(
+          expression,
+          ctx,
+          Unknown,
+          NodeExpressionId.DefaultExpressionId,
+          Some(OutputVar.switch(output))
+        )
+        ._2
+    }
+    val objExpression = expressionCompilation.map(_.compiledObject.map(Some(_))).getOrElse(Valid(None))
```
arkadius, line 86: It can be just `expressionCompilation.map(_.compiledObject).sequence`

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+    )
+
+    val compliedRecordFields = fields.zipWithIndex.map { case (field, index) =>
+      objectParametersExpressionCompiler
+        .compile(field.expression, Some(node.recordValueFieldName(index)), ctx, Unknown)
+        .map(CompiledRecordField(field, index, _))
+    }
+
+    val recordValuesCompilationResult = compliedRecordFields.traverse { validatedField =>
+      validatedField.andThen { compiledField =>
```
arkadius, line 137: It can be just `.map {` (without wrapping with `Valid`)

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+      validatedField.map { compiledField =>
+        RecordValidator.IndexedField(
+          compiledField.field.name,
+          compiledField.typedExpression,
+          compiledField.index
+        )
+      }
+    }
+    val additionalValidationResult = RecordValidator.validateRecord(
+      typedFieldsValidations.collect { case Valid(typedField) => typedField }
```
arkadius, line 158: So we validate uniqueness only when field's expression is valid. Why? Why not validate uniqueness always?

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+      parameters = None,
+      validationContext = outputVar.map(ctx.withVariable(_, typedObject)).getOrElse(Valid(ctx)),
+      compiledObject = compiledFields,
+      expressionType = Some(typedObject)
+    )
+
+    combineErrors(compilationResult, additionalValidationResult)
+  }
+
+  private object ExpressionCompilerAdapter {
```
arkadius, line 186: I don't exactly get what we adapt here? IMO making this `compileExpression` method just 1st level private method would decrease the noise.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+
+    case class ExpressionCompilation[R](
```
arkadius, line 188: This name is misleading. Sth is missing here. It is not a "process of compilation itself" but rather some result of compilation? Also, I see some redundancy here. For me, this class looks like a simple wrapper for `ValidatedNel[ProcessCompilationError, TypedExpression]` that additionally holds `fieldName`. If you put this `ValidatedNel` into `newValidated`, you can compute the current validated by `newValidated.map(_.expression)` and `typedExpression` by `newValidated.toOption`.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  private object ValidationAdapter {
+
+    def validateMaybeBoolean(
+        expression: Option[TypedExpression],
+        fieldName: String
+    )(
+        implicit nodeId: NodeId
+    ): Validated[NonEmptyList[PartSubGraphCompilationError], Unit] = {
```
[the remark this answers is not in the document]
mslabek: Done

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: done

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: I'm still not used to cats it seems :D Done

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: done

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: It was just easier to do both mandatory value and unique key validations at once, didn't think about it honestly. Corrected now at the cost of a little bit more complexity.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: Yes, it doesn't make much sense besides some code organisation. I wanted to make this code more readable (thats why I extracted this class from the NodeCompiler) and limit the scope at the same time - thats why there's ValidationAdapter, RecordValidator and this ExpressionCompilerAdapter. Do you think other private objects are ok or flattening this structure or doing something else would be better?

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: Yes, this is one of the remains from the previous implementation in NodeCompiler. I've reduced it to a simple wrapper and adapted remaining code.

designer/client/src/components/graph/node-modal/Variable.tsx
```diff
@@ -68,7 +68,7 @@ export default function Variable({
                 {renderFieldLabel("Variable Name")}
             </LabeledInput>
             <EditableEditor
-                fieldName="expression"
+                fieldName="$expression"
```
Dzuming, line 71: What is the reason for the change of the param name to the `$` prefix? Is it our new convention? 
As is see, we still have https://github.com/TouK/nussknacker/pull/4953/files#diff-cca035cb72539c4c5ea0a2f36b88c9a47476645295c68c0cb477d34c3720fc41L31 the field name without `$` prefix `$fields-${index}-key` (I mean key which is the field name).

docs/Changelog.md
```diff
@@ -27,6 +27,10 @@
 * [#5033](https://github.com/TouK/nussknacker/pull/5033) Updated Scala 2.13 to 2.13.12
 * [#4887](https://github.com/TouK/nussknacker/pull/4887) New parameter validator - `ValidationExpressionParameterValidator` that allows to use SpEL (or any other) expression to validate parameters
 * [#5077](https://github.com/TouK/nussknacker/pull/5077) Add an option to set schema on connections in SQL enricher
+* [#4953](https://github.com/TouK/nussknacker/pull/4953) Add stricter validation in base components:
```
arkadius, line 30: @coutoPL @pielas This changelog is ok for previous and further Realease Managers? Does it match "Added entry in Changelog.md describing the change from the perspective of a public distribution user" rule?

mslabek: Forgot to push that change - I propose a generic "Improved node validation" here in changelog and this extended description 

"Add stricter node validation:
  * Boolean expressions in Switch and Filter nodes now required not null values
  * Variable values in MapVariable, FragmentOutput and Variable are now mandatory
  * Field names in MapVariable, FragmentOutput are required to be unique"

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
arkadius: I see that you already reuse these patterns between record and other base components. IMO we should extract two new files: `RecordValidator` and `(Base)ComponentValidatorHelper`

designer/client/src/components/graph/node-modal/Variable.tsx
Dzuming: Ok, and what about this one `$fields-${index}-key`, shouldn't it be `$fields-${index}-$key`?

mslabek: I think this is the first case besides fragment parameters when were introducing nested fieldNames like this so there's no convention. For me the $ before nested name `key` could make sense if we were mapping these errors and after mapping one field would expect a fieldName `$key` but when we do it flatly like currently, I can't see the point, unless for consistency with other changes? WDYT?

Dzuming (approved):

designer/client/src/components/graph/node-modal/Variable.tsx
mslabek: Ok, thats reasonable, I'll update for consistency :)

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
```diff
@@ -0,0 +1,318 @@
...
+  ): NodeCompilationResult[expression.Expression] = {
+    val (expressionCompilation, nodeCompilation) =
+      compileExpression(
+        variable.value,
+        ctx,
+        expectedType = Unknown,
+        outputVar = Some(OutputVar.variable(variable.varName))
+      )
+
+    val additionalValidationResult: ValidatedNel[ProcessCompilationError, Unit] =
```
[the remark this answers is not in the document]
mslabek: Agreed, this one I forgot to remove :) Done

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/BaseNodeCompiler.scala
mslabek: Thanks for the advice, I've extracted these two files. Done

docs/Changelog.md
arkadius: LGTM

arkadius (approved):

PR #4979, opened 2023-11-07
Add close() method to HttpRemoteEnvironment

piotrp:
## Describe your changes

`HttpRemoteEnvironment` creates `HttpExt` in its constructor, which makes it bound to implicitly passed `ActorSystem`. Shutting down `ActorSystem` may cause an exception to be logged because Akka HTTP still depends on it. To prevent that we need to explicitly close its connection pool.

Currently we're using:
```scala
      try {
        // checkScenarioMigration(...)
      } finally {
        Await.ready(remoteEnv.http.shutdownAllConnectionPools(), 5.seconds)
      }
```
but it's better not to depend in internal implementation. 

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
@@ -3,7 +3,7 @@ package pl.touk.nussknacker.ui.process.migrate
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets
 import akka.actor.ActorSystem
-import akka.http.scaladsl.Http
+import akka.http.scaladsl.{Http, HttpExt}
 import akka.http.scaladsl.marshalling.Marshal
 import akka.http.scaladsl.model.Uri.{Path, Query}
 import akka.http.scaladsl.model.headers.{Authorization, BasicHttpCredentials, RawHeader}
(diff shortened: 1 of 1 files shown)


mslabek (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -103,6 +108,10 @@ class HttpRemoteEnvironment(
     )
   }
 
+  override def close(): Unit = Await.ready(http.shutdownAllConnectionPools(), closeTimeout)
+
+  @unused
```
coutoPL, line 113: is it needed? the annotation.
BTW, I thought that this annotation was introduced in Scala 2.13. But the cross-compile doesn't complain ... weird.

piotrp: Depends on how you see it. I added it because it's an API that I believe should be available here, but it's unused in Nussknacker. This annotation is meant exactly for this use case.

piotrp: I can remove it if you don't want it here

coutoPL: If the method is not used, do we need to have it? If the method can be used somewhere else, eg. in some Nu extension, IMO, the annotation is not needed.

piotrp: Done, annotation removed

coutoPL (approved):

PR #4985, opened 2023-11-08
Add createdAt as a second sort param to place newest at the beginning of lists

Filemon279:
Right now deployments are sorted by `runAt` and it might produce problems when in some specific cases you want to run instant batch without waiting for schedule (using custom actions). Then `runAt` is still set by CRON but deployment starts immediately. In such case we can produce > 1 records with `runAt` set to same value and NU cannot handle states properly.

## Checklist before merge


Filemon279:
Moved to https://github.com/TouK/nussknacker/pull/4992

PR #4992, opened 2023-11-10
Sort periodic process deployments not only by runAt but also by createdAt

Filemon279:
Right now deployments are sorted by `runAt` and it might produce problems when in some specific cases you want to run instant batch without waiting for schedule (using custom actions). Then `runAt` is still set by CRON but deployment starts immediately. In such case we can produce > 1 records with `runAt` set to same value and NU cannot handle states properly.

## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+1.12.4 (?? Oct 2023)
+* [#4992](https://github.com/TouK/nussknacker/pull/4992) Fix: Sort PeriodicProcessDeployments by `runAt` and `createdAt` and instead of `runAt` only.
+
 1.12.3 (26 Oct 2023)
 1.12.2 (25 Oct 2023)
 1.12.1 (25 Oct 2023)
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 1.12.4 (?? Oct 2023)
-* [#4992](https://github.com/TouK/nussknacker/pull/4992) Fix: Sort PeriodicProcessDeployments by `runAt` and `createdAt` and instead of `runAt` only.
+* [#4992](https://github.com/TouK/nussknacker/pull/4992) Fix: List of periodic deployments is now sorted not only by schedule time but also by its creation time.
 
 1.12.3 (26 Oct 2023)
 1.12.2 (25 Oct 2023)
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
@@ -571,7 +571,7 @@ object PeriodicProcessService {
         lastActiveDeploymentStatusForEachSchedule.find(_.status == status)
 
       def last(status: PeriodicProcessDeploymentStatus) =
-        lastActiveDeploymentStatusForEachSchedule.findLast(_.status == status)
+        lastActiveDeploymentStatusForEachSchedule.reverse.find(_.status == status)
 
       first(PeriodicProcessDeploymentStatus.Deployed)
         .orElse(last(PeriodicProcessDeploymentStatus.Failed))
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
@@ -419,8 +419,8 @@ class PeriodicProcessService(
         scheduleData.latestDeployments.map { deployment =>
           DeploymentStatus(
             deployment.id,
-            deployment.createdAt,
             scheduleId,
+            deployment.createdAt,
             deployment.runAt,
             deployment.state.status,
             scheduleData.process.active,
@@ -428,8 +428,7 @@ class PeriodicProcessService(
           )
         }
       }
-      .sortBy(_.runAtWithCreatedAt)
-      .reverse
+      .sorted(DeploymentStatus.ordering.reverse)
 
     for {
       activeSchedules <- getLatestDeploymentsForActiveSchedules(name, MaxDeploymentsStatus)
@@ -499,9 +498,7 @@ object PeriodicProcessService {
     def limitedAndSortedDeployments: List[DeploymentStatus] =
       (activeDeploymentsStatuses ++ inactiveDeploymentsStatuses.take(
         MaxDeploymentsStatus - activeDeploymentsStatuses.size
-      ))
-        .sortBy(_.runAtWithCreatedAt)
-        .reverse
+      )).sorted(DeploymentStatus.ordering.reverse)
 
     // We present merged name to be possible to filter scenario by status
     override def name: StatusName = mergedStatusDetails.status.name
@@ -568,13 +565,13 @@ object PeriodicProcessService {
       */
     def pickMostImportantActiveDeployment: Option[DeploymentStatus] = {
       val lastActiveDeploymentStatusForEachSchedule =
-        latestDeploymentForEachSchedule(activeDeploymentsStatuses)
-          .sortBy(_.runAtWithCreatedAt)
+        latestDeploymentForEachSchedule(activeDeploymentsStatuses).sorted
+
       def first(status: PeriodicProcessDeploymentStatus) =
         lastActiveDeploymentStatusForEachSchedule.find(_.status == status)
 
       def last(status: PeriodicProcessDeploymentStatus) =
-        lastActiveDeploymentStatusForEachSchedule.reverse.find(_.status == status)
+        lastActiveDeploymentStatusForEachSchedule.findLast(_.status == status)
 
       first(PeriodicProcessDeploymentStatus.Deployed)
         .orElse(last(PeriodicProcessDeploymentStatus.Failed))
@@ -589,7 +586,7 @@ object PeriodicProcessService {
         .groupBy(_.scheduleId)
         .values
         .toList
-        .map(_.sortBy(_.runAtWithCreatedAt).reverse.head)
+        .map(_.min(DeploymentStatus.ordering.reverse))
     }
 
   }
@@ -597,8 +594,8 @@ object PeriodicProcessService {
   case class DeploymentStatus( // Probably it is too much technical to present to users, but the only other alternative
       // to present to users is scheduleName+runAt
       deploymentId: PeriodicProcessDeploymentId,
-      createdAt: LocalDateTime,
       scheduleId: ScheduleId,
+      createdAt: LocalDateTime,
       runAt: LocalDateTime,
       // This status is almost fine but:
       // - we don't have cancel status - we have to check processActive as well (isCanceled)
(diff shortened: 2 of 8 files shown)


docs/Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+1.12.4 (?? Oct 2023)
+* [#4992](https://github.com/TouK/nussknacker/pull/4992) Fix: Sort PeriodicProcessDeployments by `runAt` and `createdAt` and instead of `runAt` only.
```
coutoPL, line 4: please edit it taking into consideration the users' perspective

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -427,7 +428,8 @@ class PeriodicProcessService(
           )
         }
       }
-      .sortBy(_.runAt)(Ordering[LocalDateTime].reverse)
+      .sortBy(_.runAtWithCreatedAt)
+      .reverse
```
coutoPL, line 432: it looks like the approach with "reversing" the ordering is a little bit efficient (we don't reverse the whole collection but sort with the final order.

Maybe let's create a private val reversedLocalDataTimeOrdering and reuse them all over the class? WDYT?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -566,7 +569,7 @@ object PeriodicProcessService {
     def pickMostImportantActiveDeployment: Option[DeploymentStatus] = {
       val lastActiveDeploymentStatusForEachSchedule =
         latestDeploymentForEachSchedule(activeDeploymentsStatuses)
-          .sortBy(_.runAt)(Ordering[LocalDateTime])
+          .sortBy(_.runAtWithCreatedAt)
       def first(status: PeriodicProcessDeploymentStatus) =
         lastActiveDeploymentStatusForEachSchedule.find(_.status == status)
```
coutoPL, line 575: maybe:
```
 def last(status: PeriodicProcessDeploymentStatus) =
        lastActiveDeploymentStatusForEachSchedule.findLast(_.status == status)
        ```

File: engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala [hunk 4]
```diff
@@ -566,7 +569,7 @@ object PeriodicProcessService {
     def pickMostImportantActiveDeployment: Option[DeploymentStatus] = {
       val lastActiveDeploymentStatusForEachSchedule =
         latestDeploymentForEachSchedule(activeDeploymentsStatuses)
-          .sortBy(_.runAt)(Ordering[LocalDateTime])
+          .sortBy(_.runAtWithCreatedAt)
```
coutoPL, line 572: it's already sorted. Do we need to sort it again? Or we may ask: how do we know it's sorted? Maybe the sorting inside the `latestDeploymentForEachSchedule` method is not needed if it's hard to express it differently than with the name (it can be done using types, but let's not take this into consideration now). 

So, I see two possible improvements here:
1. get rid of sorting inside the implementation of `latestDeploymentForEachSchedule` and sort in places where the function is used
2. change the name of the method to give a reader the hint that the output collection is sorted (and how). Then in this line, we could just reverse the collection.

IMO the 1st idea is better. WDYT? Maybe another idea?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -605,6 +609,8 @@ object PeriodicProcessService {
       runtimeStatusOpt: Option[StatusDetails]
   ) {
 
+    val runAtWithCreatedAt: (LocalDateTime, LocalDateTime) = (runAt, createdAt)
```
coutoPL, line 612: AFAIS we create this `val` only for the purpose of sorting in this service. WDYT about creating Ordering (or even better, the cats' Order), well-named object and use `.sort(ourWellNamedOrder)` instead of `.sortBy` and this artificial `val`. WDYT?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/model/PeriodicProcessDeployment.scala
```diff
@@ -9,6 +9,7 @@ import java.time.{Clock, LocalDateTime}
 // TODO: We should separate schedules concept from deployments - fully switch to ScheduleData and ScheduleDeploymentData
 case class PeriodicProcessDeployment(
     id: PeriodicProcessDeploymentId,
+    createdAt: LocalDateTime,
```
coutoPL, line 12: maybe it'd be better to move it closer to the `runAt` field? ;)

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -586,14 +589,15 @@ object PeriodicProcessService {
...
-        .map(_.sortBy(_.runAt)(Ordering[LocalDateTime].reverse).head)
+        .map(_.sortBy(_.runAtWithCreatedAt).reverse.head)
     }
 
   }
 
   case class DeploymentStatus( // Probably it is too much technical to present to users, but the only other alternative
       // to present to users is scheduleName+runAt
       deploymentId: PeriodicProcessDeploymentId,
+      createdAt: LocalDateTime,
```
coutoPL, line 600: maybe it'd be better to move it closer to the `runAt` field? ;)

coutoPL (changes requested):

docs/Changelog.md
Filemon279: Thanks, changed a little bit

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
Filemon279: That is true, I've changed it to use reversed ordering instead of reverse sort with standard ordering.

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
Filemon279: Hmm I am not sure it is fully sorted already. The sort operation inside `latestDeploymentForEachSchedule` is taking place inside `List[List[A]]` and it is taking place to get latest deployment for each schedule and produce `List[A]` which I don't think is sorted yet :thinking: 

As for me I found it quite intuitive to read and I wouldn't change it too much - still I am open to suggestions, I might be wrong :slightly_smiling_face:

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
Filemon279: That is true, thanks - I've created companion object with implicit ordering (I've used Ordering since we already use it inside this scope) and now we can just use `sorted` and for `reverse` use reverted ordering rather than `reverse` on sorted list. WDYT of that approach?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
Filemon279: We are going to face crossCompile problems if we do that :(

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
```diff
@@ -309,7 +310,7 @@ class SlickPeriodicProcessesRepository(
         (
           rowNumber() :: Over
             .partitionBy((deployment.periodicProcessId, deployment.scheduleName))
-            .sortBy(deployment.runAt.desc),
+            .sortBy(deployment.runAt.desc, deployment.createdAt.desc),
```
coutoPL, line 313: should we use `DeploymentStatus.ordering` here too?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
```diff
@@ -354,7 +355,7 @@ class SlickPeriodicProcessesRepository(
               .filter(deployment =>
                 deployment.periodicProcessId === process.id && (deployment.scheduleName === scheduleName || deployment.scheduleName.isEmpty && scheduleName.isEmpty)
               )
-              .sortBy(_.runAt.desc)
+              .sortBy(a => (a.runAt.desc, a.createdAt.desc))
```
coutoPL, line 358: and here?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
coutoPL: ok, so let's leave it

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
Filemon279: It is slick query - the `sortBy` gets `slick.lifted.Ordered`

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
Filemon279: same

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
coutoPL: oh, right. Didn't notice. Sorry. 
One more question: do we need to sort in the DB too? Do these two sorting methods (in the service and in the repository) are related?

Filemon279: I think it is since it came on tests in another project that is was required as well - still I am not 100% sure how exactly. I thought though, that since we are sorting here by `runAt` we should add `createdAt` as well since someone might use E.g. `getLatestDeploymentsForActiveSchedules` directly and it would be inconsistent with other methods.

coutoPL: so, maybe at least we should leave here a comment that it should be changed when `DeploymentStatus.ordering` is changed. It would be great to have a dynamic ordering here, but AFAIU this is not simple. The comment seems to be an acceptable compromise.

Filemon279: Good idea, I've added appropriate comment :+1:

coutoPL (approved):
LGTM!

Filemon279:
Thanks for review :)

PR #5006, opened 2023-11-14
Fixup to async DefaultAsyncExecutionConfigPreparer

raphaelsolarski:
Fixup to async DefaultAsyncExecutionConfigPreparer - using class loader release hook instead of counting prepareExecutionContext/close method calls:

* Closed ExecutorService could leak into execution after restart because lack of synchronization in close method (cached reference to Some in asyncExecutionContext var for thread)
* prepare/close counter could fall below zero and lead to odd behaviour in cases when close was called without a prior open method call (e.g. in case of operator initializeState error - look into org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.initializeStateAndOpenOperators and org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke)

## Checklist before merge

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/async/DefaultAsyncExecutionConfigPreparer.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/async/DefaultAsyncExecutionConfigPreparer.scala
@@ -1,20 +1,16 @@
 package pl.touk.nussknacker.engine.process.async
 
-import java.util.concurrent._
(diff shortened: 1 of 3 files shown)


raphaelsolarski:
I'm still thinking about enabling this mechanism to work in our tests, because unfortunately there is a slightly different classloader arrangement in them, so DefaultAsyncExecutionConfigPreparer is not loaded in FlinkUserCodeClassLoader but in AppClassLoader

raphaelsolarski:
fixed in https://github.com/TouK/nussknacker/pull/6055 and https://github.com/TouK/nussknacker/pull/6204

PR #5051, opened 2023-11-24
Add mode to run updates in query enricher

philemone:
## Checklist before merge

### DIFF
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/query/QueryResultStrategy.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/query/QueryResultStrategy.scala
@@ -1,6 +1,6 @@
 package pl.touk.nussknacker.sql.db.query
 
-import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
+import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypingResult}
 import pl.touk.nussknacker.sql.db.schema.TableDefinition
 
 object QueryResultStrategy {
@@ -8,6 +8,7 @@ object QueryResultStrategy {
   def apply(name: String): Option[QueryResultStrategy] = name match {
     case ResultSetStrategy.`name`    => Some(ResultSetStrategy)
     case SingleResultStrategy.`name` => Some(SingleResultStrategy)
+    case UpdateResultStrategy.`name` => Some(UpdateResultStrategy)
     case _                           => None
   }
 
@@ -20,6 +21,7 @@ sealed trait QueryResultStrategy {
     this match {
       case ResultSetStrategy    => tableDef.resultSetType
       case SingleResultStrategy => tableDef.rowType
+      case UpdateResultStrategy => Typed[UpdateResultStrategy.updateResultType]
     }
 
 }
@@ -30,4 +32,9 @@ case object ResultSetStrategy extends QueryResultStrategy {
 
 case object SingleResultStrategy extends QueryResultStrategy {
   val name: String = "Single result"
+}
+
+case object UpdateResultStrategy extends QueryResultStrategy {
+  type updateResultType = Integer
+  val name: String = "Update result"
 }
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/query/QueryResultStrategy.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/query/QueryResultStrategy.scala
@@ -36,5 +36,5 @@ case object SingleResultStrategy extends QueryResultStrategy {
 
 case object UpdateResultStrategy extends QueryResultStrategy {
   type updateResultType = Integer
-  val name: String = "Update result"
+  val name: String = "Updates count"
 }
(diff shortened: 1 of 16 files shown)


components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/JdbcMetaDataProvider.scala
```diff
@@ -32,7 +32,7 @@ class JdbcMetaDataProvider(getConnection: () => Connection) extends DbMetaDataPr
     Using.resource(getConnection()) { connection =>
       Using.resource(connection.prepareStatement(query)) { statement =>
         TableMetaData(
-          TableDefinition(statement.getMetaData),
+          Option(statement.getMetaData).fold(TableDefinition(Nil))(TableDefinition(_)),
```
arkadius, line 35: Does Nu is still functional when getMetaData returns null? Shouldn't we communicate it as an Error to user and fail fast instead of pretending that everything is ok?

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricher.scala
```diff
@@ -167,14 +171,13 @@ class DatabaseQueryEnricher(val dbPoolConfig: DBPoolConfig, val dbMetaDataProvid
     }
   }
 
-  // SyntaxErrorException should have message meaningful for users, for others (connection problem?) it's better
-  // to return generic message and log details...
   private def messageFromSQLException(query: String, sqlException: SQLException): String = sqlException match {
     case e: SQLSyntaxErrorException =>
       e.getMessage
-    case e =>
+    case e => // For some users full error msg can be significant. Allow to choose full error message in config
```
arkadius, line 177: And for which user it won't be significant? Why not to just enable it always? What concerns do you have around that?

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/query/QueryResultStrategy.scala
```diff
@@ -31,3 +33,8 @@ case object ResultSetStrategy extends QueryResultStrategy {
 case object SingleResultStrategy extends QueryResultStrategy {
   val name: String = "Single result"
 }
+
+case object UpdateResultStrategy extends QueryResultStrategy {
+  type updateResultType = Integer
+  val name: String = "Update result"
```
arkadius, line 39: I think that this description is meaningless to user. WDYT about renaming it to `Updates count` or `Updated records count` or `Affected records count`?

arkadius (commented):
Nice, some comments added

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/JdbcMetaDataProvider.scala
philemone: Yes, Nu is still functional :) What is more, for update queries there is no metadata passed through, so it would not work without this change

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricher.scala
philemone: Well, I wasn't bold enough to argue with previous comment which stated `[...] should have message meaningful for users, for others (connection problem?) it's better to return generic message and log details...`

But I personally think that it is better to show error to the user - Seeing that you have similar though I've removed flag and set error always to be shown :slightly_smiling_face:

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/JdbcMetaDataProvider.scala
arkadius: Ok, so why you don't return `Option[TableMetada]` and write a comment that for updates it is `None`? I see some hidden assumptions here :)

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricher.scala
arkadius: Ok, now I understand. In which situation have you occurred other problem than `SQLSyntaxErrorException`?

philemone: When you are using column which doesn't exists inside query - then in we would get e.g. `PSQLException`

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/JdbcMetaDataProvider.scala
philemone: I couldn't see them, sorry :) The idea with `Option` is fair enough and lead to better error handling inside db components. I've wrapped `TableMetada` in to `Option` and added cases to handle error when it is missing inside select query. For update queries it is a valid value.

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricher.scala
arkadius: Uh, so it's look like we have to choose a compromise between technical information leakage like:
- problems with authentication
- connection problems 

that wasn't configured by a user and  helpfulness of the message.

Let's:
- add a comment that we made this informed choice
- adds a logs before returning message to make sure that also admin would have an access to these information

arkadius (approved):
One comment added, please fix it before merge

components/sql/src/main/scala/pl/touk/nussknacker/sql/service/DatabaseQueryEnricher.scala
philemone: You are right - thanks, I've added both comment and log entry :+1:

PR #5081, opened 2023-11-28
[NU-1286] logging improvements in case of Nu Designer API errors

coutoPL:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NuDesignerErrorToHttp.scala
```diff
@@ -32,21 +34,42 @@ object NuDesignerErrorToHttp extends LazyLogging with FailFastCirceSupport {
...
+
+  private def logError(error: NuDesignerError): Unit = error match {
+    case error: NotFoundError =>
+      logger.debug(s"Not found error: ${error.getMessage}. ${returnedHttpStatusInfo(error)}", error)
+    case error: FatalError =>
+      logger.error(s"Fatal error: ${error.getMessage}. ${returnedHttpStatusInfo(error)}", error)
+    case error: BadRequestError =>
+      logger.debug(s"Bad request error: ${error.getMessage}. ${returnedHttpStatusInfo(error)}", error)
+    case error: IllegalOperationError =>
+      logger.error(
```
arkadius, line 55: This "error" appears e.g. when smb invokes an action on a scenario and the scenario is in a state that doesn't allow this operation.   It is a kind of problem on the client side. We don't have control on the client side, so the client can generate a flood of error logs. I recommend using debug in this place instead.

arkadius (approved):
One minor comment added

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NuDesignerErrorToHttp.scala
coutoPL: After discussion, I've decided to modify FatalError and OtherError too, to be consistent with the described above strategy.

PR #5153, opened 2023-12-04
[NU-1291] ProcessingTypeDataProvider - caching transformed processing type data and combined data

arkadius:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
```diff
@@ -34,75 +32,126 @@ trait ProcessingTypeDataProvider[+T, +C] {
...
+    stateValue
   }
 
-  def mapCombined[CC](fun: (=> C) => CC): ProcessingTypeDataProvider[T, CC] = {
+}
+
+object ProcessingTypeDataProvider {
 
-    new ProcessingTypeDataProvider[T, CC] {
+  val noCombinedDataFun: () => Nothing = () =>
```
DeamonDev, line 83: I would change this first occurrence of `()` to `Unit` in order to distinguish the `Unit` datatype from its only value which is `()`.

arkadius: But in this context, `()` is not a Unit, but empty parameters list :) It's just a function!

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
```diff
@@ -24,8 +24,6 @@ import pl.touk.nussknacker.ui.security.api.LoggedUser
  */
 trait ProcessingTypeDataProvider[+T, +C] {
```
coutoPL, line 25: it looks like the name is too generic. I can find in the interface references to ProcessingType but I don't know what is the `Data` in this context. Moreover, we can find references to LoggedUser in the methods. Why? It's not obvious how T refers to ProcessingType and LoggedUser.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
```diff
@@ -34,75 +32,126 @@ trait ProcessingTypeDataProvider[+T, +C] {
...
+    case (k, ValueWithPermission(v, AnyUser)) => (k, Some(v))
+    case (k, ValueWithPermission(v, UserWithAccessRightsToCategory(category))) if user.can(category, Permission.Read) =>
+      (k, Some(v))
+    case (k, _) => (k, None)
+  }
 
   // TODO: We should return a generic type that can produce views for users with access rights to certain categories only.
   //       Thanks to that we will be sure that no sensitive data leak
-  def combined: C
+  def combined: C = state.getCombined()
```
coutoPL, line 52: it's hard to figure out what the "combined: C" is.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
```diff
@@ -34,75 +32,126 @@ trait ProcessingTypeDataProvider[+T, +C] {
...
+        allValues
+      )
+
+    }
 
 }
 
-class MapBasedProcessingTypeDataProvider[T, C](map: Map[ProcessingType, ValueWithPermission[T]], getCombined: => C)
-    extends ProcessingTypeDataProvider[T, C] {
+trait ProcessingTypeDataState[+T, +C] {
```
coutoPL, line 122: IMO this name is also too generic. We have `ValueWithPermission` in the interface and it's not obvious for me why it's there and how it relates to the State.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataReload.scala
```diff
@@ -29,46 +16,31 @@ trait Initialization {
...
-
-      override def combined: CombinedProcessingTypeData = reloader.current.combined
-    }
-    val lazyInitializedReloader = new ProcessingTypeDataReload with Initialization {
-      override def init(): Unit = reloader
-
-      override def reloadAll(): Unit = reloader.reloadAll()
-    }
-    (provider, lazyInitializedReloader)
+  @volatile private[processingtypedata] var state
```
coutoPL, line 24: maybe we can use AtomicReference instead of volatile?

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/UIProcessValidatorSpec.scala
```diff
@@ -1044,28 +1015,26 @@ private object UIProcessValidatorSpec {
...
-          None,
-          None,
-          Some(List(CompileTimeEvaluableValueValidator)),
-          Some("label")
-        )
-      ) ++ FlinkStreamingPropertiesConfig.properties)
-    )
+    Map(
+      "requiredStringProperty" -> ScenarioPropertyConfig(
+        None,
```
coutoPL, line 1020: let's use named arguments for a better readablity

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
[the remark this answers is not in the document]
arkadius: I'll comment on it in comment next to `ProcessingTypeDataProvider`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
arkadius: `ProcessingType` is a kind of `ApplicationContext` - similar to this one from `Spring`. It could be said that we have "applications inside application". One `ProcessingType` can't see data from other `ProcessingType's`. Services run inside one `ProcessingType` scope behave differently from services run in another scope.

`Data` is the generic name here by purpose. The purpose of this interface was to make accessible any data, hidden by this interface. You are starting from the whole application context which is `ProcessingTypeData` class (probably name should contain "Root"). Then you want to pass somewhere only a piece of information inside it, so you do some transformation, the same for each context.

- `T` is these `Data` available for certain `ProcessingType`
- `C` is `CombinedData`. We have a limited set of data which is a kind of intersection between all `ProcessingTypes`. We can discuss this further because it is a longer story.

We have `LoggedUser` because our `ProcessingTypes` are associated with `Category` e.g. Fraud Detection, Marketing. Given user has access to certain categories - see `LoggedUser.can`. `ProcessingTypeDataProvider` is a kind of gateway to all around `ProcessingType` so it is the first place where we can authorize user. Almost every endpoint is inside `ProcessingType`, and in each of them we at the beginning authorize user access to given `ProcessingType`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
arkadius: What exactly do you propose? Should I describe why we combine something in a comment?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
arkadius: With regards to `ValueWithPermission`: Due to the fact, that `ProcessingType's` concept is deeply rooted in our application, you can see it in many tests. Because of that, we use `ProcessingTypeDataProvider.apply(Map)` instead of creating reloadable `ProcessingTypeDataReload`, and we specify that any user will have an access to given `ProcessingType` Data, not only those with `AccessRightsToCategory`. Also, we wanted to have a backward compatibility to previous `categoriesConfig` (see `LegacyConfigProcessCategoryService` and notice next to `CategoryConfig`), so I decided to use `AnyUser`access also for the legacy config. It would be difficult to check categories access using the old format, because `ProcessCategoryService` is inside `CombinedProcessingTypeData`. We can discuss it offline. Maybe we should break compatibility and leave `AnyUser` access only for tests purpose?

With regards to `ProcessingTypeDataState` name. I have no cloud, how to call it. I need it to provide caching. It keeps a state (value) that is cached and restricted by `ProcessingTypeDataProvider`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataReload.scala
coutoPL: `synchronized` blocks queued threads, so `AtomicReference` is better in this aspect. But at the same time, there are no effects in the `stateValue`, so the blocking is good enough.

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/ScenarioPropertiesValidatorTest.scala
```diff
@@ -28,27 +28,25 @@ class ScenarioPropertiesValidatorTest extends AnyFunSuite with Matchers {
...
-        optionalFieldName -> ScenarioPropertyConfig(None, Some(StringParameterEditor), None, Some(label)),
-        optFixedFieldName -> ScenarioPropertyConfig(
-          None,
-          Some(FixedValuesParameterEditor(possibleValues)),
-          Some(List(FixedValuesValidator(possibleValues))),
-          Some(label)
-        )
+    Map(
+      reqFieldName -> ScenarioPropertyConfig(
+        None,
```
coutoPL, line 33: here (in this class) we could use named parameters too

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
coutoPL: I think I didn't understand what is `C` before. WDYT about explaining in the `ProcessingTypeDataProvider` trait comment relations between `ProcessingType`, `LoggedUser`, `Data` and `CombinedData`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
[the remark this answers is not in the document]
arkadius: Sure, I forgot about that. Thank you for catching it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
arkadius: 1. Done
2. Done
3. I'd rather don't want to break this convention. We can discuss the convention around implicit parameters in the whole project separately.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataReload.scala
arkadius: Yeah but on the other `AtomicReference`'s related operations can be processed many times so we should be sure that our transformation has no side effects. In the reload case it unfortunately has side effects.

coutoPL: oh, I missed the `loadMethod` and its side-effect. So, synchronized FTW here ;)

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/ScenarioPropertiesValidatorTest.scala
arkadius: Done

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceSpec.scala
```diff
@@ -78,14 +79,18 @@ class DeploymentServiceSpec
...
   private val processingTypeDataProvider: ProcessingTypeDataProvider[DeploymentManager, Nothing] =
     new ProcessingTypeDataProvider[DeploymentManager, Nothing] {
-      override def forType(processingType: ProcessingType)(implicit user: LoggedUser): Option[DeploymentManager] =
-        all.get(processingType)
 
-      override def all(implicit user: LoggedUser): Map[ProcessingType, DeploymentManager] = Map(
-        TestProcessingTypes.Streaming -> deploymentManager
-      )
+      override val state: ProcessingTypeDataState[DeploymentManager, Nothing] =
+        new ProcessingTypeDataState[DeploymentManager, Nothing] {
```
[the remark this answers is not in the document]
coutoPL: ah, the cursed var ;) TBH we could use e.g. `Supplier[DeploymentManager]` instead of using var. To loosen this coupling. But it's up to you. In the end, it's not such a big deal.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtypedata/ProcessingTypeDataProvider.scala
coutoPL: Great comment. It is much easier to grasp the sense of the trait now.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceSpec.scala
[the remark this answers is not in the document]
arkadius: I'll leave it as it is. It would be complicated with this `Supplier`. We'd have to change provided type to `DeploymentService`

PR #5209, opened 2023-12-08
[NU-1291] ModelDefinition using ComponentType instead passing around Maps for each ComponentType

arkadius:
## Describe your changes
- Mostly this change takes advantage of `ComponentType` in definition preparation. `ModelDefinition` became a simple wrapper for a `Map[ComponentInfo, SomeComponentDefinition]`. It reduced the amount of boilerplate 
- Thanks to that `ModelDefinitionExtractor` now is reduced to two steps: `ModelDefinitionFromConfigCreatorExtractor.extractModelDefinition` ++ `ComponentsFromProvidersExtractor.extractComponents`
- Thanks to that we are checking component duplicates not only across `ComponentProvider`s but also in the whole model with `ConfigCreator` - it blew up tests in many places
- To make test clean-up easier behaviour of `LocalModelData` was also change - now it doesn't discover components from `ComponentProviders` which was harmful in many situations. Instead, it takes `List[ComponentDefinition]`.
It had an influence also on testkit which uses `LocalModelData` under the hood.
- As a side effect, testkit doesn't wrap test components in in a fake `ComponentProvider` anymore
- Teskits's predefined set of components was changed - it is described in the `MigrationGuide`
- Teskits now uses default model with their global variables instead of empty global variables
- Some `ConfigCreator` usage in tests were translated into `List[ComponentDefinition]`. It made us closer to `ConfigCreator` removal, but still a lot of work is ahead
- `ComponentsUiConfig` become a value class
- `DefaultNamespacedObjectNaming` was moved into `interpreter`. lighter `ObjectNaming.OriginalNames` was added to `components-api`
- Now component's  UI configuration is respected not only for components created by `ConfigCreator` but also by `ComponentProvider`.
- Filtering by `!disable` works also for global variables now

## Checklist before merge


extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
```diff
@@ -2,9 +2,11 @@ package pl.touk.nussknacker.engine.api.process
 
 import pl.touk.nussknacker.engine.api.component.{ComponentId, SingleComponentConfig}
 
-// TODO: rename it? its no longer just a value with categories
+// TODO: This is deprecated API, remove it after ConfiCreator API will be removed
 case class WithCategories[+T](value: T, categories: Option[List[String]], componentConfig: SingleComponentConfig) {
 
+  def availableForCategory(category: String): Boolean = categories.isEmpty || categories.exists(_.contains(category))
```
mslabek, line 8: This can be removed - its implemented locally in one place that uses a method like this

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/FlinkBaseComponentProvider.scala
```diff
@@ -19,10 +19,23 @@ class FlinkBaseComponentProvider extends ComponentProvider {
...
+  }
+
+  override def isCompatible(version: NussknackerVersion): Boolean = true
+
+  override def isAutoLoaded: Boolean = true
+}
+
+object FlinkBaseComponentProvider {
+
+  def Components: List[ComponentDefinition] =
```
mslabek, line 34: Shouldn't this be a val?

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
```diff
@@ -105,14 +104,6 @@ class DevProcessConfigCreator extends ProcessConfigCreator {
           processObjectDependencies,
           FlinkKafkaAvroSinkImplFactory
         )
-      ),
-      "kafka" -> all(
```
mslabek, line 109: Is it redundant?

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentStaticDefinition.scala
```diff
@@ -1,19 +1,36 @@
...
+) extends BaseComponentDefinition {
 
   def withComponentConfig(componentConfig: SingleComponentConfig): ComponentStaticDefinition =
     copy(componentConfig = componentConfig)
 
   val hasNoReturn: Boolean = returnType.isEmpty
 
 }
+
+sealed trait ComponentTypeSpecificData
```
mslabek, line 24: Can't this be an Option[CustomComponentSpecificData]?

mslabek (approved):
Small suggestions, nothing blocking, LGTM!

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/model/ModelDefinition.scala
```diff
@@ -1,111 +1,81 @@
...
-      processingType: String
-  ): List[(ComponentIdWithName, T)] =
-    sinkFactories.map { case (name, obj) =>
-      val id = componentIdProvider.createComponentId(
-        processingType,
-        ComponentInfo(ComponentType.Sink, name)
-      )
-      ComponentIdWithName(id, name) -> obj
-    }.toList
+  def addComponents(componentsToAdd: List[(String, T)]): ModelDefinition[T] = {
```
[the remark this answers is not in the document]
arkadius: Yeah, I wasn't sure how to call it. I though that maybe `withComponents` will be interpreted as a `copy(components = newComponents)` which will introduce another confusion, but maybe it smaller evil. I'll change it to `withComponents`

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/WithCategories.scala
arkadius: Good catch

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/FlinkBaseComponentProvider.scala
arkadius: Yes, it should, I'll change it

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/DevProcessConfigCreator.scala
arkadius: Before this change, duplicates were checked only across `ComponentProvider`s. Components from ConfigCreators wasn't considered. It caused the silent shadowing of components. I fixed that, some tests shown me that we had an invalid configuration so I fixed it. These components are already delivered by `kafka` `ConfigProvider` which is already in `classPath` in configuration.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentStaticDefinition.scala
arkadius: I'm not sure about that. I was considering even adding additional `*ComponentSpecificData` which will extends `NoComponentTypeSpecificData` and have some assertion that each `ComponentType` has related `ComponentSpecificData`. Thanks to that we will be sure that if we check `componentType` before, e.g. `asCustomComponentData` won't throw an `Exception`. `Option[ComponentTypeSpecificData]` could be interpreted by someone as a data that we don't know yet and I intended it to be a kind of inheritance but used by composition. Is it worth it? WDYT?

mslabek (approved):

PR #5216, opened 2023-12-11
Nu 971 remove frontend validators

Dzuming:
## Checklist before merge


designer/client/src/components/modals/ValidationLabels.tsx
```diff
@@ -21,35 +23,20 @@ const LabelsContainer = styled("div")({
...
     // It's possible that expression is valid and it's type is known, but a different type is expected.
     return (
         <LabelsContainer>
-            {isValid ? (
+            {isEmpty(fieldErrors) ? (
                 <LimitedValidationLabel title={validationLabelInfo}>{validationLabelInfo}</LimitedValidationLabel>
             ) : (
-                renderErrorLabels()
+                fieldErrors.map((fieldErrors) => (
+                    <LimitedValidationLabel key={fieldErrors.message} title={fieldErrors.message} type="ERROR">
```
mslabek, line 36: Now that we are display all errors coming from BE, I don't think we can use message as key. When typing an expression that returns two exactly same errors like `a + a` (without having a in context), then removing `a`, then adding it again we get 3 errors. React throws `Uncaught Error: Children cannot be added or removed during a reorder operation.`

Can we generate ids on FE without messing up ordering of errors?

designer/client/src/types/node.ts
```diff
@@ -53,7 +53,7 @@ export type Field = Parameter | FragmentInputParameter;
 export interface Parameter {
     uuid: string;
     name: string;
-    expression: Expression;
+    expression?: Expression; //TODO: Check if we need this type
```
mslabek, line 56: In what situations is this null?

designer/client/src/components/AddProcessDialog.tsx
```diff
@@ -6,62 +6,46 @@ import { useProcessNameValidators } from "../containers/hooks/useProcessNameVali
...
         }
-    }, [isFragment, isValid, navigate, passProps, value]);
+    }, [isFragment, fieldErrors, navigate, passProps, value]);
 
-    const { t } = useTranslation();
     const buttons: WindowButtonProps[] = useMemo(
         () => [
             { title: t("dialog.button.cancel", "Cancel"), action: () => passProps.close() },
-            { title: t("dialog.button.create", "create"), action: () => createProcess(), disabled: !isValid },
+            { title: t("dialog.button.create", "create"), action: () => createProcess(), disabled: Boolean(fieldErrors) },
```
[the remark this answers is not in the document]
Dzuming: You are right, fixed ✔️

designer/client/src/types/node.ts
Dzuming: I think it's something redundant, I checked and there is no use in removing it. Reverted ✔️

designer/client/src/components/modals/ValidationLabels.tsx
Dzuming: Nice catch! I didn't expect the messages to be the same, and I ignored this key (I just copied a preview approach). I just changed the key to the index, which resolves the uniqueness issue.

designer/client/src/components/AddProcessDialog.tsx
```diff
@@ -6,62 +6,46 @@ import { useProcessNameValidators } from "../containers/hooks/useProcessNameVali
...
-                navigate(visualizationUrl(processId));
-            } catch (error) {
-                if (error?.response?.status == 400) {
-                    //TODO: change to pass error from BE as whole object not just the message
-                    setProcessNameError({ fieldName: "processName", message: error?.response?.data, description: "", typ: "" });
-                } else {
-                    throw error;
-                }
-            }
+            await HttpService.createProcess(processId, processCategory, isFragment);
```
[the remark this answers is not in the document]
Dzuming: I removed them intentionally, because my assumption was, that it would just duplicate frontend errors. Now we check the uniqueness and mandatory of the field name. Are there more cases on the backend to return errors?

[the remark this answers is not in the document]
mslabek: BE additionally validates scenario/fragment id checking for: trailing, leading spaces, blank values and we have engine-specific custom validations like for flink - returning `Only digits, letters, underscore (_), hyphen (-) and space in the middle are allowed`.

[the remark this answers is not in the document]
Dzuming: Thanks for the explanation, changed ✔️

mslabek (approved):

JulianWielga (approved):

designer/client/src/types/definition.ts
```diff
@@ -42,7 +42,6 @@ export interface UIParameter {
     name: string;
     typ: TypingResult;
     editor: $TodoType;
-    validators: $TodoType;
```
arkadius, line 45: `validators` are also in `ParameterConfig`. We need to remove them as well

Dzuming: Good catch, Removed ✔️

PR #5219, opened 2023-12-11
[NU-1227] Notifications api endpoint with tapir

ForrestFairy:
## Describe your changes
Replaced `NotificationResources` with `NotificationApiEndpoints` and `NotificationApiHttpService`.

The structure of notifications is based on `Notification.scala`. The `JsonCodec` in Notification file is replaced with `derevo(encoder, decoder, schema)` to introduce `Schema` for `Notification`. Also a `Schema` for `ProcessName` is introduced inside `Notification` file.

Query parameter `'after'` is removed - it isn't used anywhere and we don't want to support it further.

Can look in future into `getUserActionsAfter` in `ProcessActionRepository.scala` if it used anywhere else than in `NotificationService` as it was used with `'after'` parameter which is removed

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/services/NotificationApiHttpService.scala
```diff
@@ -0,0 +1,37 @@
...
+    notificationService: NotificationService
+)(implicit executionContext: ExecutionContext)
+    extends BaseHttpService(config, getProcessCategoryService, authenticator)
+    with LazyLogging {
+
+  private val notificationApiEndpoints = new NotificationApiEndpoints(authenticator.authenticationMethod())
+
+  expose {
+//    Was before ->  Symbol("after").as[Instant].optional
+    val notificationsAfter = Some(Instant.now())
```
coutoPL, line 26: in akka-http-based implementation we have:
```scala
parameter(Symbol("after").as[Instant].optional)
```

It's a declaration of an optional query parameter decoded as Instant. 

It seems that in Tapir-based implementation we should use sth like that:
```scala
query[Option[Instant]]("after")
```
in endpoint declaration.

Make sure that we use the same format for decoding from String to Instant (probably it's ISO 8601). We should mention that in Open API doc too.

coutoPL (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NotificationApiEndpoints.scala
```diff
@@ -0,0 +1,29 @@
...
+
+class NotificationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpointDefinitions {
+
+  lazy val notificationEndpoint: SecuredEndpoint[Option[Instant], Unit, List[Notification], Any] =
+    baseNuApiEndpoint
+      .summary("Endpoint to display notifications")
+      .tag("Notifications")
+      .get
+      .in("notifications")
+      .in(query[Option[Instant]]("after").example(Some(Instant.parse("2023-07-29T19:30:00Z"))))
```
coutoPL, line 21: it would be nice to add tests for `after` != None:
1. with date-time passed in UTC (like in the example)
2. with date-time passed in eg. our local date time
3. with date-time invalid format

coutoPL (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NotificationApiEndpoints.scala
ForrestFairy: `after` parameter is removed altogether as it is not used anywhere

coutoPL (approved):
LGTM

PR #5234, opened 2023-12-13
OpenAPI enricher: Security configuration documentation clarification.

arkadius:
## Checklist before merge

### DIFF
--- a/docs/integration/OpenAPI.md
+++ b/docs/integration/OpenAPI.md
@@ -5,17 +5,18 @@ sidebar_position: 3
 # OpenAPI enrichers
 
 ## Overview
-                              
+
 Nussknacker can use services documented with OpenAPI specification.
-We use Swagger to parse OpenAPI, versions 2.x and 3.x are supported 
+We use Swagger to parse OpenAPI, versions 2.x and 3.x are supported
 (version 2.x should be considered as deprecated).
 
 Nussknacker applies following rules when mapping OpenAPI services to enrichers:
+
 - Each operation is mapped to one enricher
 - If operationId is configured it's used as enricher name, otherwise we create it as concatenation
-of HTTP method, path and parameters (to provide some level of uniqueness)
+  of HTTP method, path and parameters (to provide some level of uniqueness)
 - Parameters (path, query of header) are used to define enricher parameters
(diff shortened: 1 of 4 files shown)


docs/integration/OpenAPI.md
```diff
@@ -64,20 +67,26 @@ components {
...
+| url                    | true     |         | URL of the [*OpenAPI document*](https://swagger.io/specification/v3/). It contains specification of the service you want to interact with.                |
+| rootUrl                | false    |         | The URL of the service. If not specified, the URL of the service is taken from the *OpenAPI document*.                                                    |
+| allowedMethods         | false    | ["GET"] | Usually only GET services should be used as enrichers are meant to be idempotent and not change data                                                      |
+| namePattern            | false    | .*      | Regexp for filtering operations by operationId (i.e. enricher name)                                                                                       |
+| security               | false    |         | Configuration for [authentication](https://swagger.io/docs/specification/authentication/) for each `securitySchemas` defined in the OpenAPI configuration |
```
ZbyszekMM, line 76: The term "OpenAPI configuration" used here may be misleading, as we used term "OpenAPI document" in lines 72 and 73. 

Swagger people did not come up with a good term for the artifact which contains a definition of the interface. The document referenced in line 72 uses term "OpenAPI document" for this. The document you link to in line 76 uses term "OpenAPI definition". 
Term "OpenAPI specification" sounds like a definition of an OpenAPI standard, not a specification of your particular interface or set of interfaces.
Term "OpenAPI definition" also sounds to me like a definition of an OpenAPI standard
Term "OpenAPI configuration" - probably does not fit at all. 

Looking at all this terminology mess (something that really confused me), for me the preferred terms to use would be to use  _Interface specification_ or _API specification_.

The remaining candidates have some shortcomings.  
2. OpenAPI file (used more often than OpenAPI document, does not sound like a definition)
3. OpenAPI document (seems not to be used widely, sounds like a definition)
4. OpenAPI interface definition (a bit too long, but OK)

arkadius: I like the `OpenAPI interface definition`. We already use the `definition` in URLs where we expose our req-resp definition. I changed it (I hope) in all places where was used other term in the same context.

PR #4978, opened 2023-11-06
Expand FragmentParameter with hintText, initialValue, required, inputMode

wrzontek:
## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [x] Code is cleaned from temporary changes and commented out lines
- [x] Parts of the code that are not easy to understand are documented in the code
- [x] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [x] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [x] Verify that PR will be squashed during merge
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change

PR #5241, opened 2023-12-14
FragmentParameter valueCompileTimeValidation

wrzontek:
## Checklist before merge


interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/fragment/FragmentDefinitionExtractor.scala
```diff
@@ -143,6 +140,19 @@ class FragmentComponentDefinitionExtractor(
         )
       }
       .getOrElse(EditorExtractor.extract(parameterData, config))
+
+    val customExpressionValidator = fragmentParameter.valueCompileTimeValidation.flatMap(expr => {
+      expressionCompiler
+        .compileWithoutContextValidation(expr.validationExpression, fragmentParameter.name, Typed[Boolean])
```
bartektartanus, line 146: `fragmentParameter.name` or `#value`?

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
```diff
@@ -35,27 +38,66 @@ class FragmentParameterValidator(
...
-    ).sequence.map(_ => ())
+      fixedExpressionsValidationResult,
+      validationExpressionResult
+    ).sequence.map(_.flatten)
+  }
+
+  private def compileExpressionValidator(
+      fragmentParameter: FragmentParameter,
+      typ: TypingResult
+  )(implicit nodeId: NodeId) = fragmentParameter.valueCompileTimeValidation match {
```
arkadius, line 55: Let's use just `fragmentParameter.valueCompileTimeValidation.flatMap(...).sequence` instead of pattern matching

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -136,30 +135,15 @@ class NodeCompiler(
...
-        )
-      }
+    case frag @ FragmentInputDefinition(id, _, _) =>
+      val parameterDefinitions =
+        fragmentDefinitionExtractor.extractParametersDefinition(frag, contextWithOnlyGlobalVariables)
+      val variables: Map[String, TypingResult] = parameterDefinitions.value.map(a => a.name -> a.typ).toMap
+      val validationContext                    = contextWithOnlyGlobalVariables.copy(localVariables = variables)
 
-      definitions.getComponent(ComponentType.Source, id) match {
+      val compilationResult = definitions.getComponent(ComponentType.Source, id) match {
```
arkadius, line 144: We need to understand this part :) 
1. Why this code is looking for some componen't definition? Shouldn't be always `None` here? 
2. `StubbedFragmentInputTestSource` takes `fragmentDefinitionExtractor` which looks weird. We already have `parameterDefinitions`.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
```diff
@@ -35,27 +38,77 @@ class FragmentParameterValidator(
...
+    case Some(expr) =>
+      if (expr.validationExpression.expression.isBlank) {
```
arkadius, line 57: Instead of checking `isBlank`, check if `typedExpression.expression` is an instance of `NullExpression`. Thanks to that it will work better with other expression languages and we will keep this `isBlank` logic in one place

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
arkadius: 1. I understand that the case with `None` handling is for ad-hoc tests but why this code also do something with Some(...) case? When it happens?

arkadius: @philemone can you help us to understand your code? :)

wrzontek: i think I don't understand your question, the Some(...) case is for ad-hoc testing, the None is the "normal behaviour", when just compiling a fragment

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
arkadius: I can't find my comment about the type, but let's use `Option` instead of `List`  - I see only one optional validator returned by this method

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
arkadius: In "normal behaviour" you don't have any Source with id = fragment id in definition. Why would you have it? definitions are something that you can use when you referring to something e.g. you use source component "kafka" in your nodes so you have to check the definition to figure out if you are using it correctly.

Here we are compiling fragment so we can't take a look at where is this fragment definition because we are just trying to figure out this definition :)

arkadius: ahh I read it again. You're right!

arkadius (approved):

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
philemone: I'm glad you worked that out :slightly_smiling_face:

docs/Changelog.md
```diff
@@ -42,6 +42,7 @@
 * 1.12.5 (1 Dec 2023)
 ------------------------
 * [#5110](https://github.com/TouK/nussknacker/pull/5110) Fix: The compare option doesn't display fragment input properties between the two versions
+* [#5241](https://github.com/TouK/nussknacker/pull/5241) FragmentParameter uses `ValidationExpressionParameterValidator` through new field: `valueCompileTimeValidation`
```
arkadius, line 45: It won't be added to 1.12, we almost have 1.13! :) Please bundle it with `Expand FragmentParameter`. TBH both of them don't look like comments for a human being. Can we do sth with this?

wrzontek: i've changed and batched  it with  changelog of https://github.com/TouK/nussknacker/pull/4978

(the 1.12.5 was barely visible because of the *, fixed it)

PR #5245, opened 2023-12-14
[NU-1330] REST API - validators removal from scenario properties config

mslabek:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -224,7 +218,6 @@ object UIProcessObjectsFactory {
         }
         .toMap
 
-    val transformed = modelDefinition.transform(createUIComponentDef)
```
mslabek, line 227: This was unused. The `transform` function was used only in one test so I've removed it, unless it's needed for convenience?

arkadius (approved):

PR #5256, opened 2023-12-19
[NU-1042] node windows history fixed

JulianWielga:
## Checklist before merge


designer/client/src/components/graph/node-modal/node/FragmentContent.tsx
```diff
@@ -1,39 +1,45 @@
...
+import { usePromiseEffect } from "../../../../containers/hooks/usePromiseEffect";
 
 export function FragmentContent({ nodeToDisplay }: { nodeToDisplay: FragmentNodeType }): JSX.Element {
     const processCounts = useSelector(getProcessCounts);
     const processDefinitionData = useSelector(getProcessDefinitionData);
 
     const [fragmentContent, setFragmentContent] = useState<Process>(null);
+    const { openFragmentNodes } = useModalDetailsIfNeeded();
+
+    usePromiseEffect(
```
Dzuming, line 21: If I understand correctly what we want to achieve here is:
- set process details
- open fragment node and node of this fragment if it exists. I'm not sure if `openFragmentNodes` describes it correctly. To me, it sounds like there are multiple fragments to open. Maybe, something like openFragmentAndHisNode will be a better name? 
- Do it only once on init? Not sure about this one, but I don't see other cases

WDYT about changing this usePromiseEffect additional abstraction to something like

```
  const processDetailsInitialized = useRef(false);

    useEffect(() => {
        if (fragmentContent) return;
        if (!NodeUtils.nodeIsFragment(nodeToDisplay)) return;
        if (processDetailsInitialized.current) return;

        const handleFragmentContent = async () => {
            const id = nodeToDisplay?.ref.id;
            const { data } = await HttpService.fetchProcessDetails(id);
            const fetchedProcessDetails = correctFetchedDetails(data, processDefinitionData);
            setFragmentContent(fetchedProcessDetails.json);
            return fetchedProcessDetails;
        };

        handleFragmentContent().then((fetchedProcessDetails) => {
            openFragmentNodes(fetchedProcessDetails.json);
        });

        processDetailsInitialized.current = true;
    }, [fragmentContent, nodeToDisplay, openFragmentNodes, processDefinitionData]);
```
?

We will remove additional usePromiseEffect abstraction, which  according to the name sound more like https://gist.github.com/finom/f97a6a4f344690a2486668bb3e207e38

designer/client/src/components/graph/types.ts
```diff
@@ -13,7 +13,7 @@ import {
 import { Capabilities } from "../../reducers/selectors/other";
 import { ProcessType } from "../Process/types";
 
-export interface GraphProps {
+type ScenarioGraphProps = {
```
Dzuming, line 16: Nice 👏

designer/client/src/containers/hooks/useSearchQuery.tsx
```diff
@@ -1,26 +1,71 @@
...
+    }, [parsed]);
+
+    return [parsed, updateQuery];
+}
+
+export function replaceSearchQuery(params: QueryRecord): void {
+    const { history, location } = window;
+    const currentState = history.state || {};
+    const currentURL = new URL(location.href);
+    currentURL.search = stringifyQuery(params);
```
Dzuming, line 51: Is it the reason to not use native UrlSearchParams here?

```
 const currentURL = new URL(location.href);
  const searchParams = new URLSearchParams(params);
  currentURL.search = searchParams.toString();
```

JulianWielga: we are using https://www.npmjs.com/package/query-string which gives us encode, decode, arrays etc. i just tried to unify and remove `navigate`.

Dzuming: I even found similar doubts here https://dev.to/nerdyman/replacing-query-string-with-native-urlsearchparams-4kdg. It's not a blocker, so I can leave with this solution. You can resolve this comment if you think that there is no reason to change it.

Dzuming (approved):

PR #5262, opened 2023-12-20
[NU-1222]  Node endpoint tests with rest assured 

ForrestFairy:
## Describe your changes
Rewritten tests from `NodesResources` and they are now in `NodesApiSpec` to not use Akka Http and instead use RESTasured.

Also introduced new tests to cover more functionality with tests.

## Checklist before merge


designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,680 @@
...
+            None,
+            None
+          )
+
+          sendRequestAsJsonAsAllpermuser(request)
+            .when()
+            .post(s"$nuDesignerHttpAddress/api/nodes/${process.id}/validation")
+            .Then()
+            .statusCode(200)
+            .body("validationErrors[0]", equalTo(null))
```
coutoPL, line 142: I'm not sure, but doesn't the test pass when there was more than one `validationErrors` and the first is null? Using JSON path (because it's a json path string, right?), in general, isn't enough to check the whole JSON response

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,680 @@
...
+          )
+
+          sendRequestAsJsonAsAllpermuser(request)
+            .when()
+            .post(s"$nuDesignerHttpAddress/api/nodes/${process.id}/validation")
+            .Then()
+            .statusCode(200)
+            .body(
+              equalsJson(
+                Encoder[NodeValidationResult]
```
coutoPL, line 162: plain json (for sure) will be better

designer/server/src/main/scala/pl/touk/nussknacker/ui/suggester/ExpressionSuggester.scala
```diff
@@ -62,7 +62,7 @@ object ExpressionSuggester {
 
 }
 
-@JsonCodec(decodeOnly = true)
+@JsonCodec
```
coutoPL, line 65: the same here

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureExtensions.scala
```diff
@@ -22,6 +22,26 @@ trait NuRestAssureExtensions {
 
   }
 
+  implicit class BasicAuth[T <: RequestSpecification](requestSpecification: T) {
```
coutoPL, line 25: great! Can we refactor the other API-related specs and use these extensions?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+            .Then()
+            .statusCode(200)
+            .body(
+              equalsJson(s"""{
+                   |  "content": "\\nSamples:\\n\\n| id  | value |\\n| --- | ----- |\\n| a   | generated |\\n| b   | not existent |\\n\\nResults for a can be found [here](http://touk.pl?id=a)\\n",
+                   |  "type": "MarkdownAdditionalInfo"
+                   |}""".stripMargin)
+            )
+
+          given()
```
coutoPL, line 75: I wonder if we shouldn't have two tests instead of this one? At the moment it's hard to understand why there are two different responses and how the two requests differs (from the business point of view)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+
+  "The endpoint for nodes" - {
+    "additional info when" - {
+      "authenticated should" - {
+        "return additional info for existing process" in {
+          createSavedProcess(process, TestCategories.Category1, TestProcessingTypes.Streaming)
+
+          given()
+            .basicAuth("allpermuser", "allpermuser")
+            .and()
```
coutoPL, line 43: do we need to add the `and`?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+            )
+        }
+      }
+      "not authenticated should" - {
+        "forbid access" in {
+          createSavedProcess(process, TestCategories.Category1, TestProcessingTypes.Streaming)
+
+          given()
+            .auth()
+            .none
```
coutoPL, line 131: it would be nice to have `noAuth` extension too ;)

coutoPL: and I started to wonder if in case of "forbid access" cases we should pass no credentials or maybe pass invalid credentials for some existing user? WDYT?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+            .and()
+            .jsonBody(s"""{
```
coutoPL, line 133: maybe in the case of tests where the request's body is not so important, we should have a `val` with the example body string somewhere and just reference it here? 

In this test, the important thing is the endpoint and credentials (not the body).  The same when we test the case where the process doesn't exist. 

IMO the body should be inlined in tests in which the body is important in case of tested functionality.
WDYT?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+                 |    },
+                 |    "branchVariableTypes": null,
+                 |    "outgoingEdges": null
+                 |}""".stripMargin
+            )
+            .when()
+            .post(s"$nuDesignerHttpAddress/api/nodes/${process.name}/validation")
+            .Then()
+            .statusCode(200)
+            .body(
```
coutoPL, line 208: WDYT about extension `equalsJsonBody`?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+                 |    },
+                 |    "branchVariableTypes": null,
+                 |    "outgoingEdges": null
+                 |}""".stripMargin
+            )
+            .when()
+            .post(s"$nuDesignerHttpAddress/api/nodes/${process.name}/validation")
+            .Then()
+            .statusCode(200)
+            .body("validationErrors[0].typ", equalTo("ExpressionParserCompilationError"))
```
coutoPL, line 368: why we cannot validate the whole JSON?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
```diff
@@ -0,0 +1,1093 @@
...
+                 |        },
+                 |        "metaDataType": "StreamMetaData"
+                 |    },
+                 |    "name": "test"
+                 |}""".stripMargin
+            )
+            .post(s"$nuDesignerHttpAddress/api/properties/${process.name}/validation")
+            .Then()
+            .statusCode(200)
+            .body("validationErrors[0].typ", equalTo("InvalidPropertyFixedValue"))
```
coutoPL, line 655: the same question here

coutoPL (changes requested):

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
ForrestFairy: I have been using `auth.none` as I thought if user wasn't authenticated, it wouldn't matter if it was due to wrong or no credentials and I thought that `auth.none` would be better visible in code than some wrong credentials, but I am open to change if we think it's better the other way

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiSpec.scala
coutoPL: oh, it seems that I didn't express myself clearly. I didn't mean "local val", but class level val. 

Eg. we have this test:
```scala
    "not authenticated should" - {
        "forbid access" in {
          createSavedProcess(process, TestCategories.Category1, TestProcessingTypes.Streaming)
          val properJson = s"""{
                               |    "id": "1",
                               |    "service": {
                               |        "id": "otherService",
                               |        "parameters": []
                               |    },
                               |    "output": "out",
                               |    "additionalFields": null,
                               |    "type": "Enricher"
                               |}""".stripMargin
          given()
            .noAuth()
            .jsonBody(properJson)
            .when()
            .post(s"$nuDesignerHttpAddress/api/nodes/${process.name}/additionalInfo")
            .Then()
            .statusCode(401)
            .body(
              equalTo("The resource requires authentication, which was not supplied with the request")
            )
        }
      }
```

When you read it, you are interested in checking if request `POST /api/nodes/${process.name}/additionalInfo` cannot be handled when no auth was passed. In this proper example, you are not interested in the body of the request. You don't want to pass to the request body some malformed string, obviously. But as far as the body is valid, it's irrelevant here. 

So, when we define eg.:
```
private val exampleNodesAdditionalInfoRequestBody = 
   s"""{
                               |    "id": "1",
                               |    "service": {
                               |        "id": "otherService",
                               |        "parameters": []
                               |    },
                               |    "output": "out",
                               |    "additionalFields": null,
                               |    "type": "Enricher"
                               |}""".stripMargin
```

somewhere in the Spec class (as far we agree it's not so important, it should probably be placed at the end of the file). 
Then the test will look like this:

```scala
    "not authenticated should" - {
        "forbid access" in {
          createSavedProcess(process, TestCategories.Category1, TestProcessingTypes.Streaming)

          given()
            .noAuth()
            .jsonBody(exampleNodesAdditionalInfoRequestBody)
            .when()
            .post(s"$nuDesignerHttpAddress/api/nodes/${process.name}/additionalInfo")
            .Then()
            .statusCode(401)
            .body(
              equalTo("The resource requires authentication, which was not supplied with the request")
            )
        }
      }
```

Now, it's much more readable, because not important things (emphasizing - in this case!) were moved out of the test.

coutoPL (changes requested):

coutoPL (approved):
LGTM

PR #5266, opened 2023-12-21
[NU-1372] Block accessing methods by field names in record

mslabek:
## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -10,7 +10,9 @@
 * [#5280](https://github.com/TouK/nussknacker/pull/5280) Security improvement: Checking if user has access rights to fragment's Category for fragments served by definitions API
 * [#5303](https://github.com/TouK/nussknacker/pull/5303) Added `skipNodeResults` parameter to API endpoints that return scenario validation results
 * [#5323](https://github.com/TouK/nussknacker/pull/5323) Removed support for [the legacy categories configuration format](https://nussknacker.io/documentation/docs/1.12/installation_configuration_guide/DesignerConfiguration/#scenario-type-categories)
-* [#5266](https://github.com/TouK/nussknacker/pull/5266) Fixed order of access of methods and keys in records
+* [#5266](https://github.com/TouK/nussknacker/pull/5266) Security improvement: fixed order of accessing no-parameter methods and keys in records in expressions. Examples:
+  * Previously `{class: "str"}.class` would return the result of calling `getClass()`
+  * Now `{class: "str"}.class` returns `str`
 
 1.13.0 (Not released yet)
 -------------------------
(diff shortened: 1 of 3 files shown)


interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/internal/propertyAccessors.scala
```diff
@@ -162,14 +162,18 @@ object propertyAccessors {
   }
 
   object MapPropertyAccessor extends PropertyAccessor with ReadOnly {
-
+    // if map does not contain property this should return false so that `NoParamMethodPropertyAccessor` can be used
     override def canRead(context: EvaluationContext, target: scala.Any, name: String): Boolean =
-      true
+      target.asInstanceOf[java.util.Map[_, _]].containsKey(name)
```
mslabek, line 167: Don't really know what to do with this. If this always returns true then we can never call no-parameter methods on records like `empty`  - I think one solution would be for this to return false if the record doesn't have a given key - then the spel parser looks for another accessor - it finds a no-parameter method through `NoParamMethodPropertyAccessor` and uses that. 

But if it returns true only if it contains given key then for example in case we have a typed record with a field `foo` and in runtime the record does not have that key then we get a spel exception - not a `null` like previously. I think a `null` in that case is desired behaviour - there is a test for this - https://github.com/TouK/nussknacker/blob/cac675ce8518f828393c567e05d82055ce86296b/interpreter/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala#L591-L614. 

Don't see any simple solution and don't know if I've explained it clearly - would gladly talk it over offline :smiley:

docs/Changelog.md
```diff
@@ -10,6 +10,7 @@
 * [#5280](https://github.com/TouK/nussknacker/pull/5280) Security improvement: Checking if user has access rights to fragment's Category for fragments served by definitions API
 * [#5303](https://github.com/TouK/nussknacker/pull/5303) Added `skipNodeResults` parameter to API endpoints that return scenario validation results
 * [#5323](https://github.com/TouK/nussknacker/pull/5323) Removed support for [the legacy categories configuration format](https://nussknacker.io/documentation/docs/1.12/installation_configuration_guide/DesignerConfiguration/#scenario-type-categories)
+* [#5266](https://github.com/TouK/nussknacker/pull/5266) Fixed order of access of methods and keys in records
```
coutoPL, line 13: TBH it's hard to understand what's that. Are you able to write from the perspective of a user?

arkadius: I agree, @mslabek , you find a security leak, point it out here :) order is only a way to fix it

mslabek: I've added an extended explanation with an example. 

As to if this is a security problem - I can't get any meaningful vulnerability from this. I've tried a lot of different approaches to get arbitrary code execution or anything - can't get to classloader, runtime but that is probably part our configuration of spel parser and part of its internal implementation. But that implementation may change so I agree it's kind of a security risk.

mslabek: I've shortened the description to be actually readable for users

coutoPL (approved):
LGTM

docs/Changelog.md
coutoPL: yeah, this is it!

arkadius (approved):

PR #5271, opened 2023-12-22
Change AdditionalUIConfigProvider to be similar to FragmentParameter.

wrzontek:
## Checklist before merge


docs/MigrationGuide.md
```diff
@@ -7,6 +7,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 ### Code API changes
 
+* [#TODO](https://github.com/TouK/nussknacker/pull/TODO) Changed AdditionalUIConfigProvider to be more in line with FragmentParameter (TODO more explanation?)
```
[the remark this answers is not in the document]
wrzontek: i thought that 1.13 was already cut off/frozen, but yes it would be simpler if this PR would be part of 1.13

what's the procedure for merging into release branches, do I just duplicate this PR to 1.13?

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+  def getAllForProcessingType(processingType: String): Map[ComponentId, AdditionalUIConfig]
+
+  def getScenarioPropertiesUIConfigs(processingType: String): Map[String, ScenarioPropertyConfig]
+
+}
+
+object AdditionalUIConfigProvider {
+  val empty = new DefaultAdditionalUIConfigProvider(Map.empty, Map.empty)
+
+  case class AdditionalUIConfig(
```
arkadius, line 35: Let's move these two classes outside the object. We used to do this way and after all, it was very messy, We very often had long versions like `AdditionalUIConfigProvider.AdditionalUIConfig` in imports, it was impossible to extract the class to a separate file without changing imports.

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  case class ParameterAdditionalUIConfig(
+      required: Boolean = false,
+      initialValue: Option[FixedExpressionValue],
+      hintText: Option[String],
+      valueEditor: Option[ValueInputWithFixedValues],
+      valueCompileTimeValidation: Option[ParameterValueCompileTimeValidation],
+  ) {
+
+    def toParameterConfigWithoutValidators: ParameterConfig = ParameterConfig(
```
arkadius: I see two options:
1. We keep `ParameterAdditionalUIConfig.valueCompileTimeValidation`, remove `WithoutValidators` from method names and left only a comment that currently it doesn't work - pros: there is a change, that this API won't change when we add the support for validators, cons: we expose some API that looks like support sth, but it doesn't
2. We remove `ParameterAdditionalUIConfig.valueCompileTimeValidation` and remove `WithoutValidators` from method names. Pros/cons: the same as above but reversed.

The current approach is sth in the middle, but only with cons

wrzontek: ok, I think I prefer the first method

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -37,7 +38,7 @@ object UIProcessObjectsFactory {
 
   def prepareUIProcessObjects(
       modelDataForType: ModelData,
-      modelDefinition: ModelDefinition[ComponentStaticDefinition],
+      modelDefinition: ModelDefinition[ComponentStaticDefinition], // todo enrich with AdditionalUIConfigProvider (?)
```
wrzontek, line 41: junk comment from another PR, sorry

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+case class ParameterAdditionalUIConfig(
+    required: Boolean = false,
+    initialValue: Option[FixedExpressionValue],
+    hintText: Option[String],
+    valueEditor: Option[ValueInputWithFixedValues],
+    valueCompileTimeValidation: Option[ParameterValueCompileTimeValidation],
+) {
+
+  def toParameterConfigWithoutValidators: ParameterConfig = ParameterConfig(
+    defaultValue = initialValue.map(_.expression),
```
arkadius, line 64: What about the whole `FragmentParameterValidator.validateFixedExpressionValues` stuff? I think we need to do some refactor and reuse this code. We shouldn't allow to break component definition using this API. Also, I figured out, that this code around `FixedValues` in fragments is in the wrong place. It is used in `extractValidatorsToParam` as a part of `FragmentCompleteDefinitionExtractor` which suppose to contain only validator's related stuff. The logic around other things suppose to be inside `FragmentWithoutValidatorsDefinitionExtractor` like the name describes.

docs/MigrationGuide.md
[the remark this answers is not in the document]
arkadius: It is always an open question, if we should add sth to the cut-offed branch. The Release Manager should consider the situation. The big advantage of merging it to 1.13 is that we won't change the API to often. We actually change the API too often as you can see in the migration guide, but we should try to this rarely. The disadvantage is that Release Manager might do some tests again. Just talk with RM and ask him, what does he think about that concept.

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
wrzontek: `validateFixedExpressionValues` required `expressionCompiler` and `validationContext`

not relying on them was the whole reason behind creating FragmentWithoutValidatorsDefinitionExtractor (though the aim was to ignore `valueCompileTimeValidation`, I didn't consider that it also doesn't validate the fixed values - this currently isn't a big problem as FragmentWithoutValidatorsDefinitionExtractor is used on already defined, validated beforehand, fragments -  but it should be made clearer)

if we require `AdditionalUIConfig.toSingleComponentConfig` to rely on `expressionCompiler` and `validationContext` it'd be difficult to use it in `prepareUIProcessObjects` - perhaps after refactoring the usage of AdditionalUIConfigProvider to happen at the level of ModelData or thereabouts it'd be easier

>We shouldn't allow to break component definition using this API

I agree, but this PR isn't a step back, with the old API (or just through config) one could provide nonsensical editors/validators for parameters and break them completely, here the worst that can happen is a `Fixed list` with no valid values.

arkadius: > validateFixedExpressionValues required expressionCompiler and validationContext
Ahh, I missed it. So it looks like we need to make some things clearer e.g. rename `extractValidatorsToParam` and extend comment next to `FragmentWithoutValidatorsDefinitionExtractor` that it also not validates `initialValue` and `valueEditor`.

> perhaps after refactoring the usage of AdditionalUIConfigProvider to happen at the level of ModelData or thereabouts it'd be easier
I agree. Let's add some TODO 

 > I agree, but this PR isn't a step back, with the old API ...
Yeah, the previous one already was too much low level. I thought that with a little effort and extraction of some code we can make things better. But I see know, that it isn't easy and also I was cheated by classes/methods names in the previous one PR which caused this wrong assumption. Please add some mentioned above clarifications / naming tweaks and we can move forward

arkadius: Now this code looks much better. It is easier to understand what are the responsibilities of each classes. Now, what do you think about reusing the `FragmentParameterValidator.validateFixedValuesList` also for non-fragment components? It doesn't need expression compiler.

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+}
+
+case class ParameterAdditionalUIConfig(
+    required: Boolean = false,
+    initialValue: Option[FixedExpressionValue],
+    hintText: Option[String],
+    valueEditor: Option[ValueInputWithFixedValues],
+    valueCompileTimeValidation: Option[ParameterValueCompileTimeValidation],
```
arkadius, line 58: Let's add a TODO that it isn't supported yet

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
wrzontek: yep, i've reused it and a bit more, good idea

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/ValueEditorValidator.scala
```diff
@@ -0,0 +1,84 @@
...
+}
+import pl.touk.nussknacker.engine.api.editor.DualEditorMode
+import pl.touk.nussknacker.engine.graph.node.FragmentInputDefinition.{FragmentClazzRef, ValueInputWithFixedValues}
+
+object ValueEditorValidator {
+
+  def validateAndGetEditor( // this method doesn't validate the compilation validity of FixedExpressionValues (it requires validationContext and expressionCompiler, see FragmentParameterValidator.validateFixedExpressionValues)
+      valueEditor: ValueInputWithFixedValues,
+      initialValue: Option[FixedExpressionValue],
+      refClazzName: Option[String],
```
arkadius, line 22: 1. Let's use value class (`FragmentClazzRef`) instead of `String`
2. Let's extract two methods - one which validates against `clazzRef` and another which validates rest. Thanks to that we avoid passing None for components. Also, I think that the first method (`validateAgainstClazzRef`) should be in `FragmentParameterValidation` because for components it is impossible to use it yet.

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -55,12 +54,26 @@ case class ParameterAdditionalUIConfig(
...
-  def toParameterConfig: ParameterConfig = ParameterConfig(
-    defaultValue = initialValue.map(_.expression),
-    editor = valueEditor.map(EditorExtractor.extract),
+  def toParameterConfig(paramName: String): ParameterConfig = ParameterConfig(
+    defaultValue = initialValue.map(
+      _.expression
+    ), // TODO validating initialValue with context and expressionCompiler will be similar to handling 'valueCompileTimeValidation'
+    editor = valueEditor.flatMap(editor =>
+      ValueEditorValidator
+        .validateAndGetEditor(
```
arkadius, line 68: Let's use named parameters

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/ValueEditorValidator.scala
```diff
@@ -12,21 +12,20 @@ import pl.touk.nussknacker.engine.api.definition.{
   ParameterEditor
 }
 import pl.touk.nussknacker.engine.api.editor.DualEditorMode
-import pl.touk.nussknacker.engine.graph.node.FragmentInputDefinition.{FragmentClazzRef, ValueInputWithFixedValues}
+import pl.touk.nussknacker.engine.graph.node.FragmentInputDefinition.{
```
arkadius, line 15: 1. I just realize that you still keep these classes inside `FragmentInputDefinition`. We need to move them out from there. They become not only Fragment specific. 
2. We have a dedicated module (`common-api`) which was intended to be used in such situation when we have some part of api shared between components and scenario graph. Let's move `FixedExpressionValue` class there and reuse it instead of duplicating

docs/MigrationGuide.md
```diff
@@ -7,6 +7,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 ### Code API changes
 
+* [#5271](https://github.com/TouK/nussknacker/pull/5271) Changed AdditionalUIConfigProvider to be more in line with FragmentParameter (TODO more explanation?)
```
arkadius, line 10: What about this TODO?

designer/server/src/main/scala/pl/touk/nussknacker/ui/additionalconfig/AdditionalUIConfigProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Trait allowing the provision of UI configuration for components and scenario properties, without requiring a model reload.
+ *
+ * TODO: The current implementation allows providing configs only for standard components - meaning that base components aren't handled.
+ * TODO: Currently the value of 'valueCompileTimeValidation' has no effect, it'll be supported in the future but is included now to keep the API stable.
+ *       Validating and extracting a parameter's validators requires access to expressionCompiler and validationContext,
+ *       this is currently hard to achieve where AdditionalUIConfigProvider is used (UIProcessObjectsFactory),
+ *       after refactoring it to be used at the level of ModelData or so it should be easy, and support for 'valueCompileTimeValidation' will be possible
+ */
+trait AdditionalUIConfigProvider extends Serializable {
```
arkadius, line 17: I think that it needs to be in some api module :) `component-api` or `exetensions-api`. How otherwise you would provide your own implementation of plugin?

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/UIProcessObjectsFactory.scala
```diff
@@ -61,7 +61,7 @@ object UIProcessObjectsFactory {
       combinedComponentsConfig,
       additionalUIConfigProvider
         .getAllForProcessingType(processingType)
-        .mapValuesNow(_.toSingleComponentConfig)
+        .mapValuesNow(toSingleComponentConfig)
```
arkadius, line 64: Let's use explicit `ComponentAdditionalConfigConverter` call

arkadius (approved):
One little comment added

PR #5275, opened 2023-12-22
[NU-1361] Move fragment input definition validation to BE

mslabek:
### Scope
Moves remaining fragment parameters validation to BE - that is checking if:
1. Parameter name is unique
2. Parameter name is a valid identifier.

To ensure consistency, I've centralized the logic of checking variable identifier validity in api.

### Problems
When names are duplicated, the validation doesn't return more specific errors than about names. The errors are displayed based on name so if it wasn't fail-fast, we'd get duplicated errors for parameters with duplicated names. And  we shouldn't rely on index since that causes problems with drag'n'drop.

A good way to solve this would be to rely on an identifier from FE. To not pass this data to db, a custom encoder could skip the id field. I can leave it as is - this behaviour should be exactly the same from user perspective or I can make this change if we think this is a clean enough solution.

## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
@@ -3,26 +3,15 @@ package pl.touk.nussknacker.engine.api.context
 import cats.data.Validated.{Invalid, Valid}
 import cats.data._
 import cats.implicits._
-import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.{InvalidVariableOutputName, OverwrittenVariable}
-import pl.touk.nussknacker.engine.api.context.ValidationContext.{empty, validateVariableName}
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.OverwrittenVariable
+import pl.touk.nussknacker.engine.api.context.ValidationContext.empty
 import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 import pl.touk.nussknacker.engine.api.NodeId
-
-import javax.lang.model.SourceVersion
+import pl.touk.nussknacker.engine.api.validation.Validations.validateVariableName
 
 object ValidationContext {
 
   def empty: ValidationContext = ValidationContext()
-
-  def isVariableNameValid(name: String): Boolean = SourceVersion.isIdentifier(name)
-
-  def validateVariableName(name: String, paramName: Option[String])(
-      implicit nodeId: NodeId
-  ): ValidatedNel[PartSubGraphCompilationError, String] = {
-    // TODO: add correct and more precise error messages
-    if (isVariableNameValid(name)) Valid(name)
-    else Invalid(InvalidVariableOutputName(name, paramName)).toValidatedNel
-  }
 
 }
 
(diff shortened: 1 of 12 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
```diff
@@ -13,6 +13,17 @@ import javax.lang.model.SourceVersion
 object ValidationContext {
 
   def empty: ValidationContext = ValidationContext()
+
+  def isVariableNameValid(name: String): Boolean = SourceVersion.isIdentifier(name)
+
+  def validateVariableName(name: String, paramName: Option[String])(
```
wrzontek, line 19: is this the best place for these methods? they don't depend on the Context at all
maybe just create a separate object for them

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -163,12 +165,36 @@ class NodeCompiler(
           )
       }
 
-      val parameterExtractionValidation =
-        NonEmptyList.fromList(parameterDefinitions.written).map(invalid).getOrElse(valid(List.empty))
+      val parameterNameValidation = parameterDefinitions.value
```
wrzontek, line 168: i'd move this logic to a helper function, this function is complicated enough as is

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
```diff
@@ -13,6 +13,17 @@ import javax.lang.model.SourceVersion
...
+      implicit nodeId: NodeId
+  ): ValidatedNel[PartSubGraphCompilationError, String] = {
+    // TODO: add correct and more precise error messages
+    if (isVariableNameValid(name)) Valid(name)
+    else Invalid(InvalidVariableOutputName(name, paramName)).toValidatedNel
```
wrzontek, line 24: i think `InvalidVariableOutputName` is not a perfect name, it's should just be `InvalidVariableName`, the 'output' part looks like a legacy leftover

wrzontek: its text in PrettyValidationError.scala also should change

arkadius (approved):
Good job

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
arkadius: Yeah, I think that we won't find better.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
mslabek: Good point, done :+1:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ValidationContext.scala
mslabek: I didn't notice that, I think you're right. I've changed it and also message in OverwrittenVariable error.

mslabek: FYI, the error message from InvalidVariableName is not correct since the we use normal java variable identifier rules and they are not as simple as `only letters, numbers or '_', cannot be empty` :smile: I've left a TODO to provide more precise error messages.

wrzontek (approved):
LGTM

PR #5280, opened 2023-12-29
UI definition refactor

arkadius:
## Describe your changes

Changes in the definition API:
  * `/processDefinitionData/componentIds` endpoint is removed
  * `/processDefinitionData/*` response changes:
    * `services`, `sourceFactories`, `sinkFactories`, `customStreamTransformers` and `fragmentInputs` maps inside `processDefinition` were replaced by
      one `components` map with key in format `$componentType-$componentName` and moved into top level of response
    * `typesInformation` inside `processDefinition` was renamed into `classes`, moved into top level of response 
      and nested `clazzName` inside each element was extracted
    * `nodeId` field inside `edgesForNodes` was renamed into `componentId` in the flat `$componentType-$componentName` format
    * `defaultAsyncInterpretation` field was removed

Also, FE doesn't filter components by `categories` anymore - this job is already done on BE side. `categories` are still returned to FE in the response (it will be changed soon), but they aren't used by FE.

On the BE side:
* `ComponentIdProvider` moved back into designer module
* `ModelDefinitionWithComponentIds` is used locally in one place - in other places standard `ModelDefinition` is used instead
* `FragmentRepository` uses `FetchingProcessRepository` instead of its custom implementation
* `FragmentRepository` is filtering components by `processingType` instead of `category` 

## Checklist before merge


designer/client/src/common/ProcessUtils.ts
```diff
@@ -171,36 +179,51 @@ class ProcessUtils {
...
+        const foundDefinition = components?.[this.findComponentId(node)];
         const emptyDefinition = {
             parameters: null,
             returnType: null,
         };
         return foundDefinition || emptyDefinition;
     };
 
-    private findDefinitionsForType = (processDefinition: ProcessDefinition, node?: NodeType): Record<string, NodeObjectTypeDefinition> => {
+    findComponentId = (node?: NodeType): string | null => {
```
wrzontek, line 191:
```suggestion
    resolveComponentId = (node?: NodeType): string | null => {
```
i'm not familiar with TS naming schemes, but to me `find` doesn't fit, it's all done on static information, no searching required

designer/client/src/common/ProcessUtils.ts
```diff
@@ -244,15 +267,10 @@ class ProcessUtils {
...
-    findNodeDefinitionIdOrType = (node: NodeType) => this.findNodeDefinitionId(node) || node.type || null;
-
-    getNodeBaseTypeCamelCase = (node: NodeType) => node.type && node.type.charAt(0).toLowerCase() + node.type.slice(1);
-
     findNodeConfigName = (node: NodeType): string => {
         // First we try to find id of node (config for specific custom node by id).
-        // If it is falsy then we try to extract config name from node type (config for build-in components e.g. variable, join).
-        // If all above are falsy then it means that node is special process properties node without id and type.
-        return this.findNodeDefinitionId(node) || this.getNodeBaseTypeCamelCase(node) || "$properties";
+        // When is missing it means that node is special process properties node without id
```
wrzontek, line 272:
```suggestion
        // When it is missing it means that the node is a special process properties node without id
```
and the old comment above should be 
` // First we try to find the id of the node (configs for specific custom nodes are resolved by id).
`

articles are hard

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/fragment/FragmentRepository.scala
```diff
@@ -1,132 +1,56 @@
...
+  final def fetchLatestFragmentsSync(processingType: ProcessingType)(implicit user: LoggedUser): List[FragmentDetails] =
+    Await.result(fetchLatestFragments(processingType), 10 seconds)
 
-  def loadFragments(versions: Map[String, VersionId], category: String): Set[FragmentDetails]
+  def fetchLatestFragments(processingType: ProcessingType)(implicit user: LoggedUser): Future[List[FragmentDetails]]
 
-  // FIXME: get by id in DB
-  def get(id: String): Option[FragmentDetails] = loadFragments().find(_.canonical.metaData.id == id)
+  // FIXME: async version should be used instead
+  final def fetchLatestFragmentSync(processName: ProcessName)(implicit user: LoggedUser): Option[FragmentDetails] =
```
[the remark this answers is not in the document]
arkadius: I changed to fragmentName, good point

designer/client/src/common/ProcessUtils.ts
arkadius: Yeah, `find` is confusing in this context, but I think the `determine` word will be a little better than `resolve`, we use it quite often on the BE side in similiar context.

designer/client/src/common/ProcessUtils.ts
arkadius: Yep, this comment had more problems than only articles, I rephrased it more deeply.

wrzontek (approved):
LGTM :+1:

PR #5313, opened 2024-01-05
Date, aggregation and numeric helpers changes

paw787878:
## Checklist before merge


engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -171,6 +172,75 @@ object aggregates {
...
+    override def computeOutputType(input: typing.TypingResult): Validated[String, typing.TypingResult] = {
+
+      if (!input.canBeSubclassOf(Typed[Number])) {
+        Invalid(s"Invalid aggregate type: ${input.display}, should be: ${Typed[Number].display}")
+      } else {
+        Valid(ForLargeFloatingNumbersOperation.promoteSingle(input))
+      }
+    }
+
+    override def computeStoredType(input: typing.TypingResult): Validated[String, typing.TypingResult] =
```
arkadius, line 238: Uhh, I forgot that we have this clever mechanism, that generates flink's type information based on the `TypingResult`. Take a look at `TypingResultAwareTypeInformationDetection`. As far I remember, it was important:
1. For the time optimization reasons
2. To make state readable across deployments - I mean, if the state is stored using wrong TypeInformation, features like redeploy could not work. To redeploy you need to hit the deploy button for already running scenario.

I think that we should consider:
- Doing some tests around this
- Changing the type into Map[String, Any] and typing result to `TypedObjectTypingResult` with two fields: sum and count. The type of `sum` should be the same as returned by `computeOutputType`

arkadius (commented):
It looks ok, I added one important notice about state serialization

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -171,6 +172,84 @@ object aggregates {
...
+        case null =>
+          // will be replaced to Double.Nan in alignToExpectedType iff return type is known to be Double
+          null
+        case sum: java.lang.Double     => Predef.double2Double(sum / count)
+        case sum: java.math.BigDecimal => (BigDecimal(sum) / BigDecimal(count)).bigDecimal
+      }
+    }
+
+    override def alignToExpectedType(value: AnyRef, outputType: TypingResult): AnyRef = {
+      if (value == null && outputType == Typed(classOf[lang.Double])) {
```
arkadius, line 231: It can be just `Typed[Double]`, we always convert primitives to wrappers in `Typed.typedClass`. Thanks to that we can use scala idiomatic types

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -171,6 +172,84 @@ object aggregates {
...
+        aggregate1.count + aggregate2.count
+      )
+
+    override def result(finalAggregate: Aggregate): AnyRef = {
+      val count = finalAggregate.count
+      finalAggregate.sum match {
+        case null =>
+          // will be replaced to Double.Nan in alignToExpectedType iff return type is known to be Double
+          null
+        case sum: java.lang.Double     => Predef.double2Double(sum / count)
```
arkadius, line 225: I think that it can be just `sum / count` - scala should handle this conversion on their own.

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
```diff
@@ -171,6 +172,84 @@ object aggregates {
...
+        case sum: java.math.BigDecimal => (BigDecimal(sum) / BigDecimal(count)).bigDecimal
+      }
+    }
+
+    override def alignToExpectedType(value: AnyRef, outputType: TypingResult): AnyRef = {
+      if (value == null && outputType == Typed(classOf[lang.Double])) {
+        Predef.double2Double(Double.NaN)
```
arkadius, line 232: I think that it can be just `Double.NaN` - scala should handle this conversion on their own.

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
paw787878: Intellij does not show error when I remove double2Double, but when I run test which uses this code there is an error "the result type of an implicit conversion must be more specific than AnyRef". I added .asInstanceOf[AnyRef] instead

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/aggregates.scala
paw787878: Łukasz Jędrzejewski suggested that in esp they add @TypeInfo  to case classes to speed up the serialization. I did it in commit https://github.com/TouK/nussknacker/pull/5313/commits/7abb8bdce557805262831d3603d99feeefca9b6f . Does this solve this problem?

arkadius: It almost handles it correctly. I checked in the debugger such a code:
```
TypeInformation.of(classOf[AverageAggregatorState])
```
and I see that one problem that remain is that `fieldType` for sum is still generic.
We can easily fix that by replacing the `sum` with nullable `sumDouble` and `sumBigDecimal`. Unfortunatelly, `CaseClassTypeInfoFactory` doesn't handles either sum types (`Either`, sealed traits) or `Option`.

arkadius (approved):
Good job

PR #5322, opened 2024-01-09
[NU-1398] removed unnecessary colon character

pielas:
## Checklist before merge

### DIFF
--- a/designer/client/src/http/HttpService.ts
+++ b/designer/client/src/http/HttpService.ts
@@ -374,7 +374,7 @@ class HttpService {
     }
 
     changeProcessName(processName, newProcessName): Promise<boolean> {
-        const failedToChangeNameMessage = i18next.t("notification.error.failedToChangeName", "Failed to change scenario name:");
+        const failedToChangeNameMessage = i18next.t("notification.error.failedToChangeName", "Failed to change scenario name");
         if (newProcessName == null || newProcessName === "") {
             this.#addErrorMessage(failedToChangeNameMessage, i18next.t("notification.error.newNameEmpty", "Name cannot be empty"), true);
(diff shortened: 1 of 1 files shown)


designer/client/src/http/HttpService.ts
```diff
@@ -374,7 +374,7 @@ class HttpService {
     }
 
     changeProcessName(processName, newProcessName): Promise<boolean> {
-        const failedToChangeNameMessage = i18next.t("notification.error.failedToChangeName", "Failed to change scenario name:");
+        const failedToChangeNameMessage = i18next.t("notification.error.failedToChangeName", "Failed to change scenario name");
```
arkadius, line 377: Do you know how will behave the case when scenario name is empty? I see that in this case this message is somehow bundled with "Name cannot be empty"?

pielas: I couldn't reach such state from the UI. When I tried set empty name in the properties window I received "Scenario name is mandatory and cannot be empty" message right under name field

arkadius: So it looks like it is a death code? If so, let's remove it.

pielas: I don't think it's a dead code. It looks like this `if` is something just in case and I think it's better to leave it there. In terms of what this code does, `#addErrorMessage` function has such signature `#addErrorMessage(message: string, error: string, showErrorText: boolean)` and it seems that "Name cannot be empty" is passed as error parameter and it's not concatenated with message(at least at this point). In such case my change shouldn't have any negative impact on code under if clause, even if it has somehow been reached.

arkadius (approved):

PR #5324, opened 2024-01-09
[NU-1399] provide value type to suggester

Dzuming:
## Describe your changes
image

## Checklist before merge

### DIFF
--- a/designer/client/src/components/graph/node-modal/fragment-input-definition/settings/variants/fields/ValidationFields.tsx
+++ b/designer/client/src/components/graph/node-modal/fragment-input-definition/settings/variants/fields/ValidationFields.tsx
@@ -1,8 +1,8 @@
-import React from "react";
+import React, { useMemo } from "react";
 import { useTranslation } from "react-i18next";
 import { FieldName, onChangeType, ValueCompileTimeValidation } from "../../../item";
 import { SettingRow, fieldLabel } from "./StyledSettingsComponnets";
-import { NodeValidationError, VariableTypes } from "../../../../../../../types";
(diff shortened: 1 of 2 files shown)


designer/client/src/components/graph/node-modal/fragment-input-definition/settings/variants/fields/ValidationFields.tsx
```diff
@@ -25,11 +26,20 @@ export default function ValidationFields({
...
+    typ,
 }: ValidationFields) {
     const { t } = useTranslation();
 
     const validationExpressionFieldName: FieldName = `$param.${name}.$validationExpression`;
 
+    const extendedVariableType = useMemo(
+        () => ({
+            ...variableTypes,
+            value: { refClazzName: typ.refClazzName, type: "TypedClass", params: [] } as ReturnedType,
```
mslabek, line 38: This request template is not always correct. When you select a parameter type with a generic parameter like `List[Unknown]` the response for /suggestions is `Type's generic parameters don't match expected type parameters: found List[], expected List[E]. This may be caused by passing incorrect arguments or using type aliases.` For these types, FE should pass these type parameters in the `params` array. 

You can see which types I mean when you check the `processDefinitionData` response in the `typesInformation` array.

I think if BE provides refClazzName then it should provide info about its parameters. But without changing BE I guess you could try to get that info from the `typesInformation` request (I'm guessing these class names are unique and it should work?)

Dzuming: You are right, I changed it to  use data from `typesInformation` ✔️

mslabek (approved):
LGTM!

PR #5327, opened 2024-01-09
Parallelize scenario migrations testing

Elmacioro:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
@@ -51,8 +51,7 @@ trait RemoteEnvironment {
   // TODO This method is used by an external project. We should move it to some api module
   def testMigration(
       processToInclude: ScenarioWithDetails => Boolean = _ => true,
-      batchingExecutionContext: ExecutionContext,
-      batchingRequestTimeout: FiniteDuration
+      batchingExecutionContext: ExecutionContext
   )(
       implicit ec: ExecutionContext,
       user: LoggedUser
@@ -123,7 +122,11 @@ class HttpRemoteEnvironment(
   def closeAsync(): Future[Unit] = http.shutdownAllConnectionPools()
 }
 
-final case class StandardRemoteEnvironmentConfig(uri: String, batchSize: Int = 10)
+final case class StandardRemoteEnvironmentConfig(
+    uri: String,
+    batchSize: Int = 10,
+    batchTimeout: FiniteDuration = 120 seconds
+)
 
 //TODO: extract interface to remote environment?
 trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironment {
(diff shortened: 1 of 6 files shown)


Elmacioro:
After consulting with @piotrp I will also:

- parallelize fetchGroupByGroup operation
- Modify logging to show how many threads we're using for testing migration
- Add some meaningful prefixes for ForkJoinPool threads so it looks better in thread dumps/ profiler

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/TestModelMigrations.scala
```diff
@@ -91,6 +95,22 @@ class TestModelMigrations(
 
   }
 
+  private def processInParallel(
+      input: List[MigratedProcessDetails],
+      maxParallelism: Int
+  )(process: MigratedProcessDetails => TestMigrationResult): List[TestMigrationResult] = {
+    val pool        = new ForkJoinPool(maxParallelism)
```
arkadius, line 102: Is it a good idea to create a pool for every request? @piotrp wdyt? Maybe we should Move this pool outside to the place where we know if it is a good idea or not - to the place of `testMigrations` invocation

piotrp: I'm not sure, right now `testMigrations` is simple to use - you just pass it an additional integer (which we may redefine to `maxParallelism: Int = 1` to make it even simpler). I think migrations are heavy enough so that creating a thread pool here is justified, especially that this method is meant for performing operations in bulk.

piotrp: An idea: maybe let's do it Scala-way, and reuse `ExecutionContext` that is already being passed to `HttpRemoteEnvironment`? It's currently unused. Then, in cases where we don't care about degree of parallelism, we can just pass default EC.

Elmacioro: I updated the PR with following changes:

- I removed unused ExecutionContext from HttpRemoteEnvironment
- I modified testMigrations to use ExecutionContext instead of ForkJoinPool directly
- I parallelized fetchGroupByGroup operation
- I created custom ExecutionContext with thread prefixes for easier time with thread dumps

After @piotrp comment I will try to utilize previously unused ExecutionContext that I deleted from HttpRemoteEnvironment instead of creating one for collections operations.

piotrp: Don't restore EC to HttpRemoteEnvironment, it's enough that you have one passed to testMigration

Elmacioro: Ok, I added a commit where I use EC which is already passed to the method implicitly.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -222,16 +224,21 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
-          for {
-            current <- acc
-            fetched <- fetchProcessesDetails(processesGroup)
-          } yield current ::: fetched
-      }
+      .toVector
+    // We create ParVector manually instead of calling par for compatibility with Scala 2.12
+    val parallelCollection = new ParVector(groupedBasicProcesses)
+    val taskSupport        = new ExecutionContextTaskSupport(ec)
+    parallelCollection.tasksupport = taskSupport
```
piotrp, line 234: nit: you can make it in just one line, `parallelCollection.tasksupport = new ExecutionContextTaskSupport(ec)`
(or close this thread if you don't want to)

piotrp (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
Elmacioro: Oh, I missed it! 
I inlined ExecutionContextTaskSupport instantiation.

Elmacioro:
I will abstain from merging for the time being as I'll cherry-pick commits from this PR to preview/esp_1.12.4 branch and test the changes.

Elmacioro:
I updated PR to use two ExecutionContexts instead of one as using only one EC with strictly restricted parallelism made things significantly worse performance-wise during tests. The new EC will be only used for operations on parallel collections while the old implicit one will take care of the tasks that it used to.

Elmacioro:
In the latest commit I modified fetchGroupByGroup to be executed in parallel bounding the number of maximum concurrent requests to the number of threads in the batchingExecutionContext. Previous approach did not work as I intended since foldLeft forced fetching to be executed in sequence even when called on parallel collection.

arkadius (approved):
LGTM

PR #5363, opened 2024-01-14
[NU-1361] Add unique fragment output validation

mslabek:
## Describe your changes
Fragment output name validation is done only from the context of scenario using the fragment. This is bad since all validations that can be corrected on the level of fragment authoring should be displayed on that level. This change introduces the first (I think) validation of this kind in context of fragment validation.

To improve UX, this change changes the type of global errors to contain list of node ids. These node ids are linked in the tips panel and highlighted on the visualization. When entered into node, these errors are not visible as they cannot be fixed without triggering global validation. This is also not ideal but deals with the problem of disappearing validation messages when they do not appear from the context of node validation.

Other global errors are also ported to the new mechanism.

## Checklist before merge


Dzuming (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -285,18 +287,27 @@ object ProcessCompilationError {
 
   final case class DisablingManyOutputsFragment(id: String, nodeIds: Set[String]) extends ProcessCompilationError
 
-  final case class DisablingNoOutputsFragment(id: String) extends ProcessCompilationError {
+  // TODO local: check if this displays ok
```
wrzontek, line 290: todo local

wrzontek: same below

mslabek: thanks! btw they di not display properly but it was the same before the change. Will make these changes in other PRs

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -38,6 +38,8 @@ object ProcessCompilationError {
 
   }
 
+  sealed trait GlobalError { self: ProcessCompilationError => }
```
arkadius, line 41: 1. Let's call it `GlobalScenarioGraphError` or `ScenarioGraphLevelError` or sth similar. Global/local naming is always confusing. You never now when you'll have sth that need to be "even more global" or "even more local"
2. Let's add a comment that these errors should be thrown when the error can't be fixed by node parameters editing. We also treat otherwise node ids.

arkadius (approved):
I only left one comment about naming/comments in the code. It looks very promising, I'm looking forward to migrate all current global errors to this mechanism.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
mslabek: Done :+1:

designer/client/src/components/tips/error/ErrorTips.tsx
```diff
@@ -1,29 +1,46 @@
...
-    const nodeErrors = errors.invalidNodes;
-    const propertiesErrors = errors.processPropertiesErrors;
+    const { globalErrors, processPropertiesErrors, invalidNodes } = errors;
 
-    const globalErrorsTips = (globalErrors: NodeValidationError[]) => <div>{globalErrors.map((error) => globalError(error, null))}</div>;
+    const invalidNodeIds = Object.keys(invalidNodes);
 
-    const globalError = (error: NodeValidationError, suffix) => (
+    const renderGlobalErrorSimpleMessage = (error: NodeValidationError) => (
         <span key={uuid4()} title={error.description}>
```
Dzuming, line 15: We cannot use uuid4() ad a key, because it will force re-render of the component on every react pre-render. We need to use something stable as a key, which won't change during the lifecycle of the component.

mslabek: Thanks, havent thought about it. I changed it to use index. I also removed displaying fieldname since for global errors its always null and removed the function.

Dzuming (approved):

PR #5371, opened 2024-01-15
DictParameterEditor

wrzontek:
## Checklist before merge


designer/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
```diff
@@ -32,9 +32,43 @@ class DictsFlowTest
...
   private val Key            = "foo"
   private val Label          = "Foo"
+  private val DictId         = "dict"
 
   override def nuTestConfig: Config = ConfigWithScalaVersion.TestsConfigWithEmbeddedEngine
 
+  test("query dict entries by label pattern") {
+    val response1 = httpClient.send(
+      quickRequest
+        .get(uri"$nuDesignerHttpAddress/api/processDefinitionData/$Streaming/dict/$DictId/entry?label=fo")
```
arkadius, line 42: I know that it isn't your orignal code, but let's change the Label to sth different. Currently, it could be confusing, because label looks like the key.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ExpressionCompiler.scala
```diff
@@ -219,9 +228,12 @@ class ExpressionCompiler(expressionParsers: Map[String, ExpressionParser]) {
...
 
   def withExpressionParsers(modify: PartialFunction[ExpressionParser, ExpressionParser]): ExpressionCompiler =
-    new ExpressionCompiler(expressionParsers.map { case (k, v) =>
-      k -> modify.lift(v).getOrElse(v)
-    })
+    new ExpressionCompiler(
+      expressionParsers.map { case (k, v) =>
+        k -> modify.lift(v).getOrElse(v)
+      },
+      dictRegistry
```
arkadius, line 235: Take a look for the usage of this method :) There is only one:
```
uiValidator.withValidator(_.withExpressionParsers {
    case spel: SpelExpressionParser =>
      spel.typingDictLabels
  })
```
The problem that we have is that we don't want to query labels on the engine side and this problem I think is still valid for your case - I'm not sure about that. Even if this method works correctly for your case, I think that we need to change the abstraction level of this method to avoid confusions e.g. rename it to `typingDictLabels`, rename `dictRegistry` to `dcitParameterRegistry` and add a comment that during dict parameter validation, there is always a key inside the expression.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
```diff
@@ -89,4 +91,30 @@ object Validations {
...
+      .sequence
+      .map(_ => ())
+  }
+
+  private def validateDictParameter(definition: Parameter, parameter: NodeParameter, dictRegistry: DictRegistry)(
+      implicit nodeId: NodeId
+  ): ValidatedNel[PartSubGraphCompilationError, Unit] = definition.editor match {
+    case Some(DictParameterEditor(dictId)) =>
+      dictRegistry
+        .labelByKey(dictId, parameter.expression.expression)
```
arkadius, line 114: How it would work on the engine side? We don't want to engine query the registry - take a look at `DictRegistry.toEngineRegistry`

designer/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
```diff
@@ -32,9 +32,43 @@ class DictsFlowTest
   private val EndNodeId      = "end"
   private val Key            = "foo"
   private val Label          = "Foo"
+  private val DictId         = "dict"
 
   override def nuTestConfig: Config = ConfigWithScalaVersion.TestsConfigWithEmbeddedEngine
 
+  test("query dict entries by label pattern") {
```
arkadius, line 39: The purpose of these *Flow* tests is to show e2e a flow that user follows. Thanks to that we avoid the situation that something works locally, but after integrating all mechanisms, it makes no sense. Do you know this meme?
image

We need a test that would follow the flow:
- user creates a scenario
- user uses the editor to query for a key
- user saves the scenario with the provided key
- user invokes a test based on this scenario

Current, tests do similar thing using dictionary variables (take a look at usage of `saveProcessAndTestIt`). They had a gap in the second step, because when we wrote them, the suggestion mechanism was on the FE side. When we moved it into BE side, we forgot to add this missing part.

I'm not sure it this test will work correctly in the 4th step - we don't want to Flink validate dictionaries.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
wrzontek: `keyByLabel` is the one that cannot be used on engine side, so it should be fine, unless i'm misunderstanding something

wrzontek:
note to self: remember to add notes to changelog and migrationguide after undrafting

common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
```diff
@@ -1,17 +1,32 @@
...
 @JsonCodec case class Expression(language: String, expression: String)
 
 object Expression {
 
   object Language {
-    val Spel         = "spel"
-    val SpelTemplate = "spelTemplate"
+    val Spel             = "spel"
+    val SpelTemplate     = "spelTemplate"
+    val Literal          = "literal" // TODO remove if not needed?
```
arkadius: YAGNI, if you haven't used yet, move it back to the test code

common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
```diff
@@ -1,17 +1,32 @@
...
+    val DictLabelWithKey = "dictLabelWithKey"
```
arkadius, line 15: Can we stick with key -> label order. Do you feal that the reversed order in this names is somehow very important? key -> label seems more natural

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/ExpressionCompiler.scala
wrzontek: renamed/clarified it into `withLabelsDictTyper`, as that was it's only usage

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DictResources.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import akka.http.scaladsl.marshalling.ToResponseMarshallable
+import akka.http.scaladsl.model.{HttpResponse, StatusCodes}
+import akka.http.scaladsl.server.{Directives, Route}
+import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport
+import pl.touk.nussknacker.engine.api.dict.DictQueryService
+
+import scala.concurrent.ExecutionContext
+
+class DictResources(implicit ec: ExecutionContext) extends Directives with FailFastCirceSupport {
```
arkadius, line 11: Sorry for the late announcement but we create every new endpoint in tapir. Can you rewrite it to tapir?

designer/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
wrzontek: created such a test

common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
```diff
@@ -1,17 +1,32 @@
...
   }
 
   def spel(expression: String): Expression = Expression(Language.Spel, expression)
 
   def spelTemplate(expression: String): Expression = Expression(Language.SpelTemplate, expression)
+
+  def literal(expression: String): Expression = Expression(Language.Literal, expression)
+
+  @JsonCodec
+  case class DictLabelWithKeyExpression(dictId: String, label: String, key: String)
```
arkadius, line 25: 1. Why we need `dictId`? It is already available in the definition of component (in the editor). Which dictId will be more important - this stored in the json or in the definition?
2. I can't see any logic in the scenario UI resolving part. How to you handle changes in the label?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DictResources.scala
wrzontek: will do in separate PR

common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
wrzontek: i've fiddled with this for a while, and it's hard to do nicely because of `expression: String`, i've added a TODO to change it, but I can't do it right now

I think the way I did it (one class with optional label) is slightly better because of only one parser

common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
```diff
@@ -1,17 +1,29 @@
...
+    val SpelTemplate     = "spelTemplate"
+    val DictKeyWithLabel = "dictKeyWithLabel"
   }
 
   def spel(expression: String): Expression = Expression(Language.Spel, expression)
 
   def spelTemplate(expression: String): Expression = Expression(Language.SpelTemplate, expression)
+
+  @JsonCodec
+  case class DictKeyWithLabelExpression(key: String, label: Option[String])
```
arkadius, line 22: Please leave a comment why label is optional

interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/ProcessDictSubstitutor.scala
```diff
@@ -19,44 +20,89 @@ import pl.touk.nussknacker.engine.spel.ast.{
...
     }
 
     rewriter.rewriteProcess(process)
   }
 
+  private def removeLabelFromDictKeyExpression(expr: Expression) =
+    DictKeyWithLabelExpressionParser
+      .parseDictKeyWithLabelExpression(expr.expression)
+      .map(expr => Expression.dictKeyWithLabel(expr.key, None))
+      .leftMap(e => throw new IllegalStateException(s"Errors parsing DictKeyWithLabel expression: $expr: $e"))
```
arkadius, line 61: Use `valueOr`

interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/ProcessDictSubstitutor.scala
```diff
@@ -19,44 +20,89 @@ import pl.touk.nussknacker.engine.spel.ast.{
...
-              s"Found ${substitutions.size} substitutions in expression: ${process.name} > ${nodeExpressionId.nodeId.id} > ${nodeExpressionId.expressionId}. " +
-                s"Expression: '${expr.expression}' replaced with '$afterSubstitution'"
-            )
-          afterSubstitution
-        }
-        .getOrElse(expr.expression)
-      expr.copy(expression = substitutedExpression)
+
+      if (expr.language == Expression.Language.DictKeyWithLabel) {
+        if (!isReverse)
```
arkadius, line 46: Let's avoid double negation. I mean:
```
if (isReverse) { } else {}
```

interpreter/src/main/scala/pl/touk/nussknacker/engine/dict/ProcessDictSubstitutor.scala
```diff
@@ -19,44 +20,89 @@ import pl.touk.nussknacker.engine.spel.ast.{
...
+      .get
+
+  private def addLabelToDictKeyExpression(
+      process: CanonicalProcess,
+      optionalExpressionTypingInfo: Option[ExpressionTypingInfo],
+      nodeExpressionId: NodeExpressionId
+  ) =
+    optionalExpressionTypingInfo match {
+      case Some(DictKeyWithLabelExpressionTypingInfo(key, label, _)) => Expression.dictKeyWithLabel(key, label)
+      case _ =>
```
arkadius, line 72: I'm not sure if we should handle `None` the same way. Substition is done even if you got validation error on the compilation step. I think that we should return something like missing label that user could fix on the FE side by choosing another label. WDYT?

interpreter/src/main/scala/pl/touk/nussknacker/engine/spel/parser/DictKeyWithLabelExpressionParser.scala
```diff
@@ -0,0 +1,64 @@
...
+import pl.touk.nussknacker.engine.api.definition.{AdditionalVariable => _}
+import pl.touk.nussknacker.engine.api.expression.{Expression => CompiledExpression, _}
+import pl.touk.nussknacker.engine.api.generics.ExpressionParseError
+import pl.touk.nussknacker.engine.api.typed.typing
+import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypingResult}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.engine.graph.expression.Expression.DictKeyWithLabelExpression
+import pl.touk.nussknacker.engine.spel.SpelExpressionParseError.KeyWithLabelExpressionParsingError
+
+case class DictKeyWithLabelExpressionTypingInfo(key: String, label: Option[String], typingResult: TypingResult)
```
arkadius, line 17: I see that `typingResult` is always `Unknown`. If so, let's remove it from the parameters and move to the method.

arkadius (approved):
Good job

PR #5379, opened 2024-01-16
Fix serialization of database lookup enricher

paw787878:
## Checklist before merge

### DIFF
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
@@ -21,7 +21,6 @@ class MetaDataProviderFactory extends Serializable {
       props.put(k, v)
     }
     val getConnection: () => Connection = () => {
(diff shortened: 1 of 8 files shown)


components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
```diff
@@ -13,23 +13,23 @@ object MetaDataProviderFactory {
...
-      new DriverDataSource(
-        dbPoolConfig.url,
-        dbPoolConfig.driverClassName,
-        props,
-        dbPoolConfig.username,
-        dbPoolConfig.password
-      )
-    }
     val getConnection: () => Connection = () => {
+      val ds = ThreadUtils.withThisAsContextClassLoader(getClass.getClassLoader) {
```
paw787878, line 24: czy tutaj zrobiłem ok, nie jestem pewien czy to nie będzie mieć negatywnych konsekwencji

arkadius:
I'm closing it. As we discuss, we should use flink test scenario runner which doesn't need to make this component serializable

PR #5384, opened 2024-01-16
Make default timeout of RichKafkaConsumer methods public

piotrp:
## Checklist before merge


utils/kafka-test-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/RichKafkaConsumer.scala
```diff
@@ -76,7 +76,7 @@ class RichKafkaConsumer[K, M](consumer: Consumer[K, M]) extends LazyLogging {
 }
 
 object RichKafkaConsumer {
-  private val DefaultSecondsToWait = 30
+  val DefaultSecondsToWait: Int = 30
```
lciolecki, line 79: Can you add comment why it's public? Some can change it again as private in the future..

piotrp: It made my IDEA break, probably because you use it as a default value in public functions. But I want to do a few more checks.

PR #5391, opened 2024-01-17
[Nu-1407] Improve validation in numeric toNumber helper method

DeamonDev:
## Checklist before merge

### DIFF
--- a/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
+++ b/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
@@ -60,18 +60,15 @@ trait numeric extends MathUtils with HideToString {
   @GenericType(typingFunction = classOf[ToNumberTypingFunction])
   def toNumber(@ParamName("stringOrNumber") stringOrNumber: Any): java.lang.Number = stringOrNumber match {
     case s: CharSequence =>
-      val ss                                       = s.toString
-      val tryByte: Try[java.lang.Byte]             = Try(java.lang.Byte.parseByte(ss))
-      val tryShort: Try[java.lang.Short]           = Try(java.lang.Short.parseShort(ss))
-      val tryInteger: Try[java.lang.Integer]       = Try(java.lang.Integer.parseInt(ss))
-      val tryLong: Try[java.lang.Long]             = Try(java.lang.Long.parseLong(ss))
-      val tryFloat: Try[java.lang.Float]           = Try(java.lang.Float.parseFloat(ss))
-      val tryDouble: Try[java.lang.Double]         = Try(java.lang.Double.parseDouble(ss))
-      val tryBigDecimal: Try[java.math.BigDecimal] = Try(new java.math.BigDecimal(ss))
-      val tryBigInteger: Try[java.math.BigInteger] = Try(new java.math.BigInteger(ss))
+      val ss = s.toString
 
-      val tries: List[Try[java.lang.Number]] =
-        List(tryByte, tryShort, tryInteger, tryLong, tryFloat, tryDouble, tryBigDecimal, tryBigInteger)
+      val tries: List[Try[java.lang.Number]] = List(
+        Try(java.lang.Integer.parseInt(ss)),
+        Try(java.lang.Long.parseLong(ss)),
+        Try(java.lang.Double.parseDouble(ss)),
+        Try(new java.math.BigDecimal(ss)),
+        Try(new java.math.BigInteger(ss))
+      )
 
       val firstSuccess: Option[Number] = tries
         .collectFirst { case Success(value) =>
--- a/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
+++ b/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
@@ -62,6 +62,7 @@ trait numeric extends MathUtils with HideToString {
     case s: CharSequence =>
       val ss = s.toString
 
+      // we pick the narrowest type as possible to reduce the amount of memory and computations overheads
       val tries: List[Try[java.lang.Number]] = List(
         Try(java.lang.Integer.parseInt(ss)),
         Try(java.lang.Long.parseLong(ss)),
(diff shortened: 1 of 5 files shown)


utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
```diff
@@ -57,7 +59,27 @@ trait numeric extends MathUtils with HideToString {
   @Documentation(description = "Parse string to number")
   @GenericType(typingFunction = classOf[ToNumberTypingFunction])
   def toNumber(@ParamName("stringOrNumber") stringOrNumber: Any): java.lang.Number = stringOrNumber match {
-    case s: CharSequence     => new java.math.BigDecimal(s.toString)
+    case s: CharSequence =>
+      val ss                                       = s.toString
+      val tryByte: Try[java.lang.Byte]             = Try(java.lang.Byte.parseByte(ss))
```
arkadius, line 64: I think that the better approach will be behaving as java behaves when someone declares literal number - I mean skipping byte, short and float cases. I think that these cases are only for some hard computations when you need to save memory. In most business cases, they introduce more noise than it benefits.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
```diff
@@ -57,7 +59,27 @@ trait numeric extends MathUtils with HideToString {
...
+      val tryShort: Try[java.lang.Short]           = Try(java.lang.Short.parseShort(ss))
+      val tryInteger: Try[java.lang.Integer]       = Try(java.lang.Integer.parseInt(ss))
+      val tryLong: Try[java.lang.Long]             = Try(java.lang.Long.parseLong(ss))
+      val tryFloat: Try[java.lang.Float]           = Try(java.lang.Float.parseFloat(ss))
+      val tryDouble: Try[java.lang.Double]         = Try(java.lang.Double.parseDouble(ss))
+      val tryBigDecimal: Try[java.math.BigDecimal] = Try(new java.math.BigDecimal(ss))
+      val tryBigInteger: Try[java.math.BigInteger] = Try(new java.math.BigInteger(ss))
+
+      val tries: List[Try[java.lang.Number]] =
```
arkadius, line 73: Let's skip local variables above and keep everything in the inlined list. Thanks to that we avoid the mistake that this list misses some case from above.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
```diff
@@ -57,7 +59,27 @@ trait numeric extends MathUtils with HideToString {
...
+        List(tryByte, tryShort, tryInteger, tryLong, tryFloat, tryDouble, tryBigDecimal, tryBigInteger)
+
+      val firstSuccess: Option[Number] = tries
+        .collectFirst { case Success(value) =>
+          value
+        }
+        .orElse(Some(new java.math.BigDecimal(ss)))
```
arkadius, line 80: `orElse(Some(...).get` is the same as `getOrElse(...)`

utils/default-helpers/src/test/scala/pl/touk/nussknacker/engine/util/functions/CollectionUtilsSpec.scala
```diff
@@ -189,9 +189,9 @@ class CollectionUtilsSpec extends AnyFunSuite with BaseSpelSpec with Matchers {
     evaluateType("#COLLECTION.sum({#int, #long})", variables) shouldBe "Number".valid
   }
 
-  test("sum should fall back to Double on unknown Number type") {
```
arkadius, line 192: Please add tests:
1. doing the same but using simple plus operator
2. showing that we behave different than before in the runtime

DeamonDev: I've added tests

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
DeamonDev: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
DeamonDev: Done

utils/default-helpers/src/test/scala/pl/touk/nussknacker/engine/util/functions/CollectionUtilsSpec.scala
arkadius: What about tests mentioned in 2. point?

utils/default-helpers/src/test/scala/pl/touk/nussknacker/engine/util/functions/NumericUtilsSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import cats.implicits._
+
+import java.math.BigInteger
+
+class NumericUtilsSpec extends AnyFunSuite with BaseSpelSpec with Matchers {
+
+  test("numeric") {
```
arkadius, line 11: I would call it `test("string to number conversion should return narrow numeric types in the runtime")` or sth similar. Thanks to that we would know in the future, why we introduced this change.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/numeric.scala
```diff
@@ -57,7 +59,23 @@ trait numeric extends MathUtils with HideToString {
   @Documentation(description = "Parse string to number")
   @GenericType(typingFunction = classOf[ToNumberTypingFunction])
   def toNumber(@ParamName("stringOrNumber") stringOrNumber: Any): java.lang.Number = stringOrNumber match {
-    case s: CharSequence     => new java.math.BigDecimal(s.toString)
+    case s: CharSequence =>
+      val ss = s.toString
+
+      val tries: List[Try[java.lang.Number]] = List(
```
arkadius, line 65: I would add a comment that we pick the narrowest type as possible to reduce the amount of memory and computations overheads ,

arkadius (approved):
LGTM, I added some small comments imrovements that could help us to understand why we introduced changes in the `toNumber` runtime.

PR #5399, opened 2024-01-18
ValidationExpressionParameterValidatorToCompile to allow config from file or from AdditionalUIConfigProvider to support validation

wrzontek:
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -256,6 +255,31 @@ case object JsonValidator extends ParameterValidator {
 
 }
 
+case class ValidationExpressionParameterValidatorToCompile(
```
arkadius, line 258: Can you move `ValidationExpressionParameterValidator` into the `interpreter` module? It is important because:
1. `CompiledExpression` leaked to the `component-api` and I want it to came back into `extensions-api`
2. In the future I would like to split `Parameter` into two classes: accessible for the component developers and second one, ready to use by compiler, with defined properties - see TODO there. This change will help to achieve that

I know that there will be a little issue with the fact that `ParameterValidator` is sealed trait and we use it with circe, but I belive that you will handle it.

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
```diff
@@ -62,8 +55,17 @@ object FragmentParameterValidator {
       fragmentParameter.name,
     )
 
-    val validationExpressionResult =
-      compileExpressionValidator(fragmentParameter, validationContext(fragmentParameter.name), expressionCompiler)
+    val validationExpressionResult = fragmentParameter.valueCompileTimeValidation
+      .map(validation =>
+        expressionCompiler
```
arkadius, line 60: I think that we don't need it anymore. Take a look that `FragmentCompleteDefinitionExtractor.extractParametersDefinition` is invoked always by `NodeCompiler`. We can leverage the fact, that parameters definition is "compiled" there, so we can just return `ValidationExpressionParameterValidatorToCompile`  in the definition. After that we can remove dependency to `ExpressionCompiler` and merge some `Fragment*DefinitionExtractor`s

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
wrzontek: do you mean for me to also move `CompileExpression` to `extensions-api`? from what i quickly checked, it doesn't seem to be a simple move, and it shouldn't be a part of this PR imo

interpreter/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
wrzontek: good idea

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
arkadius: You are right, I forgot that it accidentally leaded also into here: https://github.com/TouK/nussknacker/pull/5373/files#diff-dcbf8a2291f73522ea043e9e7d33ddb0dc8317c3c513c90b9c9a7a13a61e7e7f 
I already have a PR that fixes it. I'll move this class into the extension-api after merge of this PR

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/fragment/FragmentParametersDefinitionExtractor.scala
```diff
@@ -18,18 +18,22 @@ import pl.touk.nussknacker.engine.definition.component.parameter.defaults.{
...
+}
 import pl.touk.nussknacker.engine.graph.expression.Expression
 import pl.touk.nussknacker.engine.graph.node.FragmentInputDefinition.FragmentParameter
 import pl.touk.nussknacker.engine.graph.node.{FragmentInput, FragmentInputDefinition}
 
 /*
- * This class exists as a more lightweight alternative to FragmentParametersCompleteDefinitionExtractor - it doesn't rely on ExpressionCompiler and ValidationContext
- * However, it doesn't validate the parameters' initialValue and valueEditor and doesn't extract (and validate the correctness of) the parameters' validators - use in cases where it's not needed
+ * This class doesn't validate the parameters' initialValue and valueEditor (e.g. values can be of incorrect type).
+ * If they are not validated it is not a breaking issue anyway as a process using these incorrect values will fail validation.
```
arkadius, line 31: Is it still the true? IMO we should validate things in the place where user can fix them. What is the reason that we don't validate it?

arkadius (approved):
Good job, I left one question.

interpreter/src/main/scala/pl/touk/nussknacker/engine/definition/fragment/FragmentParametersDefinitionExtractor.scala
wrzontek: well it would still require validationcontext and expression compiler which was the initial reason behind the `FragmentCompleteDefinitionExtractor` and `FragmentWithoutValidatorsDefinitionExtractor`, maybe i should have left that in the comment

PR #5419, opened 2024-01-22
Remove system properties from merged model config

piotrp:
## Checklist before merge

### DIFF
--- a/extensions-api/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
+++ b/extensions-api/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.engine.modelconfig
 
-import com.typesafe.config.{Config, ConfigFactory}
+import com.typesafe.config.impl.ConfigImpl
+import com.typesafe.config.{Config, ConfigFactory, ConfigResolveOptions}
 import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.ConfigWithUnresolvedVersion
 import pl.touk.nussknacker.engine.modelconfig.ModelConfigLoader.defaultModelConfigResource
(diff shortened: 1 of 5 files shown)


extensions-api/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
```diff
@@ -100,8 +101,15 @@ abstract class ModelConfigLoader extends Serializable with LazyLogging {
...
       and have baseUrl taken from application config
      */
-    // We want to respect overrides (like system properties) and standard fallbacks (like reference.conf)
-    ConfigFactory.load(classLoader, withFallbackToDefaults(inputConfig, classLoader))
+    // We want to respect overrides (via environment variables) and standard fallbacks (like reference.conf).
+    // Code below works like 'ConfigFactory.load(ClassLoader,Config)' but it:
+    // - doesn't include system properties
+    // - always resolves overrides from environment variables (works as if 'config.override_with_env_vars' was always enabled')
+    ConfigFactory
+      .systemEnvironmentOverrides()
```
arkadius, line 109: IMO we should remove these overrides for the sake of clarity that it is done only on the designer stage but is not mandatory for now, feel free to omit this.

arkadius (approved):

extensions-api/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
piotrp: Ok, I will remove them

extensions-api/src/main/scala/pl/touk/nussknacker/engine/modelconfig/ModelConfigLoader.scala
```diff
@@ -100,8 +101,11 @@ abstract class ModelConfigLoader extends Serializable with LazyLogging {
       service1Url: ${baseUrl}/service1
       and have baseUrl taken from application config
      */
-    // We want to respect overrides (like system properties) and standard fallbacks (like reference.conf)
-    ConfigFactory.load(classLoader, withFallbackToDefaults(inputConfig, classLoader))
+    // We want to respect standard fallbacks (like reference.conf), but leave resolution of overrides and environment
```
arkadius, line 104: Actually, resolution of environment variables is sometime that here as well (behind `resolve(ConfigResolveOptions.defaults)`. We only skip system properties resolution and environment variables override mechanism

piotrp: Can you give an example? I tried to write a test for that (in `DefaultModelConfigLoaderTest`) and it fails in the call you mentioned with `Could not resolve substitution to a value: ${PATH}`.

Modified defaultModelConfig.conf:
```conf
{
  property1: shouldBeOverridenToo
  property2: ${property1}Suffix
  otherProperty: true
  envPathProperty: ${PATH}
}
```

Test:
```scala
  test("should resolve environment variables") {
    val config = LocalModelData(inputConfig, List.empty).modelConfig
    val envPath = System.getenv("PATH")

    envPath shouldNot be(null)
    config.getString("envPathProperty") shouldBe envPath
  }
```

arkadius: strange, I do the same on your branch and it works for me :) https://github.com/TouK/nussknacker/pull/5421/files

PR #4687, opened 2023-09-18
Align CustomAction api with predefined actions api (deploy, cancel)

gskrobisz:
## Describe your changes
The first step of CustomAction api refactor
1. Change type returned by invokeCustomAction to `Future[CustomActionResult]` to be coherrent with predefined deploy/cancel actions.
2. Move deployment comment validation from `ManagementResources` deploy/cancel routes into `DeploymentService`. On the service level the comment is a parameter of the invoked action. Here service is responsible for the whole transaction: all validations, interactions with repositories and invoking action itself.

Next steps:
3. Move `CustomActionInvokerService` to `DeploymentService` (according to the TODO comment)
4. Implement cancel/deploy as "custom action".
5. Add validators to `CustomActionParameter`. The validators are expected behave in the same way as in additional properties or node parameters. Adding validator to the parameter is available via custom action configuration.
6. Replace `deploymentComment` validation (configured by `DeploymentCommentSettings.validationPattern`) with a parameter validator.
7. Add dynamic validation like `api/nodes/.../validation` or `api/properties/.../validation` and handle validation results in FE

Also in next steps: introduce asynchronous actions.

Profits:
1. one coherrent behaviour of parameter validation for action, node and property
1. one coherrent api and the same approach for all actions, both predefined and custom
2. a control over action parameters validation, each NU instance is able to define its own validation rules (e.g. for cancel and deploy)
3. no hardcoded validations (e.g. deploymentComment)

## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [x] Code is cleaned from temporary changes and commented out lines
- [ ] Code formatting checked
- [ ] Parts of the code that are not easy to understand are documented in the code
- [x] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [x] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [x] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #5438, opened 2024-01-25
Move invokeCustomAction to DeploymentService

gskrobisz:
## Describe your changes

Previous changes: https://github.com/TouK/nussknacker/pull/4687

Here:
3. Move CustomActionInvokerService to DeploymentService (according to the TODO comment)
- implementtion of `invokeCustomAction` is moved to `DeploymentServiceImpl`
- invokeCustomAction handles errors in the same way as other methods, there is no need for CustomActionError
- no need for `CustomActionNotImplemented`, missing implementation `???` is handled via `NuDesignerErrorToHttp`
- `ProcessActionType` has fixed values, 

Next steps:
4. Implement cancel/deploy as "custom action"
- deployment comment should be provided as action parameter, see routing changes in https://github.com/TouK/nussknacker/pull/4720/files#diff-2369735d4142be8fa1e8634e2605a86c838798201e225376be8fc465fd6ec0feR136
- change http response from Conflict to Forbidden for actions that are illegal or not allowed (e.g. because of invalid process status) 

## Checklist before merge


engine/development/deploymentManager/src/main/scala/pl/touk/nussknacker/development/manager/DevelopmentDeploymentManagerProvider.scala
```diff
@@ -172,29 +172,18 @@ class DevelopmentDeploymentManager(actorSystem: ActorSystem)
...
+  ): Future[CustomActionResult] = {
+    val processName = actionRequest.processVersion.processName
+    val statusOpt = customActionStatusMapping
+      .collectFirst { case (customAction, status) if customAction.name == actionRequest.name => status }
+
+    statusOpt match {
+      case Some(newStatus) =>
+        asyncChangeState(processName, newStatus)
+        Future.successful(CustomActionResult(actionRequest, s"Done ${actionRequest.name}"))
+      case _ => ???
```
arkadius, line 184: IMO it should be inside Future.apply - the whole block probably should be. Generally, it is not a good idea to throw exceptions inside a method that declares that returns the `Future`. Invoking service can handle wrong such a case.

arkadius (commented):
Some minor comments added

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NuDesignerErrorToHttp.scala
```diff
@@ -16,26 +16,30 @@ object NuDesignerErrorToHttp extends LazyLogging with FailFastCirceSupport {
...
     import akka.http.scaladsl.server.Directives._
     ExceptionHandler { case NonFatal(e) =>
-      complete(errorToHttpResponse(e))
+      complete(errorToHttpResponse(unwrapException(e)))
     }
   }
 
-  private def errorToHttpResponse: PartialFunction[Throwable, HttpResponse] = {
+  def unwrapException(e: Throwable): Throwable =
+    if (Option(e.getCause).nonEmpty) {
```
arkadius, line 24: IMO it's not a good approach. Let's say that someone throw NuDesignerError with some cause. Will be it handled correctly? I thought about:
```
e match {
case SomeSpecificClassNameThatWeAreSureThatWeWantExtractCause => e.getCasue
case _ => e
}
```
By `SomeSpecificClassNameThatWeAreSureThatWeWantExtractCause` you said something about BoxedException? Am I right?

arkadius (approved):
Good job

PR #5447, opened 2024-01-26
Fixup: jdk downgraded from 17 to 11 in lite runner image (due to issues with InaccessibleObjectException)

raphaelsolarski:
The lack of open java.base/java.util module caused runtime issues with the following spel expression constructs:
```
* {"a": 1}.getOrDefault(“a”, -1)
* {"a": 1}.get("a")
* {1,2,3}.get(0)
* {1,2,3}.lastIndexOf(3)
* {1,2,3}.indexOf(3)
* {1,2,3}.contains(1) 
* {1,2,3}.toString
```

Alternatively we can add `--add-opens=java.base/java.util=ALL-UNNAMED` to `engine/lite/runtime-app/src/universal/bin/nu-engine-entrypoint.sh` but still we will have different jdk in CI and different in a production use.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -384,14 +384,10 @@ def flinkLibScalaDeps(scalaVersion: String, configurations: Option[String] = Non
 
 lazy val commonDockerSettings = {
   Seq(
-    dockerBaseImage       := forScalaVersion(
-      scalaVersion.value,
-      "eclipse-temurin:17-jre-jammy",
-      (
-        2,
-        12
-      ) -> "eclipse-temurin:11-jre-jammy" // jre11, cause for jdk17 minimum scala version is 2.12.15, we use 2.12.10
-    ),
+    // designer should run on java11 since it may run Flink in-memory-cluster, which does not support newer java and we want to have same jre in both designer and lite-runner
+    // to make analysis of problems with jre compatibility easier using testing mechanism and embedded server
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -384,14 +384,7 @@ def flinkLibScalaDeps(scalaVersion: String, configurations: Option[String] = Non
...
   Seq(
-    dockerBaseImage       := forScalaVersion(
-      scalaVersion.value,
-      "eclipse-temurin:17-jre-jammy",
-      (
-        2,
-        12
-      ) -> "eclipse-temurin:11-jre-jammy" // jre11, cause for jdk17 minimum scala version is 2.12.15, we use 2.12.10
-    ),
+    dockerBaseImage       := "eclipse-temurin:11-jre-jammy",    // designer should run on java11 since it may run Flink in-memory-cluster, which does not support newer java and we want to have same jdk in both designer and lite-runner
```
arkadius, line 387: ... to make analysis of problems with jre compatibility easier using testing mechanism and embedded server
+ Add a TODO that we want to support jre >= 17 but before that we have to remove scala 2.12 compatibility and we should handle opening of modules for spel reflectional access to class memebers

arkadius (approved):
I left one clarification that would be nice to see to remember why we pick this solution

PR #5486, opened 2024-01-31
Cleaned up typos and syntax mistakes in scenarios_authoring docs

Diamekod0221:
## Describe your changes
A few docs improvements so they're more readable. The for each section should probably also be changed, but my technical expertise isn't good enough to write it and the documentation isn't understandable for me.

## Checklist before merge


docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
 
 # Basic Components
 
-Nodes work with a data records. They can produce, fetch, send, collect data or organize data flow. Each node has at least two parameters: `Name` and `Description`. Name has to be unique in a scenario. Description is a narrative of your choice.
+Nodes work with data records. They can produce, fetch, send and collect data or organize the flow. Each node has at least two parameters: `Name` and `Description`. Name has to be unique in a scenario. Description is a narrative of your choice.
```
paw787878, line 7: maybe it is better to write "data flow" here, to be more precise, even if this means we use word data twice

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 
-As you can see in the `variable` configuration form below, Nussknacker inferred the data type of the `#input` variable from the information already available to Nussknacker.
+As you can see in the `variable` configuration form below, Nussknacker has inferred the data type of the `#input` variable. Nussknacker can do this based on info stored in the previous components.
 
 ![alt_text](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_variable1.png "Variable declaration form")
 
 
-In the next example `#input` variable is used to create an expression returning a boolean value. If the input Kafka topic contains JSON objects and they contain `operation` field, the value of this field can be obtained in the following way:
+In the next example `#input` variable is used to create an expression returning a boolean value. If the input Kafka topic contains JSON objects and they contain an `operation`, the value can be obtained using the following pattern:
```
wrzontek, line 26:
```suggestion
In the next example `#input` variable is used to create an expression returning a boolean value. If the input Kafka topic contains JSON objects which contain a field named `operation`, the value of the field can be obtained in the following way:
```

paw787878, line 26: maybe "the following expression" instead of "the following pattern"

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
-Most of the nodes, with source and sink nodes being notable exceptions, have both input and at least one output flow.
+Most of the nodes have both input and at least one output flow. Source and sink nodes are an exception.
 
 Sinks and filters can be disabled by selecting `Disable` checkbox.
 
 &nbsp;
 ## Variable
 
-A Variable component is used to declare a new variable; in the simplest form a variable declaration looks like in the example  below. As the event was read from the Kafka topic, the `#input` variable holds its content and  its value is assigned to a newly declared `myFirstVariable` variable.
+A Variable component is used to declare a new variable. In its simplest form a variable declaration looks like the example  below. As the event has been read from a Kafka topic, the `#input` variable holds event contents. Event value is assigned to a newly declared `myFirstVariable` variable.
```
wrzontek, line 16:
```suggestion
A Variable component is used to declare a new variable. In its simplest form a variable declaration looks like the example  below. As the event has been read from a Kafka topic, the `#input` variable holds event contents. Here the event's value is assigned to a newly declared `myFirstVariable` variable.

File: docs/scenarios_authoring/BasicNodes.md [hunk 4]
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 ![alt_text](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_variable0.png   "Scenario with variable declaration")
 
 
-As you can see in the `variable` configuration form below, Nussknacker inferred the data type of the `#input` variable from the information already available to Nussknacker.
+As you can see in the `variable` configuration form below, Nussknacker has inferred the data type of the `#input` variable. Nussknacker can do this based on info stored in the previous components.
```
wrzontek, line 21:
```suggestion
As you can see in the `variable` configuration form below, Nussknacker has inferred the data type of the `#input` variable. Nussknacker can do this based on the information available from the previous components.
```

docs/scenarios_authoring/BasicNodes.md
paw787878, line 26: maybe "an 'operation' field" would be better, as it explicitly states what is this 'operation' entity

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
+The same outcome can be achieved using a plain `Variable` component. Just make sure to write a valid SpEL expression.
 
 
 ![alt_text](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_mapmariable1.png "record-variable declaration using a plan Variable component")
 
 &nbsp;
 ## Filter
 
-Filter passes records which satisfy the filtering condition to `true sink`.
+Filters let through records that satisfy a filtering condition.
```
paw787878, line 52: maybe "Filters let through only records ..."

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 ![filter graph single](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_filter0.png)
 
-You can additionally define `false sink`. Records from the `source` which meet the filter's condition go to the `true sink`, and others go to the `false sink`.
+You can additionally also define an additional `false sink`. Records from the `source` which meet the filter's conditions are going to be redirected to the `true sink`, while others end up in the `false sink`.
```
wrzontek, line 56:
```suggestion
You can additionally also define an additional `false sink`. Records from the `source` which meet the filter's conditions are going to be directed to the `true sink`, while others end up in the `false sink`.
```

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 It distributes incoming records among output branches in accordance with the filtering conditions configured for those branches.
 
 ![choice graph](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_choice0.png)
 
-Each record from the `source` is tested against the condition defined for outgoing node. If `#input.color` is `blue` record goes to the `blue sink`.  
-If `#input.color` is `green` record goes to the `green sink`. For every other value record goes to the `sink for others` because condition `true` is always true.
-Order of evaluation of conditions is the same as visible in form. You can modify the order using drag & drop functionality.
-Order is also visible on graph in edges description as a number. Be aware that layout button can change displayed order of nodes, but it has no influence on order of evaluation.
+After a record leaves`source`. When it arrives in `choice`, the record value is tested against each of the defined conditions.  If `#input.color` is `blue`, the record ends up in `blue sink`.
```
wrzontek, line 72:
```suggestion
After a record leaves`source` it arrives in `choice`, the record's value is tested against each of the defined conditions.  If `#input.color` is `blue`, the record ends up in `blue sink`.  
```

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 ![choice graph](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_choice0.png)
 
-Each record from the `source` is tested against the condition defined for outgoing node. If `#input.color` is `blue` record goes to the `blue sink`.  
-If `#input.color` is `green` record goes to the `green sink`. For every other value record goes to the `sink for others` because condition `true` is always true.
-Order of evaluation of conditions is the same as visible in form. You can modify the order using drag & drop functionality.
-Order is also visible on graph in edges description as a number. Be aware that layout button can change displayed order of nodes, but it has no influence on order of evaluation.
+After a record leaves`source`. When it arrives in `choice`, the record value is tested against each of the defined conditions.  If `#input.color` is `blue`, the record ends up in `blue sink`.  
+If `#input.color` is `green`,the record will be sent to the `green sink`. For every other value, the record is sent to `sink for others` because condition `true` is always true.
```
wrzontek, line 73:
```suggestion
If `#input.color` is `green`,the record is sent to the `green sink`. For every other value, the record is sent to `sink for others` because condition `true` is always true.
```

docs/scenarios_authoring/BasicNodes.md
paw787878, line 72: First sentence in this line seems to end abruptly

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
+Order of evaluation of conditions is the same as is visible in the configuration form - top to bottom. You can modify the order using the drag & drop functionality.
+Order is also visible on the designer graph in an edge's (arrow's) description as a number. Be aware that the layout button can change displayed order of nodes, but it has no influence on order of evaluation.
```
wrzontek, line 75: (arrow's)  doesn't clarify much, either expand it `(the arrow connecting nodes)` it or remove it

docs/scenarios_authoring/Intro.md
```diff
@@ -6,34 +6,34 @@ sidebar_position: 1
 
 ## Intended audience
 
-Nussknacker provides a drag and drop visual authoring tool (Nussknacker Designer) allowing to define decision algorithms – we call them scenarios – without the need to write code. This document is intended for those who will use Nussknacker Designer to configure processing scenarios. Nussknacker is a low-code platform; prior knowledge of SQL, JSON and concepts like variables and data types will help master authoring of the data processing scenarios in Nussknacker. 
+Nussknacker provides a drag and drop visual authoring tool (Nussknacker Designer). This tool allows the user to define decision algorithms – we call them scenarios – without the need to write code. This document is intended for those who will use Nussknacker Designer to configure scenario processing. Nussknacker is a low-code platform; prior knowledge of SQL, JSON and familiarity with concepts like variables and data types will help master data processing with Nussknacker.
```
wrzontek, line 9:
```suggestion
Nussknacker provides a drag and drop visual authoring tool (Nussknacker Designer). This tool allows the user to define decision algorithms – we call them scenarios – without the need to write code. This document is intended for those who will use Nussknacker Designer to configure processing scenarios. Nussknacker is a low-code platform; prior knowledge of SQL, JSON and familiarity with concepts like variables and data types will help master data processing with Nussknacker. 
```

docs/scenarios_authoring/Intro.md
```diff
@@ -6,34 +6,34 @@ sidebar_position: 1
...
 
 **Please try the [Demo](/quickstart/demo) to quickly understand how to move around Nussknacker Designer, create a simple scenario and see SpEL in action.**
 
 &nbsp;
 ## Nussknacker scenario diagram
 
-Nussknacker scenario represents a decision algorithm in a form of a graph. The decisions are based on data, which can be anything from clicks on a website, bank transactions, to readings from sensors. These data are processed by Nussknacker, according to the scenario (algorithm). The results generated by Nussknacker are also data; this time they communicate the result or decisions.
+Nussknacker scenario represents a decision algorithm in the form of a graph. The decisions are based on data, which can be anything from clicks on a website, to bank transactions, to readings from sensors. This data is processed by Nussknacker, according to the defined scenario(algorithm). Applying the scenario template to the input the data yields some output from Nussknacker. Either result data or information about decisions to be made.
```
wrzontek, line 16: it's kinda awkward, maybe something like this:?

The input data passes through the steps defined in the scenario. The output is either processed data or information about the actions (decisions) made, depending on the scenarion.

docs/scenarios_authoring/Intro.md
```diff
@@ -6,34 +6,34 @@ sidebar_position: 1
...
 
-Every scenario starts with a source of data because we have to know what kind of data we want to work with. The rest of the scenario is a sequence (Directed Acyclic Graph or DAG to be more precise) of different nodes:
+Every scenario starts with a source of data. We have to specify what kind of data we want to work with. The rest of the scenario is a sequence (Directed Acyclic Graph or DAG to be more precise) of different nodes:
 - flow control functions: filter, switch, split etc.
 - data enrichments from external sources (JDBC, OpenAPI)
 - aggregates in different types of time windows (available with Flink engine)
 - custom, tailor-made components, which extend default functionality
-- and more
+- and additional features
```
wrzontek, line 23:
```suggestion
- and more
```

wrzontek (commented):
this probably would be better handled by someone more involved with docs than random developers, but i left some comments anyway

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: "Nodes can manipulate the data flow by producing, fetching, sending or collecting data"?

docs/scenarios_authoring/Intro.md
```diff
@@ -86,7 +87,7 @@ If the event which arrived to some node originally came from the Kafka topic, th
 Consult Kafka [documentation](https://kafka.apache.org/33/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html) for the exact meaning of those fields. 
 
 
-The `#meta` variable carries meta information about the currently executed scenario. The following meta information elements are available:
+The `#meta` variable carries meta information about the scenario in execution. The following meta information elements are available:
```
paw787878, line 90: Question to nu-devs: does this tag access "dynamic" data, available only when scenario is being executed, or only static data such as scenario name? If it accesses only static data maybe here we should write "about the current scenario" without mentioning execution.

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: For my personal taste, pattern is more suitable here as it is more related to this object.field access concept.

Diamekod0221: I'm not sure whether this specification is needed for a non-technical user. How would he know what "field" means here? I think it makes things more complicated. I could probably settle for 'operation' value which is technically incorrect, but probably more understandable to most people.

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: Isn't that tad close to a tautology? :D

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: Right, typo. Will improve

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: I think your wording is very complex. You specify plural objects first but then switch to evaluating a singular field of a Singular JSON. I think that could be pretty confusing to a reader that isn't very familiarized with the context. Additionally, as written in the comment above, I think refraining from using "field" here would be better, even though I understand the willingness to be technically precise.

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: I agree with the suggestion to change, but I think "After that, the event's..." would be better, as Here sounds where ambiguous

docs/scenarios_authoring/BasicNodes.md
paw787878: the following pattern" is also ok

docs/scenarios_authoring/Intro.md
Diamekod0221: "Applying the scenario template to the input the data yields output from Nussknacker. The output is either processed data or information about the actions (decisions) made, depending on the scenario."?

docs/scenarios_authoring/BasicNodes.md
paw787878: I think that if someone knows what JSON is that person may also know what json object field is. In that case "they contain an `operation`" may mean to them less then "they contain an `operation` field". But it is more of an sugggestion, it can stay like this.

docs/scenarios_authoring/BasicNodes.md
paw787878: User might think that they let through also some other kind of records :) But on the other hand no one is gong to think that so it can stay

docs/scenarios_authoring/Intro.md
wrzontek: `scenario procesing` here implies we're processing scenarios
`processing scenarios` here implies we're defining scenarios how data is processed

we're doing the second one

docs/scenarios_authoring/BasicNodes.md
ZbyszekMM, line 16: Variables are not easy to. We ended up with the following statement on variables [here](https://nussknacker.io/documentation/docs/1.12/scenarios_authoring/Intro/#variables): _Variables are attributes of the data record, they do not exist by themselves._

In this context stating _event's value_ will be confusing. Event does not have value, it has attributes in the form of variables. For me correct formulations would be _content of #input variable is assigned to myFirstVariable_.

Finally, this fragment of docs was written when we did not have R-R processing mode, so term "event" is used. May be it could be replaced with more neutral "data record" term, or even better, we could run away from mentioning Kafka here.

docs/scenarios_authoring/Intro.md
ZbyszekMM: what about _use .. Designer to configure scenario's logic_ rather than _use .. Designer to configure processing scenarios_?

docs/scenarios_authoring/Intro.md
```diff
@@ -6,16 +6,16 @@ sidebar_position: 1
...
+Nussknacker scenario represents a decision algorithm in the form of a graph. The decisions are based on data, which can be anything from clicks on a website, to bank transactions, to readings from sensors. This data is processed by Nussknacker, according to the defined scenario(algorithm). Applying the scenario template to the input the data yields some output from Nussknacker. Either result data or information about decisions to be made.
 
-Every scenario starts with a source of data because we have to know what kind of data we want to work with. The rest of the scenario is a sequence (Directed Acyclic Graph or DAG to be more precise) of different nodes:
+Every scenario starts with a source of data. We have to specify what kind of data we want to work with. The rest of the scenario is a sequence (Directed Acyclic Graph or DAG to be more precise) of different nodes:
```
ZbyszekMM, line 18: These first 2 sentences (or the original first sentence) are not clear. We need to explain here that in the streaming processing mode one specifies Kafka topic from which events are read. In the R-R processing mode "it is on the diagram for aesthetical reason" (surely a better justification is needed). Anyway, current and previous formulation contains unfinished thought.

docs/scenarios_authoring/Intro.md
```diff
@@ -24,16 +24,16 @@ Every scenario starts with a source of data because we have to know what kind of
...
+In the **Request-Response** processing mode it is a request data record which enters a scenario. The best and easiest way to understand how this request will be processed by Nussknacker's scenario is to think of it as of Streaming mode with a singular event. All the considerations from the previous paragraph apply. The most important trait of a **Request-Response** scenario is that it's synchronous: some other computer system sends a request to Nussknacker and awaits a response. That request is the input to the scenario and the output - the decision - is a response. Since the other system is awaiting a response, there has to be exactly one. The natural question to ask is what will happen when there are nodes in the scenario which "produce" additional data records - for-each or split. The topic of how to handle such situations is covered  [here](RRDataSourcesAndSinks.md#scenario-response-in-scenarios-with-split-and-for-each-nodes).
```
ZbyszekMM, line 29: singular? What is this about? 

[https://www.askdifference.com/single-vs-singular/](https://www.askdifference.com/single-vs-singular/)

docs/scenarios_authoring/Intro.md
```diff
@@ -46,36 +46,37 @@ SpEL [Spring Expression Language](https://docs.spring.io/spring-framework/docs/3
...
+Every SpEL expression returns a value of one of the predefined SpEL data types, like integer, double, boolean, map, etc. Data types in Nussknacker can be a confusing aspect at the beginning, as depending on the context in which data is processed or displayed, different data type schemes will be used - please refer to the [SpEL Cheat Sheet page](Spel.md#data-types-and-structures) for more information. 
 
 In some contexts data type conversions may be necessary - conversion functions are described [here](Spel.md#type-conversions).
 
 &nbsp;
 ## Variables
 
-Nussknacker uses variables as containers for data. Variables have to be declared; a `variable` or `record-variable` component are used for this. Once declared, a hash sign `"#"` is used to refer to a variable from a SpEL expression. Variables are attributes of the data record, they do not exist by themselves. 
+Nussknacker uses variables as containers for data. Variables have to be declared; a `variable` or `record-variable` component is used for this. Once declared, a hash sign `"#"` is used to refer to a variable from a SpEL expression. Variables are part of a data record, they do not exist by themselves.
```
ZbyszekMM, line 71: _Variables are part of a data record_ vs original _attributes of a data record_. What is better? Which of these words carries idea of adding new variables to the data record? Btw, what is wrong with term _attribute_?

docs/scenarios_authoring/Intro.md
```diff
@@ -46,36 +46,37 @@ SpEL [Spring Expression Language](https://docs.spring.io/spring-framework/docs/3
...
 
-In the Streaming processing mode the `#input` variable is associated with the event which originally came from the Kafka topic. In the Request-Response processing mode the `#input` variable carries the request data of REST call which invoked Nussknacker scenario. Both in the Streaming and Request-Response cases some nodes not only terminate the input events, but also create new ones. As the result, the #input data record is no longer available after such a node, while the newly created data record (and the variable associated with it) is available "downstream". 
+In **Streaming** processing mode the `#input` variable is associated with an event that is provided by a Kafka topic. In the **Request-Response** processing mode the `#input` variable carries the request data of a REST call which invoked Nussknacker scenario. Both in the Streaming and Request-Response cases some nodes not only terminate input events but also create new ones. As the result, the `#input` data record is no longer available after such a node. The newly created data record (and the variable associated with it) is available "downstream" -in subsequent nodes.
```
ZbyszekMM, line 75: if we do not like formulation _which originally came from the Kafka topic_, may  be _which was originally read from the Kafka topic_ would be better. I would never say that _event was provided by a Kafka topic_.

docs/scenarios_authoring/Intro.md
ZbyszekMM: What kind of improvement does this formulation bring ("scenario in execution")? Is it really more clear?

docs/scenarios_authoring/BasicNodes.md
paw787878: You can write "Nodes can manipulate the data flow by producing, fetching, sending or collecting data", or it can stay as is

docs/scenarios_authoring/Intro.md
paw787878: I see that in meta we can access for instance #meta.properties.checkpointIntervalInSeconds, so it seems that this is an "dynamic" data.

We have here situation similar to "process" and "program" distinction. I would go with
"The `#meta` variable carries meta information about the execution of scenario"

docs/scenarios_authoring/Intro.md
Diamekod0221: I think singular is a better choice here as it outlines the fact that there's just one request processed. But maybe this wording is too complex for this kind of document.

docs/scenarios_authoring/Intro.md
Diamekod0221: This is just a matter of taste, I think both are technically ok.

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: I agree, but I would like to underline the fact illustrated in the graphic that first the record value is assigned to input and then to another variable. I have changed the wording to underline that and omit kafka.

docs/scenarios_authoring/Intro.md
Diamekod0221: I think maybe the wording is not optimal, but I would restrain from explaining specific modes here. I think the intro should just present an abstract concept of a data source from which nussknacker can take data.

docs/scenarios_authoring/Intro.md
Diamekod0221: Updated to attributes.

docs/scenarios_authoring/Intro.md
Diamekod0221: I have changed the wording to underline the fact that meta is a dynamically allocated variable.

Diamekod0221 (commented):
Changed the wordings a bit according to comments, still waiting for an approve.

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 
-As you can see in the `variable` configuration form below, Nussknacker inferred the data type of the `#input` variable from the information already available to Nussknacker.
+As you can see in the `variable` configuration form below, Nussknacker has inferred the data type of the `#input` variable. Nussknacker can do this based on the information available from the previous components.
 
 ![alt_text](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_variable1.png "Variable declaration form")
 
 
-In the next example `#input` variable is used to create an expression returning a boolean value. If the input Kafka topic contains JSON objects and they contain `operation` field, the value of this field can be obtained in the following way:
+In the next example `#input` variable is used to create an expression returning a boolean value. If the input Kafka topic contains JSON objects and they contain an `operation` field, the value of the field can be obtained using the following pattern:
```
ZbyszekMM, line 26: You removed mention of Kafka topic in line 16, so I think we should not mention Kafka here either

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 
 `#input.operation`
 
-Note that internally Nussknacker converts JSON’s object into SpEL’s map.
-
-
 
 ![alt_text](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_variable2.png "Screenshot_tooltip")
 
+Note that internally Nussknacker converts the JSON object into a SpEL map.
```
ZbyszekMM, line 34: We no longer use 'map' term - it is a[record](https://nussknacker.io/documentation/docs/scenarios_authoring/BasicNodes/#recordvariable) these days. 

Moreover, "internally converts" is debatable. I think Nu [maps scenario variables](https://nussknacker.io/documentation/docs/scenarios_authoring/Spel/#data-types-and-structures) to Java types; but I think this info is not critical here and can be omitted.

docs/scenarios_authoring/BasicNodes.md
```diff
@@ -4,77 +4,75 @@ sidebar_position: 2
...
 It distributes incoming records among output branches in accordance with the filtering conditions configured for those branches.
 
 ![choice graph](../autoScreenshotChangeDocs/Auto_Screenshot_Change_Docs_-_basic_components_-_choice0.png)
 
-Each record from the `source` is tested against the condition defined for outgoing node. If `#input.color` is `blue` record goes to the `blue sink`.  
-If `#input.color` is `green` record goes to the `green sink`. For every other value record goes to the `sink for others` because condition `true` is always true.
-Order of evaluation of conditions is the same as visible in form. You can modify the order using drag & drop functionality.
-Order is also visible on graph in edges description as a number. Be aware that layout button can change displayed order of nodes, but it has no influence on order of evaluation.
+After a record leaves`source` it arrives in `choice`, the record's value is tested against each of the defined conditions.  If `#input.color` is `blue`, the record ends up in `blue sink`.
```
ZbyszekMM, line 72: "the record's value is tested" --> a bit too imprecise:

- Variables are attributes of data records. 
- It is #input.color _do not know what_ (data field?) that is tested here.

docs/scenarios_authoring/Intro.md
```diff
@@ -6,16 +6,16 @@ sidebar_position: 1
...
 
 **Please try the [Demo](/quickstart/demo) to quickly understand how to move around Nussknacker Designer, create a simple scenario and see SpEL in action.**
 
 &nbsp;
 ## Nussknacker scenario diagram
 
-Nussknacker scenario represents a decision algorithm in a form of a graph. The decisions are based on data, which can be anything from clicks on a website, bank transactions, to readings from sensors. These data are processed by Nussknacker, according to the scenario (algorithm). The results generated by Nussknacker are also data; this time they communicate the result or decisions.
+Nussknacker scenario represents a decision algorithm in the form of a graph. The decisions are based on data, which can be anything from clicks on a website, to bank transactions, to readings from sensors. This data is processed by Nussknacker, according to the defined scenario(algorithm). Applying the scenario template to the input the data yields output from Nussknacker. The output is either processed data or information about the actions (decisions) made, depending on the scenario.
```
ZbyszekMM, line 16: What is a scenario templated? Why so complicated?

docs/scenarios_authoring/Intro.md
```diff
@@ -24,16 +24,16 @@ Every scenario starts with a source of data because we have to know what kind of
...
 
-SpEL [Spring Expression Language](https://docs.spring.io/spring-framework/docs/3.2.x/spring-framework-reference/html/expressions.html) is a powerful expression language that supports querying and manipulating data objects. What exactly does the term _expression_ mean and why SpEL is an _expression language_? In programming language terminology, an _expression_ is a combination of values and functions that are combined to create a new value. SpEL allows to write expressions only; therefore it is an expression language. Couple examples:
+SpEL [Spring Expression Language](https://docs.spring.io/spring-framework/docs/3.2.x/spring-framework-reference/html/expressions.html) is a powerful expression language that supports querying and manipulating data objects. What exactly does the term _expression_ mean and why is SpEL an _expression language_? In programming language terminology, an _expression_ is a union of values and functions that are joined to create a new value. SpEL only allows you to write expressions; therefore it is considered an expression language. A couple examples:
```
ZbyszekMM, line 36: A couple examples -> A couple of examples

Allegedly, the first form is colloquial American English (I do not feel an expert in this). Grammarly also complains on the current version.

Diamekod0221: Yes, it's a casual statement. Changed it so grammarly doesn't complain :)

docs/scenarios_authoring/Intro.md
Diamekod0221: I've simplified the introduction here. I think the template wording is very adequate - we define a template, based on which nussknacker processes flowing data. Each part is processed according to our scenario diagram - that is the definition of a template. I can accept some other word that would be suitable if you can come up with one

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: Changed it to be more precise about attributes usage

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: I think internally converts is ok, we don't need to dabble in specifics here. 
Changed map to record.

docs/scenarios_authoring/BasicNodes.md
Diamekod0221: removed.

docs/scenarios_authoring/Intro.md
ZbyszekMM: I consulted the template aspect with another team member and answer is that term template may be too difficult here, being another difficult concept to use.

PR #5491, opened 2024-02-01
[Nu-5490] Expand Custom Action with display policy

DeamonDev:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
```diff
@@ -91,6 +91,7 @@ object ProcessService {
     case class ValidateAndResolve(includeValidationNodeResults: Boolean = true) extends ValidationMode
   }
 
+  @JsonCodec case class GetLastlyDeployedVersionId(versionId: Option[Long])
```
philemone, line 94: Maybe we could remove verb `Get` from class name ? WDYT?

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/definition/package.scala
```diff
@@ -135,6 +141,14 @@ package object definition {
...
       name = action.name,
       allowedStateStatusNames = action.allowedStateStatusNames,
+      displayPolicy = action.displayPolicy match {
+        case Some(displayPolicy) =>
+          displayPolicy match {
+            case deployment.CurrentlyViewedProcessVersionIsDeployed =>
+              Some(UICustomActionDisplayPolicy.CurrentlyViewedProcessVersionIsDeployed)
+          }
+        case None => None
+      },
```
philemone, line 151: WDYT about changing this to simple `flatMap`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -133,6 +133,12 @@ class ProcessesResources(
             processService.getProcessActions(processId.id)
           }
         }
+      } ~ path("processes" / ProcessNameSegment / "lastlyDeployedVersionId") { processName =>
```
philemone, line 136: I might be just picking on this but `lastly` does not sound right for me in this context. What about `last` -> `lastDeployedVersionId`? And similar change in classes ans usages. You can ignore it if you will :upside_down_face:

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/definition/package.scala
DeamonDev: I think `map` even would be better!

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ProcessService.scala
DeamonDev: Yeahh, that sounds much better :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
DeamonDev: Yeahh, that sounds much better :)

designer/client/src/components/toolbars/status/ProcessInfo.tsx
```diff
@@ -28,6 +33,9 @@ const ProcessInfo = memo(({ id, buttonsVariant, children }: ToolbarPanelProps) =
     const unsavedNewName = useSelector((state: RootState) => getProcessUnsavedNewName(state));
     const processState = useSelector((state: RootState) => getProcessState(state));
     const customActions = useSelector((state: RootState) => getCustomActions(state));
+    const versionId = useSelector(getProcessVersionId);
+
+    console.log(versionId);
```
philemone, line 38: :slightly_smiling_face:

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/CustomAction.scala
```diff
@@ -32,3 +33,8 @@ case class CustomActionParameter(name: String, editor: ParameterEditor)
 case class CustomActionRequest(name: String, processVersion: ProcessVersion, user: User, params: Map[String, String])
 
 case class CustomActionResult(req: CustomActionRequest, msg: String)
+
+sealed trait CustomActionDisplayPolicy
+
+// TODO: Add more
+case object CurrentlyViewedProcessVersionIsDeployed extends CustomActionDisplayPolicy
```
gskrobisz, line 40: I'm not sure if it is good approach. Those policies seem to be instance-specific, but each NU instance can be configured with different custom actions, different policies. 

What happens when I want to add my own custom policy? E.g I want a button that is enabled when scenario diagram contains specified node. Do I need to hardcode it here and in designer/client/src/helpers/customActionDisplayabilityResolver.ts ?

DeamonDev: Answering to your very last question, the answer is yes. If such hardcoded solution cannot be taken into account, the natural improvement is to create some configuration language (implemented using yaml, json, whatever), send it thru' the backend and teach frontend to read it. And of course we should not then use selaed traits.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -133,6 +133,12 @@ class ProcessesResources(
             processService.getProcessActions(processId.id)
           }
         }
+      } ~ path("processes" / ProcessNameSegment / "lastDeployedVersionId") { processName =>
+        processId(processName) { processId =>
+          complete {
+            processService.getLastDeployedVersionId(processId.id)
```
gskrobisz, line 139: Can we use results of `processService.getLatestProcessWithDetails` or `processService.getProcessWithDetails` instead of adding new endpoint? See `ScenarioWithDetails.lastDeployedAction`, there is some info on current and last deployed version.

PR #5505, opened 2024-02-06
[NU-1408][regression] anonymous user handling in case of Tapir's endpoints

mk-software-pl:
## Describe your changes

Akka-http routes have anonymous user handling when no authorization credentials were passed. The functionality was not ported to Tapir-based endpoints. It's a regression.

This PR adds the functionality to Tapir's endpoints.

## Checklist before merge


ForrestFairy (approved):
LGTM, just a question if we should add a note about this to Changelog?

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
```diff
@@ -1,3 +1,85 @@
...
+      case PassedAuthCredentials(value) => value
+      case AnonymousAccess =>
+        new String(Base64.getEncoder.encode(s"$anonymousUsername:${crypter.encrypt(anonymousPass)}".getBytes))
+    }
+
+  }
+
+}
+
+trait Crypter {
```
mslabek, line 51: maybe should be `sealed`? Or do we want to leave extensibility?

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
```diff
@@ -1,3 +1,85 @@
...
+
+object AuthCredentials {
+
+  val anonymousUsername     = "anonymous"
+  private val anonymousPass = "anonymous"
+
+  final case class PassedAuthCredentials(value: String) extends AuthCredentials
+  case object AnonymousAccess                           extends AuthCredentials
+
+  def fromStringToPassedAuthCredentials(value: String): AuthCredentials = {
```
mslabek, line 18: Shouldn't this and other stuff from here be in `security` module? Do we want to expose all this on api level? My understanding is we are telling users that they can use it and we have to communicate changes of this code then.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
```diff
@@ -46,18 +59,81 @@ trait AuthenticationResources extends Directives with FailFastCirceSupport {
...
+
+  override protected final val authCredentialsCrypter: Crypter = AesCrypter
+
+  override def authenticationMethod(): EndpointInput[AuthCredentials] = {
+    anonymousUserRole match {
+      case Some(_) =>
+        // In WithAnonymousAccessFallback the encrypted password of the anonymous user is added as an Authorization
+        // header credentials. Here, using the same crypter, is decrypted and AuthCredentials object is created.
+        // The crypter is used to prevent the anonymous credentials from being used for explicit authentication
+        // (by passing the credentials with the credentials with the HTTP request).
```
mslabek, line 91: Can't really understand this (but I'm not really familiar with the subject so it may be just me). What would happen if we didn't use the crypter?

designer/server/work/.gitkeep
mslabek, line 1: Do we want to remove these kind of files? I don't have an opinion. We have these files but with `.keep` name for example in dist module.

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
mk-software-pl: yeah, it can be sealed.

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
mk-software-pl: i think, it could be hidden somehow, but at the moment the whole `AnonymousAccess` trait is exposed as part of the API. The `AuthCredentials` ADT is used there so, it'd require a non-trivial refactor.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
mk-software-pl: the previous implementation worked like this:
```bash
curl -v "http://.../api/users/" 
```

returned `anonymous` user data.

But when you did it like this:
```bash
curl -v -u anonymous:anonymous "http://.../api/users/" 
```

you got 401. 

My solution (or hack) works like this:
when in akka http response is info that the request failed because of no credentials (and the anonymous role is configured), add the anonymous user credentials to the request and try again. 

Because of the encryption I can distinguish if our code added the credentials or the credentials come from the external world. The encryption here acts like a signature ;)

designer/server/work/.gitkeep
mk-software-pl: In `.gitignore` we have `designer/server/work` entry. So it means, we should ignore the folder. And `.gitkeep` file is added to a folder, to track the folder in git. So, I don't think we need it (assuming that .gitignore is more important). The `work` folder is created automatically. 

IMO we can remove it.

mslabek (approved):

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
```diff
@@ -1,3 +1,85 @@
...
+        PassedAuthCredentials(value)
+    }
+  }
+
+  implicit class Stringify(authCredentials: AuthCredentials) {
+
+    def stringify(crypter: Crypter): String = authCredentials match {
+      case PassedAuthCredentials(value) => value
+      case AnonymousAccess =>
+        new String(Base64.getEncoder.encode(s"$anonymousUsername:${crypter.encrypt(anonymousPass)}".getBytes))
```
arkadius, line 44: I wonder why this Auth credential need to look like this. Why we encode password and add the user instead of use some secure random and check if this secure random is this one that we generated. Can't it be just as simple? At the end we just want to check if our fake credentials are our own.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
```diff
@@ -46,18 +59,81 @@ trait AuthenticationResources extends Directives with FailFastCirceSupport {
...
+  protected def authenticateReally(): AuthenticationDirective[AuthenticatedUser]
+
+  protected def authenticateReally(credentials: PassedAuthCredentials): Future[Option[AuthenticatedUser]]
+
+  override protected final val authCredentialsCrypter: Crypter = AesCrypter
+
+  override def authenticationMethod(): EndpointInput[AuthCredentials] = {
+    anonymousUserRole match {
+      case Some(_) =>
+        // In WithAnonymousAccessFallback the encrypted password of the anonymous user is added as an Authorization
```
arkadius, line 88: In order in making this more understable let's:
1. Highlight that is the hack
2. Start with explanation of limitation of tapir
3. Use "fake Authorization header" to highlight that it isn't some kind of normal situation :)
4. Mention that we run some part of logic is done twice - the best with notice like `see withAnonymousAccessFallbackOnMissingCredentials` - the best would be to keep this part of code closer

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
```diff
@@ -46,18 +59,81 @@ trait AuthenticationResources extends Directives with FailFastCirceSupport {
...
+      .createAuthenticationResources(config, classLoader)
+  }
+
+}
+
 trait AnonymousAccess extends Directives {
-  val anonymousUserRole: Option[String]
+  this: AuthenticationResources =>
+
+  protected def anonymousUserRole: Option[String]
```
arkadius, line 77: Let's add a TODO, that it shouldn't be in this interface. We should have some external configuration for this instead of adding this field to each security extension's configuration.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
```diff
@@ -5,15 +5,21 @@ import akka.http.scaladsl.server._
...
 trait AuthenticationResources extends Directives with FailFastCirceSupport {
-  val name: String
-  val frontendStrategySettings: FrontendStrategySettings
+  def name: String
+
+  protected def frontendStrategySettings: FrontendStrategySettings
+
+  protected def rawAuthCredentialsMethod: EndpointInput[String]
+
+  protected def authCredentialsCrypter: Crypter = NoOpCrypter
```
arkadius, line 22: Let's remove this default, someone can introduce a dirty bug in their security extension and blame as for that. This class mixing two responsibilities: Handling common logic that is the same for all authentications and this specific for each authentication strategy. I think that we should start to split this. Eventually we should use some composition and keep in the interface only changing parts but we could start with moving this common things to the companion object. WDYT? Or maybe split it now? It was < 100 LOC before this change

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/PatientScalaFutures.scala
```diff
@@ -10,6 +10,6 @@ import org.scalatest.time.{Millis, Seconds, Span}
 trait PatientScalaFutures extends BasePatientScalaFutures {
 
   final override implicit def patienceConfig: PatienceConfig =
-    PatienceConfig(timeout = scaled(Span(5, Seconds)), interval = scaled(Span(50, Millis)))
+    PatienceConfig(timeout = scaled(Span(10, Seconds)), interval = scaled(Span(50, Millis)))
```
arkadius, line 13: We already have a solution for that - use `VeryPatientScalaFutures` :)

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
mk-software-pl: not relevant anymore

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
mk-software-pl: not relevant anymore

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
mk-software-pl: not relevant anymore

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
```diff
@@ -1,3 +1,85 @@
...
 
-final case class AuthCredentials(value: String) extends AnyVal
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.Base64
+import javax.crypto.{Cipher, KeyGenerator, SecretKey}
+import scala.util.{Failure, Success, Try}
+
+sealed trait AuthCredentials
+
+object AuthCredentials {
```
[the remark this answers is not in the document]
mk-software-pl: not relevant anymore

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/NuAkkaHttpServerInterpreterForTapirPurposes.scala
```diff
@@ -40,6 +60,58 @@ class NuAkkaHttpServerInterpreterForTapirPurposes(implicit val executionContext:
...
+        }
+    }
+
+    private def isMissingAuthHeader(response: HttpResponse) = {
+      val wwwAuthHeaderOpt = response.headers.find {
+        case `WWW-Authenticate`(_) => true
+        case _                     => false
+      }
+      wwwAuthHeaderOpt match {
+        case Some(_) if response.status.intValue() == 401 => true
```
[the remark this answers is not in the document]
mk-software-pl: not relevant anymore

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/NuAkkaHttpServerInterpreterForTapirPurposes.scala
```diff
@@ -40,6 +60,58 @@ class NuAkkaHttpServerInterpreterForTapirPurposes(implicit val executionContext:
     }
   }
 
+  private implicit class WithAnonymousAccessFallback(val route: Route) {
+
+    def withAnonymousAccessFallbackOnMissingCredentials(): Route = { context: RequestContext =>
+      route
+        .apply(context)
+        .flatMap {
+          case complete @ RouteResult.Complete(_) if isMissingAuthHeader(complete.response) =>
```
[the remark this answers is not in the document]
mk-software-pl: not relevant anymore

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationResources.scala
mk-software-pl: I've refactored it a little bit

arkadius (approved):
Nice

PR #5508, opened 2024-02-06
[Draft] Added Validators to CustomActionParameter

Diamekod0221:
## Describe your changes
Add parameter validators to CustomAction. Expose an endpoint for CustomAction validation. Check if we need new Validators for CustomActions. Hook up validators to CustomActions as currently needed.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementApiEndpoints.scala
```diff
@@ -0,0 +1,60 @@
...
+import sttp.tapir.json.circe.jsonBody
+import sttp.tapir.derevo.schema
+
+class ManagementApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpointDefinitions {
+
+  private lazy val baseProcessManagementEndpoint = baseNuApiEndpoint.in("processManagement")
+
+  lazy val customActionValidationEndpoint: SecuredEndpoint[
+    (ProcessName, CustomActionRequest),
+    CustomActionValidationError,
```
gskrobisz, line 31: See comments in https://github.com/TouK/nussknacker/commit/7cd849ac0b77e2b0cdae606a84c2757b4fca66e9

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
```diff
@@ -27,6 +27,7 @@ import pl.touk.nussknacker.ui.process.deployment.{DeploymentManagerDispatcher, D
 import pl.touk.nussknacker.ui.process.processingtype.ProcessingTypeDataProvider
 import pl.touk.nussknacker.ui.process.test.{RawScenarioTestData, ResultsWithCounts, ScenarioTestService}
 import pl.touk.nussknacker.ui.security.api.LoggedUser
+import pl.touk.nussknacker.ui.validation.CustomActionValidator
```
gskrobisz, line 30: Seems to be not used.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesApiEndpoints.scala
```diff
@@ -482,6 +482,7 @@ object NodesApiEndpoints {
       final case object NoPermission                                    extends NodesError with CustomAuthorizationError
       final case class MalformedTypingResult(msg: String)               extends NodesError
 
+      // FIXME: what does this message mean??
```
gskrobisz, line 485: Please, remove comment. I guess here we need one-way serialization only to communicate exceptions.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -148,7 +157,7 @@ class DeploymentServiceImpl(
   private def validateDeploymentComment(comment: Option[String]): Future[Option[DeploymentComment]] =
     DeploymentComment.createDeploymentComment(comment, deploymentCommentSettings) match {
       case Valid(deploymentComment) => Future.successful(deploymentComment)
-      case Invalid(exc)             => Future.failed(ValidationError(exc.getMessage))
+      case Invalid(exc)             => Future.failed(CustomActionValidationError(exc.message))
```
gskrobisz, line 160: Use CustomActionValidator.validate here to validate deployment comment.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -593,13 +603,25 @@ class DeploymentServiceImpl(
           params
         )
         customActionOpt = manager.customActionsDefinitions.find(_.name == actionName)
-        _ <- existsOrFail(customActionOpt, CustomActionNonExisting(actionName))
+        _ <- existsOrFail(customActionOpt, CustomActionNonExistingError(s"${actionName.value} doesn't exist"))
+        validator = new CustomActionValidator(manager.customActionsDefinitions)
+        // TODO: remove this validation prosthesis
```
gskrobisz, line 608: Please remove this validation prosthesis and use action parameters validation.

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -391,11 +399,12 @@ class AkkaHttpBasedRouteProvider(
...
           appApiHttpService,
           componentsApiHttpService,
-          userApiHttpService,
+          managementApiHttpService,
+          nodesApiHttpService,
           notificationApiHttpService,
           scenarioActivityApiHttpService,
           scenarioParametersHttpService,
-          nodesApiHttpService
+          userApiHttpService
```
gskrobisz, line 407: Why the order is changed?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+import pl.touk.nussknacker.engine.deployment.{
+  CustomActionDefinition,
+  CustomActionParameter,
+  CustomActionValidationResult
+}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.restmodel.CustomActionRequest
+import pl.touk.nussknacker.ui.BadRequestError
+
+class CustomActionValidator(val allowedActions: List[CustomActionDefinition]) {
```
gskrobisz, line 18: Lets rename to ScenarioActionParameterValidator, we are aiming here to validation at "action parameter" level. And because of that I recommend to pass single action properties like this:
```
class ScenarioActionParameterValidator(actionParameters: Map[String, CustomActionParameter])
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+      checkedParams
+    )
+  }
+
+}
+
+object CustomActionValidationError {
+  def apply(message: String): CustomActionValidationError = new CustomActionValidationError(message)
+
+  implicit val encodeCustomActionValidationError: Encoder[CustomActionValidationError] =
```
gskrobisz, line 144: This should be in ManagementApiEndpoint

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.restmodel.CustomActionRequest
+import pl.touk.nussknacker.ui.BadRequestError
+
+class CustomActionValidator(val allowedActions: List[CustomActionDefinition]) {
+
+  def validateCustomActionParams(
+      request: CustomActionRequest
+  ): Either[CustomActionValidationError, CustomActionValidationResult] = {
```
gskrobisz, line 22: Too much happens here. It seems that with `Either[CustomActionValidationError, CustomActionValidationResult]` you have 2 levels of error handling:
- Left(CustomActionValidationError) is returned when validation errors are detected
- Right(CustomActionValidationResult) contains validation errors.

Here we should return something simple like:
```
CustomActionValidationResult(parameterErrors: List[NodeValidationError])
```
In the code below we should handle validation error as Validated/ValidatedNel, instead of exception (see CustomActionValidationError and MismatchedParamsError). Also, keep in mind that:
- no validation error = ok, goes straight to 200 OK
- some validation errors = still ok, we should return list of invalid fields with reason of invalidity
- technical exception = not ok, goes straight to 500 server error (?)

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+                .collect { case Invalid(i) => i }
+            case None =>
+              throw new IllegalStateException
+          }
+        )
+      }
+    }
+  }
+
+  private def checkForMissingKeys(
```
gskrobisz, line 92: See similar validation in ScenarioPropertiesValidator

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+        val customActionParams      = foundAction.parameters
+        val requestParamsMap        = getRequestParamsMap(request, customActionParams)
+
+        val validated = validateParams(requestParamsMap, customActionParams)
+        getValidationResult(validated)
+      }
+    }
+  }
+
+  private def getRequestParamsMap(request: CustomActionRequest, customActionParams: List[CustomActionParameter]) = {
```
gskrobisz, line 43: See similar validation in `ScenarioPropertiesValidator`, it seems that here you want do `getMissingRequiredPropertyValidationResults` and `getUnknownPropertyValidationResults`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,160 @@
...
+
+    hasErrors match {
+      case Right(true) =>
+        Right(CustomActionValidationResult.Invalid(validatedParams.getOrElse(throw new IllegalStateException)))
+      case Right(false) => Right(CustomActionValidationResult.Valid)
+      case Left(va)     => Left(va)
+    }
+  }
+
+  def validateCustomActionParams(
```
gskrobisz, line 119: `validateCustomActionParams` is not in the responsibliity scope of validator, should be moved to DeploymentServiceImpl.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ManagementResourcesSpec.scala
```diff
@@ -421,8 +421,8 @@ class ManagementResourcesSpec
       ProcessTestData.sampleProcessName,
       CustomActionRequest(ScenarioActionName("non-existing"))
     ) ~> check {
-      status shouldBe StatusCodes.NotFound
-      responseAs[String] shouldBe "non-existing is not existing"
+      status shouldBe StatusCodes.BadRequest
```
gskrobisz, line 424: Non existing action should be "not found".

extensions-api/src/main/scala/pl/touk/nussknacker/engine/deployment/CustomActionDefinition.scala
```diff
@@ -26,7 +28,13 @@ case class CustomActionDefinition(
...
 
-//TODO: validators, defaultValue, hint, labelOpt?
-case class CustomActionParameter(name: String, editor: ParameterEditor)
+case class CustomActionParameter(name: String, editor: ParameterEditor, validators: Option[List[ParameterValidator]])
 
 case class CustomActionResult(msg: String)
+
+sealed trait CustomActionValidationResult
+
+object CustomActionValidationResult {
```
gskrobisz, line 37: You should use `Validated` and `ValidatedNel` where Valid and Invalid are already implemented.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementResources.scala
Diamekod0221: imports updated

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
Diamekod0221: Used Validator

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
Diamekod0221: I have alligned components alphabetically to easily check order.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: Changed to injecting single aciton into the constructor. I think the naming is ok for now, we can clean up when the change to action parameter comes

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: Extracted error logic to ManagementApiHttpService, CustomActionValidator now returns ValidatedNel[SubGraphCompilationError, Unit]

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: removed

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: added a checkForMissingParameters

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: moved

gskrobisz: I still see here `validateCustomActionParams`and `fromCommand`. Lets keep that validator responsible only for validation on action level, and move that translation from command outside, to `DeploymentServiceImpl`.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ManagementResourcesSpec.scala
```diff
@@ -422,7 +422,9 @@ class ManagementResourcesSpec
       CustomActionRequest(ScenarioActionName("non-existing"))
     ) ~> check {
       status shouldBe StatusCodes.NotFound
-      responseAs[String] shouldBe "non-existing is not existing"
+      responseAs[
+        String
+      ] shouldBe s"Couldn't find definition of action non-existing for scenario ${ProcessTestData.sampleProcessName} when trying to validate"
```
gskrobisz, line 427: Maybe "Couldn't find definition of action non-existing" is sufficient?

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/CustomActionValidatorTest.scala
```diff
@@ -0,0 +1,198 @@
...
+      "testStatus1" :: "testStatus2" :: Nil,
+      testCustomActionParams
+    )
+
+  private val noParamsCustomAction =
+    CustomActionDefinition(ScenarioActionName("noparams"), "testStatus" :: Nil, Nil)
+
+  private val validator = new CustomActionValidator(testCustomAction)
+
+  type TestParams = Map[String, String]
```
gskrobisz, line 48: It it used only in one place. Seems to be obsolete.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,130 @@
...
+import pl.touk.nussknacker.engine.api.NodeId
+import pl.touk.nussknacker.engine.api.context.PartSubGraphCompilationError
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.MismatchParameter
+import pl.touk.nussknacker.engine.api.definition.MandatoryParameterValidator
+import pl.touk.nussknacker.engine.api.deployment.CustomActionCommand
+import pl.touk.nussknacker.engine.api.parameter.ParameterName
+import pl.touk.nussknacker.engine.deployment.{CustomActionDefinition, CustomActionParameter}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.restmodel.CustomActionRequest
+import pl.touk.nussknacker.ui.{BadRequestError, NotFoundError}
```
gskrobisz, line 14: Unused import {BadRequestError, NotFoundError}

designer/client/src/components/modals/CustomActionDialog.tsx
```diff
@@ -38,6 +40,28 @@ function CustomActionForm(props: CustomActionFormProps): JSX.Element {
...
+    const validationRequest: CustomActionValidationRequest = {
+        actionName: action.name,
+        params: state,
+    };
+    const debouncedValidateCustomAction = debounce(async () => {
+        const data = await HttpService.validateCustomAction(processName, validationRequest);
+        setErrors(data.validationErrors);
+    }, 300);
+
+    useEffect(() => {
```
Dzuming, line 55: You can merge, these useEffect, and update dependencies

```
 useEffect(() => {
        debouncedValidateCustomAction();

      return () => {
        debouncedValidateCustomAction.cancel();
      };
    }, [debouncedValidateCustomAction]);
```

designer/client/src/components/modals/CustomActionDialog.tsx
```diff
@@ -38,6 +40,28 @@ function CustomActionForm(props: CustomActionFormProps): JSX.Element {
...
     useEffect(() => onChange(state), [onChange, state]);
 
+    const [errors, setErrors] = useState<NodeValidationError[]>([]);
+
+    const processName = useSelector(getProcessName);
+    const validationRequest: CustomActionValidationRequest = {
+        actionName: action.name,
+        params: state,
+    };
+    const debouncedValidateCustomAction = debounce(async () => {
```
Dzuming, line 50: You can memoize this action

```
const debouncedValidateCustomAction = useMemo(
        () =>
            debounce(async () => {
                const data = await HttpService.validateCustomAction(processName, validationRequest);
                setErrors(data.validationErrors);
            }, 300),
        [processName, validationRequest],
    );
```

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/CustomActionValidatorTest.scala
```diff
@@ -0,0 +1,198 @@
...
+    CustomActionDefinition(
+      testCustomActionName,
+      "testStatus1" :: "testStatus2" :: Nil,
+      testCustomActionParams
+    )
+
+  private val noParamsCustomAction =
+    CustomActionDefinition(ScenarioActionName("noparams"), "testStatus" :: Nil, Nil)
+
+  private val validator = new CustomActionValidator(testCustomAction)
```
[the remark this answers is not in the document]
Diamekod0221: The ones that create their own validator use some custom validator state. I could maybe rename to customValidator for clarity and define the whole setup block outside of test scope for consistency?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ManagementResourcesSpec.scala
Diamekod0221: I think it is good to write the scenario name here to make sure that we are actually processing what we want. More information shouldn't hurt :)

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/CustomActionValidatorTest.scala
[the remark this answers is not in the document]
Diamekod0221: I have renamed to `defaultValidator` and `customValidator` for clarity. There's only one test that uses a custom validator so i think it's okay to use the current structure.

designer/server/src/test/scala/pl/touk/nussknacker/ui/validation/CustomActionValidatorTest.scala
Diamekod0221: removed

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ManagementResourcesSpec.scala
gskrobisz: Ok, lets keep the scenario name. The part "when trying to validate" is not true. With current implementation it's more like "when trying to execute an action" and is obsolete.

designer/client/src/components/modals/CustomActionDialog.tsx
Dzuming: We discussed it internally, and setTimeout will fit our needs in this case

Dzuming (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ManagementApiHttpService.scala
```diff
@@ -0,0 +1,82 @@
...
+            scenarioIdWithName = ProcessIdWithName(scenarioId, processName)
+            actionDefinition <- getActionDefinition(scenarioIdWithName, req.actionName)
+            validator = new CustomActionValidator(actionDefinition)
+            resultsDto <- getValidationResultsDto(validator, req)
+          } yield resultsDto
+        }
+      }
+  }
+
+  private def getScenarioIdByName(scenarioName: ProcessName): EitherT[Future, NoScenario, ProcessId] = {
```
arkadius, line 48: I extracted a helper trait for that: `ScenarioHttpServiceExtensions` to avoid code duplication. Can you reuse it?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+import pl.touk.nussknacker.engine.api.context.PartSubGraphCompilationError
+import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.MismatchParameter
+import pl.touk.nussknacker.engine.api.definition.MandatoryParameterValidator
+import pl.touk.nussknacker.engine.api.deployment.CustomActionCommand
+import pl.touk.nussknacker.engine.api.parameter.ParameterName
+import pl.touk.nussknacker.engine.deployment.{CustomActionDefinition, CustomActionParameter}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.restmodel.CustomActionRequest
+
+class CustomActionValidator(allowedAction: CustomActionDefinition) {
```
arkadius, line 15: Do you see an option to unify this with `ScenarioPropertiesValidator`? It is a huge duplication of quite easy to broken business logic. Also users may be disgusted if they find out that this logic behave different for some corner cases.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+            s"Couldn't find a matching parameter in action definition for this param: $name",
+            "",
+            ParameterName(name),
+            nodeId.id
+          ).invalidNel[Unit]
+      }
+    }.sequence_
+  }
+
+  private def toValidatedNel(param: CustomActionParameter, expressionValue: String, parameterName: String)(
```
arkadius, line 78: This name is confusing. It looks like you do some dumb conversion here but in fact, here is done the core of the logic - validations

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      customActionParams: List[CustomActionParameter]
+  )(implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, Unit] = {
+
+    customActionParams
+      .collect { case param if param.validators.nonEmpty => (param.name, param.validators) }
+      .flatMap { case (name, validators) =>
+        validators.collect { case validator => validator }.map {
+          _.isValid(
+            paramName = ParameterName(name),
+            expression = Expression.spel("None"),
```
arkadius, line 104: What the heck? Why do you put some `None` into spel expression?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  private def validateParams(
+      requestParamsMap: Map[String, String],
+      customActionParams: List[CustomActionParameter]
+  )(implicit nodeId: NodeId): ValidatedNel[PartSubGraphCompilationError, Unit] = {
+
+    requestParamsMap match {
+      case paramsMap if paramsMap.nonEmpty =>
+        handleNonEmptyParamsRequest(paramsMap, customActionParams)
+      case emptyMap if emptyMap.isEmpty =>
```
arkadius, line 54: Why the empty map is some special case? Eventually you traverse all params so what's the difference?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -646,17 +649,33 @@ class DeploymentServiceImpl(
...
         )
         customActionOpt = manager.customActionsDefinitions.find(_.actionName == actionName)
-        _ <- existsOrFail(customActionOpt, CustomActionNonExisting(actionName))
+        customAction <- existsOrFail(
+          customActionOpt,
+          CustomActionNonExistingError(
+            s"Couldn't find definition of action ${actionName.value} for scenario ${processIdWithName.name} when trying to validate"
+          )
+        )
+        _ <- validateActionCommand(actionCommand, customAction)
```
arkadius, line 658: I can't see any test covering this important case

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -646,17 +649,33 @@ class DeploymentServiceImpl(
...
         _ = checkIfCanPerformActionInState(actionName, processDetails, processState)
         invokeActionResult <- DBIOAction.from(manager.processCommand(actionCommand))
       } yield invokeActionResult
     )
   }
 
+  private def validateActionCommand(actionCommand: CustomActionCommand, customAction: CustomActionDefinition) = {
+    val validator        = new CustomActionValidator(customAction)
+    val validationResult = validator.validateCustomActionParams(actionCommand)
+    val validationFlag   = validationResult
```
arkadius, line 668: Flag? Flag is a boolean parameter. It:
- isn't boolean
- isn't parameter - it's the result

Also, this local variable doesn't add any value, you already have `validationResult`

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: I think in the current state we could possibly merge these two in a separate refactor. I don't think that's a good idea, because ScenarioProperties is a different object from CustomAction. I think merging this makes the code unnecesarily complex, especially if it turns out that the logic isn't exactly the same or we have to change it in the future. It's just 90 extra lines. 
Your user argument is convincing, but we should just make sure the external results are the same.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: In case that the map is empty I'm checking if there are validators that will pop off for an empty request (like MandatoryParameter). Because of this i have divided this check into two cases.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
Diamekod0221: Please specify which line do you mean. This: 
` customAction <- existsOrFail(
          customActionOpt,
          CustomActionNonExistingError(
            s"Couldn't find definition of action ${actionName.value} for scenario ${processIdWithName.name} when trying to validate"
          )
        )
        ` ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
Diamekod0221: Tru, leftover code. Fixed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: It's an attempt to validate an empty expression when the user doesn't input anything. Is there some more subtle way?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: Changed the name.

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/ManagementApiHttpService.scala
```diff
@@ -0,0 +1,83 @@
...
+            scenarioIdWithName = ProcessIdWithName(scenarioId, processName)
+            actionDefinition <- getActionDefinition(scenarioIdWithName, req.actionName)
+            validator = new CustomActionValidator(actionDefinition)
+            resultsDto <- getValidationResultsDto(validator, req)
+          } yield resultsDto
+        }
+      }
+  }
+
+  private def getScenarioIdByName(
```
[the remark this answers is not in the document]
Diamekod0221: I don't want to use it because it would effectively replace 3 lines of code with 7 or more and add complexity.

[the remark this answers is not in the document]
arkadius: What plan do you recommend to deal with this code duplication (getting scenario details, checking if scenario exists)?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
arkadius: I don't understand your answer. Why do you validate an empty expression when the user doesn't input anything?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
arkadius: And for one element map with missing some parameter value that is mandatory you don't need to do the same?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
arkadius: No, this:
```
DBIOAction.failed(new IllegalStateException(s"Validation failed for: ${actionCommand.actionName}"))
    }
```

arkadius: Also IMO it shouldn't be 500, it is bad request - not an illegal state of the application

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/ManagementApiHttpService.scala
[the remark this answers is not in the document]
Diamekod0221: I think your trait is a good idea, but in this case it's a one liner 
`
            scenarioId <- EitherT.fromOptionF(processService.getProcessId(processName), NoScenario(processName))`
            that doesn't make use of all the functionality the trait provides and is basically a wrapped function call. Adding some extra code and complexity on top of it doesn't make that much sense to me

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/CustomActionValidator.scala
Diamekod0221: you are right, I forgot to clean this up when i extracted 
`private def checkForMissingParams(
      requestParamsMap: Map[String, String],
      customActionParams: List[CustomActionParameter]`
      This emptyParamsMapRequest can probably go along with the None expression check. I will update it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
Diamekod0221: I have changed the error type and added a test case.

arkadius (approved):
one comment left. fix it and you can merge

PR #5521, opened 2024-02-09
[NU-1068] DeploymentManagerProvider: ability to return validation errors + DeploymentManagerDependencies

arkadius:
…easier passing through methods

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
@@ -101,12 +101,11 @@ class AkkaHttpBasedRouteProvider(
       additionalUIConfigProvider = createAdditionalUIConfigProvider(resolvedConfig, sttpBackend)
       typeToConfig <- prepareProcessingTypeData(
         config,
-        deploymentServiceSupplier,
-        AllDeployedScenarioService(dbRef, _),
         processingTypeDataStateFactory,
         additionalUIConfigProvider,
+        deploymentServiceSupplier,
+        AllDeployedScenarioService(dbRef, _),
         sttpBackend,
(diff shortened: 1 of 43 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -500,21 +501,32 @@ class AkkaHttpBasedRouteProvider(
...
-  )(implicit executionContext: ExecutionContext): Resource[IO, ProcessingTypeDataReload] = {
-    implicit val sttpBackendImplicit: SttpBackend[Future, Any] = sttpBackend
+      executionContext: ExecutionContext,
+  ): Resource[IO, ProcessingTypeDataReload] = {
     Resource
       .make(
         acquire = IO(
-          new ProcessingTypeDataReload(() =>
+          new ProcessingTypeDataReload({ () =>
+            def getDeploymentManagerDependencies(processingType: ProcessingType) = {
```
lukasz-bigorajski, line 512: Maybe it's good idea to extract/move this function outside: `prepareProcessingTypeData`?

lukasz-bigorajski (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
arkadius: This function uses 3 of the method arguments.
Maybe I'm missing something but I can't see an option to extract it without some strange things like:
a) Multiple parameters lists - one with these 3 parameters and the second one with the `processingType`
b) Extracting a new internal class which takes these 3 parameters in the constructor and the `processingType` as a parameter of the single method

Do you see some better approach?

PR #5525, opened 2024-02-12
Exclude Finished batches from isDeployed filter, rename filter to isActive

piotrp:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessActionEntityFactory.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessActionEntityFactory.scala
@@ -1,7 +1,7 @@
 package pl.touk.nussknacker.ui.db.entity
 
 import pl.touk.nussknacker.engine.api.deployment.ProcessActionState.ProcessActionState
-import pl.touk.nussknacker.engine.api.deployment.{ProcessActionId, ProcessActionType}
+import pl.touk.nussknacker.engine.api.deployment.{ProcessActionId, ProcessActionState, ProcessActionType}
 import pl.touk.nussknacker.engine.api.deployment.ProcessActionType.ProcessActionType
 import pl.touk.nussknacker.engine.api.process.{ProcessId, VersionId}
 import slick.lifted.{ForeignKeyQuery, ProvenShape, TableQuery => LTableQuery}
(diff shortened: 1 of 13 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessActionEntityFactory.scala
```diff
@@ -92,5 +92,6 @@ final case class ProcessActionEntityData(
 
   lazy val createdAtTime: Instant           = createdAt.toInstant
   lazy val performedAtTime: Option[Instant] = performedAt.map(_.toInstant)
-  lazy val isDeployed: Boolean              = actionType.equals(ProcessActionType.Deploy)
+  lazy val isDeployed: Boolean =
+    actionType.equals(ProcessActionType.Deploy) && !state.equals(ProcessActionState.ExecutionFinished)
```
wrzontek, line 96: perhaps this change should be reflected in some UT (i think the only changes in tests in this PR are renames)

arkadius: let's use `==` instead of `equals`, we are in scala, not some prehistoric java! ;-)   
W8 I can't see usage of that. It looks like a death code. Isn't it?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -53,7 +53,7 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](
         { process =>
           expr.flatten.foldLeft(true: Rep[Boolean])((x, y) => x && y(process))
         },
-        query.isDeployed
```
arkadius, line 56: What about deeper usages of this? e.g. `fetchLatestProcessesQuery`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ProcessActionEntityFactory.scala
arkadius: Yes yes yes! All undocumented as externally used death code should be always removed. I even looked for some tool for detecting that but didn't find any reliable one :(

PR #5534, opened 2024-02-13
Add fixed values parameter editor for timestampField in kafka source

vder:
## Checklist before merge

### DIFF
--- a/utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
+++ b/utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
@@ -20,28 +20,28 @@ object DelayedKafkaSourceFactory {
   final val TimestampFieldParamName = "timestampField"
 
   // TODO: consider changing to lazy parameter and add the same parameter also in "not delayed" kafka sources
-  final val FallBackTimestampFieldParameter = Parameter
+  final val fallbackTimestampFieldParameter = Parameter
     .optional(TimestampFieldParamName, Typed[String])
     .copy(
       editor = Some(DualParameterEditor(simpleEditor = StringParameterEditor, defaultMode = DualEditorMode.RAW))
     )
 
   def timestampFieldParameter(typingResult: Option[TypingResult]): Parameter = {
-
-    val editor = typingResult
+    val editorOpt = typingResult
       .collect { case TypedObjectTypingResult(fields, _, _) => fields.toList }
       .map(_.collect {
         case (paramName, typing) if TimestampUtils.supportedTimestampTypes.contains(typing) =>
           FixedExpressionValue(s"'${paramName}'", paramName)
       })
       .filter(_.nonEmpty)
+      .map(_.sortBy(_.label))
       .map(FixedValuesParameterEditor(_))
       .map(DualParameterEditor(_, DualEditorMode.SIMPLE))
       .orElse(Some(DualParameterEditor(simpleEditor = StringParameterEditor, defaultMode = DualEditorMode.RAW)))
 
     Parameter
       .optional(TimestampFieldParamName, Typed[String])
-      .copy(editor = editor)
+      .copy(editor = editorOpt)
   }
 
   def extractTimestampField(params: Params): String = params.extract[String](TimestampFieldParamName).getOrElse("")
--- a/utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
+++ b/utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
@@ -35,6 +35,7 @@ object DelayedKafkaSourceFactory {
       })
       .filter(_.nonEmpty)
       .map(_.sortBy(_.label))
+      .map(FixedExpressionValue("", "") :: _)
       .map(FixedValuesParameterEditor(_))
       .map(DualParameterEditor(_, DualEditorMode.SIMPLE))
       .orElse(Some(DualParameterEditor(simpleEditor = StringParameterEditor, defaultMode = DualEditorMode.RAW)))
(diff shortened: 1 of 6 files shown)


utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/delayed/DelayedUniversalKafkaSourceFactory.scala
```diff
@@ -73,4 +80,22 @@ class DelayedUniversalKafkaSourceFactory(
...
 
+  protected def timestampFieldParamStep(implicit nodeId: NodeId): NodeTransformationDefinition = {
+    case step @ TransformationStep(
+          (`topicParamName`, DefinedEagerParameter(topic: String, _)) ::
+          (SchemaVersionParamName, DefinedEagerParameter(version: String, _)) :: Nil,
+          _
+        ) =>
+      val preparedTopic = prepareTopic(topic)
+      val versionOption = parseVersionOption(version)
+      val typingResultValidationResult = determineSchemaAndType(
```
arkadius, line 91: We have such a concept as a `TransformationStep.state`. Let's put this information into it during and pass it between steps. Thanks to that we avoid unnecessary requests to the schema registry.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
```diff
@@ -20,12 +20,30 @@ object DelayedKafkaSourceFactory {
...
 
   // TODO: consider changing to lazy parameter and add the same parameter also in "not delayed" kafka sources
-  final val TimestampFieldParameter = Parameter
+  final val fallbackTimestampFieldParameter = Parameter
     .optional(TimestampFieldParamName, Typed[String])
     .copy(
       editor = Some(DualParameterEditor(simpleEditor = StringParameterEditor, defaultMode = DualEditorMode.RAW))
     )
 
+  def timestampFieldParameter(typingResult: Option[TypingResult]): Parameter = {
```
arkadius, line 29: 1. Let's rename `typingResult` to sth like `kafkaRecordValueType`
2. Let's leave a TODO that it is not a final solution and in case of further development (e.g. nested fields adding etc.) we should rather focus on the final solution allowing to define spel expression.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
```diff
@@ -20,12 +20,30 @@ object DelayedKafkaSourceFactory {
...
+    val editorOpt = typingResult
+      .collect { case TypedObjectTypingResult(fields, _, _) => fields.toList }
+      .map(_.collect {
+        case (paramName, typing) if TimestampUtils.supportedTimestampTypes.contains(typing) =>
```
arkadius, line 33: Would be nice to see some test showing which fields are available in editor. If you think that it is too much and it is better to focus on the final solution it's fine to me to mention in the proposed TODO that we picked this compromise instead.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
```diff
@@ -20,12 +20,30 @@ object DelayedKafkaSourceFactory {
...
+          FixedExpressionValue(s"'${paramName}'", paramName)
+      })
+      .filter(_.nonEmpty)
+      .map(_.sortBy(_.label))
+      .map(FixedValuesParameterEditor(_))
+      .map(DualParameterEditor(_, DualEditorMode.SIMPLE))
+      .orElse(Some(DualParameterEditor(simpleEditor = StringParameterEditor, defaultMode = DualEditorMode.RAW)))
```
arkadius, line 40: Would be nice to see test showing that even if we give some fields to the user or no field has matching type, user can provide his/her own field. If you think that it is too much and it is better to focus on the final solution it's fine to me to mention in the proposed TODO that we picked this compromise instead.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
vder: done.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
vder: ok I've made some digging and managed to test parameterEditors without too much code.

utils/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/delayed/DelayedKafkaSourceFactory.scala
vder: same here

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/delayed/DelayedUniversalKafkaSourceFactory.scala
vder: wasnt sure how to properly do validations before last step so I put in state whole valueSchemaValidationResult

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
```diff
@@ -229,10 +229,12 @@ class UniversalKafkaSourceFactory(
 
 object UniversalKafkaSourceFactory {
 
-  case class UniversalKafkaSourceFactoryState(
+  trait UniversalKafkaSourceFactoryState
```
arkadius, line 232: Why this common trait for both cases? And what does mean this Final prefix? Which one state is non final?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/delayed/DelayedUniversalKafkaSourceFactory.scala
arkadius: Does this state-related approach reduce unnecessary requests to the schema registry? :-)

arkadius (approved):

PR #5535, opened 2024-02-13
[NU-1068] ModelClassLoader - ability to specify working directory for the tests purpose

arkadius:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/LocalProcessingTypeDataStateFactory.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/LocalProcessingTypeDataStateFactory.scala
@@ -13,6 +13,8 @@ import pl.touk.nussknacker.ui.process.processingtype.{
 }
 import pl.touk.nussknacker.ui.util.LocalNussknackerWithSingleModel.{category, typeName}
 
+import java.nio.file.Path
+
 class LocalProcessingTypeDataStateFactory(
     modelData: ModelData,
     deploymentManagerProvider: DeploymentManagerProvider,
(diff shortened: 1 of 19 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/LocalProcessingTypeDataStateFactory.scala
```diff
@@ -22,7 +24,8 @@ class LocalProcessingTypeDataStateFactory(
   override def create(
       designerConfig: ConfigWithUnresolvedVersion,
       getDeploymentManagerDependencies: ProcessingType => DeploymentManagerDependencies,
-      additionalUIConfigProvider: AdditionalUIConfigProvider
+      additionalUIConfigProvider: AdditionalUIConfigProvider,
+      workingDirectoryOpt: Option[Path]
```
lciolecki, line 28: Why Opt? I haven't seen much use for this kind of standard.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/ProcessingTypeConfig.scala
```diff
@@ -1,14 +1,10 @@
...
 import com.typesafe.config.Config
-import net.ceedubs.ficus.readers.ValueReader
-
-import java.io.File
-import java.net.URL
 
 case class ProcessingTypeConfig(
     deploymentManagerType: String,
-    classPath: List[URL],
+    classPath: List[String],
```
lciolecki, line 7: Isn't it an API's class? Maybe we should put down an entry in MigrationGuide? Why we change URL to String? At the end you convert these strings to URI..

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/LocalProcessingTypeDataStateFactory.scala
arkadius: `find -name *.scala -exec grep 'Opt: ' {} \; | wc -l` returned 127 entries - it is quite common. I've seen three approaches in scala:
1. using `Opt` suffix
2. using `maybe` prefix
3. skipping Opt/maybe (I assume that this one you propose)
 
1. and 2. approach has the benefit that whey you do map on this field you don't have to think how to call the mapped value (`workingDirectoryOpt.map { workingDirectory=>`). How do you name the mapped values in the proposed approach?

lciolecki: So we can make agreement that in the future Option value we will named with postfix Opt? It's fine by me.

lciolecki (approved):

lukasz-bigorajski (approved):

PR #5540, opened 2024-02-14
Backport 1.13-esp: better logging for scenario state fetching  

lciolecki:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
@@ -493,8 +493,11 @@ class DeploymentServiceImpl(
           case CompletedNormally(value) =>
             value
           case CompletedByTimeout(value) =>
+            logger.trace(s"Checking fetching scenario state ${processIdWithName.name} with policy $freshnessPolicy")
             logger
-              .warn(s"Timeout: $timeout occurred during waiting for response from engine for ${processIdWithName.name}")
+              .warn(
+                s"Timeout: $timeout occurred during waiting for response from engine for ${processIdWithName.name} with policy: $freshnessPolicy"
+              )
             value
         }
       }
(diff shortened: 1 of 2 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -493,8 +493,11 @@ class DeploymentServiceImpl(
           case CompletedNormally(value) =>
             value
           case CompletedByTimeout(value) =>
+            logger.trace(s"Checking fetching scenario state ${processIdWithName.name} with policy $freshnessPolicy")
```
raphaelsolarski, line 496: Do we need the trace log if we already have logger.warn with same data in the next line?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/rest/HttpFlinkClient.scala
```diff
@@ -93,7 +92,10 @@ class HttpFlinkClient(config: FlinkConfig)(implicit backend: SttpBackend[Future,
           .reverse
       }
       .map(WithDataFreshnessStatus.fresh)
-      .recoverWith(recoverWithMessage("retrieve Flink jobs"))
+      .recoverWith {
+        logger.trace(s"Checking fetching scenario state $jobName with policy $freshnessPolicy")
```
raphaelsolarski, line 96: Why you moved `"Checking fetching scenario state $jobName with policy $freshnessPolicy"` log from metod start to failure handling?

lciolecki: I want to trace from the begging how was looked call this method in case of error.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
lciolecki: In fact tracing in HttpFlinkClient is enough.

raphaelsolarski (approved):

PR #5575, opened 2024-02-19
Fixup: fetching/parsing batch periodic json only when needed

raphaelsolarski:


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesTable.scala
```diff
@@ -104,13 +108,62 @@ trait PeriodicProcessesTableFactory {
 
   }
 
-  object PeriodicProcesses extends TableQuery(new PeriodicProcessesTable(_))
+  class PeriodicProcessesWithUnitTable(tag: Tag) extends PeriodicProcessesTable(tag) {
```
arkadius, line 111: Let's call it `PeriodicProcessesWithoutJson` or `PeriodicProcessesMetadataOnly`

arkadius (approved):
LGTM

PR #5591, opened 2024-02-21
Fixup: fetching/parsing batch periodic json only when needed for 1.13 esp

raphaelsolarski:


raphaelsolarski:
closed due to https://github.com/TouK/nussknacker/pull/5621

PR #5611, opened 2024-02-23
[NU-1296] Ad-hoc tests with ververica

ForrestFairy:
With a new default parameter "isTest" in a couple of places ad-hoc te…st works without adding "flink-dropwizard-metrics-deps" to classpath

## Checklist before merge

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkEngineRuntimeContextImpl.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkEngineRuntimeContextImpl.scala
@@ -4,7 +4,7 @@ import org.apache.flink.api.common.functions.RuntimeContext
 import pl.touk.nussknacker.engine.api.JobData
 import pl.touk.nussknacker.engine.api.runtimecontext.{ContextIdGenerator, IncContextIdGenerator}
 import pl.touk.nussknacker.engine.flink.api.FlinkEngineRuntimeContext
-import pl.touk.nussknacker.engine.util.metrics.MetricsProviderForScenario
+import pl.touk.nussknacker.engine.util.metrics.{MetricsProviderForScenario, NoOpMetricsProviderForScenario}
 
 case class FlinkEngineRuntimeContextImpl(jobData: JobData, runtimeContext: RuntimeContext)
     extends FlinkEngineRuntimeContext {
@@ -12,5 +12,14 @@ case class FlinkEngineRuntimeContextImpl(jobData: JobData, runtimeContext: Runti
 
   override def contextIdGenerator(nodeId: String): ContextIdGenerator =
     new IncContextIdGenerator(jobData.metaData.id + "-" + nodeId + "-" + runtimeContext.getIndexOfThisSubtask)
+}
+
+case class FlinkTestEngineRuntimeContextImpl(jobData: JobData, runtimeContext: RuntimeContext)
+    extends FlinkEngineRuntimeContext {
+  override val metricsProvider: MetricsProviderForScenario = NoOpMetricsProviderForScenario
+
+  override def contextIdGenerator(nodeId: String): ContextIdGenerator = {
+    new IncContextIdGenerator(jobData.metaData.id + "-" + nodeId + "-" + runtimeContext.getIndexOfThisSubtask)
+  }
 
 }
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerData.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerData.scala
@@ -36,11 +36,18 @@ class FlinkProcessCompilerData(
     val componentUseCase: ComponentUseCase
 ) {
 
-  def open(runtimeContext: RuntimeContext, nodesToUse: List[_ <: NodeData]): Unit = {
+  def open(runtimeContext: RuntimeContext, nodesToUse: List[_ <: NodeData], isTest: Boolean = false): Unit = {
     val lifecycle = compilerData.lifecycle(nodesToUse)
-    lifecycle.foreach {
-      _.open(FlinkEngineRuntimeContextImpl(jobData, runtimeContext))
+    if (!isTest) {
+      lifecycle.foreach {
+        _.open(FlinkEngineRuntimeContextImpl(jobData, runtimeContext))
+      }
+    } else {
+      lifecycle.foreach {
+        _.open(FlinkTestEngineRuntimeContextImpl(jobData, runtimeContext))
+      }
     }
+
   }
 
   def close(nodesToUse: List[_ <: NodeData]): Unit = {
@@ -70,7 +77,7 @@ class FlinkProcessCompilerData(
   def restartStrategy: RestartStrategies.RestartStrategyConfiguration = exceptionHandler.restartStrategy
 
   def prepareExceptionHandler(runtimeContext: RuntimeContext): FlinkExceptionHandler = {
-    exceptionHandler.open(FlinkEngineRuntimeContextImpl(jobData, runtimeContext))
+    exceptionHandler.open(FlinkTestEngineRuntimeContextImpl(jobData, runtimeContext))
     exceptionHandler
   }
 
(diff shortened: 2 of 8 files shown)


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -72,7 +74,8 @@ class FlinkProcessRegistrar(
       process: CanonicalProcess,
       processVersion: ProcessVersion,
       deploymentData: DeploymentData,
-      resultCollector: ResultCollector
+      resultCollector: ResultCollector,
+      isTest: Boolean = false
```
arkadius, line 78: We already have a `ComponentUseCase`. I wonder if we can reuse this concept.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -132,15 +146,22 @@ class FlinkProcessRegistrar(
       val exceptionHandlerPreparer = (runtimeContext: RuntimeContext) =>
         compilerDataForProcessPart(None)(runtimeContext.getUserCodeClassLoader).prepareExceptionHandler(runtimeContext)
       val jobData = compilerData.jobData
+      val engineRuntimeContext =
+        if (isTest) { eng =>
```
arkadius, line 150: It would be nice to move this factory logic to the one place where we control how things are created. We already have the `FlinkProcessCompilerDataFactory` which is our IoC container for the flink runtime.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkEngineRuntimeContextImpl.scala
```diff
@@ -4,13 +4,22 @@ import org.apache.flink.api.common.functions.RuntimeContext
...
 
 case class FlinkEngineRuntimeContextImpl(jobData: JobData, runtimeContext: RuntimeContext)
     extends FlinkEngineRuntimeContext {
   override val metricsProvider: MetricsProviderForScenario = new FlinkMetricsProviderForScenario(runtimeContext)
 
   override def contextIdGenerator(nodeId: String): ContextIdGenerator =
     new IncContextIdGenerator(jobData.metaData.id + "-" + nodeId + "-" + runtimeContext.getIndexOfThisSubtask)
+}
+
+case class FlinkTestEngineRuntimeContextImpl(jobData: JobData, runtimeContext: RuntimeContext)
```
arkadius, line 17: DRY. It looks like the original author of `FlinkEngineRuntimeContextImpl` cut the corners and instead of having logic of creation of `MetricsProviderForScenario`, moved this logic into the body of the class. We shouldn't follow this approach but clean up this.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerData.scala
```diff
@@ -38,9 +38,17 @@ class FlinkProcessCompilerData(
 
   def open(runtimeContext: RuntimeContext, nodesToUse: List[_ <: NodeData]): Unit = {
     val lifecycle = compilerData.lifecycle(nodesToUse)
-    lifecycle.foreach {
-      _.open(FlinkEngineRuntimeContextImpl(jobData, runtimeContext))
+    componentUseCase match {
+      case ComponentUseCase.TestRuntime =>
+        lifecycle.foreach {
```
arkadius, line 43: DRY. Only engine runtime context creation looks different for both cases. Opening behave the same

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerData.scala
```diff
@@ -70,8 +78,15 @@ class FlinkProcessCompilerData(
   def restartStrategy: RestartStrategies.RestartStrategyConfiguration = exceptionHandler.restartStrategy
 
   def prepareExceptionHandler(runtimeContext: RuntimeContext): FlinkExceptionHandler = {
-    exceptionHandler.open(FlinkEngineRuntimeContextImpl(jobData, runtimeContext))
-    exceptionHandler
+    componentUseCase match {
+      case ComponentUseCase.TestRuntime =>
+        exceptionHandler.open(FlinkTestEngineRuntimeContextImpl(jobData, runtimeContext))
```
arkadius, line 83: DRY. after removing duplication in open, you'll get a nice method that can be used here.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SourceMetricsFunction.scala
```diff
@@ -22,3 +23,20 @@ private[registrar] class SourceMetricsFunction[T](sourceId: String) extends Proc
   }
 
 }
+
+private[registrar] class TestSourceMetricsFunction[T](sourceId: String) extends ProcessFunction[T, T] {
```
arkadius, line 27: DRY. Please try to pass `EngineRuntimeContext` here. We did it in other places - take a look at things like `convertToEngineRuntimeContext: RuntimeContext => EngineRuntimeContext`. In case if it will be too hard to achieve, you can always extract this metric provider creation logic to one function.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
arkadius: It isn't in one place, I see at least four duplicated places. Take a look at comments below

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -132,11 +143,18 @@ class FlinkProcessRegistrar(
...
         compilerDataForProcessPart(None)(runtimeContext.getUserCodeClassLoader).prepareExceptionHandler(runtimeContext)
       val jobData = compilerData.jobData
+
       FlinkCustomNodeContext(
         jobData,
         nodeComponentId.nodeId,
         compilerData.processTimeout,
-        convertToEngineRuntimeContext = FlinkEngineRuntimeContextImpl(jobData, _),
+        // TODO: this has to be changed
+        convertToEngineRuntimeContext = eng => {
```
ForrestFairy, line 152: In proposed version in every place where `FlinkEngineRuntimeContextImpl` is used, `withProperMetricsProvider` method is called. This creates `FlinkEngineRuntimeContextImpl` with proper `MetricsProviderForScenario` (dummy provider for test, normal one otherwise).

This works everywhere apart from here in `FlinkProcessRegistrar`. Here, any use of `componentUseCase` results in `org.apache.flink.api.common.InvalidProgramException: The implementation of the RichMapFunction is not serializable` exception

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkEngineRuntimeContextImpl.scala
ForrestFairy: I'm not sure if that's what you meant, if not then I'll need some help with it. I've tried to base the `MetricsProviderForScenario` on the `ComponentUseCase` but there is one place where it doesn't work, as a serialization error is thrown (see comment below).
Apart from this problem, everywhere else one method is used which looks the same for test and normal use, so there is less code duplication. But again, I'm not sure if it's enough

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/MetricsProviderForFlink.scala
```diff
@@ -0,0 +1,19 @@
+package pl.touk.nussknacker.engine.process.compiler
+
+import org.apache.flink.api.common.functions.RuntimeContext
+import pl.touk.nussknacker.engine.api.process.ComponentUseCase
+import pl.touk.nussknacker.engine.util.metrics.{MetricsProviderForScenario, NoOpMetricsProviderForScenario}
+
+object MetricsProviderForFlink {
+
+  def getMetricsProvider(
```
arkadius, line 9: Let's rename it to `createMetricProvider`. `get` looks like sth is already created and you only return it.

arkadius (approved):

PR #5612, opened 2024-02-23
[Nu-5583] Rewrite environment migration endpoint to Tapir

DeamonDev:
## Describe your changes

Currently, upon clicking the "Migrate" button in the designer, an endpoint is triggered on the primary environment (PE), which then hits the secondary environment (**SE**) via the Akka HTTP client. The issue is that **PE** hits **SE** four times, using four different endpoints. This causes problems in maintaining the migration endpoint because any changes to these endpoints must be considered for the migration endpoint. Below is the simplified code currently executed from **PE** to **SE**:

```
for {
    validation <- validateScenarioGraph(localGraph, processName)
    remoteProcessE <- fetchProcessDetails(processName)
    _ <- remoteProcessE match {
            case Right(processDetails) => ...
            case Left(RemoteEnvCommunicationError.NotFound) => 
                createProcessOnRemote()
        }
    _ <- saveProcess

} yield ()
```

Additionally, one of the endpoints: `/processes/{processName}/{processCategory} `is only used here. Thus, after consolidating the logic of the four endpoints into one, it can be removed.

Let's examine where the above logic is called, i.e., the migration endpoint on **PE**:

```
path(ProcessNameSegment / VersionIdSegment / "migrate") { (procesName, version) => 
    withProcess(
        processName,
        version,
        details => 
            remoteEnvironment.migrate(details) //(1)
    )
}
```

Therefore, the line (1) should be changed to call the new endpoint (one) which will be executed on **SE**. This endpoint will accept basic data for scenario migration as proposed by @arkadius  and defined as `ScenarioWithDetailsForMigrations`. Additional parameters will include: `environmentId`, `processingMode`, and _engineSetupName_. Thus, I suggest creating a DTO:

```
    @derive(encoder, decoder, schema)
    final case class MigrateScenarioRequest(
        environmentId: String,
        processingMode: ProcessingMode,
        engineSetupName: EngineSetupName,
        scenarioWithDetailsForMigrations: ScenarioWithDetailsForMigrations
    )
```

The additional parameters mentioned above are available in the `details` object above.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/services/MigrationApiHttpService.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  private val remoteEnvironmentApiEndpoints = new MigrationApiEndpoints(authenticator.authenticationMethod())
+  private val passUsernameInMigration       = true
+
+  expose {
+    remoteEnvironmentApiEndpoints.migrateEndpoint
+      .serverSecurityLogic(authorizeKnownUser[NuDesignerError])
+      .serverLogicEitherT { implicit loggedUser => migrateScenarioRequest =>
+        {
+          val scenarioWithDetailsForMigrations = migrateScenarioRequest.scenarioWithDetailsForMigrations
```
arkadius, line 53: 1. Let's extract a domain service for that instead of keeping everything in the infrastructure layer.
2. I see that you added a lot of code but didn't remove any. What's going one? Can you provide some wider context on what are you planning to do in the PR description?

DeamonDev: I basically rewritten the content of `pl.touk.nussknacker.ui.process.migrate` `migrate` function. In this way we don't rely on four endpoints.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/scenariodetails/ScenarioWithDetailsForMigrations.scala
```diff
@@ -35,6 +46,8 @@ trait BaseScenarioWithDetailsForMigrations {
   def isArchived: Boolean
   def isFragment: Boolean
   def processingType: ProcessingType
+  def processingMode: ProcessingMode
```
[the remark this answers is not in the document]
DeamonDev: `processingMode` is needed because I need `ScenarioParameters` in order to create new process.

designer/server/src/main/scala/pl/touk/nussknacker/ui/services/MigrationApiHttpService.scala
arkadius: So what do you planning with the old one? What about tests?

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/scenariodetails/ScenarioWithDetailsForMigrations.scala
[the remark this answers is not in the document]
arkadius: Yeah, I forgot that migration code is working on the legacy scenario creating endpoint in order to handle the situation of migration between new and old version. Now we can fix that. Can you do that on the separate PR that would also clean up the old endpoint? It was left only for this purpose.

build.sbt
```diff
@@ -1491,9 +1491,10 @@ lazy val extensionsApi = (project in file("extensions-api"))
   .settings(
     name := "nussknacker-extensions-api",
     libraryDependencies ++= Seq(
-      "org.springframework"      % "spring-expression" % springV,
+      "org.springframework"          % "spring-expression" % springV,
+      "com.softwaremill.sttp.tapir" %% "tapir-core"        % tapirV,
```
coutoPL, line 1495: tapir code in extensions-api?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -6,9 +6,14 @@ import io.circe.generic.extras.semiauto.{deriveConfiguredDecoder, deriveConfigur
 import io.circe.{Decoder, Encoder, HCursor}
 import pl.touk.nussknacker.engine.api.process.ProcessName
 import pl.touk.nussknacker.engine.api.CirceUtil._
+import sttp.tapir.Schema
 
 @JsonCodec case class LayoutData(x: Long, y: Long)
 
+object LayoutData {
+  implicit val schema: Schema[LayoutData] = Schema.derived
```
coutoPL, line 14: IMO tapir schema should be defined close to tapir endpoint definitions. We should also ask ourselves if circe coders should be defined here too

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -116,4 +121,6 @@ object ProcessAdditionalFields {
 
   implicit val circeEncoder: Encoder[ProcessAdditionalFields] = deriveConfiguredEncoder
 
+  implicit val schema: Schema[ProcessAdditionalFields] = Schema.derived
```
coutoPL, line 124: the same as above

common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
```diff
@@ -83,6 +83,12 @@ object StreamMetaData {
 
 }
 
+case class CustomMetaData(customProperties: Map[String, String]) extends ScenarioSpecificData {
```
coutoPL, line 86: final case class?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
DeamonDev: I agree in a case when this is DTO used only for the sake of given endpoint. But what if structure is used a substructure in multiple endpoints (like `ScenarioGraph` class in my case)? Then according to your advice it would cause code repetition.

mk-software-pl: if it's shared by several tapir services, the schema and codecs should be defined in tapir's endpoints package (see eg. `TapirCodecs`).
Adding this dependency to Nu API module is a rather bad idea.

DeamonDev: Okay, thanks for pointing this out!

DeamonDev: I relaxed the schema since I do not think it is worth the price. Instead I could expand provided example. Of course it might be marked as TODO for the brave ones.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpoints.scala
```diff
@@ -0,0 +1,171 @@
...
+      Codec.string.map(
+        Mapping.from[String, FatalError](deserializationException)(_.getMessage)
+      )
+
+    @derive(encoder, decoder, schema)
+    final case class MigrateScenarioRequest(
+        sourceEnvironmentId: String,
+        processingMode: ProcessingMode,
+        engineSetupName: EngineSetupName,
+        scenarioWithDetailsForMigrations: ScenarioWithDetailsForMigrations
```
mk-software-pl, line 166: WDYT about renaming it to  `scenarioToMigrate`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -0,0 +1,202 @@
...
+import pl.touk.nussknacker.ui.process.repository.UpdateProcessComment
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+import pl.touk.nussknacker.ui.uiresolving.UIProcessResolver
+import pl.touk.nussknacker.ui.validation.FatalValidationError
+import pl.touk.nussknacker.ui.{NuDesignerError, UnauthorizedError}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+trait MigrationService {
```
mk-software-pl, line 32: let's remove the interface. When you don't know how to name the implementation and you add the "Impl" suffix, it almost always means that you don't need the interface at this moment.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -0,0 +1,202 @@
...
+      _ <- processService
+        .updateProcess(processIdWithName, updateScenarioCommand)
+        .withSideEffect(response =>
+          response.processResponse.foreach(resp => notifyListener(OnSaved(resp.id, resp.versionId)))
+        )
+        .map(_.validationResult)
+
+    } yield ()
+
+    val transformedFuture: Future[Either[NuDesignerError, Unit]] =
```
mk-software-pl, line 146: why not like this:
```
      future.transform[Either[NuDesignerError, Unit]] {
        case Failure(e: NuDesignerError) => Success(Left(e))
        case Failure(e: Throwable) => Failure(e)
        case Success(()) => Success(Right(()))
      }
```
?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers
+    with RestAssuredVerboseLogging {
+
+  private val sourceEnvironmentId = "DEV"
+
+  private val exampleProcessName = ProcessName("test2")
+  private val illegalProcessName = ProcessName("#test")
+
+  private val exampleScenario =
```
mk-software-pl, line 41: let's move all the less important things after the test cases. Test cases are the most important part of a Spec, so they should be exposed.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+          createSavedScenario(exampleScenario, Category1)
+        )
+        .when()
+        .basicAuthAllPermUser()
+        .jsonBody(validRequestData)
+        .post(s"$nuDesignerHttpAddress/api/migrate")
+        .Then()
+        .statusCode(200)
+
+      given()
```
mk-software-pl, line 111: this is the assertion right? 
Please, take a look at the `ScenarioActivityApiHttpServiceBusinessSpec` and the `verifyCommentExists` assertion. Should we extract it to some helper? (eg. `NuRestAssureExtensions`?)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+        .applicationState(
+          createSavedScenario(exampleScenario, Category1)
+        )
+        .when()
+        .basicAuthAllPermUser()
+        .jsonBody(validRequestData)
+        .post(s"$nuDesignerHttpAddress/api/migrate")
+        .Then()
+        .statusCode(200)
```
mk-software-pl, line 109: let's assert the empty plain body too

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+      given()
+        .applicationState(
+          createSavedScenario(validFragment, Category1)
+        )
+        .when()
+        .basicAuthAllPermUser()
+        .jsonBody(validRequestDataForFragment)
+        .post(s"$nuDesignerHttpAddress/api/migrate")
+        .Then()
+        .statusCode(200)
```
mk-software-pl, line 182: let's assert the empty plain body too

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsSecuritySpec.scala
```diff
@@ -0,0 +1,108 @@
...
+    with WithRichConfigRestAssuredUsersExtensions
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers
+    with RestAssuredVerboseLogging {
+
+  private val sourceEnvironmentId = "DEV"
+
+  private val exampleProcessName = ProcessName("test")
+
+  private val exampleScenario =
```
mk-software-pl, line 35: the same comment as in the spec above

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsSecuritySpec.scala
```diff
@@ -0,0 +1,108 @@
...
+       |    "validationResult": ${validationResult.asJson.noSpaces},
+       |    "history": null,
+       |    "modelVersion": null
+       |  }
+       |}
+       |""".stripMargin
+
+  private val requestData: String = prepareRequestData(exampleProcessName.value)
+
+  "The endpoint for scenario migration between environments should" - {
```
mk-software-pl, line 67: can we follow the convention defined in the other SecuritySpecs?
we test authentication success, authentication failure, no credentials case (anonymous user)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsSecuritySpec.scala
```diff
@@ -0,0 +1,108 @@
...
+import pl.touk.nussknacker.test.config.WithRichDesignerConfig.TestCategory.Category1
+import pl.touk.nussknacker.test.config.{
+  WithMockableDeploymentManager,
+  WithRichConfigRestAssuredUsersExtensions,
+  WithRichDesignerConfig
+}
+import pl.touk.nussknacker.test.{NuRestAssureExtensions, NuRestAssureMatchers, RestAssuredVerboseLogging}
+import pl.touk.nussknacker.ui.process.marshall.CanonicalProcessConverter
+
+class MigrationApiEndpointsSecuritySpec
```
mk-software-pl, line 20: we agreed that the naming convention for the tapir API specs looks like this:
```
{name of the http service}{Business|Security}Spec
``` 
could you please rename it?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+import pl.touk.nussknacker.test.config.WithRichDesignerConfig.TestCategory.Category1
+import pl.touk.nussknacker.test.config.{
+  WithMockableDeploymentManager,
+  WithRichConfigRestAssuredUsersExtensions,
+  WithRichDesignerConfig
+}
+import pl.touk.nussknacker.test.{NuRestAssureExtensions, NuRestAssureMatchers, RestAssuredVerboseLogging}
+import pl.touk.nussknacker.ui.process.marshall.CanonicalProcessConverter
+
+class MigrationApiEndpointsBusinessSpec
```
mk-software-pl, line 25: we agreed that the naming convention for the tapir API specs looks like this:
```
{name of the http service}{Business|Security}Spec
``` 
could you please rename it?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResourcesSpec.scala
```diff
@@ -217,6 +213,16 @@ class RemoteEnvironmentResourcesSpec
       Future.successful(Right(testMigrationResults))
     }
 
+    override def migrate(
+        processingMode: ProcessingMode,
+        engineSetupName: EngineSetupName,
+        scenarioWithDetailsForMigrations: ScenarioWithDetailsForMigrations
+    )(implicit ec: ExecutionContext, loggedUser: LoggedUser): Future[Either[NuDesignerError, Unit]] = {
+      val localScenarioGraph = scenarioWithDetailsForMigrations.scenarioGraphUnsafe
```
mk-software-pl, line 221: because it's unsafe, maybe we should be sure that the expectation won't be thrown outside the Future?

mk-software-pl (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
```diff
@@ -88,4 +131,23 @@ object TapirCodecs {
...
+
+  object ProcessNameCodec {
+    implicit val processNameSchema: Schema[ProcessName] = Schema.string
+  }
+
+  object ScenarioWithDetailsForMigrationsCodec {
+
+    import ProcessNameCodec._
+
+    implicit val scenarioWithDetailsForMigrationsSchema: Schema[ScenarioWithDetailsForMigrations] = Schema.anyObject
```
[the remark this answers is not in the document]
DeamonDev: I wrote it in another comment I would like to relax it ATM and left it as TODO. Typing this is no trivial since
there are a lot of substructures and some of them are not Tapir-schema friendly (e.g. `TypingResult`)

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
DeamonDev: I tried to mimic what I've seen in the neighbour package(s), for example `pl.touk.nussknacker.ui.notifications.NotificationService`. Also I think that was suggested by @arkadius in his comment to this PR: 

"Let's extract a domain service for that instead of keeping everything in the infrastructure layer."

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
mk-software-pl: ok, let's add the proper `// TODO:`

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
mk-software-pl: For me @arkadius wants you to keep `MigrationApiService` as thin as possible. And it's good. But when your domain service has only one implementation (and you have no idea what the other implementation could look like), there is no sense in keeping the interface. That's why I propose to remove it (if you want to you can do the same for NotificationService trait)

arkadius: I agree. If there is only one implementation, the benefits of having the interface separated are lower than the drawbacks that it causes. Eventually, I would merge the interface with implementation in every place where we have such a situation to avoid confusion.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpoints.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  val exampleGraph = CanonicalProcessConverter.toScenarioGraph(exampleProcess)
+
+  private val errorValidationResult =
+    ValidationResults.ValidationResult.success
+
+  lazy val migrateEndpoint: SecuredEndpoint[MigrateScenarioRequest, NuDesignerError, Unit, Any] =
+    baseNuApiEndpoint
+      .summary("Migration between environments service")
+      .tag("migration")
```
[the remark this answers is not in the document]
mk-software-pl: Capitalized ;) to be consistent with other tags

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
mk-software-pl: I'm not sure if for the determined developer, the comment will be clear. Maybe we could add the link to this discussion?

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
mk-software-pl: I think we misunderstood each other. The `MigrationApiHttpService` was OK. We asked you to remove the `MigrationService` trait and rename the `MigrationServiceImpl` to `MigrationService`. That's all.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpointsBusinessSpec.scala
mk-software-pl: no comment, no fix

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpoints.scala
```diff
@@ -0,0 +1,171 @@
...
+import derevo.circe._
+import pl.touk.nussknacker.engine.api.component.ProcessingMode
+import pl.touk.nussknacker.engine.api.process.ProcessName
+import pl.touk.nussknacker.engine.build.ScenarioBuilder
+import pl.touk.nussknacker.engine.deployment.EngineSetupName
+import pl.touk.nussknacker.restmodel.BaseEndpointDefinitions
+import pl.touk.nussknacker.restmodel.BaseEndpointDefinitions.SecuredEndpoint
+import pl.touk.nussknacker.restmodel.scenariodetails.ScenarioWithDetailsForMigrations
+import pl.touk.nussknacker.restmodel.validation.ValidationResults
+import pl.touk.nussknacker.restmodel.validation.ValidationResults.{NodeValidationError, NodeValidationErrorType}
```
mk-software-pl, line 13: unused import

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -81,6 +87,11 @@ final case class MigrationValidationError(errors: ValidationErrors)
       s"Cannot migrate, following errors occurred: ${messages.mkString(", ")}"
     })
 
+case class MissingScenarioGraphError(msg: String)
```
mk-software-pl, line 90: final case class

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpBusinessSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+    with WithRichScenarioActivitySpecAsserts
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers
+    with RestAssuredVerboseLogging {
+
+  "The endpoint for scenario migration between environments should" - {
+    "migrate scenario and add update comment" in {
+      given()
+        .applicationState(
+          createSavedScenario(exampleScenario, Category1)
```
mk-software-pl, line 33: `Suspicious forward reference in class`
let's declare the examples and other private vals declared after the test as lazy. It will solve the warning

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpSecuritySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers
+    with RestAssuredVerboseLogging {
+
+  "The endpoint for scenario migration between environments when" - {
+    "authenticated should" - {
+      "return response for allowed scenario" in {
+        given()
+          .applicationState(
+            createSavedScenario(exampleScenario, Category1)
```
mk-software-pl, line 36: the same in this file

mk-software-pl (changes requested):

mk-software-pl (approved):
LGTM

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -49,10 +56,9 @@ trait RemoteEnvironment {
   def processVersions(processName: ProcessName)(implicit ec: ExecutionContext): Future[List[ScenarioVersion]]
 
   def migrate(
-      localProcess: ScenarioGraph,
-      processName: ProcessName,
-      parameters: ScenarioParameters,
-      isFragment: Boolean
+      processingMode: ProcessingMode,
+      engineSetupName: EngineSetupName,
+      scenarioToMigrate: ScenarioWithDetailsForMigrations
```
arkadius, line 61: 1. Let's remove this temporary class and extract minimal set of information that we need for migration
2. Let's consider various combinations of backward/forward incompatible changes and versioning schemas

arkadius (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -182,40 +193,19 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
-          )
-        case Right(_) => EitherT.rightT[Future, NuDesignerError](())
-        case Left(RemoteEnvironmentCommunicationError(StatusCodes.NotFound, _)) =>
-          val userToForward = if (passUsernameInMigration) Some(loggedUser) else None
-          createProcessOnRemote(processName, parameters, isFragment, userToForward)
-        case Left(other) => EitherT.leftT[Future, NuDesignerError](other)
-      }
-      usernameToPass = if (passUsernameInMigration) Some(RemoteUserName(loggedUser.username)) else None
-      _ <- EitherT {
-        saveProcess(
```
arkadius, line 211: Please cleanup methods that are dead code after your changes.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
```diff
@@ -104,6 +104,9 @@ object ScenarioBuilder {
   def streamingLite(id: String) =
     new ProcessMetaDataBuilder(MetaData(id, LiteStreamMetaData()))
 
+  def withCustomMetaData(id: String, properties: Map[String, String]) =
```
arkadius, line 107: Why do you need that?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpBusinessSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+  ValidationResult
+}
+import pl.touk.nussknacker.test.base.it.{NuItTest, WithRichConfigScenarioHelper}
+import pl.touk.nussknacker.test.config.WithRichDesignerConfig.TestCategory.Category1
+import pl.touk.nussknacker.test.config.{WithMockableDeploymentManager, WithRichDesignerConfig}
+import pl.touk.nussknacker.test.processes.WithRichScenarioActivitySpecAsserts
+import pl.touk.nussknacker.test.{NuRestAssureExtensions, NuRestAssureMatchers, RestAssuredVerboseLogging}
+import pl.touk.nussknacker.ui.process.marshall.CanonicalProcessConverter
+
+class MigrationApiHttpBusinessSpec
```
arkadius, line 22: Change the name of classes to be able to use go to test feature in idea (ctrl+t). To achieve that, test class need to have the same prefix as production class - in this case MigrationApiHttpService*. The same with the second test class

mk-software-pl:
oh, and the Changelog needs to be updated too ;)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
```diff
@@ -83,7 +82,7 @@ class MigrationApiHttpSecuritySpec
           .jsonBody(requestData)
           .post(s"$nuDesignerHttpAddress/api/migrate")
           .Then()
-          .statusCode(401)
+          .statusCode(500)
```
arkadius, line 85: wtf? :)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceBusinessSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+        .applicationState(
+          createSavedScenario(exampleScenario, Category1)
+        )
+        .when()
+        .basicAuthAllPermUser()
+        .jsonBody(validRequestData)
+        .post(s"$nuDesignerHttpAddress/api/migrate")
+        .Then()
+        .statusCode(200)
+        .verifyCommentExists(exampleProcessName.value, "Scenario migrated from DEV by allpermuser")
```
arkadius, line 45: a comment prove that migration is done? :) I thought that it is a minor side effect and new shape of scenario graph is the most important :)

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceBusinessSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+      given()
+        .applicationState(
+          createSavedScenario(validFragment, Category1)
+        )
+        .when()
+        .basicAuthAllPermUser()
+        .jsonBody(validRequestDataForFragment)
+        .post(s"$nuDesignerHttpAddress/api/migrate")
+        .Then()
+        .statusCode(200)
```
arkadius, line 84: status code 200 and empty response body means that it works?

scenario-api/src/main/scala/pl/touk/nussknacker/engine/build/ScenarioBuilder.scala
arkadius: Let's write a comment why we change production api or remove redundant properties verification

DeamonDev: I need that for the sake of UI validation. Based on our private discussion I will remove my factory method thanks to removal of metadata checks.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
DeamonDev: Shame on me, I should mark somehow for further investigation. Short story short: after merging master today this started returning `500`. I had a lot to do and since I like green color in CI I decided to change it for a moment. I should add some commentary for that.  Sorry 4 that

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceBusinessSpec.scala
DeamonDev: You're right I should request for that new shape.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceBusinessSpec.scala
DeamonDev: I think I wanted to check that _it does not fail_.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -0,0 +1,182 @@
...
+  def migrate(
+      migrateScenarioRequest: MigrateScenarioRequest
+  )(implicit loggedUser: LoggedUser): Future[Either[NuDesignerError, Unit]] = {
+    val sourceEnvironmentId = migrateScenarioRequest.sourceEnvironmentId
+    val targetEnvironmentId = config.getString("environment")
+    val processingMode      = migrateScenarioRequest.processingMode
+    val processCategory     = migrateScenarioRequest.processCategory
+    val engineSetupName     = migrateScenarioRequest.engineSetupName
+    val parameters          = ScenarioParameters(processingMode, processCategory, engineSetupName)
+    val processingType      = migrateScenarioRequest.processingType
```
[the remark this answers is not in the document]
arkadius: What about that?

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithSimplifiedScenarioActivitySpecAsserts.scala
```diff
@@ -0,0 +1,99 @@
...
+        .Then()
+        .statusCode(200)
+        .body(
+          matchJsonWithRegexValues(
+            s"""
+               |{
+               |  "comments": [
+               |    {
+               |      "id": "${regexes.digitsRegex}",
+               |      "processVersionId": 1,
```
arkadius, line 34: Don't fix parameters. If you think that it is not necessary to check it, use regexes instead. But TBH author of the comment is quite important thing to check. I remember that there is some trick that we are forwarding the user but this user is only visible in the comment body, not as an author of the comment.
Also

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithRichScenarioActivitySpecAsserts.scala
```diff
@@ -0,0 +1,99 @@
...
+             |}
+             |""".stripMargin
+        )
+    }
+
+  }
+
+  implicit class VerifyAttachmentsExists[T <: ValidatableResponse](validatableResponse: T) {
+
+    def verifyAttachmentsExists(scenarioName: String): ValidatableResponse = {
```
arkadius, line 71: Duplicated, not used. Please check the diff before you give it to someone to review.

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithRichScenarioActivitySpecAsserts.scala
```diff
@@ -0,0 +1,99 @@
...
+               |""".stripMargin
+          )
+        )
+    }
+
+  }
+
+  implicit class VerifyEmptyCommentsAndAttachments[T <: ValidatableResponse](validatableResponse: T) {
+
+    def verifyEmptyCommentsAndAttachments(scenarioName: String): ValidatableResponse = {
```
arkadius, line 51: Duplicated, not used. Please check the diff before you give it to someone to review.

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithSimplifiedScenarioActivitySpecAsserts.scala
```diff
@@ -0,0 +1,99 @@
...
+trait WithSimplifiedScenarioActivitySpecAsserts
+    extends AnyFreeSpecLike
+    with NuItTest
+    with WithSimplifiedDesignerConfig
+    with WithSimplifiedConfigRestAssuredUsersExtensions
+    with NuRestAssureMatchers {
+
+  implicit class VerifyCommentExists[T <: ValidatableResponse](validatableResponse: T) {
+
+    def verifyCommentExists(scenarioName: String, commentContent: String): ValidatableResponse = {
```
arkadius, line 20: DRY - it is also in `WithRichScenarioActivitySpecAsserts`

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
DeamonDev: Done, sorry for over-sighting it. I'll try to look more carefully next time.

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithSimplifiedScenarioActivitySpecAsserts.scala
DeamonDev: It is hard not to DRY in that case, since we have already divisions: 

- `WithRichConfigRestAssuredUsersExtensions` vs `WithSimplifiedConfigRestAssuredUsersExtensions`
- `WithRichDesignerConfig` vs `WithSimplifiedDesignerConfig`

arkadius (commented):
I didn't finish the review, but I already have some comments. I'll back to it probably tomorrow

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -0,0 +1,208 @@
...
+          case Left(e) => Future.successful[Either[NuDesignerError, ValidationResults.ValidationResult]](Left(e))
+          case Right(uiProcessResolverO) =>
+            uiProcessResolverO match {
+              case Some(uiProcessResolver) =>
+                Future.successful(
+                  FatalValidationError.renderNotAllowedAsError(
+                    uiProcessResolver.validateBeforeUiResolving(scenarioGraph, processName, isFragment)
+                  )
+                )
+              case None => Future.successful(Left(new MissingProcessResolverError(loggedUser, processingType)))
```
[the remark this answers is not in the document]
DeamonDev: Ad. chasing my own tail. I just wanted to avoid [triangles of death](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)).

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
```diff
@@ -71,9 +72,12 @@ class RemoteEnvironmentResources(
                 version,
                 details =>
                   remoteEnvironment.migrate(
+                    details.processingMode,
+                    details.engineSetupName,
+                    details.processCategory,
+                    details.processingType,
```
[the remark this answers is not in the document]
DeamonDev: It is stupid now, but I created second PR which makes it not to be true anymore: [https://github.com/TouK/nussknacker/pull/5748](https://github.com/TouK/nussknacker/pull/5748) 

I decided to separate these 4 clarity.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -0,0 +1,208 @@
...
+        }
+      }
+
+    transformedFuture
+  }
+
+  private def resolveValidatedProcessingType(
+      validatedProcessingType: Validated[NuDesignerError, ProcessingType]
+  ): Future[ProcessingType] =
+    validatedProcessingType.fold(Future.failed, Future.successful)
```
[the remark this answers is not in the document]
DeamonDev: You are right. I exaggerated here.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
arkadius: You could use `EitherT` - it is natural consequence when you wan to to have Future[Either in the method's signature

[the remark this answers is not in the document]
DeamonDev: Could we agree to add these changes in a new PR and atm merge it as it is? 

https://github.com/TouK/nussknacker/pull/5771/files

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithScenarioActivitySpecAsserts.scala
```diff
@@ -0,0 +1,99 @@
...
+        .Then()
+        .body(
+          matchJsonWithRegexValues(
+            s"""
+               |{
+               |  "comments": [],
+               |  "attachments": [
+               |    {
+               |      "id": "${regexes.digitsRegex}",
+               |      "processVersionId": 1,
```
arkadius, line 85: You still fix values as I reported you in  https://github.com/TouK/nussknacker/pull/5612#discussion_r1523548905

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
arkadius: Sure

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  private def prepareRequestData(scenarioName: String): String =
+    s"""
+       |{
+       |  "sourceEnvironmentId": "$sourceEnvironmentId",
+       |  "processingMode": "Unbounded-Stream",
+       |  "engineSetupName": "Mockable",
+       |  "processName": "${scenarioName}",
+       |  "isFragment": false,
+       |  "processingType": "streaming1",
```
arkadius, line 115: We use `TestProcessingType` / `TestCategory` in such cases

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResourcesSpec.scala
```diff
@@ -217,6 +214,21 @@ class RemoteEnvironmentResourcesSpec
...
+        engineSetupName: EngineSetupName,
+        processCategory: String,
+        scenarioGraph: ScenarioGraph,
+        processName: ProcessName,
+        isFragment: Boolean
+    )(implicit ec: ExecutionContext, loggedUser: LoggedUser): Future[Either[NuDesignerError, Unit]] = {
+      val localScenarioGraph = scenarioGraph
+      migrateInvocations = localScenarioGraph :: migrateInvocations
+      Future.successful(Right(()))
+    }.recover[Either[NuDesignerError, Unit]] { case e: IllegalStateException =>
```
arkadius, line 228: WTF?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
DeamonDev: Fortunately, we dont need anymore `processingType` in request. I removed it from the DTO some time ago but it seems that the excess is ignored during the serialization process.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpoints.scala
```diff
@@ -0,0 +1,159 @@
+package pl.touk.nussknacker.ui.api
```
arkadius, line 1: Now we keep these classes inside api/description

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
arkadius: LOL i forgot about that. We should add more restrive requests checks in tests. The part regarding the category is still valid

DeamonDev: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/MigrationApiEndpoints.scala
DeamonDev: Done

arkadius (approved):
Good job, fix tests before merging

PR #5621, opened 2024-02-27
fetching/parsing batch periodic json only when needed (stop parsing during status check)

raphaelsolarski:


arkadius (approved):
LGTM

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
```diff
@@ -48,17 +49,36 @@ object PeriodicProcessesRepository {
...
       .fold(e => throw new IllegalArgumentException(e), identity)
     PeriodicProcess(
       processEntity.id,
       model.DeploymentWithJarData(
         processVersion = processVersion,
-        canonicalProcess = processEntity.processJson,
         inputConfigDuringExecutionJson = processEntity.inputConfigDuringExecutionJson,
-        jarFileName = processEntity.jarFileName
+        jarFileName = processEntity.jarFileName,
+        process = processEntity.processJson.get // todo: add exception ?
```
piotrp, line 62: Yes, please , with some basic information from context, so that if it ever fails we'll have at least something useful in logs

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesTable.scala
```diff
@@ -98,13 +102,62 @@ trait PeriodicProcessesTableFactory {
 
   }
 
-  object PeriodicProcesses extends TableQuery(new PeriodicProcessesTable(_))
+  class PeriodicProcessesWithUnitTable(tag: Tag) extends PeriodicProcessesTable(tag) {
```
piotrp, line 105: Maybe `PeriodicProcessWithoutJson`, to keep names consistent?

piotrp (approved):

PR #5627, opened 2024-02-27
[NU-1470] Sql file based basic source

mslabek:
## Checklist before merge


engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  def extractTablesFromFlinkRuntime(
+      createTableStatements: List[SqlStatement]
+  ): List[Either[SqlExtractorError, SqlDataSourceConfig]] = {
+    val settings = EnvironmentSettings
+      .newInstance()
+      .withBuiltInCatalogName(builtInCatalogName)
+      .withBuiltInDatabaseName(buildInDatabaseName)
+      .build()
+    val tableEnv = TableEnvironment.create(settings)
```
arkadius, line 34: I see some a problem here. This method creates fat objects like `ResourceManager`. Each of them are `Closable` but `TableEnvironment` isn't. I see that it is on the Flink's roadmap in https://cwiki.apache.org/confluence/display/FLINK/FLIP-36%3A+Support+Interactive+Programming+in+Flink. This method will be run multiple times, e.g. we have a model reloading mechanism :-/ Do you have some idea how to deal with this?

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,127 @@
...
+    def extractOne(
+        statement: SqlStatement
+    ): Either[SqlExtractorError, SqlDataSourceConfig] = {
+      implicit val statementImplicit: SqlStatement = statement
+      for {
+        _ <- tryExecuteStatement(statement, tableEnv)
+        // Need to grab table from env because table from result has wrong schema
+        tableName <- tableEnv
+          .listTables()
+          .headOption
```
arkadius, line 69: What about the rest of the list? :)  I see that we broken a YAGNI rule a bit. IMO we should start with more naive implementation:
- execute all statements
- do the rest with all tables.

Currently, it looks like handles some edge cases but it is hard to define which one it handles and which not. We can dive into this or we can just ignore edge cases and follow the YAGNI rule

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,127 @@
...
+    connector: Connector,
+    schema: Schema,
+    sqlCreateTableStatement: SqlStatement
+)
+
+object SqlDataSourceConfig {
+  type Connector = String
+}
+
+final case class Schema(columns: List[Column], typingResult: TypingResult)
```
arkadius, line 126: `columns` are not used in the production code

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+      FinalResults.forValidation(context, Nil, None)(initializer.validationContext)
+  }
+
+  override def implementation(
+      params: Params,
+      dependencies: List[NodeDependencyValue],
+      finalStateOpt: Option[State]
+  ): Source = {
+    val tableName     = tableNameParam.extractValue(params)
+    val selectedTable = getSelectedTableUnsafe(tableName)
```
arkadius, line 59: You can leverage the `finalStateOpt` instead of doing this twice.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,127 @@
...
+  )(implicit statement: SqlStatement): Either[SqlExtractorError, TableMetaData] = {
+    tableWithMetadata.getOptions.asScala
+      .get(connectorKey)
+      .toRight(SqlExtractorError(ConnectorMissing))
+      .map(TableMetaData)
+  }
+
+}
+
+final case class SqlDataSourceConfig(
```
arkadius, line 115: It suggest that this class is a "dumb" representation of the object from the configuration. But in fact it is a smart object with determined typing result and extracted table name. We usually name this kind of object as `Definition` - I don't know if it is the correct name or not. Take a look at `pl.touk.nussknacker.sql.db.schema.TableDefinition`.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,127 @@
...
+    tableName: String,
+    connector: Connector,
```
arkadius, line 117: not used

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
```diff
@@ -0,0 +1,83 @@
...
+    s"tableApi-source-sql-${config.componentNameSuffix}"
+
+  private def extractDataSourceConfigFromSqlFile(filePath: String): List[SqlDataSourceConfig] = {
+    val sqlStatements = readSqlFromFile(Paths.get(filePath))
+    val results       = extractTablesFromFlinkRuntime(sqlStatements)
+
+    results.flatMap {
+      case Right(config) => Some(config)
+      case Left(error) =>
+        logger.error(error.toString)
```
arkadius, line 58: I don't think that it is a correct approach :) We should fail.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/SqlExtractorErrors.scala
```diff
@@ -0,0 +1,50 @@
+package pl.touk.nussknacker.engine.flink.table.extractor
+
+import pl.touk.nussknacker.engine.flink.table.extractor.SqlStatementReader.SqlStatement
+
+final case class SqlExtractorError(
+    errorType: SqlExtractionErrorType,
```
arkadius, line 6: Neight of this fields is used. YAGNI.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/SqlExtractorErrors.scala
```diff
@@ -0,0 +1,50 @@
...
+sealed trait SqlExtractionErrorType {
+  def message: String
+}
+
+object StatementNotExecuted extends SqlExtractionErrorType {
+
+  // TODO: explain our required syntax - for example we don't handle explicit catalog / db name
+  override val message: String =
+    """
+      |Could not execute sql statement. The statement may be malformed. The statement has to be a CREATE TABLE statement
```
arkadius, line 30: Would be nice to have the statement content in this messages :)

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+  override def isCompatible(version: NussknackerVersion): Boolean = true
+
+  override def isAutoLoaded: Boolean = false
+
+}
+
+final case class SqlFileDataSourceConfig(
+    componentNameSuffix: String,
```
arkadius, line 81: Why this suffix? It smells like YAGNI ;-) Keep in mind that we already have the built-in mechanism prefixing components - see `ComponentProviderConfig.componentPrefix`

engine/flink/tests/src/test/scala/pl/touk/nussknacker/devmodel/SqlKafkaPingPongTest.scala
```diff
@@ -0,0 +1,92 @@
...
+       | ) WITH (
+       |  'connector' = 'kafka',
+       |  'topic' = '$inputTopicNameTest1',
+       |  'properties.bootstrap.servers' = '${kafkaServer.kafkaAddress}',
+       |  'properties.group.id' = 'someConsumerGroupId',
+       |  'scan.startup.mode' = 'earliest-offset',
+       |  'format' = 'json'
+       |);""".stripMargin
+
+  private lazy val sqlTablesDefinitionFilePath = {
```
arkadius, line 40: Instead of creating the temp file, we can replace `SqlFileDataSourceConfig.sqlFilePath` with URL and use pass a resources location. WDYT?

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+    with SourceFactory {
+
+  override type State = Nothing
+
+  private val tableNamParamName = "Table"
+
+  private val tableNameParam: ParameterWithExtractor[String] = {
+    val param                    = ParameterWithExtractor.mandatory[String](tableNamParamName)
+    val possibleTableParamValues = configs.map(c => FixedExpressionValue(s"'${c.tableName}'", c.tableName))
+    param.copy(parameter =
```
arkadius, line 26: Ah this api is ugly :( But it is not a part of this change

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
mslabek: That was a great idea, I've implemented a simpler approach with less obstructions and now theoretically even arbitrary catalogs should work (if we add dependencies to classpath).

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
mslabek: I've flattened the `DataSourceTableDefinition` and removed unused fields.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
mslabek: Done :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
mslabek: removed :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
mslabek: Switched it to throw exception and blow up :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/SqlExtractorErrors.scala
mslabek: They were supposed to be logged/part of exception message - I've added a message val based on these fields that is part of exception now

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/SqlExtractorErrors.scala
mslabek: Same as comment above

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
mslabek: Right, controlling naming at this point is needless. Removed :+1:

engine/flink/tests/src/test/scala/pl/touk/nussknacker/devmodel/SqlKafkaPingPongTest.scala
mslabek: I wanted to use a path relative to the base path since it's easier to pass this path through env variables in designer configuration and in flink container. For this test I have to compose a file with content depending on kafka address, so I don't think I can do it like that.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
mslabek: Changed it a bit to first construct a parameter and then wrap it into the `ParameterWithExtractor`

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
arkadius: Let's leave a TODO

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    val tableDefinitions = for {
+      catalogName  <- tableEnv.listCatalogs().toList
+      catalog      <- tableEnv.getCatalog(catalogName).toScala.toList
+      databaseName <- catalog.listDatabases.asScala.toList
+      tableName    <- tableEnv.listTables(catalogName, databaseName).toList
+      // table path may be different for some catalog-managed tables - for example JDBC catalog adds a schema name to
+      // table name when listing tables, but querying using tablePath with catalog.database.schema.tableName throws
+      // exception
+      tablePath = s"$catalogName.$databaseName.$tableName"
```
arkadius, line 41: How flink-sql deal with such a long names? Haven't they got some kind of default catalog / default database concepts?

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+      )
+    }
+
+    SqlDataSourcesDefinition(results.tableDefinitions, sqlStatements)
+  }
+
+  private def readSqlFromFile(pathToFile: Path): List[SqlStatement] =
+    Try {
+      val fileContent = ResourceLoader.load(pathToFile)
+      SqlStatementReader.readSql(fileContent.mkString)
```
arkadius, line 65: This line can't throw any exception. Am I right?

mslabek: Yes, it just splits the lines and trims them

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/SqlComponentProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+    } match {
+      case Failure(exception) =>
+        logger.warn(s"Couldn't parse sql tables definition: $exception")
+        List.empty
```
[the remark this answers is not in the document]
mslabek: Done :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
mslabek: Normally you can use `USE` statements which change env context - https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/dev/table/sql/use/#use-catalog. There is also the default catalog and database created when table env is initialized. In this implementation I wanted to not depend on context.

engine/flink/tests/src/test/scala/pl/touk/nussknacker/devmodel/SqlKafkaPingPongTest.scala
mslabek: done :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/extractor/DataSourceSqlExtractor.scala
mslabek: Added a TODO with a link to this comment

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  override def nodeDependencies: List[NodeDependency] = List(TypedNodeDependency[NodeId])
+
+  override val allowedProcessingModes: Option[Set[ProcessingMode]] = Some(Set(ProcessingMode.UnboundedStream))
+
+  private def getSelectedTableUnsafe(tableName: String): DataSourceTableDefinition =
+    defs.tableDefinitions
+      .find(_.tableName == tableName)
+      .getOrElse(throw new IllegalStateException("Table with selected name not found."))
```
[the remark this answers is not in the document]
arkadius: Ah, indeed. I had some brain fog, it is a few numbers above. So in fact, it is Illegal state, because this situation would happen only if something went wrong in our compiler.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
```diff
@@ -0,0 +1,70 @@
...
+
+class SqlSourceFactory(defs: SqlDataSourcesDefinition) extends SingleInputDynamicComponent[Source] with SourceFactory {
+
+  override type State = DataSourceTableDefinition
+
+  private val tableNameParamName = "Table"
+
+  private val tableNameParam: ParameterWithExtractor[String] = {
+    val possibleTableParamValues =
+      defs.tableDefinitions.map(c => FixedExpressionValue(s"'${c.tableName}'", c.tableName))
```
arkadius, line 24: We usually adds an extra "non-picked" option which is the default option. Thanks to that we avoid situation that we pick the table that user don't want to pick. Take a look at other places of `FixedValuesParameterEditor` usage.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/SqlSourceFactory.scala
```diff
@@ -0,0 +1,70 @@
...
+    val parameter = Parameter[String](
+      name = tableNameParamName
+    ).copy(
+      editor = Some(FixedValuesParameterEditor(possibleTableParamValues)),
+      validators = List(FixedValuesValidator(possibleTableParamValues))
```
arkadius, line 29: AFAIK you don't need to add this validators. They will be enriched automatically based on the editor.

PR #5632, opened 2024-02-27
Remove unnecessary Optional wrapping in Params class

bartektartanus:
## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
@@ -23,7 +23,7 @@ final case class Params(nameToValueMap: Map[String, Any]) {
       .getOrElse(throw new IllegalArgumentException(cannotFindParamNameMessage(paramName)))
   }
 
-  private def rawValueExtract(paramName: String) = nameToValueMap.get(paramName).flatMap(Option.apply)
+  private def rawValueExtract(paramName: String) = nameToValueMap.get(paramName)
 
   private def cannotFindParamNameMessage(paramName: String) =
     s"Cannot find param name [$paramName]. Available param names: ${nameToValueMap.keys.mkString(",")}"
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -48,6 +48,7 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   * `ComponentId` was renamed to `DesignerWideComponentId`
   * new `ComponentId` is serialized in json to string in format `$componentType-$componentName` instead of separate fields (`name` and `type`)
   * `NodeComponentInfo.componentInfo` was renamed to `componentId`
+* [#5438](https://github.com/TouK/nussknacker/pull/5438) Removed sealed trait `CustomActionError`, now `CustomActionResult` is always used
 * [#5465](https://github.com/TouK/nussknacker/pull/5465) [#5457](https://github.com/TouK/nussknacker/pull/5457) Typed related changes
   * `CommonSupertypeFinder` shouldn't be created directly anymore - `CommonSupertypeFinder.*` predefined variables should be used instead,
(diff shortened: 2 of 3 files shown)


arkadius (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
```diff
@@ -23,7 +23,7 @@ final case class Params(nameToValueMap: Map[String, Any]) {
       .getOrElse(throw new IllegalArgumentException(cannotFindParamNameMessage(paramName)))
   }
 
-  private def rawValueExtract(paramName: String) = nameToValueMap.get(paramName).flatMap(Option.apply)
+  private def rawValueExtract(paramName: String) = nameToValueMap.get(paramName)
```
coutoPL, line 26: AFAIR, it is needed by the database lookup component. That's why the tests failed

docs/MigrationGuide.md
```diff
@@ -48,6 +48,7 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   * `ComponentId` was renamed to `DesignerWideComponentId`
   * new `ComponentId` is serialized in json to string in format `$componentType-$componentName` instead of separate fields (`name` and `type`)
   * `NodeComponentInfo.componentInfo` was renamed to `componentId`
+* [#5438](https://github.com/TouK/nussknacker/pull/5438) Removed sealed trait `CustomActionError`, now `CustomActionResult` is always used
```
coutoPL, line 51: what's this?

docs/MigrationGuide.md
```diff
@@ -111,6 +112,7 @@ To see the biggest differences please consult the [changelog](Changelog.md).
     * `SingleGenericContextTransformationWrapper` to `SingleInputDynamicComponentWrapper`
     * `SourceFactoryGenericContextTransformationWrapper` to `SourceFactoryDynamicComponentWrapper`
     * `JoinGenericContextTransformationWrapper` to `JoinDynamicComponentWrapper`
+  * type `NodeTransformationDefinition` (inside `DynamicComponent`) renamed to `ContextTransformationDefinition`
```
coutoPL, line 115: what's this?

coutoPL (changes requested):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
bartektartanus: But with this `Option` wrap, user can get error like this:
`Cannot find param name [foo]. Available param names: bar,foo,xyz`
when `foo` exists in `nameToValueMap` with `null` value.
Which is a little bit strange.

docs/MigrationGuide.md
bartektartanus: Missing migration guide entry, that will be helpful when upgrading Nu to 1.14

docs/MigrationGuide.md
bartektartanus: Missing migration guide entry, that will be helpful when upgrading Nu to 1.14

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
arkadius: SPeL is "null oriented", user can provide an expression `null`. So we have two optional things in this equation:
- parameter is missing - probably because someone import some outdated scenario version which wasn't migrated or developer didn't provide the migration for new changes in the component
- parameter is defined, and the expression has eagerly known, `null` value.

Currenlty we flatten both things which can generate confusions. If db lookup component base on this flattening, IMO we should fix it. It would be nice to provide a test that show both situations. We have a lot of combinations in  `ProcessTestHelpers`

mk-software-pl: done

coutoPL (approved):
LGTM

PR #5635, opened 2024-02-28
Scala List to Java List for result typing inside DecisionTable & Add test

philemone:
… 

## Checklist before merge

### DIFF
--- a/engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
+++ b/engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
@@ -21,6 +21,8 @@ object DecisionTable extends EagerService with SingleInputDynamicComponent[Servi
 
   override type State = Unit
 
+  private type DecisionTableResponse = java.util.List[java.util.Map[String, Any]]
+
   private object BasicDecisionTableParameter {
     val name = "Basic Decision Table"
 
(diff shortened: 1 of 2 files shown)


engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
```diff
@@ -21,6 +21,8 @@ object DecisionTable extends EagerService with SingleInputDynamicComponent[Servi
 
   override type State = Unit
 
+  private type DecisionTableResponse = java.util.List[java.util.Map[String, Any]]
```
coutoPL, line 24: maybe just `Output`?

engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
```diff
@@ -92,14 +94,14 @@ object DecisionTable extends EagerService with SingleInputDynamicComponent[Servi
         implicit ec: ExecutionContext,
         collector: InvocationCollectors.ServiceInvocationCollector,
         componentUseCase: ComponentUseCase
-    ): Future[Any] = Future {
+    ): Future[DecisionTableResponse] = Future {
```
coutoPL, line 97: there is no sense in changing type here

coutoPL (approved):
two small comments

engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
philemone: Hmm, I know that it is not propagated further but it forces checks on this type with whats inside Future, it can show that we've changes `filterRows` method to, let's say, Map[Int, Boolean] - with `Any` it will just quietly pass

engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
philemone: Good point, it will be more generic - changed

philemone:
Moved to https://github.com/TouK/nussknacker/pull/5638 (merge to release/1.14)

PR #5643, opened 2024-02-29
cherrypick: fetching/parsing batch periodic json only when needed (stop parsing during status check)

raphaelsolarski:
cherrypick: https://github.com/TouK/nussknacker/pull/5621


arkadius (approved):

arkadius (changes requested):
Isn't is already available on staging?

raphaelsolarski:
> Isn't is already available on staging?

Yes. After merging 1.14 into staging it is no longer needed.

PR #5647, opened 2024-02-29
Align deploymentservice actions

gskrobisz:
## Describe your changes
1. Extract common steps performed in `DeploymentServiceImpl` during deploy and cancel (validate input parameters, fetch scenario details, check if action is allowed, register new action, order command execution, handle command result)
2. ~~Removed `processingType` validation from `markActionExecutionFinished`~~

TODO:
4. ~~!! Bring back critical section: exclusive table lock.~~
5. ~~Extract some methods etc., make the code nice and tidy.~~

To be continued in separate PR:
6. Align those common steps with invokeCustomAction. This step needs to allow custom actions in action repository (replace current "action name" as ProcessActionType (fixed value enum) with ScenarioActionName (value class)).
7. Rename `OnDeployActionSuccess` and `OnDeployActionFailed`. Now we can have the same action-audit-log and action-response -handling for all actions, not only "Deploy". Aaaand this "Deploy" is misleading, it occurs also for "cancel" operations.
8. Enable async for all actions (see the comment for `DeploymentServiceImpl.deployProcessAsync`)

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
@@ -70,29 +70,56 @@ class DeploymentServiceImpl(
       processId: ProcessIdWithName,
       comment: Option[String]
   )(implicit user: LoggedUser, ec: ExecutionContext): Future[Unit] = {
-    val actionType = ProcessActionType.Cancel
 
-    validateDeploymentComment(comment).flatMap { comment =>
-      checkCanPerformActionAndAddInProgressAction[Unit](
-        processId,
-        actionType,
-        _.lastDeployedAction.map(_.processVersionId),
-        _ => None
-      ).flatMap { case (processDetails, actionId, versionOnWhichActionIsDone, buildInfoProcessIngType) =>
-        runDeploymentActionWithNotifications(
+    implicit val freshnessPolicy: DataFreshnessPolicy = DataFreshnessPolicy.Fresh
+    // TODO: use actionName (value class, can be user-defined) instead of actionType (enum, fixed values)
+    val actionType = ProcessActionType.Cancel
+    val actionName = ScenarioActionName(actionType)
+
+    for {
+      // 1. common for all commands/actions
+      // 1.1. validate input parameters
+      // TODO: extract param definitions to do generic "parameter validation"
+      _ <- validateDeploymentComment(comment)
+      // 1.2. fetch scenario data
+      processDetailsOpt <- dbioRunner.run(
+        processRepository.fetchLatestProcessDetailsForProcessId[CanonicalProcess](processId.id)
+      )
+      processDetails <- dbioRunner.run(existsOrFail(processDetailsOpt, ProcessNotFoundError(processId.name)))
+      // 1.3. calculate which scenario version is affected by the action: latest for deploy, deployed for cancel
+      versionOnWhichActionIsDone = processDetails.lastDeployedAction.map(_.processVersionId)
+      buildInfoProcessingType    = None
+      // 1.4. check if action is performed on proper scenario (not fragment, not archived)
+      _ = checkIfCanPerformActionOnScenario(actionName, processDetails)
+      // 1.5. check if action is allowed for
+      inProgressActionTypes <- dbioRunner.run(actionRepository.getInProgressActionTypes(processDetails.processId))
+      processState          <- dbioRunner.run(getProcessState(processDetails, inProgressActionTypes))
+      _ = checkIfCanPerformActionInState(actionName, processDetails, processState)
+      // 1.6. create new action, action is started with "in progress" state, the whole command execution can take some time
+      actionId <- dbioRunner.run(
+        actionRepository.addInProgressAction(
+          processDetails.processId,
           actionType,
-          actionId,
-          processId,
           versionOnWhichActionIsDone,
-          comment,
-          buildInfoProcessIngType
-        ) {
-          dispatcher
-            .deploymentManagerUnsafe(processDetails.processingType)
-            .processCommand(CancelScenarioCommand(processId.name, user.toManagerUser))
-        }.map(_ => ())
-      }
-    }
+          buildInfoProcessingType
+        )
+      )
+      // 2. command specific section
+      deploymentComment = comment.map(DeploymentComment.unsafe)
+      command           = CancelScenarioCommand(processId.name, user.toManagerUser)
+      // 3. again, common for all commands/actions
+      _ = runActionAndHandleResults(
+        actionType,
+        actionId,
+        versionOnWhichActionIsDone,
+        deploymentComment,
+        processDetails
+      ) {
+        dispatcher
+          .deploymentManagerUnsafe(processDetails.processingType)
+          .processCommand(command)
+      }.map(_ => ())
+    } yield ()
   }
 
   // Inner Future in result allows to wait for deployment finish, while outer handles validation
@@ -101,46 +128,63 @@ class DeploymentServiceImpl(
   // - deployment on engine side - it is longer part, the result will be shown as a notification
   // TODO: refactor this into smaller steps, currently "validators" in this method have a lot of side effects and the whole thing is pretty confusing
   override def deployProcessAsync(
-      processIdWithName: ProcessIdWithName,
+      processId: ProcessIdWithName,
       savepointPath: Option[String],
       comment: Option[String]
   )(implicit user: LoggedUser, ec: ExecutionContext): Future[Future[Option[ExternalDeploymentId]]] = {
-    val actionType = ProcessActionType.Deploy
-    validateDeploymentComment(comment).flatMap { validatedComment =>
-      checkCanPerformActionAndAddInProgressAction[CanonicalProcess](
-        processIdWithName,
-        actionType,
-        d => Some(d.processVersionId),
-        d => Some(d.processingType)
-      ).flatMap { case (processDetails, actionId, versionOnWhichActionIsDone, buildInfoProcessIngType) =>
-        validateBeforeDeploy(processDetails, actionId).transformWith {
-          case Failure(ex) =>
-            dbioRunner.runInTransaction(actionRepository.removeAction(actionId)).transform(_ => Failure(ex))
-          case Success(validationResult) =>
-            // we notify of deployment finish/fail only if initial validation succeeded
-            val deploymentFuture = runDeploymentActionWithNotifications(
-              actionType,
-              actionId,
-              processIdWithName,
-              versionOnWhichActionIsDone,
-              validatedComment,
-              buildInfoProcessIngType
-            ) {
-              dispatcher
-                .deploymentManagerUnsafe(processDetails.processingType)
-                .processCommand(
-                  RunDeploymentCommand(
-                    validationResult.processVersion,
-                    validationResult.deploymentData,
-                    validationResult.resolvedScenario,
-                    savepointPath
-                  )
-                )
-            }
-            Future.successful(deploymentFuture)
-        }
+
+    implicit val freshnessPolicy: DataFreshnessPolicy = DataFreshnessPolicy.Fresh
+    val actionType                                    = ProcessActionType.Deploy
+    val actionName                                    = ScenarioActionName(actionType)
+
+    for {
+      validatedComment <- validateDeploymentComment(comment)
+      processDetailsOpt <- dbioRunner.run(
+        processRepository.fetchLatestProcessDetailsForProcessId[CanonicalProcess](processId.id)
+      )
+      processDetails <- dbioRunner.run(existsOrFail(processDetailsOpt, ProcessNotFoundError(processId.name)))
+      versionOnWhichActionIsDone = Some(processDetails.processVersionId)
+      buildInfoProcessingType    = Some(processDetails.processingType)
+      _                          = checkIfCanPerformActionOnScenario(actionName, processDetails)
+      _                     <- dbioRunner.run(actionRepository.lockActionsTable)
+      inProgressActionTypes <- dbioRunner.run(actionRepository.getInProgressActionTypes(processDetails.processId))
+      processState          <- dbioRunner.run(getProcessState(processDetails, inProgressActionTypes))
+      _ = checkIfCanPerformActionInState(actionName, processDetails, processState)
+      actionId <- dbioRunner.run(
+        actionRepository.addInProgressAction(
+          processDetails.processId,
+          actionType,
+          versionOnWhichActionIsDone,
+          buildInfoProcessingType
+        )
+      )
+      // 2. command specific section
+      deployedScenarioData <- prepareDeployedScenarioData(processDetails, actionId)
+      command = RunDeploymentCommand(
+        deployedScenarioData.processVersion,
+        deployedScenarioData.deploymentData,
+        deployedScenarioData.resolvedScenario,
+        savepointPath
+      )
+      result <- validateBeforeDeploy(processDetails, deployedScenarioData).transformWith {
+        case Failure(ex) =>
+          dbioRunner.runInTransaction(actionRepository.removeAction(actionId)).transform(_ => Failure(ex))
+        case Success(_) =>
+          // we notify of deployment finish/fail only if initial validation succeeded
+          val deploymentFuture = runActionAndHandleResults(
+            actionType,
+            actionId,
+            versionOnWhichActionIsDone,
+            validatedComment,
+            processDetails
+          ) {
+            dispatcher
+              .deploymentManagerUnsafe(processDetails.processingType)
+              .processCommand(command)
+          }
+          Future.successful(deploymentFuture)
       }
-    }
+    } yield result
   }
 
   // TODO: this is temporary step: we want ParameterValidator here. The aim is to align deployment and custom actions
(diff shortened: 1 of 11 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
+      command = RunDeploymentCommand(
+        deployedScenarioData.processVersion,
+        deployedScenarioData.deploymentData,
+        deployedScenarioData.resolvedScenario,
+        savepointPath
+      )
+      // TODO: move validateBeforeDeploy before creating an action
+      result <- validateBeforeDeploy(ctx.processDetails, deployedScenarioData).transformWith {
+        case Failure(ex) =>
+          dbioRunner.runInTransaction(actionRepository.removeAction(ctx.actionId)).transform(_ => Failure(ex))
```
Diamekod0221, line 135: Perhaps we could extract some name for this for readability?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
+  /**
+    * Common validations and operations for a command execution.
+    * @return gathered data for further command execution
+    */
+  private def prepareCommandContextWithAction[PS: ScenarioShapeFetchStrategy](
+      processId: ProcessIdWithName,
+      actionType: ProcessActionType,
+      getVersionOnWhichActionIsDone: ScenarioWithDetailsEntity[PS] => Option[VersionId],
+      getBuildInfoProcessingType: ScenarioWithDetailsEntity[PS] => Option[ProcessingType]
+  )(implicit user: LoggedUser, ec: ExecutionContext): Future[CommandContext[PS]] = {
```
Diamekod0221, line 164: are you sure that the for comprehension in deploy and cancel will work properly if you use Future here? I'm concerned about the situation that we had before, where failing the future would still run side effects. I think some testing of this method would be a good idea?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
+          versionOnWhichActionIsDone,
+          buildInfoProcessingType
+        )
+      } yield CommandContext(processDetails, actionId, versionOnWhichActionIsDone, buildInfoProcessingType)
+    )
   }
 
+  // TODO: Use buildInfo explicitly instead of ProcessingType-that-is-used-to-calculate-buildInfo
+  private case class CommandContext[PS: ScenarioShapeFetchStrategy](
+      processDetails: ScenarioWithDetailsEntity[PS],
```
arkadius, line 197: I see a little problem here. `processDetails` can contain scenario shape for the latest version and  `versionOnWhichActionIsDone` is for some other version. Why we can't replace `processRepository.fetchLatestProcessDetailsForProcessId` with `fetchProcessDetailsForId`. Thanks to that we avoid confusion and we can remove `versionOnWhichActionIsDone` as it is already in `ScenarioWithDetailsEntity`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
+      actionId: ProcessActionId,
+      versionOnWhichActionIsDone: Option[VersionId],
+      buildInfoProcessingType: Option[ProcessingType]
```
arkadius, line 200: Why not just use `ScenarioWithDetailsEntity.processingType`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -240,73 +249,118 @@ class DeploymentServiceImpl(
...
+      ps: ProcessState
+  )(implicit user: LoggedUser): List[ScenarioActionName] = {
+    val defaultStateActions = ps.allowedActions.map(ScenarioActionName(_))
+    val customActions = dispatcher
+      .deploymentManagerUnsafe(processDetails.processingType)
+      .customActionsDefinitions
+      .collect {
+        case a if a.allowedStateStatusNames.contains(ps.status.name) => a.name
+      }
+      .distinct
```
arkadius, line 274: IMO we should return a `Set` to avoid that we'd have duplicates after the concatenation below

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -240,73 +249,118 @@ class DeploymentServiceImpl(
...
     }
   }
 
-  protected def prepareDeploymentData(user: User, deploymentId: DeploymentId): DeploymentData = {
-    DeploymentData(deploymentId, user, Map.empty)
+  private def allowedActionsForState[PS: ScenarioShapeFetchStrategy](
+      processDetails: ScenarioWithDetailsEntity[PS],
+      ps: ProcessState
+  )(implicit user: LoggedUser): List[ScenarioActionName] = {
+    val defaultStateActions = ps.allowedActions.map(ScenarioActionName(_))
```
arkadius, line 267: Why `default`? WDYT about `actionsDefinedInState` and `actionsDefinedInCustomActions`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
-      }
-    }
+      )
+      // 2. command specific section
+      deploymentComment = comment.map(DeploymentComment.unsafe)
+      command           = CancelScenarioCommand(processId.name, user.toManagerUser)
+      // 3. common for all commands/actions
+      _ = runActionAndHandleResults(
+        actionType,
+        ctx.actionId,
```
arkadius, line 91: You already have a nice context class, why do you pass all fields separately? :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -72,77 +72,134 @@ class DeploymentServiceImpl(
...
+      )
+      // 2. command specific section
+      deployedScenarioData <- prepareDeployedScenarioData(ctx.processDetails, ctx.actionId)
+      command = RunDeploymentCommand(
+        deployedScenarioData.processVersion,
+        deployedScenarioData.deploymentData,
+        deployedScenarioData.resolvedScenario,
+        savepointPath
+      )
+      // TODO: move validateBeforeDeploy before creating an action
```
arkadius, line 132: :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -240,73 +249,118 @@ class DeploymentServiceImpl(
...
-    val actionString = s"${actionType.toString.toLowerCase} (id: ${actionId.value}) of ${processIdWithName.name}"
+    val actionString = s"${actionType.toString.toLowerCase} (id: ${actionId.value}) of ${processDetails.name}"
     actionFuture.transformWith {
-      case Failure(failure) =>
-        logger.error(s"Action: $actionString finished with failure", failure)
+      case Failure(exception) =>
+        logger.error(s"Action: $actionString finished with failure", exception)
         val performedAt = clock.instant()
-        processChangeListener.handle(OnDeployActionFailed(processIdWithName.id, failure))
+        // TODO: rename to OnActionFailed
```
arkadius, line 318: :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -486,7 +540,7 @@ class DeploymentServiceImpl(
       .getOrElse(state)
   }
 
-  override def markActionExecutionFinished(processingType: ProcessingType, actionId: ProcessActionId)(
```
arkadius, line 489: Why do you remove processing type checking? Currently, one DeploymentManager that is configured in some 
 processing type can modify actions maintained by another DeploymentManager

docs/Changelog.md
```diff
@@ -40,14 +40,14 @@
   * Fixed the ability to download an unrelated attachment from a given scenario
 * [#5522](https://github.com/TouK/nussknacker/pull/5522), [#5519](https://github.com/TouK/nussknacker/pull/5519) Scenario status caching more often
 * [#5505](https://github.com/TouK/nussknacker/pull/5505) Fix: anonymous user handling regression
-* [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
-    * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
+* [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) [#5610](https://github.com/TouK/nussknacker/pull/5610) Improvement in DeploymentManager API: Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
```
arkadius, line 43: The target branch is staging so it should be rather in 1.15.0. Am I right?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: So far for `deploy` we have `ScenarioWithDetailsEntity.processingType`, for `cancel` we have `None`. I guess this is similar to the "VersionOnWhichActionIsDone", and that we want to keep track of buildInfo of deployed scenario. But... processingType does not change and we want to keep buildInfo for the point in time when action is invoked.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Fixed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Right now the only side effect is registered action and we have tests for that in DeploymentServiceSpec. What test scenario do you think of?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Fixed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Good point :) fixed.

docs/Changelog.md
gskrobisz: Fixed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Fixed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: It does not prevent modifying "not-my-action" when using "not-my-action-id" with "not-my-processing-type" either. 

That validation does not collide with action-to-customaction alignment, I excluded from this PR.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
arkadius: I think that we can try to break compatibility and store buildInfo / version for every action but we can do this later

docs/Changelog.md
arkadius: I can't see this "fix". It is sill in 1.14 section.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
arkadius: What about that?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
Diamekod0221: I think if the registered action is covered in other tests then it's ok.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: I like that idea to run action on a scenario in given version (last version to deploy or last deployed version to cancel). There is that one edge-case where "cancel" relies on missing version (scenario -> lastDeployedAction -> processVersionId):
```
            // Currently we don't send notifications and don't add finished action into db if version id is missing.
            // It happens during cancel when there is no finished deploy action before it, but some scenario is running on engine.
            // TODO: We should send notifications and add action db entry in that cases as well as for normal finish.
            //       Before we can do that we should check if we somewhere rely on fact that version is always defined -
            //       see ProcessAction.processVersionId
            logger.info(s"Action $actionString finished for action without version id - skipping listener notification")
            removeInvalidAction(ctx.actionId)
```
WDYT?

arkadius: Ah I forgot about that edge case. Correct me if I'm wrong. Do you propose to not allow to cancel scenario in such situation? What user can do then? Should he/she ask admin to kill such a scenario? Is it possible to make this field somehow optional and handle also this case?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -66,83 +66,164 @@ class DeploymentServiceImpl(
...
-                    validationResult.processVersion,
-                    validationResult.deploymentData,
-                    validationResult.resolvedScenario,
-                    savepointPath
-                  )
-                )
-            }
-            Future.successful(deploymentFuture)
-        }
+        getCurrentVersion,
```
arkadius, line 124: Please inline this function. It is not obvious which version is current as we use latest in most places in the code.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
```diff
@@ -66,83 +66,164 @@ class DeploymentServiceImpl(
...
+  }
+
+  /**
+    * Common validations and operations for a command execution.
+    * @return gathered data for further command execution
+    */
+  private def prepareCommandContextWithAction[PS: ScenarioShapeFetchStrategy](
+      processIdWithName: ProcessIdWithName,
+      actionType: ProcessActionType,
+      getScenarioWithVersionOnWhichActionIsDone: ScenarioWithDetailsEntity[PS] => DB[
```
arkadius, line 162: It is important to name somehow what is passed in this function maybe we should rename this to `extract...FromLatestScenarioDetails` ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
arkadius: 2. Also please name it more explicit - `versionOnWhichActionIsDone` part was pretty good, we can adapt it to this situation

arkadius: Please name it latestScenarioDetails to make it clear that it is latest

arkadius (approved):
one minor naming comment left

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceImpl.scala
gskrobisz: Done, renamed.

PR #5712, opened 2024-03-11
Fixed NPE when ForEachTransformer is invoked with null elements parmeter

gskrobisz:
## Describe your changes

Before:
When ForEachTransformer is invoked with `elements = null` null pointer exception is raised (transient exception, it is not handled via ExceptionHandler/ExceptionConsumer, leads to scenario failure and restart).

After:
When `elements = null` the component does not emit any data, just like when elements is an empty list.

Alternative solution:
When `elements = null` the component raises non-transient exception to make clear to the user that something erroneous happend (via EspExceptionConsumer) without restarting the whole scenario. WDYT?

## Checklist before merge

### DIFF
--- a/engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
+++ b/engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
@@ -27,11 +27,13 @@ object ForEachTransformer extends CustomStreamTransformer with Serializable {
           .flatMap(ctx.lazyParameterHelper.lazyMapFunction(elements))
           .flatMap(
             (valueWithContext: ValueWithContext[util.Collection[AnyRef]], c: Collector[ValueWithContext[AnyRef]]) => {
-              valueWithContext.value.asScala.zipWithIndex
-                .map { case (partToRun, index) =>
-                  new ValueWithContext[AnyRef](partToRun, valueWithContext.context.appendIdSuffix(index.toString))
-                }
-                .foreach(c.collect)
+              Option(valueWithContext.value).foreach(
+                _.asScala.zipWithIndex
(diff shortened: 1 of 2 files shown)


engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/ForEachTransformer.scala
```diff
@@ -27,11 +27,13 @@ object ForEachTransformer extends CustomStreamTransformer with Serializable {
           .flatMap(ctx.lazyParameterHelper.lazyMapFunction(elements))
           .flatMap(
             (valueWithContext: ValueWithContext[util.Collection[AnyRef]], c: Collector[ValueWithContext[AnyRef]]) => {
```
mslabek, line 29: I don't know the topic fully, but `valueWithContext: ValueWithContext[util.Collection[AnyRef]]` declaration makes me think that there should be some Collection there. 

I think there's 2 types of situations where a null can end up here:
1. Someone enters a self-contained expression that evaluates to null
2. Someone enters a reference to a list coming from external system - like a list field from a kafka source - and that evaluates to null

For the 1. situation I think we should tackle this at parameter level. I don't see why it would be useful to have a foreach that doesn't do anything. So I would propose adding a NotNull validation to it.

For the 2. situation I think this we should signal this as exception. I think that if configurator expects a null in list field he should do a null check before? And if he omitted that I feel like administrator would like to know about that and communicate it to the configurator. Although I don't have practical experience in these roles so I'm just guessing. WDYT?

Also, do we have another way of handling nulls in our code at component implementation level?

lciolecki: Remember that: `util.Collection[AnyRef]` is java, so null is posible here :)

arkadius:
Closing due to inactivity

PR #5724, opened 2024-03-13
Improvements: Allow Designer to run locally with remote debugging & normalize NUSSKNACKER_SCALA_VERSION env param name

lciolecki:
- passing -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address at run.sh
- replace SCALA_VERSION by NUSSKNACKER_SCALA_VERSION

## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -7,6 +7,7 @@
   * openapi-circe-yaml: 0.6.0 -> 0.7.4 
 * [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
     * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
+* [#5724](https://github.com/TouK/nussknacker/pull/5724) Improvements: Allow Designer to run locally with remote debugging & normalize NUSSKNACKER_SCALA_VERSION env param name
 
 1.14.0 (21 Mar 2024)
 -------------------------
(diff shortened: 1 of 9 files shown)


docs/Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+1.15.0 (Not released yet)
+* [#5724](https://github.com/TouK/nussknacker/pull/5724) Improvements: Allow Designer to run locally with remote debugging & normalize SCALA_VERSION env param name
```
arkadius, line 4: I think that we shouldn't expose this in changelog. It is interesting for developers instead of users. Instead, we should mention about that in contributing guide

docs/MigrationGuide.md
```diff
@@ -26,6 +26,9 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 ### Other changes
 
 * [#5574](https://github.com/TouK/nussknacker/pull/5574) Removed the support for the plugable expression languages: `ExpressionConfig.languages` removed
+* [#5724](https://github.com/TouK/nussknacker/pull/5724) Improvements: Run Designer locally
+  * Introduce `JAVA_DEBUG_PORT` to run the Designer locally with remote debugging capability
+  * Removed `NUSSKNACKER_SCALA_VERSION` from running Designer locally, please use `SCALA_VERSION` instead
```
arkadius, line 31: Why did you change it from `NUSSKNACKER_SCALA_VERSION` to `SCALA_VERSION`? Thanks to nussknacker prefix, it was possible to set up this on the system level (in `.profile` file). I used it quite often when for some reasons I had to switch to 2.12 for a longer period.

lciolecki: Hmm, ok it has sense.. But right now we use SCALA_VERSION on run scripts like ./runServer / CI scripts and NUSSKNACKER_SACLA_VERSION on sbt.. IMO, we should normalize it and use one convention everywhere. We can stay with NUSSKNACKER_SCALA_VERSION, WDYT?

arkadius (approved):

PR #5759, opened 2024-03-19
Explicit defaults for scenario properties

gskrobisz:
## Describe your changes

AS IS:
Default values for scenario properties are introduced in several places: 
1. in `DeploymentManagerProvider.scenarioPropertiesConfig` (e.g. FlinkStreamingPropertiesConfig.properties)
2. via configuration `scenarioPropertiesConfig`
3. in `DeploymentManagerProvider.metaDataInitializer` there are some static overwrites (e.g. `overridingProperties = Map(StreamMetaData.parallelismName -> "1", StreamMetaData.spillStateToDiskName -> "true")`)
4. in TypeSpecificData (e.g. `StreamMetaData.spillStateToDisk = Some(true)` which by accident corresponds to metaDataInitializer)
5. in some places in code (e.g. `DefaultAsyncInterpretationValueDeterminer.DefaultValue` and `ProcessConfigCreator.asyncExecutionContextPreparer`)

Implicit defaults lead to 3-state logic of "IO mode" and "Spill state to disk", true / false / server default, where server default is burried somewhere in those places above.
Some scenario properties are used to initialize scenario metaData (with typeSpecificData).
MetaData control the behaviour of scenario.

Comments:
1. MetaDataInitializer: "TODO: set the defaults in one place without overriding"
2. DefaultAsyncInterpretationValueDeterminer: // ... in the future this default will be changed to true

Purpose of the change:
Introduce explicit default values for scenario properties:
- to avoid confusing 3-state logic
- user knows what is the configuration of scenario
- the explicit configuration is imprinted into scenario version

TO BE:
**1. Remove "server default" option from "IO mode" and "Spill state to disk". It seems to be redundant to other two options.
2. Migrate all implicit properties (stored as empty string) to explicit values.**
3. Remove `DefaultAsyncInterpretationValueDeterminer`, it duplicates MetaDataInitializer.create
4. Move `asyncExecutionConfig.defaultUseAsyncInterpretation` to scenarioPropertiesConfig
5. Remove `ProcessConfigCreator.asyncExecutionContextPreparer`
6. `TypeSpecificData` keeps property names and default values.
7. `DeploymentManagerProvider.scenarioPropertiesConfig` defines explicit static default values.
8. `ProcessingTypeData.createDeploymentData` combines DeploymentManagerProvider.scenarioPropertiesConfig and config to build DeploymentData (to display in UI and use inside deployment)
9. `MetaDataInitializer.overrideDefaultProperties` should not override, it is more like "evaluate some initial MD values that cannot be statically defined (like slug in request-response)"

## Checklist before merge

### DIFF
--- a/common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
+++ b/common-api/src/main/scala/pl/touk/nussknacker/engine/api/TypeSpecificData.scala
@@ -44,10 +44,10 @@ object FragmentSpecificData {
 
 // TODO: rename to FlinkStreamMetaData
 case class StreamMetaData(
-    parallelism: Option[Int] = None,
+    parallelism: Option[Int] = Some(1),
     // we assume it's safer to spill state to disk and fix performance than to fix heap problems...
     spillStateToDisk: Option[Boolean] = Some(true),
-    useAsyncInterpretation: Option[Boolean] = None,
+    useAsyncInterpretation: Option[Boolean] = Some(false),
     checkpointIntervalInSeconds: Option[Long] = None
 ) extends ScenarioSpecificData {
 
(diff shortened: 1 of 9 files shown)


arkadius:
Closing due to inactivity

PR #5779, opened 2024-03-21
[NU-1588] Add API versioning for migrations between environments

DeamonDev:
## Describe your changes

The changes are described in the associated issue.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -205,15 +222,48 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
-    invokeForSuccess(
-      HttpMethods.POST,
-      List("migrate"),
-      Query.Empty,
-      HttpEntity(migrateScenarioRequest.asJson.noSpaces),
-      List.empty
-    )
+
+    val f: EitherT[Future, NuDesignerError, Unit] = for {
+      remoteNuVersion <- EitherT(fetchRemoteNuVersion)
```
mk-software-pl, line 227: why do we have to do it? We are in the Source Nussknacker. We want to migrate the scenario to the Target Nussknacker. I thought we would be sending two things `(source_nu_version, scenario_data_in_source_nu_version_format)`.  At the target Nussknacker side, we will be handling it in the following way:
1. check version of the data
2. decide what parser (or adapter, whatever) to user
3. convert the source format to the current, target format

Why do we have to do it in the two steps like here?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -205,15 +222,48 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
+        versionComparisonResult
+      )
+
+      versionTag = decidedMigrateScenarioRequest match {
+        case _: MigrateScenarioRequestV1_14 => "V1_14"
+        case _: MigrateScenarioRequestV1_15 => "V1_15"
+      }
+
+      res <- EitherT(
+        invokeForSuccess(
```
mk-software-pl, line 245: we have a Tapir endpoint, so we should use it here in the Tapir's way (https://tapir.softwaremill.com/en/latest/client/sttp.html)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -205,15 +222,48 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
+          HttpMethods.POST,
+          List("migrate"),
+          Query.Empty,
+          HttpEntity(decidedMigrateScenarioRequest.asJson.noSpaces),
+          List(
+            RawHeader("X-MigrateDtoVersion", versionTag)
```
mk-software-pl, line 251: could you please explain why we need it?

mk-software-pl (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/AppApiHttpService.scala
```diff
@@ -114,6 +114,13 @@ class AppApiHttpService(
       }
   }
 
+  expose {
+    appApiEndpoints.nuVersionEndpoint
```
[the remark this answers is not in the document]
mk-software-pl: let's talk about [this](https://github.com/TouK/nussknacker/pull/5779#discussion_r1551762339) first

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
```diff
@@ -108,9 +114,30 @@ object TapirCodecs {
...
+
+    implicit val migrateScenarioRequestEncoder: Encoder[MigrateScenarioRequest] = Encoder.instance {
+      case v14: MigrateScenarioRequestV1_14 => v14.asJson
+      case v15: MigrateScenarioRequestV1_15 => v15.asJson
+    }
+
+    implicit val migrateScenarioRequestDecoder: Decoder[MigrateScenarioRequest] = List[Decoder[MigrateScenarioRequest]](
+      Decoder[MigrateScenarioRequestV1_14].widen,
+      Decoder[MigrateScenarioRequestV1_15].widen,
+    ).reduceLeft(_ or _)
```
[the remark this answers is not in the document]
DeamonDev: I followed [official](https://circe.github.io/circe/codecs/adt.html#adts-encoding-and-decoding) approach. This has the advantage of not being dependent of some dynamic values.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -142,16 +141,34 @@ object MigrationApiEndpoints {
...
         engineSetupName: EngineSetupName,
         processCategory: String,
         scenarioGraph: ScenarioGraph,
         processName: ProcessName,
         isFragment: Boolean,
-    )
+    ) extends MigrateScenarioRequest
+
+    @derive(encoder, decoder)
+    final case class MigrateScenarioRequestV1_15(
```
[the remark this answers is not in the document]
DeamonDev: No, but I wanted to show how this process works.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -44,7 +47,7 @@ class MigrationService(
   private val passUsernameInMigration = true
 
   def migrate(
-      migrateScenarioRequest: MigrateScenarioRequest
+      migrateScenarioRequest: CurrentScenarioMigrateRequest
```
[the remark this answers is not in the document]
DeamonDev: Why?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: No, the version check has to be made on the source side. What if source version is greater than the target one? In my approach source then lowers its version as much as possible.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: Its only relevant for testing purposes, see `StandardRemoteEnvironmentSpec`. Thanks to that I am able to test the version sent was proper one.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
mk-software-pl: because it's a Data Transport Object. Belongs to the transport layer by definition. It's simply not clean approach.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
mk-software-pl: This approach is one of several. the one in which you try to fit it by the structure is highly error-prone and not so straightforward.

IMO it's not request metadata. As I was saying - this is the feature that allows to migrate some version of the scenario to another version of the scenario. It's a domain logic! The version is a domain thing here. I don't think it should be passed with the header.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
mk-software-pl: The intention is not clear. 
So what business value from the user perspective we're going to deliver with this PR?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -34,7 +33,7 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
         jsonBody[MigrateScenarioRequest].example(
           Example.of(
             summary = Some("example of migration request between environments"),
-            value = MigrateScenarioRequest(
+            value = MigrateScenarioRequestV1_15(
```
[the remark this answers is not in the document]
mk-software-pl: do you use the endpoint already on production?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mk-software-pl: why mocking version endpoint is a problem?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mk-software-pl: Ah, the requirement about migration from a higher version to a lower one. It's a pity we don't have the requirements listed :) 

Why the new endpoint, not the existing one?

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationApiAdapterService.scala
```diff
@@ -0,0 +1,114 @@
...
+    val versionPattern = """(\d+)\.(\d+)\.(\d+).*""".r
+
+    val localResult: Try[Regex.Match]  = Try(versionPattern.findFirstMatchIn(localNuVersion).get)
+    val remoteResult: Try[Regex.Match] = Try(versionPattern.findFirstMatchIn(remoteNuVersion).get)
+
+    localResult match {
+      case Success(localVersionPatternMatch) =>
+        remoteResult match {
+          case Success(remoteVersionPatternMatch) =>
+            val localMajorVersion = localVersionPatternMatch.group(1).toInt
```
mk-software-pl, line 91: let's use https://github.com/kevin-lee/just-semver to not struggle with the version

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
DeamonDev: But we are not talking about version of the scenario. We are talking about Nussknacker version which is irrelevant for the targget Nussknacker, since he always lifts up DTO to his highest version.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
DeamonDev: The business value is that future generations of Nu developers will be able to version the request and they will be able to mitigate adding new version based on this duplicate.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
DeamonDev: Yes

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: I dont get the question :(

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: It's not a problem. I just wanted to test more. But it is not so relevant. I can change that one.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mk-software-pl: sorry, I was referring to the new version endpoint. We already have an endpoint which returns a Nu version (and some other things)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
mk-software-pl: I just realized that we focus on the Nu version. Maybe we should also consider the approach when we focus on the ScenarioData (graph + the other things from the migration DTO) version. Let's say we have V1 now (in Nu 1.14). In Nu 1.15 there are no changes in the ScenarioData, so there is still V1 (so Nu 1.14 and 1.15 works with the V1). In Nu 1.16 there is some change, so we introduce V2. 

Won't it be simpler?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: Yes, but this endpoint has some legacy quirks and gave me weird responses. Are u Ok I'll go with my Version only and comment it should be removed when that more general one stabilize?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TapirCodecs.scala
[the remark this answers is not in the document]
DeamonDev: Yeah, but OTOH when 1.15 changes nothing then we not introduce anything new. And with current approach we get explicit breaking versions history.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mk-software-pl: what quirks exactly?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
DeamonDev: Yeah, but OTOH it is Akka-specific file. So I would rewrite it to all the way Tapir while rewriting `RemoteEnvironmentResources` into Tapir.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -44,7 +46,7 @@ class MigrationService(
   private val passUsernameInMigration = true
 
   def migrate(
-      migrateScenarioRequest: MigrateScenarioRequest
+      migrateScenarioRequest: CurrentMigrateScenarioRequest
```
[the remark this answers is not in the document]
DeamonDev: Good idea. I also wondered about that but the reason I stayed with this was we really are talking about request :D But if we could go with Description I would be even more happy :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -118,49 +167,58 @@ class MigrationService(
...
+          uiProcessResolverO match {
+            case Some(uiProcessResolver) =>
+              Future.successful[Either[NuDesignerError, ValidationResults.ValidationResult]](
+                FatalValidationError.renderNotAllowedAsError(
+                  uiProcessResolver.validateBeforeUiResolving(scenarioGraph, processName, isFragment)
+                )
+              )
+            case None =>
+              Future.failed(
+                new IllegalStateException(
```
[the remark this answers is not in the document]
DeamonDev: Yeah, for `Either`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -50,6 +50,16 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
       .errorOut(nuDesignerErrorOutput)
       .withSecurity(auth)
 
+  lazy val apiVersionEndpoint: SecuredEndpoint[Unit, NuDesignerError, Int, Any] =
+    baseNuApiEndpoint
+      .summary("current version of the migrations between environments API")
+      .tag("Migrations")
+      .get
+      .in("migrate" / "apiVersion")
```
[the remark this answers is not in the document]
mk-software-pl: there is one more problem with this ... we use the verb "migrate", but it'd be better to use the noun "migration". But what to do with the existing "/api/migrate" endpoint?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -50,6 +50,16 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
+      .out(plainBody[Int])
+      .errorOut(nuDesignerErrorOutput)
```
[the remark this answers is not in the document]
mk-software-pl: please take a look:
Screenshot 2024-04-16 at 12 27 41

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -118,49 +167,58 @@ class MigrationService(
...
+  }
 
-    transformedFuture
+  private def validateProcessingTypeAndUIProcessResolver(
+      scenarioGraph: ScenarioGraph,
+      processName: ProcessName,
+      isFragment: Boolean,
+      processingType: ProcessingType
+  )(implicit loggedUser: LoggedUser) = {
+    EitherT[Future, NuDesignerError, ValidationResults.ValidationResult](
```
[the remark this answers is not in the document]
mk-software-pl: I see the code looks like this now:

```
    EitherT[Future, NuDesignerError, ValidationResults.ValidationResult](
      processResolver
        .forProcessingTypeEUnsafe(processingType) match {
        case Left(e) => Future.successful[Either[NuDesignerError, ValidationResults.ValidationResult]](Left(e))
        case Right(uiProcessResolver) =>
          Future.successful[Either[NuDesignerError, ValidationResults.ValidationResult]](
            FatalValidationError.renderNotAllowedAsError(
              uiProcessResolver.validateBeforeUiResolving(scenarioGraph, processName, isFragment)
            )
          )
      }
    )
```

Why you don't want to use `fromEither` here? IMO `Future.successful[Either[NuDesignerError, ValidationResults.ValidationResult]]` usages are a boilerplate here.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -38,76 +47,162 @@ class MigrationService(
...
+    val versionsDifference               = localScenarioDescriptionVersion - remoteScenarioDescriptionVersion
+
+    val liftedMigrateScenarioRequestE: Either[ApiAdapterServiceError, MigrateScenarioData] =
+      if (versionsDifference > 0) {
+        migrationApiAdapterService.adaptUp(migrateScenarioData, versionsDifference)
+      } else Right(migrateScenarioData)
+
+    liftedMigrateScenarioRequestE match {
+      case Left(apiAdapterServiceError) =>
+        EitherT(
```
mk-software-pl, line 71: `EitherT.leftT(MigrationApiAdapterError(apiAdapterServiceError))`

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -38,76 +47,162 @@ class MigrationService(
...
+          Future[Either[NuDesignerError, Unit]](
+            Left(
+              MigrationApiAdapterError(apiAdapterServiceError)
+            )
+          )
+        )
+      case Right(liftedMigrateScenarioRequest) =>
+        liftedMigrateScenarioRequest match {
+          case currentMigrateScenarioRequest: CurrentMigrateScenarioData =>
```
mk-software-pl, line 80: let's have it flattened with the outer pattern matching

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -38,76 +47,162 @@ class MigrationService(
...
+    useLegacyCreateScenarioApi: Boolean,
+    migrationApiAdapterService: MigrationApiAdapterService
 )(implicit val ec: ExecutionContext) {
 
   private val passUsernameInMigration = true
 
   def migrate(
-      migrateScenarioRequest: MigrateScenarioRequest
+      migrateScenarioData: MigrateScenarioData
+  )(implicit loggedUser: LoggedUser): EitherT[Future, NuDesignerError, Unit] = {
```
mk-software-pl, line 58: IMO we should not use here `NuDesignerError` but the specific for the MigrationService errors hierarchy

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
mk-software-pl: I see FIXME. What stops us from changing it now?

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -788,6 +790,40 @@ paths:
...
+      - Migrations
+      summary: current version of the scenario description version being used
+      operationId: getApiMigrateScenarioDescriptionVersion
+      responses:
+        '200':
+          description: ''
+          content:
+            text/plain:
+              schema:
+                type: integer
```
mk-software-pl, line 805: maybe we should move the integer to JSON object? It'd be easier to extend it in a backward compatible way in the future (and costs almost nothing)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
 object MigrationApiEndpoints {
 
   object Dtos {
+    sealed trait MigrationError
```
coutoPL, line 118: maybe it'd be better to move it to the `MigrationService` companion object? According to the rule which says the infrastructure layer should have no access to the core/domain layer

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
+
+    object MigrationError {
```
coutoPL, line 120: I'd propose more human names for the errors:
* `MigrationError.InvalidScenario`
* `MigrationError.CannotMigrateArchivedScenario`
* `MigrationError.InsufficientPermission(missingPermission: Permission)`

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
+              errors.processPropertiesErrors.map(_.message) ++ errors.invalidNodes.map { case (node, nerror) =>
+                s"$node - ${nerror.map(_.message).mkString(", ")}"
+              }
+            s"Cannot migrate, following errors occurred: ${messages.mkString(", ")}"
+          })
+        )
+
+      final case class RemoteEnvironmentCommunicationError(message: String) extends MigrationError
+
+      final case class Unknown(message: String) extends MigrationError
```
coutoPL, line 141: I don't think we need it. It should be simply converted to HTTP 500. So, in the place where we instantiate the Unknown, there should be an exception thrown.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -195,25 +216,73 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
+          engineSetupName,
+          processCategory,
+          scenarioGraph,
+          processName,
+          isFragment
+        )
+      versionsDifference = localScenarioDescriptionVersion - remoteScenarioDescriptionVersion
+      transformedMigrateScenarioRequestE =
+        if (versionsDifference > 0)
+          migrationApiAdapterService.adaptDown(migrateScenarioRequest, versionsDifference)
```
coutoPL, line 237: WDYT about changing the interface a little bit? IMO we could look like this:
```
def adaptTo(data: D, targetDataVersion: Int): Either[ApiAdapterServiceError, D] =
```

It seems we can encapsulate more adapting logic inside the adapter

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -195,25 +216,73 @@ trait StandardRemoteEnvironment extends FailFastCirceSupport with RemoteEnvironm
...
-      HttpMethods.POST,
-      List("migrate"),
-      Query.Empty,
-      HttpEntity(migrateScenarioRequest.asJson.noSpaces),
-      List.empty
     )
   }
 
+  private def fetchRemoteMigrationScenarioDescriptionVersionAux(implicit ec: ExecutionContext): Future[ApiVersion] = {
+    invoke[ApiVersion](
```
coutoPL, line 275: let's leave a todo, that we could use a client that is generated from Tapir endpoint

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -25,7 +25,7 @@ import pl.touk.nussknacker.engine.util.multiplicity.{Empty, Many, Multiplicity,
 import pl.touk.nussknacker.engine.{ConfigWithUnresolvedVersion, DeploymentManagerDependencies, ModelDependencies}
 import pl.touk.nussknacker.processCounts.influxdb.InfluxCountsReporterCreator
 import pl.touk.nussknacker.processCounts.{CountsReporter, CountsReporterCreator}
-import pl.touk.nussknacker.ui.api._
+import pl.touk.nussknacker.ui.api.{MigrationApiHttpService, _}
```
coutoPL, line 28: `import pl.touk.nussknacker.ui.api._`

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ApiAdapter.scala
```diff
@@ -0,0 +1,6 @@
+package pl.touk.nussknacker.ui.util
+
+trait ApiAdapter[D] {
```
coutoPL, line 3: I wonder if we shouldn't have here 3 generic params:
* previous version
* current version
* next version

Obviously, we cannot assume there will be always possible to lift or downgrade the version so each method should rather have declarations like that:

```scala
def liftVersion: CURRENT_VERSION => Option[NEXT_VERSION]
def downgradeVersion: CURRENT_VERSION => Option[PREV_VERSION]
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ApiVersion.scala
```diff
@@ -0,0 +1,7 @@
+package pl.touk.nussknacker.ui.util
+
+import derevo.circe._
+import derevo.derive
+
+@derive(encoder, decoder)
+final case class ApiVersion(value: Int)
```
coutoPL, line 7: * it's used in `MigrationApiEndpoints` so it should be defined there
* should we add `@derive(schema)` as well?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/VersionedData.scala
```diff
@@ -0,0 +1,5 @@
+package pl.touk.nussknacker.ui.util
+
+trait VersionedData {
+  def currentVersion(): Int
```
coutoPL, line 4: maybe without `()`? it could be easily overridden as `val`

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/ReflectionBasedUtils.scala
```diff
@@ -0,0 +1,23 @@
+package pl.touk.nussknacker.test.utils.domain
+
+import org.reflections.Reflections
+import org.reflections.util.ConfigurationBuilder
+
+import scala.jdk.CollectionConverters._
+
+object ReflectionBasedUtils {
+
+  def findSubclassesOf[T](clazz: Class[T], packageName: String): List[Class[_ <: T]] = {
```
coutoPL, line 10: WDYT about improving it a little bit?
```scala
def findSubclassesOf[T: ClassTag](packageName: String): List[Class[_ <: T]] = {
    val baseClass = classTag[T].runtimeClass.asInstanceOf[Class[T]]
    val reflections = new Reflections(
      new ConfigurationBuilder().forPackages(baseClass.getPackageName, packageName)
    )

    reflections
      .getSubTypesOf(baseClass)
      .asScala
      .toList
      .sortBy(_.getName)
  }
```

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/ReflectionBasedUtils.scala
```diff
@@ -0,0 +1,23 @@
+package pl.touk.nussknacker.test.utils.domain
+
+import org.reflections.Reflections
+import org.reflections.util.ConfigurationBuilder
+
+import scala.jdk.CollectionConverters._
+
+object ReflectionBasedUtils {
```
coutoPL, line 8: It would be nice if you could use it in `NuDesignerApiAvailableToExpose#findApiEndpointsClasses` too

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpointsSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.scalatest.freespec.AnyFreeSpecLike
+import org.scalatest.matchers.should.Matchers
+import pl.touk.nussknacker.test.utils.domain.ReflectionBasedUtils.findSubclassesOf
+import pl.touk.nussknacker.test.utils.domain.migration.MigrationUtils.allSubclassesContainVersionField
+import pl.touk.nussknacker.ui.api.description.MigrationApiEndpoints.Dtos.MigrateScenarioRequestDto
+
+class MigrationApiEndpointsSpec extends AnyFreeSpecLike with Matchers {
+
+  "all subclasses of MigrateScenarioRequestDto" - {
+    "contain version field of type Int" in {
```
coutoPL, line 12: hmm, what if we make a slight change:
```scala
sealed trait MigrateScenarioRequestDto {
  def version: Int
}
```

It looks like after doing this, the test won't be needed

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/StandardRemoteEnvironmentSpec.scala
```diff
@@ -185,56 +253,49 @@ class StandardRemoteEnvironmentSpec
...
-        case AddProcess(isFragment) =>
-          remoteProcessList = expectedProcessDetails.name :: remoteProcessList
-          triedToAddProcess = true
-          addedFragment = Some(isFragment)
-
-          Marshal(ProcessTestData.validScenarioDetailsForMigrations).to[RequestEntity].map { entity =>
-            HttpResponse(OK, entity = entity)
+        case Migrate() =>
+          /*
+            FIXME: Currently, MigrateScenarioRequestDtoV1 and MigrateScenarioRequestDtoV2 are "the same" and they serve
```
coutoPL, line 272: TBH we should consider dropping the MigrateScenarioRequestDtoV2 (as a last phase of the review changes). Or just comment it out and add a proper note. 

It's because V1 and V2 are the same. At the moment we recommend changing the shape of V2 when the first inconsistency between scenario versions takes place. But from the API perspective, it would be a breaking change. So, maybe it's better to focus on the fact that currently we have only one version and make sure (by making use of comment nodes) that in the future, another developer will be able to add the new version.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceBusinessSpec.scala
```diff
@@ -27,6 +29,23 @@ class MigrationApiHttpServiceBusinessSpec
...
+    "return current api version" in {
+      given()
+        .when()
+        .basicAuthAllPermUser()
+        .get(s"$nuDesignerHttpAddress/api/migration/scenario/description/version")
+        .Then()
+        .statusCode(200)
+        .body(
+          "value",
+          equalTo[Int](MigrationApiAdapters.adapters.keySet.size + 1)
```
coutoPL, line 44: I don't see anything wrong with hardcoding the version here.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
 object MigrationApiEndpoints {
 
   object Dtos {
+    sealed trait MigrationError
+
+    object MigrationError {
+      final case class Validation(errors: ValidationErrors)                                extends MigrationError
+      final case class MigrationToArchived(processName: ProcessName, environment: String)  extends MigrationError
+      final case class Unauthorized(user: LoggedUser)                                      extends MigrationError
+      final case class MigrationApiAdapter(apiAdapterServiceError: ApiAdapterServiceError) extends MigrationError
```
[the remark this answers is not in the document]
DeamonDev: No business explanation: it may be replaced with throwing IllegalStateException.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
+
+            val messages = errors.globalErrors.map(_.error.message) ++
+              errors.processPropertiesErrors.map(_.message) ++ errors.invalidNodes.map { case (node, nerror) =>
+                s"$node - ${nerror.map(_.message).mkString(", ")}"
+              }
+            s"Cannot migrate, following errors occurred: ${messages.mkString(", ")}"
+          })
+        )
+
+      final case class RemoteEnvironmentCommunicationError(message: String) extends MigrationError
```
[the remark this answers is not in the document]
DeamonDev: No business explanation: it may be replaced with throwing IllegalStateException.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -98,6 +115,88 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
   object Dtos {
+    sealed trait MigrationError
+
+    object MigrationError {
+      final case class Validation(errors: ValidationErrors)                                extends MigrationError
+      final case class MigrationToArchived(processName: ProcessName, environment: String)  extends MigrationError
+      final case class Unauthorized(user: LoggedUser)                                      extends MigrationError
+      final case class MigrationApiAdapter(apiAdapterServiceError: ApiAdapterServiceError) extends MigrationError
+
+      implicit val validationErrorCodec: Codec[String, Validation, CodecFormat.TextPlain] =
```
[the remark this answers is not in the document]
DeamonDev: Show is not a covariant functor.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
coutoPL: ok, so it's better to throw IllegalStateException as soon as we detect the illegal state

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ApiAdapter.scala
DeamonDev: Yeah typesafety would be great. But note that composition of such adapters becomes more problematic. And it is easy to end up in the world full of `Foo[D, Any, Any, Any, Any]`. I'll mark it for the future improvement.

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/ReflectionBasedUtils.scala
DeamonDev: Done :)

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/StandardRemoteEnvironmentSpec.scala
DeamonDev: I would change the comment to encourage the user to add `V3`. The reasons is having comments is not so restrictive and might lead to catastrophic bugs.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
DeamonDev: I tried to follow analogous tapir services. I'll leave TODO for someone to think about it :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
DeamonDev: Ok, done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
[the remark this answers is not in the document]
mk-software-pl: please take a look at my improvement suggestions: 
https://github.com/TouK/nussknacker/pull/5975

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
mk-software-pl: please take a look at my improvement suggestions: 
https://github.com/TouK/nussknacker/pull/5975

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mk-software-pl: please take a look at my improvement suggestions: 
https://github.com/TouK/nussknacker/pull/5975

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ApiVersion.scala
mk-software-pl: please take a look at my improvement suggestions: 
https://github.com/TouK/nussknacker/pull/5975

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpointsSpec.scala
mk-software-pl: please take a look at my improvement suggestions: 
https://github.com/TouK/nussknacker/pull/5975

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/StandardRemoteEnvironmentSpec.scala
mk-software-pl: But do we need to have `MigrateScenarioRequestDtoV2` at the moment? If it's only for the example purpose, maybe we can just comment out all the code related to V2 and in all these places we could leave a note that it can be uncommented and adopted. WDYT?

mk-software-pl (commented):
1. please take a look at my suggestions in the draft branch
2. IMO [this](https://github.com/TouK/nussknacker/pull/5779#discussion_r1580849175) is the most important thing to solve before merging this PR

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/StandardRemoteEnvironmentSpec.scala
DeamonDev: Okay, that would be good compromise

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/migrate/StandardRemoteEnvironmentSpec.scala
```diff
@@ -185,56 +249,47 @@ class StandardRemoteEnvironmentSpec
...
-
-          Marshal(ProcessTestData.validScenarioDetailsForMigrations).to[RequestEntity].map { entity =>
-            HttpResponse(OK, entity = entity)
-          }
-        case UpdateProcess() if remoteProcessList contains expectedProcessDetails.name =>
-          onMigrate(Unmarshal(request).to[UpdateScenarioCommand])
-
-          Marshal(ValidationResult.errors(Map(), List(), List())).to[RequestEntity].map { entity =>
+        case Migrate() =>
+          /*          parseBodyToJson(request).as[MigrateScenarioRequestDtoV2] match {
```
mk-software-pl, line 267: let's add some comments also here.

mk-software-pl (approved):
LGTM

PR #5780, opened 2024-03-21
Fix Scala case class with fields in body serialization

Elmacioro:
## Describe your changes

Currently when we try to serialize `@TypeInfo` case class with fields defined inside of its body we get `IndexOutOfBoundsException`. 
For example serializing case class:
```
@TypeInfo(classOf[SimpleCaseClassWithAdditionalField.TypeInfoFactory])
final case class SimpleCaseClassWithAdditionalField(constructorField: String) {
  val bodyField: String = "body " + constructorField
}

object SimpleCaseClassWithAdditionalField {
  class TypeInfoFactory extends CaseClassTypeInfoFactory[SimpleCaseClassWithAdditionalField]
}
```

Will throw the following exception:

```
1
java.lang.IndexOutOfBoundsException: 1
	at scala.runtime.Statics.ioobe(Statics.java:196)
	at pl.touk.nussknacker.engine.flink.api.typeinfo.caseclass.SimpleCaseClassWithAdditionalField.productElement(CaseClassSerializationTest.scala:61)
	at pl.touk.nussknacker.engine.flink.api.typeinfo.caseclass.CaseClassSerializer.serialize(CaseClassSerializer.scala:102)
	at pl.touk.nussknacker.engine.flink.api.typeinfo.caseclass.CaseClassSerializer.serialize(CaseClassSerializer.scala:32)
	at pl.touk.nussknacker.engine.flink.api.typeinfo.caseclass.CaseClassSerializationTest.serializeAndDeserialize(CaseClassSerializationTest.scala:47)
	at pl.touk.nussknacker.engine.flink.api.typeinfo.caseclass.CaseClassSerializationTest.$anonfun$new$2(CaseClassSerializationTest.scala:35)
	at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
...
```

It happens because serializers are created for all of the class fields, including the ones defined inside of its body. However when values are actually passed to the serializer it serializes values only for the primary constructor.
With the proposed changes we create serializers for the primary constructor fields that should actually be serialized.

## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -7,6 +7,7 @@
   * openapi-circe-yaml: 0.6.0 -> 0.7.4 
 * [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
     * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
+* [#5780](https://github.com/TouK/nussknacker/pull/5780) Modified `CaseClassTypeInfoFactory` so Scala case classes are serialized properly
 
 1.14.0 (21 Mar 2024)
 -------------------------
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -7,7 +7,7 @@
   * openapi-circe-yaml: 0.6.0 -> 0.7.4 
 * [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
     * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
-* [#5780](https://github.com/TouK/nussknacker/pull/5780) Modified `CaseClassTypeInfoFactory` so Scala case classes are serialized properly
+* [#5780](https://github.com/TouK/nussknacker/pull/5780) Fixed Scala case classes serialization when a class has additional fields in its body
 
 1.14.0 (21 Mar 2024)
 -------------------------
--- a/engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
+++ b/engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
@@ -17,22 +17,9 @@ abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfo
       t: Type,
       genericParameters: java.util.Map[String, TypeInformation[_]]
   ): TypeInformation[T] = {
-    val runtimeClassType = classTag[T].runtimeClass
-    val mirror           = runtimeMirror(getClass.getClassLoader)
-
-    val fields = mirror
-      .classSymbol(runtimeClassType)
-      .primaryConstructor
-      .asMethod
-      .paramLists
-      .head
-    val fieldNames = fields.map(_.name.decodedName.toString)
-    val fieldTypes = fields.map { field =>
-      val fieldClass = mirror.runtimeClass(field.typeSignature.typeSymbol.asClass)
-      TypeExtractor.getForClass(fieldClass)
-    }
-
-    val classType = runtimeClassType.asInstanceOf[Class[T]]
+    val runtimeClassType         = classTag[T].runtimeClass
+    val (fieldNames, fieldTypes) = getClassFieldsInfo(runtimeClassType)
+    val classType                = runtimeClassType.asInstanceOf[Class[T]]
     new CaseClassTypeInfo[T](classType, Array.empty, fieldTypes.toIndexedSeq, fieldNames) {
       override def createSerializer(config: ExecutionConfig): TypeSerializer[T] = {
         new ScalaCaseClassSerializer[T](
@@ -41,6 +28,22 @@ abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfo
         )
       }
     }
+  }
+
+  private def getClassFieldsInfo(runtimeClassType: Class[_]): (List[String], List[TypeInformation[_]]) = {
+    val mirror = runtimeMirror(getClass.getClassLoader)
+    val fields = mirror
+      .classSymbol(runtimeClassType)
+      .primaryConstructor
+      .asMethod
+      .paramLists
+      .head
+    val fieldNames = fields.map(_.name.decodedName.toString)
+    val fieldTypes = fields.map { field =>
+      val fieldClass = mirror.runtimeClass(field.typeSignature)
+      TypeExtractor.getForClass(fieldClass)
+    }
+    (fieldNames, fieldTypes)
   }
 
 }
--- a/engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
+++ b/engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
@@ -8,6 +8,7 @@ import org.apache.flink.api.java.typeutils.runtime.NullableSerializer
 
 import java.lang.reflect.Type
 import scala.reflect._
+import scala.reflect.runtime.universe._
 
 // Generic class factory for creating CaseClassTypeInfo
 abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfoFactory[T] with Serializable {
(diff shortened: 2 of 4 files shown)


engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
```diff
@@ -17,8 +17,8 @@ abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfo
       genericParameters: java.util.Map[String, TypeInformation[_]]
   ): TypeInformation[T] = {
     val tClass     = classTag[T].runtimeClass.asInstanceOf[Class[T]]
-    val fieldNames = tClass.getDeclaredFields.map(_.getName).toList
-    val fieldTypes = tClass.getDeclaredFields.map(_.getType).map(TypeExtractor.getForClass(_))
+    val fieldNames = tClass.getDeclaredConstructors.head.getParameters.map(_.getName).toList
```
coutoPL, line 20: let's extract the constructor to val:
```scala
val primaryConstructor = tClass.getDeclaredConstructors.head
val fields = primaryConstructor.getParameters.map(_.getName).toList
...
```

engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
```diff
@@ -17,8 +17,8 @@ abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfo
...
+    val fieldTypes = tClass.getDeclaredConstructors.head.getParameterTypes.map(TypeExtractor.getForClass(_))
```
coutoPL, line 21: are we sure that the first constructor on the list is the primary one? I wonder if cannot be different on different JVM implementations. Maybe there is a way to check it?

coutoPL (commented):
Please add changelog entry (Looking from the client perspective)

engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
Elmacioro: Before implementing it this way i locally checked that the first one should be the primary one.
Unfortunately there does not seem to be an easy way to check it. 
We have to use Java reflection classes here and Java doesn't know anything about  case classes' primary constructors.
I tried mixing Scala reflection to get what we need and ended up with this...
```
abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfoFactory[T] with Serializable {

  override def createTypeInfo(
                               t: Type,
                               genericParameters: java.util.Map[String, TypeInformation[_]]
                             ): TypeInformation[T] = {
    val mirror = runtimeMirror(getClass.getClassLoader)
    val classType = classTag[T].runtimeClass
    val primaryConstructor = mirror.classSymbol(classType).primaryConstructor.asMethod
    val fieldNames = primaryConstructor.paramLists.flatten.map(_.name.toString)
    val fieldTypes = primaryConstructor.paramLists.flatten.map(_.typeSignature.getClass).map(TypeExtractor.getForClass(_))

    new CaseClassTypeInfo[T](classType.asInstanceOf[Class[T]], Array.empty, fieldTypes.toIndexedSeq, fieldNames) {
      override def createSerializer(config: ExecutionConfig): TypeSerializer[T] = {
        new ScalaCaseClassSerializer[T](
          classType.asInstanceOf[Class[T]],
          fieldTypes.map(typeInfo => NullableSerializer.wrap(typeInfo.createSerializer(config), true)).toArray
        )
      }
    }
  }

}
```
It passes the tests but looks horrible... With Scala reflection we can get the primary constructor with the `primaryConstructor` call.

engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
Elmacioro: Done

engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
Elmacioro: Well... After consulting the approach with Java and Scala reflection mixture I added a new proposed solution. It's a bit different from the one above as `fieldTypes` didn't contain correct types' sigantures. I also added a test with additional secondary constructor.

engine/flink/scala-utils/src/test/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassSerializationTest.scala
```diff
@@ -0,0 +1,92 @@
...
+
+object SimpleCaseClassWithAdditionalField {
+  class TypeInfoFactory extends CaseClassTypeInfoFactory[SimpleCaseClassWithAdditionalField]
+}
+
+@TypeInfo(classOf[SimpleCaseClassWithMultipleConstructors.TypeInfoFactory])
+final case class SimpleCaseClassWithMultipleConstructors(firstField: String, secondField: Double) {
+  val bodyField: String = "body " + firstField
+
+  def this(someField: Int, someSecondField: String) = {
```
coutoPL, line 84: maybe we can add sth like that too:
```
def this(someField: Int, someSecondField: String, toIgnore: String) = {
    this(someSecondField, someField)
  }
```

WDYT?

engine/flink/scala-utils/src/test/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassSerializationTest.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    val serializer   = getSerializer[SimpleCaseClassWithAdditionalField]
+    val deserialized = serializeAndDeserialize(serializer, input)
+
+    serializer shouldBe a[ScalaCaseClassSerializer[_]]
+    deserialized shouldEqual input
+  }
+
+  test("Case class with secondary constructor should be the same after serialization and deserialization") {
+    val input = SimpleCaseClassWithMultipleConstructors("value", 2.2)
```
coutoPL, line 42: WDYT about using the second constructor here?
```
new SimpleCaseClassWithMultipleConstructors(2.2, "value")
```

docs/Changelog.md
```diff
@@ -7,6 +7,7 @@
   * openapi-circe-yaml: 0.6.0 -> 0.7.4 
 * [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
     * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
+* [#5780](https://github.com/TouK/nussknacker/pull/5780) Modified `CaseClassTypeInfoFactory` so Scala case classes are serialized properly
```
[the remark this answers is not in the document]
Elmacioro: True, I modified the changelog message

engine/flink/scala-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassTypeInfoFactory.scala
```diff
@@ -16,14 +17,26 @@ abstract class CaseClassTypeInfoFactory[T <: Product: ClassTag] extends TypeInfo
...
       genericParameters: java.util.Map[String, TypeInformation[_]]
   ): TypeInformation[T] = {
-    val tClass     = classTag[T].runtimeClass.asInstanceOf[Class[T]]
-    val fieldNames = tClass.getDeclaredFields.map(_.getName).toList
-    val fieldTypes = tClass.getDeclaredFields.map(_.getType).map(TypeExtractor.getForClass(_))
+    val runtimeClassType = classTag[T].runtimeClass
+    val mirror           = runtimeMirror(getClass.getClassLoader)
 
-    new CaseClassTypeInfo[T](tClass, Array.empty, fieldTypes.toIndexedSeq, fieldNames) {
+    val fields = mirror
```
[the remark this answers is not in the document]
Elmacioro: That's a good idea. I moved it outside

engine/flink/scala-utils/src/test/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassSerializationTest.scala
Elmacioro: Yes, it might be good to test it with a constructor that has more parameters than the primary one.

engine/flink/scala-utils/src/test/scala/pl/touk/nussknacker/engine/flink/api/typeinfo/caseclass/CaseClassSerializationTest.scala
Elmacioro: It's a good idea to check it like that. I changed it in the code

coutoPL (approved):
LGTM

PR #5783, opened 2024-03-22
[NU-1512] Added allowedProcessingModes field to components endpoint

pielas:
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ProcessingMode.scala
```diff
@@ -32,4 +34,13 @@ object ProcessingMode {
     case other                               => throw new IllegalArgumentException(s"Not known processing mode: $other")
   }
 
+  sealed trait AllowedProcessingModes
```
mk-software-pl, line 37: I think that this sealed trait hierarchy is more related to `Component` than to `ProcessingMode` (even it uses the `ProcessingMode` case class). WDYT about moving it close to the `Component` definition?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/Source.scala
```diff
@@ -61,23 +63,39 @@ object SourceFactory {
...
+      inputType,
+      AllowedProcessingModes.SetOf(NonEmptySet.one(ProcessingMode.UnboundedStream))
+    )
 
   def noParamUnboundedStreamFactory[T: TypeTag](source: => Source)(implicit ev: T =:!= Nothing): SourceFactory =
-    NoParamSourceFactory(_ => source, Typed.fromDetailedType[T], Some(Set(ProcessingMode.UnboundedStream)))
+    NoParamSourceFactory(
+      _ => source,
+      Typed.fromDetailedType[T],
+      AllowedProcessingModes.SetOf(NonEmptySet.one((ProcessingMode.UnboundedStream)))
```
mk-software-pl, line 76: if we add `apply` like this:
```scala
    object SetOf {
      def apply(processingMode: ProcessingMode, processingModes: ProcessingMode*): SetOf = {
        new SetOf(NonEmptySet.of(processingMode, processingModes: _*))
      }
    }
```

we'd be able to write it like this:
```scala
AllowedProcessingModes.SetOf(ProcessingMode.UnboundedStream)
```

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/component/package.scala
```diff
@@ -45,6 +55,30 @@ package object component {
     def sortMethod(component: ComponentListElement): (String, String) = (component.name, component.id.value)
   }
 
+  implicit val allowedProcessingModesEncoder: Encoder[AllowedProcessingModes] = Encoder.instance {
+    case AllowedProcessingModes.AllProcessingModes            => ProcessingMode.all.asJson
```
mk-software-pl, line 59: We have Enumeratum in scope. WDYT about improving this `all` method? At the moment, we have to remember to add the new case to the `all` val.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
```diff
@@ -24,15 +25,16 @@ object ScenarioParametersDeterminer {
       case Nil => throw new IllegalStateException(s"Empty list of components for processing type: $processingType")
       case nonEmptyList =>
         val intersection = nonEmptyList.foldLeft(ProcessingMode.all) {
-          case (acc, Some(componentProcessingModes)) => acc.intersect(componentProcessingModes)
-          case (acc, None)                           => acc
+          case (acc, AllowedProcessingModes.SetOf(componentProcessingModes)) =>
```
mk-software-pl, line 28: we could have a `ToProcessingModes` extension class:

```scala
implicit class ToProcessingModes(allowedProcessingModes: AllowedProcessingModes) extends AnyVal {
  def toProcessingModes: Set[ProcessingMode] = ???
}
```
and doesn't have to pattern match on it here

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
```diff
@@ -24,15 +25,16 @@ object ScenarioParametersDeterminer {
...
+            acc.intersect(componentProcessingModes.toSortedSet)
+          case (acc, AllowedProcessingModes.AllProcessingModes) => acc
         }
         intersection.toList match {
           case oneMode :: Nil => oneMode
           // FIXME: some proper errors and tests for that
           case Nil =>
             val componentsWithDefinedAllowedProcessingModes = componentsToProcessingMode.collect {
```
mk-software-pl, line 36: WDYT about changing `collect` to `flatMap` here? IMO `collect` is not so safe in the context of potential extending sealed trait

mk-software-pl (commented):
Some tests failed. 
`NuDesignerApiAvailableToExposeYamlSpec` failed because for some reason OpenAPI was not generated. You can generate the new version of it manually by running `sbt generateDesignerOpenApi`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
```diff
@@ -15,24 +16,24 @@ object ScenarioParametersDeterminer {
       processingType: ProcessingType,
   ): ProcessingMode = {
     val componentsToProcessingMode = components.map { component =>
-      val allowedProcessingModes = component.implementation.allowedProcessingModes
+      val allowedProcessingModes = component.component.allowedProcessingModes
```
mk-software-pl, line 19: `component.component` looks weird. Is it intentional?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
```diff
@@ -15,24 +16,24 @@ object ScenarioParametersDeterminer {
...
         }
         intersection.toList match {
           case oneMode :: Nil => oneMode
           // FIXME: some proper errors and tests for that
           case Nil =>
-            val componentsWithDefinedAllowedProcessingModes = componentsToProcessingMode.collect {
-              case (id, Some(modes)) => id -> modes
+            val componentsWithDefinedAllowedProcessingModes = componentsToProcessingMode.flatMap {
+              case (id, AllowedProcessingModes.SetOf(modes)) => id -> modes :: Nil
+              case (_, AllowedProcessingModes.All)           => Nil
```
mk-software-pl, line 36: is this ok? we ignore the modes in the `All` case

designer/server/src/test/scala/pl/touk/nussknacker/ui/definition/component/DefaultComponentServiceSpec.scala
```diff
@@ -1,14 +1,16 @@
...
 import org.scalatest.Inside.inside
 import org.scalatest.OptionValues
 import org.scalatest.exceptions.TestFailedException
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import pl.touk.nussknacker.engine.ModelData
+import pl.touk.nussknacker.engine.api.component.Component.AllowedProcessingModes
 import pl.touk.nussknacker.engine.api.component.ComponentType._
-import pl.touk.nussknacker.engine.api.component._
+import pl.touk.nussknacker.engine.api.component.{ProcessingMode, _}
```
mk-software-pl, line 13: just do `import pl.touk.nussknacker.engine.api.component._`

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -1480,6 +1482,12 @@ components:
     AdditionalInfo:
       oneOf:
       - $ref: '#/components/schemas/MarkdownAdditionalInfo'
+    All:
```
mk-software-pl, line 1485: It shows that Schema for `allowedProcessingModes` is wrong. I saw that we encode it as an array of processing modes. Here it looks like we should provide `AllowedProcessingModes` object

docs/Changelog.md
```diff
@@ -8,6 +8,7 @@
   * openapi-circe-yaml: 0.6.0 -> 0.7.4 
 * [#5438](https://github.com/TouK/nussknacker/pull/5438) [#5495](https://github.com/TouK/nussknacker/pull/5495) Improvement in DeploymentManager API:
     * Alignment in the api of primary (deploy/cancel) actions and the experimental api of custom actions.
+* [#5783](https://github.com/TouK/nussknacker/pull/5783) Added allowedProcessingModes field to components endpoint
```
mk-software-pl, line 11: let's show it from the user's black box perspective. If there is no such perspective and the change is e.g. technical, so maybe we should consider adding it to the migration guide.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
pielas: Yes, this is intentional. Such behaviour has been achieved implicitly by using `.collect`. Moreover, it looks like in this case we are only interested in components which are allowed in separated sets of processing modes. Components that are allowed in all processing modes don't collide with any set of allowed processing modes and taking them into consideration could introduce noise

docs/Changelog.md
pielas: in this case it would be probably better not to mention this change anywhere. I think it doesn't fit to migration guide because this change doesn't affect current functionality in any way

arkadius: It affect code interfaces. Take a look at the summary of changes - there are changes in `components-api` module. It means, that authors of custom components will probably need to do modifications in their code.

From the user perspective, there also will be changes but they are hidden now - they will be available after the changes on the FE side. It would be nice to have links to all related PR's next to this entry but the author of FE side probably won't add the link to this PR.  Do you guys have idea how to handle it?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ScenarioParametersDeterminer.scala
mk-software-pl: WDYT about leaving a comment? Because I'm not sure if it's clear why we skip modes in this case

docs/Changelog.md
mk-software-pl: assuming the REST Components API is the public API (not the internal one), we can say that from now on, we expose in some endpoint the information about processing modes. 

And @Dzuming please remember to add information in the changelog from the UI perspective, when your change is done.

pielas: ok, I adapted changelog entry according to your suggestions, please verify

mk-software-pl: it's ok for me. According to the @arkadius's comment, I guess we need to add a Migration Guide's entry with the changes in the components-api module

arkadius: Yes, please also add a MG entry

arkadius: Thanks. It's fine now for me. @mk-software-pl resolve the comment if it is fine for you as well

mk-software-pl (approved):
LGTM

PR #5793, opened 2024-03-26
Add show and hide option for variables in test case

Elmacioro:
## Describe your changes

Currently, when we test a scenario with ad hoc tests or tests from a file we get textAreas with contents displayed fully as shown on the screen.

Zrzut ekranu z 2024-03-26 12-09-45

When contents are large it becomes cumbersome to scroll to find the variable that actually interests us, and generally it's not too pleasant to read.
To solve this I added mechanism with the "eye" symbol similar to the one introduced in this merged PR: https://github.com/TouK/nussknacker/pull/5690

If a variable's contents fit into 4 lines of text in textArea we do not show the "eye" icon.
Else, we render the icon in its "active" form hiding large content as shown on the screen below.

Zrzut ekranu z 2024-03-26 11-10-20

Additionally, I removed the unused space that occurred when the fade effect was active on a textArea (screen below). Currently, it takes the same space as other textAreas (screen above).

pusteMiejsce

## Checklist before merge

### DIFF
--- a/designer/client/src/components/graph/node-modal/tests/TestResults.tsx
+++ b/designer/client/src/components/graph/node-modal/tests/TestResults.tsx
@@ -2,11 +2,11 @@ import { isEmpty, isObject, join } from "lodash";
 import React from "react";
 import InfoIcon from "@mui/icons-material/Info";
 import NodeTip from "../NodeTip";
-import TestValue from "./TestValue";
 import { useTestResults } from "../TestResultsWrapper";
 import { NodeId } from "../../../../types";
 import { Box, FormControl, FormLabel } from "@mui/material";
 import { useTranslation } from "react-i18next";
+import TestResultsVariables from "./TestResultsVariables";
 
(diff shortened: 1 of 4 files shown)


designer/client/src/components/graph/node-modal/tests/TestResults.tsx
```diff
@@ -29,10 +29,7 @@ export default function TestResults({ nodeId }: { nodeId: NodeId }): JSX.Element
             {Object.keys(results.testResultsToShow.context.variables)
                 .sort((a, b) => a.localeCompare(b))
                 .map((key, ikey) => (
```
Dzuming, line 31: Perhaps we can improve readability by renaming `ikey` to `index`, which technically is the second map argument.

Dzuming (approved):
I added a small comment, Looks good!

designer/client/src/components/graph/node-modal/tests/TestResults.tsx
Elmacioro: Done, changed to `index`!

PR #5815, opened 2024-04-03
Validate blank `dictId` in fragment definition

wrzontek:
## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
@@ -351,7 +351,7 @@ object ProcessCompilationError {
       extends DuplicateFragmentOutputNames
       with ScenarioGraphLevelError
 
-  final case class EmptyMandatoryParameterConfigurationField(nodeId: String, qualifiedParamName: ParameterName)
+  final case class EmptyMandatoryField(nodeId: String, qualifiedFieldName: ParameterName)
       extends PartSubGraphCompilationError
       with InASingleNode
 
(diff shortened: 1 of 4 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -351,6 +351,10 @@ object ProcessCompilationError {
       extends DuplicateFragmentOutputNames
       with ScenarioGraphLevelError
 
+  final case class EmptyMandatoryParameterConfigurationField(nodeId: String, qualifiedParamName: ParameterName)
```
arkadius, line 354: Do you created a new error, because `EmptyMandatoryParameter` has `message` and `description`? If so, let's remove these parameters and reuse this class

wrzontek: i created a new error because it's logically a different thing, `dictId` here is not a "parameter" but a "parameter of a parameter" (which i called a "parameter configuration field" here)

arkadius: I think that we mix "field in the form" with "parameter of component" in these classes. Taka a look at:
`MissingParameters(Set(ParameterName("OutputVariable")`. What does "ParameterConfiguration" mean it this context?

wrzontek: both `EmptyMandatoryParameter` and `MissingParameters`  are awkward to use here with clear pointing to the qualified field name.

>Do you created a new error, because EmptyMandatoryParameter has message and description? If so, let's remove these parameters and reuse this class

those fields are used and non-constant, see it's three versions in `ParameterValidator.scala`

i've renamed the new error, if you still want to handle it differently let me know

PR #5816, opened 2024-04-03
[NU-1577] EitherT usages refactor

arkadius:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/util/EitherTImplicits.scala
```diff
@@ -1,29 +0,0 @@
-package pl.touk.nussknacker.ui.util
-
-import cats.Monad
-import cats.data.EitherT
-import cats.implicits.toFunctorOps
-
-import scala.language.higherKinds
-
-object EitherTImplicits {
```
arkadius, line 9: I removed this class because it moved the moment when a developer had to learn Monad Transformers a step further and other examples in the code showed that it was wrong.
We decided to stick with Monad Transformers so we need to be close to all available catz syntax extensions to deal with different transformations between types.

ForrestFairy (approved):

PR #5824, opened 2024-04-05
[NU-1579] decision table component - labels change

mk-software-pl:
## Describe your changes

- "Basic Decision Table" -> "Decision Table"
- "Expression" -> "Filtering expression"

## Checklist before merge


ForrestFairy (approved):
LGTM

engine/common/components-tests/src/test/scala/pl/touk/nussknacker/engine/common/components/DecisionTableSpec.scala
```diff
@@ -214,6 +214,9 @@ trait DecisionTableSpec
        |}""".stripMargin
   }
 
+  private val decisionTableParamName    = ParameterName("Decision Table")
+  private val filterExpressionParamName = ParameterName("Filtering expression")
```
lciolecki, line 218: Why do you copy these two params instead of  using BasicDecisionTableParameter and FilterDecisionTableExpressionParameter? IMO params from these objects should be public in case of someone wants to compare something in tests?

mk-software-pl: Why? Because of DRY? IMO you should not modify the code for test purposes. And the DRY principle don't have to be applied here

lciolecki: Remember that other projects could also use our test toolkit and can use decision table as well. IMO we should do that in the same way as is done in Kafka.

coutoPL: But what I proposed was to add the `decisionTable` method to the ScenarioBuilder's DSL. Why do you think that other projects won't be able to use it?

mk-software-pl: What I proposed was to add the `decisionTable` method to the ScenarioBuilder's DSL. Why do you think that external projects won't be able to use it?

lciolecki: Ahh now I see ;p Ok let's move on.

mk-software-pl: @lciolecki please take a look now

lciolecki (approved):
LGTM

PR #5825, opened 2024-04-05
[NU-1477] Deployments API: ability to pass parameters into sources

arkadius:
## Checklist before merge


designer/server/src/test/scala/pl/touk/nussknacker/ui/api/DeploymentApiHttpServiceBusinessSpec.scala
```diff
@@ -75,16 +74,24 @@ class DeploymentApiHttpServiceBusinessSpec
...
-    .source("source", "table", "Table" -> Expression.spel("'transactions'"))
+    .source(sourceNodeId, "table", "Table" -> Expression.spel("'transactions'"))
     .emptySink(
       "sink",
       "table",
       "Table" -> Expression.spel("'transactions_summary'"),
       "Value" -> Expression.spel("#input")
     )
 
+  private val correctDeploymentRequest = s"""{
```
mslabek, line 89: (Not related to this line) When I tried to test this locally using batch dev setup, I get a "Comment is required." message. So to test manually this I have to disable the `deploymentCommentSettings` globally. Or is there another way / other way to pass the comment?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/NodesDeploymentData.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.api.component
+
+import io.circe.generic.extras.semiauto.{deriveUnwrappedDecoder, deriveUnwrappedEncoder}
+import io.circe.{Decoder, Encoder}
+import pl.touk.nussknacker.engine.api.NodeId
+
+final case class NodesDeploymentData(dataByNodeId: Map[NodeId, NodeDeploymentData])
```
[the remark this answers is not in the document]
arkadius: At the beginning I named it *Parameters but then I realize that we have parameters in nodes and it means sth else. Someone could think that in `DeploymentNodesParameters.parameters` should be passed the exact structure as in nodes - so if you want to configure offset you should have a separate parameter for offset, if want to configure param1 you should have param1 etc. In 2nd layer (nodes) we name it `NodeDependency` which can be anything that node need. `Parameter` is a special case of `NodeDependency`. It isn't visible in the class structure but it is used in most places together. TL;DR I'm a little bit confused, because we currently doesn't have such a clean model. If we cleaned up 1st and 2nd layer, the decision would be easy, but without this, we can introduce sth that looks like cleaned module but when someone dig deeper, he/she realize that it isn't.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/DeploymentApiHttpServiceBusinessSpec.scala
arkadius: I'll check it

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/NodesDeploymentData.scala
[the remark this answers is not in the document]
mslabek: Ok, its more compilcated than I thought, lets leave it as it.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/DeploymentApiHttpServiceBusinessSpec.scala
arkadius: I added the new request parameter: `comment`. I needed to extract another test class because designer config is setup on class level. Now it should work also on the staging

mslabek (approved):

PR #5843, opened 2024-04-09
[NU-1524] Add endpoint to provide statistics URLs

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -2,7 +2,8 @@ package pl.touk.nussknacker.ui.api
 
 import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints
-import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticUrlResponseDto
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticError.InvalidURL
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.{StatisticError, StatisticUrlResponseDto}
 import pl.touk.nussknacker.ui.security.api.AuthenticationResources
 import pl.touk.nussknacker.ui.statistics.UsageStatisticsReportsSettingsDeterminer
 
@@ -15,14 +16,21 @@ class StatisticsApiHttpService(
     extends BaseHttpService(authenticator)
     with LazyLogging {
 
-  private val endpoints = new StatisticsApiEndpoints()
+  private val endpoints = new StatisticsApiEndpoints(authenticator.authenticationMethod())
 
   expose {
-    endpoints.statisticUrlEndpoint
-      .serverLogicSuccess { _ =>
+    endpoints.statisticUsageEndpoint
+      .serverSecurityLogic(authorizeKnownUser[StatisticError])
+      .serverLogic { _ => _ =>
         determiner
           .determineStatisticsUrl()
-          .map(maybeUrl => StatisticUrlResponseDto(maybeUrl.toList))
+          .map(maybeUrl => success(StatisticUrlResponseDto(maybeUrl.toList)))
+          .recover {
+            case e: IllegalStateException => {
+              logger.warn("Statistics URL construction failed", e)
+              businessError(InvalidURL)
+            }
+          }
       }
   }
 
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -1,0 +1,29 @@
+package pl.touk.nussknacker.ui.api
+
+import com.typesafe.scalalogging.LazyLogging
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticUrlResponseDto
+import pl.touk.nussknacker.ui.security.api.AuthenticationResources
+import pl.touk.nussknacker.ui.statistics.UsageStatisticsReportsSettingsDeterminer
+
+import scala.concurrent.ExecutionContext
+
+class StatisticsApiHttpService(
+    authenticator: AuthenticationResources,
+    determiner: UsageStatisticsReportsSettingsDeterminer
+)(implicit executionContext: ExecutionContext)
+    extends BaseHttpService(authenticator)
+    with LazyLogging {
+
+  private val endpoints = new StatisticsApiEndpoints()
+
+  expose {
+    endpoints.statisticUrlEndpoint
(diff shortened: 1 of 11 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
```diff
@@ -0,0 +1,49 @@
...
+class StatisticsApiEndpoints extends BaseEndpointDefinitions {
+
+  import StatisticsApiEndpoints.Dtos._
+
+  lazy val statisticUrlEndpoint: PublicEndpoint[Unit, Unit, StatisticUrlResponseDto, Any] =
+    baseNuApiEndpoint
+      .summary("Statistics URL service")
+      .tag("Statistics")
+      .get
+      .in("statistic" / "url")
```
mk-software-pl, line 22: WDYT about abstracting over the current implementation with the URL? I'm talking about the name of the endpoint. Like eg. "statistics/usage"?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpServiceSecuritySpec.scala
```diff
@@ -0,0 +1,40 @@
...
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers {
+
+  private val nuVersion = BuildInfo.version
+
+  "The statistic URL endpoint when" - {
+    "require no auth to return single URL" in {
+      given()
+        .when()
+        .noAuth()
```
mk-software-pl, line 24: Are we sure? This potentially leads to an information leak. Maybe the frontend should only send the info where a user is logged. But the leak will be there too (are we sure that the information is not user's specific?). Maybe we should discuss this.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
```diff
@@ -0,0 +1,49 @@
...
+        )
+      )
+
+}
+
+object StatisticsApiEndpoints {
+
+  object Dtos {
+    @derive(encoder, decoder, schema)
+    final case class StatisticUrlResponseDto private (urls: List[String])
```
mk-software-pl, line 46: Maybe we can represent a lemonlabs URL or at least a java URL to make sure the URL is valid?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
lukasz-bigorajski: "statistics/usage" sounds great

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpServiceSecuritySpec.scala
lukasz-bigorajski: I agree, I will change that

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    object StatisticError {
+      final case object InvalidURL extends StatisticError
+
+      private def deserializationException =
+        (ignored: Any) => throw new IllegalStateException("Deserializing errors is not supported.")
+
+      implicit val invalidURLCodec: Codec[String, InvalidURL.type, CodecFormat.TextPlain] = {
+        Codec.string.map(
+          Mapping.from[String, InvalidURL.type](deserializationException)(e => "Constructed URL is invalid.")
```
mk-software-pl, line 84: 1. maybe we should use `StatisticError` here instead of `InvalidURL.type`? And be prepared for new cases (this would be a cosmetic change only)
2. from the caller's perspective this error message can be misleading because they call statistics error and expect to get statistics (in some form). The caller doesn't pass any URL, so it may be not obvious at the first glance what "Constructed URL is invalid" means. From their perspective it's a problem with providing statistics. So, maybe the message should be changed. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
```diff
@@ -0,0 +1,37 @@
...
+  expose {
+    endpoints.statisticUsageEndpoint
+      .serverSecurityLogic(authorizeKnownUser[StatisticError])
+      .serverLogic { _ => _ =>
+        determiner
+          .determineStatisticsUrl()
+          .map(maybeUrl => success(StatisticUrlResponseDto(maybeUrl.toList)))
+          .recover {
+            case e: IllegalStateException => {
+              logger.warn("Statistics URL construction failed", e)
```
mk-software-pl, line 30: error level?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -109,12 +109,12 @@ class UsageStatisticsReportsSettingsDeterminer(
     fetchNonArchivedScenariosInputData: () => Future[List[ScenarioStatisticsInputData]]
 )(implicit ec: ExecutionContext) {
 
-  def determineStatisticsUrl(): Future[Option[String]] = {
+  def determineStatisticsUrl(): Future[Option[URL]] = {
```
mk-software-pl, line 112: WDYT about changing the return type to `Either[Error, Option[URL]`?

lukasz-bigorajski: It's changed in this PR: https://github.com/TouK/nussknacker/pull/5867

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
lukasz-bigorajski: For me the error level is when application crashes or cannot work because of it, and the warn level is when sth occurred exceptional but application works normally. Is there in NU different strategy?

mk-software-pl: I'm not aware of any Nu strategy :P But your explanation seems to be reasonable 👍

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
lukasz-bigorajski: Done

mk-software-pl (approved):
LGTM!

PR #5860, opened 2024-04-12
[NU-1522] Extend statistics mechanism

ForrestFairy:
## Describe your changes
Extend statistics mechanism to gather more information about the use of Nussknacker.

Statistic of custom components used should be improved in the future, for example by looking at component providers. But priority is to get as much statistics as possible before 1.15 release.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -46,20 +54,72 @@ object UsageStatisticsReportsSettingsDeterminer {
...
+        }.sequence
+      } yield activity
+    }
+
+    def fetchComponentList(): Future[List[ComponentListElement]] = {
+      implicit val user: LoggedUser = NussknackerInternalUser.instance
+      componentService.getComponentsList
+    }
+
+    def fetchScenarioActions(): Future[List[List[ProcessAction]]] = {
```
lukasz-bigorajski, line 100: Shouldn't it be: `Future[List[ProcessAction]]`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -46,20 +54,72 @@ object UsageStatisticsReportsSettingsDeterminer {
...
+      implicit val user: LoggedUser = NussknackerInternalUser.instance
+      processService
+        .getLatestProcessesWithDetails(
```
lukasz-bigorajski, line 103: We invoke `processService.getLatestProcessesWithDetails` for the 3 times, maybe we could invoke it only once and then filter for apropriate elements?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
     } yield basicStatistics ++ scenariosStatistics
 
+    val queryParamsFromScenarios = fetchNonArchivedScenariosInputData()
+      .map {
+        case List() => Map.empty[String, String]
+        case scenariosInputData =>
+          val scenarios = scenariosInputData.count(!_.isFragment)
+
+          //        Nodes stats
+          val nodes        = scenariosInputData.map(_.nodes).sorted
```
lukasz-bigorajski, line 203: sortedNodes?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
       scenariosStatistics = scenariosInputData.map(determineStatisticsForScenario).combineAll.mapValuesNow(_.toString)
       fingerprint <- fingerprintSupplier(config, nuFingerprintFileName)
       basicStatistics = determineBasicStatistics(fingerprint, config)
     } yield basicStatistics ++ scenariosStatistics
 
+    val queryParamsFromScenarios = fetchNonArchivedScenariosInputData()
+      .map {
+        case List() => Map.empty[String, String]
+        case scenariosInputData =>
+          val scenarios = scenariosInputData.count(!_.isFragment)
```
lukasz-bigorajski, line 200: scenariosCount?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+
+          //        Nodes stats
```
lukasz-bigorajski, line 202: I would remove empty spaces and new lines

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+          val nodes        = scenariosInputData.map(_.nodes).sorted
+          val nodesMedian  = nodes.get(nodes.length / 2).getOrElse(0)
+          val nodesAverage = nodes.sum / scenariosInputData.length
+
+          //        Category stats
+          val categories = scenariosInputData.map(_.processCategory).toSet.size
+
+          //        Version stats
+          val versions        = scenariosInputData.map(_.processVersion.value).sorted
```
lukasz-bigorajski, line 211: sortedVersions?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+        case scenariosInputData =>
+          val scenarios = scenariosInputData.count(!_.isFragment)
+
+          //        Nodes stats
+          val nodes        = scenariosInputData.map(_.nodes).sorted
+          val nodesMedian  = nodes.get(nodes.length / 2).getOrElse(0)
+          val nodesAverage = nodes.sum / scenariosInputData.length
+
+          //        Category stats
+          val categories = scenariosInputData.map(_.processCategory).toSet.size
```
lukasz-bigorajski, line 208: categoriesCount?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+
+          //        Version stats
+          val versions        = scenariosInputData.map(_.processVersion.value).sorted
+          val versionsMedian  = versions.get(versions.length / 2).getOrElse(-1)
```
lukasz-bigorajski, line 212: In nodesMedian there is `orElse(0)` and here is -1, why is like that?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+            "v_m" -> versionsMedian,
+            "v_v" -> versionsAverage,
+            "a_n" -> authors,
+            "m_f" -> fragmentsMedian,
+            "v_f" -> fragmentsAverage,
+          ).mapValuesNow(_.toString)
+      }
+    val activityQueryParams = fetchActivity()
+      .map {
+        case List()           => Map.empty[String, String]
```
lukasz-bigorajski, line 236: Nil?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+          //        Category stats
+          val categories = scenariosInputData.map(_.processCategory).toSet.size
+
+          //        Version stats
+          val versions        = scenariosInputData.map(_.processVersion.value).sorted
+          val versionsMedian  = versions.get(versions.length / 2).getOrElse(-1)
+          val versionsAverage = versions.sum / scenariosInputData.length
+
+          //        Author stats
+          val authors = scenariosInputData.map(_.createdBy).toSet.size
```
lukasz-bigorajski, line 216: authorsCount?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+            "m_f" -> fragmentsMedian,
+            "v_f" -> fragmentsAverage,
+          ).mapValuesNow(_.toString)
+      }
+    val activityQueryParams = fetchActivity()
+      .map {
+        case List()           => Map.empty[String, String]
+        case listOfActivities =>
+          //        Attachment stats
+          val attachments = listOfActivities
```
lukasz-bigorajski, line 239: sortedAttachmentCountList?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+            .map { activity =>
+              activity.attachments
+            }
+            .map(_.length)
+            .sorted
+          val attachmentMedian  = attachments.get(attachments.length / 2).getOrElse(0)
```
lukasz-bigorajski, line 245: I'm wondering if we needs median for the attachments or whether they exist. Maybe we could leave only average and instead of exists, we could calculate total sum?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+          val attachmentExist   = attachments.count(_ > 0)
+
+          //        Comment stats
+          val comments = listOfActivities
+            .map { activity =>
+              activity.comments
+            }
+            .map(_.length)
+            .sorted
+          val commentsMedian  = comments.get(comments.length / 2).getOrElse(0)
```
lukasz-bigorajski, line 256: Maybe totalSum instead of median is better?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+            "v_c" -> commentsAverage
+          ).mapValuesNow(_.toString)
+      }
+
+    val componentStats =
+      fetchComponentList().map {
+        case List() => Map.empty[String, String]
+        case componentList =>
+          val withoutFragments = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
+          val byName           = withoutFragments.groupBy(_.name)
```
lukasz-bigorajski, line 273: componentsByNameCount?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+            "c_n" -> byName.size.toString
+          )
+      }
+
+    val uptimeStats =
+      fetchScenarioActions().map {
+        case List() => Map.empty[String, String]
+        case listForScenarios =>
+          val uptimes = listForScenarios.map { listOfActions =>
+            val durationsOfDeployment = calculateUpTimeStats(listOfActions).filter(_.isDefined).sequence.get.sorted
```
lukasz-bigorajski, line 285: Maybe this is better?
```scala
calculateUpTimeStats(listOfActions).flatten.sorted match {
              case Nil => 0
              case deploymentsUpTimeStats => deploymentsUpTimeStats.sum / deploymentsUpTimeStats.length 
            }
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+              durationsOfDeployment.sum / durationsOfDeployment.length
+            } else {
+              0
+            }
+
+          }
+          val uptimeAverage = uptimes.sum / uptimes.length
+          val uptimeMedian  = uptimes.get(uptimes.length / 2).get
+          Map(
+            "m_u" -> uptimeMedian,
```
lukasz-bigorajski, line 297: IMHO median for uptime does not make sense. Instead of this, I would add min and max

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+    val queryParamsFromScenarios = fetchNonArchivedScenariosInputData()
+      .map {
+        case List() => Map.empty[String, String]
+        case scenariosInputData =>
+          val scenarios = scenariosInputData.count(!_.isFragment)
+
+          //        Nodes stats
+          val nodes        = scenariosInputData.map(_.nodes).sorted
+          val nodesMedian  = nodes.get(nodes.length / 2).getOrElse(0)
+          val nodesAverage = nodes.sum / scenariosInputData.length
```
lukasz-bigorajski, line 205: Maybe we could add min and max?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+
+          //        Category stats
+          val categories = scenariosInputData.map(_.processCategory).toSet.size
+
+          //        Version stats
+          val versions        = scenariosInputData.map(_.processVersion.value).sorted
+          val versionsMedian  = versions.get(versions.length / 2).getOrElse(-1)
+          val versionsAverage = versions.sum / scenariosInputData.length
```
lukasz-bigorajski, line 213: Maybe we could add min and max?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+  }
+
+  private def calculateUpTimeStats(processActions: List[ProcessAction]): List[Option[Long]] = {
+    val iter                    = processActions.reverseIterator
+    var hasStarted              = false
+    var lastActionTime: Instant = Instant.now
+    val listOfUptime = iter.map {
+      case ProcessAction(_, _, _, _, _, performedAt, ScenarioActionName.Deploy, _, _, _, _, _) =>
+        if (hasStarted) {
+          val uptime = performedAt.getEpochSecond - lastActionTime.getEpochSecond
```
lukasz-bigorajski, line 317: If lastActionTime is assinged to `now` I am sure that the uptime is below 0 because now is bigger than performedAt from past

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,14 +185,161 @@ class UsageStatisticsReportsSettingsDeterminer(
...
+
+    // Combine all stats
+    queryParamsFromScenarios
+      .zipWith(activityQueryParams)(_ ++ _)
+      .zipWith(componentStats)(_ ++ _)
+      .zipWith(uptimeStats)(_ ++ _)
+      .zipWith(generalStats)(_ ++ _)
+  }
+
+  private def calculateUpTimeStats(processActions: List[ProcessAction]): List[Option[Long]] = {
```
lukasz-bigorajski, line 310: scenario

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -148,5 +358,10 @@ private[statistics] case class ScenarioStatisticsInputData(
     processingMode: ProcessingMode,
     deploymentManagerType: DeploymentManagerType,
     // For fragments status is empty
-    status: Option[StateStatus]
+    status: Option[StateStatus],
+    nodes: Int,
+    processCategory: String,
+    processVersion: VersionId,
+    createdBy: String,
+    fragmentsUsed: Int
```
lukasz-bigorajski, line 366: fragmentsUsedCount?

lukasz-bigorajski (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
ForrestFairy: But the scenario would need to be started before by another `ProcessAction` which would overwrite `lastActionTime` as we go from the oldest `ProcessAction` to newest. But to be sure I can add to get max between the calculation and 0

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
ForrestFairy: I think of it as List of `ProcessAction` per scenario, hence List of Lists, but I may be wrong in my understanding

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
ForrestFairy: Doesn't apply anymore as we will gather statistics only for currently running scenarios

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
ForrestFairy: Now it's `Future[List[ProcessSaction]]` as we gather statistics only for currently running scenarios and so we only need `lastDeployedAction`

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: I see the `getLatestProcessesWithDetails` in `fetchActivity` and `fetchNonArchivedScenarioParameters`, maybe we can `processId` to `ScenarioStatisticsInputData` and then pass to `fetchActivity(scenariosInputData)`. What do you think?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: Maye we extract function to calculate median on List[Long]?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: I only meant that Nil is equivalent to List()

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: I would move `.size` from map to val declaration

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,138 @@
...
+      val versionsMax     = sortedVersions.head
+      val versionsMin     = sortedVersions.last
+      //        Author stats
+      val authorsCount = scenariosInputData.map(_.createdBy).toSet.size
+      //        Fragment stats
+      val fragments        = scenariosInputData.map(_.fragmentsUsedCount).sorted
+      val fragmentsMedian  = fragments.get(fragments.length / 2).getOrElse(0)
+      val fragmentsAverage = fragments.sum / scenariosCount
+      //          Uptime stats
+      val lastActions = scenariosInputData.map(_.lastDeployedAction).filter(_.isDefined).sequence.get
```
lukasz-bigorajski, line 38: get on Option is unsafe, maybe we could use: `toList.flatten` or map actions to uptime and then use getOrElse?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,13 +175,19 @@ class UsageStatisticsReportsSettingsDeterminer(
     }
   }
 
-  private[statistics] def determineQueryParams(): Future[Map[String, String]] =
+  private[statistics] def determineQueryParams(): Future[Map[String, String]] = {
     for {
       scenariosInputData <- fetchNonArchivedScenariosInputData()
       scenariosStatistics = scenariosInputData.map(determineStatisticsForScenario).combineAll.mapValuesNow(_.toString)
```
lukasz-bigorajski, line 181: I would move this logic to the ScenarioStatistics class too.

designer/server/src/test/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminerTest.scala
```diff
@@ -170,34 +214,80 @@ class UsageStatisticsReportsSettingsDeterminerTest
...
       UsageStatisticsReportsConfig(enabled = true, Some(sampleFingerprint), None),
       fingerprintSupplier,
-      () => Future.successful(List(nonRunningScenario, runningScenario, fragment, k8sRRScenario))
+      () => Future.successful(List(nonRunningScenario, runningScenario, fragment, k8sRRScenario)),
+      () => Future.successful(processActivityList),
+      () => Future.successful(componentList)
     ).determineQueryParams().futureValue
 
     val expectedStats = Map(
+      "a_n"     -> "1",
```
lukasz-bigorajski, line 271: For better readability I would replace the string keys with the ADT, fe:
`AuthorsCount.name -> "1"`

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
lukasz-bigorajski, line 38: IMHO is not the correct way to calculate an uptime. `lastDeployedAction` is a deploy action but is necessarily the last one (you can check it here:pl.touk.nussknacker.ui.process.repository.DBFetchingProcessRepository.fetchLatestProcessDetailsByQueryAction), fe. the last one can be cancel. IMHO we should take all actions and if the last one is the deploy one then we can calculate the uptime, otherwise we cannot do any calculations.

lukasz-bigorajski (changes requested):
It's looks good. I added only few comments ;)

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,178 @@
...
+  // - processing mode: streaming, r-r, batch
+  // - dm type: flink, k8s, embedded, custom
+  // - status: active (running), other
+  // We have two options:
+  // 1. To aggregate statistics for every combination - it give us 3*4*2 + 3*4 = 36 parameters
+  // 2. To aggregate statistics for every dimension separately - it gives us 2+3+4+1 = 10 parameters
+  // We decided to pick the 2nd option which gives a reasonable balance between amount of collected data and insights
+  private[statistics] def determineStatisticsForScenario(inputData: ScenarioStatisticsInputData): Map[String, Int] = {
+    Map(
+      "s_s"     -> !inputData.isFragment,
```
lukasz-bigorajski, line 136: Maybe we could move these keys to parameters too?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,178 @@
...
+      "s_pm_rr" -> (inputData.processingMode == ProcessingMode.RequestResponse),
+      "s_dm_f"  -> (inputData.deploymentManagerType == flinkDeploymentManagerType),
+      "s_dm_l"  -> (inputData.deploymentManagerType == liteK8sDeploymentManagerType),
+      "s_dm_e"  -> (inputData.deploymentManagerType == liteEmbeddedDeploymentManagerType),
+      "s_dm_c"  -> !knownDeploymentManagerTypes.contains(inputData.deploymentManagerType),
+      "s_a"     -> inputData.status.contains(SimpleStateStatus.Running),
+    ).mapValuesNow(if (_) 1 else 0)
+  }
+
+  private def calculateMedian[T: Numeric](list: List[T]): T = {
```
lukasz-bigorajski, line 149: I will fell calm if this function could handle empty list too :). Maybe we should extract analogous function for the average? Additionally I would extract min and max functions which would handle empty list

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,178 @@
...
+      val versionsMedian  = calculateMedian(sortedVersions)
+      val versionsAverage = sortedVersions.sum / scenariosInputData.length
+      val versionsMax     = sortedVersions.head
+      val versionsMin     = sortedVersions.last
+      //        Author stats
+      val authorsCount = scenariosInputData.map(_.createdBy).toSet.size
+      //        Fragment stats
+      val fragments        = scenariosInputData.map(_.fragmentsUsedCount).sorted
+      val fragmentsMedian  = calculateMedian(fragments)
+      val fragmentsAverage = fragments.sum / scenariosCount
```
lukasz-bigorajski, line 39: Fragments count? Do we have test for it?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,178 @@
...
+      val withoutFragments      = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
+      val componentsByNameCount = withoutFragments.groupBy(_.name).size
+
+      Map(
+        ComponentsCount -> componentsByNameCount
+      ).map { case (k, v) => (k.toString, v.toString) }
+    }
+  }
+
+  private val flinkDeploymentManagerType = DeploymentManagerType("flinkStreaming")
```
lukasz-bigorajski, line 116: Variables in the class should be before function like in the determiner class

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -0,0 +1,178 @@
...
+  // We have four dimensions:
+  // - scenario / fragment
+  // - processing mode: streaming, r-r, batch
+  // - dm type: flink, k8s, embedded, custom
+  // - status: active (running), other
+  // We have two options:
+  // 1. To aggregate statistics for every combination - it give us 3*4*2 + 3*4 = 36 parameters
+  // 2. To aggregate statistics for every dimension separately - it gives us 2+3+4+1 = 10 parameters
+  // We decided to pick the 2nd option which gives a reasonable balance between amount of collected data and insights
+  private[statistics] def determineStatisticsForScenario(inputData: ScenarioStatisticsInputData): Map[String, Int] = {
```
lukasz-bigorajski, line 134: Maybe we could move the logic with combining results of this function here because here we have calculation for single data here and combining these results in the other place

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -122,13 +136,22 @@ class UsageStatisticsReportsSettingsDeterminer(
     }
   }
 
-  private[statistics] def determineQueryParams(): Future[Map[String, String]] =
```
lukasz-bigorajski, line 125: I guess we can remove: `[statistics]` now

lukasz-bigorajski (changes requested):
Last comments and finish

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
ForrestFairy: And currently it doesn't? `Get` gives `Some(value)` if given index exists in a list or `None` if it's outside the scope. Then `getOrElse` transforms `Option` to either value or zero.
At least that's my understanding but may be wrong

lukasz-bigorajski: Sorry my bad, it's safe :)

lukasz-bigorajski: I remember why I wrote it. I saw and average calculation and I generalised it to the median. So other remarks I would still consider to do :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
ForrestFairy: It is used in tests, after moving in `ScenarioStatisticsTest`

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
ForrestFairy: I changed it to fragmentsUsedCount so it won't be mistaken with amount of created fragments

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
lukasz-bigorajski:
```
def calculateAverage[T: Numeric](list: List[T]): T = {
  val result = implicitly[Numeric[T]].toInt(list.sum) / list.length
  implicitly[Numeric[T]].fromInt(result)
}

calculateAverage(List[Int]())
```
```
def calculateAverage[T](list: List[T])(implicit evidence$1: Numeric[T]): T

java.lang.ArithmeticException: / by zero
  at calculateAverage(<console>:2)
  ... 41 elided
```

lukasz-bigorajski (approved):

PR #5867, opened 2024-04-15
[NU-1574] Statistics error handling

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -27,7 +27,7 @@ class StatisticsApiHttpService(
           .prepareStatisticsUrl()
           .map {
             case Left(CannotGenerateStatisticsError) => businessError(ApiCannotGenerateStatisticError)
-            case Right(url)                          => success(StatisticUrlResponseDto(url.toList))
+            case Right(maybeUrl)                     => success(StatisticUrlResponseDto(maybeUrl.toList))
           }
       }
   }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -1,24 +1,23 @@
 package pl.touk.nussknacker.ui.api
 
-import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints
-import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticError.{DbError, InvalidURL}
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticError.{
+  DbError => ApiDbError,
+  InvalidURL
+}
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.{StatisticError, StatisticUrlResponseDto}
 import pl.touk.nussknacker.ui.security.api.AuthenticationResources
-import pl.touk.nussknacker.ui.statistics.UsageStatisticsReportsSettingsDeterminer
+import pl.touk.nussknacker.ui.statistics.{DbError, UrlError, UsageStatisticsReportsSettingsDeterminer}
 
-import java.sql.SQLException
 import scala.concurrent.ExecutionContext
 
 class StatisticsApiHttpService(
     authenticator: AuthenticationResources,
     determiner: UsageStatisticsReportsSettingsDeterminer
 )(implicit executionContext: ExecutionContext)
-    extends BaseHttpService(authenticator)
-    with LazyLogging {
+    extends BaseHttpService(authenticator) {
 
-  private val endpoints    = new StatisticsApiEndpoints(authenticator.authenticationMethod())
-  private val errorMessage = "Statistics URL construction failed"
+  private val endpoints = new StatisticsApiEndpoints(authenticator.authenticationMethod())
 
   expose {
     endpoints.statisticUsageEndpoint
@@ -26,17 +25,13 @@ class StatisticsApiHttpService(
       .serverLogic { _ => _ =>
         determiner
           .determineStatisticsUrl()
-          .map(maybeUrl => success(StatisticUrlResponseDto(maybeUrl.toList)))
-          // TODO: how should we handle JDK and DB errors?
-          .recover {
-            case e: IllegalStateException => {
-              logger.warn(errorMessage, e)
-              businessError(InvalidURL)
-            }
-            case e: SQLException => {
-              logger.warn(errorMessage, e)
-              businessError(DbError)
-            }
+          .map {
+            case Left(error) =>
+              error match {
+                case DbError  => businessError(ApiDbError)
+                case UrlError => businessError(InvalidURL)
+              }
+            case Right(url) => success(StatisticUrlResponseDto(url.toList))
           }
       }
   }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -1,13 +1,13 @@
 package pl.touk.nussknacker.ui.api
 
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints
-import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.StatisticError.{
-  DbError => ApiDbError,
-  InvalidURL
+import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.{
+  CannotGenerateStatisticError => ApiCannotGenerateStatisticError,
+  StatisticError,
+  StatisticUrlResponseDto
 }
-import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.{StatisticError, StatisticUrlResponseDto}
 import pl.touk.nussknacker.ui.security.api.AuthenticationResources
-import pl.touk.nussknacker.ui.statistics.{DbError, UrlError, UsageStatisticsReportsSettingsDeterminer}
+import pl.touk.nussknacker.ui.statistics.{CannotGenerateStatisticsError, UsageStatisticsReportsSettingsDeterminer}
 
 import scala.concurrent.ExecutionContext
 
@@ -24,14 +24,10 @@ class StatisticsApiHttpService(
       .serverSecurityLogic(authorizeKnownUser[StatisticError])
       .serverLogic { _ => _ =>
         determiner
-          .determineStatisticsUrl()
+          .prepareStatisticsUrl()
           .map {
-            case Left(error) =>
-              error match {
-                case DbError  => businessError(ApiDbError)
-                case UrlError => businessError(InvalidURL)
-              }
-            case Right(url) => success(StatisticUrlResponseDto(url.toList))
+            case Left(CannotGenerateStatisticsError) => businessError(ApiCannotGenerateStatisticError)
+            case Right(url)                          => success(StatisticUrlResponseDto(url.toList))
           }
       }
   }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
@@ -2,12 +2,12 @@ package pl.touk.nussknacker.ui.api
 
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints
 import pl.touk.nussknacker.ui.api.description.StatisticsApiEndpoints.Dtos.{
-  CannotGenerateStatisticError => ApiCannotGenerateStatisticError,
+  CannotGenerateStatisticError,
   StatisticError,
   StatisticUrlResponseDto
 }
 import pl.touk.nussknacker.ui.security.api.AuthenticationResources
-import pl.touk.nussknacker.ui.statistics.{CannotGenerateStatisticsError, UsageStatisticsReportsSettingsDeterminer}
+import pl.touk.nussknacker.ui.statistics.UsageStatisticsReportsSettingsDeterminer
 
 import scala.concurrent.ExecutionContext
 
@@ -26,8 +26,8 @@ class StatisticsApiHttpService(
         determiner
           .prepareStatisticsUrl()
           .map {
-            case Left(CannotGenerateStatisticsError) => businessError(ApiCannotGenerateStatisticError)
-            case Right(maybeUrl)                     => success(StatisticUrlResponseDto(maybeUrl.toList))
+            case Left(_)         => businessError(CannotGenerateStatisticError)
+            case Right(maybeUrl) => success(StatisticUrlResponseDto(maybeUrl.toList))
           }
       }
   }
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
@@ -5,9 +5,9 @@ import db.util.DBIOActionInstances.DB
 import pl.touk.nussknacker.ui.db.DbRef
 import slick.jdbc.JdbcProfile
 
+import java.sql.SQLException
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.higherKinds
-import scala.util.{Failure, Success}
 
 class DBIOActionRunner(dbRef: DbRef) extends LazyLogging {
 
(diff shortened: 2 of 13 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/FingerprintService.scala
```diff
@@ -16,26 +16,30 @@ class FingerprintService(dbioRunner: DBIOActionRunner, fingerprintRepository: Fi
...
+      case Some(fingerprintFromConfig) => Future.successful(Right(new Fingerprint(fingerprintFromConfig)))
       case None                        => fetchOrGenerate(fingerprintFileName)
     }
   }
 
-  private def fetchOrGenerate(fingerprintFileName: FileName): Future[Fingerprint] =
+  private def fetchOrGenerate(fingerprintFileName: FileName): Future[Either[StatisticError, Fingerprint]] =
     for {
-      dbFingerprint <- dbioRunner.run(fingerprintRepository.read())
+      dbFingerprint <- dbioRunner.safeRun(fingerprintRepository.read(), _ => DbError)
```
mk-software-pl, line 33: shouldn't we have a `read` method that returns an `Option[]`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
```diff
@@ -18,6 +20,23 @@ class DBIOActionRunner(dbRef: DbRef) {
...
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    safeRun(action.transactionally, error)
+
+  def safeRun[T, E](action: DB[T], error: Throwable => E)(
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    run(action).transformWith {
+      case Failure(ex) => {
+        logger.warn("Exception occurred during database access", ex)
```
mk-software-pl, line 34: I don't think we need this error. We can handle the exception at the business level somehow and the exception in the log will be misleading

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
```diff
@@ -18,6 +20,23 @@ class DBIOActionRunner(dbRef: DbRef) {
...
+  def safeRunInTransaction[T, E](action: DB[T], error: Throwable => E)(
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    safeRun(action.transactionally, error)
+
+  def safeRun[T, E](action: DB[T], error: Throwable => E)(
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    run(action).transformWith {
+      case Failure(ex) => {
```
mk-software-pl, line 33: non-fatal only?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
```diff
@@ -1,42 +1,37 @@
...
-              logger.warn(errorMessage, e)
-              businessError(InvalidURL)
-            }
-            case e: SQLException => {
-              logger.warn(errorMessage, e)
-              businessError(DbError)
-            }
+          .map {
+            case Left(error) =>
+              error match {
```
mk-software-pl, line 30: can we have this flattened?
Like this:
```
case Left(DbError) => 
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -49,15 +57,21 @@ object UsageStatisticsReportsSettingsDeterminer {
...
+              ScenarioStatisticsInputData(
+                scenario.isFragment,
+                scenario.processingMode,
+                deploymentManagerTypeByProcessingType(scenario.processingType),
+                scenario.state.map(_.status)
+              )
             )
           )
         }
+        .recoverWith { case ex: SQLException =>
```
mk-software-pl, line 71: IMO at this level we should not deal with SQLExceptions explicitly. IMO this case should be handled at the repository implementation level.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
```diff
@@ -1,42 +1,37 @@
...
-  private val endpoints    = new StatisticsApiEndpoints(authenticator.authenticationMethod())
-  private val errorMessage = "Statistics URL construction failed"
+  private val endpoints = new StatisticsApiEndpoints(authenticator.authenticationMethod())
 
   expose {
     endpoints.statisticUsageEndpoint
       .serverSecurityLogic(authorizeKnownUser[StatisticError])
       .serverLogic { _ => _ =>
         determiner
           .determineStatisticsUrl()
```
mk-software-pl, line 27: I'm not sure if `determine` is still the best choice.  WDYT about `prepare` instead?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/FingerprintService.scala
```diff
@@ -16,26 +16,30 @@ class FingerprintService(dbioRunner: DBIOActionRunner, fingerprintRepository: Fi
     implicit ec: ExecutionContext
 ) extends LazyLogging {
 
-  def fingerprint(config: UsageStatisticsReportsConfig, fingerprintFileName: FileName): Future[Fingerprint] = {
+  def fingerprint(
+      config: UsageStatisticsReportsConfig,
+      fingerprintFileName: FileName
+  ): Future[Either[StatisticError, Fingerprint]] = {
```
mk-software-pl, line 22: from the perspective of FingerprintService and its `fingerprint` method, we return a Fingerprint obj or information that the fingerprint cannot be generated (and IMO there is not need to say it's because of db error - it doesn't matter from the perspective of the client who uses it). So I'd not use `StatisticError` here

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/StatisticError.scala
```diff
@@ -0,0 +1,5 @@
+package pl.touk.nussknacker.ui.statistics
+
+sealed trait StatisticError
+case object DbError  extends StatisticError
```
mk-software-pl, line 4: IMO we have two cases here:
1. we cannot provide statistics because we don't have persistent fingerprint 
2. we cannot provide statistics because there was a problem with URL creation

One is related to some unexpected DB errors, and the latter is related to some development bug (is there an expected case where the URL cannot be created?). So, maybe have only one StatisticError implementation, like "CannotGenerateStatistics" (why? it's not sth that the layer above is interested in, but in can be logged, obviously)

nussknacker-dist/src/universal/conf/logback.xml
```diff
@@ -51,7 +51,7 @@
         <appender-ref ref="ERROR-FILE"/>
     </root>
 
-    <logger name="pl.touk.nussknacker" level="${NUSSKNACKER_LOG_LEVEL:-INFO}"/>
+    <logger name="pl.touk.nussknacker" level="${NUSSKNACKER_LOG_LEVEL:-DEBUG}"/>
```
mk-software-pl, line 54: let's revert to INFO

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/StatisticError.scala
lukasz-bigorajski: Done

nussknacker-dist/src/universal/conf/logback.xml
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/FingerprintService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
lukasz-bigorajski: I agree

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: It's implemented deep in the old repository, should I provide alternative implementation with the error handling or for this case we can leave as it is and handle the exception here or maybe should we ignore that some exceptions can bubble in Future.failed?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/FingerprintService.scala
lukasz-bigorajski: we have and in the second parameter we provide the mapping function from exception to domain

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/StatisticError.scala
mk-software-pl: ok, great, but WDYT about sth like that?
```scala
sealed trait StatisticError
object CannotGenerateStatisticsError extends StatisticError
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
```diff
@@ -18,6 +19,20 @@ class DBIOActionRunner(dbRef: DbRef) {
...
+  def safeRunInTransaction[T, E](action: DB[T])(error: Throwable => E)(
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    safeRun(action.transactionally)(error)
+
+  def safeRun[E, T](action: DB[T])(error: Throwable => E)(
+      implicit ec: ExecutionContext
+  ): Future[Either[E, T]] =
+    run(action.map(t => Right(t)))
+      .recover { case ex: SQLException =>
```
mk-software-pl, line 32: TBH we're cheating a little bit here. We catch `SQLException` but in the `error` conversion we have a `Throwable`. We do hide the SqlException, but in the end, we give the user a way to do something with it. 

Ideally, there should be no SQLException in the domain/core layer. It should be encapsulated in the repository. Maybe we can try to do it?

Currently we use in to run this code:
```scala
    val dbResult = for {
      dbFingerprint <- fingerprintRepository.read()
      result <- dbFingerprint match {
        case Some(dbValue) => DBIO.successful(new Fingerprint(dbValue))
        case None =>
          val generated = readFingerprintFromFile(fingerprintFileName).getOrElse(randomFingerprint)
          logger.info(s"Generated fingerprint $generated")
          fingerprintRepository
            .write(generated)
            .map(_ => new Fingerprint(generated))
      }
    } yield result
```

Are we able to add a new method to the `FingerprintRepository` interface? 
I'd see its declaration like this:
```scala
def readOrSave(freshFingerprint: Fingerprint): F[Either[Error, Fingerprint]]

sealed trait Error
object Error {
  case object SaveError extends Error
}
```

The code I show at the beginning (after some tweaks) would be its implementation. The SQLException will be caught in the repository implementation. At the core level, we should not bother of the underlying exceptions and just bubble them up. The error, we are interested in, is passed in the Left. 

WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
```diff
@@ -36,9 +44,13 @@ object UsageStatisticsReportsSettingsDeterminer {
       processService: ProcessService,
       // TODO: Instead of passing deploymentManagerTypes next to processService, we should split domain ScenarioWithDetails from DTOs - see comment in ScenarioWithDetails
       deploymentManagerTypes: ProcessingTypeDataProvider[DeploymentManagerType, _],
-      fingerprintSupplier: (UsageStatisticsReportsConfig, FileName) => Future[Fingerprint]
+      fingerprintSupplier: (
```
mk-software-pl, line 47: WDYT about refactoring it a little bit?
In this case, the supplier is nothing but FignerprintService. Why cannot we use it here?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
mk-software-pl: IMO there is no sense to consider all possible problems here. We should only consider the things we can predict. For eg. no scenarios in the returned list. But I don't think we should bother that eg. there won't be a connection to handle our case. We cannot do much with it. So, it's better to forget all these things and let the upper-level handler do its job (here: caching the exception, logging, and returning HTTP 500 - I guess these things are done now ;))

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminer.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBIOActionRunner.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/FingerprintService.scala
```diff
@@ -40,7 +45,10 @@ class FingerprintService(dbioRunner: DBIOActionRunner, fingerprintRepository: Fi
         }
       }
     } yield result
-    dbioRunner.runInTransaction(dbResult)
+    dbioRunner.safeRunInTransaction(dbResult) { ex =>
```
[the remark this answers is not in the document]
lukasz-bigorajski: I've done this by next suggestion. I got rid of the big transaction and I added transaction on new operation readOrSave which is doing double check on DB.

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/StatisticError.scala
lukasz-bigorajski: Done

designer/server/src/test/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminerTest.scala
```diff
@@ -19,26 +22,31 @@ import scala.concurrent.Future
...
-    with OptionValues
+    with EitherValues
     with PatientScalaFutures
     with TableDrivenPropertyChecks {
 
   private val sampleFingerprint = "fooFingerprint"
-  private val fingerprintSupplier = (_: UsageStatisticsReportsConfig, _: FileName) =>
-    Future.successful(new Fingerprint(sampleFingerprint))
+
+  private val fingerprintService: FingerprintService = mock[FingerprintService](
```
mk-software-pl, line 31: maybe `mockedFingerprintService`?

mk-software-pl (approved):

designer/server/src/test/scala/pl/touk/nussknacker/ui/statistics/UsageStatisticsReportsSettingsDeterminerTest.scala
lukasz-bigorajski: Done

PR #5875, opened 2024-04-15
[NU-1597] Add idleness to Kafka source

mslabek:
## Checklist before merge


engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
```diff
@@ -43,11 +43,11 @@ import java.time.Duration
 import java.util.Properties
 import scala.jdk.CollectionConverters._
 
-class FlinkKafkaSource[T](
+abstract class FlinkKafkaSource[T](
```
mslabek, line 46: I did and experiment with making this class abstract and removing `timestampAssignerForTest` and `typeInformation` because it is not instantiated directly anywhere - only as `FlinkConsumerRecordBasedKafkaSource` which overrides these. The tests are passing so I think they were never used.

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -27,8 +27,9 @@ case class KafkaConfig(
     schemaRegistryCacheConfig: SchemaRegistryCacheConfig = SchemaRegistryCacheConfig(),
     avroAsJsonSerialization: Option[Boolean] = None,
     kafkaAddress: Option[String] = None,
-    // TODO: remove this feature flag in future release
-    useNamingStrategyForConsumerGroupId: Boolean = true
+    // TODO question: move it into a generic map / move it into nested config class / leave it here?
```
mslabek, line 30: Not sure about this - its a similar concept to maxOutOfOrderness which is configured in `kafkaEspProperties` - but I don't see a reason besides consistency to have this in a generic string map

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -27,8 +27,9 @@ case class KafkaConfig(
...
+    enableIdleTimeout: Boolean = true,
+    idleTimeout: FiniteDuration = 3 minutes
```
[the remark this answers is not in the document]
mslabek: I did it like that at first, but I want to have:
- a default timeout that is turned on if someone leaves it empty
- a way to disable the idleness entirely

Is it possible to do that with one optional field? For `outOfOrderness` it may work ok because I guess you can set the outOfOrdeness to 0 to effectively disable that, but here I think we should have a way to disable setting idleness with a default.

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -49,6 +60,12 @@ case class KafkaConfig(
 
 }
 
+private object KafkaEspPropertiesConfig {
```
lciolecki, line 63: Please also change the name to KafkaEspPropertiesConfig? We don't need here Esp ;)

lciolecki: The namespace should be written in a constant style: https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables. I wonder, it should also be public / available to external projects?

docs/integration/KafkaIntegration.md
```diff
@@ -276,13 +275,15 @@ Important thing to remember is that Kafka server addresses/schema registry addre
...
+|---------------------------------------------------|------------|----------------------------|--------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| kafkaEspProperties.defaultMaxOutOfOrdernessMillis | Medium     | long                       | 60000 (60 seconds) | Configuration of [bounded out of orderness watermark generator](https://ci.apache.org/projects/flink/flink-docs-stable/docs/dev/datastream/event-time/built_in/#fixed-amount-of-lateness) used by Kafka sources |
+| idlenessConfig.enableIdleTimeout                  | Medium     | boolean                    | true               | Enabling [idleness](https://nightlies.apache.org/flink/flink-docs-stable/docs/connectors/datastream/kafka/#idleness) used by Kafka sources                                                                      |
```
arkadius, line 283: One more polishing that we can do. We can rename it to `idleTimeout.enabled` and `idleTimeout.duration`. WDYT?

arkadius (approved):
Good job

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
mslabek: I didn't want to name it KafkaPropertiesConfig not to potentially confuse someone that this is set in that map (kafkaProperies is not kafkaEspProperies). WDYT about just moving it to KafkaConfig as public constants?

docs/integration/KafkaIntegration.md
mslabek: Done :+1:

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/KafkaConfig.scala
```diff
@@ -60,9 +73,11 @@ object KafkaConfig {
   import net.ceedubs.ficus.Ficus._
   import net.ceedubs.ficus.readers.ArbitraryTypeReader._
   import net.ceedubs.ficus.readers.EnumerationReader._
-  import TopicsExistenceValidationConfig._
 
-  val defaultGlobalKafkaConfigPath = "kafka"
+  val defaultGlobalKafkaConfigPath                              = "kafka"
```
lciolecki, line 77: What about this name?

mslabek: Done :+1:

lciolecki (approved):

PR #5881, opened 2024-04-16
Interpreter module renamed to scenario-compiler

arkadius:
## Checklist before merge


pielas:
Maybe it would be good to add a comment like `Due to legacy reasons, this class is named after an Interpreter. However, in new terminology, this is a Scenario Compiler` in places like this https://github.com/TouK/nussknacker/blob/8675b11208669323dc61da70df4a0fc0fb5de2de/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala#L24

arkadius:
> Maybe it would be good to add a comment like `Due to legacy reasons, this class is named after an Interpreter. However, in new terminology, this is a Scenario Compiler` in places like this
> 
> https://github.com/TouK/nussknacker/blob/8675b11208669323dc61da70df4a0fc0fb5de2de/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala#L24
> 
> WDYT?

This class is the the interpreter. Take a look that the public method `interpret` takes node and `Context` (of variables) and returns the new `Context` (of variables). 

The big picture is that both the `designer` module and runtime modules use `ProcessCompiler` - it is the entrypoint class in the new `scenario-compiler` module. The compiler as a compilation result returns parts of the scenario graph that are interpreted only on the `runtime` side. So in fact, we have the `interpreter`. With a little bit of effort we could move the `interpter` into the separate module that would be used only by `runtime` modules. I can add a TODO for that. WDYT?

pielas:
> > Maybe it would be good to add a comment like `Due to legacy reasons, this class is named after an Interpreter. However, in new terminology, this is a Scenario Compiler` in places like this
> > https://github.com/TouK/nussknacker/blob/8675b11208669323dc61da70df4a0fc0fb5de2de/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala#L24
> > 
> > WDYT?
> 
> This class is the the interpreter. Take a look that the public method `interpret` takes node and `Context` (of variables) and returns the new `Context` (of variables).
> 
> The big picture is that both the `designer` module and runtime modules use `ProcessCompiler` - it is the entrypoint class in the new `scenario-compiler` module. The compiler as a compilation result returns parts of the scenario graph that are interpreted only on the `runtime` side. So in fact, we have the `interpreter`. With a little bit of effort we could move the `interpter` into the separate module that would be used only by `runtime` modules. I can add a TODO for that. WDYT?

at first glance I thought that this is compiler. Leaving a TODO sounds good

arkadius:
> > > Maybe it would be good to add a comment like `Due to legacy reasons, this class is named after an Interpreter. However, in new terminology, this is a Scenario Compiler` in places like this
> > > https://github.com/TouK/nussknacker/blob/8675b11208669323dc61da70df4a0fc0fb5de2de/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/Interpreter.scala#L24
> > > 
> > > WDYT?
> > 
> > 
> > This class is the the interpreter. Take a look that the public method `interpret` takes node and `Context` (of variables) and returns the new `Context` (of variables).
> > The big picture is that both the `designer` module and runtime modules use `ProcessCompiler` - it is the entrypoint class in the new `scenario-compiler` module. The compiler as a compilation result returns parts of the scenario graph that are interpreted only on the `runtime` side. So in fact, we have the `interpreter`. With a little bit of effort we could move the `interpter` into the separate module that would be used only by `runtime` modules. I can add a TODO for that. WDYT?
> 
> at first glance I thought that this is compiler. Leaving a TODO sounds good

Ok, done

pielas (approved):
LGTM!

PR #5889, opened 2024-04-17
[NU-1527] decision table params validation improvement

coutoPL:
## Describe your changes

**Before:** 
When the decision-table component's "Decision Table" and "Match Condition" parameters are filled and then the first parameter is changed to a value that doesn't validate, the second parameter disappears. 

**After:**
The second parameter doesn't disappear. 

## Checklist before merge


engine/common/components-tests/src/test/scala/pl/touk/nussknacker/engine/common/components/DecisionTableSpec.scala
```diff
@@ -173,6 +173,20 @@ trait DecisionTableSpec
                     )
                   )
                 )
+              ),
+              ExpressionParserCompilationError(
+                message = "There is no property 'age' in type: Record{}",
```
lciolecki, line 178: Hmm this message for business people can be scary..  Why not:
`
There is no property 'age' in table?
`

mk-software-pl: Agree. But the same message users will see in any component that has SpEL expression ;) I mean, it's not something that is specific in case of the Decision Table component. Probably can be improved, but it's rather related to SpEL validation improvements.

mk-software-pl: or maybe ... I have one idea, how to provide a little improvement in this case inside the Decision Component. Will check

arkadius (approved):
LGTM, one syntax question added

engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
```diff
@@ -83,14 +79,51 @@ object DecisionTable extends EagerService with SingleInputDynamicComponent[Servi
...
+
+  private def allParametersDefined(context: ValidationContext, dependencies: List[NodeDependencyValue])(
+      implicit nodeId: NodeId
+  ): ContextTransformationDefinition = {
+    case TransformationStep(
+          (firstParamName, FailedToDefineParameter) ::
+          (secondParamName, _) :: Nil,
+          _
+        )
+        if firstParamName == BasicDecisionTableParameter.name &&
```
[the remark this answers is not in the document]
mk-software-pl: yeah, it looks better, but the is no way to do it now (when the name is defined in the object)

engine/common/components-tests/src/test/scala/pl/touk/nussknacker/engine/common/components/DecisionTableSpec.scala
mk-software-pl: @lciolecki check it out now ;)

lciolecki (approved):
So supper! :)

engine/common/components/src/main/scala/pl/touk/nussknacker/engine/common/components/DecisionTable.scala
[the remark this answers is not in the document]
arkadius: Ah! What a nasty limitation

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/tabularDataDefinition/TabularDataDefinitionParser.scala
```diff
@@ -97,7 +98,8 @@ object TabularDataDefinitionParser extends ExpressionParser {
                   errorMessage =
                     s"Column has '${coordinates.columnName.aType.getSimpleName}' type but its value cannot be converted to the type."
```
mslabek, line 99: The error message could be better - user just has to know that there is a type mismatch between cell and column, doesn't need to know about conversion.

mslabek (approved):
Besides the error message, LGTM!

PR #5891, opened 2024-04-17
Add decision table documentation

Diamekod0221:
## Describe your changes
Added documentation for usage of decision table in Enrichers.md
Added relevant images and a showcases fragment.

## Checklist before merge


docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+![alt_text](img/decisionTableExpression.png "Expression matching input with rows.")
+
+The output from the component will be a list of matched *whole* rows (of type record). So, in our case, if one of inputs is (True, True, "myCar"), then the output would be a list consisting of a single recordList[(True, True, True)]. If you want to only get the _result columns_, then you need to transform the data afterwards with a different component.
+
+You can add as many input and result columns as you would like. The _Match condition_ is a fully functioning expression, so you can be quite flexible with it. You could omit some of the checks on the input rows. In our example, we could check for engine control only with *#engine = #ROW.Engine_starts*. It is also up to you which columns you define as input and which as output, you can treat them completely differently in two components that have the same table definition.
+
+[Download showcase fragment](showcases/decisionTableShowcase.json)
```
lciolecki, line 65: Why fragment, not just tghe scenario?

Diamekod0221: Idk, I just wrote a fragment for this...

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
 
 ![alt_text](img/databaseLookupEnricher.png "databaseLookup Enricher")
 
+### DecisionTable enricher
+
+This component allows the user to match input to a pre-defined result. Please be wary that this component works a bit different than a classical decision table([Decision tables in Wikipedia](https://en.wikipedia.org/wiki/Decision_table)). It consists of the fields:
+- Name: the name of the component.
```
mk-software-pl, line 47: is it a component name or rather a node name?

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+- Decision Table: a user defined table of type TabularTypedData, discussed below.
```
mk-software-pl, line 48: TBH `TabularTypedData` is our internal type which leaked into UI (I reported this fact but at the moment there is a decision to not bother with this). We should assume that users don't have to know (and the do NOT know) what the type is. 

Let's describe that:
1. the table could have many columns
2. each column has a type
3. the cells are validated
4. no value in a cell means `null`

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+- Match condition: a Boolean SpEL expression that matches input to table rows, discussed below.
```
mk-software-pl, line 49: 1. `matches input to table rows` - I don't think it's possible to grasp the idea without checking the example. 
I'd rather say that for each row Nu evaluates the expression and based on the result (true/false) the currently analyzed row will be picked or not.

2. `discussed below.` - IMO we should deeply explain what the params offer. The example is an addition. Not a main part of the description of the component. 

3. Please, write a little bit about the `#ROW` variable. Explaining what's this.

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+- Output: name of the output variable which will hold data generated by the component.
```
mk-software-pl, line 50: let's write explicitly more about the "data generated". What's the type of the data? What does the data hold?

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+- Description: user-defined description of the particular component.
```
mk-software-pl, line 51: the same here: component or node?

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+### DecisionTable enricher
+
+This component allows the user to match input to a pre-defined result. Please be wary that this component works a bit different than a classical decision table([Decision tables in Wikipedia](https://en.wikipedia.org/wiki/Decision_table)). It consists of the fields:
+- Name: the name of the component.
+- Decision Table: a user defined table of type TabularTypedData, discussed below.
+- Match condition: a Boolean SpEL expression that matches input to table rows, discussed below.
+- Output: name of the output variable which will hold data generated by the component.
+- Description: user-defined description of the particular component. 
+
+The component has two main elements: _Decision Table_ which is a user-defined table of values and _Match condition_ - a SpEL expression which Nu will use to match the input to the table values.
```
mk-software-pl, line 53: I'd split the component doc into two parts: the description and the example.
In the example part it'd be great if we could say something about:

1. node input (showing the example of the processing input)
2. the decision table node configuration:
  a) what we have in the table
  b) what we have in the match condition
3. node output (showing the example of the output)

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,6 +41,28 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
...
+ 
+First we have to define a fixed-value decision table. We will use a simple car mechanic concept - a decision table that helps to define whether a car should be serviced. The blue columns represent our input values and the red one is our decision column.
```
mk-software-pl, line 55: We make the distinction that some columns represent input and some decisions, but in case of further description, it isn't important. Because the current implementation of the Decision Table component isn't really a decision table (sic!). I don't think we need to do the distinction. If we do, we should show the example in the wider context. Eg. w the next node in which we could process the decision table node output by taking the first element of the list and extracting the last column (accessing the proper record field) that we interpret as a decision (but at this stage, not in the decision table node).

docs/scenarios_authoring/Enrichers.md
mk-software-pl: IMO it's better to have it as a scenario. 

But I see a much more important problem with the example. It's great to have it, but on the other hand, it's very dangerous because for sure we miss the moment when the scenario will be broken (because of reasons ... eg. JSON format changes). It'd be nice to have it tested. Eg. we could have a mechanism that takes all the example scenarios used in the docs and tries to import them (and checks if there are no errors). Without a mechanism like that, the example will be outdated (soon or later), and the users' feeling about the doc will be worse than without the example.

mk-software-pl (changes requested):

docs/scenarios_authoring/Enrichers.md
Diamekod0221: We can have the scenarios imported on staging and scrape the website for correctness or have some check like you said. At the very least we can do "example scenario - designed for version x.y". Then the user will be prepared that it can be outdated and a slightly broken one will still be way better than nothing in my opinion. It doesn't have to be perfect.

docs/scenarios_authoring/Enrichers.md
Diamekod0221: Name of the specific component, which is also a node in the graph.

docs/scenarios_authoring/Enrichers.md
Diamekod0221: 1. I would rephrase it as `will be used to match input to table rows `. I feel that the phrasing you proposed is too complex.

2. I don't think it is a good idea to try and explain these two (decision table, match condition) complex fields in a quick summary. That is precisely why i have extracted them. 

3. Sure, I forgot about that

docs/scenarios_authoring/Enrichers.md
Diamekod0221: Tbh this has just been added as a remark about flexibility. I think further examples are unnecessary, maybe we should remove the mention of this altogether.

docs/scenarios_authoring/Enrichers.md
mk-software-pl: 2. I'd insist on splitting the doc into two parts: the explanation and the example. The example is just an addition. An example should not be used as an explanation for a concept.

docs/scenarios_authoring/Enrichers.md
Diamekod0221: The documentation you refer to "A node is a component used in a scenario". I can change it to "name of the particular node" if you want to be specific about the fact that this is the name of node instance.

docs/scenarios_authoring/Enrichers.md
Diamekod0221: I have added an input screenshot. I'm trying to add an output screenshot but staging keeps dying on ad hoc testing ;_;

docs/scenarios_authoring/Enrichers.md
arkadius: Name and description are parameters common for every component. WDYT, we should describe these parameters for every component type or we should describe it in one place?

docs/scenarios_authoring/Enrichers.md
Diamekod0221: Ok, it's done

docs/scenarios_authoring/Enrichers.md
Diamekod0221: I have added a quick field summary at the top. I don't agree with your " The example is just an addition. ". This is a pretty straightforward, well-known case of functionality. We should present how it works in our software. An example is the perfect way. I also don't see how we can do without it? If the current version finds out to be too little, please argument your idea

mk-software-pl (changes requested):
The example is great. But I have a few small comments.

mk-software-pl (approved):

docs/scenarios_authoring/Enrichers.md
```diff
@@ -41,12 +41,111 @@ DatabaseLookupEnricher is a specialized look-up component; it returns all column
 
 ![alt_text](img/databaseLookupEnricher.png "databaseLookup Enricher")
 
+### Decision Table
```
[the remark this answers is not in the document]
mk-software-pl: But currently, it's in enrichers in the creator panel.
Screenshot 2024-05-24 at 13 16 19

The Product Team agrees that it's not an enricher and it will be changed in the future, but for now, let's have it here and be consistent with the Nu Designer UI.

PR #5910, opened 2024-04-19
Use ScenarioPropertiesConfigFinalizer in NewProcessPreparer and ScenarioPropertiesValidatior

wrzontek:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/ScenarioPropertiesConfigFinalizer.scala
```diff
@@ -12,7 +12,12 @@ class ScenarioPropertiesConfigFinalizer(
   def finalizeScenarioProperties(
       scenarioPropertiesConfig: Map[String, ScenarioPropertyConfig],
   ): Map[String, ScenarioPropertyConfig] = {
-    additionalUIConfigProvider.getScenarioPropertiesUIConfigs(processingType) |+| scenarioPropertiesConfig
+    additionalUIConfigProvider
+      .getScenarioPropertiesUIConfigs(processingType)
+      .filter { case (propertyName, _) =>
+        // configs from additionalUIConfigProvider should only override existing configs, not create new ones
```
arkadius, line 18: Why?

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -200,7 +200,8 @@ class AkkaHttpBasedRouteProvider(
       val newProcessPreparer = processingTypeDataProvider.mapValues { processingTypeData =>
         new NewProcessPreparer(
           processingTypeData.deploymentData.metaDataInitializer,
-          processingTypeData.deploymentData.scenarioPropertiesConfig
+          processingTypeData.deploymentData.scenarioPropertiesConfig,
```
arkadius, line 203: Why can't we just somehow merge `deploymentData.scenarioPropertiesConfig` with these from `additionalUIConfigProvider` at the very beginning of the process in the place where they are already merged with properties from config - the same as we do with component's parameters?

wrzontek: I guess we could, but this way changes to the return value of `getScenarioPropertiesUIConfigs` take effect immediately, without requiring model reload.

wrzontek: wait, let me check some things

wrzontek: i've managed to rather easily get rid of this limitation

>   // `ScenarioPropertyConfig.validators` currently does nothing (only usage goes to createUIScenarioPropertyConfig)

including them in ScenarioPropertiesValidator pretty much makes them equal to `normal` scenario properties, so i've also removed the filter from the thread above

arkadius (approved):

ForrestFairy (approved):
Let's merge it

wrzontek:
nondeterministic cypress tests failure
check out the cypress PR's, they change the same file back and forth

not fault of this PR so i'll merge it

PR #5914, opened 2024-04-20
[Nu-1474] Add Table API primitive types

mslabek:
## Checklist before merge


build.sbt
```diff
@@ -1758,7 +1758,8 @@ lazy val flinkBaseComponentsTests = (project in flink("components/base-tests"))
     name := "nussknacker-flink-base-components-tests",
     libraryDependencies ++= Seq(
       "org.apache.flink" % "flink-connector-files" % flinkV % Test,
-      "org.apache.flink" % "flink-csv"             % flinkV % Test
+      "org.apache.flink" % "flink-csv"             % flinkV % Test,
+      "org.apache.flink" % "flink-json"            % flinkV % Test
```
mslabek, line 1762: added this to have a more readable test input file instead of csv

nussknacker-dist/src/universal/conf/tables-definition.sql
```diff
@@ -1,17 +1,40 @@
...
     client_id STRING,
-    amount INT,
-    datetime DATE
+    amount    INT,
+    datetime  DATE
+) WITH (
+      'connector' = 'datagen'
+);
+
+CREATE TABLE all_types_generated
```
mslabek, line 19: added this for manual testing

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/TableTypeConversions.scala
```diff
@@ -15,18 +17,39 @@ object TableTypeConversions {
...
+
+    case klass if klass == classOf[java.time.LocalDate]     => Some(DataTypes.DATE)
+    case klass if klass == classOf[java.time.LocalTime]     => Some(DataTypes.TIME)
+    case klass if klass == classOf[java.time.LocalDateTime] => Some(DataTypes.TIMESTAMP)
+    case klass if klass == classOf[java.time.Instant]       => Some(DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE)
+
+    case _ => None
   }
 
+  private val DecimalTypeWithDefaultPrecisionAndScale = DataTypes.DECIMAL(10, 0)
```
mslabek, line 53: This is an assumed hardcoded value because we don't have (?) a way to pass information about things like BigDecimals precision and scale in our type system. We could either:
1. (I'm guessing here) Modify our `TypedClass` to carry more information about types
2. Make a 'spacious' default assumption - for example DECIMAL(38,8) so it would be good enough for most purposes. The problem is that when someone uses a decimal with 9 digits after decimal point we don't throw exception, Flink doesn't raise any exceptions, logs - we just get a null value :(

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/TableTypeConversions.scala
```diff
@@ -14,19 +16,39 @@ object TableTypeConversions {
...
+
+    case klass if klass == classOf[java.time.LocalDate]     => Some(DataTypes.DATE)
+    case klass if klass == classOf[java.time.LocalTime]     => Some(DataTypes.TIME)
+    case klass if klass == classOf[java.time.LocalDateTime] => Some(DataTypes.TIMESTAMP)
+    case klass if klass == classOf[java.time.Instant]       => Some(DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE)
+
+    case _ => None
   }
 
+  private val DecimalTypeWithDefaultPrecisionAndScale = DataTypes.DECIMAL(38, 8)
```
arkadius, line 52: 1. Let's write an explanation, that 38 is max and 8 is arbitrary chosen reasonable scale 
2. Can you show in test this case when null is produced?

mslabek: done :+1: I wrote a test for when null was getting produced when I was testing manually but it behaves normally - it rounds the value to the default scale. I'll try to reproduce this tomorrow again

mslabek: Couldn't reproduce it now - I'm guessing it was a problem with test data or bad configuration of tables

engine/flink/components/base-tests/src/test/resources/tables/primitives/primitives.json
```diff
@@ -0,0 +1 @@
+{"string":"bd23cc017c0927ae2a7769177eb0cf2ae6aa2d01013e142170f61c5659abd37665bb44dc47c525750673a80a6d48fcd9e665","boolean":true,"tinyInt":-1,"smallInt":-11796,"int":1781149910,"bigint":6740817575276386117,"float":2.0957512E38,"double":6.366705250892882E307,"decimal":2312106163,"date":"2024-04-20","time":"02:26:40","timestamp":"2024-04-20 00:26:40.604","timestampLtz":"2024-04-20 00:26:40.604Z"}
```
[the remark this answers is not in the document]
mslabek: Its indirectly used by a `should do file-to-file ping-pong for all primitive types` test in `TableFileSinkTest` by the `input` table. I generated it using the datagen connector

nussknacker-dist/src/universal/conf/tables-definition.sql
mslabek: Done :+1: I had to add it in 2 places - also in `packageZipTarball` because it wasn't filtering when the filter was only in `Universal / mappings`, so I extracted the filter as a separate function and reused it

arkadius (approved):

PR #5943, opened 2024-04-24
Fix always enabled generate button

mslabek:
## Checklist before merge


Dzuming (approved):

lciolecki:
When shouldn't this button be available?

mslabek:
@lciolecki FE asks  `http://localhost:8080/api/testInfo/{scenario-name}/capabilities` with current scenario graph and BE responds with info about what kind of test-like-actions can be done. For example:
```json
{"canBeTested":false,"canGenerateTestData":false,"canTestWithForm":false}
```.
So if `canGenerateTestData` is true then it should be enabled. As to how its determined on BE its a long story :smile:

PR #5968, opened 2024-04-25
[NU-1520] Add register statistics endpoint

lukasz-bigorajski:
## Checklist before merge


docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -3350,6 +3402,68 @@ components:
       type: object
     SqlParameterEditor:
       type: object
+    StatisticDto:
+      required:
+      - name
+      - value
+      type: object
+      properties:
```
Dzuming, line 3410: Maybe it's worth, collecting the context of the event, for example, a user on the Firefox browser clicks x times into a button, and on the Safari browser y times and extends it for the NU version. So on version 1.14, we had x clicks, and on version 1.15 y clicks. Also, maybe it's worth having device context, for type (touch device, laptop device), and resolution. 

I know it's a lot of additional data, but IMO, if we are not able to collect them because of query string params length limit, I recon, we should consider changing the approach and find a way to send them, via body as POST method

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -3350,6 +3402,68 @@ components:
...
+        name:
+          $ref: '#/components/schemas/StatisticName'
+        value:
+          type: integer
+          format: int64
+    StatisticName:
+      type: string
+      enum:
```
Dzuming, line 3418: I'm wondering, if for readability's sake, shouldn't add an event type prefix, in all names, 

`SELECT_SCENARIO_FILTER`
`CLICK_EDIT_UNDO`
`CLICK_TEST_GENERATED` -> because someone can open the modal, but for some reason doesn't generate the file. The other thing is a statistic, how many user opens the generate file dialog and don't generate them?

Also, if we have `SCENARIO_FILTER` which is about selecting the filter, does it make sense to send an unselected filter statistic?

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -3350,6 +3402,68 @@ components:
...
       type: object
+    StatisticDto:
+      required:
+      - name
+      - value
+      type: object
+      properties:
+        name:
+          $ref: '#/components/schemas/StatisticName'
+        value:
```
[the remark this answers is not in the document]
lukasz-bigorajski: I meant count of the events in case we want to report more than one event of given type at once. So I will rename this field

docs-internal/api/nu-designer-openapi.yaml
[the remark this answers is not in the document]
lukasz-bigorajski: What kind of eventId? I think we need only need the event name for now because the key in DB will be the name and the current date.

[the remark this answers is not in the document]
lukasz-bigorajski: We don't plan sessions so the distinction between users is needles.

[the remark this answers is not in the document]
Dzuming: I meant some kind of unique event identifier. If we will use the key from a DB it's ok.

docs-internal/api/nu-designer-openapi.yaml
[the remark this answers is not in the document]
Dzuming: ok, but doesn't make it as a list in such a case?

docs-internal/api/nu-designer-openapi.yaml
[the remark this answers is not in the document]
Dzuming: > We don't plan sessions so the distinction between users is needles.

I'm not sure if I get it. But we still have multiple users right? So if we will be able to recognize some patterns in case of specific anonymous user, it can be somehow useful, and in the future if the pattern changed, we can ask the question to yourself, what happened.

lukasz-bigorajski: Maybe we should talk about it because it looks like a bigger task? Maybe statistics with variety of devices and resolution will be sufficient? Because if we add next dimension like a device and user context then we will have sessions and if we want implement session then solution like the matomo has big sense.

docs-internal/api/nu-designer-openapi.yaml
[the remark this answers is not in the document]
Dzuming: Another question, so the timestamp will be added during DB save, right?

docs-internal/api/nu-designer-openapi.yaml
lukasz-bigorajski: Yes, I agree. I will change it

docs-internal/api/nu-designer-openapi.yaml
Dzuming: We discussed it internally, and for now, since we won't have a session user concept, we will just collect a count of the events. If we want to get the context like browser, resolution, etc. we will introduce another event type, which will be sent initially on the page open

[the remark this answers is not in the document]
Dzuming: We discussed it internally, and since there won't be a session context, there is no sense in providing this kind of information. We will just collect a count of specific events.

Dzuming (approved):
LGTM!

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -1801,6 +1801,51 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RegisterStatisticsRequestDto'
+        required: true
+      responses:
+        '204':
+          description: ''
+        '400':
```
mk-software-pl, line 1819: if the response doesn't carry any business value, IMO, it should be skipped and not shown in the API spec

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
```diff
@@ -73,6 +84,72 @@ object StatisticsApiEndpoints {
...
 
+    @derive(encoder, decoder, schema)
+    final case class RegisterStatisticsRequestDto private (statistics: List[StatisticDto])
+
+    @derive(encoder, decoder, schema)
+    final case class StatisticDto private (name: StatisticName)
+
+    sealed trait StatisticName extends EnumEntry with UpperSnakecase
+
+    object StatisticName extends Enum[StatisticName] with CirceEnum[StatisticName] {
```
mk-software-pl, line 95: the name of the statistic is part of our API now. Changing the name implicates changing the API (the compatibility break). Maybe we should have a test that shows us that someone changed the name of any of these objects (and as a result, they changed the API)?

I see two solutions:
1. business spec in which we use every possible statistic name (in the future, new statistic names won't be included and the developer who is going to extend them, probably will miss adding them to the test)
2. `NuDesignerApiAvailableToExposeYamlSpec` - we can add a test which gets all names from the yaml and checks if the set looks exactly as we want it to look like

docs-internal/api/nu-designer-openapi.yaml
lukasz-bigorajski: It's a default mapping in Tapir if the jsonBody is used as the input.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/StatisticsApiEndpoints.scala
lukasz-bigorajski: Done

mk-software-pl (approved):
LGTM

PR #5980, opened 2024-04-26
fix fragment flaky test

Dzuming:
## Checklist before merge


designer/client/cypress/e2e/fragment.cy.ts
```diff
@@ -293,8 +297,11 @@ describe("Fragment", () => {
             .find("[data-testid='form-helper-text']")
             .should("be.visible");
 
+        cy.get("[data-testid=window]").find("section").scrollTo("top");
+        cy.get("[data-testid=window]").find('[data-testid="content-size"]').matchImage();
+
         cy.get("[data-testid=window]").find("section").scrollTo("bottom");
-        cy.get("[data-testid=window]").matchImage({ maxDiffThreshold: 0.01 });
```
ForrestFairy, line 297: We don't need threshold anymore in any place?

ForrestFairy (approved):
LGTM, just one question to better understand

designer/client/cypress/e2e/fragment.cy.ts
Dzuming: We have a default threshold defined in the cypress config `designer/client/cypress.config.ts` which should be enough 

```
pluginVisualRegressionMaxDiffThreshold: 0.007,
```

AFAIR increasing this value, in this case, was a try to remove flakiness from this test some time ago, which failed   ;)

PR #6003, opened 2024-05-07
[Nu-1642] search changes with zoom and moving a view

Dzuming:
## Describe your changes
- We're foregoing automatic zooming after searching. The search itself will not affect the zoom level in the scenario view.
- Upon hovering over a search result, we'll only highlight the node without changing the zoom level.
- Upon clicking a search result, we'll center that node in the view and zoom in if it's zoomed out beyond a certain threshold (meaning if someone has a very large scenario in view, we'll zoom in a bit; if they're zoomed in on a few nodes, we won't adjust it).
## Checklist before merge


designer/client/src/components/graph/Graph.tsx
```diff
@@ -281,9 +281,12 @@ export class Graph extends React.Component<Props> {
         this.panAndZoom.fitContent(area, this.viewport);
     }, 250);
 
-    fitToNodes = (nodeIds: NodeId[] = []): void => {
-        const cells = this.graph.getCells().filter((c) => nodeIds.includes(c.id as string));
-        this.fit(cells);
+    fitToNode = (nodeId: NodeId): void => {
+        const cellsToFit = this.graph.getCells().filter((c) => c.id === nodeId);
+        const area = cellsToFit?.length ? this.graph.getCellsBBox(cellsToFit) : this.processGraphPaper.getContentArea();
```
mslabek, line 286: 1. I think the `cellsToFit?` null check is redundant since `.filter` will return empty array if it doesnt find anything
2. Isn't it guaranteed that we find exactly one node? If so, we could use `.find` instead of `filter`

designer/client/src/components/graph/Graph.tsx
```diff
@@ -281,9 +281,12 @@ export class Graph extends React.Component<Props> {
...
+
+        const withCurrentZoomValue = this.zoom > 0.63;
```
mslabek, line 288: maybe name this 0.63 and deduplicate it with the 0.63 line after?

designer/client/src/components/graph/PanZoomPlugin.ts
```diff
@@ -84,10 +84,11 @@ export class PanZoomPlugin {
         this._enabled = enabled;
     }
 
-    fitContent(content: g.Rect, updatedViewport?: Viewport): void {
+    fitContent(content: g.Rect, updatedViewport?: Viewport, predefinedScale?: number | undefined): void {
         this.viewport = updatedViewport;
 
-        const scale = Math.min(1, 0.78 / Math.max(content.width / this.viewport.width, content.height / this.viewport.height));
+        const autoScaleValue = Math.min(1, 0.78 / Math.max(content.width / this.viewport.width, content.height / this.viewport.height));
```
mslabek, line 90: also maybe name the 0.78 something?

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Process view should have node search toolbar #0.png
mslabek, line 1: seems like a non-deterministic screenshot? Do you remember was this a problem previously?

mslabek (commented):
Clicked through it - functionality-wise no problems and I agree with the change. Added some technical comments

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Process view should have node search toolbar #0.png
Dzuming: Hmm, why do you think it's non-deterministic :)? We changed the zooming behavior after the search, that's why this screenshot changed.

designer/client/src/components/graph/Graph.tsx
Dzuming: 1. Changed ✔️ 
2. Good idea, maybe it shows our intentions in a clearer way, the only downside is `this.graph.getCellsBBox([cellToFit])` but IMO it's acceptable

designer/client/src/components/graph/Graph.tsx
Dzuming: Good idea, let's get rid off these automagic numbers. Done ✔️

designer/client/src/components/graph/PanZoomPlugin.ts
Dzuming: I'm not sure what is the proper name for this number, I just named it `zoomThreshold`

mslabek (approved):

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Process view should have node search toolbar #0.png
Dzuming: Yeah, I tried to remove this header from the screenshot, but there is no proper selector in jointjs (However in theory we select here, only the nodes, but maybe it's an issue with svg). About the position, that's how it works now:
-  the zooming in value is pretty high, because, there are only 4 nodes
- Since we discussed with a business that, we won't manipulate the zoom value on the search
- After node search and result selection, the view is centered on the node

That's why, we only can see 3 nodes on the screen

PR #6008, opened 2024-05-08
[NU-1520] Save statistics data to quest db

lukasz-bigorajski:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/StatisticsDb.scala
```diff
@@ -0,0 +1,24 @@
...
+
+trait StatisticsDb[F[_]] extends AutoCloseable {
+  def initialize(): Unit
+  def write(statistics: Map[String, Long]): F[Unit]
+  def read(): F[Map[String, String]]
+}
+
+object StatisticsDb {
+
+  def create(): Resource[IO, StatisticsDb[Future]] =
```
coutoPL, line 16: because it's related to the specific implementation (QuestDB), shouldn't we move it to `QuestDb` class?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/StatisticsDb.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import cats.effect.{IO, Resource}
+
+import scala.concurrent.Future
+import scala.language.higherKinds
+
+trait StatisticsDb[F[_]] extends AutoCloseable {
+  def initialize(): Unit
+  def write(statistics: Map[String, Long]): F[Unit]
+  def read(): F[Map[String, String]]
```
coutoPL, line 11: Why do we get `Map[String, Long]` and here return `Map[String, String]`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/StatisticsDb.scala
```diff
@@ -0,0 +1,24 @@
+package pl.touk.nussknacker.ui.db.timeseries
+
+import cats.effect.{IO, Resource}
+
+import scala.concurrent.Future
+import scala.language.higherKinds
+
+trait StatisticsDb[F[_]] extends AutoCloseable {
+  def initialize(): Unit
```
coutoPL, line 9: I think it's DB-specific and we don't have to include it in the interface. There are two many problems with such methods. Like:
1. who can call it? After all, it's public
2. why does it return `Unit`? What about effects?
3. How can I prevent it from being called more than once?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/StatisticsDb.scala
```diff
@@ -0,0 +1,24 @@
+package pl.touk.nussknacker.ui.db.timeseries
+
+import cats.effect.{IO, Resource}
+
+import scala.concurrent.Future
+import scala.language.higherKinds
+
+trait StatisticsDb[F[_]] extends AutoCloseable {
```
coutoPL, line 8: 1. the fact that the DB should be closed is also specific to a given implementation. If we expose the instance using a factory method that returns `Resource`, there is no need to declare it as `AutoCloseable`

2. I wonder if we shouldn't have `StatisticsRepository` interface, instead of the `StatisticsDb`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpService.scala
```diff
@@ -35,7 +39,14 @@ class StatisticsApiHttpService(
...
       .serverSecurityLogic(authorizeKnownUser[Unit])
-      .serverLogicSuccess { _ => _ =>
+      .serverLogicSuccess { _ => request =>
+        // todo change to groupMapReduce in scala 2.13
+        val groupedByName = request.statistics
+          .groupBy(_.name.entryName)
+          .map { case (k, v) =>
+            k -> v.size.toLong
+          }
+        db.write(groupedByName)
```
coutoPL, line 49: we "swallow" the potential exception here. Maybe it would be nice to log it? 
BTW shouldn't we create a private method for this logic?

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureExtensions.scala
```diff
@@ -114,7 +114,7 @@ trait NuRestAssureExtensions {
 
   }
 
-  implicit class ExtractLong[T <: ValidatableResponse](validatableResponse: T) {
+  implicit class ExtractParameter[T <: ValidatableResponse](validatableResponse: T) {
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/StatisticsApiHttpServiceBusinessSpec.scala
```diff
@@ -60,18 +68,145 @@ class StatisticsApiHttpServiceBusinessSpec
...
+          }
+        }
+      case None =>
+        logger.info("Cannot extract query params")
+        false
+    }
+
+    override def describeTo(description: Description): Unit = description.appendValue(queryParamsMatchers)
+
+    private def extractQueryParams(queryParamsPath: String): Option[Map[String, String]] =
```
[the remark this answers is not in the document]
lukasz-bigorajski: None means that there is some error while splitting parameters, if there are no parameters then there will be empty map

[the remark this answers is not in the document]
coutoPL: maybe we can fail fast (we're at the test level, so it looks like we can) and throw an illegal argument ex (with a descriptive message) instead?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/FEStatisticsRepository.scala
```diff
@@ -0,0 +1,9 @@
+package pl.touk.nussknacker.ui.db.timeseries
+
+import scala.concurrent.ExecutionContext
+import scala.language.higherKinds
+
+trait FEStatisticsRepository[F[_]] {
+  def write(statistics: Map[String, Long])(implicit ec: ExecutionContext): F[Unit]
```
coutoPL, line 7: the `ExecutionContext` is strictly related to `Future`. We declare a generic F as an effect, so let's move the execution context to the repository implementation (at the constructor level)

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+  private def recreateTables(): Unit = Try {
+    logger.info("Recreating table")
+    cairoEngine.drop(dropTablesQuery, ctx)
+    QuestDbFEStatisticsRepository.createDirAndConfigureLogging()
+    cairoEngine.ddl(createTableQuery, ctx)
+  }.recover { case ex =>
+    logger.warn("Exception occurred while tables recreate", ex)
+    ex
```
coutoPL, line 116: if we are going to propagate the `ex` we should throw it. If not, we can remove this line.

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+  override def write(statistics: Map[String, Long])
+                    (implicit ec: ExecutionContext): Future[Unit] = withExceptionHandling(
+    () => {
+      statistics.foreach { entry =>
+        val row = statsWalTableWriter.newRow(Os.currentTimeMicros())
+        row.putStr(0, entry._1)
+        row.putLong(1, entry._2)
+        row.append()
+      }
+      statsWalTableWriter.commit()
```
coutoPL, line 53: because of the blocking nature of the QuestDB (AFAIR), we should have a separate execution context to not make any unexpected, unintended influence on the designer. It seems that a fixed-size pool will be ok. But we have to be sure that if the pool exhausts (because of some problem with the QuestDB or our implementation), the errors won't be propagated in Futures returned by write/read methods. I see two potential solutions here:

1.  usage level recover - the code that uses the repository assumes that there can be some problem but the business level decision is to skip it and continue the main logic 
2. decorator level recover - we provide a decorator that is supposed to add the recovered logic around the read & write methods of `FEStatisticsRepository`. It would be better to create the decorator inside the class which uses the repository to explicitly show that ignoring the problems with the repository is a business decision.

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+      buffer.toMap
+    },
+    Map.empty[String, Long]
+  )
+
+  private def close(): Unit = {
+    recordCursorFactory.close()
+    statsWalTableWriter.close()
+    applyWal2TableJob.close()
+    cairoEngine.close()
```
coutoPL, line 83: the engine not created in this class so the closing should be at the outside too

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  private def initialize(): Unit =
+    cairoEngine.ddl(createTableQuery, ctx)
+
+  private def withExceptionHandling[T](dbAction: () => T, onError: => T)(
```
coutoPL, line 89: dbAction param can be declared "by name" too, right?
```scala
dbAction: => T
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  private def createDirAndConfigureLogging(): File = {
+    val nuDir: File = createRootDirIfNotExists()
+    configureLogging(nuDir)
+    nuDir
+  }
+
+  private def createRootDirIfNotExists(): File = {
+    val dir   = Try(Option(System.getProperty("java.io.tmpdir"))).toOption.flatten.getOrElse("/tmp")
```
coutoPL, line 144: WDYT about using better-files here? it simplifies the code like this one

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  private val createTableQuery =
+    "CREATE TABLE IF NOT EXISTS statistics (name string, count long, ts timestamp) TIMESTAMP(ts) PARTITION BY DAY WAL"
+
+  private val selectQuery =
+    s"""
+       |   SELECT name,
+       |          sum(count)
+       |     FROM statistics
```
coutoPL, line 176: FE statistics?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -0,0 +1,194 @@
...
+
+}
+
+object QuestDbFEStatisticsRepository extends LazyLogging {
+
+  def create(): Resource[IO, FEStatisticsRepository[Future]] =
+    Resource.make(
+      acquire = for {
+        // todo move this ExecutorService properties to a configuration
+        executorService <- IO(new ThreadPoolExecutor(2, 4, 60L, TimeUnit.SECONDS, new ArrayBlockingQueue[Runnable](8)))
```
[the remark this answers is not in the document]
lukasz-bigorajski: It's much butter ;)

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
lukasz-bigorajski: Great tool!

coutoPL (approved):
LGTM!

PR #6010, opened 2024-05-08
[NU-1648] Fix ad hoc tests with schemas containing nested fields

jedrz:
## Checklist before merge

### DIFF
--- a/utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/parameters/TestingParametersSupport.scala
+++ b/utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/parameters/TestingParametersSupport.scala
@@ -9,15 +9,15 @@ object TestingParametersSupport {
   def joinWithDelimiter(key1: String, key2: String) = s"$key1$delimiter$key2"
 
   // Used to un-flat map with concat name eg. { a.b -> _ } => { a -> { b -> _ } }. Reverse kinda joinWithDelimiter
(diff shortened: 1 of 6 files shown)


utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/parameters/TestingParametersSupport.scala
```diff
@@ -10,20 +10,39 @@ object TestingParametersSupport {
 
   // Used to un-flat map with concat name eg. { a.b -> _ } => { a -> { b -> _ } }. Reverse kinda joinWithDelimiter
   def unflattenParameters(flatMap: Map[ParameterName, AnyRef]): Map[String, AnyRef] = {
+    unflattenParametersInternal(flatMap.map { case (ParameterName(name), value) => name -> value })
```
philemone, line 13: It probably was me who named it `flatMap` but now I think that it is unfortunate name for scala variable :face_in_clouds: Maybe `flatParameters` / `flatParams` or something like this?

philemone (approved):
nice, LGTM :+1:

utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/parameters/TestingParametersSupport.scala
jedrz: You're right. It's a better name.

coutoPL (approved):
LGTM

PR #6053, opened 2024-05-14
[NU-1649] Add impersonate mechanism

Elmacioro:
## Describe your changes

Currently, when a technical user performs an action on behalf of a business user, they become an author of that action instead of the business user who initiated it.

In these changes I introduced a possibility to impersonate a business user. During mapping from `AuthenticatedUser` to `LoggedUser`, when a technical user has appropriate permission (`Overrideusername`) and `impersonatedUser` field is present, `impersonatedUser` will be mapped as `LoggedUser` and their permissions will be used for authorization.

As a result of these changes new column `impersonated_by` was added to 4 database tables:
- `process_attachments`
- `process_comments`
- `process_actions`
- `processes`

With that it should be possible to extinguish which technical user acted on behalf of a business user for certain operations.

In the next PR I will introduce changes to the process migration mechanism to utilize new impersonate mechanism instead of passing `remoteUsername` values.

## Checklist before merge


docs/Changelog.md
```diff
@@ -4,6 +4,7 @@
 -------------------------
 
 * [#6008](https://github.com/TouK/nussknacker/pull/6008) Add embedded QuestDB as database for FE statistics.
+* [#6053](https://github.com/TouK/nussknacker/pull/6053) Add impersonate mechanism for technical users
```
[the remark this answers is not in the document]
arkadius: @Elmacioro we have to update this https://nussknacker.io/documentation/docs/installation_configuration_guide/DesignerConfiguration/#users-roles-and-permissions documentation. 
BTW please remove mention about `AdminTab` - it is some obsolete garbage.
After updating the docs, I think that this will be exposed to user. Not the configurator, but for the admin who can integrate another application with designer. WDYT @mk-software-pl ?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -66,9 +85,16 @@ final case class CommonUser(
...
+  override def canImpersonate: Boolean = globalPermissions
+    .map(_.toLowerCase)
+    .contains(Permission.OverrideUsername.toString.toLowerCase)
+
 }
 
 final case class AdminUser(id: String, username: String) extends LoggedUser {
   override def can(category: String, permission: Permission): Boolean = true
   override val isAdmin: Boolean                                       = true
+  override val impersonatedBy: Option[LoggedUser]                     = None
```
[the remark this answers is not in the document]
arkadius: Why admin user can't be impersonated? I thought that isAdmin is a kind of wildcard for user that have access to all permissions to all categories.

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
```diff
@@ -153,14 +153,10 @@ class MigrationService(
       .leftMap(MigrationError.from(_))
   }
 
-  private def checkLoggedUserCanOverrideProcess(processId: ProcessId, forwardedUsername: Option[RemoteUserName])(
-      implicit loggedUser: LoggedUser
-  ) = {
+  private def checkLoggedUserCanOverrideProcess()(implicit loggedUser: LoggedUser) = {
```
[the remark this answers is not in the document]
arkadius: It looks like something is totally not ok ;-) . Impersonation is handled at the beginning so here we should check only if LoggedUserCanWriteToProcess

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
mk-software-pl: you may say that admin rights are too powerful to allow for impersonation of such type of user. But if the admin is just a regular user with all permissions, maybe we can allow to do that. But let's check it first

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
mk-software-pl: ok, so it means we can get rid of this method 👍

docs/Changelog.md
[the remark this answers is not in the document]
Elmacioro: I updated the doc and removed the `AdminTab` mention

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
Elmacioro: After some changes I made it possible to impersonate Admin user. 

I introduced a marker trait `RealLoggedUser` (Representing real users without impersonation: `CommonUser` and `AdminUser`) and created `ImpersonatedUser` having references to the impersonated and impersonating `RealLoggedUsers`.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -48,13 +51,29 @@ object LoggedUser {
...
+
 }
 
 final case class CommonUser(
     id: String,
     username: String,
     categoryPermissions: Map[String, Set[Permission]] = Map.empty,
-    globalPermissions: List[GlobalPermission] = Nil
+    globalPermissions: List[GlobalPermission] = Nil,
+    impersonatedBy: Option[LoggedUser] = None
```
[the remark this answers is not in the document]
Elmacioro: After modifying the code after the review i moved `impersonatedBy` to the `ImpersonatedUser` class - it's no longer present for common and admin user.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -2,13 +2,16 @@ package pl.touk.nussknacker.ui.security.api
...
+import pl.touk.nussknacker.security.Permission
 import pl.touk.nussknacker.security.Permission.Permission
 
 sealed trait LoggedUser {
   val id: String
   val username: String
   val isAdmin: Boolean
+  val impersonatedBy: Option[LoggedUser]
   def can(category: String, permission: Permission): Boolean
+  def canImpersonate: Boolean
```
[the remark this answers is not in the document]
Elmacioro: Thank you for the suggestion! I moved the `canImpersonate` check to the extension method and introduced `ImpersonatedUser` class extending `LoggedUser`. It made it easier to check which type of user can impersonate. 

I'm still learing how to do things the "Scala Way" so such suggestions are really helpful, thank you! 😃

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatedToLoggedUserConverter.scala
```diff
@@ -0,0 +1,28 @@
...
+
+final case class ImpersonationError(message: String)
+
+object AuthenticatedToLoggedUserConverter {
+
+  def convertToLoggedUser(
+      authenticatedUser: AuthenticatedUser,
+      rules: List[ConfigRule]
+  ): Either[ImpersonationError, LoggedUser] = {
+    val baseLoggedUser = LoggedUser(authenticatedUser, rules)
```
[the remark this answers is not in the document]
Elmacioro: I removed this class and moved the creation logic to the `LoggedUser` `create` function

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatedUser.scala
```diff
@@ -1,6 +1,19 @@
 package pl.touk.nussknacker.ui.security.api
 
-final case class AuthenticatedUser(id: String, username: String, roles: Set[String])
+private[api] sealed trait UserBasicInfo {
+  val id: String
+  val username: String
+  val roles: Set[String]
+}
+
+final case class ImpersonatedUser(id: String, username: String, roles: Set[String]) extends UserBasicInfo
```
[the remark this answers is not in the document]
Elmacioro: This `ImpersonatedUser` class was supposed to represent the impersonated user data before `AuthenticatedUser` => `LoggedUser` mapping.  After intoducing `ImpersonatedUser` on the `LoggedUser` side i removed it with along with UserBasicInfo as it wasn't too readable

designer/server/src/main/scala/pl/touk/nussknacker/ui/migrations/MigrationService.scala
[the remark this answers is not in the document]
Elmacioro: Oh, snap! Yes, now it's not necessary. After removing it I had to fix two migration tests as they were actually passing by accident as the test user lacked `Overrideusername/Impersonate` permission

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/BaseHttpService.scala
```diff
@@ -54,7 +55,10 @@ abstract class BaseHttpService(
         case Some(user) if user.roles.nonEmpty =>
           // TODO: This is strange that we call authenticator.authenticate and the first thing that we do with the returned user is
           //       creation of another user representation based on authenticator.configuration. Shouldn't we just return the LoggedUser?
-          success(LoggedUser(user, authenticator.configuration.rules))
+          convertToLoggedUser(user, authenticator.configuration.rules) match {
+            case Right(loggedUser) => success(loggedUser)
+            case Left(_)           => securityError(AuthorizationError)
```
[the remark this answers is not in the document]
Elmacioro: I added the actuall error type. Thank you for the Scala insight!

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/CommentEntityFactory.scala
```diff
@@ -75,6 +78,7 @@ trait CommentActions {
             processVersionId = processVersionId,
             content = c.value,
             user = loggedUser.username,
+            impersonatedBy = loggedUser.impersonatedBy.map(_.username),
```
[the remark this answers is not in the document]
Elmacioro: Done, modified entities to use ids

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
arkadius: This structure is much better

arkadius (approved):
LGTM

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/CommentEntityFactory.scala
```diff
@@ -46,6 +48,7 @@ final case class CommentEntityData(
     processVersionId: VersionId,
     content: String,
     user: String,
+    impersonatedBy: Option[String],
```
mk-software-pl, line 51: I see some inconsistency here.
- in the `user` field we have a name of a user

but:
- in the `impersonatedBy` we have an ID of a user
- there is no foreign key in the schema definition
- the name of the fields doesn't suggest it's an ID

extensions-api/src/main/scala/pl/touk/nussknacker/security/Permission.scala
```diff
@@ -4,11 +4,11 @@ import pl.touk.nussknacker.security
...
-  val Read             = Value("Read")
-  val Write            = Value("Write")
-  val Deploy           = Value("Deploy")
-  val Demo             = Value("Demo")
-  val OverrideUsername = Value("OverrideUsername") // Used for migration username forwarding
+  val Read        = Value("Read")
+  val Write       = Value("Write")
+  val Deploy      = Value("Deploy")
+  val Demo        = Value("Demo")
+  val Impersonate = Value("Impersonate") // Used for migration username forwarding
```
mk-software-pl, line 11: It's a backward incompatible change. Shouldn't we mention it in the migration guide?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatedUser.scala
```diff
@@ -1,6 +1,17 @@
 package pl.touk.nussknacker.ui.security.api
 
-final case class AuthenticatedUser(id: String, username: String, roles: Set[String])
+sealed trait BaseAuthenticationUserInfo {
```
mk-software-pl, line 3: do we need it now?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -11,12 +13,33 @@ sealed trait LoggedUser {
...
 object LoggedUser {
 
-  def apply(
+  def create(
       authenticatedUser: AuthenticatedUser,
       rules: List[ConfigRule]
-  ): LoggedUser = {
+  ): Either[CreationError, LoggedUser] = {
+    val loggedUser = LoggedUser(authenticatedUser, rules)
+    authenticatedUser.impersonatedAuthenticationUser match {
```
mk-software-pl, line 25: we can flatten it a little bit:
```scala
    authenticatedUser.impersonatedAuthenticationUser match {
      case None =>
        Right(loggedUser)
      case Some(impersonatedUser) if loggedUser.canImpersonate =>
        Right(
          ImpersonatedUser(
            impersonatedUser = LoggedUser(impersonatedUser, rules),
            impersonatingUser = loggedUser
          )
        )
      case Some(_) =>
        Left(ImpersonationNotAllowed)
    }
```

WDYT?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -48,27 +71,55 @@ object LoggedUser {
...
+final case class AdminUser(id: String, username: String) extends RealLoggedUser {
   override val isAdmin: Boolean                                       = true
+  override def can(category: String, permission: Permission): Boolean = true
+}
+
+final case class ImpersonatedUser(impersonatedUser: RealLoggedUser, impersonatingUser: RealLoggedUser)
+    extends LoggedUser {
+  override val id: String                                             = impersonatedUser.id
+  override val username: String                                       = impersonatedUser.username
+  override val isAdmin: Boolean                                       = impersonatedUser.isAdmin
```
mk-software-pl, line 116: WDYT about boy-scouting here a little bit? AFAIS we don't really need the `isAdmin` field. Maybe we can remove it?

docs/Changelog.md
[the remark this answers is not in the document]
mk-software-pl: But currently, from the users' perspective, there is no impersonation support yet, right? I've checked where `AuthenticatedUser` is instantiated and I didn't find the place where we create an instance with the `impersonatedAuthenticationUser` field != None. You've mentioned the next PR. Maybe we should add this changelog entry in the next PR instead?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
mk-software-pl: In the impersonation solutions, there is often a mechanism that tells if a given user can be impersonated. We don't have it yet (and maybe there is no need to have it). But IMO it is better not allowing to impersonate the admin role. Do you need it in your case? Maybe at the moment we don't have to do it?

docs/Changelog.md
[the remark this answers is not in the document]
Elmacioro: Currently none of the implementations of `AuthenticationResources` (which provides `AuthenticatedUser`) use this mechanism in Nu (it will change with next PR with migrations).
However, somebody integrating another application with the designer can have their own `AuthenticationResources` implementation (for example for SSO) with impersonated user.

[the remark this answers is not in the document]
arkadius: Uhh, so we don't have any end to end test as well? It is not acceptable! :) With regards to our DoD, this should be a part of this change.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
arkadius: As I mentioned before, admin currently is only a user with all permissions to all categories. Our admin doesn't have more power than any other user. IMO we shouldn't add this distinction because we don't have such a requirement  at this moment and it will be confusing in the current roles model - if we want to have such a  distinction, we should add more clarifications in docs around users/roles handling.

docs/Changelog.md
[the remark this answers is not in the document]
arkadius: Also, as we discusses, with a little bit effort we could make this impersonation logic not plugin-specific. Why not to use the same header, no matter which authentication mechanism we pick?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
mk-software-pl: AFAIS the admin role is created by adding a proper flag to the user. I'm not sure if this is equal to the situation where the user has a full set of permissions. Then in the code, we explicitly check in some places if this is the admin role. That's why I see it as a security hole which is better to avoid if there is no use case to impersonate the admin.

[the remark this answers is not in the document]
mk-software-pl: In the `BaseHttpService` class we have the `authorizeAdminUser` method. Do you think it should be replaced by e.g. `authorizerAllPermUser` or just removed? Because it's used in the context of two endpoints: config reload and GET config.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/UserApiEndpoints.scala
```diff
@@ -90,6 +90,21 @@ object DisplayableUser {
         categoryPermissions = Map.empty,
         globalPermissions = Nil
       )
+    case ImpersonatedUser(impersonatedUser, impersonatingUser) =>
+      new DisplayableUser(
+        id = impersonatedUser.id,
+        isAdmin = impersonatedUser.isAdmin,
+        username = impersonatedUser.username,
+        categories = allUserAccessibleCategories.toList.sorted,
+        categoryPermissions = impersonatedUser match {
```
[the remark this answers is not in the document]
Elmacioro: Yeah, that's a good idea. I extracted that to methods.

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/CommentEntityFactory.scala
Elmacioro: Hmmm, the situation seems to be a bit more complex.
User's id is not an ID in a database sense because AFAIK there is no user table with information about users. Instead id is an identity that is passed from `AuthenticationResources` implementation to `AuthenticatedUser` an then to the `LoggedUser`. 

In Nu there seems to be a confusion between user's name and `identity`. It is `username` that is saved under the `user` field although it should probably be also the identity however it would probably introduce difficulties with displaying the user's name in the designer.

Initially `impersonatedBy` was representing impersonating user's name. However, I changed it to represent identity after Arek's comment. @arkadius  WDYT?

extensions-api/src/main/scala/pl/touk/nussknacker/security/Permission.scala
Elmacioro: Right, I mentioned the change in the migration guide

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticatedUser.scala
Elmacioro: True, it's not needed anymore. I removed it

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
Elmacioro: Nice! I modified it to be this way, thank you!

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
Elmacioro: Yes, it doesn't seem to be needed as a field. I removed it.

docs/Changelog.md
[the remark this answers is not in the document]
Elmacioro: Ok, I'll try to modify it so we could use the header independently of AuthenticationResources implementation. It will also make it easier to add an end to end test for this PR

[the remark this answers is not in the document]
Elmacioro: I introduced `AuthenticationManager` that coordinates anonymous access, impersonated access and plugin specific authentication with `AuthenticationResources`. 
The impersonated access should be possible by attaching `Impersonate-User-Data` header by a user with `Impersonate` permission. I added some tapir and akka based tests to check whether user gets impersonated with correct set of permissions.

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/AnonymousAccess.scala
```diff
@@ -0,0 +1,28 @@
...
+  def handleAuthorizationFailedRejection: Directive0 = handleRejections(
+    RejectionHandler
+      .newBuilder()
+      // If the authorization rejection was caused by anonymous access,
+      // we issue the Unauthorized status code with a challenge instead of the Forbidden
+      .handle { case AuthorizationFailedRejection => authenticationResources.authenticate() { _ => reject } }
+      .result()
+  )
+
+  def authenticateWithAnonymousAccess(): Future[Option[AuthenticatedUser]] = anonymousUserRole
```
mk-software-pl, line 22: WDYT about sth like this?
```scala
  def authenticateWithAnonymousAccess(): Future[Option[AuthenticatedUser]] = Future {
    anonymousUserRole.map(role => AuthenticatedUser.createAnonymousUser(Set(role)))
  }
```

and let's add `implicit executionContext: ExecutionContext` to the constructor. When you take a look at the Future.successful method def you will find sth like this:

```scala
final def successful[T](result: T): Future[T]
```
For some reason, it doesn't use `: =>` (by-name param) so the `T` will be evaluated before creating the Future. It means that all exception that will be thrown during evaluation won't be placed in the Future context (as Future.failed). So, this means that `Future.successful` should be used only when you have already evaluated value (or at least some trivial code that won't throw for sure)

In our case, we call the `createAnonymousUser` method. From this place, we cannot assume the code won't throw any exception. Using `Future.apply` solves the problem.

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/AnonymousAccess.scala
```diff
@@ -0,0 +1,28 @@
+package pl.touk.nussknacker.ui.security.accesslogic
```
mk-software-pl, line 1: the class is used only in the context of Authentication Manager only. Maybe we can move it to the the manager class and make it private? WDYT?

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/ImpersonatedAccess.scala
```diff
@@ -0,0 +1,45 @@
...
+        }
+      case None => provide(user)
+    }
+  }
+
+  def authenticateWithImpersonation(
+      impersonatingUserAuthCredentials: PassedAuthCredentials,
+      impersonatedUserData: ImpersonatedUserData
+  )(implicit ec: ExecutionContext): Future[Option[AuthenticatedUser]] = {
+    for {
```
mk-software-pl, line 31: WDYT about doing it like this?
```scala
    OptionT(authenticationResources.authenticate(impersonatingUserAuthCredentials))
      .map { user => AuthenticatedUser.createImpersonatedUser(user, impersonatedUserData) }
      .value
```

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,75 @@
...
+    anonymousAccess.anonymousUserRole match {
+      case Some(role) =>
+        authenticationResources.authenticate().optional.flatMap {
+          case Some(user) => impersonatedAccess.provideOrImpersonateUser(user)
+          case None =>
+            anonymousAccess.handleAuthorizationFailedRejection.tmap(_ =>
+              AuthenticatedUser.createAnonymousUser(Set(role))
+            )
+        }
+      case None => authenticationResources.authenticate().flatMap(impersonatedAccess.provideOrImpersonateUser)
```
mk-software-pl, line 34: does it make sense to impersonate the anonymous user?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,75 @@
...
+        impersonatedAccess.authenticateWithImpersonation(impersonatingUserAuthCredentials, impersonatedUserData)
+    }
+  }
+
+  def authenticationEndpointInput(): EndpointInput[AuthCredentials] = {
+    authenticationResources
+      .authenticationMethod()
+      .and(ImpersonatedAccess.headerEndpointInput)
+      .map(
+        Mapping.fromDecode[(Option[String], Option[String]), AuthCredentials] {
```
mk-software-pl, line 54: maybe we can map the headerEndpointINput only? Sth like this: `.and(ImpersonatedAccess.headerEndpointInput.map(...))`

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,75 @@
...
+          case (Some(passedCredentials), None) => DecodeResult.Value(PassedAuthCredentials(passedCredentials))
+          case (Some(passedCredentials), Some(impersonatedUserData)) =>
+            decode[ImpersonatedUserData](impersonatedUserData) match {
```
mk-software-pl, line 57: I'm not sure if we should pass all the impersonated user data. I suspect that this is done for the sake of oauth2, right? 

In the case of basic auth it's simple because we have the user on the backend side. We can (and TBH we should) verify if the user exists and has proper permissions. For oauth2 it's not a case.

security/src/test/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticationSpec.scala
```diff
@@ -46,32 +46,12 @@ class BasicAuthenticationSpec extends AnyFunSpec with Matchers with ScalatestRou
     }
   )
 
-  it("should permit an anonymous user") {
```
mk-software-pl, line 49: why did we remove the tests?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/MigrationApiHttpServiceSecuritySpec.scala
```diff
@@ -46,31 +46,31 @@ class MigrationApiHttpServiceSecuritySpec
...
-          .basicAuthReader()
+          .basicAuthLimitedReader()
           .jsonBody(requestData)
           .post(s"$nuDesignerHttpAddress/api/migrate")
           .Then()
           .statusCode(401)
-          .equalsPlainBody("The supplied user [reader] is not authorized to access this resource")
+          .equalsPlainBody("The supplied user [limitedReader] is not authorized to access this resource")
       }
       "forbid access for user with limited writing permissions" in {
```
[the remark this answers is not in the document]
Elmacioro: It worked as `limitedReader` did not have any writing permissions. I modifed the test to use `limitedWriter` who has writing permission to `Category1` but not `Category2` so the logic is more reasonable.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NuDesignerApiAvailableToExposeYamlSpec.scala
```diff
@@ -164,7 +163,6 @@ object NuDesignerApiAvailableToExpose {
   private def createInstanceOf(clazz: Class[_ <: BaseEndpointDefinitions]) = {
     val basicAuth = auth
       .basic[Option[String]]()
-      .map { AnonymousAccess.optionalStringToAuthCredentialsMapping(false) }
```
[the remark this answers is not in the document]
Elmacioro: Well, after looking into it - we need this and a bit more.
I added the mapping for anonymous access and also for impersonation.

Since we have an additional optional header now I also had to udpate the OpenAPI spec file.
Thank you for pointing this one out.

extensions-api/src/main/scala/pl/touk/nussknacker/security/AuthCredentials.scala
```diff
@@ -1,10 +1,20 @@
 package pl.touk.nussknacker.security
 
+import io.circe.generic.JsonCodec
+
 sealed trait AuthCredentials
 
 object AuthCredentials {
 
   final case class PassedAuthCredentials(value: String) extends AuthCredentials
-  case object AnonymousAccess                           extends AuthCredentials
```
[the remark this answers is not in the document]
Elmacioro: I added the mention to migration guide.

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/AnonymousAccess.scala
Elmacioro: Oh, I was not aware of that. Thank you! I modified it so potential errors will be placed in the context.

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/AnonymousAccess.scala
Elmacioro: True, I moved it and made it an  `AuthenticationManager` internal class

security/src/main/scala/pl/touk/nussknacker/ui/security/accesslogic/ImpersonatedAccess.scala
Elmacioro: Looks nicer! Thank you!

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: Currently it would be possible if technical user sent impersonated user data with an `Anonymous` role in the header.
I don't think there is much need for impersonating anonymous user but since a user with `Impersonate` permission can impersonate any user he could also do that with an anonymous one.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: I mapped the header input separately and used it for final mapping. As this logic is part of impersonation access I moved it to the `ImpersonatedAccess` companion object.

security/src/test/scala/pl/touk/nussknacker/ui/security/basicauth/BasicAuthenticationSpec.scala
Elmacioro: I wanted to move some tests regaring anonymous access to the `AuthenticationManager` since it's now its resposibility to handle it. However i did it a bit too heavy handed and removed two tests that should still be in the `BasicAuthenticationSpec` - I added them back.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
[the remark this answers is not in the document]
Elmacioro: After offline consultations we decided that there is an Admin user. I modified the documentation to include information about it.
After changes admin user cannot be impersonated by default. 
However i made it possible to impersonate an Admin by setting `isAdminImpersonationPossible` flag in the authorization configuration to `true`. I made it possible as some external systems working with Nussknacker need that option.

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/CommentEntityFactory.scala
Elmacioro: After offline consultations two columns regaring impersonation will be added. First one holding the Id and second the name of an impersonating user. 
As there are some plans for switching from using username in favor of identity in Nu db tables this will let us simply delete the additional `impersonatedByUsername` column once it's not needed. For now I would add it so it's possible to present impersonating user's name on the frontend.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: As I mentioned in one of the comments https://github.com/TouK/nussknacker/pull/6053#discussion_r1622667524 new mechanism with `ImpersonationContext` was introduced. Impersonation is currently implemeneted for Basic Auth and I left a TODO comment for OAuth2.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/ImpersonationContext.scala
```diff
@@ -0,0 +1,25 @@
+package pl.touk.nussknacker.ui.security.api
+
+import com.typesafe.config.Config
+import sttp.client3.SttpBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ImpersonationContext {
```
mk-software-pl, line 8: I'm not sure if we shouldn't do it in a little bit different way. 

Let's assume that we have these two things: `AuthenticationManager` and `AuthenticationResources`. The `AuthenticationManager` is just a wrapper or adapter for the `AuthenticationResources` class. It's fine for now. 

The `AuthenticationResources` interface shows us that it's responsible for providing authentication methods (implementations). For me, it'd be nice, if it could say more: 
* I'm responsible for the impersonation support
* I'm responsible for the anonymous access support

It could look like that:
```scala
trait AuthenticationResources extends ... {
  this: ImpersonationSupport with AnonymousAccessSupport
}
```

and the `ImpersonationSupport` trait could look like this:
```scala
trait ImpersonationSupport {

  def getImpersonatedUserData(impersonatedUserIdentity: String): Option[ImpersonatedUserData]

}

trait NoImpersonationSupport extends {
   protected def getImpersonatedUserData(impersonatedUserIdentity: String): Option[ImpersonatedUserData] = None 
}
```

it seems that the things from ImpersonatedAccess we could move to AuthenticationManager.

The same could be done in case of the anonymous access.

Then the e.g. `OAuth2AuthenticationResources` implementation could look like this:
```scala
class OAuth2AuthenticationResources(...)
  extends AuthenticationResources
   with NoImpersonationSupport
   with AnonymousAccessSupport { ... }
```
WDYT about this appoarch? TBH it's rather a little modification compared to the current solution.

docs/Changelog.md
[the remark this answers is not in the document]
Elmacioro: I modified the changelog entry to mention basic auth and the API

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: After the changes where we do not directly pass roles in the impersonation header and pass impersonated user's identity instead it does not seem to be possible anymore.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/UserApiEndpoints.scala
```diff
@@ -70,26 +70,48 @@ object DisplayableUser {
...
+        username = user.username,
         // Sorting for stable tests results
         categories = allUserAccessibleCategories.toList.sorted,
-        categoryPermissions = Map.empty,
-        globalPermissions = Nil
+        categoryPermissions = categoryPermissionsOf(user),
+        globalPermissions = globalPermissionsOf(user)
       )
+    case ImpersonatedUser(impersonatedUser, _) =>
+      new DisplayableUser(
```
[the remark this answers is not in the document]
Elmacioro: Ok, i modified it to be cleaner.

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -535,13 +543,18 @@ class AkkaHttpBasedRouteProvider(
         } ~ pathPrefix("api") {
           apiResourcesWithoutAuthentication.reduce(_ ~ _)
         } ~ pathPrefix("api") {
-          authenticationResources.authenticate() { authenticatedUser =>
+          authenticationManager.authenticate() { authenticatedUser =>
             authorize(authenticatedUser.roles.nonEmpty) {
-              val loggedUser = LoggedUser(
-                authenticatedUser = authenticatedUser,
-                rules = authenticationResources.configuration.rules
+              val maybeLoggedUser = LoggedUser.create(
```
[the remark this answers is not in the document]
Elmacioro: Oh right, I modified Akka routes to also return 403 with the same message as Tapir endpoints.

docs/installation_configuration_guide/DesignerConfiguration.md
```diff
@@ -141,18 +145,29 @@ have one.
 
 Currently supported permissions:
 
-* AdminTab - shows Admin tab in the UI (right now there are some useful things kept there including search components
-  functionality).
+* Impersonate - Enables technical user to impersonate a business user and act on their behalf with their permissions.
+
+### Impersonation
+
+Nussknacker has an impersonation mechanism, allowing system's technical users to perform actions on behalf of
```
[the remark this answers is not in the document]
Elmacioro: Ok, I added the mention.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -48,27 +71,78 @@ object LoggedUser {
     }
   }
 
+  private def createImpersonatedUser(
+      impersonatingUser: RealLoggedUser,
+      impersonatedLoggedUser: RealLoggedUser,
+      isAdminImpersonationPossible: Boolean
+  ): Either[CreationError, LoggedUser] =
+    if (isAdminImpersonationPossible || (!isAdminImpersonationPossible && !impersonatedLoggedUser.isAdmin))
```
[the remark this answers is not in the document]
Elmacioro: Ok, it looks better.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -48,27 +71,78 @@ object LoggedUser {
...
   override def can(category: String, permission: Permission): Boolean = true
-  override val isAdmin: Boolean                                       = true
+}
+
+final case class ImpersonatedUser(impersonatedUser: RealLoggedUser, impersonatingUser: RealLoggedUser)
+    extends LoggedUser {
+  override val id: String                                             = impersonatedUser.id
+  override val username: String                                       = impersonatedUser.username
+  override def can(category: String, permission: Permission): Boolean = impersonatedUser.can(category, permission)
+  val impersonatedBy: RealLoggedUser                                  = impersonatingUser
```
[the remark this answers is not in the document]
Elmacioro: No, i removed it. It seemed to be an artifact of the past changes.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+  def authenticate(authCredentials: AuthCredentials): Future[Option[AuthenticatedUser]] = authCredentials match {
+    case passedCredentials @ PassedAuthCredentials(_) => authenticationResources.authenticate(passedCredentials)
+    case NoCredentialsProvided                        => authenticateWithAnonymousAccess()
+    case ImpersonatedAuthCredentials(impersonatingUserAuthCredentials, impersonatedUserIdentity) =>
+      authenticateWithImpersonation(impersonatingUserAuthCredentials, impersonatedUserIdentity)
+  }
+
+  def authenticationEndpointInput(): EndpointInput[AuthCredentials] =
+    authenticationResources
+      .authenticationMethod()
```
mk-software-pl, line 45: at the moment `authenticationMethod` returns `Option[String]` aliased as `Credentials`. WDYT about changing it to `Option[PassedAuthCredentials]`?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+      .and(impersonationHeaderEndpointInput)
```
mk-software-pl, line 46: it would be great if we could write it like that:
```scala
 authenticationResources
      .authenticationMethod()
      .withPossibleImpersonation()
```

and it could be implemented like this:
```
  
  implicit class ImpersonationConsideringInputEndpoint(underlying: EndpointInput[Option[String]]) {
    def withPossibleImpersonation(): EndpointInput[AuthCredentials] = {
      underlying
        .and(impersonationHeaderEndpointInput)
        .map(mappedAuthenticationEndpointInput)
    }
    
    private def impersonationHeaderEndpointInput = ???
    private def mappedAuthenticationEndpointInput = ???
```

We could retuse this extension class in tests too (see NuDesignerApiAvailableToExposeYamlSpec)

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+      .map(mappedAuthenticationEndpointInput)
+
+  private def provideOrImpersonateUser(user: AuthenticatedUser): Directive1[AuthenticatedUser] = {
+    optionalHeaderValueByName(impersonateHeaderName).flatMap {
+      case Some(impersonatedUserIdentity) =>
+        authenticationResources.getImpersonatedUserData(impersonatedUserIdentity) match {
+          case Some(impersonatedUserData) =>
+            provide(AuthenticatedUser.createImpersonatedUser(user, impersonatedUserData))
+          case None => reject
```
mk-software-pl, line 55: I've previously added a comment about more symmetric implementation. I'm not sure but here it looks like it behaves a little bit differently than in the Tapir's case (the `None`s, in Tapir's case, are flattened, so it looks like it doesn't behave exactly the same). Maybe it'll be easier to see it when you do the changes mentioned in my previous comment.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+      case None => provide(user)
+    }
+  }
+
+  private def authenticateWithImpersonation(
+      impersonatingUserAuthCredentials: PassedAuthCredentials,
+      impersonatedUserIdentity: ImpersonatedUserIdentity,
+  ): Future[Option[AuthenticatedUser]] = (for {
+    impersonatingUser <- OptionT(authenticationResources.authenticate(impersonatingUserAuthCredentials))
+    impersonatedUserData <- OptionT(
```
mk-software-pl, line 66: `OptionT` has many helpful factory methods.
E.g here we could do it like this:
```scala
impersonatedUserData <- OptionT.fromOption[Future](authenticationResources.getImpersonatedUserData(impersonatedUserIdentity.value))
```

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  val mappedAuthenticationEndpointInput: Mapping[(Option[String], Option[String]), AuthCredentials] =
+    Mapping.fromDecode[(Option[String], Option[String]), AuthCredentials] {
+      case (Some(passedCredentials), None) => DecodeResult.Value(PassedAuthCredentials(passedCredentials))
+      case (Some(passedCredentials), Some(identity)) =>
+        DecodeResult.Value(
+          ImpersonatedAuthCredentials(PassedAuthCredentials(passedCredentials), ImpersonatedUserIdentity(identity))
+        )
+      case (None, None) => DecodeResult.Value(NoCredentialsProvided)
+      case _            => DecodeResult.Missing
```
mk-software-pl, line 99: let's do it explicity: `(None, Some(_))`

It means that the creds were not passed, but there was the impersonation header present. We return `Missing`. Maybe we should add a comment about why here we return Missing instead of `NoCredentialsProvided`. It's not obvious at first glance. 

WDYT?

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -2,21 +2,44 @@ package pl.touk.nussknacker.ui.security.api
...
+          loggedUser,
+          LoggedUser(impersonatedUser, rules),
+          isAdminImpersonationPossible
+        )
+      case Some(_) =>
+        Left(ImpersonationNotAllowed)
+    }
+  }
+
   def apply(
```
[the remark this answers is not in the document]
Elmacioro: I checked and we use additional `apply` methods extensively in tests for creating logged users.
Hovewer now they are all responsible for creation of `RealLoggedUser` so I moved them to `RealLoggedUser` companion object. As sigantures do not overlap anymore I renamed the `LoggedUser` `create` method to `apply` and it's responsible for creation of both real and impersonated users.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+}
+import sttp.tapir._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuthenticationManager(authenticationResources: AuthenticationResources)(implicit ec: ExecutionContext) {
+
+  lazy val authenticationRules: List[AuthenticationConfiguration.ConfigRule] =
+    authenticationResources.configuration.rules
+  lazy val isAdminImpersonationPossible: Boolean = authenticationResources.configuration.isAdminImpersonationPossible
```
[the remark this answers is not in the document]
Elmacioro: I created mentioned methods and renamed the class to `AuthManager`.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
```diff
@@ -0,0 +1,107 @@
...
+  )
+
+  private def authenticateWithAnonymousAccess(): Future[Option[AuthenticatedUser]] = Future {
+    authenticationResources.getAnonymousRole.map(role => AuthenticatedUser.createAnonymousUser(Set(role)))
+  }
+
+}
+
+object AuthenticationManager {
+  val impersonateHeaderName = "Impersonate-User-Identity"
```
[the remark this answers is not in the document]
Elmacioro: I omitted the `X-` prefix on purposie since it's no longer recommended to add it to custom header names https://www.rfc-editor.org/rfc/rfc6648
I added the `Nu` prefix, though.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: True, it's better. At some point I had it this way but changed it for some reason.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: Done, I introduced the implcicit class.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: As I mentioned it doesn't seem trivial to make it more symmetric. 
I looked into the logic and in case of Tapir I did not differenate between failure in impersonating user's authentication and failure in fetching impersonated user's data. I modified the code to operate on `Either` in case of Tapir authentication to handle that situation. I moved `SecurityError` classes to the security module. It makes more sense for them to be there and it made it easier to handle error responses for both Tapir and Akka.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: Oh, yes it's better!

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthenticationManager.scala
Elmacioro: I added it explicitly and commented why we return that.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthManager.scala
```diff
@@ -0,0 +1,162 @@
...
+  BaseAuthenticationError,
+  BaseAuthorizationError,
+  ImpersonatedUserDataNotFoundError,
+  ImpersonationMissingPermissionError
+}
+import sttp.tapir._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuthManager(authenticationResources: AuthenticationResources)(implicit ec: ExecutionContext) {
```
mk-software-pl, line 31: WDYT about creating two traits `AkkaBasedAuthManager` and moving there the code related to the akka things? In the future, we will just remove it. At the moment there are not so many (non-trivial) common code lines in this class. Moreover, we are not able to easily make these two cases symmetric. Reading the code of the AuthManager is not easy because akka and trapir things are mixed. Maybe it's worth to exclude the akka things to the private trait (for a sake of easy getting rid of it in the future).

mk-software-pl (approved):
Just minor comments left. 
LGTM. Great change!

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
```diff
@@ -0,0 +1,20 @@
+package pl.touk.nussknacker.ui.security.api
+
+sealed trait SecurityError
+sealed trait AuthorizationError  extends SecurityError
+sealed trait AuthenticationError extends SecurityError
+
+object SecurityError {
+  case object BaseAuthorizationError extends AuthorizationError
```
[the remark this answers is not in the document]
Elmacioro: Changed the name to `InsufficientPermission`

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  case object ImpersonationMissingPermissionError extends AuthorizationError {
+    val errorMessage = "The supplied authentication is not authorized to impersonate"
+  }
+
+  case object BaseAuthenticationError extends AuthenticationError
```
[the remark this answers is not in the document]
Elmacioro: Changed the name to `CannotAuthenticateUser`

security/src/test/scala/pl/touk/nussknacker/ui/security/api/AuthManagerSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  private val config = ConfigFactory.parseString(s"""
+        authentication: {
+          method: "BasicAuth"
+          anonymousUserRole: "Anonymous"
+          usersFile: "classpath:basic-users.conf"
+        }
+      """.stripMargin)
+
+  private val authenticationResources = AuthenticationResources(config, classLoader, testingBackend)
```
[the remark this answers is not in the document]
Elmacioro: Done

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthManager.scala
```diff
@@ -0,0 +1,162 @@
...
+    authenticationResources
+      .authenticationMethod()
+      .withPossibleImpersonation()
+
+  def authorizeDirective(user: AuthenticatedUser)(secureRoute: LoggedUser => Route): Route =
+    Directives.authorize(user.roles.nonEmpty) {
+      LoggedUser(user, authenticationRules, isAdminImpersonationPossible) match {
+        case Left(ImpersonationNotAllowed) =>
+          complete(StatusCodes.Forbidden -> ImpersonationMissingPermissionError.errorMessage)
+        case Right(loggedUser) => secureRoute(loggedUser)
```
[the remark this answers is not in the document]
Elmacioro: Done

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthManager.scala
```diff
@@ -0,0 +1,173 @@
...
+  def authenticationEndpointInput(): EndpointInput[AuthCredentials] =
+    authenticationResources
+      .authenticationMethod()
+      .withPossibleImpersonation()
+
+  def authorize(user: AuthenticatedUser): Either[AuthorizationError, LoggedUser] = {
+    if (user.roles.nonEmpty)
+      // TODO: This is strange that we call authenticator.authenticate and the first thing that we do with the returned user is
+      //       creation of another user representation based on authenticator.configuration. Shouldn't we just return the LoggedUser?
+      LoggedUser(user, authenticationRules, isAdminImpersonationPossible) match {
```
arkadius, line 57: 1. By convention, companion's object apply method rather returns companion class instance. For method returning either we can create a named factory method e.g. `LoggedUser.create(...)`.
2. It can be simplified to `.left.map(...)`

security/src/main/scala/pl/touk/nussknacker/ui/security/api/ImpersonationSupport.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.ui.security.api
+
+trait ImpersonationSupport {
+  def getImpersonatedUserData(impersonatedUserIdentity: String): Option[ImpersonatedUserData]
+}
+
+trait NoImpersonationSupport extends ImpersonationSupport {
+  override def getImpersonatedUserData(impersonatedUserIdentity: String): Option[ImpersonatedUserData] = None
```
arkadius, line 8: I think that we should rather return `Left[ImpersonationNotSupported]` or at least throw an exception in this case instead of retuning `None`. `None` suggests that it is a client problem, that sent id of user which doesn't exists but it might be either client problem that uses impersonation but it shouldn't or server problem, that configuration is invalid.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
```diff
@@ -2,23 +2,89 @@ package pl.touk.nussknacker.ui.security.api
...
+          .contains(Permission.Impersonate.toString.toLowerCase)
+    }
+
+    def isAdmin: Boolean = user match {
+      case _: AdminUser        => true
+      case _: CommonUser       => false
+      case u: ImpersonatedUser => u.impersonatedUser.isAdmin
+    }
+
+    def getImpersonatingUserId: Option[String] = user match {
```
arkadius, line 67: is scala we rather omit `get`

arkadius (approved):

security/src/main/scala/pl/touk/nussknacker/ui/security/api/AuthManager.scala
Elmacioro: Done, I renamed it to `create` and simplified logic as proposed

security/src/main/scala/pl/touk/nussknacker/ui/security/api/ImpersonationSupport.scala
Elmacioro: I intorduced a solution based on `Either` instead of exceptions.
We didn't want to change `getImpersonatedUserData` signature to `Either[NoImpersonationSupport, Option[ImpersonatedUserData]]` as while implementing `getImpersonatedUserData` we should not care about the `NoImpersonationSupport` situation. So I introduced `getImpersonatedUserDataWithSupportCheck` implemented for both `ImpersonationSupport` and `NoImpersonationSupport` which wraps `getImpersonatedUserData` helping with error handling.

I also added custom responses for both Tapir and Akka in case somebody tries to impersonate while authentication mechanism does not support it.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/LoggedUser.scala
Elmacioro: Ok, I removed the prefix

docs/MigrationGuide.md
```diff
@@ -6,6 +6,30 @@ To see the biggest differences please consult the [changelog](Changelog.md).
...
+        * `authenticationMethod()` returns `EndpointInput[Option[PassedAuthCredentials]]` instead of `EndpointInput[AuthCredentials]`.
+          The `Option[PassedAuthCredentials]` should hold the value that will be passed to the mentioned `authenticate(authCredentials)`.
+        * `AuthenticationResources` extends `AnonymousAccessSupport` trait:
+          * `AnonymousAccessSupport` has one method `getAnonymousRole()` which returns anonymous role name. If you do not want to have
+            an anonymous access mechanism for your authentication method you can extend your `AuthenticationResources`
+            implementation with `NoAnonymousAccessSupport` trait.
+        * `AuthenticationResources` has a field `impersonationSupport` of type `ImpersonationSupport`:
+          * `ImpersonationSupport` is a trait stating whether authentication method supports impersonation.
+            If you don't want impersonation support you can assign `NoImpersonationSupport` object to it.
+            If you wish to have it - assign `ImpersonationSupport` abstract class to it and
```
arkadius, line 26: `ImpersonationSupported` should be here

PR #6055, opened 2024-05-14
[NU-1657] Fixup to async DefaultAsyncExecutionContextPreparer

DeamonDev:
## Describe your changes

We've run into the problem a few times now with the scenario getting a closed (Terminated) pool of threads from before the restart.
The idea is to remove the mechanism of atomic counting of the number of calls to open/close the pool. Instead, we use the "closeAsyncExecutionContext" hook and in the case of pool closure, we simply unregister user's class loader from Flink and reset the static variable: DefaultServiceExecutionContextPreparer#asyncExecutionContext.

BEWARE:  Asynchronous tests can still have this problem, because the tests share the same class loader, and in parallel execution that class loader is not unloaded - resulting in the pool not being closed!

## Checklist before merge


raphaelsolarski (approved):
Epic!

PR #6105, opened 2024-05-27
[NU-1585] resolve active save button after scenario save issue

Dzuming:
## Describe your changes
- It seems like it was some kind of race condition, which fetchTestFormParameters updating a redux history, after a display process action.
- Also we fetchTestFormParameters and displayTestCapabilities on every useTestWithFormAvailability re-render, instead of  scenarioGraph change
 
## Checklist before merge


designer/client/src/actions/nk/process.ts
```diff
@@ -15,7 +15,6 @@ export function fetchProcessToDisplay(processName: ProcessName, versionId?: Proc
         dispatch({ type: "PROCESS_FETCH" });
 
         return HttpService.fetchProcessDetails(processName, versionId).then((response) => {
-            dispatch(displayTestCapabilities(processName, response.data.scenarioGraph));
```
JulianWielga, line 18: this way test capabilities may be never fetched.

JulianWielga (approved):

PR #6118, opened 2024-05-29
[NU-1520] Workaround for handle multiple instances of questDB

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
@@ -14,6 +14,7 @@ import pl.touk.nussknacker.ui.db.timeseries.QuestDbFEStatisticsRepository.{creat
 
 import java.nio.charset.StandardCharsets
 import java.nio.file.StandardOpenOption
+import java.util.UUID
 import java.util.concurrent.{ArrayBlockingQueue, ThreadPoolExecutor, TimeUnit}
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService, Future}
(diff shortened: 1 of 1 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
```diff
@@ -181,7 +182,9 @@ object QuestDbFEStatisticsRepository extends LazyLogging {
   }
 
   private def createRootDirIfNotExists(): File = {
-    val nuDir = File.temp.createChild("nu", asDirectory = true)
+    val nuDir = File.temp
+      .createChild("nu", asDirectory = true)
+      .createChild(UUID.randomUUID().toString, asDirectory = true)
```
arkadius, line 187: So after a restart, we will lose data?

lukasz-bigorajski: Yes, it's workaround to work with multiple instances. If it's not critical we can close PR.

arkadius: Ok, so as a temporary workaround it is fine. Let's only remember to fix it before the release. Maybe some jira task with fix version would help with that?

arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/timeseries/QuestDbFEStatisticsRepository.scala
arkadius: It is not critical IMO, we can leave for a while with this. There is an easy workaround for that - we just need to stop designer during tests and resume it after tests are finished

lukasz-bigorajski: Ok, so I close PR

PR #6134, opened 2024-06-05
Fix last action data determining

wrzontek:
## Describe your changes
This PR fixes the discrepant (and in the case of `lastDeployedActionData` incorrect in both cases) behaviour of `lastStateActionData/lastDeployedActionData` determining between `fetchLatestProcessDetailsByQueryAction` and `fetchProcessDetailsForVersion`

Here are the behaviours BEFORE the changes in this PR: ('list' is `fetchLatestProcessDetailsByQueryAction`, 'single' is `fetchProcessDetailsForVersion`)

lastActionData:
 - 'single' and 'list' - latest action out of those that are `Finished/ExecutionFinished`

lastStateActionData:
- 'single' - latest action out of those that (are `Finished/ExecutionFinished` _**AND**_ are state actions)
- 'list' - latest action out of those that (are `Finished/ExecutionFinished`) **_IF IT IS_** a state action

lastDeployedActionData (could be renamed to currentlyDeployedActionData IMO):
- 'single' - latest action out of those that are `Finished/ExecutionFinished` if it is `name==Deploy` and `state==Finished`
- 'list' - latest action out of those that are `Finished` if it is `name==Deploy`

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
@@ -76,11 +76,17 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](
         actionRepository
           .getLastActionPerProcess(ProcessActionState.FinishedStates, Some(ScenarioActionName.StateActions))
       )
-      // for last deploy action we are not interested in ExecutionFinished deploys - we don't want to show them in the history
+      // For last deploy action we are not interested in Deploys that are Finished, but not ExecutionFinished, and that are not Cancelled
+      // so that the presence of such an action means that the process is currently deployed
       lastDeployedActionPerProcess <- fetchActionsOrEmpty(
         actionRepository
-          .getLastActionPerProcess(ProcessActionState.FinishedStates, Some(Set(ScenarioActionName.Deploy)))
-      ).map(_.filter { case (_, action) => action.state == ProcessActionState.Finished })
+          .getLastActionPerProcess(
+            ProcessActionState.FinishedStates,
+            Some(Set(ScenarioActionName.Deploy, ScenarioActionName.Cancel))
+          )
+      ).map(_.filter { case (_, action) =>
+        action.actionName == ScenarioActionName.Deploy && action.state == ProcessActionState.Finished
+      })
 
       latestProcesses <- fetchLatestProcessesQuery(query, lastDeployedActionPerProcess.keySet, isDeployed).result
     } yield latestProcesses
@@ -182,9 +188,13 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](
       processVersion = processVersion,
       lastActionData = actions.headOption,
       lastStateActionData = actions.find(a => ScenarioActionName.StateActions.contains(a.actionName)),
-      // for last deploy action we are not interested in ExecutionFinished deploys - we don't want to show them in the history
-      lastDeployedActionData =
-        actions.find(_.actionName == ScenarioActionName.Deploy).filter(_.state == ProcessActionState.Finished),
+      // For last deploy action we are not interested in Deploys that are Finished, but not ExecutionFinished, and that are not Cancelled
+      // so that the presence of such an action means that the process is currently deployed
+      lastDeployedActionData = actions
+        .find(action => Set(ScenarioActionName.Deploy, ScenarioActionName.Cancel).contains(action.actionName))
+        .filter(action =>
+          action.actionName == ScenarioActionName.Deploy && action.state == ProcessActionState.Finished
+        ),
       isLatestVersion = isLatestVersion,
       tags = Some(tags),
       history = Some(
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
@@ -77,9 +77,10 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](
           .getLastActionPerProcess(ProcessActionState.FinishedStates, Some(ScenarioActionName.StateActions))
       )
       // for last deploy action we are not interested in ExecutionFinished deploys - we don't want to show them in the history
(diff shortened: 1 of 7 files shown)


wrzontek:
TODO: unit test

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
```diff
@@ -77,9 +77,10 @@ abstract class DBFetchingProcessRepository[F[_]: Monad](
           .getLastActionPerProcess(ProcessActionState.FinishedStates, Some(ScenarioActionName.StateActions))
       )
       // for last deploy action we are not interested in ExecutionFinished deploys - we don't want to show them in the history
-      lastDeployedActionPerProcess <- fetchActionsOrEmpty(
-        actionRepository.getLastActionPerProcess(Set(ProcessActionState.Finished), Some(Set(ScenarioActionName.Deploy)))
-      )
+      lastDeployedActionPerProcess = lastActionPerProcess.filter { case (_, action) =>
+        action.actionName == ScenarioActionName.Deploy && action.state == ProcessActionState.Finished
```
arkadius, line 81: Why is it better?  After that we won't see deploy actions if there are some actions after it

wrzontek: >After that we won't see deploy actions if there are some actions after it
that was already the case

see the PR description, the only thing that changed was it made to behave the same as `fetchProcessDetailsForVersion`

wrzontek: i do think this code is still ugly, not very readable, and potentially just plain wrong, but the objective of this quick PR was to at least make it consistent

i've got a bigger fix (and performance improvement) cooking in a separate PR (that is probably too large to be backported to 1.15)

arkadius: Why are you sure that `fetchProcessDetailsForVersion` shouldn't behave like `fetchLatestProcessDetailsByQueryAction` ?

wrzontek: ok, perhaps this requires more analysis, here's how i understand the intended 'meaning' of the fields

- '`lastActionData`' - latest Finished/ExecutionFinished action (could perhaps be relaxed to latest action regardless of ProcessActionState), both methods behave the same

- '`lastStateActionData`' - latest Finished/ExecutionFinished state action, so last action that succeeded and impacted the process state (Cancel, Deploy, Pause) - IMO `fetchProcessDetailsForVersion` behaves correctly

- '`lastDeployedActionData`' (the tricky one) - latest Finished deploy action, that hasn't yet been ExecutionFinished, so if a process has such an action, it means that it is currently deployed (and this 'meaning' is used by `fetchLatestProcessDetailsByQueryAction -> fetchLatestProcessesQuery`) - IMO both behave incorrectly

wrzontek: so if we want to achieve the (in my opinion) desired outcome for `lastDeployedActionData` we'd have:

in `fetchLatestProcessDetailsByQueryAction`
```
lastDeployedActionPerProcess <- fetchActionsOrEmpty(
  actionRepository
    .getLastActionPerProcess(ProcessActionState.FinishedStates, Some(Set(ScenarioActionName.Deploy)))
).map(_.filter { case (_, action) => action.state == ProcessActionState.Finished })
```

and in `fetchProcessDetailsForVersion`
```
// for last deploy action we are not interested in ExecutionFinished deploys - we don't want to show them in the history
lastDeployedActionData =
  actions.find(_.actionName == ScenarioActionName.Deploy).filter(_.state == ProcessActionState.Finished),
```

arkadius: Yeah, it looks like 3rd option (making both methods work the same, but otherwise than any of them) ;-) is correct. Now we need a test for that :)

arkadius: I miss one more thing. I understood @gskrobisz that he discovered that we have a bug in two places:
1. In the history panel - "DEPLOYED" badge is missing - I assume that `lastDeployedActionData` will fix it
2. In the status visible on the scenario list - I thought that `lastStateActionData` is used for that. Isn't it?

wrzontek:
the failing (old) UT fails with the following:
1. scenario is deployed (action Deploy with state Finished)
2. scenario is cancelled (action Cancel with state Finished)

and the new code fails, because it considers it still Deployed (perhaps the `Cancel` action could also cause the present `Deploy` action to transition to state `ExecutionFinished`? It would require special treatment around `DeploymentService.runActionAndHandleResults->ProcessActionRepository.markActionAsFinished`)

the old code passed the test, but mostly by accident, any action would have 'covered' the Deploy, even if it was `CustomAction("do literally nothing")`

I'll change 'lastDeployedActionData' to: the latest action out of those that are `Finished/ExecutionFinished` and are `Deploy/Cancel` is `Deploy-Finished`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/DBFetchingProcessRepository.scala
wrzontek: > 1.In the history panel - "DEPLOYED" badge is missing - I assume that `lastDeployedActionData` will fix it
> 2.In the status visible on the scenario list - I thought that `lastStateActionData` is used for that. Isn't it?

1. yes, pretty sure that's correct
2. FE actually doesn't use `lastStateActionData` at all anywhere, but BE uses it to determine `ScenarioWithDetails.state` (in `DeploymentService.getProcessState`) which FE uses

wrzontek:
only fail looks like an unrelated cypress error

arkadius (approved):
LGTM

ForrestFairy (approved):
LGTM, let's merge it

PR #6138, opened 2024-06-05
[NU-1618] sync with installation example repo

mk-software-pl:
## Describe your changes

Nussknacker installation example are published to the separate [Nu installation example repo](https://github.com/TouK/nussknacker-installation-example) during release

## Checklist before merge


.github/workflows/scripts/publishNuInstallationExample.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "NU_INSTALLATION_EXAMPLE_ACCESS_TOKEN variable has to be defined"
+    exit 1
+fi
+
+if [ -z "$NUSSKNACKER_VERSION" ]; then
+    echo "NUSSKNACKER_VERSION variable has to be defined"
+    exit 1
+fi
+
+rm -rf nu-installation-example-repo
```
lukasz-bigorajski, line 13: This dir should not be clean up after push?

lukasz-bigorajski (approved):
LGTM

.github/workflows/scripts/publishNuInstallationExample.sh
coutoPL: TBH it should not be needed. I did it for easier local testing (the alternative would be checking if the directory doesn't exist and failing the script). IMO it's easier to just make sure the directory is not there (even if we trust there isn't)

coutoPL: oh, and we don't care about cleaning after the whole process because a runner always does the cleaning for us. But on the other hand the script should not leave any trashes when we call it manually, so I've added the cleanup step

PR #6159, opened 2024-06-07
New custom action behaviour

vder:
## Describe your changes
In current implementation of custom actions state of the  actions depends only on scenario's status. The following change makes this mechanism  little more flexible. Now buttons can be enabled/disabled according to the visible scenario version and last deployment action as well. In particular we want to enable custom action only in case when deployed version of the scenario is visible in designer.

## Checklist before merge


designer/client/src/components/toolbars/scenarioActions/buttons/CustomActionButton.tsx
```diff
@@ -11,29 +11,23 @@ import UrlIcon from "../../../UrlIcon";
...
 
     const icon = action.icon ? <UrlIcon src={action.icon} FallbackComponent={DefaultIcon} /> : <DefaultIcon />;
 
-    const statusName = processStatus?.name;
-    const available = !disabled && action.allowedStateStatusNames.includes(statusName);
-
-    const toolTip = available
-        ? null
-        : t("panels.actions.custom-action.tooltips.disabled", "Disabled for {{statusName}} status.", { statusName });
+    const toolTip = disabled ? t("panels.actions.custom-action.tooltips.disabled", "Disabled for current state.") : null;
```
Dzuming, line 23: Why do we change the previous message? IMO From the user's perspective it sounds better to have the info about status in this tooltip

designer/client/src/actions/nk/process.ts
```diff
@@ -27,12 +29,13 @@ export function fetchProcessToDisplay(processName: ProcessName, versionId?: Proc
...
     return (dispatch) =>
-        HttpService.fetchProcessState(processName).then(({ data }) =>
+        HttpService.fetchProcessState(processName).then(({ data }) => {
             dispatch({
                 type: "PROCESS_STATE_LOADED",
                 processState: data,
-            }),
-        );
+            });
+            dispatch(updateEnabledCustomActions(processName));
```
Dzuming, line 37: I don't understand what we want to achieve. IIUC Previously we determined the custom action enable/disable state, based on the scenario state. Could you describe in few words, we do we want to change this behavior?

designer/client/src/components/toolbars/scenarioActions/buttons/CustomActionButton.tsx
vder: Ok I've added some additional description in `Describe your changes` section. So I hope it brings some insight behind the change. And replying to this very question I removed  particular status name here  as I thought it can be little misleading as status of the button don't rely only on scenario status now.  Additionally this err message was the only reason to pass this status here as an function's argument.

designer/client/src/actions/nk/process.ts
vder: I've added some description above.  In this particular case we want to enable buttons only in case when deployed version of the scenario is visible. So after switching to old version or after creation of the new one we have to have opportunity to update state.

Dzuming: Thanks for the explanation! 

Ok, so in this case, we will make two API requests every 10 seconds

- `api/processes/<scenario_name>/status`
- `api/processManagement/customAction/<scenario>}/<scenario_version>`

and `/status` send in the response

```
{
    ...
    "allowedActions": [
        "DEPLOY",
        "CANCEL"
    ],
    ...
}
```
We duplicate a logic here. Did you check the possibility of extending the existing endpoint `allowedActions` property, and for example pass the status of specific scenarios version in the `/status` endpoint? Do you know about this approach?

Also, I can see a problem with adding it to the `loadProcessState`, because, it can be an option, that we will ask for the status, only in the case of the current version (of course if we extend `/status` endpoint, it will have the sense  to ask for a specific version status)

vder: I've got mixed felling about that :).  From one side I was told that we wanna keep custom action as a separate logic  and separate endpoint, from the other one I have another task to make custom action configurable from the files and make it a little closer to the build-in actions (opposed to the current solution where custom action is part of the deployment manager). So from that perspective I guess I could use (or at least try to use) `api/processes/<scenario_name>/status` endpont.

arkadius:
Closing due to inactivity

PR #6174, opened 2024-06-11
[NU-1704] Extend user statistics with the components used

ForrestFairy:
## Describe your changes
Created StatisticKeys for components and a map to easily map component name to its proper StatisticKey or to 'Custom'

URL with statistics only include components which are used at least once to shorten the length of the average URL.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ComponentMap.scala
ForrestFairy, line 1: I am not sure if this map should be in `ScenarioStatistics` which would make it even longer, so it's in a separate file for now, but I am open to put it in there

lukasz-bigorajski (commented):
IMHO it's ok if map is in another file. I will send you a PR about hardcoded names to discuss it

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -213,3 +233,90 @@ case object ActiveScenarioCount    extends StatisticKey("s_a")
 case object NuSource               extends StatisticKey("source") // f.e docker, helmchart, docker-quickstart, binaries
 case object NuFingerprint          extends StatisticKey("fingerprint")
 case object NuVersion              extends StatisticKey("version")
+
+object ComponentKeys {
+  case object Aggregate                         extends StatisticKey("ag")
```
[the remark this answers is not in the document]
ForrestFairy: Should we then change the prefix for comments statistics to `co_` for example?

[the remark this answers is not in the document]
lukasz-bigorajski: Ok, it's good idea.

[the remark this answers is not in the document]
ForrestFairy: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -115,16 +116,51 @@ object ScenarioStatistics {
     }
   }
 
-  def getComponentStatistic(componentList: List[component.ComponentListElement]): Map[String, String] = {
+  def getComponentStatistic(
+      componentList: List[component.ComponentListElement],
```
lukasz-bigorajski, line 120: Can we leave todo comment about removing components from restmodel?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -115,16 +116,51 @@ object ScenarioStatistics {
...
+      components: List[ComponentDefinitionWithImplementation]
+  ): Map[String, String] = {
     if (componentList.isEmpty) {
       Map.empty
     } else {
-      val withoutFragments      = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
-      val componentsByNameCount = withoutFragments.groupBy(_.name).size
+      val builtinNames = List("choice", "filter", "record-variable", "split", "variable")
```
lukasz-bigorajski, line 126: Can we take these names from components-api e.g. `BuiltInComponentId.Filter.name` and place it as static var?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -115,16 +116,51 @@ object ScenarioStatistics {
...
+      // Get number of available components to check how many custom components created
+      val withoutFragments = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
+      val componentsWithUsageByName: Map[String, Long] =
+        withoutFragments
+          .map { comp =>
+            components.find(compo => compo.name.equals(comp.name)) match {
+              case Some(comps) =>
+                if (comps.component.getClass.toString.contains("pl.touk.nussknacker")) {
```
lukasz-bigorajski, line 134: Can we replace this with: `getPackageName.startsWith`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -174,14 +210,25 @@ object ScenarioStatistics {
     else orderedList.last
   }
 
+  def mapNameToStat(name: String): String = {
```
lukasz-bigorajski, line 213: IMHO we should not rely on hashCode. We should better simply remove vowels or calculate and shorten md5 or sha hash.

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -213,3 +260,90 @@ case object ActiveScenarioCount    extends StatisticKey("s_a")
 case object NuSource               extends StatisticKey("source") // f.e docker, helmchart, docker-quickstart, binaries
 case object NuFingerprint          extends StatisticKey("fingerprint")
 case object NuVersion              extends StatisticKey("version")
+
+object ComponentKeys {
```
lukasz-bigorajski, line 264: If we introduce this kind of enum then we should have some test which will check all the values (if hash is calculated well) to have ability to detect changes in components names.

lukasz-bigorajski (changes requested):

lukasz-bigorajski:
I would also add some components statistics asserts in integration test

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -213,3 +258,102 @@ case object ActiveScenarioCount    extends StatisticKey("s_a")
 case object NuSource               extends StatisticKey("source") // f.e docker, helmchart, docker-quickstart, binaries
 case object NuFingerprint          extends StatisticKey("fingerprint")
 case object NuVersion              extends StatisticKey("version")
+
+sealed abstract class ComponentKeys(val value: String, val name: String) extends StringEnumEntry {
+  override def toString: String = value
+}
+
+object ComponentKeys extends StringEnum[ComponentKeys] {
```
lukasz-bigorajski, line 266: IMHO we was close to best solution at the beginning because map was good but we should only do checking of custom component based on packageName, but only what we should additional was extract components names to components-api.

```scala
val componentsMappings = Map(
  ComponentsApi.ComponentEnum.AggregateName -> "some short name"
)
```
and then in test we should if all values from enum has mapping.

But we what should do now? Imho I would leave this enum but maybe we should take shortName like that: 
```scala
and then:
```scala
  // TODO extract components name to components-api
  private def mapNameToStat(name: String): Option[String] = {
    ComponentKeys.withNameOption(name).map(_.shortName)
  // shortNames can be invented by us
  }
```

lukasz-bigorajski: > extract components names to components-ap

We cannot do this, so imho what we should do is leave mapping like it was with vowels and check mapping in some integration test and remove this enum with mappings. Writing such test would be hard so I would leave todo and write such test later

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -174,14 +214,19 @@ object ScenarioStatistics {
     else orderedList.last
   }
 
+  private def mapNameToStat(name: String): String = {
```
arkadius, line 217: Component name could be not enough to identify component. E.g. we have kafka source and kafka sink. `ComponentId` have both things

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
arkadius: Why do we still use DTO's (`ComponentListElement`) in the domain logic, we should base on only domain objects (`ComponentDefinitionWithImplementation`)

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -115,16 +119,52 @@ object ScenarioStatistics {
...
       Map.empty
     } else {
-      val withoutFragments      = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
-      val componentsByNameCount = withoutFragments.groupBy(_.name).size
 
-      Map(
-        ComponentsCount -> componentsByNameCount
-      ).map { case (k, v) => (k.toString, v.toString) }
+      // Get number of available components to check how many custom components created
+      val withoutFragments = componentList.filterNot(comp => comp.componentType == ComponentType.Fragment)
```
arkadius, line 132: We could add `getAlignedComponentsWithBuiltInComponents` next to the `getAlignedComponentsWithBuiltInComponentsAndFragments`. Thanks to that we don't need to filter out fragments here. Now we add some filter and then we filter out them, it looks strange.

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
ForrestFairy: We use `ComponentListElement` here to get `usageCount` for each component. I knew that it can be provided by `ComponentService` so I did it this way, but if there's another way I can change it. Or if there is currently no other way, I can try to make one

lukasz-bigorajski (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
ForrestFairy: This problem will be resolved once we completely switch from using `ComponentService` and that will be when we'll have other way to get `usageCount` for components

designer/server/src/main/scala/pl/touk/nussknacker/ui/statistics/ScenarioStatistics.scala
```diff
@@ -22,6 +22,10 @@ object ScenarioStatistics {
   private val knownDeploymentManagerTypes =
     Set(flinkDeploymentManagerType, liteK8sDeploymentManagerType, liteEmbeddedDeploymentManagerType)
 
+  private val vowelsRegex = "[aeiouAEIOU-]"
```
arkadius, line 25: yY are missing here?

arkadius (approved):
LGTM, one comment left

PR #6182, opened 2024-06-13
Deployment api unignore + additional logging 

arkadius:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/DeploymentRepository.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/DeploymentRepository.scala
@@ -64,8 +64,6 @@ class DeploymentRepository(dbRef: DbRef, clock: Clock)(implicit ec: ExecutionCon
       }
       .sequence
       .map(_.combineAll)
-      // For the performance reasons it is better to run all updates in the one session, transactionally should enforce it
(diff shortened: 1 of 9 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/DeploymentRepository.scala
```diff
@@ -64,8 +64,6 @@ class DeploymentRepository(dbRef: DbRef, clock: Clock)(implicit ec: ExecutionCon
       }
       .sequence
       .map(_.combineAll)
-      // For the performance reasons it is better to run all updates in the one session, transactionally should enforce it
-      .transactionally
```
mslabek, line 68: Is this comment on transactionally not applicable now?

mslabek (approved):
One clarifying question, other than that LGTM :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/DeploymentRepository.scala
arkadius: The comment is applicable but I'm not sure that `transactionally` was even needed. Maybe the fact that it is in one db action is enough. I don't want to check it now so I think that it is better to remove `transactionally` and add it when it is necessary.

PR #6195, opened 2024-06-14
[NU-1665] Table source test data generation

mslabek:
## Checklist before merge


e2e-tests/src/test/scala/pl/touk/nussknacker/DockerBasedInstallationExampleNuEnvironment.scala
```diff
@@ -19,7 +19,7 @@ import java.io.{File => JFile}
 trait DockerBasedInstallationExampleNuEnvironment extends BeforeAndAfterAll with BeforeAndAfterEach with LazyLogging {
   this: Suite =>
 
-  private lazy val specSetupService = unsafeContainerByServiceName("spec-setup")
+  private val specSetupService = unsafeContainerByServiceName("spec-setup")
```
mslabek, line 22: The companion object of the trait `DockerBasedInstallationExampleNuEnvironment` is not instantiated automatically when the trait is instantiated - it is instantiated when it is first accessed. Without something like this non-lazy val refering to that companion object, the docker env is not started before test that extend this test suite.

A good answer on SO for this with Scastie worksheet is here:
https://stackoverflow.com/questions/55477793/order-of-initialization-in-scala

utils/test-utils/src/main/resources/logback-test.xml
```diff
@@ -61,5 +61,6 @@
     <!-- skuber by default logs all requests to k8s on INFO -->
     <logger name="skuber" level="${SKUBER_LOG_LEVEL:-WARN}"/>
     <logger name="com.azure" level="INFO"/>
+    <logger name="org.testcontainers" level="DEBUG"/>
```
mslabek, line 64: Added a debug logging for testcontainers. Was really useful and I think it will be useful as default.

build.sbt
```diff
@@ -1813,6 +1810,7 @@ lazy val experimentalFlinkTableApiComponents = (project in flink("components/dev
         "org.apache.flink" % "flink-table-api-java-bridge" % flinkV,
         "org.apache.flink" % "flink-table-planner-loader"  % flinkV,
         "org.apache.flink" % "flink-table-runtime"         % flinkV,
+        "org.apache.flink" % "flink-clients"               % flinkV,
```
mslabek, line 1813: This is added since the `flinkTable.jar` is loaded on classpath of minicluster and it needs flink-clients. Ultimately I think we should have a better approach for loading things like minicluster dependencies / formats / connectors, but I don't see any simple solution.

build.sbt
```diff
@@ -2020,7 +2018,10 @@ lazy val e2eTests = (project in file("e2e-tests"))
   .settings {
     // TODO: it'd be better to use scalaVersion here, but for some reason it's hard to disable existing task dynamically
     forScalaVersion(defaultScalaV) {
-      case (2, 12) => doTest
+      case (2, 12) => {
+        System.setProperty("addDevArtifacts", "true")
```
mslabek, line 2022: This is why `addDevArtifacts` was changed to a def. Instead of just setting it on CI workflow definition, I wanted to have the e2eTests/test sbt task behave the same on CI and when launching locally.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/RowToTestDataConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+import pl.touk.nussknacker.engine.flink.table.TableDefinition
+import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
+
+import scala.collection.immutable.ListMap
+
+class RowToTestDataConverter(classLoaderForEncoder: ClassLoader) {
+
+  private val encoder = new BestEffortJsonEncoder(failOnUnknown = false, classLoader = classLoaderForEncoder)
+
+  def rowsToTestData(rows: List[Row], tableDefinition: TableDefinition): TestData = TestData(rows.map(row => {
```
mslabek, line 14: Taking the `tableDefinition` allows for deterministic and controllable ordering of fields. This way its always in the same order as columns in the CREATE TABLE statement.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/RowToTestDataConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.apache.flink.types.Row
+import pl.touk.nussknacker.engine.api.test.{TestData, TestRecord}
+import pl.touk.nussknacker.engine.flink.table.TableDefinition
+import pl.touk.nussknacker.engine.util.json.BestEffortJsonEncoder
+
+import scala.collection.immutable.ListMap
+
+class RowToTestDataConverter(classLoaderForEncoder: ClassLoader) {
+
+  private val encoder = new BestEffortJsonEncoder(failOnUnknown = false, classLoader = classLoaderForEncoder)
```
mslabek, line 12: Working on solving this better through using Flink's json Connector.

build.sbt
mslabek: If we're adding table components to public distribution its not necessary

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+
+class BatchDataGenerationSpec
+    extends AnyFreeSpecLike
+    with DockerBasedInstallationExampleNuEnvironment
+    with Matchers
+    with VeryPatientScalaFutures
+    with NuRestAssureExtensions
+    with NuRestAssureMatchers {
+
+  "Batch scenario generate file function should generate random results according to defined schema" in {
```
coutoPL, line 26: TBH I'm not sure if I understand what we test here.

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    given()
```
coutoPL, line 27: IMO it'd be better if we could create a well-named private function that allows us (without looking into its details and without wondering what `POST /api/processes` does) to easily grasp the sense of the "given" block of this test

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+          |    "category" : "Default",
+          |    "isFragment" : false,
+          |    "processingMode" : "Bounded-Stream"
+          |}
+          |""".stripMargin)
+      .post("http://localhost:8080/api/processes")
+      .Then()
+      .statusCode(201)
+
+    val testResultContent = given()
```
coutoPL, line 43: The same goes for the "when" block. The extracting part should be also encapsulated in there.

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      .preemptiveBasicAuth("admin", "admin")
+      .jsonBody(toScenarioGraph(simpleBatchTableScenario).asJson.spaces2)
+      .post(s"http://localhost:8080/api/testInfo/SumTransactions/generate/10")
+      .Then()
+      .statusCode(200)
+      .extract()
+      .body()
+      .asString()
+
+    val firstLine  = testResultContent.split('\n').head
```
coutoPL, line 55: why only head? what about the rest? there is no rest? or the rest is not important to test?

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -86,9 +99,105 @@ class TableSource(
 //  TODO: add implementation during task with test from file
   override def testRecordParser: TestRecordParser[RECORD] = ???
 
+  override def generateTestData(size: Int): TestData = {
```
coutoPL, line 102: It's rather complicated logic and AFAIS is unrelated to the internals of the class, so maybe we can move it to a separate trait to make this class more readable.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -86,9 +99,105 @@ class TableSource(
...
+    conf
+  }
+
 }
 
 object TableSource {
   private type RECORD = java.util.Map[String, Any]
-  private val filteringInternalViewName = "filteringView"
+  private val filteringInternalViewName                 = "filteringView"
+  private val dataGenerationInputInternalTableName      = "testDataInputTable"
```
coutoPL, line 200: does it mean that two Nu Designer users (working on two different batch scenarios) won't be able to use the Nu Scenario Test functionality at the same time?

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: How about `Generate file endpoint should generate records with randomized values for scenario with table source`?

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: I've extracted the `createBatchScenario` as a separate method

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: I changed it to use a response body matcher extension

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: Rest is not important - I'm just testing if the response body has a record that matches table schema, so it can just be the first record. I changed the api call to request only 1 record

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    val testRecord = parse(firstLine).getOrElse(fail()).asObject.get.toMap("record").asObject.get.toMap
+
+    testRecord("client_id").isString shouldBe true
```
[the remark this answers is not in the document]
mslabek: done :+1:

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
mslabek: True. I extracted it into a separate class - I don't see a benefit to modeling it as a trait.

As to the logic, I've extracted some methods, but its not going to be scala idiomatic since its mostly interacting with Flink's API where it just returns Unit. 

I think this class will change a lot after we add support for testing on data from the generator to accomodate reuse, but thats for another PR.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -86,9 +99,105 @@ class TableSource(
...
+          .forConnector("datagen")
+          .option("number-of-rows", size.toString)
+          .schema(tableDefinition.toFlinkSchema)
+          .build()
+      )
+
+      val tempDir    = Files.createTempDirectory("tableSourceDataDump-")
+      val tempDirUrl = tempDir.toUri.toURL
+      logger.debug(s"Created temporary directory for dumping test data at: '$tempDirUrl'")
+
```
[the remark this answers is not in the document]
mslabek: I've extracted some methods, but its not going to be scala idiomatic since its mostly interacting with Flink's API where it just returns Unit. 

I think this class will change a lot after we add support for testing on data from the generator to accomodate reuse, but thats for another PR.

engine/flink/components/dev-table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
mslabek: Unless there is some minicluster implementation details I don't know about, we are creating a clean environment every time the generate file functionality is called, so it shouldn't be a problem.

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      .when()
+      .request()
+      .basicAuthAdmin()
+      .jsonBody(toScenarioGraph(simpleBatchTableScenario).asJson.spaces2)
+      .post(
+        s"$designerServiceUrl/api/testInfo/${simpleBatchTableScenario.name.value}/generate/$numberOfRecordsToGenerate"
+      )
+      .Then()
+      .statusCode(200)
+      .matchPlainBodyWithRegex(expectedRecordRegex)
```
coutoPL, line 53: WDYT about doing it like that:
```scala
      .body(
        matchJsonWithRegexValues(
          s"""
             |{
             |  "sourceId": "sourceId",
             |  "record": {
             |    "datetime":"${regexes.localDateRegex}",
             |    "client_id":"[a-z\\\\d]{100}",
             |    "amount":"${regexes.decimalRegex}"
             |  }
             |}""".stripMargin
        )
      )
```

IMO it's much more readable. 

And yes, regexes are not so readable, but BTH regex is not readable by definition :P But we can create well-named `val`s and use them here. 

Eg.
```scala
trait NuRestAssureMatchers {

  object regexes {
    val localDateRegex = "^\\\\d{4}-\\\\d{2}-\\\\d{2}\\\\s\\\\d{2}:\\\\d{2}:\\\\d{2}\\\\.\\\\d{3}$$"
    val zuluDateRegex = "^\\\\d{4}-\\\\d{2}-\\\\d{2}T\\\\d{2}:\\\\d{2}:\\\\d{2}(?:\\\\.\\\\d{3,6})?Z$$"
    // ujson treats all numbers as a double
    val digitsRegex = "^\\\\d+\\\\.0$$"
    val decimalRegex = "^\\\\d+(\\\\.\\\\d+)?([eE]\\\\d+)?$$"
  }

```

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+  "Generate file endpoint should generate records with randomized values for scenario with table source" in {
+    val expectedRecordRegex = {
+      val timestampRegex                    = """\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2}\.\d{3}""".r
+      val stringRegex                       = """[a-z\d]{100}""".r
+      val decimalWith15Precision2ScaleRegex = """\d{0,13}\.\d{0,2}""".r
+      val recordRegex =
+        s"""\\{"datetime":"$timestampRegex","client_id":"$stringRegex","amount":$decimalWith15Precision2ScaleRegex}""".r
+      s"""\\{"sourceId":"sourceId","record":$recordRegex}""".r
+    }
+    val numberOfRecordsToGenerate = 1
```
coutoPL, line 38: 1. I'm not sure if inlining the value is not batter than externalized val used only in one place.
2. why only 1?

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -29,12 +29,13 @@ import pl.touk.nussknacker.engine.flink.table.source.TableSource._
 import pl.touk.nussknacker.engine.flink.table.utils.RowConversions
 
 class TableSource(
-    tableDefinition: TableDefinition,
+    val tableDefinition: TableDefinition,
```
coutoPL, line 32: I'm not sure if we really need to expose it

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterDataGenerator.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import java.net.URL
+import java.nio.charset.StandardCharsets
+import java.nio.file.{DirectoryStream, Files, Path}
+import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Success, Try}
+
+class FlinkMiniClusterDataGenerator(flinkTableSchema: Schema) extends LazyLogging {
+
+  // TODO: check the minicluster releases memory properly and if not, refactor to reuse one minicluster per all usages
```
coutoPL, line 21: I see that at least an executor is created. I see no method to release the resources tho :(
But maybe we can create the `env` in the constructor of this class? At least each run of test generation within the same instance of node, won't create the mini cluster. 

(obviously, in the future, we could do as it's described in the comment)

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterDataGenerator.scala
```diff
@@ -0,0 +1,132 @@
...
+  def generateTestData(amountOfRecordsToGenerate: Int): TestData =
+    // setting context classloader because Flink in multiple places relies on it and without this temporary override it doesnt have
+    // the necessary classes
+    ThreadUtils.withThisAsContextClassLoader(getClass.getClassLoader) {
+      val env = TableEnvironment.create(miniClusterTestingEnvConfig)
+      createGeneratorTable(env, amountOfRecordsToGenerate)
+      val (tempDirForOutputTable, tempDirUrl) = createTempDirectoryForTestResults()
```
coutoPL, line 28: test results or test data?

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterDataGenerator.scala
```diff
@@ -0,0 +1,132 @@
...
+  private def insertDataAndAwait(env: TableEnvironment, inputTableName: String, outputTableName: String): Unit = {
+    val inputTable = env.from(inputTableName)
+    inputTable.insertInto(outputTableName).execute().await()
+  }
+
+  private def createTempDirectoryForTestResults(): (Path, URL) = {
+    val tempDir    = Files.createTempDirectory(tempTestDataOutputTablePrefix)
+    val tempDirUrl = tempDir.toUri.toURL
+    logger.debug(s"Created temporary directory for dumping test data at: '$tempDirUrl'")
+    tempDir -> tempDirUrl
```
coutoPL, line 66: you can convert Path to URL whenever you want (and it's a rather light operation), so I don't think we should return the tuple here

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterDataGenerator.scala
```diff
@@ -0,0 +1,132 @@
...
+  private def createGeneratorTable(env: TableEnvironment, amountOfRecordsToGenerate: Int): Unit = env.createTable(
+    dataGenerationInputInternalTableName,
+    TableDescriptor
+      .forConnector("datagen")
+      .option("number-of-rows", amountOfRecordsToGenerate.toString)
+      .schema(flinkTableSchema)
+      .build()
+  )
+
+  private def createOutputFileTable(env: TableEnvironment, tempDirUrl: URL): Unit = env.createTable(
```
coutoPL, line 78: maybe we should generate the temp dir inside this method and instead of returning Unit, we will return the file where the output for table will be stored?

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterDataGenerator.scala
mslabek: Ok, when I implemented this, I checked that the TableSource and FlinkMiniClusterDataGenerator are initialized each time the test data generation is called :/ So even when we reuse the minicluster on FlinkMiniClusterDataGenerator or even component level, it doesn't matter :/ Outside of some larger refactor I don't see a solution

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
coutoPL: First of all, the media type for this kind of response should not be a text by application/x-ndjson. 
I think we should create `MatchNdJsonWithRegexValues` based on `MatchJsonWithRegexValues`. 

Currently, we expect that the body is a JSON (and we parse the body to JSON). And the expected part is a single string which is being read as JSON (with potential regex fields).

In the new `MatchNdJsonWithRegexValues` the body will be split by `\n` first and then each line will be parsed as JSON. The expected value will be a list of strings (not a single string).

In this solution we don't compare strings but JSON objects, so we can format the JSON string (in the `then` part) as we want.

utils/test-utils/src/main/scala/pl/touk/nussknacker/test/NuRestAssureMatchers.scala
```diff
@@ -148,7 +153,8 @@ object NuRestAssureMatchers extends NuRestAssureMatchers {
     ): Validated[String, Unit] = {
       stringToRegex(expectedValueRegex).map { regex =>
         currentValueString match {
-          case regex() => Validated.Valid(())
+          case regex()   => Validated.Valid(())
+          case regex(_*) => Validated.Valid(())
```
mslabek, line 157: without this case the check failed in cases where groups were used

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: Done :+1: Had to adjust the matching logic a bit to enable groups - https://github.com/TouK/nussknacker/pull/6195#discussion_r1670475283

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  private val simpleBatchTableScenario = ScenarioBuilder
+    .streaming("SumTransactions")
+    .source("sourceId", "table", "Table" -> "'transactions'".spel)
+    .emptySink("end", "dead-end")
+
+  private val designerServiceUrl = "http://localhost:8080"
+
+  "Generate file endpoint should generate records with randomized values for scenario with table source" in {
+    val expectedRecordRegex =
```
coutoPL, line 31: now, we can inline the "json-regex" ;) The reader will pay attention to this when they are reading the "then" part. Currently, the reader sees it, when they start to read code of the test, but it'll be relevant later - almost at the end of the test.

coutoPL (approved):
one minor comment. 
LGTM! Thanks

e2e-tests/src/test/scala/pl/touk/nussknacker/BatchDataGenerationSpec.scala
mslabek: right, done :+1:

PR #6211, opened 2024-06-18
Do not mark scenario as finished when there is no job at Flink

DeamonDev:
## Checklist before merge

### DIFF
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
@@ -242,6 +242,8 @@ class PeriodicProcessService(
             status
           ) && deployment.state.status != PeriodicProcessDeploymentStatus.Finished =>
         markFinished(deployment, processState).needsReschedule(value = true)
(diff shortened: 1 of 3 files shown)


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -242,8 +242,6 @@ class PeriodicProcessService(
             status
           ) && deployment.state.status != PeriodicProcessDeploymentStatus.Finished =>
         markFinished(deployment, processState).needsReschedule(value = true)
-      case None if deployment.state.status == PeriodicProcessDeploymentStatus.Deployed =>
```
jedrz, line 245: I think by removing this case we would introduce the risk of the job staying in the deployed status indefinitely, even though it may never actually be deployed to Flink, or it may finish or fail and disappear from the web console after the Nussknacker designer has exited.

Perhaps we should consider introducing a new status - deploying? We would wait for a running job to appear within, e.g. 10 minutes. If the job does not appear, we would mark it as failed.

WDYT?

DeamonDev: That's good idea. But I would propose to expand it in another PR.

PR #6213, opened 2024-06-18
Improvement: Load resource config only in test context

lciolecki:
## Checklist before merge

### DIFF
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -23,7 +23,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 * [#6437](https://github.com/TouK/nussknacker/pull/6437) Removed deprecated operation to create a scenario:
   POST `/api/processes/{name}/{category}`. POST `/api/processes` should be used instead.
 * [#6213](https://github.com/TouK/nussknacker/pull/6213) Improvement: Load resource config only in test context
-  * Default behaviour for `WithConfig` was broken, and now for load application.conf file from test resources you need to override `configFilename`
+  * `WithConfig` from `test-utils` modules behaviour changes: now it only parses given config, 
+    without resolving reference configs, system env variables etc.
 
 ## In version 1.16.0
 
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -22,6 +22,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 * [#6437](https://github.com/TouK/nussknacker/pull/6437) Removed deprecated operation to create a scenario:
   POST `/api/processes/{name}/{category}`. POST `/api/processes` should be used instead.
+* [#6213](https://github.com/TouK/nussknacker/pull/6213) Improvement: Load resource config only in test context
+  * Default behaviour for `WithConfig` was broken, and now for load application.conf file from test resources you need to override `configFilename`
 
 ## In version 1.16.0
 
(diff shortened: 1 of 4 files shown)


utils/test-utils/src/main/scala/pl/touk/nussknacker/test/WithConfig.scala
```diff
@@ -1,10 +1,24 @@
...
+import scala.util.Try
 
 trait WithConfig {
 
-  protected lazy val config: Config = resolveConfig(ConfigFactory.load())
+  // Here is the place where we can change the entry config file
+  protected val configFilename: String = "application.conf"
+
+  protected lazy val config: Config = {
+    val config = Try(IOUtils.resourceToString(s"/$configFilename", StandardCharsets.UTF_8)).toOption
```
arkadius, line 16: Why this `Try().toOption`? It is always better to fail fast instead of postponing the problem to the place where you don't know what's going on

lciolecki: Hmm, we can do that. But I saw that we have some tests without this file and exception was thrown in these tests..

arkadius: Hmm, I think that I'm getting your intention now. Before your change, was loaded default chain of configs, which essentially was an empty config for such a case. If we resign from usage of this chain and do everything by our own, we just check if file exists and if not, return empty config

lciolecki: image
Old version if applicaiton.conf doesn't exit in test resource creates empty / default config..

lciolecki: My proposition:

```
  // Here is the place where we can change the entry config file
  protected val configFilename: Option[String] = None

  protected lazy val config: Config = {
    val config = configFilename
      .map(file => IOUtils.resourceToString(s"/$file", StandardCharsets.UTF_8))
      .map(ConfigFactory.parseString)
      .getOrElse(ConfigFactory.empty())

    resolveConfig(config.resolve())
  }
  ```
  WDYT?

Comment (arkadius) [reply utils/test-utils/src/main/scala/pl/touk/nussknacker/test/WithConfig.scala:16]: Looks good

File: docs/Changelog.md [hunk 2]
```diff
@@ -26,6 +26,7 @@
 * [#6437](https://github.com/TouK/nussknacker/pull/6437) Removed deprecated operation to create a scenario:
   POST `/api/processes/{name}/{category}`. POST `/api/processes` should be used instead.
 * [#6415](https://github.com/TouK/nussknacker/pull/6415) Added "Component group" field to fragment properties, which allows selection of the group of components in the Creator Panel in which the fragment will be available
+* [#6213](https://github.com/TouK/nussknacker/pull/6213) Improvement: Load resource config only in test context
```
arkadius, line 29: There is something wrong with this explanation, I don't understand it :) TBH I don't know if it is worth to mention about this change to our users, this test-utils module is rather an internal thing, it generates only a noise for users, migration guide entry is enough for me.

docs/MigrationGuide.md
```diff
@@ -16,6 +16,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   * Now method `returnType` from `EagerServiceWithStaticParameters` requires implicit nodeId param
 * [#6437](https://github.com/TouK/nussknacker/pull/6437) Removed deprecated operation to create a scenario:
   POST `/api/processes/{name}/{category}`. POST `/api/processes` should be used instead.
+* [#6213](https://github.com/TouK/nussknacker/pull/6213) Improvement: Load resource config only in test context
```
arkadius, line 19: `WithConfig` from test-utils modules behaviour changes: it only parse given config, without resolving reference configs, system variables etc.

arkadius (approved):

PR #6238, opened 2024-06-21
Fix FE ScenarioPropertyConfig's attribute to align to the BE response

Elmacioro:
## Describe your changes

Currently there is a missmatch between what FE expects and BE provides while fetching `ProcessDefinitionData`.
Frontend's `ScenarioPropertiesConfig` contains attribute `values: Array<PossibleValue>` however backend does not return that. Instead it returns `defaultValue: Option[String]` which is ignored by the FE.

The `values` attribute isn't utilized on Frontend so I aligned the attribute to the BE response:
```  
@JsonCodec final case class UiScenarioPropertyConfig(
      defaultValue: Option[String],
      editor: ParameterEditor,
      label: Option[String]
  )
````
Zrzut ekranu z 2024-06-21 09-24-17

I noticed that mismatch when I had to use the `defaultValue` in an external project. 

## Checklist before merge


Dzuming (approved):
LGTM

PR #6248, opened 2024-06-21
[MAINTENANCE] Remove implicit conversion from string to spel

lukasz-bigorajski:
## Checklist before merge


components/sql/src/test/scala/pl/touk/nussknacker/sql/service/DatabaseLookupLiteRuntimeTest.scala
```diff
@@ -58,12 +58,12 @@ class DatabaseLookupLiteRuntimeTest
         "sql-lookup-enricher",
         "output",
         "sql-lookup-enricher",
-        "Table"      -> "'PERSONS'",
-        "Key column" -> "'ID'",
-        "Key value"  -> "#input",
-        "Cache TTL"  -> ""
+        "Table"      -> "'PERSONS'".spel,
```
lciolecki, line 61: I wonder, why just not pass "PERSON" and let's do this `'` magic in .spel? :)

lukasz-bigorajski: I think it's not possible because we pass every expression to spel as String and we have to have a way to distinguish between a string and function or variable

lciolecki: See `LiteralSpEL`

lukasz-bigorajski: What about that expression: `"#input != null".spel`? This expression cannot be quoted and LiteralSpEL will quote this. IMHO what we can do is to add another method `asStringSpel` which will add quote to passed string.

benchmarks/src/test/scala/pl/touk/nussknacker/engine/benchmarks/interpreter/ManyParamsInterpreterBenchmark.scala
```diff
@@ -8,7 +8,7 @@ import pl.touk.nussknacker.engine.api.{Context, MethodToInvoke, ParamName, Servi
 import pl.touk.nussknacker.engine.build.ScenarioBuilder
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
 import pl.touk.nussknacker.engine.graph.expression.Expression
-import pl.touk.nussknacker.engine.spel.Implicits._
```
arkadius, line 11: I can't see removal of Implicits object. Do you removed it? If not, please remove the dead code. Also, please add a migration guide entry.

mk-software-pl (approved):
LGTM

but migration guide entry with information about removing the implicit conversion and guidance that we should use SpelExtension instead would be nice to have ;)

lukasz-bigorajski:
> LGTM
> 
> but migration guide entry with information about removing the implicit conversion and guidance that we should use SpelExtension instead would be nice to have ;)

https://github.com/TouK/nussknacker/pull/6248/files#diff-2a20a169a727b678d0b8fbac5f7e82f36da9900b8ef7cc82692e7fd703e05ebc

arkadius (approved):

PR #6258, opened 2024-06-24
[NU-1719] Component statistics without restmodel

ForrestFairy:
## Describe your changes
Remove creating `ComponentListElement` DTOs just to get component usages.
Instead, map of usages per `DesignerWideComponentId` is provided by `ComponentService` and then mapped by id to available components

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/component/ComponentService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/component/ComponentService.scala
@@ -32,6 +32,11 @@ trait ComponentService {
       implicit user: LoggedUser
   ): Future[XError[List[ComponentUsagesInScenario]]]
 
+  def getUsagesPerDesignerWideComponentId(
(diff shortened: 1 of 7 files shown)


lukasz-bigorajski (commented):
About counting components stats I will write later

lukasz-bigorajski:
Having 2 logics for counting usages can be misleading and error prone so we should use logic that we have currently. I am looking at componentsService now and I wonder if it will be enough to add method:
```
def getComponentsUsages(implicit user: LoggedUser): Future[Map[DesignerWideComponentId, Long]]
```
and then:
```
  override def getComponentsUsages(implicit user: LoggedUser): Future[Map[DesignerWideComponentId, Long]] =
    getUserAccessibleComponentUsages
```
but I have to confirm how is matched `DesignerWideComponentId` and `ComponentId`

lukasz-bigorajski (approved):
LGTM

Have you manually tested this components usage count on local nu?

PR #6282, opened 2024-06-27
[NU-1682] topic auto create changes

mk-software-pl:
## Describe your changes

- `topicsExistenceValidationConfig.enabled` default is `true` now
- Kafka setting `auto.create.topics.enable` is only taken into consideration in the case of Sink's topics existence validation. So when `topicsExistenceValidationConfig.enabled: true`:
  - the existence of a topic used in a Source will be checked always  
  - the existence of a topic used in a Sink will be checked when `auto.create.topics.enable: false` is set in Kafka

## Checklist before merge


examples/installation/.env
```diff
@@ -1 +1 @@
-NUSSKNACKER_VERSION=1.15.1
+NUSSKNACKER_VERSION=1.15.3
```
mslabek, line 1: 1. How does this bump relate to the synchronization of external repo https://github.com/TouK/nussknacker-installation-example? I thought this had to be merged first into release branch and then released through GH action? But now I see that in that repo it has version 1.15.3.

2. When should this be bumped? Is it part of the release procedure?

utils/kafka-components-utils/src/it/scala/pl/touk/nussknacker/engine/kafka/validator/CachedTopicsExistenceValidatorTest.scala
```diff
@@ -1,115 +1,130 @@
...
+      TopicExistenceValidationException(NonEmptyList.one(notExistingSourceTopic))
+    )
   }
 
   test("should not validate not existing topic when validation disabled") {
-    val v = new CachedTopicsExistenceValidator(
-      kafkaConfig.copy(
+    val validator = new CachedTopicsExistenceValidator(
+      defaultKafkaConfig.copy(
         kafkaProperties = Some(Map("bootstrap.servers" -> "broken address")),
```
mslabek, line 37: i think this broken address is redundant and makes test fail for the wrong reason - `Failed to create new KafkaAdminClient`. but after commenting it out it fails for the right reason.

engine/flink/management/src/it/scala/pl/touk/nussknacker/engine/management/streaming/FlinkStreamingDeploymentManagerSpec.scala
```diff
@@ -72,7 +72,7 @@ class FlinkStreamingDeploymentManagerSpec extends AnyFunSuite with Matchers with
   test("cancel of not existing job should not fail") {
     deploymentManager
       .processCommand(DMCancelScenarioCommand(ProcessName("not existing job"), user = userToAct))
-      .futureValue shouldBe (())
+      .futureValue shouldBe ()
```
mslabek, line 75: After this when compiling with scala 2.12 in IDEA I get:
```
Adaptation of argument list by inserting () is deprecated: leaky (Object-receiving) target makes this especially dangerous.
        signature: AnyShouldWrapper.shouldBe(right: Any): org.scalatest.Assertion
  given arguments: <none>
 after adaptation: AnyShouldWrapper.shouldBe((): Unit)
      .futureValue shouldBe ()
```

utils/kafka-components-utils/src/it/scala/pl/touk/nussknacker/engine/kafka/validator/CachedTopicsExistenceValidatorTest.scala
```diff
@@ -1,115 +1,130 @@
...
-    kafkaEspProperties = None,
-    consumerGroupNamingStrategy = None,
-    avroKryoGenericRecordSchemaIdSerialization = None,
-    topicsExistenceValidationConfig = TopicsExistenceValidationConfigForTest.config
-  )
+    extends BaseCachedTopicsExistenceValidatorTest(
+      kafkaAutoCreateEnabled = false
+    ) {
 
   test("should validate existing topic") {
```
mslabek, line 21: For consistency with tests that have autoCreateEnabled=true I'd change this test name to be `should validate existing source topic` and `should validate not existing sink topic` for the next one. Identical names for these tests will be more readable since name of the test class is also printed and from there the reader will find out the difference was `autoCreate` setting.

docs/Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+1.17.0 (Not released yet)
+* [6282](https://github.com/TouK/nussknacker/pull/6184) `topicsExistenceValidationConfig.enabled` default was changed from `false` to `true`
```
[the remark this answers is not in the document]
mk-software-pl: 1. the info about the change from the perspective of the user. There is no distinction on the type of users yet.
2. ok, added.

examples/installation/docker-compose.yml
```diff
@@ -107,25 +107,17 @@ services:
 
   ### KAFKA-related services:
 
-  zookeeper:
```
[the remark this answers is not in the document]
mk-software-pl: yes, it should work now (but maybe the additional step with creating the topics will be needed).

examples/installation/.env
mk-software-pl: The pipeline script, during publishing, should set the proper version. The e2e tests also set the current build version. TBH we don't need it. I'm removing it.

utils/kafka-components-utils/src/it/scala/pl/touk/nussknacker/engine/kafka/validator/CachedTopicsExistenceValidatorTest.scala
mk-software-pl: Good point. They are more symmetrical now.

mslabek (approved):

PR #6302, opened 2024-07-01
[NU-6301] Restrict SpEL expression based on caret position

DeamonDev:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/suggester/ExpressionSuggester.scala
```diff
@@ -30,6 +30,25 @@ class ExpressionSuggester(
       scenarioPropertiesNames
     )
 
+  private def restrictExpressionByCaretPosition2d(
```
bartektartanus, line 33:
```suggestion
  private def truncateExpressionByCaretPosition2d(
```

bartektartanus:
Please add some test for that new case.

designer/server/src/main/scala/pl/touk/nussknacker/ui/suggester/ExpressionSuggester.scala
```diff
@@ -30,6 +30,25 @@ class ExpressionSuggester(
       scenarioPropertiesNames
     )
 
+  private def truncateExpressionByCaretPosition2d(
```
arkadius, line 33: All heuristics should be in internals (implementation class) as it is a part of the code that would change frequently. Also it would be nice to keep all heuristics somehow organized to help diagnosis what happen in some surprising situations - see `shouldInsertDummyVariable`

designer/server/src/main/scala/pl/touk/nussknacker/ui/suggester/ExpressionSuggester.scala
```diff
@@ -39,11 +58,23 @@ class ExpressionSuggester(
...
-          validationContext
-        )
+        spelExpressionSuggester
+          .expressionSuggestions(
+            expression,
+            caretPosition2d.normalizedCaretPosition(expression.expression),
+            validationContext
+          )
+          .flatMap {
+            case Nil =>
```
arkadius, line 68: `Nil` is not a good marker for invalid expression. We will retry suggestions even for normal cases. This heuristic should be applied as a final option, when all clean solutions can't handle it.

arkadius (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/suggester/ExpressionSuggester.scala
arkadius: This comment is still valid

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
```diff
@@ -34,12 +37,47 @@ class SpelExpressionSuggester(
...
+    val futureSuggestionsOption =
+      expressionSuggestionsAux(
+        expression,
+        caretPosition2d.normalizedCaretPosition(expression.expression),
+        validationContext
+      )
+
+    val newExpression: Expression = truncateExpressionByCaretPosition2d(expression, caretPosition2d)
+
+    val futureSuggestionsAfterTruncatingExpressionByCaretPositionOption = expressionSuggestionsAux(
```
arkadius, line 55: This code should be invoked lazyly, only when the first method doesn't work.

arkadius (approved):

PR #5907, opened 2024-04-19
Gracefully stop deployment actors (and wait for free names) on PeriodicDeploymentManager close()

wrzontek:
## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [ ] Code is cleaned from temporary changes and commented out lines
- [ ] Parts of the code that are not easy to understand are documented in the code
- [ ] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [ ] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #6314, opened 2024-07-03
Stop deployment and reschedule actors less gracefully on reload

wrzontek:
## Describe your changes
The changes in https://github.com/TouK/nussknacker/pull/5907 proved insufficient, because `gracefulStop` patiently waits for the actor to finish processing ALL it's queued messages, which is a bit too graceful.

After the change, it will only finish it's currently processing message then stop.

## Checklist before merge

### DIFF
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/Utils.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/Utils.scala
@@ -1,16 +1,16 @@
 package pl.touk.nussknacker.engine.management.periodic
 
-import akka.actor.{ActorNotFound, ActorPath, ActorRef, ActorSystem}
+import akka.actor.{ActorRef, ActorSystem, Props}
 import com.typesafe.scalalogging.LazyLogging
 
 import scala.concurrent.duration.DurationInt
-import scala.concurrent.{Await, ExecutionContext}
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 object Utils extends LazyLogging {
 
-  private val ActorResolutionTimeout = 10 seconds
-  private val ActorResolutionPause   = 50 milliseconds
-  private val ActorResolutionRetries = 50
+  private val ActorCreationPause   = 50 milliseconds
+  private val ActorCreationRetries = 50
 
   def runSafely(action: => Unit): Unit = try {
     action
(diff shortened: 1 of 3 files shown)


wrzontek:
Also this could probably be put into 1.16

wrzontek:
Note: clearing the queue for `DeploymentActor` and `RescheduleFinishedActor` isn't a problem, as all they're doing is calling methods from `PeriodicProcessService` on interval, they don't have any internal state

wrzontek:
locally it's all good, something weird with CI

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/Utils.scala
```diff
@@ -22,24 +18,25 @@ object Utils extends LazyLogging {
...
 
-  def gracefulStopActor(actorRef: ActorRef, actorSystem: ActorSystem): Unit = {
+  def stopActorAndWaitUntilItsNameIsFree(actorRef: ActorRef, actorSystem: ActorSystem): Unit = {
     import actorSystem.dispatcher
-    logger.info(s"Gracefully stopping $actorRef")
+    logger.info(s"Stopping $actorRef")
 
-    val gracefulStopFuture = for {
-      _ <- gracefulStop(actorRef, GracefulStopTimeout)
+    actorSystem.stop(actorRef)
```
arkadius, line 25: Why so brutal stop? Shouldn't we try with graceful stop and after some timeout eventually perform less graceful stop? We need at least some comment with explanation to stop someone from refactoring it to the previous version ;-)

engine/flink/management/periodic/src/test/scala/pl/touk/nussknacker/engine/management/periodic/UtilsSpec.scala
```diff
@@ -24,10 +30,38 @@ class UtilsSpec extends TestKit(ActorSystem("UtilsSpec")) with AnyWordSpecLike w
...
+      while (messageCounter < 1000) {
+        actorRef ! s"message number $messageCounter"
+        messageCounter += 1
+      }
+
+      Thread.sleep(1000)
+      // with gracefulStop this times out, because the PoisonPill is queued after the many normal messages
+      Utils.stopActorAndWaitUntilItsNameIsFree(actorRef, system)
+
+      val actorRef2 = system.actorOf(Props(new TestActor), actorName)
```
arkadius, line 62: Hmm this new test is failed.
image

wrzontek: it only fails on CI, and I don't get why
are the tests on CI somehow "more parallel" than locally?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/Utils.scala
wrzontek: this stop still allows it to finish processing current message, and it's sufficient as I explained here: https://github.com/TouK/nussknacker/pull/6314#issuecomment-2208580681

I could add some comments to the code i guess

engine/flink/management/periodic/src/test/scala/pl/touk/nussknacker/engine/management/periodic/UtilsSpec.scala
wrzontek: I do get why, it's a race condition

i'll avoid it by just retrying actor creation instead of retrying actor stopping+freeing

engine/flink/management/periodic/src/test/scala/pl/touk/nussknacker/engine/management/periodic/UtilsSpec.scala
```diff
@@ -2,32 +2,104 @@ package pl.touk.nussknacker.engine.management.periodic
...
+        Thread.sleep(1000)
+
+        // if this was gracefulStop, it would take too long to stop the actor, as it would continue processing it's messages
+        system.stop(actorRef)
+      }
+
+      Await.result(Future.sequence(Seq(futureA, futureB)), Duration.Inf)
+    }
+
+    "fail to create an actor if it's name isn't freed" in {
```
wrzontek, line 92: we could change the behaviour to also call `system.stop(actor)` if it's taken, this way `def createActorWithRetry` would always succeed, but perhaps that's more powerful than what's needed, as it's only ever called after the actors are told to stop by `toClose` anyway

arkadius (approved):
LGTM

PR #6320, opened 2024-07-04
[NU-1679] table join component

arkadius:
## Checklist before merge


engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinTest.scala
```diff
@@ -0,0 +1,105 @@
...
+            ),
+            "output" -> "#input".spel,
+          )
+          .emptySink("end", TestScenarioRunner.testResultSink, "value" -> "{#input, #product}".spel)
+      )
+
+    val result = runner.runWithData(
+      scenario,
+      List(
+        OrderProduct("product", 1, -1),
```
raphaelsolarski, line 76: maybe we can add additional input rows to test whether join is done properly? e.g. orders without products and products without order.

lukasz-bigorajski (approved):
LGTM

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinTest.scala
```diff
@@ -0,0 +1,105 @@
...
+    result.validValue.successes shouldBe List(
+      List(OrderProduct("order", 10, 1), OrderProduct("product", 1, -1)).asJava,
+    )
+  }
+
+}
+
+object TableJoinTest {
+
+  // TODO: split into separate classes and pass two streams to separate source nodes
```
raphaelsolarski, line 92: Yes, I'd split e.g. to Order, Product, and OrderWithProduct.
At first glance I thought that you implemented union instead of join (the nested list in the output is not so visible)

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinComponent.scala
```diff
@@ -0,0 +1,241 @@
...
+  private def joinedBranchTypeInfo(flinkNodeContext: FlinkCustomNodeContext, outputLazyParam: LazyParameter[_]) = {
+    Types.ROW_NAMED(
+      Array(joinedKeyInternalColumnName, outputInternalColumnName),
+      Types.STRING,
+      flinkNodeContext.typeInformationDetection.forType(outputLazyParam.returnType)
+    )
+  }
+
+  private def extractMainBranchId(branchTypeByBranchId: Map[String, BranchType]) = {
+    branchTypeByBranchId.find(_._2 == BranchType.MAIN).map(_._1)
```
raphaelsolarski, line 234: Was ist das? `_._2` `_._1`?

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/RowConversions.scala
```diff
@@ -22,6 +32,32 @@ object RowConversions {
     row
   }
 
+  private def scalaMapToRow(map: Map[String, Any]): Row = {
+    val row = Row.withNames()
+    // TODO: add type alignment, e.g. bigint is represented as long by flink tables
```
raphaelsolarski, line 37: is it still valid?

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinTest.scala
arkadius: I added more rows, these pointed by you, but also case of product published after order (delayed product). After adding it, I realize that it would be nice to have a join type parameter (inner / outer) in this component so I also added it

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinTest.scala
arkadius: So far, I only polished a bit current solution (more meaningful class name, factory methods). I also changed returned type to record - take a look. Splitting it into two separate classes would be expensive - we have to add some branch-aware methods to scenario runners, so I didn't do it.

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/join/TableJoinComponent.scala
arkadius: Fixed :P

raphaelsolarski (approved):

PR #6363, opened 2024-07-10
[NU-6301] Improvement on SpEL suggestions

DeamonDev:
## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
```diff
@@ -50,16 +53,48 @@ class SpelExpressionSuggester(
...
+        expressionSuggestionsAux(
+          e,
+          e.expression.length,
+          validationContext
+        )
+      )
 
-    val firstNonEmptySuggestionFuture =
-      futureSuggestionsOption orElse futureSuggestionsAfterTruncatingExpressionByCaretPositionOption getOrElse successfulNil
+    def processOptions[A](options: LazyList[Option[Future[Iterable[A]]]]): Future[Iterable[A]] = {
```
mslabek, line 68: The name `processOptions` is pretty generic. How about something more concrete like 'firstNonEmptyFuture'? And maybe extract this as a private method to make this `expressionSuggestions` method more readable?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ExpressionSuggesterSpec.scala
mslabek, line 1: If we're going with the limitation of suggesting inside string then could you add a test that shows that this is a limitation we know about?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSuggester.scala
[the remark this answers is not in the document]
DeamonDev: We do not have full control over types: 

`val futureSuggestionsOption: Option[Future[Iterable[ExpressionSuggestion]]]`

And we create: 

`val lazyListOfFutureSuggestions: LazyList[Option[Future[Iterable[ExpressionSuggestion]]]`
which is an input of this method.

mslabek (approved):

PR #6389, opened 2024-07-15
counts refresh

JulianWielga:
## Describe your changes

https://github.com/user-attachments/assets/63d2b9f4-8a96-4ddb-8e7d-39cf81ad14bb

https://github.com/user-attachments/assets/61257d76-c77b-4ef1-be0f-8477469f7143

https://github.com/user-attachments/assets/f6368bb8-de86-4f79-9b51-faeec99e258a

## Checklist before merge


Dzuming:
Good job, I like this functionality. I found a small things to consider

1. I selected the autorefresh option but when I opened counts again, the refresh disabled is visible. Could it be possible to display the currently selected refresh value somehow?

https://github.com/user-attachments/assets/a7ec5e93-d9fe-4261-a2e2-a8274c8808e6


2. I'm able to run counts in the fragment when passing query parameters. If it's easy to validate, we probably could block this situation, but it's rather a minor issue

image

designer/client/test/VisualizationUrl-test.js
```diff
@@ -33,4 +33,15 @@ describe("validating url counts extractors", () => {
...
+
+    it("extract duration from string", () => {
+        expect(extractCountParams({ to: "0", refresh: "1s" }).refreshIn).toBe(1);
+        expect(extractCountParams({ to: "0", refresh: "1m" }).refreshIn).toBe(60);
+        expect(extractCountParams({ to: "0", refresh: "1h" }).refreshIn).toBe(3600);
+
+        expect(extractCountParams({ to: "0" }).refreshIn).toBe(false);
+        expect(extractCountParams({ to: "0", refresh: "" }).refreshIn).toBe(false);
+        expect(extractCountParams({ to: "0", refresh: "1000" }).refreshIn).toBe(false);
+        expect(extractCountParams({ to: "0", refresh: "1d" }).refreshIn).toBe(false);
```
Dzuming, line 45: "1d" is not a valid refresh value but "24h" (or more) is fine

JulianWielga: this param is just iso part after T. it doesn't have to be bulletproof. Durations like 1d, even 6h, make no sense and i don't think we should care such users. harmless, not dangerous, not worth attention

JulianWielga:
> 1. I selected the autorefresh option but when I opened counts again, the refresh disabled is visible. Could it be possible to display the currently selected refresh value somehow?

counts window has only options for counts action - not the state of fetched/displayed/refreshed counts data.
I don't want 
> 2. I'm able to run counts in the fragment when passing query parameters. If it's easy to validate, we probably could block this situation, but it's rather a minor issue

Dzuming:
> counts window has only options for counts action - not the state of fetched/displayed/refreshed counts data.
I don't want to store this state now. I changed "refresh disabled" to "no refresh" to avoid "state" association.

That makes sense, initially when I clicked through the feature, I had a feeling that it was some issue, that refresh is enabled but I can see this `refresh disabled` information. The change in the text resolved the problem.

PR #6409, opened 2024-07-17
Add dynamic import retry mechanism for component icon

Elmacioro:
## Describe your changes

Very rarely when users enter the Components tab there is an error during dynamic import of component icons.
It escapes the ErrorBoundary and user is presented with a React stacktrace.

In these changes I introduced more logging to better understand the nature of the problem as well as import retry with fallback component, so frontend tries one more time to present the user with correct data.

## Checklist before merge


designer/submodules/packages/components/src/common/nuIcon.tsx
```diff
@@ -1,8 +1,20 @@
...
 import { styled } from "@mui/material";
+import { CrackedNut } from "./CrackedNut";
 
-const Icon = loadable(async () => import("nussknackerUi/Icon"));
+// This retry mechanism was introduced as sometimes dynamic import would fail.
+const Icon = loadable(() =>
+    import("nussknackerUi/Icon").catch(async (error) => {
+        console.warn("Failed to import Icon submodule. Import retry will be carried out", error);
+        try {
+            return await import("nussknackerUi/Icon");
```
lciolecki, line 11: Only once? :( At least 3?

designer/submodules/packages/components/src/common/nuIcon.tsx
```diff
@@ -1,8 +1,20 @@
...
+        } catch (e) {
+            console.error("Import retry failed", e);
+            return { default: () => <CrackedNut /> };
```
lciolecki, line 14: This cracked nut isn't good UX stuff :( Maybe we can show user something more friendly?

Dzuming (approved):
LGTM

designer/submodules/packages/components/src/common/nuIcon.tsx
Elmacioro: This retry mechanism propably isn't a final solution as we do not fully understand why the promise sometimes rejects and whether one or even three retries would help with it.
With this approach we want to see whether a retry would even help and also gather some more information with the console logs.

designer/submodules/packages/components/src/common/nuIcon.tsx
Elmacioro: As i mentioned these changes aim to gather some more information about the root of the issue and check whether simple retries would help. If they don't then we need some kind of a signal for the user on the screen so they would check the logs that we could analyze. This cracked nut is not as flashy as in some places. It will be displayed in the place of the problematic icon and the app would still be usable.

PR #6415, opened 2024-07-18
Fragments `componentGroup` can be changed via scenario properties

wrzontek:
## Describe your changes

Screenshot from 2024-07-24 11-43-45

Screenshot from 2024-07-24 11-43-49

Screenshot from 2024-07-24 11-47-46

Screenshot from 2024-07-24 11-46-05

## Checklist before merge


wrzontek:
CI green aside from flaky tests.

designer/server/src/main/scala/db/migration/V1_054__FragmentAddComponentGroupProperty.scala
```diff
@@ -0,0 +1,28 @@
...
+import io.circe._
+import pl.touk.nussknacker.ui.db.migration.ProcessJsonMigration
+
+trait V1_054__FragmentAddComponentGroupPropertyDefinition extends ProcessJsonMigration {
+
+  override def updateProcessJson(jsonProcess: Json): Option[Json] =
+    V1_054__FragmentAddComponentGroupProperty.migrateMetadata(jsonProcess)
+}
+
+object V1_054__FragmentAddComponentGroupProperty {
```
arkadius, line 12: Is it needed? Can't we use "in-flight migration approach" I mean `FragmentSpecificData.apply` would have somewhere getOrElse (default group). BTW it is safer to not store the group until it won't be changed. We don't have separate id / name for component groups. So once we store this "fragments" group in some scenario, changing it to e.g. "Fragments" will be painfull.

wrzontek: I considered "in-flight", but that means that when you edit a fragment from before this change, and go into it's properties, the "Component group" field will be empty

if i understand correctly, it's taken straight from the process json which is taken straight from the database without altering, so to set a default on BE would require modyfing it between the DB and FE when fetching process, which currently isn't done

that's why i chose migration as the lesser evil

wrzontek: image
 :D

arkadius: Ehh, you're right

arkadius (approved):
LGTM

PR #6429, opened 2024-07-22
Improvement: Add StreamComponent interface

lciolecki:
## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
@@ -61,6 +61,11 @@ trait BoundedStreamComponent { self: Component =>
     AllowedProcessingModes.SetOf(ProcessingMode.BoundedStream)
 }
 
+trait StreamComponent { self: Component =>
+  override def allowedProcessingModes: AllowedProcessingModes =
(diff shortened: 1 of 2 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -61,6 +61,11 @@ trait BoundedStreamComponent { self: Component =>
     AllowedProcessingModes.SetOf(ProcessingMode.BoundedStream)
 }
 
+trait StreamComponent { self: Component =>
```
arkadius, line 64: Is it obvious that a "stream" component is such that is allowed for both streaming and batch? :) https://martinfowler.com/bliki/UbiquitousLanguage.html As a representative of domain experts I want to file my objection ;-) I think that the only unambiguous would sth like `BothBoundedUnboundedCompent`. But when we add it, someone can say: why there is no `BothBoundedRequestResponseCompent` and `BothUnbounedRequrestResponseCompent`. What we would say than? :)

lciolecki: Are Bounded and Unbounded now in our "Ubiquitous Language"? I don't think so ;) I agree with rest part of your opinion that someone can ask questions ;) I can agree and we can resign from this change..

PR #6269, opened 2024-06-25
Fix `TypingResult.valueOpt` for the purpose of SpEL validator

wrzontek:
## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [ ] Code is cleaned from temporary changes and commented out lines
- [ ] Parts of the code that are not easy to understand are documented in the code
- [ ] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [ ] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #6436, opened 2024-07-23
Fix TypingResult.valueOpt for the purpose of SpEL validator (fixed after rollback)

wrzontek:
## Describe your changes
Fixed version of https://github.com/TouK/nussknacker/pull/6269 after it was rolled back in https://github.com/TouK/nussknacker/pull/6408

## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -484,16 +491,22 @@ private[spel] class Typer(
...
-        .map(Typed.fromInstance)
-        .map(valid)
-        .valueOr(invalid(_, fallbackType))
-    }).getOrElse(valid(fallbackType))
+    (left.valueOpt, right.valueOpt) match {
+      case (Some(leftValue), Some(rightValue)) =>
+        Try(op(leftValue.asInstanceOf[A], rightValue.asInstanceOf[B])) match {
+          case Failure(a: ArithmeticException) =>
+            logger.debug(s"Arithmetic exception during op, using fallback. $a")
+            valid(fallbackType)
```
arkadius, line 499: Why this is valid?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
```diff
@@ -1,47 +1,101 @@
...
+  private val longClass       = Typed.typedClass[Long]
+  private val floatClass      = Typed.typedClass[Float]
+  private val doubleClass     = Typed.typedClass[Double]
+  private val booleanClass    = Typed.typedClass[Boolean]
+  private val stringClass     = Typed.typedClass[String]
+  private val byteClass       = Typed.typedClass[Byte]
+  private val bigIntegerClass = Typed.typedClass[BigInteger]
+  private val bigDecimalClass = Typed.typedClass[java.math.BigDecimal]
+
+  def encodeValue(value: Any): ValidatedNel[String, Json] = value match {
```
arkadius, line 25: Why it doesn't take the type anymore? It was redundant?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
```diff
@@ -1,47 +1,101 @@
...
+              case None                   => s"Failed to encode Record key '$encodedKey' as String".invalidNel
+            }
+          )
+        }
+        .toList
+        .sequence
+
+      encodedMap.map(values => Json.fromFields(values))
+    case value =>
+      s"No encoding logic for value $value of class ${value.getClass}".invalidNel
```
arkadius, line 58: I would write `Encoding of value [$value] of class [${value.getClass.getName}] is not supported`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
```diff
@@ -1,47 +1,101 @@
...
+    case value: Float =>
+      fromFloat(value).map(_.validNel).getOrElse(s"Could not encode $value as json.".invalidNel)
+    case value: Double =>
+      fromDouble(value).map(_.validNel).getOrElse(s"Could not encode $value as json.".invalidNel)
+
+    case vals: java.util.Collection[_] =>
+      val encodedValues = vals.asScala.map(elem => encodeValue(elem)).toList.sequence
+      encodedValues.map(values => Json.fromValues(values))
+    case vals: java.util.Map[_, _] =>
+      val encodedMap: Validated[NonEmptyList[String], List[(String, Json)]] = vals.asScala
```
arkadius, line 44: 1. There is the `ValidatedNel` type alias for that. Also, maybe let's skip the type?
2. Let's convert to list and then use sequence. Mapping of Map introduces some not necessary costs of keys recomputation.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
```diff
@@ -129,8 +137,13 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
 
   private def typedObjectWithValue(obj: HCursor): Decoder.Result[TypingResult] = for {
     valueClass <- typedClass(obj)
-    value      <- SimpleObjectEncoder.decode(valueClass, obj.downField("value"))
-  } yield TypedObjectWithValue(valueClass, value)
+    value = SimpleObjectEncoder.decodeValue(valueClass, obj.downField("value"))
+  } yield value match {
+    case Left(e) =>
+      logger.warn(s"Failed value decoding: $e")
+      valueClass
```
arkadius, line 144: We we need this fallback? Generally, it is usually better to fail fast. Thanks to that it is easier to detect the root problem and fix the bug.

docs/MigrationGuide.md
```diff
@@ -16,6 +16,11 @@ To see the biggest differences please consult the [changelog](Changelog.md).
   * Now method `returnType` from `EagerServiceWithStaticParameters` requires implicit nodeId param
 * [#6437](https://github.com/TouK/nussknacker/pull/6437) Removed deprecated operation to create a scenario:
   POST `/api/processes/{name}/{category}`. POST `/api/processes` should be used instead.
+* [#TODO](https://github.com/TouK/nussknacker/pull/TODO) Changes to `TypingResult` of SpEL expressions that are maps or lists:
+    * `TypedObjectTypingResult.valueOpt` now returns a `java.util.map` instead of `scala.collection.immutable.Map`
```
arkadius, line 20: `java.util.Map`

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
```diff
@@ -453,7 +453,7 @@ class LiteKafkaUniversalAvroSchemaFunctionalTest
         ),
         (
           rConfig(sampleInteger, recordIntegerSchema, recordArrayOfNumbersSchema, List(sampleString)),
-          invalidTypes(s"path 'field[]' actual: '${typedStr.display}' expected: 'Integer | Double'")
+          invalidTypes(s"path 'field[]' actual: '${typedStr.withoutValue.display}' expected: 'Integer | Double'")
```
arkadius, line 456: Generally, we shouldn't do any processing in the verification phase of tests. I mean we should avoid expected value transformation etc.

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalJsonFunctionalTest.scala
```diff
@@ -286,10 +286,12 @@ class LiteKafkaUniversalJsonFunctionalTest
...
-    results.isValid shouldBe true // it should be invalid, but it's so edge case that we decided to live with it
-    val runtimeError = results.validValue.errors.head
-    runtimeError.nodeId shouldBe Some("my-sink")
-    runtimeError.throwable.asInstanceOf[RuntimeException].getMessage shouldBe "#: [1,2] is not a valid enum value"
+    results.isValid shouldBe false
+    val runtimeError = results.invalidValue.head
+    runtimeError.nodeIds shouldBe Set("my-sink")
+    runtimeError.asInstanceOf[CustomNodeError].message shouldBe
+      """Provided value does not match scenario output - errors:
+        |Incorrect type: actual: 'List[Integer]([1, 2])' expected: 'List[Integer]([1, 2, 3]) | String(A)'.""".stripMargin
```
arkadius, line 294: I think that we shold rather use spel notation to display collection types - `List[Integer]({1, 2, 3})`

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionGenSpec.scala
```diff
@@ -96,7 +98,14 @@ class SpelExpressionGenSpec
           inside(validate(expr, a, b)) { case Valid(typedExpression) =>
             typedExpression.typingInfo.typingResult match {
               case TypedObjectWithValue(TypedClass(typedClass, Nil), typedValue) =>
-                typedValue shouldEqual evaluatedValue
+                (typedValue, evaluatedValue) match {
+                  case (v1: java.math.BigDecimal, v2: java.math.BigDecimal) =>
+                    // SpEL evaluation uses different precision and rounding than MathUtils
```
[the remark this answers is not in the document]
wrzontek: it wasn't detected, because the value for BigDecimal was lost in `Typed.fromInstance`, and the branch with comparing values wasn't entered

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
```diff
@@ -1,47 +1,101 @@
...
+    case `doubleClass`                  => obj.as[Double]
+    case `booleanClass`                 => obj.as[Boolean]
+    case `stringClass`                  => obj.as[String]
+    case `byteClass`                    => obj.as[Byte]
+    case `bigIntegerClass`              => obj.as[BigInteger]
+    case `bigDecimalClass`              => obj.as[java.math.BigDecimal]
+    case TypedClass(klass, List(elementType: TypingResult)) if klass == classOf[java.util.List[_]] =>
+      obj.values
+        .getOrElse {
+          throw new RuntimeException(s"$typ  -  ${obj.toString}")
```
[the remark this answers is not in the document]
wrzontek: it's debug leftover code, sorry

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
wrzontek: BigDecimal division frequently throws "Non-terminating decimal expansion", I think using the fallback (typedclass BigDecimal without value) is ok, but perhaps there's a way to round the result or something, i'll check

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionGenSpec.scala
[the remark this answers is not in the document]
wrzontek: I'm not sure why exactly, but the method 

```
  private def evaluate(expr: String, a: Any, b: Any): AnyRef = {
    val spelParser        = new org.springframework.expression.spel.standard.SpelExpressionParser()
    val evaluationContext = new StandardEvaluationContext()
    evaluationContext.setVariable("a", a)
    evaluationContext.setVariable("b", b)
    spelParser.parseRaw(expr).getValue(evaluationContext)
  }
```

returns sometimes slightly different values than the flow through `validate->parse->typeNode...->MathUtils`,

both methods use internals not really controlled by us (spel vs java.math.bigdecimal), so I think just setting a tolerance like `1` or`0.1` is enough from a test perspective

 the biggest difference i've seen after running the test a few times was:
`-0.07829300196081781493782913994294858639477752149105072021484375` (typed)
`0` (evaluated)

usually the difference is at like the 50th decimal place, so not really major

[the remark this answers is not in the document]
wrzontek: wait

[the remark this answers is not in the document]
wrzontek: yeah I can just set the scale and rounding in MathUtils to the same as is used in SpEL internals

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
wrzontek: dividing with RoundingMode.HALF_EVEN (consistent with SpEL evaluation) avoids this exception

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeEncoders.scala
wrzontek: I thought that maybe if we encounter a value that we can't encode/decode, but we can encode/decode the class, then returning the class alone (and logging the failure) is better/safer than throwing

If you have the opposite opinion, I'll change it

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
wrzontek: `AvroSchemaOutputValidator` only checks the element type of a list (as it did previously, as value wasn't even kept)

do we want to extend the validation to also check each element value? should it be part of this PR?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
wrzontek: yep, and it made it difficult to call recursively

wrzontek: though maybe it wasn't 100% reduntant, now the case with mismatched type-value isn't handled the same :thinking:

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
arkadius: I'm just saying that we should change the expected result (this string) to not use any transformations on types, I don't understand why some change in `AvroSchemaOutputValidator` would be necessary to achieve that.
Also, I'm not getting why this expected value changed. I assume that `typeStr` is typed as `TypedObjectWithValue`, also I assume, that kafka sink returns in the error message actual type as it is typed, so it should be also `TypedObjectWithValue`. Doesn't it? Which part of my assumptions is wrong? What was changed?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/SimpleObjectEncoder.scala
```diff
@@ -1,47 +1,101 @@
...
+        }
+        .toList
+        .traverse(v => decodeValue(elementType, v.hcursor))
+        .map(_.asJava)
+    case record: TypedObjectTypingResult =>
+      for {
+        fields <- obj.as[Map[String, Json]]
+        decodedFields <- fields.toList.traverse { case (fieldName, fieldJson) =>
```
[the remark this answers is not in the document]
arkadius: What about the test?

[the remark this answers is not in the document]
wrzontek: damn it i commited locally but forgot to push, i've added a test alongside a small refactor

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
wrzontek: it worked before essentially by accident: Lists with a single element kept it in `params` as `TypedObjectWithValue` (the common supertype of a single TypedObjectWithValue `t` is `t` itself, when the list had 2 or more values it was TypedClass), which is why `AvroSchemaOutputValidator` in `private def invalid` had the `typedStr` as `TypedObjectWithValue` and the resulting error message included the value

now the `params` of a TypedList are always `TypedClass` (without value), because the value is kept separately anyway (and because the behaviour was weird and inconsistent anyway)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -116,12 +116,17 @@ object typing {
...
-      val shortenedDataString =
-        if (dataString.length <= maxDataDisplaySize) dataString
-        else dataString.take(maxDataDisplaySizeWithDots) ++ "..."
-      s"${underlying.display}($shortenedDataString)"
+    override def display: String = s"${underlying.display}($shortenedDataString)"
+
+    private def shortenedDataString = {
+      val dataString = value match {
+        case l: java.util.List[_]       => l.asScala.mkString("{", ", ", "}")
+        case l: scala.collection.Seq[_] => l.mkString("{", ", ", "}")
```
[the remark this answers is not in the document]
wrzontek: i don't think so, so this case is unncessary, removing

engine/lite/components/kafka-tests/src/test/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaUniversalAvroSchemaFunctionalTest.scala
arkadius: Ah, now I get it.

arkadius (approved):

PR #6498, opened 2024-07-29
Add svg icon to fragment properties

bartektartanus:
image


arkadius (approved):
LGTM, but let's @Dzuming take a look on FE side too

Dzuming (approved):
LGTM

PR #6506, opened 2024-07-29
[NU-1556] Real data test data generation for Table Source

mslabek:
## Checklist before merge


engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
```diff
@@ -113,37 +129,25 @@ object FlinkMiniClusterTableOperations extends LazyLogging {
...
-      logger.debug(s"Successfully deleted temporary test data dumping directory at: '${dir.toUri.toURL}'")
-    } match {
-      case Failure(e) =>
-        logger.error(
-          s"Couldn't properly delete temporary test data dumping directory at: '${dir.toUri.toURL}'",
-          e
-        )
-      case Success(_) => ()
-    }
-    def deleteTable(tableName: String)(implicit env: TableEnvironment): Unit = {
```
mslabek, line 134: until we dont have a single persistent minicluster per designer, we don't need to delete tables

e2e-tests/src/test/resources/spec-setup/batch-config/batch-customizations.conf
```diff
@@ -1,21 +1,27 @@
 scenarioTypes {
-  "batch": {
-    deploymentConfig: {
-      type: "flinkStreaming"
-      restUrl: ${?FLINK_REST_URL}
+  "batch": ${baseBatchConfig}
+  "batch-with-test-on-real-data": ${baseBatchConfig} {
+    modelConfig.components.flinkTable.useRealDataForTests: true
+    category: "BatchTestOnRealData"
```
mslabek, line 5: To have a Designer for e2e tests with testing on real data and on generated data, I had to make another scenario type with another category because of scenario type limitations.

docs/Changelog.md
```diff
@@ -28,9 +28,14 @@
 * [#6415](https://github.com/TouK/nussknacker/pull/6415) Added "Component group" field to fragment properties, which allows selection of the group of components in the Creator Panel in which the fragment will be available
 * [#6462](https://github.com/TouK/nussknacker/pull/6462) Improvement of Component's API: `canHaveManyInputs` property is now 
   determined automatically, developer doesn't need to provide it by his/her own
-* [#6195](https://github.com/TouK/nussknacker/pull/6195) Added randomized test data generation for Table Source scenarios
-* [#6340](https://github.com/TouK/nussknacker/pull/6340) Added running tests on generated data for Table Source scenarios
 * [#6445](https://github.com/TouK/nussknacker/pull/6445) [#6499](https://github.com/TouK/nussknacker/pull/6499) Add support to seconds in a duration editor
+* [#6195](https://github.com/TouK/nussknacker/pull/6195) [#6340](https://github.com/TouK/nussknacker/pull/6340) [#6195](https://github.com/TouK/nussknacker/pull/6195) Added test data generation and testing for Table Source:
+  * Added test data generation with 2 modes:
+    * Random (default) - generates randomized records
```
arkadius, line 34: Let's make a Real mode default as we agree that on this stage we want to align features of table source with kafka source

docs/Changelog.md
```diff
@@ -28,9 +28,14 @@
...
+    * Real - generates records dumped from the actual configured data source 
+  * Data generation mode can be configured through the `useRealDataForTests` flag in the table components' configuration
```
arkadius, line 36: Let's rename it to `dataGenerationMode`  or sth similar with two options: random | live

e2e-tests/src/test/resources/spec-setup/batch-config/batch-customizations.conf
arkadius: It's fine but let's use live mode in the default batch scenario type and random in the new one (see the comment above)

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
```diff
@@ -179,3 +183,10 @@ object FlinkMiniClusterTableOperations extends LazyLogging {
   }
 
 }
+
+sealed trait DataGenerationMode
+
+object DataGenerationMode {
+  case object RandomDataMode                                                    extends DataGenerationMode
+  case class RealDataMode(sqlStatements: List[SqlStatement], tableName: String) extends DataGenerationMode
```
arkadius, line 191: Modes are rather things that are parameterless enums. Let's reserve this word for this purpose and use something else here. Also, I see that you use it very locally now, you could just have two methods in `FlinkMiniClusterTableOperations`

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -87,11 +87,26 @@ class TableSource(
...
+    FlinkMiniClusterTableOperations.generateTestData(size, tableDefinition.toFlinkSchema, mode)
+  }
+
 }
 
 object TableSource {
   type RECORD = java.util.Map[String, Any]
   private val filteringInternalViewName = "filteringView"
+
+  private[source] def executeSqlAndGetTable(
```
arkadius, line 103: I would split it into two methods. Also, I see that "execute SQL" / "SQL statements" are too much generic names and it can cause confusions. I think that we need at least to add `DDL` word to make it clear.

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
```diff
@@ -88,22 +90,36 @@ object FlinkMiniClusterTableOperations extends LazyLogging {
...
+  }
+
+  private def createRealDataGeneratorTable(
+      sqlStatements: List[SqlStatement],
+      tableName: String,
+  )(
+      implicit env: TableEnvironment
+  ): String = {
+    TableSource.executeSqlAndGetTable(sqlStatements, tableName, env)
+    tableName
```
arkadius, line 115: Why do you "get table" and then drop the result and use table name instead?

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
```diff
@@ -22,36 +24,36 @@ import scala.util.{Failure, Success, Try, Using}
...
       TestData(rows.map(TestRecord(_)))
     }
-  }
 
   def parseTestRecords(records: List[TestRecord], schema: Schema): List[RECORD] =
     ThreadUtils.withThisAsContextClassLoader(getClass.getClassLoader) {
       implicit val env: StreamTableEnvironment = MiniClusterEnvBuilder.buildStreamTableEnv
-      val (inputTableName, _)                  = generateTestDataTableNames
-      val inputTablePath                       = createTempFileTable(schema, inputTableName)
+      val (inputTablePath, inputTableName)     = createTempFileTable(schema)
```
arkadius, line 49: This: `env.from(inputTableName)` is wrong. You should escape the `inputTableName`. It would be nice to reuse `Table` object returned in `executeSqlAndGetTable`.

mslabek (commented):
Review fixes done :+1: Probably some tests will fail and will have to make some small corrections after this

docs/Changelog.md
```diff
@@ -28,11 +28,17 @@
...
-* [#6340](https://github.com/TouK/nussknacker/pull/6340) Added running tests on generated data for Table Source scenarios
 * [#6445](https://github.com/TouK/nussknacker/pull/6445) [#6499](https://github.com/TouK/nussknacker/pull/6499) Add support to seconds in a duration editor
 * [#6436](https://github.com/TouK/nussknacker/pull/6436) Typed SpEL list expressions will now infer their compile-time known values, instead of only the supertype of its elements. These values can be used in custom components or validators.
     * NOTE: selection (`.?`), projection (`.!`) or operations from the `#COLLECTIONS` helper cause the typed list to lose its elements' values
+* [#6195](https://github.com/TouK/nussknacker/pull/6195) [#6340](https://github.com/TouK/nussknacker/pull/6340) [#6195](https://github.com/TouK/nussknacker/pull/6195) Added test data generation and testing for Table Source:
+  * Added test data generation with 2 modes:
+    * Live (default) - creates records by pulling data from the currently configured data source
+    * Random - creates randomized records
+  * Data generation mode can be configured through the `testDataGenerationMode` in the table components configuration
```
arkadius, line 38: Shouldn't there be "live" or "random" values?

docs/Changelog.md
arkadius: Yeah, it is better

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/FlinkTableComponentProvider.scala
```diff
@@ -107,5 +109,43 @@ final case class TableSqlDefinitions(
...
+    }
+  }
+
+  sealed trait TestDataGenerationMode
+
+  object TestDataGenerationMode {
+    case object Random extends TestDataGenerationMode
+    case object Live   extends TestDataGenerationMode
+
+    implicit val configReader: ValueReader[Option[TestDataGenerationMode]] = (config: Config, path: String) => {
```
arkadius, line 139: This can be done automatically if we use scala enums and ready `EnumerationReader`

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
arkadius: It looks much more intuitive now :+1:

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/FlinkTableComponentProvider.scala
mslabek: Right, done :+1: now its able to be fully automatic for the provider config too

docs/Changelog.md
mslabek: Ahhh :D Thanks, good catch, fixed

arkadius (approved):

PR #6524, opened 2024-07-31
language of parameter's expression is based on editor type

vder:
## Describe your changes

little change to make `ConfigParameterDefaultValueDeterminer` language aware.

## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
@@ -2,11 +2,9 @@ package pl.touk.nussknacker.engine.definition.component.parameter.defaults
 
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer {
+object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer with SimpleLanguageDeterminer {
 
-  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] = {
-    // TODO: make language configurable as well
-    parameters.parameterConfig.defaultValue.map(Expression.spel)
-  }
+  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] =
+    calculateDefaultValue(parameters.determinedEditor, parameters.parameterConfig.defaultValue)
 
 }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
@@ -2,9 +2,11 @@ package pl.touk.nussknacker.engine.definition.component.parameter.defaults
 
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer with SimpleLanguageDeterminer {
+object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer {
 
-  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] =
-    calculateDefaultValue(parameters.determinedEditor, parameters.parameterConfig.defaultValue)
+  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] = {
+    val language = EditorBasedLanguageDeterminer.determineLanguageOf(parameters.determinedEditor)
+    parameters.parameterConfig.defaultValue.map(Expression(language, _))
+  }
 
 }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/OptionalityBasedDefaultValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/OptionalityBasedDefaultValueDeterminer.scala
@@ -1,39 +1,14 @@
 package pl.touk.nussknacker.engine.definition.component.parameter.defaults
 
-import pl.touk.nussknacker.engine.api.definition._
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-protected object OptionalityBasedDefaultValueDeterminer extends ParameterDefaultValueDeterminer {
+protected object OptionalityBasedDefaultValueDeterminer
+    extends ParameterDefaultValueDeterminer
+    with SimpleLanguageDeterminer {
 
-  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] = {
-    Option(parameters).filter(_.isOptional).map { _ =>
-      val lang = parameters.determinedEditor
-        .map {
-          case RawParameterEditor                   => Expression.Language.Spel
-          case simpleEditor: SimpleParameterEditor  => determineLanguageOf(simpleEditor)
-          case DualParameterEditor(simpleEditor, _) => determineLanguageOf(simpleEditor)
-        }
-        .getOrElse {
-          Expression.Language.Spel
-        }
-
-      Expression(lang, "")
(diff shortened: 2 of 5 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.definition.component.parameter.defaults
+
+import pl.touk.nussknacker.engine.api.definition._
+import pl.touk.nussknacker.engine.graph.expression.Expression
+
+trait SimpleLanguageDeterminer {
+
+  def calculateDefaultValue(
```
arkadius, line 8: It is a little bit supprising that simple language determiner calculate default value :) I'd rather expect that it would do some simple language determining ;-)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/OptionalityBasedDefaultValueDeterminer.scala
```diff
@@ -1,39 +1,14 @@
 package pl.touk.nussknacker.engine.definition.component.parameter.defaults
 
-import pl.touk.nussknacker.engine.api.definition._
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-protected object OptionalityBasedDefaultValueDeterminer extends ParameterDefaultValueDeterminer {
+protected object OptionalityBasedDefaultValueDeterminer
```
arkadius, line 5: Please use composition over inheritance

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
```diff
@@ -2,11 +2,9 @@ package pl.touk.nussknacker.engine.definition.component.parameter.defaults
 
 import pl.touk.nussknacker.engine.graph.expression.Expression
 
-object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer {
+object ConfigParameterDefaultValueDeterminer extends ParameterDefaultValueDeterminer with SimpleLanguageDeterminer {
```
arkadius, line 5: Please use composition over inheritance

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
```diff
@@ -0,0 +1,36 @@
...
+      determinedEditor: Option[ParameterEditor],
+      defaultValue: Option[String]
+  ): Option[Expression] =
+    for {
+      value <- defaultValue
+      language <- determinedEditor
+        .collect {
```
mk-software-pl, line 15: IMO it's better to avoid using `collect` in favor of `map` - in the latter case, in the future, when a new editor is added, a compiler shows us this place, and a developer should consider what to do in this new case (currently, we assume that all new cases will have language SpEL)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
```diff
@@ -0,0 +1,36 @@
+package pl.touk.nussknacker.engine.definition.component.parameter.defaults
+
+import pl.touk.nussknacker.engine.api.definition._
+import pl.touk.nussknacker.engine.graph.expression.Expression
+
+trait SimpleLanguageDeterminer {
```
wrzontek, line 6: I think this would be better as an `object`, not a `trait`, so essentially "composition over inheritance", as @arkadius says

also `EditorBasedLanguageDeterminer` would be a more telling name IMO

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
```diff
@@ -0,0 +1,36 @@
...
+      determinedEditor: Option[ParameterEditor],
+      defaultValue: Option[String]
+  ): Option[Expression] =
+    for {
+      value <- defaultValue
+      language <- determinedEditor
+        .collect {
+          case RawParameterEditor                   => Expression.Language.Spel
+          case simpleEditor: SimpleParameterEditor  => determineLanguageOf(simpleEditor)
+          case DualParameterEditor(simpleEditor, _) => determineLanguageOf(simpleEditor)
```
wrzontek, line 18:
```suggestion
          case DualParameterEditor(simpleEditor, DualEditorMode.SIMPLE) => determineLanguageOf(simpleEditor)
          case DualParameterEditor(_, DualEditorMode.RAW)               => Expression.Language.Spel
```

wrzontek:
`TypeRelatedParameterValueDeterminer.defaultStringExpression` also uses very similar logic to determine the language, i think it should also use this new common trait/object

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
```diff
@@ -2,11 +2,9 @@ package pl.touk.nussknacker.engine.definition.component.parameter.defaults
...
 
-  override def determineParameterDefaultValue(parameters: DefaultValueDeterminerParameters): Option[Expression] = {
-    // TODO: make language configurable as well
```
wrzontek, line 8: this TODO wasn't realized, just deleted, are we sure we don't want to keep it?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
vder: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/OptionalityBasedDefaultValueDeterminer.scala
vder: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
vder: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
vder: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
wrzontek: nevermind, now that i think about it, changing the mode doesn't change the language, so this suggestion should be discarded

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/SimpleLanguageDeterminer.scala
vder: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/ConfigParameterDefaultValueDeterminer.scala
vder: Mixed feelieng here, as far as I understand the logic, Editor is enough to determine the language of the expression. I would like to avoid situation where we have to define the same thing in several places as it always brings some synchro issues.

wrzontek (approved):

arkadius (approved):

PR #6545, opened 2024-08-02
TableDefinition refactoring: holding full schema information, derived…

arkadius:
… types are determined lazy based on this

## Checklist before merge


mslabek (approved):

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/FlinkMiniClusterTableOperations.scala
```diff
@@ -25,6 +26,7 @@ object FlinkMiniClusterTableOperations extends LazyLogging {
     ThreadUtils.withThisAsContextClassLoader(getClass.getClassLoader) {
       implicit val env: StreamTableEnvironment = MiniClusterEnvBuilder.buildStreamTableEnv
       val (inputTablePath, inputTableName)     = createTempFileTable(schema)
+      logger.info(s"Parser schem: $schema")
```
mslabek, line 29: leftover log

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/sink/TableSink.scala
```diff
@@ -61,7 +60,9 @@ class TableSink(
 
     val inputValueTable = tableEnv
       .fromDataStream(streamOfRows)
-      .select(tableDefinition.sinkRowDataType.getLogicalType.toRowTypeUnsafe.getFieldNames.asScala.toList.map($): _*)
+      .select(
+        tableDefinition.schema.toSinkRowDataType.getLogicalType.toRowTypeUnsafe.getFieldNames.asScala.toList.map($): _*
```
mslabek, line 64: 1. Doesn't it have to be sorted like in table source?
2. Thats a pretty long expression and its used in similar forms in other places. Maybe its worth to extract some of it to the `TableDefinition` to increase readability and deduplicate some code?

mslabek (approved):

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/source/TableSource.scala
```diff
@@ -89,26 +93,54 @@ class TableSource(
...
         )
     }
   }
 
+  // This is copy-paste of toSourceRowDataType with filtering-out of computed columns.
+  // We dont' want to generate data for computed columns - they will be added during parsing of test data
+  private def fieldsWithoutComputedColumns =
+    tableDefinition.schema.getColumns.asScala
+      .filterNot(c => c.isInstanceOf[ComputedColumn])
+      .map(c => DataTypes.FIELD(c.getName, DataTypeUtils.removeTimeAttribute(c.getDataType)))
```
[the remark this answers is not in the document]
arkadius: I want to be as compatible with `toSourceRowDataType` as it is possible - In the comment above there is a notice that it is a copy-pasted method, I can add also a comment next to this line. TBH I don't know why they remove these attributes.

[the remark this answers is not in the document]
arkadius: I extracted an extension class for `ResolvedSchema` to make it clear that it is a copy-paste. I also copied their explanation why they remove this attribute

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/sink/TableSink.scala
arkadius: 1. No, it doesn't. We should insert columns in the same order as in sink. This sorting is only needed when a Row will be serialized using TestInfo prepared by our detection
2. I extracted fields in `TableDefinition` and added some extensions to make these lines shorter

PR #6578, opened 2024-08-09
Added docsUrl to PropertiesConfig

Diamekod0221:
## Describe your changes
We want to make it possible to display a docsUrl with icon in scenario properties, based on processing type and configurable by model.conf
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
```diff
@@ -69,26 +69,26 @@ object ParameterConfig {
   val empty: ParameterConfig = ParameterConfig(None, None, None, None, None)
 }
 
-@JsonCodec case class ScenarioPropertyConfig(
+@JsonCodec case class SingleScenarioPropertyConfig(
```
arkadius, line 72: I'm not a fan of naming thing with the `Single` prefix. It breaks the encapsulation / interface segregation principle. The `scenario property config` name already saying that there is only one property. You added `Single` to resolve problem in the place where this class is used so you have cycle of dependencies: name of class depends on the place of usage which depends on the class.   

Also, please move this class to `deployment-manager-api`. This class isn't used by components, it is used by deployment managers which can defined properties config programmatically in `DeploymentManagerProvider.scenarioPropertiesConfig`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/properties/ScenarioProperties.scala
```diff
@@ -0,0 +1,25 @@
+package pl.touk.nussknacker.engine.api.properties
+
+import io.circe.generic.JsonCodec
+import pl.touk.nussknacker.engine.api.component.SingleScenarioPropertyConfig
+
+@JsonCodec case class ScenarioProperties(
```
arkadius, line 6: We don't need to keep it in the `components-api` module, let's move it to designer module

components-api/src/main/scala/pl/touk/nussknacker/engine/api/properties/ScenarioProperties.scala
```diff
@@ -0,0 +1,25 @@
...
+    propertiesConfig: Map[String, SingleScenarioPropertyConfig],
+    docsUrl: Option[String] = None
+) {
+
+  // it will overwrite docsUrl, with the approach that the latest config is the proper one.
+  def ++(newConfig: ScenarioProperties): ScenarioProperties = {
```
arkadius, line 12: I doesn't see a reason to have this confusing method. In the only place that you use it, you merge only propertiesConfig map

test.txt
```diff
@@ -0,0 +1,225 @@
+[info] welcome to sbt 1.9.6 (Eclipse Adoptium Java 11.0.21)
```
arkadius, line 1: ?! ;-)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
Diamekod0221: It makes it easier to make the distinciton between a single property and properties which are pretty close ot each other in terms of naming. I don't agree that the name depends on the place - even if we removed the Single prefix, it would still be refering to a singular entity. It is also in line with the `SingleComponentConfig`.

As for moving the class, I tried moving it to `nussknacker-deployment-manager-api` but there are a few classes in `components-api` that use this class and they break after this change. Should i somehow reroute the packages so that the classes from `deployment-manager-api` can be used in components or do we leave this?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/properties/ScenarioProperties.scala
Diamekod0221: Same problem with moving as the previous comment

components-api/src/main/scala/pl/touk/nussknacker/engine/api/properties/ScenarioProperties.scala
Diamekod0221: Ok, removed and changed to mergeConfigs in ProcessingTypeData

test.txt
Diamekod0221: mistake, removed :D

components-api/src/main/scala/pl/touk/nussknacker/engine/api/properties/ScenarioProperties.scala
arkadius: I can't see a connection. `designer` module depends on every -api module it shouldn't be problematic

Diamekod0221: Classes `DeploymentManagerProvider` `DeploymentManagerStub` need this ScenarioProperties class and it's not acessible for them when moved to  `designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/scenarioproperty/ScenarioProperties.scala`. Should i move them too?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/SingleComponentConfig.scala
Diamekod0221: I have renamed this and also the SingleComponentConfig class.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentConfig.scala
```diff
@@ -6,7 +6,7 @@ import io.circe.generic.JsonCodec
 import pl.touk.nussknacker.engine.api.definition.{ParameterEditor, ParameterValidator, SimpleParameterEditor}
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 
-case class SingleComponentConfig(
+case class ComponentConfig(
```
arkadius, line 9: Please mention about this change in the migration guide

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentAdditionalConfigConverter.scala
```diff
@@ -15,8 +15,8 @@ import pl.touk.nussknacker.engine.compile.nodecompilation.ValueEditorValidator
 
 object ComponentAdditionalConfigConverter {
 
-  def toSingleComponentConfig(componentAdditionalConfig: ComponentAdditionalConfig): SingleComponentConfig =
-    SingleComponentConfig(
+  def toSingleComponentConfig(componentAdditionalConfig: ComponentAdditionalConfig): ComponentConfig =
```
arkadius, line 18: `toComponentConfig`

engine/flink/management/dev-model/src/main/resources/defaultModelConfig.conf
```diff
@@ -60,43 +60,46 @@ components {
 
 
 scenarioPropertiesConfig {
-  environment: {
-    editor: {
-      type: "StringParameterEditor"
+  propertiesConfig {
```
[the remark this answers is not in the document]
arkadius: What about this? Also some test failes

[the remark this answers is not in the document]
Diamekod0221: I have moved the docsUrl outside of this config. Now we have `propertiesConfig` and a few lines lower `docsUrl`, there's no nested config. This is parsed into the same object as before, `ScenarioProperties(propertiesConfig, docsUrl)` on BE. I will add the docs update.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentAdditionalConfigConverter.scala
Diamekod0221: updated

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentConfig.scala
Diamekod0221: i have aded a coment

engine/flink/management/dev-model/src/main/resources/defaultModelConfig.conf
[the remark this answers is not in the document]
Diamekod0221: docs update added.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentConfig.scala
Diamekod0221: mentioned

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentAdditionalConfigConverter.scala
Diamekod0221: changed

designer/client/src/components/graph/node-modal/properties.tsx
```diff
@@ -29,7 +29,10 @@ export function Properties({
     showSwitch,
     showValidation,
 }: Props): JSX.Element {
-    const scenarioPropertiesConfig = useSelector(getScenarioPropertiesConfig);
+    const scenarioProperties = useSelector(getScenarioPropertiesConfig);
+    const scenarioPropertiesConfig = scenarioProperties?.propertiesConfig ?? {};
+    console.log("scenarioPropertiesConfig:", scenarioPropertiesConfig); // Log the value for debugging
```
gskrobisz, line 34: Please, remove console.log

Diamekod0221: removed

Dzuming (approved):
Frontend LGTM

engine/flink/management/dev-model/src/main/resources/defaultModelConfig.conf
```diff
@@ -58,8 +58,7 @@ components {
   }
 }
 
-
-scenarioPropertiesConfig {
+propertiesConfig {
```
arkadius, line 61: The previous name was good, we don't need to change it. The lone "properties" may mean many things, "scenario properties" are more explicit

docs/configuration/model/ModelConfiguration.md
```diff
@@ -179,3 +179,9 @@ scenarioPropertiesConfig {
 }
 ```
 Configuration is similar to [component configuration](#configuration-of-ui-attributes-of-components)
+
+The docs button which is displayed in the modal window of a component is by default disabled for properties. One can enable it by adding
```
arkadius, line 183: "scenario properties modal window" or "modal opened after clicking properties button" - this is not a "modal window of a COMPONENT"

engine/flink/management/dev-model/src/main/resources/defaultModelConfig.conf
Diamekod0221: ok, i have changed the naming

docs/configuration/model/ModelConfiguration.md
Diamekod0221: I have changed the wording

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerProvider.scala
```diff
@@ -25,7 +26,7 @@ trait DeploymentManagerProvider extends NamedServiceProvider {
 
   def metaDataInitializer(config: Config): MetaDataInitializer
 
-  def scenarioPropertiesConfig(config: Config): Map[String, ScenarioPropertyConfig] = Map.empty
+  def scenarioPropertiesConfig(config: Config): ScenarioProperties = ScenarioProperties.empty()
```
arkadius, line 29: Let's not change this.
If we change it, we need to mention that in the migration guide, but I can't see benefits from providing this property URL programmatically in the Deployment Manager. It is overkill.

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -165,7 +165,7 @@ class AkkaHttpBasedRouteProvider(
         val validator = new UIProcessValidator(
           processingTypeData.name,
           ProcessValidator.default(processingTypeData.designerModelData.modelData),
-          processingTypeData.deploymentData.scenarioPropertiesConfig,
+          processingTypeData.deploymentData.scenarioPropertiesConfig.scenarioPropertiesConfig,
```
arkadius, line 168: 1. It looks weird that we have some snake with doubled names `.scenarioPropertiesConfig.scenarioPropertiesConfig`
2. Both validator and newProcessPreparer don't need to know about properties URL. They need properties to:
a) validator - do validations against properties validators
b) newProcessPreparer - prepare scenario with default values that are available in the scenario properties definition
Let's avoid passing things that are not needed to classes

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/ScenarioPropertiesConfigFinalizer.scala
```diff
@@ -3,16 +3,23 @@ package pl.touk.nussknacker.ui.definition
...
+import pl.touk.nussknacker.engine.api.properties.ScenarioProperties
 
 class ScenarioPropertiesConfigFinalizer(
     additionalUIConfigProvider: AdditionalUIConfigProvider,
     processingType: ProcessingType
 ) {
 
-  def finalizeScenarioProperties(
-      scenarioPropertiesConfig: Map[String, ScenarioPropertyConfig],
+  def finalizeScenarioPropertiesParameters(
```
arkadius, line 13: The previous name was better. Why we have properties parameters now and they are available only here? 
In our domain we have:
1. Parameters of nodes
2. Properties of scenarios / fragments

Let's not mix these things

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerProvider.scala
Diamekod0221: If you're referring to changing the output type, I  absolutely don't think that leaving this as Map[...] is the good thing to do. It doesn't only make it weird from a domain standpoint where we have a domain object, which would be a natural way to use here. We will also mix up different abstraction levels in other parts of code where sometimes this Map, which is in reality a part of `ScenarioProperties`, will be mixed with `ScenarioProperties` themselves. It will also meddle with future changes to scenario properties. I think having this as a more separate, typed domain entity is way better.

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
Diamekod0221: 1. That is basically why I changed scenarioPropertiesConfig to propertiesConfig in the model.conf file. Currently i am at a loss because this is kinda ridiculous that we have scenarioPropertiesConfig.scenarioPropertiesConfig but I can't really come up with a new name since it would be nice if the domain objects had the same name as in the config, to avoid confusion. Rename `DeploymentData.scenarioPropertiesConfig` to `DeploymentData.scenarioProperties`? wdyt?

2. We're only passing `
          processingTypeData.deploymentData.scenarioPropertiesConfig.scenarioPropertiesConfig` which doesn't contain the url, I dont understand the issue.

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/ScenarioPropertiesConfigFinalizer.scala
Diamekod0221: I did it to avoid mixing it with finalizing `ScenarioProperties` object. But I guess there's no good way out of this misnaming. I have reverted this.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerProvider.scala
arkadius: I'm not saying that you have to return a map. You can wrap it with some VO if you want, but please don't mix URL pointing to documentation in one place of application with list of properties that user have to fill to run it.

Also, follow the YAGNI and KISS rules. You currently don't have the requirement to make this URL dynamically changed, it is static, configured in the config, so don't expose this ability in the Deployment Manager API.

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
arkadius: I think that this solution would be much easier if we moved this URL configuration outside of `scenarioType` config.

Do we have a requirement that this URL should be different for each processing type?
If not, we can move this URL on top level and remove it from `DeploymentData` and all other places. This change would reduce to ~20 LOC

`DeploymentData` is designed to keep objects that are derivates of scenarioType > deploymentConfig

Diamekod0221: The curent requirement is to have the url configurable for every processing mode so Batch, Streaming, etc. Does this make your idea possible?

arkadius: You mean different url for batch different for streaming?

Diamekod0221: Yes, the properties are different for these processing modes so it's natural that we have different properties docs for these.

arkadius: Ok, so it have to be inside `scenarioType` config but I would put it into separate field in the `DeploymentData` - the same as we have in the config. Thanks to that, we won't need to think how to avoid `scenarioPropertiesConfig.scenarioPropertiesConfig` repetition

engine/flink/management/dev-model/src/main/resources/defaultModelConfig.conf
```diff
@@ -97,6 +97,8 @@ scenarioPropertiesConfig {
   }
 }
 
+propertiesDocsUrl: "https://nussknacker.io/documentation/docs/configuration/model/ModelConfiguration/"
```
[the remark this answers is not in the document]
gskrobisz: Moved to dev-application.conf

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/DefinitionsService.scala
```diff
@@ -56,26 +56,34 @@ class DefinitionsService(
       val finalizedScenarioPropertiesConfig = scenarioPropertiesConfigFinalizer
         .finalizeScenarioProperties(scenarioPropertiesConfig)
 
+      import net.ceedubs.ficus.Ficus._
+      val propertiesDocsUrl = modelData.modelConfig.getAs[String]("propertiesDocsUrl")
```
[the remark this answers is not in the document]
gskrobisz: Renamed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/definition/DefinitionsService.scala
```diff
@@ -56,26 +56,34 @@ class DefinitionsService(
...
       components = components.map(component => component.component.id -> createUIComponentDefinition(component)).toMap,
       classes = modelData.modelDefinitionWithClasses.classDefinitions.all.toList.map(_.clazzName),
-      scenarioPropertiesConfig = (if (forFragment) FragmentPropertiesConfig.properties ++ fragmentPropertiesConfig
-                                  else finalizedScenarioPropertiesConfig)
-        .mapValuesNow(createUIScenarioPropertyConfig),
+      scenarioProperties = {
+        val transformedProps = (if (forFragment) FragmentPropertiesConfig.properties ++ fragmentPropertiesConfig
+                                else finalizedScenarioPropertiesConfig)
+          .mapValuesNow(createUIScenarioPropertyConfig)
+        UiScenarioProperties(propertiesConfig = transformedProps, docsUrl = propertiesDocsUrl)
```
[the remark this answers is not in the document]
Diamekod0221: Changed to parsing fragmentDocs.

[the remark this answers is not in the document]
Diamekod0221: Changed to parsing fragment docs separately.

PR #6591, opened 2024-08-13
[NU-1740] processing types reload

mk-software-pl:
## Describe your changes

- The whole model can be reloaded with `POST /api/app/processingtype/reload` now - you can use this endpoint to reconfigure Nu processing types without need to restart the app
- refactoring

## Checklist before merge


wrzontek:
>The whole model can be reloaded with POST /api/app/processingtype/reload now - you can use this endpoint to reconfigure Nu processing types without need to restart the app

This essentially boils down to the change that now the application conf is also reloaded from file during `reload`, right?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/UsageStatisticsReportsConfig.scala
```diff
@@ -2,7 +2,7 @@ package pl.touk.nussknacker.ui.config
 
 final case class UsageStatisticsReportsConfig(
     enabled: Boolean,
-    errorReportsEnabled: Boolean,
+    errorReportsEnabled: Boolean = true,
```
lukasz-bigorajski, line 5: Why this is changed?

lukasz-bigorajski (approved):
I'm realy impressed!
LGTM!

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/UsageStatisticsReportsConfig.scala
mk-software-pl: Good catch. It should be changed in common-designer.conf instead

arkadius (approved):
LGTM, one question added

designer/server/src/test/resources/config/common-designer.conf
```diff
@@ -60,6 +60,7 @@ notifications {
 
 usageStatisticsReports {
   enabled: true
+  errorReportsEnabled: true
```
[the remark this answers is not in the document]
mk-software-pl: yes, there is no default value for errorReportsEnabled so it should also be configured here. The same is done in `defaultDesignerConfig.conf`

PR #6614, opened 2024-08-20
[NU-1771] Add automatic array to list conversion in Spel

lukasz-bigorajski:
## Checklist before merge


scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/ArrayToCollectionConverter.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.core.convert.converter.ConditionalGenericConverter;
+import org.springframework.lang.Nullable;
+
+import java.lang.reflect.Array;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+// todo: should I move this java package to separate repository?
```
lukasz-bigorajski, line 30: Question which need to be answered >

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -130,7 +131,14 @@ private[spel] class Typer(
       current: TypingContext
   ): NodeTypingResult = {
 
-    def toNodeResult(typ: TypingResult) = current.toResult(TypedNode(node, TypingResultWithContext(typ)))
+    // We are converting arrays to lists everywhere
```
arkadius, line 134: 1. It is not exactly everyhwere - this  `toNodeResult` isn't used in every place. For example, in `typeChildrenNodes`, `TypedNode.apply` is used instead. 
2. This conversion will be used in some certain situations:
a) When a user uses projection/selection - am I right?
b) When a target type (e.g.) of method parameter is a List and we want to put Array into it
c) When user uses method that is not available in a source type
d) ... ?

In other places this type won't be change e.g.
aa) When we just assign array to local variables e.g. `array = "1,2".split(",")`
bb) When we pass array to parameter that is of `Array` type

In these mentioned places we can't change the type because it would cause `ClassCastException` during serialization of object by Flink. I think that we have to extract method in these places and do this conversion only when we are sure that it will be done in runtime as well

lukasz-bigorajski: I agree on this. According to this I changed implementation to handle selections, projects and passing array as parameter of list type.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/MethodDefinition.scala
```diff
@@ -30,7 +31,13 @@ sealed trait MethodDefinition {
 
   protected def isValidMethodInfo(arguments: List[TypingResult], methodTypeInfo: MethodTypeInfo): Boolean = {
     val checkNoVarArgs = arguments.length >= methodTypeInfo.noVarArgs.length &&
-      arguments.zip(methodTypeInfo.noVarArgs).forall { case (x, Parameter(_, y)) => x.canBeSubclassOf(y) }
+      arguments.zip(methodTypeInfo.noVarArgs).forall {
+        // Allow pass array as List argument because of array to list auto conversion:
+        // pl.touk.nussknacker.engine.spel.internal.ArrayToListConverter
+        case (tc @ TypedClass(klass, _), Parameter(_, y)) if klass.isArray =>
+          Typed.genericTypeClass[java.util.List[_]](tc.params).canBeSubclassOf(y)
```
arkadius, line 38: I would add `tc.canBeSubclassOf(y) || ...` just in case if method takes array

arkadius (approved):
One comment added

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/MethodDefinition.scala
lukasz-bigorajski: Done

docs/Changelog.md
```diff
@@ -66,6 +66,9 @@
 * [#6656](https://github.com/TouK/nussknacker/pull/6656)
   * Remove not working MODEL_CLASS_PATH environment variable
   * Add default DB connection name
+* [#6614](https://github.com/TouK/nussknacker/pull/6614) Array in SpeL improvements:
+  * From now onm it is possible to pass an array as a parameter of type List - e.g. `T(java.lang.String).join(',', #array)`.
```
mslabek, line 70: small typo :)
```suggestion
  * From now on it is possible to pass an array as a parameter of type List - e.g. `T(java.lang.String).join(',', #array)`.
```

arkadius (approved):

PR #6616, opened 2024-08-20
ValidationExpressionParameterValidator fails if expression is not compile-time evaluable

wrzontek:
## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/validator/ValidationExpressionParameterValidator.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/validator/ValidationExpressionParameterValidator.scala
@@ -22,6 +22,17 @@ case class ValidationExpressionParameterValidator(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit] = {
     value match {
+      case None if !expression.expression.isBlank =>
+        invalid(
+          CustomParameterValidationError(
+            message =
+              s"Value of expression '${expression.expression}' was unable to be determined at deployment time and it cannot be used with expression validation.",
+            description =
+              s"Please provide a value that is evaluable at deployment time and satisfies the validation expression '${validationExpression.original}'",
+            paramName = paramName,
+            nodeId = nodeId.id
+          )
+        )
       case None       => valid(())
       case Some(null) => valid(())
       case Some(v)    => validateValue(paramName, v)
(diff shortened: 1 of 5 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/validator/ValidationExpressionParameterValidator.scala
```diff
@@ -22,6 +22,17 @@ case class ValidationExpressionParameterValidator(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit] = {
     value match {
+      case None if !expression.expression.isBlank =>
```
arkadius, line 25: Currently, our validators are composable. Thanks to that we avoid duplicated errors from different validators. To achieve that, each validator has to ignore the scope of validation that it is not responsible for.  Because of that we have this `isBlank` checking. @bartektartanus prepared validator that check the same scope as you check here - `CompileTimeEvaluableValueValidator`, I think that it was designed to be composed with your validator. But TBH I'm not sure if it used anywhere else than in tests. @bartektartanus can you help in this area? If it isn't used, we could just merge them. If it is used, we should think a while how to connect them. Because this validator has no sense without validation done in `CompileTimeEvaluableValueValidator`

wrzontek: i've nested `CompileTimeEvaluableValueValidator` inside `ValidationExpressionParameterValidator`, my logic being that `CompileTimeEvaluableValueValidator` makes sense standing on it's own, while `ValidationExpressionParameterValidator` always requires `CompileTimeEvaluableValueValidator`

wrzontek: :thinking: I guess this can still duplicate errors if both are present

arkadius: You have some test failures

wrzontek: yep i renamed the error and didn't rename the expected error name, classic

arkadius (approved):
LGTM

PR #6546, opened 2024-08-02
Table types alignment

arkadius:
## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [ ] Code is cleaned from temporary changes and commented out lines
- [ ] Parts of the code that are not easy to understand are documented in the code
- [ ] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [ ] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #6640, opened 2024-08-22
BestEffort*Encoder naming changes:

arkadius:
- All `BestEffort*Encoder` classes renamed to fit `To<TargetFormat>(SchemaBased)Encoder` naming schema
- `JsonToNuStruct` renamed to `FromJsonDecoder` (to fit `From<SourceFormat>Decoder` naming schema)
- `ToJsonEncoder` renamed to `ToJsonEncoderCustomisation`
- `ToJsonBasedOnSchemaEncoder` renamed to `ToJsonSchemaBasedEncoderCustomisation`

## Checklist before merge

### DIFF
--- a/common-api/src/main/scala/pl/touk/nussknacker/engine/util/json/ToJsonEncoderCustomisation.scala
+++ b/common-api/src/main/scala/pl/touk/nussknacker/engine/util/json/ToJsonEncoderCustomisation.scala
@@ -2,6 +2,6 @@ package pl.touk.nussknacker.engine.util.json
 
 import io.circe.Json
 
-trait ToJsonEncoder {
+trait ToJsonEncoderCustomisation {
   def encoder(delegateEncode: Any => Json): PartialFunction[Any, Json]
 }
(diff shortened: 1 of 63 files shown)


lciolecki:
I don't like this prefix To*. Why just don't stay without this?

arkadius:
> I don't like this prefix To*. Why just don't stay without this?

The original reason why I did this refactor was this comment: https://github.com/TouK/nussknacker/pull/6546#discussion_r1727026037

@raphaelsolarski pointed out that `...SchemaEncoder` was confusing because we don't encode schema.
In the proposed by you convention, we will have the same problem - e.g. `AvroEncoder` doesn't encode avro objects but instead, it encodes our internal objects. I proposed `NuObjectToAvroEncoder` but we both agree with @raphaelsolarski that it is too much verbose and encoding already means that one side of this transformation is our objects.

lciolecki:
Ok, now I see the problem. You might be right and this convention makes more sens..

arkadius:
> Ok, now I see the problem. You might be right and this convention makes more sens..

Some approve would be nice to see then :)

lciolecki (approved):

PR #6643, opened 2024-08-22
TypeInformationDetection-static

arkadius:
## Checklist before merge


engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
```diff
@@ -27,3 +32,37 @@ trait TypeInformationDetection extends Serializable {
...
+    ServiceLoader
+      .load(classOf[TypeInformationDetection], classloader)
+      .asScala
+      .toList match {
+      case only :: Nil => only
+      case Nil =>
+        throw new IllegalStateException(
+          s"Missing ${classOf[TypeInformationDetection].getSimpleName} implementation on the classpath. " +
+            s"Classloader: ${printClassloaderDebugDetails(classloader)}. " +
+            s"Ensure that your classpath is correctly configured, probable flinkExecutor.jar is missing"
```
raphaelsolarski, line 51: `flinkExecutor.jar is probably missing`

engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
```diff
@@ -27,3 +32,37 @@ trait TypeInformationDetection extends Serializable {
   def forType[T](typingResult: TypingResult): TypeInformation[T]
 
 }
+
+object TypeInformationDetection {
+
+  // We could just inline implementation of TypeInformationDetection but we don't want expose too much implementation
```
raphaelsolarski, line 38: I don't understand the reason why we use SPI here - imo we should allow it only if we expect that it should be "overridable" in specific deployments.
What you mean by exposing too much implementation?

raphaelsolarski (changes requested):

engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/typeinformation/TypeInformationDetection.scala
raphaelsolarski: + and we don't want **to** expose too much implementation

arkadius: I was looking for a clean solution that would split single implementation from the interface. I don't want the developer of the component to see too many classes. Do you have some other nice solution that would fulfil this requirement?

raphaelsolarski (approved):

PR #6650, opened 2024-08-25
Work around Kafka source resetting to earliest offsets on some state restores

piotrp:
## Describe your changes

Added a workaround for situations when an updated scenario with Kafka source started reading from the earliest offsets.

Explanation is in code comments.

## Checklist before merge

### DIFF
--- a/engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
+++ b/engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
@@ -2,12 +2,13 @@ package pl.touk.nussknacker.engine.kafka.source.flink
 
 import cats.data.NonEmptyList
 import com.github.ghik.silencer.silent
+import com.typesafe.scalalogging.LazyLogging
 import org.apache.flink.api.common.functions.RuntimeContext
 import org.apache.flink.configuration.Configuration
 import org.apache.flink.streaming.api.datastream.DataStreamSource
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment
 import org.apache.flink.streaming.api.functions.source.SourceFunction
-import org.apache.flink.streaming.connectors.kafka.FlinkKafkaConsumer
+import org.apache.flink.streaming.connectors.kafka.{FlinkKafkaConsumer, FlinkKafkaConsumerBase}
 import org.apache.kafka.clients.consumer.ConsumerRecord
 import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.definition.Parameter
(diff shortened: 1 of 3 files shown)


engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
```diff
@@ -171,6 +175,45 @@ class FlinkKafkaConsumerHandlingExceptions[T](
...
+      s"$this must be an instance of ${classOf[FlinkKafkaConsumerBase[_]]}"
+    )
+    val restoredStateField = classOf[FlinkKafkaConsumerBase[_]].getDeclaredField("restoredState")
+    restoredStateField.setAccessible(true)
+    restoredStateField.get(this) match {
+      case null => // there is no restored stare
+      case tm: util.TreeMap[_, _] =>
+        if (tm.isEmpty) {
+          logger.info("Got empty restoredState, patching it to prevent automatic reset to the earliest offsets")
+          restoredStateField.set(this, null)
```
arkadius, line 208: How does setting this field to null prevent flink from automatic reset the consumer?

piotrp: When this field is not `null` then Flink treats all partitions that are not in state as new ones (hardcoded behavior) - it enters this branch in `open`: https://github.com/apache/flink-connector-kafka/blob/v3.1.0/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaConsumerBase.java#L578

arkadius: And when this field is set to `null` it uses specified by us resetting strategy which can be latest - here is the point of this change? I know that it might be obvious but IMO it is worth writing this.

piotrp: Added

arkadius (approved):

engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
```diff
@@ -171,6 +175,47 @@ class FlinkKafkaConsumerHandlingExceptions[T](
...
+      this.isInstanceOf[FlinkKafkaConsumerBase[_]],
+      s"$this must be an instance of ${classOf[FlinkKafkaConsumerBase[_]]}"
+    )
+    val restoredStateField = classOf[FlinkKafkaConsumerBase[_]].getDeclaredField("restoredState")
+    restoredStateField.setAccessible(true)
+    restoredStateField.get(this) match {
+      case null => // there is no restored stare
+      case tm: util.TreeMap[_, _] =>
+        if (tm.isEmpty) {
+          logger.info("Got empty restoredState, patching it to prevent automatic reset to the earliest offsets")
```
raphaelsolarski, line 207: imo we need at least WARN here

piotrp: Changed to WARN

raphaelsolarski (approved):

PR #6700, opened 2024-08-30
[Nu-1780] Add advanced search options

DeamonDev:
## Describe your changes

I've expanded the scenario nodes search graph ability in Nussknacker designer. Now we are only able to search for simple phrase and this phrase will be searched on all the properties on all the nodes. For more demanding users I propose to expand this on the following semantics: 

`id:("id1","id2",...) description:("desc1", "desc2", ...) type:("t1", "t2", ...) paramName:("p1", "p2", ...) paramValue:("pv1", "pv2", ...) outputValue:("o1", "o2", ...) "edgeExpression:("e1", "e2", ...)` 

If any of these selectors occurs as the search input text, the expression is treated as `AdvancedSearch` and the filtering logic is specified as follows: 

The search result `n` (a node) is present under advanced search request above if and only if all of the following conditions are met: 

- the `id` of the node `n` is one of: `id1, id2, ...`
- the `description` of the node `n` is one of: `desc1, desc2, ...`
- ...
- the `edgeExpression` of the node `n` is one of `e1, e2, ...`. 

So from the logic POV it is conjunction of alternatives. 

**Remark**: If some selector is not specified then it is not taken under consideration. 
**Remark II**: In the form user may write values in the quotation marks or not. For example all of the `"n1","n2"`, `n1,n2` or even `"n1",n2` will lead to the same query result. 

From the UI perspective I propose to add such form in order to generate this search query: 

image

## Checklist before merge


Dzuming:
@DeamonDev It behaves strangely on page refresh, the panel's initial state is open and then, it's closed. Could we improve it?

Uploading Screen Recording 2024-09-09 at 10.40.43.mov…

designer/client/src/components/toolbars/search/AdvancedSearchOptions.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export function AdvancedSearchOptions({
+    setFilter,
+    setCollapsedHandler,
+}: {
+    setFilter: React.Dispatch<React.SetStateAction<string>>;
+    setCollapsedHandler: React.Dispatch<React.SetStateAction<boolean>>;
+}) {
+    const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+        event.preventDefault();
```
Dzuming, line 16: We can make this logic more generic :)

```ts
        event.preventDefault();

        const formData = new FormData(event.currentTarget);

        const transformedInputs = Array.from(formData.entries()).map(([fieldName, fieldValue]) => {
            const input = (fieldValue as string) || "";
            return transformInput(input, fieldName);
        });

        const finalText = transformedInputs.join(" ").trimEnd();

        setFilter(finalText);
```

designer/client/src/components/toolbars/search/AdvancedSearchOptions.tsx
```diff
@@ -0,0 +1,86 @@
...
+                flexDirection: "column",
+                maxWidth: "300px",
+                justifyContent: "center",
+                alignItems: "center",
+            }}
+        >
+            <Typography fontWeight="bold" sx={{ m: 1 }}>
+                Advanced Search
+            </Typography>
+            <TextField sx={{ m: 1 }} size="small" label="id" name="id" />
```
Dzuming, line 69: For now, we don't use form inputs from Material, we still use our custom inputs. Can you use our custom TextFields for consistency sake?

designer/client/src/components/toolbars/search/AdvancedSearchOptions.tsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import { TextField, Button, Box, Typography } from "@mui/material";
+
+const transformInput = (input: string, fieldName: string) => {
+    return input === "" ? "" : `${fieldName}:(${input})`;
+};
+
+export function AdvancedSearchOptions({
```
Dzuming, line 8: There is no option to clear advanced filters. Maybe it's worth adding a button to handle it?

designer/client/src/components/toolbars/search/AdvancedSearchOptions.tsx
```diff
@@ -0,0 +1,86 @@
...
+                maxWidth: "300px",
+                justifyContent: "center",
+                alignItems: "center",
+            }}
+        >
+            <Typography fontWeight="bold" sx={{ m: 1 }}>
+                Advanced Search
+            </Typography>
+            <TextField sx={{ m: 1 }} size="small" label="id" name="id" />
+            <TextField size="small" sx={{ m: 1 }} label="paramValue" name="paramValue" />
```
Dzuming, line 70: All labels should match these values https://github.com/TouK/nussknacker/blob/d984547bc87156e0f0beda94e69a53f2a668a3ec/designer/client/src/components/toolbars/search/utils.ts#L99

designer/client/src/components/toolbars/search/utils.ts
```diff
@@ -92,18 +113,100 @@ export function useFilteredNodes(filterValues: string[]): {
...
+        "description:(",
+        "type:(",
+        "paramName:(",
+        "paramValue:(",
+        "outputValue:(",
+        "edgeExpression:(",
+    ]; //heuristic, but sufficient
+    const isAdvancedSearchOption = advancedFilterKeyNames.some((advancedKeyName) => filterRawText.includes(advancedKeyName));
+
+    if (isAdvancedSearchOption) {
```
Dzuming, line 183: It should work a little differently, basic search shouldn't be cleared but advanced search should be included and narrow the final result the same way as Gmail 

https://github.com/user-attachments/assets/f1e8f275-0e45-4d5c-9204-d2b0a2387b47

designer/client/src/components/toolbars/search/utils.ts
```diff
@@ -65,11 +66,31 @@ export const findFields = (filterValues: string[], node: NodeType) => {
...
+            .filter((selector) => selector.name == selectorName)
+            .flatMap(({ name, selector }) =>
+                ensureArray(selector(node))
+                    .filter((v) => matchFilters(v, filterValues))
+                    .map(() => name),
+            ),
+    );
+};
+
+function arraysEqual(arr1: string[], arr2: string[]): boolean {
```
Dzuming, line 81: can't we just use https://docs-lodash.com/v4/is-equal/ instead?

Dzuming (changes requested):

designer/client/src/components/toolbars/search/utils.ts
```diff
@@ -92,18 +113,100 @@ export function useFilteredNodes(filterValues: string[]): {
...
+}
+
+function parseRawTextToAdvancedSearch(text: string): AdvancedSearch {
+    const result: AdvancedSearch = { searchType: SearchType.ADVANCED };
+    const regex = /(\w+):\(([^)]*)\)/g;
+    let match: RegExpExecArray | null;
+
+    while ((match = regex.exec(text)) !== null) {
+        const key = match[1] as keyof Exclude<keyof AdvancedSearch, "searchType">;
+        console.log(match[2]);
```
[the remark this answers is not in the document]
DeamonDev: Of course not :D

designer/client/src/components/sidePanels/SearchLabeledInput.tsx
```diff
@@ -0,0 +1,19 @@
+import { forwardRef, PropsWithChildren } from "react";
+import { FormControl } from "@mui/material";
+import { nodeInput } from "../graph/node-modal/NodeDetailsContent/NodeTableStyled";
+import React from "react";
+
+export type SearchLabeledInputProps = PropsWithChildren<{
+    name: string;
+}>;
+
+export const SearchLabeledInput = forwardRef<HTMLInputElement, SearchLabeledInputProps>(({ children, ...props }, ref) => {
```
Dzuming, line 10: Can we adjust the style, increase the elements' width, and add padding 16px? You can use `theme.spacing`, like [here](https://github.com/TouK/nussknacker/blob/90a4c5020a7df2568ce9ebb5aa96cb619eaf2c77/designer/client/src/components/tips/Styled.tsx#L32) 
image

designer/client/src/components/table/SearchFilter.tsx
```diff
@@ -9,13 +10,41 @@ const flex = css({
...
+    collapseHandler: React.Dispatch<React.SetStateAction<boolean>>;
+}): JSX.Element {
+    const theme = useTheme();
+
+    const toggleCollapseHandler = () => {
+        props.collapseHandler((p) => !p);
+    };
+
+    return (
+        <AdvancedSearchSvg
```
Dzuming, line 24: can we add a cursor pointer for the icon, to indicate that it's a clickable element?

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+            type: t("panels.search.field.type", "Type"),
+            paramName: t("panels.search.field.paramName", "Label"),
+            paramValue: t("panels.search.field.paramValue", "Value"),
+            outputValue: t("panels.search.field.outputValue", "Output"),
+            edgeExpression: t("panels.search.field.edgeExpression", "Edge"),
+        }),
+        [t],
+    );
+
+    console.log(displayNames);
```
Dzuming, line 56: Do we need it :) ?

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+        event.preventDefault();
+
+        const formData = new FormData(event.currentTarget);
+
+        console.log(formData);
```
Dzuming, line 63: Do we need it :) ?

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+            })
+            .filter((input) => input !== "");
+
+        const finalText = transformedInputs.join(" ").trim() + " " + extractSimpleSearchQuery(filter);
+
+        setFilter(finalText);
+    };
+
+    const handleCancel = () => {
+        idRef.current.value = "";
```
Dzuming, line 78: Maybe we can improve it and just use form ref here?

```
const formRef = useRef<HTMLFormElement>()
...


 const handleCancel = () => {
        formRef.current.reset()

        setFilter(extractSimpleSearchQuery(filter));
    };

...

<Box
            ref={formRef}
            component="form"
```

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+                return transformInput(input, fieldName);
+            })
+            .filter((input) => input !== "");
+
+        const finalText = transformedInputs.join(" ").trim() + " " + extractSimpleSearchQuery(filter);
+
+        setFilter(finalText);
+    };
+
+    const handleCancel = () => {
```
Dzuming, line 77: Is it expected behavior, that when the filter is added and we want to clear it, the form is closed?


https://github.com/user-attachments/assets/ccb5a3a1-c11d-402a-915f-818bad33124d

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+            sx={{
+                display: "flex",
+                flexDirection: "column",
+                maxWidth: "300px",
+                justifyContent: "center",
+                alignItems: "center",
+            }}
+        >
+            <Typography fontWeight="bold" sx={{ m: 1 }}>
+                Advanced Search
```
Dzuming, line 102: Can we use i18n here?

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+            </SearchLabeledInput>
+            <SearchLabeledInput ref={typeRef} name="type">
+                <SearchLabel label={displayNames["type"]} />
+            </SearchLabeledInput>
+            <SearchLabeledInput ref={edgeExpressionRef} name="edgeExpression">
+                <SearchLabel label={displayNames["edgeExpression"]} />
+            </SearchLabeledInput>
+            <Box sx={{ display: "flex", flexDirection: "row", justifyContent: "space-between", width: "80%", mt: 2, mb: 1 }}>
+                <Button sx={{ width: "45%" }} size="small" variant="outlined" onClick={handleCancel}>
+                    Clear
```
Dzuming, line 127: Can we use i18n here?

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+                </Button>
+                <Button variant="contained" size="small" sx={{ width: "45%" }} type="submit">
+                    Submit
```
Dzuming, line 130: Can we use i18n here?

designer/client/src/components/toolbars/search/utils.ts
```diff
@@ -92,18 +109,92 @@ export function useFilteredNodes(filterValues: string[]): {
...
+                            .filter((e) => matchFilters(e.edgeType?.condition, [searchQuery.plainQuery]));
+
+                        groups = findFields([searchQuery.plainQuery], node)
+                            .concat(edges.length ? "edgeExpression" : null)
+                            .map((name) => displayNames[name])
+                            .filter(Boolean);
+
+                        return { node, edges, groups };
+                    } else {
+                        const edgesAux = allEdges
```
Dzuming, line 127: I didn't analyze this logic, but somewhere must be the problem, i tried to search just by the advanced search but seems like it doesn't work as expected


https://github.com/user-attachments/assets/a9223bb1-4a6e-40ab-bed9-39922f2678c7

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
```diff
@@ -0,0 +1,135 @@
...
+    while ((match = regex.exec(text)) !== null) {
+        lastIndex = regex.lastIndex;
+    }
+
+    const rest = text.slice(lastIndex).trim();
+
+    return rest;
+}
+
+export function AdvancedSearchFilters({
```
Dzuming, line 25: We should consider syncing advanced search filters, based on the simple search values, so:
1. When we clear a simple search that has advanced search queries, values in the form also should be cleared
2. When we type a query manually, it should be visible in the advanced search form 


https://github.com/user-attachments/assets/ebd3d730-83de-412e-be66-45b06a2860b3

Dzuming (changes requested):

designer/client/src/components/toolbars/search/AdvancedSearchFilters.tsx
DeamonDev: Yeah, I think it in a sense reasonable to expect that if someone want to clear filers then he wants to remove them all. If user would want to edit filter then he is able to do it on his own without clearing all the filters.

designer/client/src/components/toolbars/search/utils.ts
DeamonDev: Yeah, that was a bug. Fixed

Dzuming (approved):
LGTM

PR #6709, opened 2024-08-30
[NU-1790] Bump Flink to 1.19

mslabek:
## Checklist before merge

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/typedobject/TypedObjectBasedTypeInformation.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/typedobject/TypedObjectBasedTypeInformation.scala
@@ -1,5 +1,6 @@
 package pl.touk.nussknacker.engine.process.typeinformation.internal.typedobject
 
+import com.github.ghik.silencer.silent
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.flink.api.common.ExecutionConfig
 import org.apache.flink.api.common.typeinfo.TypeInformation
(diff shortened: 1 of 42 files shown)


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/typedobject/TypedObjectBasedTypeInformation.scala
```diff
@@ -171,7 +171,7 @@ abstract class TypedObjectBasedSerializerSnapshot[T] extends TypeSerializerSnaps
       val snapshotsToUse      = serializersSnapshots.filter(k => commons.contains(k._1))
 
       val fieldsCompatibility = CompositeTypeSerializerUtil.constructIntermediateCompatibilityResult(
```
mslabek, line 176: The signature of this is incompatible with 1.18

examples/installation/flink/flink-properties.yml
```diff
@@ -1,6 +1,5 @@
 taskmanager.numberOfTaskSlots: 8
-
-state.backend: filesystem
+state.backend.type: filesystem
```
mslabek, line 2: This will also have to be changed in quickstart

engine/flink/test-utils/src/main/scala/pl/touk/nussknacker/engine/flink/test/MiniClusterExecutionEnvironment.scala
```diff
@@ -87,6 +89,25 @@ class MiniClusterExecutionEnvironment(
     waitForJobStateWithAdditionalCheck(jobID, name, {}, expectedState: _*)(patience)
   }
 
+  def waitForJobStatusWithAdditionalCheck(
```
mslabek, line 92: For batch jobs even after the job is initialized, not all vertices are available at all times. There are moments when source is FINISHED and next vertices are not even registered leading to a passing check. Thats why for FINISHED jobs a better check is to check `executionGraph.getState`

PR #6737, opened 2024-09-03
[Nu-1794] fix problem with properties fields update

Dzuming:
## Describe your changes
- fix an issue with the properties update
- Resolve problem with logging
- Resolve fragment detection in components

## Checklist before merge


mslabek (approved):

lciolecki:
Here, it's not "problem with properties fields". There will appear bad filtering scenarios: fragments where component group was change will be not filtered.

Dzuming:
> Here, it's not "problem with properties fields". There will appear bad filtering scenarios: fragments where component group was change will be not filtered.
> 
> My proposition is to change commit name :)

The problem with components is about making a name as a link in the case of fragments, however, the name of the pr points to the problem with the properties update, there is a regression in the properties update. Check the description. this pr contains 3 separate fixes, however, I'm lazy that's why I mixed it here :P.

raphaelsolarski (approved):

PR #6766, opened 2024-09-05
[NU-1805] Scenario labels support

mateuszkp96:
## Describe your changes
Scenario labels support including:
Filtering scenario list by labels:
image
image

Adding labels to scenario;
image
image
image
image
image

Labels validation:
image
Labels validation for saved scenarios:
image

## Checklist before merge


designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
```diff
@@ -0,0 +1,315 @@
...
+        validateLabels(scenarioLabels);
+
+        if (!isEdited && scenarioLabels.length === 0) {
+            setShowEditor(false);
+        }
+
+        if (!showEditor && scenarioLabels.length !== 0) {
+            setShowEditor(true);
+        }
+    }, [scenarioLabels]);
```
Dzuming, line 172: Missing dependencies 'isEdited', 'showEditor', and 'validateLabels'

Dzuming:
I have a few general UX comments:

1. Maybe we can just hide this field from details when there is no label defined?
image

2. I'm able to add multiple tags with the same name, also this sorting is "strange" IMO because i would expect that the newest item would appear at the end of the tags,

https://github.com/user-attachments/assets/51da797e-2a7e-41f2-9d0a-562e1437a1ce

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
```diff
@@ -0,0 +1,315 @@
...
+                    getOptionKey={(option: string | LabelOption) => {
+                        if (typeof option === "string") {
+                            return option;
+                        }
+                        return toLabelValue(option);
+                    }}
+                    inputValue={input}
+                    isOptionEqualToValue={(v1: LabelOption, v2: LabelOption) => v1.value === v2.value}
+                    loading={isFetching || inputTyping}
+                    loadingText={inputTyping ? "Typing..." : "Loading..."}
```
Dzuming, line 218: Can we add i18n here?

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+    const setLabels = (labels: LabelOption[]) => {
+        const sortedLabels = labels.map(toLabelValue).sort((a, b) => a.localeCompare(b));
+        dispatch(editScenarioLabels(sortedLabels));
+    };
+
+    useEffect(() => {
+        validateLabels(scenarioLabels);
+
+        if (!isEdited && scenarioLabels.length === 0) {
```
Dzuming, line 165: How this logic related to the validateLabels? maybe we can move it to another useEffect? Also maybe it's worth to use meaningful names for this conditions

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
```diff
@@ -0,0 +1,315 @@
...
+                const validationError = response.data.validationErrors;
+                setLabelsErrors(validationError);
+            }
+        }, 500);
+    }, []);
+
+    const fetchAvailableLabels = useCallback(async () => {
+        setIsFetching(true);
+        const { data } = await HttpService.fetchScenarioLabels();
+        setIsFetching(false);
```
Dzuming, line 153: In this approach, there can be a situation, where this flag won't be set to true, for example, if fetchScenarioLabels throws an error. Can you use try {} finally {} here and move setIsFetching(false); to the finally block?

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
Dzuming: Also, I'm not sure what's the reason behind this logic, could you explain it?

Dzuming (changes requested):

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
```diff
@@ -0,0 +1,315 @@
...
+                        }
+                        return option.title;
+                    }}
+                    getOptionKey={(option: string | LabelOption) => {
+                        if (typeof option === "string") {
+                            return option;
+                        }
+                        return toLabelValue(option);
+                    }}
+                    inputValue={input}
```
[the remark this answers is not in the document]
mateuszkp96: I had problems with input when clearOnBlur was set (the input was cleared after typing values), so I used the input in the state

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
mateuszkp96: I moved them to separate effect and added proper comments

designer/client/src/components/toolbars/scenarioDetails/ScenarioLabels.tsx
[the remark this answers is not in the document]
Dzuming: I committed fix for the issue and removed controlled input

Dzuming (approved):
LGTM, as we discussed you can also add e2e test for this functionality

designer/client/cypress/e2e/labels.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+
+            cy.intercept("PUT", "/api/processes/*").as("save");
+
+            cy.get("[data-testid=AddLabel]").should("be.visible").click();
+
+            cy.get("[data-testid=LabelInput]").as("labelInput");
+
+            cy.get("@labelInput").should("be.visible").click().type("tagX");
+
+            cy.wait(1000); // wait for validation
```
Dzuming, line 28: Can't we intercept validation response and use `cy.wait('@labelvalidation')`?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ScenarioLabelsApiHttpServiceSecuritySpec.scala
```diff
@@ -0,0 +1,290 @@
...
+  "The scenario labels validation endpoint when" - {
+    "authenticated should" - {
+      "validate scenario labels" in {
+        given()
+          .when()
+          .basicAuthAdmin()
+          .body(
+            s"""
+               |{
+               |  "labels": ["tag100", "tag2"]
```
mslabek, line 142: Could you also add validation for duplicated labels? Right now if you try to have 2 of the same label I think the FE will just not make the request. WDYT about FE sending all the current labels and for `["tag2", "tag2"]` it would return error with info that labels cannot be duplicated / have to be unique?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ScenarioLabelsEntityFactory.scala
```diff
@@ -0,0 +1,37 @@
...
+import slick.lifted.{TableQuery => LTableQuery}
+import slick.sql.SqlProfile.ColumnOption.NotNull
+
+trait ScenarioLabelsEntityFactory extends BaseEntityFactory {
+
+  import profile.api._
+
+  val processesTable: LTableQuery[ProcessEntityFactory#ProcessEntity]
+
+  class ScenarioLabelsEntity(tag: Tag) extends Table[ScenarioLabelEntityData](tag, "tags") {
```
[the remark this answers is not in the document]
lciolecki: More, I think tags are more common definition the labels. I can also imagine that Components have tags. WDYT think to rename labels to tags?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/MetaData.scala
```diff
@@ -11,7 +11,11 @@ import pl.touk.nussknacker.engine.api.CirceUtil._
 
 // TODO: This class should be moved into components-api, scenario-api shouldn't use this. It should hold only properties
 //       and ScenarioRuntimeMetadata (Currently called ProcessVersion)
-@ConfiguredJsonCodec(encodeOnly = true) case class MetaData(id: String, additionalFields: ProcessAdditionalFields) {
+@ConfiguredJsonCodec(encodeOnly = true) case class MetaData(
+    id: String,
+    labels: List[String],
```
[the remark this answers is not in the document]
mateuszkp96: We agreed that we move labels in the metadata to the separate PR

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/ScenarioLabelsEntityFactory.scala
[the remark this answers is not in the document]
mateuszkp96: I used the old table `tags`. I added a migration for a new table - `scenario_labels`- and detached the `tags` table.

docs/MigrationGuide.md
mslabek, line 1: Can you add other changes to REST api? Like for example now the request for scenario PUT (under api/processes/{scenarioname}) requires `scenarioLabels` field. I don't know what other changes are there

mslabek, line 1: Please update changelog. This is an awesome feature that we need to brag about to our users, definitely will be in highlights in release notes in 1.18 :D

mslabek, line 1: Can you update the configuration docs? Generally, all configuration should be documented under Configuration section. Take a look at https://nussknacker.io/documentation/docs/configuration/DesignerConfiguration/#other-configuration-options

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/MigrationApiEndpoints.scala
```diff
@@ -52,22 +53,42 @@ class MigrationApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEn
...
-              engineSetupName = EngineSetupName("Flink"),
-              processCategory = "Category1",
-              scenarioGraph = exampleGraph,
-              processName = ProcessName("test"),
-              isFragment = false
-            )
-          )
-        )
+        jsonBody[MigrateScenarioRequestDto]
+        // FIXME uncomment examples when discriminator validation will work in the NuDesignerApiAvailableToExposeYamlSpec -
```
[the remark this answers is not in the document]
mateuszkp96: This problem occurred because it's the first DTO with a discriminator within endpoints generated from Tapir definitions. Here's the error in NuDesignerApiAvailableToExposeYamlSpec:
```
postApiMigrate > request > Example0
    $: must be valid to one and only one schema, but 2 are valid with indexes '0, 1'
postApiMigrate > request > Example1
    $: must be valid to one and only one schema, but 2 are valid with indexes '0, 1'
```
I spent some time trying to fix it, but it seems to be more complex than I thought. I wouldn't like to fix it within this PR, so I hid the problematic example. IMO it's a fine workaround for now.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ScenarioLabelsValidator.scala
```diff
@@ -0,0 +1,79 @@
...
+    } yield result).getOrElse(Validated.validNel(()))
+  }
+
+  private def validateLabels(
+      scenarioLabels: NonEmptyList[ScenarioLabel],
+      validationRules: NonEmptyList[ScenarioLabelConfig.ValidationRule]
+  ): Validated[NonEmptyList[ValidationError], Unit] = {
+    scenarioLabels
+      .traverse(label => validateScenarioLabel(validationRules, label).toValidatedNel)
+      .andThen(labels => validateUniqueness(labels))
```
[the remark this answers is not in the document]
mateuszkp96: FE sends the actual scenario labels too in a separate request. Anyway, FE shouldn't send a list with duplicates, but I would leave this check as some kind of guard

designer/server/src/main/resources/defaultDesignerConfig.conf
```diff
@@ -210,6 +210,19 @@ testDataSettings: {
...
 }
 
+scenarioLabelSettings: {
+  validationRules = [
+    {
+      validationPattern: "^[a-zA-Z0-9-_]+$",
+      validationMessage: "Scenario label can contain only alphanumeric characters, '-' and '_'"
+    }
+    {
+      validationPattern: "^.{1,4}$"
```
mslabek, line 220:
```suggestion
      validationPattern: "^.{1,20}$"
```
is it merged with dev-application.conf? can we throw this config from dev-application.conf?

mslabek (approved):
Awesome, great job :+1:

PR #6807, opened 2024-09-06
[NU-1701] Add conversions to primitive types to helper functions

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -18,11 +18,26 @@
     when entering into node form. Please be aware that changes in `tableDefinition.sql` are not refreshed.
     To do this, use `/app/processingtype/reload` API
 * [#6716](https://github.com/TouK/nussknacker/pull/6716) Fix type hints for #COLLECTION.merge function.
-* [#6695](https://github.com/TouK/nussknacker/pull/6680) From now on, arrays on UI are visible as lists but on a
+* [#6695](https://github.com/TouK/nussknacker/pull/6695) From now on, arrays on UI are visible as lists but on a 
   background they are stored as it is and SpeL converts them to lists in a runtime.
 * [#6750](https://github.com/TouK/nussknacker/pull/6750) Add varargs to `#COLLECTION.concat` and `#COLLECTION.merge`.
 * [#6778](https://github.com/TouK/nussknacker/pull/6778) SpeL: check for methods if a property for a given name does not exist.
 * [#6769](https://github.com/TouK/nussknacker/pull/6769) Added possibility to choose presets and define lists for Long typed parameter inputs in fragments.
+* [#6807](https://github.com/TouK/nussknacker/pull/6807) Add conversion functions to primitives to: `#CONV`:
+  * toNumberOrNull
+  * toString
+  * toBoolean
+  * toBooleanOrNull
+  * toInteger
+  * toIntegerOrNull
+  * toLong
+  * toLongOrNull
+  * toDouble
+  * toDoubleOrNull
+  * toBigInteger
+  * toBigIntegerOrNull
+  * toBigDecimal
+  * toBigDecimalOrNull
 
 ## 1.17
 
--- a/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
+++ b/utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
@@ -4,6 +4,8 @@ import pl.touk.nussknacker.engine.api.generics.GenericType
 import pl.touk.nussknacker.engine.api.{Documentation, HideToString, ParamName}
 import pl.touk.nussknacker.engine.util.functions.NumericUtils.ToNumberTypingFunction
 
+import scala.util.Try
+
 object conversion extends ConversionUtils
 
 trait ConversionUtils extends HideToString {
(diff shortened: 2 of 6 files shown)


docs/Changelog.md
```diff
@@ -18,10 +18,25 @@
     when entering into node form. Please be aware that changes in `tableDefinition.sql` are not refreshed.
     To do this, use `/app/processingtype/reload` API
 * [#6716](https://github.com/TouK/nussknacker/pull/6716) Fix type hints for #COLLECTION.merge function.
-* [#6695](https://github.com/TouK/nussknacker/pull/6680) From now on, arrays on UI are visible as lists but on a
+* [#6695](https://github.com/TouK/nussknacker/pull/6695) From now on, arrays on UI are visible as lists but on a 
   background they are stored as it is and SpeL converts them to lists in a runtime.
 * [#6750](https://github.com/TouK/nussknacker/pull/6750) Add varargs to `#COLLECTION.concat` and `#COLLECTION.merge`.
 * [#6778](https://github.com/TouK/nussknacker/pull/6778) SpeL: check for methods if a property for a given name does not exist.
+* [#6807](https://github.com/TouK/nussknacker/pull/6807) Add conversion functions to primitives to: `#CONV`:
```
lciolecki, line 25: Maybe we should call it #CONVERTER instead of #CONV?

lukasz-bigorajski: I would leave it as it to provide compatibility with older versions.

lciolecki (approved):
LGTM, good job!

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
```diff
@@ -20,4 +22,116 @@ trait ConversionUtils extends HideToString {
...
+  @Documentation(description = "Convert any value to String")
+  def toString(@ParamName("value") value: Any): java.lang.String = value match {
+    case v: String => v
+    case null      => null
+    case v         => v.toString
+  }
+
+  @Documentation(description = "Convert any value to Boolean or throw exception in case of failure")
+  def toBoolean(@ParamName("value") value: Any): java.lang.Boolean = value match {
+    case v: String            => java.lang.Boolean.valueOf(v)
```
mk-software-pl, line 41: IMO `java.lang.Boolean.valueOf` method is tricky. Because `java.lang.Boolean.valueOf("nu") == false` ;) I'm not sure if this is what we want here

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
```diff
@@ -20,4 +22,116 @@ trait ConversionUtils extends HideToString {
...
+    case v: Number => v.longValue()
+    case _         => null
+  }
+
+  @Documentation(description = "Convert any value to Double or throw exception in case of failure")
+  def toDouble(@ParamName("value") value: Any): java.lang.Double = value match {
+    case v: String => java.lang.Double.valueOf(numeric.toNumber(v).doubleValue())
+    case v: Number => v.doubleValue()
+    case null      => null
+    case _         => throw new IllegalArgumentException(s"Cannot convert: $value to Long")
```
mk-software-pl, line 89: "to Double"

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
```diff
@@ -20,4 +22,116 @@ trait ConversionUtils extends HideToString {
...
+  @Documentation(description = "Convert any value to Double or null in case of failure")
+  def toDoubleOrNull(@ParamName("value") value: Any): java.lang.Double = value match {
+    case v: String => Try(java.lang.Double.valueOf(numeric.toNumber(v).doubleValue())).getOrElse(null)
+    case v: Number => v.doubleValue()
+    case _         => null
+  }
+
+  @Documentation(description = "Convert any value to BigInteger or throw exception in case of failure")
+  def toBigInteger(@ParamName("value") value: Any): java.math.BigInteger = value match {
+    case v: String               => new java.math.BigInteger(v)
```
mk-software-pl, line 101: I'm not sure if we should rely on this BigInteger constructor. Maybe it's better to use the `numeric` util also here?

arkadius (approved):
LGTM

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
lukasz-bigorajski: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
lukasz-bigorajski: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
lukasz-bigorajski: Done

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
mk-software-pl: Can we do the same in case of BigDecimal constructor?

mk-software-pl (approved):
LGTM. One more comment (similar to one of my previous ones)

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
lukasz-bigorajski: Yes, but I would rather will not do it because in numeric we try to parse a double at first try and I am not sure if it's ok to do it in a such order.

PR #6823, opened 2024-09-09
[NU-1772] Scenario activities BE implementation

mgoworko:
## Describe your changes

What is done:
- new API contract for FE (in PR https://github.com/TouK/nussknacker/pull/6822)
- BE implemenetation of that API in this PR
- new table that replaces actions and comments
- possibility to save and return to FE all scenario-related activities
- full backwards compatibility with old FE API, and old Action mechanism, only changed the db underneath
- migration of data from 2 old tables - comments and actions - to new "scenario_activities" table
- some integration tests of the new API

What is left to do in further PRs
1. Veryfing what is already done and saving information of scenario activities that has not been saved until now (like incoming and outgoing migrations, automatic updates of scenarios etc.)
2. Integrating periodics into this new scenario activity mechanism.
3. Clean up endpoint with scenario details, to not include versions and comments, they will be handled by this new API

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/ScenarioActivityRepository.scala
```diff
@@ -0,0 +1,81 @@
...
+import db.util.DBIOActionInstances.DB
+import pl.touk.nussknacker.engine.api.deployment.{ScenarioActivity, ScenarioActivityId}
+import pl.touk.nussknacker.engine.api.process.{ProcessId, VersionId}
+import pl.touk.nussknacker.ui.api.description.scenarioActivity.Dtos.Legacy
+import pl.touk.nussknacker.ui.db.entity.AttachmentEntityData
+import pl.touk.nussknacker.ui.process.ScenarioAttachmentService.AttachmentToAdd
+import pl.touk.nussknacker.ui.process.repository.activities.ScenarioActivityRepository.ModifyCommentError
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+
+trait ScenarioActivityRepository {
```
arkadius, line 12: 1. I see that you directly use the repository to record activity. I already do the service (`newactivity.ActivityService`) which was designed to handle it. We try to follow three layer architecture, IMO services shouldn't call their own repositories directly (they should call services). WDYT?
2. When I see multiple similar addXYZ methods, I have a feeling that it would be better to use the command pattern. Thanks to that we can extract a parent class for similar activities and have some common handlers for similar things. Also, it is easier to distinguish between a next adding method and totally different method like `findActivities`

lukasz-bigorajski (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/ScenarioActivityRepository.scala
mgoworko: Ad 1 
I totally agree with 3-layer architecture, but would like to do it in next PR that I've started working on today. In next PR I intended to add handling of scenario activities, that were not persisted before (and were not added in this PR because of huge scope), like incoming and outgoing migrations, or technical updates. So:
- in this PR I replaced old ProcessActions with ScenarioActivities, but in some places I tried to not make too huge changes
- in next PR I would refactor it, including the service layer in between
 
 Would it be ok?
 
Ad 2 
Did you mean command pattern on repository level, or only service level? I would prefer to do it in next PR too, because it is related to the refactor from Ad 1

So I agree on both points, but would like to do this refactor in next PR I already started working on.

arkadius: The command pattern is valuable in any layer - everwhere where there are multiple similar methods. I see that you already did something in this direction in this PR as well. With regards to moving mentioned reactors to the next PR - it is fine for me :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/DbScenarioActivityRepository.scala
```diff
@@ -0,0 +1,904 @@
...
+  private def additionalPropertyFromEntity(entity: ScenarioActivityEntityData, name: String): Either[String, String] = {
+    entity.additionalProperties.properties.get(name).toRight(s"Missing additional property $name")
+  }
+
+  private def fromEntity(entity: ScenarioActivityEntityData): Either[String, (Long, ScenarioActivity)] = {
+    entity.activityType match {
+      case ScenarioActivityType.ScenarioCreated =>
+        ScenarioActivity
+          .ScenarioCreated(
+            scenarioId = scenarioIdFromEntity(entity),
```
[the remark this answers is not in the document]
arkadius: I didn't analyze it to deep. If you say so, let's stay with current approach for now.

arkadius (approved):
The parts of the code that I read are fine for me, let's w8 for @lukasz-bigorajski approve as he checked it more extensively..

designer/server/src/main/scala/db/migration/V1_055__CreateScenarioActivitiesDefinition.scala
```diff
@@ -0,0 +1,122 @@
...
+import db.migration.V1_056__MigrateActionsAndCommentsToScenarioActivitiesDefinition.logger
+import pl.touk.nussknacker.ui.db.migration.SlickMigration
+import slick.jdbc.JdbcProfile
+import slick.sql.SqlProfile.ColumnOption.NotNull
+
+import java.sql.Timestamp
+import java.util.UUID
+import scala.concurrent.ExecutionContext.Implicits.global
+
+trait V1_055__CreateScenarioActivitiesDefinition extends SlickMigration with LazyLogging {
```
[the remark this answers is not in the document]
lukasz-bigorajski: No, you shouldn't. For me it's just faster. If you use ANSI SQL the query should be the same for HSQL and postgres.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/DbScenarioActivityRepository.scala
```diff
@@ -0,0 +1,860 @@
...
+
+  def deleteComment(
+      scenarioId: ProcessId,
+      activityId: ScenarioActivityId,
+  )(implicit user: LoggedUser): DB[Either[ModifyCommentError, Unit]] = {
+    modifyActivityByActivityId(
+      activityId = activityId,
+      activityDoesNotExistError = ModifyCommentError.ActivityDoesNotExist,
+      validateCurrentValue = validateCommentExists(scenarioId),
+      modify = _.copy(comment = None, lastModifiedByUserName = Some(user.username)),
```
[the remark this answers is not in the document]
lukasz-bigorajski: I like this solution

lukasz-bigorajski (approved):
LGTM! I responded to the questions.

mgoworko:
Yesterday I tested this change using real data from prod environments. It turned out, that migration was not fully correct, and existing test did not catch that. Some data, including standalone comments, was missing.

Done after review and approve:
- added 9 tests for migrating each old process action type and standalone comments
- fixed migration, including proper handling of comment prefixes and standalone comments
- change in API contract - finishedAt field made optional in two activities, comment added for "run now" activity

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/DbScenarioActivityRepository.scala
```diff
@@ -875,17 +876,45 @@ class DbScenarioActivityRepository(override protected val dbRef: DbRef, clock: C
...
+  //  until next PR is merged there is parsing from comment content to preserve full compatibility
+  private def extractOldNameAndNewNameForRename(
+      entity: ScenarioActivityEntityData
+  ): Either[String, (String, String)] = {
+    val fromAdditionalProperties = for {
+      oldName <- additionalPropertyFromEntity(entity, "oldName")
+      newName <- additionalPropertyFromEntity(entity, "newName")
+    } yield (oldName, newName)
+
+    val legacyCommentPattern = """Rename: \[(.+?)\] -> \[(.+?)\]""".r
```
lukasz-bigorajski, line 904: I wonder whether NU was adding comments in such format or any external system? If it was any external system then this logic should not be here and maybe performed manually in target system.

designer/server/src/test/scala/db/migration/V1_057__MigrateActionsAndCommentsToScenarioActivities.scala
```diff
@@ -0,0 +1,379 @@
...
+        Some(ScenarioVersion(processVersionId)),
+      )
+    )
+  }
+
+  private def run[T](action: DBIO[T]): T = Await.result(runner.run(action), Duration.Inf)
+
+  private def processEntity(user: String, timestamp: Timestamp) = ProcessEntityData(
+    id = ProcessId(-1L),
+    name = ProcessName("2024_Q3_6917_NETFLIX"),
```
lukasz-bigorajski, line 304: I would anonymize this data.

lukasz-bigorajski (approved):
LGTM. One question.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/DbScenarioActivityRepository.scala
mgoworko: - Before this PR, NU was adding such comments itself, in `pl.touk.nussknacker.ui.process.repository.ProcessRepository`.
- In this PR it is replaced with adding `ScenarioActivity.ScenarioNameChanged` instead of plain text comment
- But old comments are not fully migrated in the db - because it would require parsing comment during the migration and extracting oldName and newName


Out of 17 types of activities, that are supposed to be handled according to analysis od NU-1772:
- 10 are fully migrated and handled in this PR
- 4 were not saved at all (ScenarioCreated, ChangedProcessingMode, OutgoingMigration, PerformedScheduledExecution), will be added in next PR
- 2 were saved, and are still saved, as plain text comments, not named activities (IncomingMigration, AutomaticUpdate)
- 1 remaining, the ScenarioNameChanged, was added in this PR (but old comments will have to be migrated alongise IncomingMigration and AutomaticUpdate comments in next PR, then this todo will be resolved)

In the next PR, ETA next week, I will add those 4 activities that were not saved until now, and will migrate those 3, that were plain text comments until now.

designer/server/src/test/scala/db/migration/V1_057__MigrateActionsAndCommentsToScenarioActivities.scala
mgoworko: done

lukasz-bigorajski (approved):

PR #6875, opened 2024-09-16
[NU-1701] Introduce extension methods to SpeL. Add canCastTo and cast…

lukasz-bigorajski:
…To to SpeL.

## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionExtractor.scala
```diff
@@ -47,8 +50,9 @@ class ClassDefinitionExtractor(settings: ClassExtractionSettings) extends LazyLo
   ): Map[String, List[MethodDefinition]] = {
     val membersPredicate = settings.visibleMembersPredicate(clazz)
     val methods          = extractPublicMethods(clazz, membersPredicate, staticMethodsAndFields)
+    val extensionMethods = extractExtensionMethods(membersPredicate, staticMethodsAndFields)
     val fields           = extractPublicFields(clazz, membersPredicate, staticMethodsAndFields).mapValuesNow(List(_))
-    filterHiddenParameterAndReturnType(methods ++ fields)
+    filterHiddenParameterAndReturnType(methods ++ fields ++ extensionMethods)
```
arkadius, line 55: I think that we shouldn't add these extensions methods into `ClassDefinition`. It is not a thing that changes for each class. It is a thing that is added to every object. We should build some wrapper for `ClassDefinition` that would inject these methods in place of usage of `methods` method. Thanks to that, we will be flexible in further changes, e.g. we could add some extension methods only in some specific context of typing. Also, thanks to this approach we avoid hacks like `ClassDefinitionAssertions`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
```diff
@@ -0,0 +1,62 @@
...
+object CastTyping {
+
+  class Typing extends TypingFunction {
+
+    override def computeResultType(
+        arguments: List[typing.TypingResult]
+    ): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+      case TypedObjectWithValue(_, clazzName: String) :: Nil =>
+        Try(Class.forName(clazzName)) match {
+          case Success(clazz) => Typed.typedClass(clazz).validNel
```
arkadius, line 32: We should check if casting is possible during typing - e.g. we shouldn't allow to `"some string".castTo("Long")`. I see that we have something like `TypingFunctionForClassMember` for that purpose

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ExpressionSuggesterSpec.scala
```diff
@@ -766,6 +770,12 @@ class ExpressionSuggesterSpec
     )
   }
 
+  test("should suggest methods for unknown") {
+    spelSuggestionsFor("#unknown.") shouldBe List(
+      suggestion("canCastTo", Typed[Boolean]),
```
mslabek, line 775: Under default settings (with no dynamic access on default), this is misleading. As I understand it, by default we allow casting only on Maps and Lists? If we are to add this suggestions on Unknown, this makes some problems, like suggestions on things like `unknown.` without `#` or `#UTIL.canCastTo()` I think we should not have any suggestions on Unknown.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
mslabek: Theres also some security concerns if we dont do some check.

1. You can infer what types are on classPath - for example: `{}.castTo("pl.touk.nussknacker.ui.server.NuAkkaHttpServerInterpreterForTapirPurposes")` will return type `NuAkkaHttpServerInterpreterForTapirPurposes
2. You can cast to types which allow for operations we shouldn't allow. For example `#input.myMap.castTo("java.util.HashMap").remove("key1")` will do a mutation.

arkadius: I think that `.remove` won't work, because this method wasn't discovered for `HashMap` type, but `Class.forName` with any class available in the classpath indeed looks like something potentially malicious (but currently I don't know how it can be used). I have one more notice - I think that the classloader can be picked incorrectly, see the implementation in JRE:
```
        Class<?> caller = Reflection.getCallerClass();
        return forName0(className, true, ClassLoader.getClassLoader(caller), caller);
```

lukasz-bigorajski: > I think that the classloader can be picked incorrectly, see the implementation in JRE|
It's true, I noticed it yesterday.
> I think that .remove won't work, because this method wasn't discovered for HashMap type
This is true
> with any class available in the classpath indeed looks like something potentially malicious 
This is also true, we can restrict the classes to what `ClassDiscovery` returns but I wonder whether this is enough.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -1353,6 +1354,70 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
...
+        ("11.canCastTo('java.lang.String')", "canCastTo", "Integer"),
+        ("true.canCastTo('java.lang.String')", "canCastTo", "Boolean"),
+        ("'true'.canCastTo('java.lang.String')", "canCastTo", "String"),
+        ("11.castTo('java.lang.String')", "castTo", "Integer"),
+        ("true.castTo('java.lang.String')", "castTo", "Boolean"),
+        ("'true'.castTo('java.lang.String')", "castTo", "String"),
+      )
+    ) { (expression, expectedMethod, expectedType) =>
+      parse[Any](expression, ctx).invalidValue.toList should matchPattern {
+        case UnknownMethodError(expectedMethod, expectedType) :: Nil =>
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ExpressionSuggesterSpec.scala
lukasz-bigorajski: I cannot agree with you on this. I think `toString` and casting methods should be available on the unknown and invoking other methods should be limited by a property: `methodExecutionForUnknownAllowed` because we cannot be sure whether other methods are available for unknown. Without this behaviour `casting` methods are rather unusable.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
mslabek: Run this test in SpelExpressionSpec:
```scala
test("debug") {
    import scala.collection.JavaConverters._
    val newctx = ctx.withVariable("hashMap", new java.util.HashMap[String, Int](Map("meh" -> 2).asJava))
    val result =
      parse[Any]("#hashMap.castTo('java.util.HashMap').remove('meh')", newctx).validExpression.evaluateSync[Any](newctx)
    println(result)
  }
```
The problem seems to be here https://github.com/TouK/nussknacker/blob/3e70f0020b115816b8aeb3801426957ba615f5f5/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/typer/MethodReferenceTyper.scala#L38 - returns None and then its mapped to Right(Unknown) and gets executed. I think this is the case for all classes which have no classDefinition

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ExpressionSuggesterSpec.scala
mslabek: Ok, I agree casting will be kind of useless when it would be working only on Lists/Arrays and Maps. Im not opposed to adding casting on Unknown and having these suggestions, there may be some edge cases to cover

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,21 @@
+package pl.touk.nussknacker.engine.extension
+
+import java.lang.reflect.Method
+
+object ExtensionMethods {
+
+  private val declarationsWithImplementations = Map(
+    classOf[Cast] -> CastImpl.apply _,
+  ).asInstanceOf[Map[Class[_], Any => Any]]
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def canCastTo(clazzType: String): Boolean =
+    Class.forName(clazzType).isAssignableFrom(target.getClass)
+
+  override def castTo[T](clazzType: String): T = Try {
+    val clazz = Class.forName(clazzType)
+    if (clazz.isInstance(target)) {
+      clazz.cast(target).asInstanceOf[T]
+    } else {
+      throw new IllegalArgumentException(s"Cannot cast: ${target.getClass} to: $clazzType")
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

docs/Changelog.md
```diff
@@ -45,6 +45,8 @@
 * [#6826](https://github.com/TouK/nussknacker/pull/6826) Security fix: added validation of expression used inside 
   indexer for Maps and Lists (for example `{1,2,3}[#otherList.remove(1) == null ? 0 : 0]`). This allowed executing 
   some types of unallowed expressions.
+* [#6840](https://github.com/TouK/nussknacker/pull/6840) Introduce extension methods in SpeL. Add canCastTo and castTo
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
lukasz-bigorajski: After some change locally I have:
```
Invalid(NonEmptyList(GenericFunctionError(java.util.HashMap is not allowed), IllegalInvocationError(Unknown)))
```

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionExtractor.scala
lukasz-bigorajski: Done

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -1353,6 +1354,70 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
...
+
+  test("should compute correct result type based on parameter") {
+    val parsed = parse[Any]("{11, 12}.castTo('java.util.Collection')", ctx).validValue
+    parsed.returnType shouldBe Typed.typedClass(classOf[java.util.Collection[_]])
+    parsed.expression.evaluateSync[Any](ctx) shouldBe a[java.util.Collection[_]]
+  }
+
+  test("should return an error if the cast return type cannot be determined at parse time") {
+    parse[Any]("{11, 12}.castTo('java.util.XYZ')", ctx).invalidValue.toList should matchPattern {
+      case ArgumentTypeError("castTo", _, _) :: Nil =>
```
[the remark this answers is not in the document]
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/AllowedCastParametersClasses.scala
```diff
@@ -0,0 +1,29 @@
...
+    namesWithTyping.get(className)
+}
+
+object AllowedCastParametersClasses {
+
+  def apply(set: ClassDefinitionSet): AllowedCastParametersClasses =
+    new AllowedCastParametersClasses(
+      namesWithTyping = set.classDefinitionsMap
+        .map { case (clazz, classDefinition) =>
+          clazz.getName -> Try(classDefinition.clazzName).toOption
```
arkadius, line 19: Why this is wrapped in Try?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
```diff
@@ -0,0 +1,58 @@
...
+object CastTyping {
+
+  def castToTyping(allowedClasses: AllowedCastParametersClasses)(
+      instanceType: typing.TypingResult,
+      arguments: List[typing.TypingResult]
+  ): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+    case TypedObjectWithValue(_, clazzName: String) :: Nil =>
+      allowedClasses.get(clazzName) match {
+        case Some(typing) => typing.validNel
+        case None         => GenericFunctionTypingError.OtherError(s"'$clazzName' is not allowed").invalidNel
```
arkadius, line 31: "Casting to ... is not allowed" maybe?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
```diff
@@ -0,0 +1,102 @@
...
+  // This functionality can be easily extended using dynamic methods discovery what you only need to do is:
+  // - Take all extension classes from class: ExtensionMethods, and then filter out those classes that are statically defined e.g. using annotation.
+  // - Filter out static and scala varargs methods (Method.isStatic and ClassDefinitionExtractor.javaVersionOfVarArgMethod)
+  // - Then invoke ClassDefinitionExtractor.extractMethod (you can easily build ClassDefinitionExtractor using settings).
+  // - After this we should filter out hidden members using ClassDefinitionExtractor.filterHiddenParameterAndReturnType.
+  // - And calculated definitions should be combined with static definitions.
+  private val extensionMethodsMap: Map[Method, Map[String, List[MethodDefinition]]] =
+    createStaticallyDefinedMethodsMap(set)
+
+  val value = new ClassDefinitionSet(
```
arkadius, line 33: Why we have set and value and both are public? :)

arkadius: Also, please move `value` and `extensionMethodsMap` initialization outside of class body, to factory method (apply/create). The general rule is to not have too much code inside the constructors. Body of class is the same as constructor in java.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ClassExtractionSettings.scala
```diff
@@ -223,8 +227,27 @@ object ClassExtractionSettings {
       )
     )
 
+  lazy val ExtensionExcludedMembers: List[ClassMemberPredicate] = List(
```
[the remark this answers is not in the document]
arkadius: Also, I think that we shouldn't put this predicates into this class. This class has wrong name but the responsibility of it is to setup which members should be visible during class members discovery process. 

I think that the only thing that we need is to check that given class has some descendants in the set of other classes and this check should be done somewhere else.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+  private def findMethodsForClass(clazz: Class[_]): Map[String, List[MethodDefinition]] = {
+    val membersPredicate = settings.visibleMembersPredicate(clazz)
+    extensionMethodsMap
+      .filterKeysNow(membersPredicate.shouldBeVisible)
+      .flatMap(_._2)
+  }
+
+  private def createStaticallyDefinedMethodsMap(
```
arkadius, line 52: We won't need to this whole reflection magic with `castClass.getDeclaredMethod`, wrapping set with  `AllowedCastParametersClasses`, use internal `VisibleMembersPredicate` class of `ClassExtractionSettings` here, if we replace usage of `ClassExtractionSettings` with one simple rule described above. This rule could be placed near `Cast`/`CastTyping` to keep all casting logic in one place.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object ExtensionMethods {
+
+  private val declarationsWithImplementations = Map[Class[_], Constructor](
+    classOf[Cast] -> new CastConstructor,
+  )
+
+  val registry: Set[Class[_]] = declarationsWithImplementations.keySet
+
+  def applies(clazz: Class[_]): Boolean = registry.contains(clazz)
```
arkadius, line 13: You could replace applies and invoke with `PartialFunction`. Thanks to that we would have a lower number of entry points to this class and the interface would be more understandable.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import java.lang.reflect.Method
+
+object ExtensionMethods {
+
+  private val declarationsWithImplementations = Map[Class[_], Constructor](
+    classOf[Cast] -> new CastConstructor,
+  )
+
+  val registry: Set[Class[_]] = declarationsWithImplementations.keySet
```
arkadius, line 11: private

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,32 @@
...
+      .get(method.getDeclaringClass)
+      .map { case constructor: CastConstructor =>
+        constructor.apply(target, classLoader)
+      }
+      .map(impl => method.invoke(impl, arguments: _*))
+      .getOrElse {
+        throw new IllegalArgumentException(s"Extension method: ${method.getName} is not implemented")
+      }
+
+  sealed trait Constructor
```
arkadius, line 26: `Constructor` sounds like something releated with handling `new ...` in Spel. Maybe just call it `ExtensionMethodImplFactory`?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  class CastConstructor extends Constructor {
```
arkadius, line 28: I would move it near `CastImpl` to keep all cast related things in one place

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -0,0 +1,32 @@
...
+  )
+
+  val registry: Set[Class[_]] = declarationsWithImplementations.keySet
+
+  def applies(clazz: Class[_]): Boolean = registry.contains(clazz)
+
+  def invoke(method: Method, target: Any, arguments: Array[Object], classLoader: ClassLoader): Any =
+    declarationsWithImplementations
+      .get(method.getDeclaringClass)
+      .map { case constructor: CastConstructor =>
```
arkadius, line 18: For now, instead of having sealed trait and pattern match on it, we could have this one:
```
def create(target: Any, classLoader: ClassLoader): Any 
```
in this interface.

lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
lukasz-bigorajski: Done

docs/scenarios_authoring/Spel.md
```diff
@@ -376,3 +376,11 @@ On the other hand, formatter created using `#DATE_FORMAT.formatter()` method wil
...
 
 For full list of available format options take a look at [DateTimeFormatter api docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html).
+
+## Casting.
+
+When a type cannot be determined by parser, the type is presented as `Unknown`. When we know what the type will be on
+runtime, we can cast a given type, and then we can operate on the cast type.
+
+E.g. having a variable `obj` of a type: `List[Unknown]` and we know the elements are strings then we can cast elements
+to String: `#obj.![#this.canCastTo('java.lang.String') ? #this.castTo('java.lang.String') : null]`.
```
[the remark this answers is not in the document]
lukasz-bigorajski: I guess it will not be easy, but I will check it in next commit.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
```diff
@@ -0,0 +1,34 @@
...
+      )
+    )
+  }
+
+  private def enrichWithExtensionMethods(
+      clazz: Class[_],
+      classDefinition: ClassDefinition,
+      castMethodDefinitions: CastMethodDefinitions
+  ): ClassDefinition =
+    castMethodDefinitions.createDefinitions(clazz) match {
```
arkadius, line 29: This `isEmpty` check is redundant, let's inline this `classDefinition.copy(methods = classDefinition.methods ++ ext)` in `apply`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
```diff
@@ -0,0 +1,117 @@
...
+
+private[extension] class CastMethodDefinitions(private val classesWithTyping: Map[Class[_], TypingResult]) {
+
+  def createDefinitions(clazz: Class[_]): Map[String, List[MethodDefinition]] =
+    classesWithTyping.filterKeysNow(c => isAssignable(clazz, c)) match {
+      case allowedClasses if allowedClasses.isEmpty => Map.empty
+      case allowedClasses                           => definitions(allowedClasses)
+    }
+
+  private def isAssignable(clazz: Class[_], targetClazz: Class[_]): Boolean =
```
arkadius, line 50: I think that inlining this method would give more benefits. I was confused, because this method doesn't check that class is assignable but also that it is not the same class. Instead of this method, I would create a local variable `childTypes` to explain why we do this filtering.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
```diff
@@ -0,0 +1,34 @@
+package pl.touk.nussknacker.engine.extension
+
+import pl.touk.nussknacker.engine.ModelData
+import pl.touk.nussknacker.engine.definition.clazz.{ClassDefinition, ClassDefinitionSet}
+
+final case class ClassDefinitionSetWithExtensionMethods private (value: ClassDefinitionSet)
+
+object ClassDefinitionSetWithExtensionMethods {
```
arkadius, line 8: This class became small and readable. Nice! I'm thinking about one more little step that we can do to make it even better. We could rename it to sth like `ExtensionMethodsDefinitionsEnricher` and rename `apply` to `enrichWithExtensionMethods`. It could be even in the same file as `ExtensionMethods`. Thanks to that we won't introduce another entity, and keeps all extension methods registry logic together. It will be easier to find out where to add another extension method. WDYT?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  def apply(modelData: ModelData): ClassDefinitionSetWithExtensionMethods =
```
arkadius, line 10: Not used

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
```diff
@@ -0,0 +1,117 @@
...
+}
+
+private[extension] object CastImpl extends ExtensionMethodsImplFactory {
+  override def create(target: Any, classLoader: ClassLoader): Any =
+    new CastImpl(target, classLoader)
+}
+
+private[extension] class CastMethodDefinitions(private val classesWithTyping: Map[Class[_], TypingResult]) {
+
+  def createDefinitions(clazz: Class[_]): Map[String, List[MethodDefinition]] =
```
arkadius, line 44: Maybe `extractDefinitions` or `queryDefinitions` as it doesn't always create anything.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/AllowedCastParametersClasses.scala
arkadius: What is done? I was only asking :)  + I see that it sill exists. `clazzName` is a field, that it won't throw an exception.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
arkadius: Nice

docs/scenarios_authoring/Spel.md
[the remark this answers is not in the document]
arkadius: Do you mean in the next PR?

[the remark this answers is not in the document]
arkadius: Also:
- We have `castToOrNull` now so it can be shortened
-  I think that it would be nice, to list each of these three methods in bullets with short explanation. Zibi would appreciate that

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ClassDefinitionSetWithExtensionMethods.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Cast.scala
lukasz-bigorajski: Done

arkadius (approved):
Nice change

docs/scenarios_authoring/Spel.md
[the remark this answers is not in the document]
lukasz-bigorajski: Done

PR #6916, opened 2024-09-19
Highlights for version 1.17

jzdr:


docs/Changelog.md
```diff
@@ -55,7 +55,15 @@
 
 #### Highlights
 
-(Not available yet)
+##### End-user
+
+* Flink Catalogs integration. Thanks to Catalogs, Nu can be used to act on data stored in Data Lakehouses.
+* Improvments in batch mode, including joins, testing (dry runs) and general usability.
+* Suggestions and autocompletion now also work in invalid expressions. When working on an expression you don't have to keep it valid all the time for suggestions and autocompletion to work.
```
arkadius, line 62: How autompletion is different from suggestions? :)

I think that for end-user, worth mention are also changes around SpEL expressions usability:
- Added `sortedAscBy` and `reverse` functions to `#COLLECTION` helper
- Arrays and Lists are unified now - you don't need to think, how to combine them together

SQL enricher also was improved - it does some type alignments that makes operating on input data easier.

Also, we did some performance optimization around Avro processing. The reduced CPU utilization quite significantly. Thanks to that, our users can reduce costs of data processing.

arkadius (approved):

PR #6925, opened 2024-09-20
Preserve typing information during expression compilation

Elmacioro:
## Describe your changes

Currently, there is a problem with nodes that utilize presets as sometimes labels are not present and are presented as `null`. 
This happens when we want to use a component or fragment with presets, select some values for them and save the scenario. 
If there were any errors during compilation of that node presets will have null labels.

Zrzut ekranu z 2024-09-20 12-42-53

This happens because `ProcessDictSubstituor` dynamically adds and removes presets' labels.
```
      if (expr.language == Expression.Language.DictKeyWithLabel && !expr.expression.isBlank) {
        if (isReverse)
          addLabelToDictKeyExpression(process, expr, optionalExpressionTypingInfo, nodeExpressionId)
        else
          removeLabelFromDictKeyExpression(process, expr, nodeExpressionId) // no need to keep label in BE
```

To add labels at that level we need parameters' types which are assigned during compilation. Unfortunately when there is any custom validators error during node compilation we lose typing information.
in `ExperssionCompiler.compileNodeParameters()`
```
    allCompiledParams
      .andThen(allParams => Validations.validateWithCustomValidators(allParams, paramValidatorsMap))
      .combine(redundantMissingValidation.map(_ => List()))
```
allCompiledParams contain the typing info and is of `ValidatedNel` type. When `Validations.validateWithCustomValidators(allParams, paramValidatorsMap)` is `Invalid` we get these errors without previous typing.

To solve this issue I introduced a change which will still pass typing info further even when there are some custom validator errors. Instead of using `Validated` which can represent two states: `Valid` and `Invalid`, I used `Ior` which lets us have 3 states: `Left`, `Right` and `Both`. First two are self-explanatory and `Both` can hold both failure and a success. This way we can preserve the typing info while also preserving errors from custom validation. After testing these changes locally, we preserve the typing info and have correct preset labels even when there were errors during compilation.

Zrzut ekranu z 2024-09-20 13-52-13

## Checklist before merge

### DIFF
--- a/designer/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
+++ b/designer/server/src/test/scala/pl/touk/nussknacker/ui/integration/DictsFlowTest.scala
@@ -2,6 +2,7 @@ package pl.touk.nussknacker.ui.integration
 
 import com.typesafe.config.Config
 import io.circe.Json
+import io.circe.parser._
 import io.circe.syntax.EncoderOps
 import org.scalatest.OptionValues
 import org.scalatest.funsuite.AnyFunSuiteLike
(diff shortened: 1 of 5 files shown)


arkadius (approved):
Good job, one small comment added. Merge it after fix it

PR #6958, opened 2024-09-26
Kafka exceptionHandler: retry when message is too large, log when writing to Kafka fails

piotrp:
## Describe your changes

Log errors that are thrown from Kafka producer. Try to somehow handle situations where we try to write exception info that is too large to log.

## Checklist before merge


utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
```diff
@@ -14,11 +16,123 @@ import java.io.{PrintWriter, StringWriter}
...
+      serializedValue
+    }
+  }
+
+  private def removeInputEventKeys(value: Json, valueBytes: Int, maxValueBytes: Int): Json = {
+    if (valueBytes <= maxValueBytes) {
+      return value
+    }
+
+    val removedBytesPlaceholder = "$" * valueBytes.toString.length
```
arkadius, line 42: You use valueBytes length for the placeholder based on which you compute bytesToCut, and then you replace this placeholder with removedBytes length which is different.  Also, you don't consider the size of the variable keys (I know why, but it wasn't straightforward when I read it for the first time). I see some hidden assumptions here - please document them in the comment.

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
```diff
@@ -14,11 +16,123 @@ import java.io.{PrintWriter, StringWriter}
...
+
+    val variablesWithLength               = countVariableLengths(value.hcursor.downField("inputEvent"))
+    val (variablesToRemove, removedBytes) = calculateKeysToRemove(variablesWithLength, bytesToCut)
+
+    if (removedBytes <= warningBytes) {
+      // this may happen only when doing tests with really low size limits
+      return value
+    }
+
+    val finalMessage = messageTemplate.replace(removedBytesPlaceholder, removedBytes.toString) +
```
arkadius, line 64: We could have a message template as a format for the `String.format` purpose and fill it with some placeholders for count `bytesToCut` and for final message preparation. Thanks to that we would avoid tricky dollars replacement and the template would show the full message including variable names.

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
piotrp: Explanation added, for both

utils/kafka-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaJsonExceptionSerializationSchema.scala
piotrp: I decided to do it this way (non-standard placeholder + concatenating strings) because then i don't have to do more math on the amount of space reserved for the warning message. If I used `String.format` placeholders then I would have to account for them in `warningBytes`, and I didn't want to make this code even more complicated.

piotrp:
I pushed a change that makes variable removal function public - I need it in a custom exception serializer

arkadius (approved):

PR #6969, opened 2024-10-02
Add canCastToRecord, castToRecord and castToRecordOrNull extension me…

lukasz-bigorajski:
…thods

## Checklist before merge

### DIFF
--- a/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
+++ b/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
@@ -1510,6 +1510,151 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
     ).validExpression.evaluateSync[Any](ctx) shouldBe "unknown"
   }
 
+  test("should cast to record and filter out unwanted keys") {
+    val customCtx = ctx.withVariable(
+      "undiscoveredMap",
+      ContainerOfUnknown(Map("int1" -> 2, "str1" -> "str", "int2" -> 3).asJava)
+    )
+    parse[Any](expr = "#undiscoveredMap.value", context = customCtx).validValue.returnType shouldBe Unknown
+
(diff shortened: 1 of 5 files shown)


scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -1510,6 +1510,171 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
...
   }
 
+  test("should cast to record and filter out unwanted keys") {
+    val customCtx = ctx.withVariable(
+      "undiscoveredMap",
+      ContainerOfUnknown(Map("int1" -> 2, "str1" -> "str", "int2" -> 3).asJava)
+    )
+    parse[Any](expr = "#undiscoveredMap.value", context = customCtx).validValue.returnType shouldBe Unknown
+
+    val parsed = parse[Any]("#undiscoveredMap.value.castToRecord({int1: 999, int2: 999})", customCtx)
```
raphaelsolarski, line 1520: I'm not sure about the approach of providing a sample record as a type description. Let's discuss it.

arkadius: Yep, agree. It looks like we misunderstand each other. I thought that we extend current casting methods by the ability to cast to a record described in our `TypingResult.display`. I mean that examples:
- `castTo("String")`
- `castTo("List[String]")`
- `castTo("{foo: String, bar: Integer}")`
etc. will be possible.

Basically, we don't invent anything new, take a look at how it works in Flink SQL: https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/types/#casting
You can do:
- `CAST(XYZ AS VARCHAR)`
- `CAST(XYZ AS ARRAY<VARCHAR>)`
- `CAST(XYZ AS ROW<foo VARCHAR, bar INTEGER>)`

The have also `TRY_CAST` and `COALESCE(TRY_CAST(), default)`. Sounds familiar? ;)

PR #6971, opened 2024-10-02
[NU-1772] Scenario activities BE - phase 3 - periodic deployment activities

mgoworko:
## Describe your changes

Added last missing Scenario Activity - PerformedScheduledExecution. They are generated only by periodic processes, when executed on scheduled time:
- custom DeploymentManager's are at the moment sort of plugins, that can handle processes in a custom way
- ATM periodic DeploymentManager has separate storage and logic, in general any custom DeploymentManager may 
- so now any DeploymentManager can return activities, that are specific to that manager, and are returned alongside the common ScenarioActivities

How can it be solved better in future (describe in TODO in code):
- it can work in an opposite way - we could provide some hook to the deployment manager, that will allow the deployment manager to save activities in the main repository of Scenario Activities
- it may be necessary for effective pagination
- it would demand migrating old data from external, DeploymentManager-specific data source

## Checklist before merge


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -58,6 +58,29 @@ class PeriodicProcessService(
...
+    scheduledProcessesRepository
+      .getSchedulesState(processIdWithName.name)
+      .run
+      .map(_.groupedByPeriodicProcess)
+      .map(_.flatMap(_.deployments))
+      .map(_.map { deployment =>
+        ScenarioActivity.PerformedScheduledExecution(
+          scenarioId = ScenarioId(processIdWithName.id.value),
+          scenarioActivityId = ScenarioActivityId.random,
+          user = ScenarioUser(None, UserName("Nussknacker"), None, None),
```
mk-software-pl, line 73: shouldn't we have defined `UserName.internalNuUser` (or special user, or sth like that)? And use it in all places we need sth like this?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -58,6 +58,29 @@ class PeriodicProcessService(
...
+          date = deployment.createdAt.toInstant(ZoneOffset.UTC),
+          scenarioVersionId = Some(ScenarioVersionId(deployment.periodicProcess.processVersion.versionId.value)),
+          dateFinished = deployment.state.completedAt.map(_.toInstant(ZoneOffset.UTC)),
+          scheduleName = deployment.scheduleName.display,
+          status = deployment.state.status.toString,
```
mk-software-pl, line 78: why do we have to use "toString"?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
```diff
@@ -112,6 +112,10 @@ trait PeriodicProcessesRepository {
 
   def markInactive(processId: PeriodicProcessId): Action[Unit]
 
+  def getSchedulesState(
+      processName: ProcessName
```
mk-software-pl, line 116: `scenarioName` (we agreed to not use the "Process" word anymore - `ProcessName` will be refactored in the future).

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
mgoworko: Refactored, we now have sealed trait in the ScenarioActivity domain, which represents statuses. (It was toString before, because this value is Scala enumeration)

Other change:
Activity's "date" changed from periodic "createdAt" to "runAt"; I noticed it in tests, the correct date of the activity should be the date of execution, not o scheduling, IMO it was my mistake; added field "createdAt"

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
mgoworko: Done, there is not ScenarioUser.internalNuUser. Is is the only usage ATM, in all other places we have actual, logged-in user (Here it would be possible too - we could use the user that deployed the scenario instead of "Nussknacker" - but we don't have that information available ATM)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Dtos.scala
```diff
@@ -586,6 +586,7 @@ object Dtos {
         scenarioVersionId: Option[Long],
         comment: ScenarioActivityComment,
         dateFinished: Option[Instant],
+        status: Option[String],
```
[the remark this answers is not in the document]
mgoworko: Changed to sealed trait with statuses. Serialized to human-readable string when converting to dto for HTTP API.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -70,6 +70,8 @@ trait DeploymentManager extends AutoCloseable {
 
   def customActionsDefinitions: List[CustomActionDefinition]
 
+  def scenarioActivityHandling: ScenarioActivityHandling
```
mgoworko, line 73: Done based on discussions, but slightly different than discussed, with IMO good justification:

3 cases were analyzed:
- manager does not handle any additional activities
- manager stores its custom activities itself (periodic manager - has its own database)
- manager wants to save and modify activities, stored in the core Nu database (new case, in the future periodics may work that way)

But those 3 cases are about both what manager exposes, and what abilities it wants to have.

From the point of view of what the manager exposes we have only 2 cases:
- all activities handled by Nu, manager does not store any activities
- some manager-specific activities stored and returned by manager

And from point of view of what manager wants to have provided, we have only one case, applicable to all managers - we always want to give the ability to store activities in core Nu repository. We can provide that ability always, it doesn't matter that most managers will not use it.

So:
- when it comes to what the manager exposes, we have `def scenarioActivityHandling` (trait member, because the decorator pattern is used, and self-types are troublesome)
- when it comes to what abilities we provide to the manager, we always provide `ScenarioActivityManager` inside the `DeploymentManagerDependencies`, with abilities to store and edit activities.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivityManager.scala
```diff
@@ -0,0 +1,30 @@
...
+  ): Future[Unit]
+
+  def modifyActivity(
+      scenarioActivityId: ScenarioActivityId,
+      modify: ScenarioActivity => ScenarioActivity,
+  ): Future[Either[String, Unit]]
+
+}
+
+// This stub is in API module because we don't want to extract deployment-manager-tests-utils module
```
mgoworko, line 18: The convention and comment copied from ProcessingTypeActionServiceStub

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentManagerDispatcher.scala
```diff
@@ -27,4 +31,23 @@ class DeploymentManagerDispatcher(
     managers.forProcessingTypeUnsafe(processingType)
   }
 
+  def managerSpecificScenarioActivities(
```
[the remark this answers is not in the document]
mgoworko: Done, logic moved to ScenarioActivityApiHttpService. It should be moved to service layer, which at the moment is missing for scenario activities, when there is refactor.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Dtos.scala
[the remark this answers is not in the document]
mgoworko: It is of different type in the domain - ProcessActionState, not related to the state of periodic deployment. I decided to leave it as primitive type, because ProcessActions are not yet replaced by ScenarioActivities in the Nu code.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivityManager.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def modifyActivity(
+      scenarioActivityId: ScenarioActivityId,
+      modify: ScenarioActivity => ScenarioActivity,
+  ): Future[ScenarioActivityModificationResult]
+
+}
+
+object ScenarioActivityManager {
+  sealed trait ScenarioActivityModificationResult
```
mk-software-pl, line 21: `ModificationResult` is enough ;) (the know that the context is scenario activity due to the fact the ADT is placed in the ScenarioActivityManager companion object)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
```diff
@@ -210,15 +216,25 @@ class ScenarioActivityApiHttpService(
...
+            case AllScenarioActivitiesStoredByNussknacker =>
+              Future.successful(List.empty)
+            case handling: ManagerSpecificScenarioActivitiesStoredByManager =>
+              handling.managerSpecificScenarioActivities(processIdWithName)
+          }
+        case None =>
+          Future.successful(List.empty)
+      }
+    } yield generalActivities ++ deploymentManagerSpecificActivities
+    EitherT.right(combined).map(_.map(toDto).toList.sortBy(_.date))
```
mk-software-pl, line 236: maybe we should move the DTO creation and sorting to the yield?
at the moment it's rather hidden in the monad transformer boilerplate

mk-software-pl (approved):
two small comments.
in general: LGTM!

PR #6995, opened 2024-10-08
[NU-1821][NU-1822] BASE64 helper and from/to JSON conversions

mk-software-pl:
## Describe your changes

* new `#BASE64` helper (encoding + decoding)
* new conversions (`#CONV` helper): `toJson` and `toJsonString`

## Checklist before merge


docs/Changelog.md
```diff
@@ -43,9 +43,11 @@
   * toBigIntegerOrNull
   * toBigDecimal
   * toBigDecimalOrNull
+* [#6995](https://github.com/TouK/nussknacker/pull/6995) Add `toJson` and `toJsonString` conversions
```
lciolecki, line 46: Where these conversions are available? :) I mean from which helper I can rich it?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
```diff
@@ -138,6 +139,30 @@ trait ConversionUtils extends HideToString {
     case _                       => null
   }
 
+  @Documentation(description = "Convert String value to JSON")
+  def toJson(@ParamName("value") value: String): Any = {
+    toJsonEither(value).toTry.get
```
lciolecki, line 144: I don't like that we return here Circe.. It can be a problematic type for Flink serialization / deserialization :/ On ohter hand, what we can return here? Maybe we can try convert Json to java types?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/conversion.scala
```diff
@@ -138,6 +139,30 @@ trait ConversionUtils extends HideToString {
...
+
+  @Documentation(description = "Convert JSON to String")
+  def toJsonString(@ParamName("value") value: Any): String = {
+    jsonEncoder.encode(value).noSpaces
+  }
+
+  private def toJsonEither(value: String): Either[Throwable, Any] = {
+    io.circe.parser.parse(value) match {
+      case Right(json) => Right(JsonUtils.jsonToAny(json))
+      case Left(ex)    => Left(new IllegalArgumentException(s"Cannot convert [$value] to JSON", ex))
```
lciolecki, line 160: Hmm, I'm not sure, but here we should throw Transient exception?

raphaelsolarski: AFAIK by default all exceptions except `ConnectException` are treaded as NonTransient exceptions which are not propagated to flink from interpreter and don't cause scenario restart.
The difference is that exceptions wrapped inside `NonTransientException` are not logged as unknown exceptions in `WithExceptionExtractor.extractOrThrow`.

We actually don't mark errors thrown from utils as NonTransientExceptions so IMO marking only this one doesn't make sense.

raphaelsolarski (approved):

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/DefaultConfigCreator.scala
```diff
@@ -18,6 +18,7 @@ class DefaultConfigCreator extends EmptyProcessConfigCreator {
         "DATE_FORMAT" -> anyCategory(dateFormat),
         "UTIL"        -> anyCategory(util),
         "RANDOM"      -> anyCategory(random),
+        "BASE64"      -> anyCategory(base64)
```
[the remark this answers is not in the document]
mk-software-pl: IMO it can be treated as conversion, but it's not clear at first glance. That's why I created a new, separate helper.

PR #7010, opened 2024-10-11
NU-1827 allow dynamic access via indexer operator on expressions typed to unknown

raphaelsolarski:
## Describe your changes
For now we want:
* enable by default dynamic access using indexer operator for expressions typed to Unknown
* computed type for such an access is always Unknown (or concrete type if property is known)

## Todo
- [x] Disable runtime member access checking in property accessors if dynamicPropertyAccessAllowed=True
- [x] Secure ReflectivePropertyAccessor 
- [ ] Tests for typical use cases (e.g. with untyped json based source or json parsed from util)
- [x] Documentation update
- [ ] Optional: memoization of computed method definition in ClassDefinitionSetChecking trait

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionDiscovery.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionDiscovery.scala
@@ -32,6 +32,7 @@ class ClassDefinitionDiscovery(classDefinitionExtractor: ClassDefinitionExtracto
   private val mandatoryClasses = (Set(
     classOf[java.util.List[_]],
     classOf[java.util.Map[_, _]],
+    classOf[java.util.Map.Entry[_, _]],
     classOf[java.math.BigDecimal],
     classOf[Number],
     classOf[String],
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
@@ -1,5 +1,7 @@
 package pl.touk.nussknacker.engine.definition.clazz
 
+import com.github.ghik.silencer.silent
+
 object ClassDefinitionSet {
 
   def apply(classDefinitions: Set[ClassDefinition]): ClassDefinitionSet = {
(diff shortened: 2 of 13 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionDiscovery.scala
```diff
@@ -32,6 +32,7 @@ class ClassDefinitionDiscovery(classDefinitionExtractor: ClassDefinitionExtracto
   private val mandatoryClasses = (Set(
     classOf[java.util.List[_]],
     classOf[java.util.Map[_, _]],
+    classOf[java.util.Map.Entry[_, _]],
```
arkadius, line 35: Interesting that we had examples for that, but it wasn't possible to use it :)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -228,6 +228,8 @@ private[spel] class Typer(
         case TypedNull =>
           invalidNodeResult(IllegalIndexingOperation)
         case TypedObjectWithValue(underlying, _) => typeIndexer(e, underlying)
+        case Unknown =>
```
arkadius, line 231: Can we remove the default case below? What case won't be exhausted?

arkadius (commented):
Looks good, some minor comments added. Please remember about changelog

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/propertyAccessors.scala
```diff
@@ -1,37 +1,95 @@
...
       TypedDictInstancePropertyAccessor, // must be before NoParamMethodPropertyAccessor
-      NoParamMethodPropertyAccessor,
+      new NoParamMethodPropertyAccessor(classDefinitionSet, checkAccess),
       // it can add performance overhead so it will be better to keep it on the bottom
       MapLikePropertyAccessor,
       MapMissingPropertyToNullAccessor, // must be after NoParamMethodPropertyAccessor
     )
   }
 
+  class CheckedReflectivePropertyAccessor(
```
arkadius, line 43: It is just a theoretical question, I want to know the limitation of this method checking. Let's assume that we have some variable that could have multiple types of values. Am I right that once when canRead pass method call, we won't check if method call is allowed for other types?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionDiscovery.scala
raphaelsolarski: afaik it was possible because it was artificially handled in `Typer.extractIterativeType`, but started to fail tests when I added runtime checking in property accessors.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
raphaelsolarski: The default case catches non-map and non-list TypedClass types for which we want still apply previous behaviour with `dynamicPropertyAccessAllowed` feature flag.

As we discussed - in this step we add dynamic access to unknown fields only on nodes typed to `Unknown` and only via indexer operator.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
```diff
@@ -9,11 +9,40 @@ object ClassDefinitionSet {
...
+  }
 
   def all: Set[ClassDefinition] = classDefinitionsMap.values.toSet
 
-  def get(clazz: Class[_]): Option[ClassDefinition] =
-    classDefinitionsMap.get(clazz)
+  def get(clazz: Class[_]): Option[ClassDefinition] = classDefinitionsMap.get(clazz)
+
+  def isParameterlessMethodAllowed(targetClass: Class[_], methodName: String, staticMethodsOnly: Boolean): Boolean = {
+    extractAllTypesInHierarchy(targetClass).exists(typeToCheck =>
```
arkadius, line 31: Let's move this all types hierarchy traversal to for-comprehension above so we won't do even this step in runtime

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
```diff
@@ -16,4 +16,35 @@ case class ClassDefinitionSet(classDefinitionsMap: Map[Class[_], ClassDefinition
   def get(clazz: Class[_]): Option[ClassDefinition] =
     classDefinitionsMap.get(clazz)
 
+  def isParameterlessMethodAllowed(targetClass: Class[_], method: String, onlyStaticMethods: Boolean): Boolean = {
+    // todo: some memoization
+    val classWithParentClasses = extractAllTypes(targetClass)
+    classWithParentClasses.exists(hasDirectlyDefinedParameterlessMethod(_, method, onlyStaticMethods))
```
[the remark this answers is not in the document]
arkadius: Nice, one related comment added

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
arkadius: So let's replace it with explicit pattern to make visible what is handled in this case. `case otherClass: TypedClass => ` will exhaust this pattern matching?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/propertyAccessors.scala
```diff
@@ -284,4 +356,68 @@ object propertyAccessors {
...
+
+    override protected def methodAllowed(
+        targetClass: Class[_],
+        methodName: String,
+        onlyStaticMethods: Boolean
+    ): Boolean = true
+
+  }
+
+  final case class AllowedOnlyDefinedParameterlessMethods(classDefinitionSet: ClassDefinitionSet)
```
arkadius, line 410: I think that more grammatically correct will be `OnlyDefinedParameterlessMethodsAllowed`. WDYT?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
raphaelsolarski: fixed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
arkadius: 1. Let's add a comment
2. (Optionally) Let's replace lists in `extractAllTypesInHierarchy` with `Steam` to reduce unnecessary traversal

arkadius (approved):

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/clazz/ClassDefinitionSet.scala
raphaelsolarski: fixed

arkadius (approved):

PR #7014, opened 2024-10-14
Upgrade tapir / openapiCirceYaml versions

philemone:
## Checklist before merge

### DIFF
--- a/designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
+++ b/designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
@@ -10,7 +10,7 @@ import pl.touk.nussknacker.ui.security.api.SecurityError.{
   ImpersonationNotSupportedError,
   InsufficientPermission
 }
-import sttp.model.StatusCode.{Forbidden, NotFound, NotImplemented, Unauthorized}
+import sttp.model.StatusCode.{Forbidden, NotImplemented, Unauthorized}
 import sttp.tapir.EndpointIO.Example
 import sttp.tapir._
 
@@ -29,7 +29,6 @@ trait BaseEndpointDefinitions {
 
 object BaseEndpointDefinitions {
 
-  type EndpointError[ERROR] = Either[SecurityError, ERROR]
   type SecuredEndpoint[INPUT, BUSINESS_ERROR, OUTPUT, -R] =
     Endpoint[AuthCredentials, INPUT, Either[BUSINESS_ERROR, SecurityError], OUTPUT, R]
 
@@ -68,36 +67,19 @@ object BaseEndpointDefinitions {
                 )
             ),
             oneOfVariantFromMatchType(
-              NotFound,
-              plainBody[ImpersonatedUserDataNotFoundError.type]
-                .description("Identity provided in the Nu-Impersonate-User-Identity header did not match any user")
-                .example(
-                  Example.of(
-                    summary = Some("No impersonated user's data found for provided identity"),
-                    value = ImpersonatedUserDataNotFoundError
-                  )
-                )
-            ),
-            oneOfVariantFromMatchType(
               Forbidden,
-              plainBody[ImpersonationMissingPermissionError.type]
-                .example(
-                  Example.of(
-                    summary = Some("Authorization failed, user does not have permission to impersonate"),
-                    value = ImpersonationMissingPermissionError
+              oneOfBody[SecurityError](
+                plainBody[SecurityError]
+                  .examples(
+                    List(
+                      Example.of(name = Some("InsufficientPermission"), value = InsufficientPermission),
+                      Example
+                        .of(name = Some("ImpersonationMissingPermission"), value = ImpersonationMissingPermissionError),
+                      Example.of(name = Some("ImpersonatedUserDataNotFound"), value = ImpersonatedUserDataNotFoundError)
+                    )
                   )
-                )
+              )
             ),
-            oneOfVariantFromMatchType(
-              Forbidden,
-              plainBody[InsufficientPermission.type]
-                .example(
-                  Example.of(
-                    summary = Some("Authorization failed"),
-                    value = InsufficientPermission
-                  )
-                )
-            )
           )
         )
     }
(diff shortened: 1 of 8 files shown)


designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
```diff
@@ -67,37 +66,20 @@ object BaseEndpointDefinitions {
                   )
                 )
             ),
-            oneOfVariantFromMatchType(
-              NotFound,
-              plainBody[ImpersonatedUserDataNotFoundError.type]
```
philemone, line 72: I've changed 404 to 403 for `ImpersonatedUserDataNotFoundError`. 
1. `withSecurity` is added to different endpoint definitions which also can have 404 code declared - in such a case we override some fields and we also somehow breaking openApi spec (two times 404 code defined for one endpoint) 
2. Personally I think that `withSecurity` along with unsuccessful impersonation is not 404 but 403. Should `withSecurity ever return 404? :thinking:

philemone: Now I think I could change this error to `ImpersonatedUserDataNotExists` - WDYT?

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
```diff
@@ -106,36 +88,16 @@ object BaseEndpointDefinitions {
...
-    }
-
-    implicit val impersonationPermissionErrorCodec
-        : Codec[String, ImpersonationMissingPermissionError.type, CodecFormat.TextPlain] = {
+    implicit val authorizationErrorCodec: Codec[String, SecurityError, CodecFormat.TextPlain] = {
       Codec.string.map(
-        Mapping.from[String, ImpersonationMissingPermissionError.type](_ => ImpersonationMissingPermissionError)(_ =>
-          ImpersonationMissingPermissionError.errorMessage
-        )
+        Mapping.from[String, SecurityError](_ => InsufficientPermission)(s => s.errorMessage)
```
coutoPL, line 93: `_ => InsufficientPermission` is suspicious here

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
```diff
@@ -67,37 +66,20 @@ object BaseEndpointDefinitions {
...
-                )
-            ),
             oneOfVariantFromMatchType(
               Forbidden,
-              plainBody[ImpersonationMissingPermissionError.type]
-                .example(
-                  Example.of(
-                    summary = Some("Authorization failed, user does not have permission to impersonate"),
-                    value = ImpersonationMissingPermissionError
+              oneOfBody[SecurityError](
```
coutoPL, line 71: Shouldn't we have `AuthorizationError` here instead?

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
coutoPL: the current implementation makes sense for me. the `ImpersonatedUserDataNotFoundError` means that the user was able to authenticate, but the impersonated user was not found. There was no authorization issue. 

But if it simplifies the tapir things, we can return 403 instead.

According to the name of the error: I would remove the "data" word. So `ImpersonatedUserNotExistsError` would be great IMO ;)

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
philemone: If we would have here only `InsufficientPermission` and `ImpersonationMissingPermissionError` they yes - but  we also have `ImpersonatedUserDataNotFoundError` in here

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
philemone: Ah, forgot about it, yeah, it is shady as hell

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
```diff
@@ -1,18 +1,26 @@
 package pl.touk.nussknacker.ui.security.api
 
-sealed trait SecurityError
+sealed trait SecurityError {
+  val errorMessage: String
```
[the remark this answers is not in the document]
philemone: That looks cool but If I am not mistaken it won't force a developer to provide string representation (errorMessage in this case)? There is a possibility that someone in the future will create `NewError extends SecurityError` without adding valid string representation to `Show[SecurityError]`.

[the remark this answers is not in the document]
coutoPL: hmm, in this case, `Show[SecurityError]` should stop compiling (pattern matching exhaustivity check will detect that - at least when warns are treated as errors /in our case they are/)

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
coutoPL: ok, but can we make "ImpersonatedUserDataNotFoundError" an AuthorizationError? You proposed [here](https://github.com/TouK/nussknacker/pull/7014#discussion_r1802585801) to interpret it as 403

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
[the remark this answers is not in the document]
philemone: Ah, ok - so last thing I would be concerned about: Do we prefer to separate error message from case class/object definition to somewhere else and provide it through some import and use implicit method (with pattern matching inside) to provide `.show` method rather than just call variable from that class? 

Don't get me wrong - I love cats and I don't mind using implicits but I am wondering where is it beneficial and where is not :D

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
philemone: Yes we could, then we could change that - good point.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
coutoPL: Yeah, 404 is not the best choice - it's correct. But 403 can be questioned too. As I said, the user can be authorized to impersonate users but simply there is no such user. 

But it's a common problem of RESTfull services. HTTP codes are too narrow and operations are not resources (and pretending they are doesn't make things easier). 

That's why I said: let's do this if it makes our lives a little bit simpler.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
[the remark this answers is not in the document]
coutoPL: TBH we don't have to use Show. We could have `def toString(error: SecurityError): String` method. This is just a single responsibility principle usage. In the core, you deal with some SecurityErrors. You declare them as ADT. Then at the end of the system (when you want to provide a RESTful interface) you need to change them to status codes and messages. In this the place where you say: this error maps to 403 and this one to 401. And this is the message related to the error.

So, this is not about Cats. It's rather about clean code ;)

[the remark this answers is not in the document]
philemone: That is a good point and well explained :+1:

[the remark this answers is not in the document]
philemone: I've used extension method to separate interpretations of errors from their definitions.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/BaseEndpointDefinitions.scala
philemone: Done + moved `ImpersonatedUserNotExistsError` to `401` section, since based on created model it should be there - it is part of AuthenticationError rather than AuthorizationError.

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
```diff
@@ -6,20 +6,27 @@ sealed trait AuthenticationError              extends SecurityError
...
+  case object ImpersonatedUserNotExistsError extends ImpersonationAuthenticationError
 
   case object CannotAuthenticateUser extends AuthenticationError
 
-  case object ImpersonatedUserDataNotFoundError extends ImpersonationAuthenticationError {
-    val errorMessage = "No impersonated user data found for provided identity"
-  }
+  case object ImpersonationNotSupportedError extends ImpersonationAuthenticationError
+
+  implicit class SecurityErrorWithMessage(s: SecurityError) {
```
coutoPL, line 20: let's add `extends AnyVal`

coutoPL (approved):
one minor comment.
in general LGTM

security/src/main/scala/pl/touk/nussknacker/ui/security/api/SecurityError.scala
philemone: Ah, sure -  thanks

PR #7021, opened 2024-10-15
[NU-1847] Definition service with the ability to return definition without config from additional provider

mateuszkp96:
## Describe your changes
Definition service with the ability to return definition without config from additional provider
## Checklist before merge


designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/definition/package.scala
```diff
@@ -48,6 +48,7 @@ package object definition {
       // FE need this information because branch parameters aren't changed dynamically during node validation so they never
       // should be invalidated
       branchParam: Boolean,
+      requiredParam: Boolean,
```
arkadius, line 51: It isn't used by the fronted, is it? We should avoid adding to API parameters that are not used by the FE. From a developer of this project perspective, this would be just a dead code. Dead code == to be removed. 
Solutions that I have in mind:
- Can you base on validators.exists(mandatory) rule instead on your side?
- Another API with described responsibility

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/DefinitionResources.scala
```diff
@@ -37,4 +38,20 @@ class DefinitionResources(
     }
   }
 
+  private val isFragmentParam: Directive1[Boolean] = parameter(Symbol("isFragment").as[Boolean])
+
+  private val modelParametersModeParam: Directive1[ModelParametersMode] = {
```
arkadius, line 43: 1. This `RAW` mode is a kind of quirk for external project purpose. I would document it - see comments containing "external project" and try to have it in as fewer places as possible. We can make FE don't know about the existence of this parameter at all. To achieve that, in the API we could replace it with some boolean that on the BE side will be immediately replaced by this enum.

2. Also, I would change the naming a bit - RAW/ENRICHED sounds too generic for me. We need to use `UiConfig`/`ComponentUiConfig` in the name as it is used everywhere where we touch this "customization" feature. Also, the variant that you've called "RAW" isn't raw. By "RAW" I understand something not for final usage, sth that after some additional processing will be usable and in our usages, in most cases no AdditionalUIConfig will be applied so basically both modes will return the same value.

BTW if you need some inspiration, here is the documentation: https://nussknacker.io/documentation/docs/configuration/model/ModelConfiguration/#configuration-of-ui-attributes-of-components

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentDefinitionExtractor.scala
```diff
@@ -73,35 +73,60 @@ object ComponentDefinitionExtractor {
...
+          finalEnrichedConfig,
+          additionalConfigs.groupName
+        )
+        rawComponentConfig <- filterOutDisabledAndComputeFinalUiDefinition(
+          finalRawConfig,
+          additionalConfigs.groupName
+        )
+      } yield toComponentDefinition(
+        ResolvedComponentConfig(
+          uiDefinition = enrichedComponentConfig._1,
```
arkadius, line 125: 1. Let's use name variables in for comprehension instead of _1 and _2
2. What about `uiDefinition`? I think that you'll need to show things like `icon` or `docsUrl` before the additionalUiConfig application as well on your side. Even if you don't need this at this moment, IMO we should provide them for sake of consistency.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentDefinitionExtractor.scala
```diff
@@ -73,35 +73,60 @@ object ComponentDefinitionExtractor {
...
-      val configFromAdditional                    = additionalConfigs.getConfig(componentId)
-      val combinedConfigWithoutConfigFromProvider = configFromAdditional |+| configFromDefinition |+| defaultConfig
-      val designerWideId                          = combinedConfigWithoutConfigFromProvider.componentId
-      val finalCombinedConfig = additionalConfigFromProvider(designerWideId) |+| combinedConfigWithoutConfigFromProvider
 
-      filterOutDisabledAndComputeFinalUiDefinition(finalCombinedConfig, additionalConfigs.groupName).map(f.tupled)
+      val finalEnrichedConfig = configFor(defaultConfig, withConfigFromProvider = true)
+      val finalRawConfig      = configFor(defaultConfig, withConfigFromProvider = false)
+
+      for {
```
arkadius, line 114: I think that sth is not ok in this for comprehension. You would return Some if component is not disabled in raw configuration and in enriched configuration? Why do you think that this conjunction is ok? Why not OR or maybe XOR? :) I think that we should build a different set of component definitions at all.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/definition/package.scala
mateuszkp96: I added comment that this param is used by external project

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentDefinitionExtractor.scala
mateuszkp96: I moved components extraction to a higher level, please check `pl.touk.nussknacker.engine.definition.component.Components`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentsFromProvidersExtractor.scala
```diff
@@ -26,6 +27,107 @@ object ComponentsFromProvidersExtractor {
...
+      componentDefinitionExtractionMode: ComponentDefinitionExtractionMode
+  ): Components = {
+
+    def extractComponentDefinition(configFromProvider: Map[DesignerWideComponentId, ComponentAdditionalConfig]) = {
+      ComponentDefinitionExtractor
+        .extract(
+          componentName,
+          component,
+          configFromDefinition,
+          componentsUiConfig,
```
arkadius, line 75: I thought that we agreed that for `basicComponents` we also skip `componentsUiConfig`. It is easier to think about component definitions if both `componentsUiConfig` and `configFromProvider` are treated the same.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/DesignerModelData.scala
```diff
@@ -23,7 +23,14 @@ object DesignerModelData {
   final case class DynamicComponentsStaticDefinitions(
       finalDefinitions: Map[ComponentId, ComponentStaticDefinition],
       // components definitions without enrichments from an additional provider
-      basicDefinitions: Map[ComponentId, ComponentStaticDefinition]
-  )
+      basicDefinitions: Option[Map[ComponentId, ComponentStaticDefinition]]
```
arkadius, line 26: Let's make it private

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentsFromProvidersExtractor.scala
```diff
@@ -30,30 +30,49 @@ object ComponentsFromProvidersExtractor {
 final case class Components(
     components: List[ComponentDefinitionWithImplementation],
     // components without enrichments from an additional provider
-    basicComponents: List[ComponentDefinitionWithImplementation]
+    basicComponents: Option[List[ComponentDefinitionWithImplementation]]
```
arkadius, line 33: the same here

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentsFromProvidersExtractor.scala
arkadius: Ok, but now I think that we should also rename `enrichedWithAdditionalConfig` to `enrichedWithUiConfig` and change comments such as `components without enrichments from an additional provider`.

arkadius (approved):

PR #7036, opened 2024-10-17
[NU-1772] Scenario activities - fragment, comment and attachment improvements

mgoworko:
## Describe your changes

Changes and fixes requested after preview tests:

1. Comments:
- empty-comment as no-comment
- status NOT_AVAILABLE instead of DELETED
- comment that have always had empty content can be edited

2. Attachments
- added missing endpoint for deleting
- name override: filename when file present, "File removed" when deleted

3. Scenario fragments:
- displayableName is adjusted for fragments in metadata endpoint

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
```diff
@@ -247,20 +268,17 @@ class ScenarioActivityApiHttpService(
...
-          lastModifiedBy = lastModifiedByUserName.value,
-          lastModifiedAt = lastModifiedAt,
-        )
-      case ScenarioComment.Deleted(deletedByUserName, deletedAt) =>
-        Dtos.ScenarioActivityComment(
-          content = Dtos.ScenarioActivityCommentContent.Deleted,
-          lastModifiedBy = deletedByUserName.value,
-          lastModifiedAt = deletedAt,
-        )
+    val content = scenarioComment match {
```
mk-software-pl, line 271: I'm not sure if this is the correct place to do so. 

I assume that we allowed (for some reason) activity to be saved with AvailableComment with an empty string. 
It means that we should:
1. forbid saving it (and probably introduce NonEmptyString instead of String for the field that holds the comment content)
2. write a migration

Or we could:
1. forbid saving it (or save it with no comment option)
2. encapsulate the fact that in the DB we could have available comments that are empty, and convert them to non-available comments in the repo.

Handling it at the service level it too late IMO.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Dtos.scala
```diff
@@ -107,105 +116,105 @@ object Dtos {
     private val commentRelatedActions = List("delete_comment", "edit_comment")
 
     case object ScenarioCreated extends ScenarioActivityType {
-      override def displayableName: String        = "Scenario created"
-      override def icon: String                   = "/assets/activities/scenarioModified.svg"
-      override def supportedActions: List[String] = List.empty
+      override def displayableName(scenarioType: ScenarioType): String = s"${scenarioType.entryName} created"
```
mk-software-pl, line 119: It looks very weird to me. 
We have an activity type that produces a displayable name? And it needs scenatioType to do that?
IMO the sth that produces a displayable name needs a Scenario activity (and maybe its type), but not the other way around.

Is it a way to improve this part of code?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Endpoints.scala
```diff
@@ -173,6 +179,23 @@ class Endpoints(auth: EndpointInput[AuthCredentials], streamProvider: TapirStrea
       .withSecurity(auth)
   }
 
+  val deleteAttachmentEndpoint: SecuredEndpoint[DeleteAttachmentRequest, ScenarioActivityError, Unit, Any] = {
+    baseNuApiEndpoint
+      .summary("Delete scenario attachment service")
```
mk-software-pl, line 184: Do we need the word "service" here?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Endpoints.scala
```diff
@@ -173,6 +179,23 @@ class Endpoints(auth: EndpointInput[AuthCredentials], streamProvider: TapirStrea
...
+      .tag("Activities")
+      .delete
+      .in("processes" / path[ProcessName]("scenarioName") / "activity" / "attachments" / path[Long]("attachmentId"))
+      .mapInTo[DeleteAttachmentRequest]
+      .out(statusCode(Ok))
+      .errorOut(
+        oneOf[ScenarioActivityError](
+          oneOfVariant(NotFound, plainBody[NoScenario].example(Examples.noScenarioError)),
+          oneOfVariant(InternalServerError, plainBody[NoAttachment].example(Examples.attachmentNotFoundError)),
```
mk-software-pl, line 193: why 5xx?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
```diff
@@ -90,7 +91,9 @@ object ProcessRepository {
       sourceEnvironment: String,
       targetEnvironment: String,
       sourceScenarioVersionId: Option[VersionId],
-  ) extends ModifyProcessAction
+  ) extends ModifyProcessAction {
+    override val comment: Option[Comment] = None
```
mk-software-pl, line 95: doesn't it mean that the `comment: Comment` field should be defined in one case and not be defined at the sealed trait level?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/ScenarioActivityRepository.scala
```diff
@@ -86,6 +92,14 @@ object ScenarioActivityRepository {
     case object CouldNotModifyComment extends ModifyCommentError
   }
 
+  sealed trait DeleteAttachmentError
+
+  object DeleteAttachmentError {
+    case object ActivityDoesNotExist     extends DeleteAttachmentError
+    case object AttachmentDoesNotExist   extends DeleteAttachmentError
```
mk-software-pl, line 99: not used?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
```diff
@@ -558,6 +576,15 @@ class ScenarioActivityApiHttpService(
...
   }
 
+  private def markAttachmentAsDeleted(request: DeleteAttachmentRequest, scenarioId: ProcessId)(
+      implicit loggedUser: LoggedUser
+  ): EitherT[Future, ScenarioActivityError, Unit] =
+    EitherT(
+      dbioActionRunner.run(
+        scenarioActivityRepository.markAttachmentAsDeleted(scenarioId, request.attachmentId)
+      )
+    ).leftMap(_ => NoAttachment(request.attachmentId))
```
mk-software-pl, line 586: all errors are mapped to NoAttachement? ScenarioNotFound too?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpServiceBusinessSpec.scala
```diff
@@ -229,6 +228,68 @@ class ScenarioActivityApiHttpServiceBusinessSpec
...
+        .statusCode(200)
+
+      val attachmentId = verifyAttachmentAddedActivityExists(
+        user = "allpermuser",
+        scenarioName = exampleScenarioName,
+        fileIdPresent = true,
+        filename = fileName,
+        fileStatus = "AVAILABLE",
+        overrideDisplayableName = fileName
+      ).extractString("activities[1].attachment.file.id")
```
mk-software-pl, line 251: maybe `verifyAttachmentAddedActivityExists` could return a wrapper around the currently returned `ValidatableResponse` and encapsulate the string extraction under some method?

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivity.scala
```diff
@@ -47,9 +50,9 @@ object ScenarioComment {
       lastModifiedAt: Instant,
   ) extends ScenarioComment
 
-  final case class Deleted(
-      deletedByUserName: UserName,
-      deletedAt: Instant,
+  final case class NotAvailable(
```
mk-software-pl, line 53: I'm not sure if it's easy to grasp when is not available comment? :D Especially when it has user and modification :D

mk-software-pl (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mgoworko: Added handling on repository level, both on save and on read.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Endpoints.scala
mgoworko: removed suffix "service" from all summaries in this API

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mgoworko: Yes, that is actually the only error (I removed 2 other unused variants of DeleteAttachmentError). It is executed as a single db update, so we only get a count of updated rows.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Endpoints.scala
mgoworko: That is the way BE communicates with FE in this API. It worked that way before any ScenarioActivity-related changes. Any change would involve all endpoints and would have to be coordinated with FE AFAIK.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/ProcessRepository.scala
mgoworko: done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/activities/ScenarioActivityRepository.scala
mgoworko: yes, removed, only CouldNotDeleteAttachment is used

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivity.scala
mgoworko: Changed, ScenarioComment with 2 variants: WithContent, WithoutContent

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpServiceBusinessSpec.scala
mgoworko: Name changed to `verifyAttachmentAddedActivityExistsAndReturnFileId`, with return type Option[Long]

common-api/src/main/scala/pl/touk/nussknacker/engine/api/Comment.scala
```diff
@@ -6,6 +6,13 @@ final case class Comment(content: String) extends AnyVal {
   override def toString: String = content
 }
 
+final case class NonEmptyComment private (content: String)
```
mk-software-pl, line 9: WDYT about using Refined lib and NonEmptyString as content in the `Comment` case class.
I don't think I understand why we need to have Comment and NonEmptyComment? Do we need to distinguish these cases in business logic? Or an empty content is just no comment at all? (so it could be expressed as Option[Comment] obviously)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mk-software-pl: is the second case correct?
```
      case ScenarioComment.WithContent(comment, _, _) =>
        Dtos.ScenarioActivityCommentContent.Available(comment)
      case ScenarioComment.WithoutContent(_, _) | ScenarioComment.WithContent(_, _, _) =>
        Dtos.ScenarioActivityCommentContent.NotAvailable
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
```diff
@@ -247,20 +268,17 @@ class ScenarioActivityApiHttpService(
...
-          lastModifiedAt = deletedAt,
-        )
+    val content = scenarioComment match {
+      case ScenarioComment.WithContent(comment, _, _) =>
+        Dtos.ScenarioActivityCommentContent.Available(comment)
+      case ScenarioComment.WithoutContent(_, _) | ScenarioComment.WithContent(_, _, _) =>
+        Dtos.ScenarioActivityCommentContent.NotAvailable
     }
+    Dtos.ScenarioActivityComment(
+      content = content,
```
mk-software-pl, line 278: let's inline it

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mk-software-pl: let's handle it explicitly. It'd be easier to catch changes in the future in the method declaration

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioActivity/Dtos.scala
mgoworko: Refactored a bit:
- separate displayableNameForScenario and displayableNameForFragment are explicitly defined
- the correct one is chose, when the service creates metadata for scenario/fragment

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpServiceBusinessSpec.scala
mk-software-pl: IMO it'd be better to leave the name as it was before and return the wrapper I mentioned above. When users (of the method) are interested in extracting sth they can do it, but when they are not interested in it, they won't be bothered. 

Moreover, we could easily add other methods to the wrapper that would extract totally different things from the response.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mgoworko: fixed

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mgoworko: done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
mgoworko: done

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpServiceBusinessSpec.scala
mgoworko: done, restored name, created very simple wrapper for extracting fields from activities

common-api/src/main/scala/pl/touk/nussknacker/engine/api/Comment.scala
mgoworko: For now I did it without refined (it is not present in Nu right now, definitely should be added in future). I removed NonEmptyComment and Comment now has private constructor and create method which validates string value.

One usage in Dto class was problematic, replaced with raw string in dto.

designer/server/src/test/scala/pl/touk/nussknacker/test/processes/WithScenarioActivitySpecAsserts.scala
```diff
@@ -138,3 +203,12 @@ trait WithScenarioActivitySpecAsserts
   }
 
 }
+
+object WithScenarioActivitySpecAsserts extends WithBusinessCaseRestAssuredUsersExtensions {
+
+  final case class ScenarioActivitiesResponseWrapper(validatableResponse: ValidatableResponse) {
+    def extractString(activityIndex: Int, path: String): String =
```
mk-software-pl, line 210: it's ok, but IMO it should be more specialized. Like `getAddedAttachementId: String`.

It's because the functionality is not so "symmetric" currently. We verify that the attachment was added to the scenario without other activities (see the body matcher). And then we allow getting any activity, but there are only two of them :P

So, let's make it more "symmetric"

mk-software-pl (approved):
LGTM.

One minor comment left

PR #7066, opened 2024-10-23
[NU-1823] Kafka source without topic schema

ForrestFairy:
## Describe your changes
We want to enable user to choose and use topics without schema present on schema registry.

## Checklist before merge

### DIFF
--- a/utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
+++ b/utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
@@ -1,7 +1,8 @@
 package pl.touk.nussknacker.engine.schemedkafka
 
 import cats.data.Validated.{Invalid, Valid}
-import cats.data.Writer
+import cats.data.{Validated, Writer}
+import org.apache.kafka.clients.admin.ListTopicsOptions
 import pl.touk.nussknacker.engine.api.component.Component
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError.CustomNodeError
@@ -13,11 +14,19 @@ import pl.touk.nussknacker.engine.api.process.{ProcessObjectDependencies, TopicN
 import pl.touk.nussknacker.engine.api.validation.ValidationMode
 import pl.touk.nussknacker.engine.api.{NodeId, Params}
 import pl.touk.nussknacker.engine.kafka.validator.WithCachedTopicsExistenceValidator
-import pl.touk.nussknacker.engine.kafka.{KafkaComponentsUtils, KafkaConfig, PreparedKafkaTopic, UnspecializedTopicName}
+import pl.touk.nussknacker.engine.kafka.{
+  KafkaComponentsUtils,
+  KafkaConfig,
+  KafkaUtils,
+  PreparedKafkaTopic,
+  UnspecializedTopicName
+}
 import pl.touk.nussknacker.engine.schemedkafka.schemaregistry._
 import pl.touk.nussknacker.engine.schemedkafka.schemaregistry.universal.UniversalSchemaSupportDispatcher
 import pl.touk.nussknacker.engine.kafka.UnspecializedTopicName._
 import pl.touk.nussknacker.engine.kafka.validator.TopicsExistenceValidator.TopicValidationType
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
 
 object KafkaUniversalComponentTransformer {
   final val schemaVersionParamName      = ParameterName("Schema version")
@@ -62,7 +71,10 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
   protected def getTopicParam(
       implicit nodeId: NodeId
   ): WithError[ParameterCreatorWithNoDependency with ParameterExtractor[String]] = {
-    val topics = topicSelectionStrategy.getTopics(schemaRegistryClient)
+    val allTopics = getAllTopics
+
+    val topics: Validated[SchemaRegistryError, List[UnspecializedTopicName]] = Valid(allTopics)
+    // topicSelectionStrategy.getTopics(schemaRegistryClient)
 
     (topics match {
       case Valid(topics) => Writer[List[ProcessCompilationError], List[UnspecializedTopicName]](Nil, topics)
@@ -98,15 +110,31 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
   protected def getVersionParam(
       preparedTopic: PreparedKafkaTopic[TN],
   )(implicit nodeId: NodeId): WithError[ParameterCreatorWithNoDependency with ParameterExtractor[String]] = {
-    val versions = schemaRegistryClient.getAllVersions(preparedTopic.prepared.toUnspecialized, isKey = false)
-    (versions match {
-      case Valid(versions) => Writer[List[ProcessCompilationError], List[Integer]](Nil, versions)
-      case Invalid(e) =>
-        Writer[List[ProcessCompilationError], List[Integer]](
-          List(CustomNodeError(e.getMessage, Some(topicParamName))),
-          Nil
-        )
-    }).map(getVersionParam)
+    val topicsWithSchema = topicSelectionStrategy.getTopics(schemaRegistryClient)
+    if (topicsWithSchema.exists(topics => topics.contains(preparedTopic.prepared.topicName.toUnspecialized))) {
+      val versions = schemaRegistryClient.getAllVersions(preparedTopic.prepared.toUnspecialized, isKey = false)
+      (versions match {
+        case Valid(versions) => Writer[List[ProcessCompilationError], List[Integer]](Nil, versions)
+        case Invalid(e) =>
+          Writer[List[ProcessCompilationError], List[Integer]](
+            List(CustomNodeError(e.getMessage, Some(topicParamName))),
+            Nil
+          )
+      }).map(getVersionParam)
+    } else {
+      val versionValues = List(
+        FixedExpressionValue("'Json'", "Json"),
+        FixedExpressionValue("'Plain'", "Plain")
+      )
+
+      Writer[List[ProcessCompilationError], List[FixedExpressionValue]](Nil, versionValues).map(versions =>
+        ParameterDeclaration
+          .mandatory[String](KafkaUniversalComponentTransformer.schemaVersionParamName)
+          .withCreator(
+            modify = _.copy(editor = Some(FixedValuesParameterEditor(versions)))
+          )
+      )
+    }
   }
 
   protected def getVersionParam(
@@ -210,5 +238,21 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
 
   // override it if you use other parameter name for topic
   @transient protected lazy val topicParamName: ParameterName = KafkaUniversalComponentTransformer.topicParamName
+
+  protected def getAllTopics: List[UnspecializedTopicName] = {
+    val validatorConfig = kafkaConfig.topicsExistenceValidationConfig.validatorConfig
+
+    KafkaUtils
+      .usingAdminClient(kafkaConfig) {
+        _.listTopics(new ListTopicsOptions().timeoutMs(validatorConfig.adminClientTimeout.toMillis.toInt))
+          .names()
+          .get()
+          .asScala
+          .toSet
+          .map(UnspecializedTopicName.apply)
+          .filterNot(topic => topic.name.startsWith("_"))
+      }
+      .toList
+  }
 
 }
--- a/utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
+++ b/utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
@@ -73,8 +73,8 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
   ): WithError[ParameterCreatorWithNoDependency with ParameterExtractor[String]] = {
     val allTopics = getAllTopics
 
+    // TODO: previously schemaRegistryClient made validation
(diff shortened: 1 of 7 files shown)


utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -98,15 +117,31 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
...
+      (versions match {
+        case Valid(versions) => Writer[List[ProcessCompilationError], List[Integer]](Nil, versions)
+        case Invalid(e) =>
+          Writer[List[ProcessCompilationError], List[Integer]](
+            List(CustomNodeError(e.getMessage, Some(topicParamName))),
+            Nil
+          )
+      }).map(getVersionParam)
+    } else {
+      val versionValues = List(
```
ForrestFairy, line 132: I'm not sure if we need 2 options as I think currently we try to process every message as byte array or string

At least source with "Json" option chosen seems to be working with both normal json and byte array, as I'm testing it by using kafka producer locally

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/SchemaVersionOption.scala
```diff
@@ -21,3 +26,14 @@ object SchemaVersionOption {
...
+
+case class DynamicSchemaVersion(typ: JsonTypes) extends SchemaVersionOption
+
+sealed abstract class JsonTypes(val value: Int) extends IntEnumEntry
+
+object JsonTypes extends IntEnum[JsonTypes] {
+  val values = findValues
+
+  case object Json  extends JsonTypes(1)
+  case object Plain extends JsonTypes(2)
```
raphaelsolarski, line 38: Maybe if we put `Plain` and  `Json` together in enum the enum should not be named `JsonTypes`? Maybe `ContentTypes`?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -98,15 +117,31 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
...
+        FixedExpressionValue("'Json'", "Json"),
```
raphaelsolarski, line 133: IMO we should not use the same `version` parameter for selecting content type for schemaless topics. Based on existence of schema in schema registry we should render `version` or `contentType` parameter in the source.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/SchemaVersionOption.scala
```diff
@@ -21,3 +26,14 @@ object SchemaVersionOption {
 case class ExistingSchemaVersion(version: Int) extends SchemaVersionOption
 
 case object LatestSchemaVersion extends SchemaVersionOption
+
+case class DynamicSchemaVersion(typ: JsonTypes) extends SchemaVersionOption
+
+sealed abstract class JsonTypes(val value: Int) extends IntEnumEntry
```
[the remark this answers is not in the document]
raphaelsolarski: + as I wrote in previous comments IMO we should separate selected version representation from content type.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -156,7 +158,20 @@ object JsonSchemaSupport extends ParsedSchemaSupport[OpenAPIJsonSchema] {
   override def formValueEncoder(schema: ParsedSchema, mode: ValidationMode): Any => AnyRef = {
     val encoder   = new ToJsonSchemaBasedEncoder(mode)
     val rawSchema = schema.cast().rawSchema()
-    (value: Any) => encoder.encodeOrError(value, rawSchema)
+    (value: Any) => {
+      // In ad-hoc test without schema we create object `{ "Value" = userInputInAdHoc }`, so if present we should just take the input
+      Try {
+        val temp = value.asInstanceOf[Map[String, Map[String, Any]]].head
+        if (temp._1.equals("Value")) {
```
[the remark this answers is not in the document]
raphaelsolarski:
```scala
val temp = value.asInstanceOf[Map[String, Map[String, Any]]].head
```
can be replaces with
```scala
val (key, value) = value.asInstanceOf[Map[String, Map[String, Any]]].head
```

raphaelsolarski (changes requested):

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -62,7 +73,15 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
   protected def getTopicParam(
       implicit nodeId: NodeId
   ): WithError[ParameterCreatorWithNoDependency with ParameterExtractor[String]] = {
-    val topics = topicSelectionStrategy.getTopics(schemaRegistryClient)
+    val allTopics = getAllTopics
+    val topics = allTopics match {
```
raphaelsolarski, line 77: Are we sure that the topics will be only from our tenant namespace from hermes?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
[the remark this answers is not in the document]
ForrestFairy: I have switched it back a bit, because when I try to create val to store `temp._2` value it fails. It doesn't matter if I do it as you wrote or separately create `key` and `value`. 
It is strange because there is no problem with assigning `temp._1`

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
raphaelsolarski: Where did we do this check?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -210,5 +247,25 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
 
   // override it if you use other parameter name for topic
   @transient protected lazy val topicParamName: ParameterName = KafkaUniversalComponentTransformer.topicParamName
+  @transient protected lazy val contentTypeParamName: ParameterName =
+    KafkaUniversalComponentTransformer.contentTypeParamName
+
+  protected def getAllTopics: Option[List[UnspecializedTopicName]] = {
```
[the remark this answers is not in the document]
raphaelsolarski: can we move the flag to Kafka Config level?
Let's add it to config documentation.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -210,5 +245,25 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
...
+          _.listTopics(new ListTopicsOptions().timeoutMs(validatorConfig.adminClientTimeout.toMillis.toInt))
+            .names()
+            .get()
+            .asScala
+            .toSet
+            .map(UnspecializedTopicName.apply)
+            .filterNot(topic => topic.name.startsWith("_"))
+        }
+        .toList
+    }.toOption
```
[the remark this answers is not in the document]
raphaelsolarski: what about this one?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/FeatureTogglesConfig.scala
```diff
@@ -28,7 +28,8 @@ final case class FeatureTogglesConfig(
     testDataSettings: TestDataSettings,
     enableConfigEndpoint: Boolean,
     redirectAfterArchive: Boolean,
-    componentDefinitionExtractionMode: ComponentDefinitionExtractionMode
+    componentDefinitionExtractionMode: ComponentDefinitionExtractionMode,
+    fetchKafkaTopicsWithoutSchema: Boolean
```
raphaelsolarski, line 32: we added it here but afaik we don't use it

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -156,7 +158,22 @@ object JsonSchemaSupport extends ParsedSchemaSupport[OpenAPIJsonSchema] {
...
-    (value: Any) => encoder.encodeOrError(value, rawSchema)
+    (value: Any) => {
+      // In ad-hoc test without schema we create object `{ "Value" = userInputInAdHoc }`, so if present we should just take the input
+      Try {
+        val temp = value.asInstanceOf[Map[String, Map[String, Any]]].head
+        val key  = temp._1
+        // Any try to create a variable with value temp._2 fails
+        if (key.equals("Value")) {
+          temp._2
+        } else Failure
```
raphaelsolarski, line 169: hm. returning Failure here is intentional?
If is see it right then If `key` is not equal to `Value` we will try to encode Succees(Failure()) in line 174.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
ForrestFairy: After getting all topics from kafka and concatenating them with topics from schema registry I filter them using `topicSelectionStrategy`.

Now that I think about it should I move the filtering to `getAllTopics`?

defaultModel/src/main/resources/defaultModelConfig.conf
```diff
@@ -30,4 +30,6 @@
   componentsGroupMapping {
     hidden: null
   }
+
+  fetchKafkaTopicsWithoutSchema: true
```
raphaelsolarski, line 34: the default imo should be somewhere in code (like KafkaConfig.topicsExistenceValidationConfig) - afaik if we don't use defaultModel.jar in installation then it won't be applied

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
```diff
@@ -62,7 +73,20 @@ abstract class KafkaUniversalComponentTransformer[T, TN <: TopicName: TopicValid
...
       implicit nodeId: NodeId
   ): WithError[ParameterCreatorWithNoDependency with ParameterExtractor[String]] = {
-    val topics = topicSelectionStrategy.getTopics(schemaRegistryClient)
+    val fetchAllTopics: Option[Boolean] =
+      if (modelDependencies.config.hasPath("fetchKafkaTopicsWithoutSchema")) {
+        Some(modelDependencies.config.getBoolean("fetchKafkaTopicsWithoutSchema"))
+      } else
+        None
+
+    val topics = getAllTopics(fetchAllTopics) match {
```
raphaelsolarski, line 82: imho:
1. Let's move fetchKafkaTopicsWithoutSchema to KafkaConfig and use typed config param where needed.
2. Let's use the `fetchKafkaTopicsWithoutSchema` config to select desired `TopicSelectionStrategy` once (sth like `AllNonHIddenTopics` and `ExistingSchemaRegistrySubjects`).
3. Delete `filterTopics` method from `TopicSelectionStrategy` but add kafka `Admin` to `getTopics` method

IMO it'll make the code more straightforward - we'll put logic about deciding what topics are available to single place. 

@ForrestFairy @arkadius wdyt?

raphaelsolarski (changes requested):

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
arkadius: I aggree but also let's' rename this flag to sth like `showTopicsWithoutSchema`. (`kafka` part is not needed because we are already in kafka config, fetching is a technical, implementation detail that the user is not familiar with)

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/KafkaUniversalComponentTransformer.scala
raphaelsolarski: hm. I think it is a little bit different but still should work - filtering by tenant namespace is done here:
```scala
private def getTopicParam(topics: List[UnspecializedTopicName]) = {
  ParameterDeclaration
    .mandatory[String](topicParamName)
    .withCreator(
      modify = _.copy(editor =
        Some(
          FixedValuesParameterEditor(
            nullFixedValue +: topics
              .flatMap(topic => modelDependencies.namingStrategy.decodeName(topic.name)) // <--- here
              .sorted
              .map(v => FixedExpressionValue(s"'$v'", v))
          )
        )
      )
    )
}
```

raphaelsolarski (approved):

PR #7076, opened 2024-10-24
[NU-1772] Deployment comment validation

mgoworko:
## Describe your changes

When editing or deleting comment assigned to ScenarioDeployed activity Nu should follow the same validation rules like when creating the comment. 

## Checklist before merge


mgoworko:
I added commit with refactor of non-empty comment handling - the autogenerated apply method was still used in a few places, detected by Cypress tests.

designer/server/src/test/scala/pl/touk/nussknacker/ui/config/FeatureTogglesConfigTest.scala
```diff
@@ -12,7 +12,7 @@ class FeatureTogglesConfigTest extends AnyFunSuite with Matchers {
 
   test("should be ok with no deploymentCommentSettings") {
     val config = ConfigFactory
-      .parseResources("config/business-cases/simple-streaming-use-case-designer.conf")
+      .parseResources("config/business-cases/category-used-more-than-once-designer.conf")
```
mk-software-pl, line 15: why?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioActivityApiHttpService.scala
```diff
@@ -535,26 +542,84 @@ class ScenarioActivityApiHttpService(
         scenarioActivityRepository.editComment(
           scenarioId,
           ScenarioActivityId(request.scenarioActivityId),
-          request.commentContent
+          validateComment(_, request.commentContent),
```
mk-software-pl, line 545: I'm not convinced with the current solution.

I see the following problems:
1. The validation should be done in the core layer and the repository layer should not be bothered with this kind of validation. After we validate some comment we should (at the core layer) produce an instance that holds the invariant that proves (to the repository layer) the instance is validated. 

2. The `addNewComment` method adds sth like a simple/regular comment. The `editComment` method modifies a simple/regular comment and deployment comment (this one has different validation rules). We don't have such distinction at the type level, but IMO we should have it. 

3. Comment smart constructor should have a validation logic encapsulated. IMO too much service/core logic is encapsulated in the repository.

mk-software-pl (changes requested):

mk-software-pl (approved):
let's address all the comments in a separate PR

PR #7078, opened 2024-10-24
[NU-1772] Fix line breaks in ScenarioActivities db table

mgoworko:
## Describe your changes

Plaintext activity details from old tables (for example for migration or automatic update) were migrated as part of the JSON in the new table. Newlines were incorrectly migrated as line breaks instead of \n in previous migrations.

## Checklist before merge


mk-software-pl (approved):
Do we need a changelog for that? Or this is just a fix to some features not released, so there is no need to mention that?

PR #7110, opened 2024-10-29
[NU-1836] Handle extension methods in PropertyAccessors

lukasz-bigorajski:
## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/propertyAccessors.scala
```diff
@@ -51,34 +54,63 @@ object propertyAccessors {
...
-      method.invoke(target)
-    }
+    ): Any =
+      methodInvoker.invoke(method)(target, emptyArray)
 
     override def getSpecificTargetClasses: Array[Class[_]] = null
+
+    override def createOptimalAccessor(context: EvaluationContext, target: Any, name: String): PropertyAccessor =
+      super.createOptimalAccessor(context, target, name) match {
+        case o: OptimalPropertyAccessor => new NuOptimalAccessor(o)
```
arkadius, line 86: What's going on here? Won't it itroduce a performance downgrade? Should it be used in every situation? We need to add a verbose comment here :)

PR #7115, opened 2024-10-30
Fix subtyping in dict service

Diamekod0221:
## Describe your changes
DictApiHttpService sometimes returns improper subtypes like Long type Dicts for Integer or won't return proper types like Number type dict for Long. 

Issue is caused by bugs in lang3 lib used in CanBeSubclassDeterminer - the lib would sometimes return incorrect false, and our faulty logic in TypeConversionHandler - the code would return true even if no conversion was possible.

Fixed by adding a double check in SubclassDeterminer and fixing the logic in ConversionHandler.

## Checklist before merge

## Summary by CodeRabbit

- **New Features**
	- Enhanced type conversion checks across various components, allowing for broader compatibility with numeric and collection types.
	- Introduced new methods for strict and loose type assignability checks in the `AssignabilityDeterminer` class.

- **Bug Fixes**
	- Improved error handling and validation logic in multiple areas, ensuring more accurate type compatibility assessments.

- **Documentation**
	- Updated migration guide to reflect changes in version 1.18.0, including configuration and API updates.

- **Tests**
	- Added unit tests for the new `AssignabilityDeterminer` class, covering strict and loose conversion cases.


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -74,14 +74,25 @@ object TypeConversionHandler {
...
+
+    def isFloating(candidate: Class[_]): Boolean = {
+      NumberTypesPromotionStrategy.isFloatingNumber(candidate) || candidate == classOf[java.math.BigDecimal]
+    }
+    def isDecimalNumber(candidate: Class[_]): Boolean = {
+      NumberTypesPromotionStrategy.isDecimalNumber(candidate)
+    }
+
+    boxedSuperclassCandidate match {
+      case candidate if isFloating(candidate) =>
```
arkadius, line 86: Treating specially implicit conversion for floats is not a way to go, we should have option to check can be subclass without conversion in general

arkadius (changes requested):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -206,9 +212,19 @@ trait CanBeSubclassDeterminer {
...
-  // we use explicit autoboxing = true flag, as ClassUtils in commons-lang3:3.3 (used in Flink) cannot handle JDK 11...
-  private def isAssignable(from: Class[_], to: Class[_]): ValidatedNel[String, Unit] =
-    condNel(ClassUtils.isAssignable(from, to, true), (), s"$to is not assignable from $from")
 }
 
-object CanBeSubclassDeterminer extends CanBeSubclassDeterminer
+object CanBeSubclassDeterminer extends CanBeSubclassDeterminer {
+
+  // we use explicit autoboxing = true flag, as ClassUtils in commons-lang3:3.3 (used in Flink) cannot handle JDK 11...
+  def isAssignable(from: Class[_], to: Class[_]): Boolean = {
```
arkadius, line 220: Please add a test showing when lang3 solution works wrong

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
```diff
@@ -135,7 +137,11 @@ trait CanBeSubclassDeterminer {
         (),
         f"${givenClass.display} and ${superclassCandidate.display} are not the same"
       ) orElse
-        isAssignable(givenClass.klass, superclassCandidate.klass)
+        condNel(
+          isAssignable(givenClass.klass, superclassCandidate.klass),
```
arkadius, line 141: Please use `CanBeSubclassDeterminer.isAssignable` to show that it is our class

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
gskrobisz: To clarify: we already treat floats specially, the condition here is not changed: `if` is replaced with pattern matching, the condition is extracted to separate `def` above.

The only change here is in `if isDecimalNumber` section:
```
ConversionFromClassesForDecimals.exists(ClassUtils.isAssignable(boxedGivenClass, _, true))
```
that checks `boxedGivenClass` with all possible supertypes (I guess...)
replaced with 
```
CanBeSubclassDeterminer.isAssignable(boxedGivenClass, candidate)
```
that checks  `boxedGivenClass` with exact given supertype. Aaaand that possibly collides with the `java.math.BigDecimal is quite often returned as a wrapper for all kind of numbers`.

Diamekod0221: Yeah, apart from this line the logic is exactly the same. To clarify, I think the current staging logic is buggy. It only checks whether a givenType CAN be converted, not if the conversion is to the givenSuperclass. Apart from that I've only included a fallback in case lang3 throws a false false, because the lib gives errors when it comes to decimal java types. In general the program execution flow is the same as intended before.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
Diamekod0221: Ok, changed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
Diamekod0221: Added in CanBeSubclassDeterminerSpec

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
Diamekod0221: I have moved this logic to a separate method and restored the previous logic in the handleNumberConversions method

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/CanBeSubclassDeterminer.scala
Diamekod0221: not appliable aanymore, resolve

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/StrictConversionDeterminer.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  def singleCanBeConvertedTo(
+      givenType: SingleTypingResult,
+      superclassCandidate: SingleTypingResult
+  ): ValidatedNel[String, Unit] = {
+    val givenClass     = givenType.runtimeObjType
+    val givenSuperclas = superclassCandidate.runtimeObjType
+
+    isStrictSubclass(givenClass, givenSuperclas)
```
mslabek, line 53: We can make it so the only change in logic between a strict conversion check and loose conversion check is this `singleCanBeConvertedTo`. From what i understand, the only difference should be that we should not allow precision loss on converting numbers when doing a strict check. Just instead of calling `TypeConversionHandler.canBeConvertedTo` i `TypeConversionHandler.canBeStrictlyConvertedTo` and just handle the BigDecimal and BigInteger conversions there.

Right now because theres no reuse of rest of logic we dont handle our whole type system like records or just typed object with values.

Ideally I think we'd have a single object like `ImplicitConversionDeterminer` that has only 2 public defs - `canBeConvertedTo` and `canBeStrictlyconvertedTo` . Everything else should be private so that no one else reuses it in context we dont control.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
```diff
@@ -0,0 +1,39 @@
+package pl.touk.nussknacker.engine.api.typed
+
+import org.apache.commons.lang3.ClassUtils
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class SubclassDeterminerSpec extends AnyFunSuite with Matchers {
+
+  test("Should validate assignability for decimal types") {
+    StrictConversionDeterminer.isAssignable(classOf[java.lang.Long], classOf[java.lang.Integer]) shouldBe false
```
mslabek, line 10: I think we shouldn't ever check `.isAssignable` - the only things that should be public in a ConversionDeterminer is just the check taking 2 values which are types from out typing system.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
```diff
@@ -0,0 +1,39 @@
+package pl.touk.nussknacker.engine.api.typed
+
+import org.apache.commons.lang3.ClassUtils
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class SubclassDeterminerSpec extends AnyFunSuite with Matchers {
+
+  test("Should validate assignability for decimal types") {
```
mslabek, line 9: It would be great to have a precise comparison of our 2 ways of comparing types (currently from StrictConversionDeterminer and ImplcitConversionDeterminer). I'd prefer to do it on a one test data table and use it in 2 parameterized TableDrivenPropertyChecks tests. Then we have a proof that the strict checker work on other data types like maps, lists, records, tagged types.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
Diamekod0221: I have doubts whether being very strict with public/private here is the correct way, we could possibly end up in a situation where the next person is also surprised by how this method works. I will remove the test and add a comment in isAssignable to prevent that

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  test("Should validate assignability for numerical types") {
+    StrictConversionDeterminer.isAssignable(classOf[java.lang.Long], classOf[java.lang.Double]) shouldBe true
+    StrictConversionDeterminer.isAssignable(classOf[java.lang.Float], classOf[Double]) shouldBe true
+
+    StrictConversionDeterminer.isAssignable(classOf[Integer], classOf[java.lang.Float]) shouldBe true
+    StrictConversionDeterminer.isAssignable(classOf[java.lang.Long], classOf[java.lang.Double]) shouldBe true
+  }
+
+  // to check if autoboxing lang3 is failing - we can remove our fallback from SubclassDeterminer.isAssignable if the lib works properly
```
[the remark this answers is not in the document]
Diamekod0221: removed

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
mslabek: Why leave it public? Its not / shouldn't be used in this project or outside it.

Diamekod0221: Just for testing, i agree that it shouldn't be used outside

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/StrictConversionDeterminer.scala
Diamekod0221: Refactroed to a single AssignabilityDeterminer as per your PoC

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/DictApiEndpoints.scala
```diff
@@ -81,23 +83,20 @@ class DictApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoin
...
 
-  object Dtos {
 
-    @JsonCodec
-    case class TypingResultInJson(value: Json)
+  object Dtos {
 
     @JsonCodec
-    case class DictListRequestDto(expectedType: TypingResultInJson)
+    case class DictListRequestDto(expectedType: Json)
```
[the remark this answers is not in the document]
Diamekod0221: added

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/SubclassDeterminerSpec.scala
Diamekod0221: I have added such tests in AssignabilityDeterminerSpec. I only compare the numerical types that actually differ in those two Determiners. The rest is already tested in typingResultSpec.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminerSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    ("sourceType", "targetType", "expectedStrict", "expectedLoose"),
+    (typing.Typed[Long], typing.Typed[Int], Invalid(NonEmptyList.of("")), Valid(())),
+    (typing.Typed[Double], typing.Typed[Int], Invalid(NonEmptyList.of("")), Invalid(NonEmptyList.of(""))),
+    (typing.Typed[BigDecimal], typing.Typed[Int], Invalid(NonEmptyList.of("")), Invalid(NonEmptyList.of("")))
+  )
+
+  test("isAssignableStrict should fail for looser numerical cases") {
+    forAll(looseConversionCases) { (sourceType, targetType, expectedStrict, _) =>
+      val result = AssignabilityDeterminer.isAssignableStrict(sourceType, targetType)
+      result match {
```
mslabek, line 48: This way of pattern matching in tests is not that readable and is harder to understand imo. For widening conversions both `isAssignableStrict` and `isAssignableLoose` should pass so we can just check if `isAssignableX` returns valid. For narrowing conversions `isAssignableLoose` should return valid and `isAssignableStrict` false.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -84,6 +94,28 @@ object TypeConversionHandler {
...
+
+  private def handleStrictNumberConversions(givenClass: TypedClass, superclassCandidate: TypedClass): Boolean = {
+
+    val boxedGivenClass          = ClassUtils.primitiveToWrapper(givenClass.klass)
+    val boxedSuperclassCandidate = ClassUtils.primitiveToWrapper(superclassCandidate.klass)
+    // TODO: This is probably wrong - relying on index of AllNumbers
+    (boxedGivenClass, boxedSuperclassCandidate) match {
+      case (f, t) if ClassUtils.isAssignable(f, t, true) => true
+      case (f, t) if (AllNumbers.contains(f) && AllNumbers.contains(t)) =>
+        AllNumbers.indexOf(f) >= AllNumbers.indexOf(t)
```
mslabek, line 113: This logic makes it so that in strict conversion check BigInteger can be assinged to Float or Double, which is incorrect since its like a narrowing conversion. If its corrected you can remove the TODO comment.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -84,6 +94,28 @@ object TypeConversionHandler {
...
+    handleStrictNumberConversions(givenType.runtimeObjType, superclassCandidate)
+  }
+
+  private def handleStrictNumberConversions(givenClass: TypedClass, superclassCandidate: TypedClass): Boolean = {
+
+    val boxedGivenClass          = ClassUtils.primitiveToWrapper(givenClass.klass)
+    val boxedSuperclassCandidate = ClassUtils.primitiveToWrapper(superclassCandidate.klass)
+    // TODO: This is probably wrong - relying on index of AllNumbers
+    (boxedGivenClass, boxedSuperclassCandidate) match {
+      case (f, t) if ClassUtils.isAssignable(f, t, true) => true
```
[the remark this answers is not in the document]
Diamekod0221: removed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -0,0 +1,249 @@
+package pl.touk.nussknacker.engine.api.typed
+
+import cats.data.Validated._
+import cats.data.{NonEmptyList, Validated, ValidatedNel}
+import cats.implicits.{catsSyntaxValidatedId, _}
+import org.apache.commons.lang3.ClassUtils
+import pl.touk.nussknacker.engine.api.typed.typing._
+
+/**
+  * This class determine if type can be subclass of other type. It basically based on fact that TypingResults are
```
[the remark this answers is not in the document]
Diamekod0221: corrected

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -84,6 +94,28 @@ object TypeConversionHandler {
...
 
+  def canBeStrictlyConvertedTo(givenType: SingleTypingResult, superclassCandidate: TypedClass): Boolean =
+    handleStrictConversion(givenType, superclassCandidate)
+
+  private def handleStrictConversion(givenType: SingleTypingResult, superclassCandidate: TypedClass) = {
+    handleStringToValueClassConversions(givenType, superclassCandidate) ||
+    handleStrictNumberConversions(givenType.runtimeObjType, superclassCandidate)
+  }
+
+  private def handleStrictNumberConversions(givenClass: TypedClass, superclassCandidate: TypedClass): Boolean = {
```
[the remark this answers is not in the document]
Diamekod0221: fixed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
Diamekod0221: I havechanged it so we can only assign BigInteger to BigDecimal

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminerSpec.scala
Diamekod0221: I think the pattern matching is ok, otherwise we would have to write specific match messages into the table and that cluttered it a lot.
I change the test cases so that the table is simpler.

components-api/src/test/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminerSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.scalatest.prop.TableDrivenPropertyChecks.forAll
+import org.scalatest.prop.Tables.Table
+import pl.touk.nussknacker.engine.api.typed._
+
+class AssignabilityDeterminerSpec extends AnyFunSuite with Matchers {
+
+  // Test data table: (sourceType, targetType, expectedStrict, expectedLoose)
+  val strictConversionCases = Table(
+    ("sourceType", "targetType", "expectedStrict", "expectedLoose"),
+    (typing.Typed[Int], typing.Typed[Int], Valid(()), Valid(())),                           // Same primitive type
```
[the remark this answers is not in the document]
Diamekod0221: done

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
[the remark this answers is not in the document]
Diamekod0221: removed

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/DictApiHttpServiceSpec.scala
```diff
@@ -16,17 +16,37 @@ class DictApiHttpServiceSpec
...
+            |     "params":[]
+            |     }
+            |}""".stripMargin)
+        .post(s"$nuDesignerHttpAddress/api/processDefinitionData/${Streaming.stringify}/dicts")
+        .Then()
+        .statusCode(200)
+        .equalsJsonBody(
+          s"""[
+             |]""".stripMargin
+        )
```
mslabek, line 37:
```suggestion
        .equalsJsonBody("[]")
```

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -0,0 +1,251 @@
+package pl.touk.nussknacker.engine.api.typed
+
+import cats.data.Validated._
+import cats.data.{NonEmptyList, Validated, ValidatedNel}
+import cats.implicits.{catsSyntaxValidatedId, _}
+import org.apache.commons.lang3.ClassUtils
+import pl.touk.nussknacker.engine.api.typed.typing._
+
+/**
+  * This class determine whether we can convert a class to another one. We provide two modes of conversion -
```
mslabek, line 10:
```suggestion
  * This class determine whether we can assign one type to another type - that is if its the same class, a subclass or can be converted to another type. We provide two modes of conversion -
```

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -0,0 +1,251 @@
...
+ * 1. Loose conversion is based on the fact that TypingResults are
+ * sets of possible supertypes with some additional restrictions (like TypedObjectTypingResult). It is basically how SpEL
+ * can convert things. Like CommonSupertypeFinder it's in the spirit of "Be type safe as much as possible, but also provide some helpful
+ * conversion for types not in the same jvm class hierarchy like boxed Integer to boxed Long and so on".
+ * 2. Strict conversion checks whether we can convert the type without losing precision. Eg only widening numerical types
```
mslabek, line 15: So I dug into this and its more complex unfortunately, but it doesn't matter for us, we just have to temper our expectations for how strict this strict check is :smile: 

Take a look at: https://docs.oracle.com/javase/specs/jls/se11/html/jls-5.html 5.1.2. Widening Primitive Conversion

Widening conversions are what we have in our strict check, just extended with BigInteger and BigDecimal. Thats what we have right now in the strict check.

What we don't have is check if the conversion can happen without loss of precision.

"A widening primitive conversion from float to double that is not strictfp may lose information about the overall magnitude of the converted value.

A widening primitive conversion from int to float, or from long to float, or from long to double, may result in loss of precision - that is, the result may lose some of the least significant bits of the value. In this case, the resulting floating-point value will be a correctly rounded version of the integer value, using IEEE 754 round-to-nearest mode ([§4.2.4](https://docs.oracle.com/javase/specs/jls/se11/html/jls-4.html#jls-4.2.4))."

If we allow for example int -> float, we can't say we have conversions without losing precision.

Intuitively, I'd say we should allow only precision-loss-less conversion in a method thats called `isAssignableStrict`, but `isAssignable` from ClassUtils also doesnt take that into account, so its ok for me either way, its up to you. Just if we do it with allowing precision loss, then lets remove the `without losing precision` - just say its only widening conversions.

Diamekod0221: I'd say since now it's not needed i will leave the code as it is and just reword the comment. Making it without loss of precision would require adding a lot of logic probably. I changed the comment.

Diamekod0221:
Closed because of CI unreposnsitivy and moved to https://github.com/TouK/nussknacker/pull/7246

PR #7118, opened 2024-10-31
[NU-1836] Handle extension methods in PropertyAccessors

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
@@ -64,7 +64,7 @@ private[spel] class Typer(
     staticMethodInvocationsChecking: Boolean,
     classDefinitionSet: ClassDefinitionSet,
     evaluationContextPreparer: EvaluationContextPreparer,
-    methodExecutionForUnknownAllowed: Boolean,
+    anyMethodExecutionForUnknownAllowed: Boolean,
     dynamicPropertyAccessAllowed: Boolean
 ) extends LazyLogging {
 
@@ -73,7 +73,7 @@ private[spel] class Typer(
   private lazy val evaluationContext: EvaluationContext =
     evaluationContextPreparer.prepareEvaluationContext(Context(""), Map.empty)
 
-  private val methodReferenceTyper = new MethodReferenceTyper(classDefinitionSet, methodExecutionForUnknownAllowed)
+  private val methodReferenceTyper = new MethodReferenceTyper(classDefinitionSet, anyMethodExecutionForUnknownAllowed)
 
   private lazy val typeReferenceTyper = new TypeReferenceTyper(evaluationContext, classDefinitionSet)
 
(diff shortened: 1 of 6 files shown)


arkadius (approved):
LGTM, let's @raphaelsolarski also take a look

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/propertyAccessors.scala
```diff
@@ -31,7 +33,8 @@ object propertyAccessors {
       new NoParamMethodPropertyAccessor(accessChecker),
       // it can add performance overhead so it will be better to keep it on the bottom
       MapLikePropertyAccessor,
-      MapMissingPropertyToNullAccessor, // must be after NoParamMethodPropertyAccessor
+      new ExtensionMethodsPropertyAccessor(accessChecker, set),
```
arkadius, line 36: nip: Maybe let's move it after NoParamMethodPropertyAccessor to keep map related accessors together

lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -64,7 +64,7 @@ private[spel] class Typer(
     staticMethodInvocationsChecking: Boolean,
     classDefinitionSet: ClassDefinitionSet,
     evaluationContextPreparer: EvaluationContextPreparer,
-    methodExecutionForUnknownAllowed: Boolean,
+    anyMethodExecutionForUnknownAllowed: Boolean,
```
raphaelsolarski, line 67: the `methodExecutionForUnknownAllowed` is the name of feature flag in `ExpressionConfig`, so imo lets rename it also there and in it's java counterpart or lets leave the previous name unchanged everywhere.

raphaelsolarski (approved):
lgtm, I leaved one minor comment

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
lukasz-bigorajski: I made it like this because of: https://github.com/TouK/nussknacker/pull/7110#discussion_r1821356588
I would leave as it is

arkadius: Yep, take a look @raphaelsolarski how it loos without this:
if (methodExecutionForUnknownAllowed)
...
else // assume that is not allowed
...  // execute method for unknown
:)

raphaelsolarski: Agree, but why don't we change it in other places ("ExpressionConfig and friends")?

PR #7127, opened 2024-11-06
1.18 release highlights

arkadius:
## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -6,17 +6,17 @@
 
 ##### End-user
 
-* Versions, Comments and Attachments panels were replaced by Activities panel
-* Added support for scenario labels
-* SpEL: support for navigation through fields inside variables typed as `Unknown` 
-* SpEL: conversion methods allowing to cast or convert given type to another
-* Other SpEL enhancements: `#CONV.toJson`, `#CONV.toJsonString` methods, new `#BASE64` helper, it is possible to operate on arrays as on lists and more
-* UX improvements: new UI for defining aggregations, better validation handling in ad-hoc testing feature and more
+* New Activities panel, replacing Versions, Comments and Attachments panels. Now you can browse all scenario activities on one chronological list.
+* Added scenario labels. You can now organize your scenarios and find different groups of scenarios more easily.
+* SpEL: added navigation through fields inside variables typed as Unknown. You can now access the data inside a variable, even if Nussknacker doesn't know its exact type during scenario authoring.
+* SpEL: added conversion methods to cast or convert between data types (e.g. `String` to `Integer`).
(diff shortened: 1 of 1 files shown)


docs/Changelog.md
```diff
@@ -4,7 +4,19 @@
...
 
-(Not available yet)
+##### End-user
+
+* Versions, Comments and Attachments panels were replaced by Activities panel
+* Added support for scenario labels
+* SpEL: support for navigation through fields inside variables typed as `Unknown` 
+* SpEL: conversion methods allowing to cast or convert given type to another
+* Other SpEL enhancements: `#CONV.toJson`, `#CONV.toJsonString` methods, new `#BASE64` helper, it is possible to operate on arrays as on lists and more
+* UX improvements: new UI for defining aggregations, better validation handling in ad-hoc testing feature and more
```
jzdr, line 14: I would suggest a bit more verbose version:
* New Activities panel, replacing Versions, Comments and Attachments panels. Now you can browse all scenario activities on one chronological list.
* Added scenario labels. You can now organize your scenarios and find different groups of scenarios more easily.
* SpEL: added navigation through fields inside variables typed as `Unknown`. You can now access the data inside a variable, even if Nussknacker doesn't know its exact type.
* SpEL: added conversion methods to cast and convert between data types (eg. String to Integer).
* SpEL:  various enhancements, like `#CONV.toJson` and `#CONV.toJsonString` methods, new `#BASE64` helper, possibility to treat arrays as lists, and more.
* Various UX improvements, including new UI for aggregation definitions and better validation handling in ad-hoc tests.

docs/Changelog.md
```diff
@@ -4,7 +4,19 @@
...
+
+##### Administrator
+
+* Flink upgrade to 1.19.1. Note: if you want to use Nussknacker with older versions of Flink it is possible, but it requires some extra steps to be done. See [Migration guide](MigrationGuide.md) for details
```
jzdr, line 18: Slightly rephrased:
* Flink upgrade to 1.19.1. Note: it is possible to use Nussknacker with older versions of Flink, but it requires some extra steps. See [Migration guide](MigrationGuide.md) for details.

docs/Changelog.md
arkadius: Nice, I applied this suggestion but with slight differences:
- "Nussknacker doesn't know its exact type" -> "Nussknacker doesn't know its exact type during scenario authoring" - eventually Nussknacker will know the type, but it will be done in runtime
- "added conversion methods to cast and convert" -> "added conversion methods to cast OR convert" - I'm not convinced about that, but IMO AND suggests that we added both methods for casting and converting but in fact we added only one type of behaviour: we try to cast and then try to convert. I think that we should invent some word for that such as "smart cast" :)

WDYT?

jzdr (approved):

PR #7134, opened 2024-11-07
[1.18][NU-1772] Scenario activities fixes

mgoworko:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
@@ -1,5 +1,6 @@
 package pl.touk.nussknacker.ui.process.migrate
 
+import cats.data.NonEmptyList
 import pl.touk.nussknacker.engine.api.graph.ScenarioGraph
 import pl.touk.nussknacker.engine.api.process.{ProcessId, ProcessName}
 import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
@@ -11,15 +12,23 @@ import pl.touk.nussknacker.ui.process.repository.ScenarioWithDetailsEntity
 
 final case class MigrationResult(process: CanonicalProcess, migrationsApplied: List[ProcessMigration]) {
 
-  def toAutomaticProcessUpdateAction(processId: ProcessId, labels: List[ScenarioLabel]): AutomaticProcessUpdateAction =
-    AutomaticProcessUpdateAction(
-      processId = processId,
-      canonicalProcess = process,
-      labels = labels,
-      increaseVersionWhenJsonNotChanged = true,
-      forwardedUserName = None,
-      migrationsApplies = migrationsApplied
-    )
+  def toAutomaticProcessUpdateAction(
+      processId: ProcessId,
+      labels: List[ScenarioLabel]
+  ): Option[AutomaticProcessUpdateAction] = {
+    NonEmptyList
+      .fromList(migrationsApplied)
+      .map { migrations =>
+        AutomaticProcessUpdateAction(
+          processId = processId,
+          canonicalProcess = process,
+          labels = labels,
+          increaseVersionWhenJsonNotChanged = true,
+          forwardedUserName = None,
+          migrationsApplied = migrations
+        )
+      }
+  }
 
 }
 
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
@@ -14,6 +14,7 @@ import pl.touk.nussknacker.engine.deployment.{DeploymentData, DeploymentId}
 import pl.touk.nussknacker.engine.management.periodic.PeriodicProcessService.{
   DeploymentStatus,
   EngineStatusesToReschedule,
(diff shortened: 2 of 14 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/ProcessModelMigrator.scala
```diff
@@ -11,15 +12,23 @@ import pl.touk.nussknacker.ui.process.repository.ScenarioWithDetailsEntity
...
-      canonicalProcess = process,
-      labels = labels,
-      increaseVersionWhenJsonNotChanged = true,
-      forwardedUserName = None,
-      migrationsApplies = migrationsApplied
-    )
+  def toAutomaticProcessUpdateAction(
+      processId: ProcessId,
+      labels: List[ScenarioLabel]
+  ): Option[AutomaticProcessUpdateAction] = {
```
mk-software-pl, line 18: I'm not sure if we shouldn't allow producing MigrationResult with Nil in `migrationsApplied`. WDYT?

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivity.scala
```diff
@@ -44,17 +44,56 @@ sealed trait ScenarioComment {
 
 object ScenarioComment {
```
mk-software-pl, line 45: TBH, I don't know if I get why we have this comments-like hierarchy and the `pl.touk.nussknacker.engine.api.Comment` class. And there are no connection between these two?

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
```diff
@@ -535,11 +535,19 @@ class PeriodicProcessService(
...
           Some(ScheduledExecutionStatus.Failed)
         case PeriodicProcessDeploymentStatus.RetryingDeploy =>
-          Some(ScheduledExecutionStatus.DeploymentWillBeRetried)
+          None
         case PeriodicProcessDeploymentStatus.FailedOnDeploy =>
-          Some(ScheduledExecutionStatus.DeploymentFailed)
+          None
       }
-    } yield (status, dateFinished)
+      dateCreated = instantAtSystemDefaultZone(entity.createdAt)
```
mk-software-pl, line 542: the existence of the `instantAtSystemDefaultZone` def forces me to ask the following questions:
1. why do we operate on LocalDateTime, not on Instant in the domain? are we able to improve it?
2. shouldn't we have the `ZonedDateTime` in the API responses rather than Instant? (aka, "consume utc dates, return dates with zone" principle)?

mgoworko: Current state overall:
- we operate on LocalDateTime here, because that is how they are stored in the periodic db
- this db is the main problem - there is no info about timezone for periodics at all
- even searching for deployment to run is performed on local times. So theoretically it would fail, if system timezone would change between scheduling and running (or some other hypothetical instance of Nu is running on different timezone - it uses Clock.systemDefaultZone)

This specific case:
- I may be wrong, but AFAIK all, or nearly all, APIs operate on Instant's
- instantAtSystemDefaultZone is a conversion caused by the problems related by storing LocalDateTime's in the periodic db
- API returns dates in two ways: 
  - as Instant's, then it is formatted by FE
  - as generic params, as plain text - then is it formatted by BE in systemDefaultZone. The BE implementation is prepared to accept timezone in request, but it is not done ATM. (There is comment 'TODO: pass client timezone from FE'  for similar case - pdf generation

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ScenarioActivity.scala
```diff
@@ -283,10 +307,8 @@ object ScenarioActivity {
   ) extends BatchDeploymentRelatedActivity {
 
     override def result: DeploymentResult = scheduledExecutionStatus match {
-      case ScheduledExecutionStatus.Finished                => DeploymentResult.Success(dateFinished)
-      case ScheduledExecutionStatus.Failed                  => DeploymentResult.Failure(dateFinished, None)
-      case ScheduledExecutionStatus.DeploymentWillBeRetried => DeploymentResult.Failure(dateFinished, None)
-      case ScheduledExecutionStatus.DeploymentFailed        => DeploymentResult.Failure(dateFinished, None)
```
arkadius, line 289: Why is it not an activity anymore? It looks like important information for the user: we skipped some schedule invocation because of deployment failure.

mgoworko: Restored, with fixed date handling (for error statuses there it no starting date available)

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/PeriodicProcessService.scala
mgoworko: Refactored, using ZonedDateTime, and conversions on repo level.

mgoworko: discussed and reverted because of some problems detected by tests, left for later

PR #7137, opened 2024-11-08
[NU-1872] Bump Flink to 1.20

mslabek:
## Checklist before merge

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -299,9 +299,9 @@ lazy val commonSettings =
 // Note: when updating check versions in 'flink*V' below, because some libraries must be fixed at versions provided
 // by Flink, or jobs may fail in runtime when Flink is run with 'classloader.resolve-order: parent-first'.
 // You can find versions provided by Flink in it's lib/flink-dist-*.jar/META-INF/DEPENDENCIES file.
-val flinkV                = "1.19.2"
-val flinkConnectorKafkaV  = "3.2.0-1.19"
-val jdbcFlinkConnectorV   = "3.2.0-1.19"
+val flinkV                = "1.20.1"
+val flinkConnectorKafkaV  = "3.3.0-1.20"
+val jdbcFlinkConnectorV   = "3.3.0-1.20"
 val flinkCommonsCompressV = "1.26.0"
 val flinkCommonsLang3V    = "3.12.0"
 val flinkCommonsTextV     = "1.10.0"
--- a/build.sbt
+++ b/build.sbt
@@ -300,7 +300,7 @@ lazy val commonSettings =
 // by Flink, or jobs may fail in runtime when Flink is run with 'classloader.resolve-order: parent-first'.
 // You can find versions provided by Flink in it's lib/flink-dist-*.jar/META-INF/DEPENDENCIES file.
 val flinkV                = "1.20.1"
-val flinkConnectorKafkaV  = "3.3.0-1.20"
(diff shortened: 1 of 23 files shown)


build.sbt
```diff
@@ -297,7 +297,7 @@ val scalaCheckVshort      = scalaCheckV.take(4).replace(".", "-")
 val scalaTestPlusV        =
   "3.2.18.0" // has to match scalatest and scalacheck versions, see https://github.com/scalatest/scalatestplus-scalacheck/releases
 // note: Logback 1.3 requires Slf4j 2.x, but Flink has Slf4j 1.7 on its classpath
-val logbackV                = "1.2.13"
+val logbackV                = "1.5.12"
```
piotrp, line 300: Are you sure you want to change it? Flink 1.20 provides `org.slf4j:slf4j-api:jar:1.7.36`, which is incompatible with this.

mslabek: No, you're right, I added this just to check if there are any incompatibilities like with 1.19 (there doesn't seem to be) and to check if tests pass besides this.

With previous versions of logback and slf4j when launching a Flink minicluster tests failed with:
`java.lang.NullPointerException: null
	at java.base/java.util.HashMap.putMapEntries(HashMap.java:497)
	at java.base/java.util.HashMap.putAll(HashMap.java:781)
	at java.base/java.util.Collections$SynchronizedMap.putAll(Collections.java:2604)
	at ch.qos.logback.classic.util.LogbackMDCAdapter.setContextMap(LogbackMDCAdapter.java:197)
	at org.slf4j.MDC.setContextMap(MDC.java:264)
	at org.apache.flink.util.MdcUtils.lambda$withContext$0(MdcUtils.java:48)
	at org.apache.flink.runtime.rpc.pekko.PekkoRpcActor.handleControlMessage(PekkoRpcActor.java:208)
`

After bumping the versions, the tests run ok. I wonder why our e2e tests dont blow up if we add an incompatible dependency like this :/ I'll check this

Anyways, there is a ticket for this problem: https://issues.apache.org/jira/browse/FLINK-36227 that is yet to be addressed.

I'll revert this and slf4j version bumps, see if there's a workaround and if not I'll monitor the Flink jira ticket. Thanks!

build.sbt
```diff
@@ -1926,17 +1926,23 @@ lazy val flinkTableApiComponents = (project in flink("components/table"))
...
+        "org.apache.flink"   % "flink-streaming-java"          % flinkV              % Provided,
+        "org.apache.flink"   % "flink-table-api-java"          % flinkV              % Provided,
+        "org.apache.flink"   % "flink-table-api-java-bridge"   % flinkV              % Provided,
+        "org.apache.flink"   % "flink-connector-jdbc-postgres" % jdbcFlinkConnectorV % Test,
       )
-    }
+    },
+    assembly / assemblyMergeStrategy := {
+      case PathList("org", "apache", "calcite", "sql", "fun", _) =>
+        MergeStrategy.last // taking version from flink-sql-parser instead of calcite-core
```
piotrp, line 1942: Shouldn't you just make sure that these versions match? Won't this cause issues in the future?

piotrp:
Changelog entry is missing

build.sbt
lukasz-bigorajski: What do you mean, which version should I check?
There is only a problem with `flink-sql-parser` which has dependency to `calcite-core` and in the same time it's overriding some classes from the lib.
On the other hand we have `calcite-linq4j` which should be in sync with calcite-core version and as far as I know the only way to have it in sync is to check it manually.

piotrp (approved):

PR #7167, opened 2024-11-19
[NU-7164] Add basic parser for fragment input definitions

DeamonDev:
## Checklist before merge


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/FlinkProcessCompilerDataFactory.scala
```diff
@@ -137,13 +137,18 @@ class FlinkProcessCompilerDataFactory(
...
+      modelDefinitionWithTypes.classDefinitions.all
+    )
     (ModelDefinitionWithClasses(adjustedDefinitions), dictRegistry)
   }
 
   protected def adjustDefinitions(
       originalModelDefinition: ModelDefinition,
-      definitionContext: ComponentDefinitionContext
+      definitionContext: ComponentDefinitionContext,
+      classDefinitions: Set[ClassDefinition]
```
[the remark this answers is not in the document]
DeamonDev: There is only one usage of that one and I feeded it with proper value

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/fragment/FragmentParameterTypingParser.scala
```diff
@@ -0,0 +1,45 @@
...
+        .fold(
+          // This is fallback - it may be removed and `ClassNotFound` exception may be thrown here after cleaning up the mess with `FragmentClazzRef` class
+          Typed(ClassUtils.getClass(classLoader, simpleClassName))
+        ) { classDefinition =>
+          classDefinition.clazzName
+        }
+
+    val mapPattern  = "Map\\[(.+),(.+)\\]".r
+    val listPattern = "List\\[(.+)\\]".r
+    val setPattern  = "Set\\[(.+)\\]".r
```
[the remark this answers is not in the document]
DeamonDev: Currently we are OK with that this will be shallow parsing

designer/client/src/components/graph/node-modal/editors/expression/Table/TableEditor.tsx
```diff
@@ -97,7 +97,7 @@ export const Table = ({ expressionObj, onValueChange, className, fieldErrors }:
         }
     }, [expressionObj.expression, onValueChange, rawExpression]);
 
-    const { defaultTypeOption, orderedTypeOptions } = useTypeOptions<SupportedType>();
+    const { defaultTypeOption, orderedTypeOptions } = useTypeOptions<SupportedType>(false);
```
Dzuming, line 100: Why do we need this change?, false is not a valid Supported type

designer/client/src/components/graph/node-modal/fragment-input-definition/FragmentInputDefinition.tsx
```diff
@@ -14,13 +14,13 @@ interface Props extends Omit<MapVariableProps<FragmentInputParameter>, "readOnly
     isEditMode?: boolean;
 }
 
-export function useTypeOptions<Value = string>() {
+export function useTypeOptions<Value>(useDisplay = false) {
```
Dzuming, line 17: Why do we need this `useDisplay` flag?

designer/client/src/components/graph/node-modal/editors/expression/Table/TableEditor.tsx
DeamonDev: See: https://github.com/TouK/nussknacker/pull/7167#discussion_r1850207954

designer/client/src/components/graph/node-modal/fragment-input-definition/FragmentInputDefinition.tsx
DeamonDev: See: https://github.com/TouK/nussknacker/pull/7167#discussion_r1850207954

designer/client/src/components/graph/node-modal/editors/expression/Table/TableEditor.tsx
DeamonDev: It is certainly possible but I do not want to make this refactor. I tried to do it for `TableEditor` and spend on it some time but there is some internal dependency on `.refClazzName` inside it...

DeamonDev: So long story short is that `TableEditor` uses it _somehow_ and `FragmentInputDefinition` does not use it anymore and relies on simple class names which are resolved on BE.

Dzuming: Ok, thanks for the explanation, two more questions 
1.  Based on [types](https://github.com/TouK/nussknacker/blob/49dc2be6f59f2f74fa2991ab0ac3023448c78a41/designer/client/src/types/definition.ts#L6) `.display` is required property, is it also true for `TableEditor`? If not, could you adjust this type?
2. If there is no `.display` in the case of `TableEditor`, could you simplify this logic and just make it `type.display  ||  type.refClazzName`?

DeamonDev: But there is `.display` in the `TableEditor` context. This works as follows: 

Request is performed to definition data. We obtain information containing `display` as well as `refClazzName`.  The problem is that we would like to use create `value` part of this hook-data depending on place where we call this hook (aka hook-use context). So the simplification `type.display  ||  type.refClazzName` will not working bcs this will be evaluated to `type.display` which should not be present in the context of `TableEditor`....

Dzuming: Okay, I got it. Let's return to the implementation. To be honest, I prefer your first approach (separating it into two methods and adding meaningful names). Even if the code is duplicated, IMO readability is more important.
In the current approach, this boolean says nothing and I need to go into implementation details to check what's going on here

```
const { orderedTypeOptions, defaultTypeOption } = useTypeOptions<string>(true);
```

we could change it to an object

```
const { orderedTypeOptions, defaultTypeOption } = useTypeOptions<string>({ useDisplay:true });
```

but still, I don't know what happens when I set it to false, and what exactly happens when it is true.

DeamonDev: As we discussed in DM I would go with two different hooks.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/FragmentParameterValidator.scala
```diff
@@ -147,9 +157,11 @@ object FragmentParameterValidator {
...
           dictionaries.get(dictId) match {
             case Some(dictDefinition) =>
-              val fragmentParameterTypingResult = fragmentParameter.typ
-                .toRuntimeClass(classLoader)
-                .map(Typed(_))
+              val fragmentParameterTypingParser = new FragmentParameterTypingParser(classLoader, classDefinitions)
+              val fragmentParameterTypingResult = fragmentParameterTypingParser
+                .parseClassNameToTypingResult(
+                  fragmentParameter.typ.refClazzName
+                )
```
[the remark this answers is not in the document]
DeamonDev: Not the scope of this PR

designer/client/src/components/graph/node-modal/editors/expression/Table/TableEditor.tsx
```diff
@@ -87,6 +91,27 @@ const emptySelection = {
...
+    );
+
+    const orderedTypeOptions = useMemo(() => orderBy(typeOptions, (item) => [item.label, item.value], ["asc"]), [typeOptions]);
+
+    const defaultTypeOption = useMemo(() => find(typeOptions, { label: "String" }) || head(typeOptions), [typeOptions]);
+    return {
+        orderedTypeOptions,
+        defaultTypeOption,
+    };
+}
```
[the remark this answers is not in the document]
DeamonDev: Not the scope of this PR

designer/client/src/components/graph/node-modal/editors/expression/Table/TableEditor.tsx
```diff
@@ -97,7 +121,7 @@ export const Table = ({ expressionObj, onValueChange, className, fieldErrors }:
         }
     }, [expressionObj.expression, onValueChange, rawExpression]);
 
-    const { defaultTypeOption, orderedTypeOptions } = useTypeOptions<SupportedType>();
+    const { defaultTypeOption, orderedTypeOptions } = useTableEditorTypeOptions();
```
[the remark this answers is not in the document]
DeamonDev: Not the scope of this PR

Dzuming (approved):
FE part LGTM

PR #7277, opened 2024-12-03
Sticky notes 

philemone:
## Checklist before merge

## Release Notes

- **New Features**
	- Introduced sticky notes functionality, allowing users to create, update, and delete sticky notes within scenarios.
	- Added new API endpoints for managing sticky notes: 
		- `GET /api/processes/{scenarioName}/stickyNotes`
		- `PUT /api/processes/{scenarioName}/stickyNotes`
		- `POST /api/processes/{scenarioName}/stickyNotes`
		- `DELETE /api/processes/{scenarioName}/stickyNotes/{noteId}`
	- New configuration options for sticky notes, including maximum content length and count limits.
	- Enhanced sticky notes settings for better management.

- **Bug Fixes**
	- Enhanced error handling for sticky notes operations.

- **Documentation**
	- Updated changelog and migration guide to reflect sticky notes feature and API changes.


designer/client/package.json
```diff
@@ -22,6 +22,7 @@
         "d3-transition": "3.0.1",
         "d3-zoom": "3.0.0",
         "dagre": "0.8.5",
+        "dompurify": "3.2.0",
```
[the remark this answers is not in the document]
philemone: Najnowsza wersja to 3.2.2
https://github.com/cure53/DOMPurify

designer/client/cypress/e2e/creatorToolbar.cy.ts
```diff
@@ -26,6 +26,7 @@ describe("Creator toolbar", () => {
         cy.contains(/^types$/i).click();
         cy.contains(/^services$/i).click();
         cy.contains(/^sinks$/i).click();
+        cy.contains(/^stickyNotes$/i).click();
```
gskrobisz: Label looks like technical name of variable. Can you change it to sth like "sticky notes" ?

philemone: Yeah, it is going to look nicer - changed

designer/client/src/actions/notificationActions.tsx
```diff
@@ -12,10 +12,11 @@ export function success(message: string): Action {
     });
 }
 
-export function error(message: string): Action {
+//TODO please take a look at this method and my changes, am I wrong or was it incomplete (without `error` and `showErrorText`) and had incomplete logic
+export function error(message: string, error?: string, showErrorText?: boolean): Action {
```
gskrobisz: Not sure what happens here. Why the logic is incomplete? What is the difference between message and error?

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/StickyNotesEntityFactory.scala
```diff
@@ -0,0 +1,112 @@
...
+  val processVersionsTable: LTableQuery[ProcessVersionEntityFactory#ProcessVersionEntity]
+
+  class StickyNotesEntity(tag: Tag) extends Table[StickyNoteEventEntityData](tag, "sticky_notes") {
+
+    def id                = column[StickyNoteId]("id", O.PrimaryKey, O.AutoInc)
+    def noteCorrelationId = column[StickyNoteCorrelationId]("note_correlation_id", NotNull)
+    def content           = column[String]("content", NotNull)
+    def layoutData        = column[LayoutData]("layout_data", NotNull)
+    def color             = column[String]("color", NotNull)
+    def dimensions        = column[Dimensions]("dimensions", NotNull)
```
gskrobisz, line 34: `Dimensions` looks like property of `LayoutData`. Is it on purpose to keep it separated?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StickyNotesApiHttpService.scala
```diff
@@ -0,0 +1,230 @@
...
+  expose {
+    endpoints.stickyNotesDeleteEndpoint
+      .serverSecurityLogic(authorizeKnownUser[StickyNotesError])
+      .serverLogicEitherT { implicit loggedUser =>
+        { case (scenarioName, noteId) =>
+          for {
+            scenarioId      <- getScenarioIdByName(scenarioName)
+            _               <- isAuthorized(scenarioId, Permission.Write)
+            processActivity <- deleteStickyNote(noteId)
+          } yield processActivity.toInt
```
gskrobisz, line 101: processActivity is already an Int

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/StickyNotesEntityFactory.scala
philemone: Yes, `dimensions` is specific for stickyNotes and `LayoutData` is used among nodes as well

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/StickyNotesApiHttpService.scala
philemone: That is funny but I've added it to prevent 
```
[error]...
StickyNotesApiHttpService.scala:102:19: a pure expression does nothing in statement position
[error]           } yield processActivity
```  
When you remove `.toInt` code does not compile (idea see no problem though). Scala gives us warning and we must be handling it as error :shrug:

designer/client/src/actions/notificationActions.tsx
philemone: Before my update it was like this:
```
export function error(message: string): Action {
    return Notifications.error({
        autoDismiss: 10,
        children: <Notification type={"error"} icon={<InfoOutlinedIcon />} message={message} />,
    });
}
```

but in HttpService you can find definition like this:
```
type NotificationActions = {
    ...
    error(message: string, error: string, showErrorText: boolean): void;
};
```

I was surprised that when I executed
`error("msg", "error", true)` then "msg" was shown to user. 

I've tracked it to its implementation - I've added fields I thought were missing and it worked, now you can choose if you want to show message or error, before my change I think it was impossible and flag `showErrorText` was omitted 

I might be wrong, that is why I left TODO, but for now I stand that is was bug :)

gskrobisz: With new implementation true/false flag selects which parameter is displayed, you can simplify to either `error("msg")` or `error("error")`. Please check usages of `HttpService.#addError` with showErrorTest = true because you are changing displayed message: raw error instead of user friendly message e.g. Failed to deploy, Failed to migrate

designer/client/src/common/StickyNote.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type Dimensions = { width: number; height: number };
+
+export interface StickyNote {
+    id?: string;
+    noteId: number;
+    content: string;
+    layoutData: LayoutData;
+    dimensions: Dimensions;
+    color: string;
```
bartektartanus: More strict type?
https://medium.com/@steve.alves2/how-to-type-hex-colors-in-typescript-3c3b9a32baa7

designer/client/src/common/StickyNote.ts
```diff
@@ -0,0 +1,15 @@
...
+    targetEdge?: string;
```
bartektartanus, line 13: Is this used in UI?

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Creator toolbar should allow collapse (persist) and filtering #0.png
bartektartanus, line 1: Maybe add some cypress test for creating a sticky note?

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/StickyNotesApiHttpServiceBusinessSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+        }
+        .when()
+        .basicAuthAllPermUser()
+        .get(s"$nuDesignerHttpAddress/api/processes/$exampleScenarioName/stickyNotes?scenarioVersionId=1")
+        .Then()
+        .statusCode(200)
+        .equalsJsonBody("[]")
+    }
+
+    // TODO more tests
```
bartektartanus: Currently tests are for "negative" cases (absence of note). Add more for positive/happy-path cases ;)

designer/client/src/reducers/graph/utils.ts
```diff
@@ -77,6 +79,33 @@ export function prepareNewNodesWithLayout(
     };
 }
 
+export function removeStickyNoteFromLayout(state: GraphState, stickyNoteId: number): { layout: NodePosition[]; stickyNotes: StickyNote[] } {
+    const { layout } = state;
+    const stickyNoteLayoutId = createStickyNoteId(stickyNoteId);
+    const updatedStickyNotes = state.stickyNotes.filter((n) => n.noteId !== stickyNoteId);
+    const updatedLayout = updatedStickyNotes.map((stickyNote) => {
+        return { id: createStickyNoteId(stickyNote.noteId), position: stickyNote.layoutData };
```
bartektartanus: Why are you recreating this stickyNoteId? `noteId` hasn't changed, so why are you regenerating `id` with `createStickyNoteId`?

designer/client/src/common/StickyNote.ts
philemone: good :eye: 
We plan to use it in next iteration so to avoid migrations on BE side we added it all together to FE / BE - just to fill it in in next MRs.

Dzuming:
@philemone 

1. WDYT about changing a cursor during sticky note resize to [se-resize](https://www.w3schools.com/cssref/tryit.php?filename=trycss_cursor)? It will match the node details window resizing the cursor
2. Is it possible to delete sticky notes on backspace click, when focused?
3. Is it a business requirement not to let the user add/remove sticky notes when the scenario is not saved? Maybe we can inform a user somehow, why this sticky notes option is disabled in the toolbar?
4. Could you change the text color in case of a warning message to
```
 standardWarning: ({ theme }) => ({
                            backgroundColor: theme.palette.warning.main,
                            color: blendDarken(theme.palette.text.primary, 0.9),
                        }),
```
? IMO this text contrast will fit better

Before:
image

After:
image

designer/client/package-lock.json
```diff
@@ -29,6 +29,7 @@
                 "d3-transition": "3.0.1",
                 "d3-zoom": "3.0.0",
                 "dagre": "0.8.5",
+                "dompurify": "3.2.0",
```
Dzuming: Is it something redundant? It's not mentioned in `package.json` dependencies and no usage in the code

designer/client/src/actions/nk/process.ts
```diff
@@ -56,6 +60,45 @@ export function displayTestCapabilities(processName: ProcessName, scenarioGraph:
...
+};
+
+export function fetchStickyNotesForScenario(scenarioName: string, scenarioVersionId: number): ThunkAction {
+    return (dispatch) => refreshStickyNotes(dispatch, scenarioName, scenarioVersionId);
+}
+
+export function stickyNoteUpdated(scenarioName: string, scenarioVersionId: number, stickyNote: StickyNote): ThunkAction {
+    return (dispatch) => {
+        HttpService.updateStickyNote(scenarioName, scenarioVersionId, stickyNote).then((_) => {
+            refreshStickyNotes(dispatch, scenarioName, scenarioVersionId);
```
Dzuming, line 79: Do we need this to get sticky notes call on update? The FE state should be up to date

designer/client/src/actions/notificationActions.tsx
```diff
@@ -12,7 +13,10 @@ export function success(message: string): Action {
     });
 }
 
-export function error(message: string): Action {
+// TODO This method is incomplete. I've added error and showErrorText params to show what is missing.
```
Dzuming: I'm not sure what action should be performed in case of this TODO, but we have epic in [Jira](https://touk-jira.atlassian.net/browse/NU-1702) for error handling, maybe you can add a task for it with a description of the problem and get rid of this comment for now and remove unused params?

Dzuming (commented):
I left few comments however i didn't finish a review yet

designer/client/src/actions/notificationActions.tsx
philemone: If you are working on it then ok. I am still not sure it is good idea to leave it unfixed - I leave it though, hoping it is going to be made all right :)

designer/client/src/actions/nk/process.ts
philemone: Right now yes, we do not update frontend directly so change is not visible without refresh - when you finish edit then update request is send and updated version is loaded. It is simplified not to have any unsynced state

designer/client/package-lock.json
philemone: it is! Thanks, removed (I was using it before changed to lib that we already have)

philemone:
@Dzuming 
 
1. Changed as you suggested to se-resize—looks much better! 👍
2. Hmm, it is possible but would require some work. Additionally, there is no implemented "back/forward" action for sticky notes, so someone could delete a sticky note by mistake and be unable to recover it. I would suggest waiting for feedback from the business, if you don’t mind, before addressing this and other related issues—of course, only if you’re okay with that.
3. It’s a simplification of handling two separate entities in sync with each other. See, sticky notes do not generate new scenario versions—they create and assign new changes to an existing scenario version. For example, if you made some changes to a scenario (e.g., adding a new enricher), added a sticky note with details about the added enricher, and then left without saving the changes, you’d end up with a note that doesn’t correspond to anything. Of course, there are solutions to this, but for the first version, we had to prioritize what’s important. Anyway, I’m available if you’d like to talk it through! :)
4. You’re absolutely right—I’ve used a black font, and it’s much better now. Thanks! 👍

designer/client/src/reducers/graph/utils.ts
philemone: You are right, that was redundant - removed :+1:

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/StickyNotesApiHttpServiceBusinessSpec.scala
philemone: I've added some happy cases :)

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Creator toolbar should allow collapse (persist) and filtering #0.png
philemone: Eh, I was procrastinating those :no_mouth:  but since it is now out I cannot hide from it :sweat_smile:  Cypress tests added :+1:

designer/client/src/common/StickyNote.ts
philemone: Added :+1:

Dzuming:
> @Dzuming
> 
> 1. Changed as you suggested to se-resize—looks much better! 👍
> 2. Hmm, it is possible but would require some work. Additionally, there is no implemented "back/forward" action for sticky notes, so someone could delete a sticky note by mistake and be unable to recover it. I would suggest waiting for feedback from the business, if you don’t mind, before addressing this and other related issues—of course, only if you’re okay with that.
> 3. It’s a simplification of handling two separate entities in sync with each other. See, sticky notes do not generate new scenario versions—they create and assign new changes to an existing scenario version. For example, if you made some changes to a scenario (e.g., adding a new enricher), added a sticky note with details about the added enricher, and then left without saving the changes, you’d end up with a note that doesn’t correspond to anything. Of course, there are solutions to this, but for the first version, we had to prioritize what’s important. Anyway, I’m available if you’d like to talk it through! :)
> 4. You’re absolutely right—I’ve used a black font, and it’s much better now. Thanks! 👍

2. It's ok to wait for feedback, but IMO now we have some inconsistency between the behavior of elements in a diagram because there is such an option in the case of "scenario-based" nodes
3. What is the benefit of handling sticky notes differently than the rest of the scenario version-based data and breaking consistency? From the user's point of view, he is forced to save and create a new version of the scenario only to add a sticky note, and from the code of view, there is a lot of new logic around it. A similar use case was scenario labels, which are saved on scenario save.

designer/client/src/actions/notificationActions.tsx
Dzuming: > If you are working on it then ok.

Nah, I don't think I will work on it in the near future, I just wanted to point out that we have a Jira for this kind of case, and IMO it's a better place than TODO in a code. I have a feeling that no one cares about them (some time ago I tried to recognize some of them and had no idea if it's still the case or what's going on in case of specific todo), and a Jira with a description what is a current state, why do we want specific change and the proposed solution is a much better idea.

designer/client/src/actions/nk/process.ts
Dzuming: I'm not sure if I get it. IIUC:

1. We change the sticky note
2. Redux state is updated
3. There is an update call with an empty request
4. The sticky note is still in redux

Am I missing something? When can an unsynced state occur?

Dzuming: We discussed internally that this additional call is required

Dzuming:
> It’s a simplification of handling two separate entities in sync with each other. See, sticky notes do not generate new scenario versions—they create and assign new changes to an existing scenario version. For example, if you made some changes to a scenario (e.g., adding a new enricher), added a sticky note with details about the added enricher, and then left without saving the changes, you’d end up with a note that doesn’t correspond to anything. Of course, there are solutions to this, but for the first version, we had to prioritize what’s important. Anyway, I’m available if you’d like to talk it through! :)

We discussed it internally and decided to let users test it first.

Dzuming (approved):

bartektartanus (approved):

PR #7295, opened 2024-12-06
Fetch status of all scenarios in DM in a single operation instead of separately for each scenario.

mgoworko:
## Describe your changes

Problem to be solved:

When fetching list of scenarios or status of all scenarios, for periodic scenarios there were always 2N db queries for N periodic scenarios. Those 2N queries were performed each time and endpoint /processes or /processes/status was called.

Solution:

1. Defined capability of DeploymentManager (StateQueryForAllScenariosSupported). This capability means, that the DM can fetch all scenarios in a single operation.

2. The new capability is implemented for PeriodicDeploymentManager and for FlinkStreamingDeploymentManagerProvider (which is an underlying DM of PeriodicDeploymentManager)

3. When fetching processes or status, Nu uses the more efficient implementation of `StateQueryForAllScenariosSupported` when it is present (and it is present for periodics).

4. This improvement can be implemented for other DM's too, but it is outside of the scope of this task. Moreover, we only have performance issues for periodic DM ATM.

## Checklist before merge


engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
```diff
@@ -27,13 +29,70 @@ class FlinkStreamingDeploymentManagerProvider extends DeploymentManagerProvider
...
+  ): ValidatedNel[String, DeploymentManager with StateQueryForAllScenariosSupported] = {
     import dependencies._
     val flinkConfig = deploymentConfig.rootAs[FlinkConfig]
     FlinkClient.create(flinkConfig, scenarioStateCacheTTL).map { client =>
-      CachingProcessStateDeploymentManager.wrapWithCachingIfNeeded(
-        new FlinkStreamingRestManager(client, flinkConfig, modelData, dependencies),
-        scenarioStateCacheTTL
+      val underlying  = new FlinkStreamingRestManager(client, flinkConfig, modelData, dependencies)
+      val withCaching = CachingProcessStateDeploymentManager.wrapWithCachingIfNeeded(underlying, scenarioStateCacheTTL)
+      deploymentManagerWithStateQueryForAllScenariosSupported(withCaching, underlying)
```
arkadius: It looks strange, that we pass withCaching as underlying and in second channel underlying of this underlying. Maybe we could replace inheritance by composition as it is done with with `DeploymentSynchronisationSupport`, thanks to decorators would talk only to wrapped classes.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -59,17 +59,35 @@ trait DeploymentManager extends AutoCloseable {
...
         latestVersionId,
         deployedVersionId,
         currentlyPresentedVersionId,
-      ).map(state => statusDetailsWithFreshness.map(_ => state))
+        statusDetailsWithFreshness,
+      )
     } yield stateWithFreshness
   }
 
+  final def resolvePrefetchedProcessState(
```
arkadius: Let's keep this method close to the `resolve` method and keep the order of parameters similar (list of status details is in the other position). Also, I would remove "prefetched" part from method name. It is not important in this context. BTW take a look that from the perspective of DM, you can't be sure that  `getProcessStates` do some prefetching - you assumed here that DM supports prefetching but it is optional capability

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -446,38 +447,83 @@ class DeploymentService(
...
+        deploymentManagersByProcessingTypes.map { case (processingType, manager) =>
+          manager match {
+            case dm: StateQueryForAllScenariosSupported => getProcessesStates(processingType, dm)
+            case _                                      => Future.successful(None)
+          }
+        }
+      }
+      .map(_.flatten.toMap)
+
+    prefetchedStatesByProcessingTypes
```
arkadius: Let's inline this

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -446,38 +447,83 @@ class DeploymentService(
...
       freshnessPolicy: DataFreshnessPolicy
   ): Future[F[ScenarioWithDetails]] = {
+    val scenarios = processTraverse.toList
     dbioRunner.run(
       for {
-        actionsInProgress <- getInProgressActionTypesForProcessTraverse(processTraverse)
+        actionsInProgress <- getInProgressActionTypesForScenarios(scenarios)
+        // DeploymentManager's may support fetching state of all scenarios at once (StateQueryForAllScenariosSupported capability)
+        // For those DM's we prefetch the state and then use it for resolving individual scenarios.
+        prefetchedStates <- DBIO.from(getPrefetchedStatesForSupportedManagers(scenarios))
```
arkadius: `F` is sometimes the `Id` (this method is used in context of fetching one scenario status. To fetch status of one scenario you ask all deployment managers in every processing type for a status of every scenario. Is the intention of this change? :)

engine/flink/management/periodic/src/test/scala/pl/touk/nussknacker/engine/management/periodic/DeploymentManagerStub.scala
```diff
@@ -47,7 +80,13 @@ class DeploymentManagerStub extends BaseDeploymentManager with StubbingCommands
   override def getProcessStates(
       name: ProcessName
   )(implicit freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[List[StatusDetails]]] = {
-    Future.successful(WithDataFreshnessStatus.fresh(jobStatus.toList))
+    Future.successful(WithDataFreshnessStatus.fresh(jobStatus.get(name).toList.flatten))
+  }
+
+  override def getProcessesStates()(
```
arkadius: I would call it `getAllProcessesStates()` to emphasize that it could be expensive

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
```diff
@@ -33,39 +34,57 @@ class FlinkRestManager(
   override def getProcessStates(
       name: ProcessName
   )(implicit freshnessPolicy: DataFreshnessPolicy): Future[WithDataFreshnessStatus[List[StatusDetails]]] = {
-    val preparedName = modelData.namingStrategy.prepareName(name.value)
+    getProcessesStates()
```
arkadius: 1. We have fetching of all scenario states on two levels: in `DeploymentService` and in `FlinkRestManager`. It is quite tricky because when you analyze `DeploymentService` code you see some `None` for `prefetchedStatusDetailsWithFreshness`, you think that sth is done differently but eventually you figure out that it doesn't matter because in the deepest level `getProcessesStates` is invoked for both cases
2. We use "process stateS" in two context: when we get statuses of all jobs for given scenario and when we get statuses of all jobs for all scenarios. We need to distinguish it because it is hard to figure out what's going on
3. Do we still need the cache on FlinkClient level and/or on DM level?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
mgoworko: - State was only pre-fetched from DM's resposible for scenarios, that were fetched. So if the query was only about periodic scenarios, then only PeriodicDM was queried about all states
- I made a bold assumption, that fetching all states is not heavier than fetching a single on. It is true for Flink and Periodic, but may not be true in general
- I modified the implementation, to only pre-fetch statuses from DM, if there is more than one scenario handled by that DM in the current query.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
mgoworko: done

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
mgoworko: Method resolvePrefetchedProcessState removed. It was not needed, because simple resolve can be used.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkStreamingDeploymentManagerProvider.scala
mgoworko: Done with composition

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkRestManager.scala
mgoworko: 1. In DeploymentService, I split it into 2 methods, dedicated to those 2 cases: `getProcessStateFetchingStatusFromManager` and `getProcessStateUsingPrefetchedStatus`, without Options in signatures
2. In Nu we use names:
- `getProcessState` for fetching `ProcessState`
- `getProcessStates` for fetching `List[StatusDetails]`

Method in DeploymentService renamed `getProcessesStates` -> `getAllProcessesStates`, like the one in `StateQueryForAllScenariosSupported`. Is there anything more to rename in his PR?

3.
- IMO we can't remove caching on DM level, because all DM's use it, and this improvement is limited to Flink and Periodic.
- It may be possible to now remove cache on flink client level, but it should be IMO done outside this PR, separately, after testing the behavior without cache

arkadius: Ad. 1. Looks better
Ad. 2. No, it's more clear now
Ad. 3. Let's remove client level cache in some next PR

arkadius (approved):
LGTM

PR #7335, opened 2024-12-13
[NU-1777] Separate classloader for deployment managers

coutoPL:
## Describe your changes

- deployment managers have a separate class loader (the Application ClassLoader is no longer used to load their classe
- deployment manager class loader is a parent for the model class loader
- you can configure paths for deployment managers using the `managersDirs` config setting
- removed unused env used at the SBT level (`includeFlinkAndScala`)
- new E2E test (example R-R scenario)

## Checklist before merge


build.sbt
```diff
@@ -438,7 +456,8 @@ def assemblySettings(
     includeScala: Boolean,
     filterProvidedDeps: Boolean = true
 ): List[Def.SettingsDefinition] = {
-  // This work around need to be optional because for designer module it causes excluding of scala lib (because we has there other work around for Idea classpath and provided deps)
+  // This work around need to be optional because for designer module it causes excluding of scala lib
+  // (because we have there other work around for Idea classpath and provided deps)
```
arkadius, line 439: Is it still relevant?

build.sbt
```diff
@@ -2272,9 +2289,13 @@ prepareDev := {
...
   val workTarget = (designer / baseDirectory).value / "work"
   val artifacts  =
-    (distribution / componentArtifacts).value ++ (distribution / devArtifacts).value ++ developmentTestsDeployManagerArtifacts.value ++
+    (distribution / componentArtifacts).value ++
+      (distribution / devArtifacts).value ++
+      developmentTestsDeployManagerArtifacts.value ++
       Def
-        .taskDyn(if (addManagerArtifacts) distribution / managerArtifacts else Def.task[List[(File, String)]](Nil))
+        .taskDyn(
+          if (addManagerArtifacts) distribution / deploymentManagerArtifacts else Def.task[List[(File, String)]](Nil)
```
arkadius: Won't we build some artefacts twice? I see that developmentTestsDeployManagerArtifacts contains also production artifacts now. Or sbt deduplicate them?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -1501,7 +1501,7 @@ object NodesApiEndpoints {
 
 }
 
-object TypingDtoSchemas { // todo
+object TypingDtoSchemas { // TODO:
```
arkadius: Let's remove this :)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
```diff
@@ -125,8 +128,6 @@ case class ClassLoaderModelData private (
     override val additionalConfigsFromProvider: Map[DesignerWideComponentId, ComponentAdditionalConfig],
     // This property is for easier testing when for some reason, some jars with ComponentProvider are
     // on the test classpath and CPs collide with other once with the same name.
-    // E.g. we add liteEmbeddedDeploymentManager as a designer provided dependency which also
```
arkadius, line 131: I would remove these functions now and whole comment block

build.sbt
```diff
@@ -1958,24 +1980,22 @@ lazy val designer = (project in file("designer/server"))
         CopyOptions.apply(overwrite = true, preserveLastModified = true, preserveExecutable = false)
       )
     },
+    prepareDesignerTests             := {
```
arkadius, line 1956: Why did you merge dependencies for both test scopes? Doesn't it slow down test on CI?

build.sbt
```diff
@@ -2036,6 +2057,8 @@ lazy val designer = (project in file("designer/server"))
         "io.circe"                      %% "circe-yaml"                      % circeYamlV           % Test,
         "com.github.scopt"              %% "scopt"                           % "4.1.0"              % Test,
         "org.questdb"                    % "questdb"                         % "7.4.2",
+        "org.apache.kafka"               % "kafka-clients"                   % kafkaV,
```
arkadius: What's going on here?

build.sbt
```diff
@@ -2051,24 +2074,18 @@ lazy val designer = (project in file("designer/server"))
     processReports,
     security,
     deploymentManagerApi,
+    componentsApi,
+    requestResponseComponentsApi,
```
arkadius: and here

coutoPL: Designer and Runtime should define the API of components they are supposed to work with. And with these dependencies we express it.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
coutoPL: I wanted to remove it but there is still some problem with ProcessConfigCreator implementations. Nevertheless, the lite Embedded deployment manager is not the problem anymore.

build.sbt
arkadius: Yeah, but I thought that componentsApi is only necessary. These below have some engine-specific agreement that designer shouldn't be aware of

build.sbt
coutoPL: TBH I'm not sure what exact problem was solved this way and how to check if my change outdated it.

build.sbt
coutoPL: TBH this dependency should be added at the component definition level (in the module when the kafka-related component/s is/are added). For some reason, there is no such dependency in the lite kafka component module (or one of the lite-related kafka modules). It's hard for me to unravel the relationships between these modules, so I decided to put the dependency here as well. 

Screenshot 2024-12-16 at 23 07 20

build.sbt
coutoPL: Yes, so do I. But I got this when we don't add it:

Screenshot 2024-12-16 at 23 28 35

I wonder if engine<->designer separation is as good as the models/managers<->desinger one?

build.sbt
arkadius: I see that `false` was passed only for designer assembly. It looks like we have a fat jar for designer. I don't know if someone uses it :) but ... The only provided dependencies of designer where DMs so the default `filterProvidedDeps = true` should work the same so we can remove this condition

build.sbt
mk-software-pl: the dependency was removed

build.sbt
mk-software-pl: the dependencies were removed

designer/client/docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
       KAFKA_AUTO_OFFSET_RESET: "earliest"
       SCHEMA_REGISTRY_URL: http://redpanda:8081
       CONFIG_FILE: /opt/nussknacker/conf/dev-application.conf
+      MANAGERS_DIR: /opt/nussknacker/managers
```
arkadius: Won't it be picked by default? similar to `STORAGE_DIR` and others?

build.sbt
arkadius: nice

arkadius (approved):
LGTM

PR #7336, opened 2024-12-16
[NU-1828] Processing types configs loader api

arkadius:
## Checklist before merge


designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/DesignerRootConfig.scala
```diff
@@ -0,0 +1,16 @@
+package pl.touk.nussknacker.ui.configloader
+
+import com.typesafe.config.Config
+import pl.touk.nussknacker.engine.ConfigWithUnresolvedVersion
+
+// TODO: We should extract a class for all configuration options that should be available to designer instead of returning raw hocon config.
+//       Thanks to that it will be easier to split processing type config from rest of configs and use this interface programmatically
+final case class DesignerRootConfig(rawConfig: ConfigWithUnresolvedVersion)
```
coutoPL: 1. private constructor? 
2. why not just DesignerConfig? TBH it'd be nice to have ProcessingTypesConfig too.

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoader.scala
```diff
@@ -0,0 +1,25 @@
...
+
+trait ProcessingTypeConfigsLoader {
+
+  def loadProcessingTypeConfigs(): IO[Map[String, ProcessingTypeConfig]]
+
+}
+
+object ProcessingTypeConfigsLoader extends LazyLogging {
+
+  def extractProcessingTypeConfigs(rootConfig: DesignerRootConfig): Map[String, ProcessingTypeConfig] = {
```
coutoPL: it looks like a good candidate for a method of DesignerRootConfig class (or an extension method for this class) or the ProcessingTypesConfig class

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
```diff
@@ -0,0 +1,15 @@
...
+import sttp.client3.SttpBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ProcessingTypeConfigsLoaderFactory {
+
+  def create(
+      designerRootConfigLoadedAtStart: DesignerRootConfig,
+      sttpBackend: SttpBackend[Future, Any],
+      ec: ExecutionContext
```
coutoPL: maybe we can pass it implicitly?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/DictApiEndpoints.scala
```diff
@@ -15,9 +15,8 @@ import pl.touk.nussknacker.ui.api.description.DictApiEndpoints.DictError.{
 }
 import pl.touk.nussknacker.ui.api.description.DictApiEndpoints.Dtos._
 import sttp.model.StatusCode.{BadRequest, NotFound, Ok}
-import sttp.tapir._
+import sttp.tapir.{Schema, _}
```
coutoPL: `import sttp.tapir._`

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/EachTimeLoadingRootConfigProcessingTypeConfigsLoader.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import cats.effect.IO
+import pl.touk.nussknacker.engine.ProcessingTypeConfig
+import pl.touk.nussknacker.ui.config.root.DesignerRootConfigLoader
+import pl.touk.nussknacker.ui.configloader.ProcessingTypeConfigsLoader
+
+class EachTimeLoadingRootConfigProcessingTypeConfigsLoader(designerRootConfigLoader: DesignerRootConfigLoader)
+    extends ProcessingTypeConfigsLoader {
+
+  def loadProcessingTypeConfigs(): IO[Map[String, ProcessingTypeConfig]] =
```
coutoPL: let's add `override`

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
```diff
@@ -14,27 +21,31 @@ import pl.touk.nussknacker.engine.util.config.ConfigFactoryExt
   * Result of config loading still keep version with unresolved env variables for purpose of config loading on model side - see
   * InputConfigDuringExecution and ModelConfigLoader
   */
-object DesignerConfigLoader {
+private[root] class DesignerRootConfigLoaderImpl(classLoader: ClassLoader) extends DesignerRootConfigLoader {
```
coutoPL: Let's get rid of the interface. This implementation name suggests we don't need it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
```diff
@@ -1,9 +1,16 @@
-package pl.touk.nussknacker.ui.config
+package pl.touk.nussknacker.ui.config.root
```
coutoPL: Do we need a separate package for this file?

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerAppFactory.scala
```diff
@@ -8,56 +8,76 @@ import io.dropwizard.metrics5.MetricRegistry
...
+import pl.touk.nussknacker.ui.process.processingtype.loader.{
+  ProcessingTypeDataLoader,
+  ProcessingTypesConfigBasedProcessingTypeDataLoader
+}
 import pl.touk.nussknacker.ui.server.{AkkaHttpBasedRouteProvider, NussknackerHttpServer}
+import pl.touk.nussknacker.ui.util.ActorSystemBasedExecutionContextWithIORuntime
+import sttp.client3.asynchttpclient.future.AsyncHttpClientFutureBackend
 
 import java.time.Clock
+import scala.concurrent.ExecutionContext
```
coutoPL: unused

designer/server/src/main/scala/pl/touk/nussknacker/ui/LocalNussknackerWithSingleModel.scala
```diff
@@ -49,8 +49,12 @@ object LocalNussknackerWithSingleModel {
       modelData = Map(typeName -> (category, modelData)),
       deploymentManagerProvider = deploymentManagerProvider
     )
-    val nussknackerConfig = new LoadableConfigBasedNussknackerConfig(IO.delay(DesignerConfigLoader.from(appConfig)))
-    val appFactory        = new NussknackerAppFactory(nussknackerConfig, local)
+    val designerConfigLoader = DesignerRootConfigLoader.fromConfig(appConfig)
+    val appFactory = new NussknackerAppFactory(
```
coutoPL, line 56: I wander if we can make the constructor private and handle this case using specialized (and simplified) smart constructor?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
```diff
@@ -14,27 +21,31 @@ import pl.touk.nussknacker.engine.util.config.ConfigFactoryExt
...
     for {
       baseConfig   <- IO.blocking(ConfigFactoryExt.parseUnresolved(classLoader = classLoader))
-      loadedConfig <- load(baseConfig, classLoader)
-    } yield loadedConfig
+      loadedConfig <- load(baseConfig)
+    } yield DesignerRootConfig(loadedConfig)
   }
 
-  def load(baseUnresolvedConfig: Config, classLoader: ClassLoader): IO[ConfigWithUnresolvedVersion] = {
+  def load(baseUnresolvedConfig: Config): IO[ConfigWithUnresolvedVersion] = {
```
coutoPL: I see, this method is left public only for test purposes (ConfigurationTest). Do we have to test it like that?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
```diff
@@ -6,64 +6,76 @@ import pl.touk.nussknacker.engine._
...
-        // to assert the loaded configuration is correct (fail-fast approach).
-        val combinedData = CombinedProcessingTypeData.create(processingTypesData)
+object ProcessingTypesConfigBasedProcessingTypeDataLoader extends LazyLogging {
 
-        ProcessingTypeDataState(
-          processingTypesData.mapValuesNow(toValueWithRestriction),
-          () => combinedData,
-          // We pass here new Object to enforce update of observers
-          new Object
+  def loadProcessingTypeData(
```
coutoPL: I don't like this move. At the moment we have this method only for test purposes. We could achieve the same with the previous implementation and a dummy ProcessingTypeConfigsLoader implementation passed to the constructor. Like the one:

```
class Static...Loader(processingTypesConfigs: Map[String, ProcessingTypeConfig]) extends ProcessingTypeConfigsLoader {
  override def loadProcessingTypeConfigs(): IO[Map[String, ProcessingTypeConfig]] = IO.pure(processingTypesConfigs)
}
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -713,28 +705,27 @@ class AkkaHttpBasedRouteProvider(
...
-              getDeploymentManagerDependencies(
-                additionalUIConfigProvider,
-                actionServiceProvider,
-                scenarioActivityRepository,
-                dbioActionRunner,
-                sttpBackend,
-                _
-              ),
-            )
+        acquire = IO.pure {
```
coutoPL: let's use IO.apply or IO.defer here instead (to make sure potential exceptions won't leave the IO scope)

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import sttp.client3.SttpBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ProcessingTypeConfigsLoaderFactory {
+
+  def create(
+      designerRootConfigLoadedAtStart: DesignerRootConfig,
+      sttpBackend: SttpBackend[Future, Any],
```
coutoPL: In the `ProcessingTypeConfigsLoader` trait we use IO as an effect. So, maybe we can use IO effect also in the SttpBackend?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/EachTimeLoadingRootConfigProcessingTypeConfigsLoader.scala
```diff
@@ -0,0 +1,16 @@
+package pl.touk.nussknacker.ui.config.processingtype
+
+import cats.effect.IO
+import pl.touk.nussknacker.engine.ProcessingTypeConfig
+import pl.touk.nussknacker.ui.config.root.DesignerRootConfigLoader
+import pl.touk.nussknacker.ui.configloader.ProcessingTypeConfigsLoader
+
+class EachTimeLoadingRootConfigProcessingTypeConfigsLoader(designerRootConfigLoader: DesignerRootConfigLoader)
```
coutoPL: TBH, I'm not sure we should allow passing any DesignerRootConfigLoader impl here. Currently, the name is a little lie (;)) because the config reloading depends on the DesignerRootConfigLoader impl.

I wrote about sth related [here](https://gitlab.touk.pl/nussknacker-cloud/config-loader-designer-plugin/-/merge_requests/1#note_493685). It seems that this should be a "loader that is able to load processing types from the local hocon file". It means that we should:
1. change the name
2. use DesignerRootConfigLoaderImpl directly

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/ProcessingTypeConfigsLoaderFactoryServiceLoader.scala
```diff
@@ -0,0 +1,35 @@
...
+  ProcessingTypeConfigsLoader,
+  ProcessingTypeConfigsLoaderFactory
+}
+import sttp.client3.SttpBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ProcessingTypeConfigsLoaderFactoryServiceLoader extends LazyLogging {
+
+  def loadService(designerRootConfigLoader: DesignerRootConfigLoader): ProcessingTypeConfigsLoaderFactory = {
```
coutoPL: maybe we don't have to pass any param and return None when there is no ProcessingTypeConfigsLoaderFactory to load using SPI? And the fallback would be done in the place when this method is called?

coutoPL:
let's add the jira no in the title of the PR 
> * [ ]  Related issue ID is placed at the beginning of PR title in [brackets] (can be GH issue or Nu Jira issue)

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/DesignerRootConfig.scala
arkadius: 1. ok
2. when I read this code I found it difficult to understand which config or view of config is passed around, but maybe now it is a little bit easier, I'll remove this part

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
arkadius: Take a look at `DesignerRootConfigLoader.fromConfig`. After resigning from the interface, it won't be as easy to implement this method as it is now.

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
arkadius: I can move it to the `config` package.

designer/server/src/main/scala/pl/touk/nussknacker/ui/LocalNussknackerWithSingleModel.scala
arkadius: I'm not sure if I understand you, but moving this logic into the `NussknackerAppFactory` constructor would split the logic of creating the application with one model into two places. It breaks SRP rule. But maybe I misunderstand this comment.

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
arkadius: What other option do you see? We can extract another class which will be invoked by this, but I don't think that it would increase readability.

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
coutoPL: oh, anonymous class :D I missed it. But IMO it'd be better to have two well-named implementations:
1. one instance always reads na parses file config
2. the second instance is kind of adapter - it adapts a provided Config (does the by-name param matter? Do we use it somewhere?) to the DesignerRootConfigLoader interface

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
coutoPL: The other option I see is modifying the ConfigurationTest and testing the public method, not the public-like one.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
arkadius: Hmm for me, if you feel that some code is a nice entrypoint to unit tests, you should prefer to extract this code to some separate, testable method instead of stubbing things that you don't want to test. For example. Let's say that you have a calculator which some GUI for it. In this case you should rather focus on testing logic of functions provided by the calculator instead of ensuring that every test will go through the whole graphical interface. Do you have another opinion on that?

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
arkadius: In major part of application we use `Future`, especially when we provide sttp backend (see `AdditionalUIConfigProviderFactory`, `AuthenticationProvider`) but maybe it's time to switch to IO incrementally :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
coutoPL: as disscussed

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/DesignerRootConfig.scala
arkadius: 2. We already have `ProcessingTypeConfig`. Do you mean `ProcessingTypeConfigs(configByProcessingType: Map[String, ProcessingTypeConfig])`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
arkadius: 2. Yes, this call by-name matters in `AppApiHttpServiceBusinessSpec` > `reload processing types-related model data when` where you change designer config based on `simulateChangeInApplicationConfig` flag - it is your test! :) 

Changes that I introduced so far:
- extracted named classes as we discussed
- named more explicitly these classes (try as I might ;) )
- (bonus) refactored a bit responsibility of  `DesignerConfigLoader`/`ConfigFactoryExt`. Now this business logic described in method scaladoc is in one place

WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/root/DesignerRootConfigLoader.scala
arkadius: I test public method now. In order to do that I needed to change the global state of jvm. I hope that it won't introduce flaky tests. Take a look.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
arkadius: I replaced this method by passing stub into the public method and invocation of `unsafeRunSync` on  the result as we discussed

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/EachTimeLoadingRootConfigProcessingTypeConfigsLoader.scala
arkadius: I thought about that as an explicit, named class for an implementation that just calls underlying   `DesignerRootConfigLoader.loadDesignerConfig` (no matter what it does) and transform the result. I don't think that calling `DesignerRootConfigLoaderImpl` directly is a good approach as it is against IoC approach. 

I replaced this class by inline usage of:
```
() => designerConfigLoader.loadDesignerConfig().map(_.processingTypeConfigs)
```

Is it better or I misunderstand you?

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/ProcessingTypeConfigsLoaderFactoryServiceLoader.scala
arkadius: changed

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
arkadius: changed, I hope I correctly handled `IO` to `Future` conversion for the purpose of other places

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
```diff
@@ -0,0 +1,16 @@
...
+import sttp.client3.SttpBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ProcessingTypeConfigsLoaderFactory {
+
+  def create(
+      configLoaderConfig: Config,
+      sttpBackend: SttpBackend[IO, Any],
+  )(implicit ec: ExecutionContext): ProcessingTypeConfigsLoader
```
coutoPL: let's change ExecutionContext to Runtime (as a part of Future to IO migration)

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/DesignerConfigLoader.scala
```diff
@@ -14,27 +21,39 @@ import pl.touk.nussknacker.engine.util.config.ConfigFactoryExt
...
+ */
+class SimpleConfigLoadingDesignerConfigLoader(loadConfig: => Config) extends DesignerConfigLoader {
+
+  override def loadDesignerConfig(): IO[DesignerConfig] = IO.delay(DesignerConfig.from(loadConfig))
+
+}
+
+object DesignerConfigLoader {
+
+  def apply(classLoader: ClassLoader): AlwaysLoadingFileBasedDesignerConfigLoader =
```
coutoPL: I wonder if we still need these smart constructors. 
Eg.
```
appFactory <- IO(NussknackerAppFactory(DesignerConfigLoader(getClass.getClassLoader)))
```
is less readable than:
```
appFactory <- IO(NussknackerAppFactory(AlwaysLoadingFileBasedDesignerConfigLoader(getClass.getClassLoader)))
```

(in a sense of context understanding)

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerAppFactory.scala
```diff
@@ -3,61 +3,98 @@ package pl.touk.nussknacker.ui.factory
...
+      executionContextWithIORuntime: ActorSystemBasedExecutionContextWithIORuntime,
+      sttpBackend: SttpBackend[IO, Any]
+  ): ProcessingTypeConfigsLoader = {
+    processingTypeConfigsLoaderFactoryOpt
+      .map { factory =>
+        logger.debug(
+          s"Found custom ${classOf[ProcessingTypeConfigsLoaderFactory].getSimpleName}: ${factory.getClass.getName}. Using it for configuration loading"
+        )
+        factory.create(
+          designerConfig.rawConfig.resolved.getAs[Config]("configLoader").getOrElse(ConfigFactory.empty()),
```
coutoPL: WDYT about adding `def configLoaderConfig: Config` (yeah, sounds strange ... can be `loaderConfig`) method to `DesignerConfig`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/IOToFutureSttpBackendConverter.scala
```diff
@@ -0,0 +1,27 @@
...
+  def convert(
+      ioBackend: SttpBackend[IO, Any]
+  )(implicit executionContextWithIORuntime: ExecutionContextWithIORuntime): SttpBackend[Future, Any] = {
+    import executionContextWithIORuntime.ioRuntime
+    ioBackend.mapK(
+      new FunctionK[IO, Future] {
+        override def apply[A](io: IO[A]): Future[A] = io.unsafeToFuture()
+      },
+      new FunctionK[Future, IO] {
+        override def apply[A](future: Future[A]): IO[A] = IO.fromFuture(IO.pure(future))
```
coutoPL, line 22: This is suspicious. The future passed here will be already started. Probably some IO rules are broken. OTOH, there is not much room to implement it differently. 

I agree it's a good to start replacing SttpBackend[Future, ...] with SttpBackend[IO, ...].

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/DesignerRootConfig.scala
coutoPL: yeah. Sorry. Do you think it's worth to introduce it to encapsulate the map?

designer/server/src/main/scala/pl/touk/nussknacker/ui/LocalNussknackerWithSingleModel.scala
coutoPL: There is something alarming in this constructor:
```
class NussknackerAppFactory(
    designerConfigLoader: DesignerConfigLoader,
    processingTypeConfigsLoaderFactoryOpt: Option[ProcessingTypeConfigsLoaderFactory],
    createProcessingTypeDataLoader: ProcessingTypeConfigsLoader => ProcessingTypeDataLoader
) 
```
It's not so easy to reason about. But let's move forward - we will try to do sth with it in the future.

designer/server/src/main/scala/pl/touk/nussknacker/ui/config/processingtype/EachTimeLoadingRootConfigProcessingTypeConfigsLoader.scala
coutoPL: Yeah, it seems that, without the name, it's better because it's less misleading.

coutoPL (approved):
some minor comments left.
In general - LGTM

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/DesignerRootConfig.scala
arkadius: If we had a value class for processing type, IMO it wouldn't be worth it, but in the current situation it might make sense - added

designer/config-loader-api/src/main/scala/pl/touk/nussknacker/ui/configloader/ProcessingTypeConfigsLoaderFactory.scala
arkadius: Ah, good catch

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerAppFactory.scala
arkadius: Good idea. It's follow the approach that I described in the TODO next to the class, added

designer/server/src/main/scala/pl/touk/nussknacker/ui/LocalNussknackerWithSingleModel.scala
arkadius: I found the solution that made it a little bit simpler, I moved factory SPI loading into `createApp` body.

PR #7354, opened 2024-12-17
Reduce response payload size for scenarios tab

Cichan-mcc:
## Describe your changes

Currently when we enter scenarios tab we fetch a lot of data regarding them. Some of the data isn't used at that level. It only bloats the response payload slowing down loading the tab.

In our Nussknacker instance we have a lot of scenarios so response payloads for that tab can reach around 7.6 MB. The biggest unused attribute that is sent along with the scenarios' details is `buildInfo`. We can remove it along with a few others not needed fields and we can also omit null attributes in the response JSON. By doing so we can decrease the response payload up to 25%. Which can help us a bit with long load times of scenarios tab.

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioWithDetailsConversions.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioWithDetailsConversions.scala
@@ -1,9 +1,11 @@
 package pl.touk.nussknacker.ui.process
 
+import pl.touk.nussknacker.engine.api.deployment.ProcessAction
 import pl.touk.nussknacker.engine.api.graph.ScenarioGraph
 import pl.touk.nussknacker.engine.api.process.ProcessId
 import pl.touk.nussknacker.restmodel.scenariodetails.{ScenarioParameters, ScenarioWithDetails}
 import pl.touk.nussknacker.restmodel.validation.ScenarioGraphWithValidationResult
+import pl.touk.nussknacker.ui.process.ProcessService.SkipAdditionalFields
 import pl.touk.nussknacker.ui.process.repository.ScenarioWithDetailsEntity
 
 object ScenarioWithDetailsConversions {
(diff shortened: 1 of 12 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioWithDetailsConversions.scala
```diff
@@ -55,6 +57,25 @@ object ScenarioWithDetailsConversions {
     )
   }
 
+  def skipAdditionalFields(details: ScenarioWithDetails, skipOptions: SkipAdditionalFields): ScenarioWithDetails = {
```
lciolecki, line 60: I would rather prefer do not fetch this data from the DB at all. On the other hand I'm not sure is it now possible?

Dzuming (approved):
FE LGTM

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ProcessesResourcesSpec.scala
```diff
@@ -129,6 +129,30 @@ class ProcessesResourcesSpec
     }
   }
 
+  test("/api/processes should return lighter details without ProcessAction's additional fields") {
+    createDeployedScenario(processName, category = Category1)
+    Get(s"/api/processes") ~> withReaderUser() ~> applicationRoute ~> check {
+      status shouldEqual StatusCodes.OK
+      val loadedProcess = responseAs[List[ScenarioWithDetails]]
+
+      loadedProcess.head.lastAction should matchPattern {
```
lciolecki: Shouldn't we test that field doesn't exist in a response? Now you testsing Optional, so there still can be passed null - as I understand.

Cichan-mcc: You're right I forgot to modify the test after removing nulls from the JSON response!
I modified the test to verify whether unneceseary fields are omitted and no nulls are present in JSON.

lciolecki (approved):

PR #7356, opened 2024-12-17
[NU-1932] Big decimal uses default scale

paw787878:
## Checklist before merge


utils/utils/src/main/java/pl/touk/nussknacker/springframework/util/NumberUtilsConsts.java
```diff
@@ -0,0 +1,5 @@
+package pl.touk.nussknacker.springframework.util;
+
+public class NumberUtilsConsts {
```
arkadius: Let's rename this class to some name that would explain why we do this trick e.g. `BigDecimalScaleEnsurer`, move all setScale invocations and write a notice that this is work around for spel's behaviour with an explanation how spel behaves. Basically let's move our online conversion to comment :)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Conversion.scala
```diff
@@ -68,11 +69,15 @@ object ToBigDecimalConversion extends ToNumericConversion[JBigDecimal] {
 
   override def convertEither(value: Any): Either[Throwable, JBigDecimal] =
     value match {
-      case v: JBigDecimal => Right(v)
-      case v: JBigInteger => Right(new JBigDecimal(v))
+      case v: JBigDecimal => Right(v.setScale(Math.max(v.scale(), NumberUtilsConsts.DEFAULT_BIG_DECIMAL_SCALE)))
```
lciolecki: Let's extract:

```
v.setScale(Math.max(v.scale(), NumberUtilsConsts.DEFAULT_BIG_DECIMAL_SCALE))
```

to separated method.

utils/utils/src/main/java/pl/touk/nussknacker/springframework/util/NumberUtilsConsts.java
paw787878: done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/Conversion.scala
paw787878: done

arkadius (commented):
Please add changelog entry and fix tests

arkadius (approved):

PR #7364, opened 2024-12-19
PeriodicDM as decorator for any DM with data stored in main Nu DB

mgoworko:
## Describe your changes

This PR is a huge refactor of the periodic DM.

What is done:
- The PeriodicDM is no longer a separate DM. Its modules are removed in code.
- Periodic DM is now a decorator, that can be applied to any (with implemented support, see point below) DeploymentManager, based on flag in config `supportsPeriodicExecution`
- Any DM may implement support for periodic execution - in that case, they must implement PeriodicDeploymentEngineHandler (which contains methods that handle deployment on a specific engine), right now implemented only for Flink.
- DB tables - datasource of PeriodicDM - are now moved to the main Nussknacker DB (previously were completely separated, with different schema and separate migration numbering, could be a completely different db). 
- The old datasource definition is marked as legacy. Can still be used (so it is not necessary to migrate DB immediately), but should be switched to new mechanism.

What was NOT done on purpose:
- The general structure and location of PeriodicDM is changed, but there are little changes in the internals of the DM. It is intentional. The structural changes introduced in this PR, like moving repo to core and making PeriodicDM a quasi-decorator, will allow future changes. But it would be way too much in this PR.
- The DB structure is not extensively refactored. There are some optimizations, like not storing full scenario JSON and allowing underlying DMs other than Flink, but overall tables are almost the same. They are now in main Nu db, so migrations can be introduced in future PRs

TODO in future PRs:
- rename Periodic => Schedule (or variants of that word) in designer modules (already changed in the API)
- add better support for migration from legacy db to new db tables

## Checklist before merge


designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -132,3 +137,25 @@ trait DeploymentSynchronisationSupported extends DeploymentSynchronisationSuppor
 }
 
 case object NoDeploymentSynchronisationSupport extends DeploymentSynchronisationSupport
+
+sealed trait PeriodicExecutionSupport
```
arkadius: 1. As this change introduces new API, IMO it is time to rethink naming. I'm for getting rid of `periodic` word usage and replacing it with `scheduler`/`scheduling`. Periodic is the very hermetic name for what we are trying to achieve. Everyone calls it scheduling. Especially user can configure one. delayed deployment so there nothing will be done periodically.
2. I feel confused what where ends  `PeriodicExecutionSupport` responsibility and starts `PeriodicDeploymentEngineHandler` responsibility. Am I right that handled is responsible for things related with scenario deployment and `PeriodicExecutionSupport` for scheduling itself?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -132,3 +137,25 @@ trait DeploymentSynchronisationSupported extends DeploymentSynchronisationSuppor
...
+
+trait PeriodicExecutionSupported extends PeriodicExecutionSupport {
+
+  def engineHandler(
+      modelData: BaseModelData,
+      dependencies: DeploymentManagerDependencies,
+      deploymentConfig: Config,
+  ): PeriodicDeploymentEngineHandler
```
arkadius: Let's not introduce new phrases such as `deployment engine`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
```diff
@@ -12,17 +12,15 @@ import scala.concurrent.duration._
   * @param rescheduleCheckInterval {@link RescheduleFinishedActor} check interval.
   * @param deployInterval {@link DeploymentActor} check interval.
   * @param deploymentRetry {@link DeploymentRetryConfig}  for deployment failure recovery.
-  * @param jarsDir Directory for jars storage.
   * @param maxFetchedPeriodicScenarioActivities Optional limit of number of latest periodic-related Scenario Activities that are returned by Periodic DM.
   */
 case class PeriodicBatchConfig(
-    db: Config,
+    db: Option[Config],
```
arkadius: Let's keep naming consistent and not introduce another `batch` word. Also let's add a TODO that we want to remove this db parameter after we migrate to the new approach. What about other configuration option? I guess that `processingType` will be also removed?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessesManager.scala
```diff
@@ -0,0 +1,103 @@
...
+import pl.touk.nussknacker.engine.api.deployment.periodic.model.DeploymentWithRuntimeParams
+import pl.touk.nussknacker.engine.api.process.{ProcessName, VersionId}
+import pl.touk.nussknacker.engine.canonicalgraph.CanonicalProcess
+import pl.touk.nussknacker.ui.process.periodic.model.PeriodicProcessDeploymentStatus.PeriodicProcessDeploymentStatus
+import pl.touk.nussknacker.ui.process.periodic.model._
+
+import java.time.LocalDateTime
+import scala.concurrent.Future
+
+trait PeriodicProcessesManagerProvider {
```
arkadius: Do we need to keep it abstract? Can't we have one implementation? Also, I feel  confused about responsibility of this class. We have repository, we have service, and this manager what exactly does? It looks like it only forward evething to repository. Why we need to have a provider instead of creating the implementing class directly

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
```diff
@@ -0,0 +1,74 @@
...
+      periodicExecutionSupported: PeriodicExecutionSupported,
+      periodicProcessesManagerProvider: PeriodicProcessesManagerProvider,
+      modelData: ModelData,
+      deploymentConfig: Config,
+      dependencies: DeploymentManagerDependencies,
+  ): DeploymentManager = {
+    logger.info("Decorating DM with periodic functionality")
+    import net.ceedubs.ficus.Ficus._
+    import net.ceedubs.ficus.readers.ArbitraryTypeReader._
+    val periodicBatchConfig = deploymentConfig.as[PeriodicBatchConfig]("deploymentManager")
```
arkadius: Can we have all scheduling related config in one field? E.g.:
```
scheduling {
  enabled: true
  ... other configuration options
}
```
? 
`deploymentManager` looks weird here as all `deploymentConfig` is a configuration of a deployment manager.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
```diff
@@ -0,0 +1,74 @@
...
+      schedulePropertyExtractorFactory = schedulePropertyExtractorFactory,
+      processConfigEnricherFactory = processConfigEnricherFactory,
+      listenerFactory = periodicProcessListenerFactory,
+      periodicBatchConfig = periodicBatchConfig,
+      originalConfig = deploymentConfig,
+      additionalDeploymentDataProvider = additionalDeploymentDataProvider,
+    )
+  }
+
+  def additionalScenarioProperties: Map[String, ScenarioPropertyConfig] = Map(cronConfig)
```
arkadius, line 107: R u sure that it will work in our client's deployment? I thought that our client have more complex structure of properties than single property with schedule

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
mgoworko: 1. Changes only in API module, not changed in designer
- package `periodic` -> `scheduler`
- name changes from Periodic to Schedule/Scheduled/or removed "Periodic" prefix

2. I changed it, the interface now looks like that:
- `SchedulingSupported` is a trait, that must provide everything needed to enable scheduling
- `ScheduledExecutionPerformer` is an interface, that must be implemented to actually run scenarios on a specific engine-DM

```
trait SchedulingSupported extends SchedulingSupport {

  def createScheduledExecutionPerformer(
      modelData: BaseModelData,
      dependencies: DeploymentManagerDependencies,
      deploymentConfig: Config,
  ): ScheduledExecutionPerformer

  def customSchedulePropertyExtractorFactory: Option[SchedulePropertyExtractorFactory]

  def customProcessConfigEnricherFactory: Option[ProcessConfigEnricherFactory]

  def customPeriodicProcessListenerFactory: Option[ScheduledProcessListenerFactory]

  def customAdditionalDeploymentDataProvider: Option[AdditionalDeploymentDataProvider]

}
```

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
mgoworko: Name changed to `ScheduledExecutionPerformer`. It is a small interface, that the DeploymentManager must support, in order to perform scheduled executions (actually run scenario on a specific engine). Described in a comment above.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessesManager.scala
mgoworko: Removed interfaces PeriodicProcessesManagerProvider, PeriodicProcessesManager. The repository is used directly, without this intermediate layer.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
mgoworko: Done, new config section, also renamed class in code PeriodicBatchConfig -> SchedulingConfig.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
mgoworko: Renamed PeriodicBatchConfig -> SchedulingConfig. Field name `db` changed to `legacyDb` with comment, that it is legacy and will be removed in the future.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
mgoworko: We can still configure those additional properties in the DeploymentManagerProvider. For our client, we will still have to implement our custom DMProvider, which will override some fields of PeriodicExecutionSupported. I created POC, everything seemed to work fine. For client, we do not have to override additionalScenarioProperties.

```
trait PeriodicExecutionSupported extends PeriodicExecutionSupport {

  def engineHandler <- not overriden, from FlinkDM

  // overriden for client, with custom services 

  def customSchedulePropertyExtractorFactory: Option[PeriodicSchedulePropertyExtractorFactory]

  def customProcessConfigEnricherFactory: Option[ProcessConfigEnricherFactory]

  def customPeriodicProcessListenerFactory: Option[PeriodicProcessListenerFactory]

  def customAdditionalDeploymentDataProvider: Option[AdditionalDeploymentDataProvider]

}
```

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
arkadius: Ad.1 When we change it in designer as well?
Ad. 2 Much better, but it should be `customScheduledProcessListenerFactory`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
arkadius: What about other fields? I guess that `processingType` is also to remove because this config is already inside `processingType`. So why user have to specify it twice?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
arkadius: Ad. 2.1 With regards to `ScheduledExecutionPerformer`, take a look at occurrence of `engineHandler` in the PR

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
mgoworko: Other fields seem necessary, except processingType. But we cannot remove processing type at this moment. 
- this processing type is an identifier of the DM that stores data in the scheduling DB tables
- it can be different (and in default Nu config is different) than the "true" processingType"
- so it would make the migration much more difficult, I'd rather leave it without changes in this PR

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
mgoworko: 1. Information added in PR description. It will be just refactor, but a huge one, which I really want to avoid in this change. I can do it if I have some time in near future.
2,, 2.1. - done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
arkadius: > it can be different (and in default Nu config is different) than the "true" processingType"

I don't understand that but let me know if I'm correct or wrong. After migration it can be removed? If so, let's add a comment the same as with legacy db

mgoworko: It cannot be removed after migration, because it is a parameter, that identifies the DM, that uses the scheduling mechanism in the db entries. It was a mistake in the past, that it is not the same as the actual, true name of the processingType. All DB queries filter by `processingType`.  

This value can be removed from configuration only after simultaneously modifying all rows in the periodic db + at the same time changing this value in the config to be equal to the true value of processingType.

mgoworko: Added comment in code
```
 // The `processingType` value should be removed in the future, because it should correspond to the real value of processingType.
    // But at the moment it may not be equal to the value of processingType of the DM that uses scheduling mechanism.
    // Therefore, we must keep the separate value SchedulingConfig, until we ensure consistency between the real processingType and the one defined here.
```

arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicBatchConfig.scala
arkadius: Ok now I'm getting it and this comment is ok - let's clean up this mess eventually.

PR #7370, opened 2024-12-20
replaced spel parser

JulianWielga:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new parsing structure utilizing the `ts-spel` library for improved functionality.
	- Added methods for parsing lists and objects in the `AggMapLikeParser` class.

- **Bug Fixes**
	- Updated expected outputs in test cases to reflect changes in parsing logic.

- **Tests**
	- Expanded test coverage with additional scenarios for parsing lists and maps, including complex object structures.


JulianWielga:
@coderabbitai summary

JulianWielga:
changes to ts-spel appended with patch are here https://github.com/JulianWielga/ts-spel/tree/nu-changes

designer/client/src/components/graph/node-modal/aggregate/parser.test.ts
```diff
@@ -41,11 +44,13 @@ describe("AggMapLikeParser", () => {
         [`"aaa"`, null],
         [`{}`, {}],
         [`#AGG.map({})`, {}],
+        [`#AGG2.map({})`, null],
```
arkadius: Let's add a test case for `#AGG.map({:})` (empty map)

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
```diff
@@ -1,236 +1,57 @@
...
+        return text.slice(el.start, el.end).trim();
     }
 
-    private tokenizeInput(input: string) {
-        const lexResult = this.lexer.tokenize(input);
-        this.fullText = input;
-        this.input = lexResult.tokens;
+    #parseObject(input: string, ast?: Ast) {
+        switch (ast?.type) {
+            case "InlineList":
```
arkadius: Let's write a note that it is a work around for a bug in the ts-spel parser that it treats empty map as empty list (I'm not sure If I understand you correctly)

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
```diff
@@ -1,236 +1,57 @@
...
+        },
+    ) {
+        return text.slice(el.start, el.end).trim();
     }
 
-    private tokenizeInput(input: string) {
-        const lexResult = this.lexer.tokenize(input);
-        this.fullText = input;
-        this.input = lexResult.tokens;
+    #parseObject(input: string, ast?: Ast) {
```
arkadius: I would call it more precisely. It looks like it parse map aggregate aggregators, not a general object

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
```diff
@@ -1,236 +1,57 @@
...
-        this.lexer = new Lexer(tokens);
-        this.parseObject = withLogs(this.parseObject.bind(this));
-        this.parseList = withLogs(this.parseList.bind(this));
-        this.performSelfAnalysis();
     }
 
-    parseList(input: string): Array<string> {
-        this.tokenizeInput(input);
-        return this.groupBy() || null;
+    #parseList(input: string, ast?: Ast) {
```
arkadius: The same as above - I'd call it parseListToStringExpression or parseAggregateKey

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
JulianWielga: it's not a bug. it's the way it works. {} means empty list. i want to treat empty list as empty object in this case - that's all. there is no sense and case where ":" would be inserted. test case shows this clearly.

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
JulianWielga: it returns object - Record<string, string> and parses both: map and map wrapped with #AGG.map to the same format.

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
arkadius: Why do you want to treat empty list as empty object? In which situation empty list in `parseToObject`? As far as I understand it correctly, for empty strategy of aggregations it should be `#AGG.map({:})` and for matching `aggregateBy` expression it should be `{:}`. Empty list is invalid on BE side for both cases.

designer/client/src/components/graph/node-modal/aggregate/aggMapLikeParser.tsx
arkadius: Ok, now I'm getting it. Thanks to better names that you provided now, it is visible without getting deep into this code. IMO it would be better to split this method into two methods. One for `aggregator` and one for `aggregateBy`. These are two different expressions with different format. Why to merge them? We saved ~10 lines of code.  The risk of some bug now or after changes on BE side around this is higher now.

arkadius (approved):
Added comments worth checking but not blockers

arkadius (approved):

PR #7376, opened 2024-12-31
[NU-1941] Evaluate value every time for count > 1 in sample-generator

mslabek:
## Checklist before merge

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -34,8 +34,11 @@
   this fixes issue with unexpected low scale when performing division on BigDecimals which were created in such conversion.
 * [#7379](https://github.com/TouK/nussknacker/pull/7379) Removed CustomAction mechanism.
 * Changes to `periodic` component (renamed to `sample-generator`):
-    * [#7368](https://github.com/TouK/nussknacker/pull/7368) Component rename: `periodic` to `sample-generator`
-    * [#7373](https://github.com/TouK/nussknacker/pull/7373) Improvements to `period` editor
+  * [#7368](https://github.com/TouK/nussknacker/pull/7368) Component rename: `periodic` to `sample-generator`
+  * [#7373](https://github.com/TouK/nussknacker/pull/7373) Improvements to `period` editor
+  * [#7376](https://github.com/TouK/nussknacker/pull/7376) Previously, when count was > 1, the value was evaluated once
+    and emitted times count. For example: if the value was evaluated to be a random UUID and count was 5, one UUID was
+    generated and emitted 5 times. Now in one count batch each value is evaluated separately.
 
 ## 1.18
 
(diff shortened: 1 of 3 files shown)


docs/Changelog.md
```diff
@@ -32,7 +32,12 @@
 * [#7341](https://github.com/TouK/nussknacker/pull/7341) Added possibility to choose presets and define lists for Integer typed parameter inputs in fragments.
 * [#7356](https://github.com/TouK/nussknacker/pull/7356) Integers converted to BigDecimals have scale 18,
   this fixes issue with unexpected low scale when performing division on BigDecimals which were created in such conversion.
-* [#7368](https://github.com/TouK/nussknacker/pull/7368) Component rename: `periodic` to `sample-generator`
+* Changes to `periodic` / `sample-generator` component:
```
arkadius: Instead of slash I would add `(new sample-gerator)` or renamed to

arkadius, line 35: Instead of slash I would add `(new `sample-gerator)` or renamed to

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/SampleGeneratorSourceFactory.scala
```diff
@@ -88,6 +89,14 @@ class SampleGeneratorSourceFactory(timestampAssigner: TimestampWatermarkHandler[
     }
   }
 
+  class ContextMultiplierPerCountFunction(count: Int) extends FlatMapFunction[Context, Context] with Serializable {
```
arkadius: I think that we should create a new context for each invocation as well. Thanks to that it be possible to correlate each generated value with things that happened in the scenario

docs/Changelog.md
mslabek: done :+1:

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/SampleGeneratorSourceFactory.scala
mslabek: The context we create at this level is actually only used for evaluation of the `value` parameter. Its simply just a temporary context just for that and its not used later. Later, a proper context is created in `FlinkContextInitializingFunction`.

I added a comment and I also made this source extend StandardFlinkSource instead of FlinkSource and removed some duplicate logic through that.

arkadius: Oh I see, we could extend `FlinkSource` instead of `StandardFlinkSource` and control when the context is generated. This temporary context id will be visible in errors topic and in test mechanism if we add a support for that. I'm not forcing to do it because current solution is also fine for now but if you want you can change it that way.

arkadius (approved):

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/SampleGeneratorSourceFactory.scala
mslabek: Unfortunately, couldn't manage to use this approach to make it cleaner. Ultimately, the contexts are created per each event, so it should work regardless with current approach.

PR #7387, opened 2025-01-03
[NU-1828] Notification reloading creator panel when configuration rel…

arkadius:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/notifications/NotificationService.scala
```diff
@@ -55,12 +57,16 @@ class NotificationServiceImpl(
     val limit = now.minusMillis(config.duration.toMillis)
     scope match {
       case NotificationsScope.NotificationsForLoggedUser(user) =>
-        notificationsForUserActions(user, limit)
+        for {
+          notificationsForUserActions <- notificationsForUserActions(user, limit)
+          globalNotifications = fetchGlobalNotificationsAndTriggerEviction(limit)
```
mslabek: This can be deduplicated with lines 66-67. Like:
```scala
    def fetchUserAndGlobalNotifications(user: LoggedUser): Future[List[Notification]] = {
      for {
        userActions <- notificationsForUserActions(user, limit)
        global = fetchGlobalNotificationsAndTriggerEviction(limit)
      } yield userActions ++ global
    }

    scope match {
      case NotificationsScope.NotificationsForLoggedUser(user) =>
        fetchUserAndGlobalNotifications(user)

      case NotificationsScope.NotificationsForLoggedUserAndScenario(user, processName) =>
        for {
          userAndGlobalNotifications <- fetchUserAndGlobalNotifications(user)
          scenarioNotifications <- notificationsForScenarioActivities(user, processName, limit)
        } yield userAndGlobalNotifications ++ scenarioNotifications
    }
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala
```diff
@@ -719,7 +724,15 @@ class AkkaHttpBasedRouteProvider(
               _
             ),
           )
-          new ReloadableProcessingTypeDataProvider(laodProcessingTypeDataIO)
+          val firstConfigurationLoad = new AtomicBoolean(true)
+          val loadAndNotifyIO = laodProcessingTypeDataIO
+            .map { state =>
+              if (!firstConfigurationLoad.compareAndSet(true, false)) {
+                globalNotificationRepository.saveEntry(Notification.configurationReloaded)
```
mslabek: imo it would be also ok if we notified FE to reload configuration on first time too, since configuration can change not only after reload but also each time BE starts

designer/client/src/containers/Notifications.tsx
```diff
@@ -44,9 +44,15 @@ const prepareNotification =
...
+    (
+        { scenarioName, toRefresh }: BackendNotification,
+        currentScenarioName: string,
+        processVersionId: number,
+        currentProcessingType: string,
+        currentIsFragment: boolean,
+    ): ThunkAction =>
     (dispatch) => {
-        if (!scenarioName || scenarioName !== currentScenarioName) {
+        if (scenarioName && scenarioName !== currentScenarioName) {
```
mslabek, line 55: This new condition causes a problem outside view of scenario. When you load scenario list view, the notifications also fire and reloading configuration will result in a call under `api/processDefinitionData/undefined?isFragment=undefined` . I think need different conditions for refreshing creator vs state and activity since for creator we have no scenario name.

designer/server/src/main/scala/pl/touk/nussknacker/ui/notifications/NotificationService.scala
```diff
@@ -107,4 +113,10 @@ class NotificationServiceImpl(
     } yield notificationsForScenarioActivities
   }
 
+  private def fetchGlobalNotificationsAndTriggerEviction(limit: Instant) = {
+    val globalNotifications = globalNotificationRepository.fetchEntries(limit)
```
mslabek, line 118: right now if we have multiple refreshes previously to an FE client being lanched, the FE will have N notifications to refresh config, which will result in N calls to api for definition of processing type. It could either be solved by just sending the single latest notification of reload or by flattening them on FE into one request.

designer/client/src/containers/Notifications.tsx
arkadius: Fixed

designer/server/src/main/scala/pl/touk/nussknacker/ui/notifications/NotificationService.scala
arkadius: I found the resolution of this problem on FE side more suitable. Take a look

mslabek (approved):
LGTM!

PR #7391, opened 2025-01-07
Handle all finished periodic scenarios waiting for reschedule instead of one

gskrobisz:
## Describe your changes

Problem occurs when multiple periodic scenarios waiting for reschedule.

For periodic scenarios "handleFinished" processed only one scenario. When this scenario fails to be rescheduled (and stays in "deployed" state) in the next iteration "handleFinished" fetches the same list of periodic scenarios to reschedule. And those scenarios have no chance to be rescheduled.

Here I change "reschedule only first of waiting scenarios" to "reschedule all of waiting scenarios".

## Checklist before merge

### DIFF
--- a/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
+++ b/engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
@@ -325,7 +325,7 @@ class SlickPeriodicProcessesRepository(
     getLatestDeploymentsForEachSchedule(
       processesHavingDeploymentsWithMatchingStatus,
       deploymentsPerScheduleMaxCount = 1
-    ).map(_.values.headOption.getOrElse(SchedulesState(Map.empty)))
+    ).map(schedulesForProcessNames => SchedulesState(schedulesForProcessNames.values.map(_.schedules).reduce(_ ++ _)))
   }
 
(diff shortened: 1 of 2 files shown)


engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
```diff
@@ -325,7 +325,7 @@ class SlickPeriodicProcessesRepository(
     getLatestDeploymentsForEachSchedule(
       processesHavingDeploymentsWithMatchingStatus,
       deploymentsPerScheduleMaxCount = 1
-    ).map(_.values.headOption.getOrElse(SchedulesState(Map.empty)))
+    ).map(schedulesForProcessNames => SchedulesState(schedulesForProcessNames.values.map(_.schedules).reduce(_ ++ _)))
```
mateuszkp96: Let's use foldLeft instead of reduce (reduce would throw an exception on an empty schedules map))

mateuszkp96 (changes requested):

engine/flink/management/periodic/src/main/scala/pl/touk/nussknacker/engine/management/periodic/db/PeriodicProcessesRepository.scala
gskrobisz: fixed

mateuszkp96 (approved):
LGTM

PR #7404, opened 2025-01-09
[NU-1944] Fix spel evaluation error when using conversion extensions methods or array.get extension method

arkadius:
## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
@@ -78,26 +78,30 @@ object ExtensionMethods {
 abstract class ExtensionMethod[R: ClassTag] {
   val argsSize: Int
   def invoke(target: Any, args: Object*): R
+  // This method gets the same arguments as passed to invoke
   def returnType(args: Object*): Class[_] = classTag[R].runtimeClass.asInstanceOf[Class[R]]
 }
 
 object ExtensionMethod {
 
-  def NoArg[R: ClassTag](method: Any => R): ExtensionMethod[R] = new ExtensionMethod {
-    override val argsSize: Int                         = 0
-    override def invoke(target: Any, args: Object*): R = method(target)
-  }
+  def NoArg[R: ClassTag](method: Any => R): ExtensionMethod[R] =
+    new ExtensionMethod[R] {
+      override val argsSize: Int                         = 0
+      override def invoke(target: Any, args: Object*): R = method(target)
+    }
 
-  def NoArg[R: ClassTag](method: Any => R, clazzProvider: () => Class[_]): ExtensionMethod[R] = new ExtensionMethod {
-    override val argsSize: Int                         = 0
-    override def invoke(target: Any, args: Object*): R = method(target)
-    override def returnType(args: Object*): Class[_]   = clazzProvider()
-  }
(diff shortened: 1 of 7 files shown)


scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -1733,6 +1733,15 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
     }
   }
 
+  test("should convert value to given type with correct return type") {
```
paw787878, line 1741: You can also add tests from bug [NU-1944]

  test("should be able to run indexer on created with toMap map") {
    evaluate[Int]("""{{key: 1, value: 5}}.toMap[1]""") shouldBe 5
  }

  test("should be able to run indexer on created with toMapOrNull map") {
    evaluate[Int]("""{{key: 1, value: 5}}.toMapOrNull[1]""") shouldBe 5
  }

Those tests pass on this commit

arkadius: Dodane ;)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -87,18 +88,23 @@ object ExtensionMethod {
     override def invoke(target: Any, args: Object*): R = method(target)
   }
 
+  def NoArg[R: ClassTag](method: Any => R, clazzProvider: () => Class[_]): ExtensionMethod[R] = new ExtensionMethod {
```
paw787878: clazzProvider parameter can be replaced with simply clazz: Class[_], though now I see that You may want to keep some convention so that it acts similar to SingleArg. In that case it can stay

lukasz-bigorajski (approved):
LGTM

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
```diff
@@ -77,7 +78,7 @@ object ExtensionMethods {
 abstract class ExtensionMethod[R: ClassTag] {
   val argsSize: Int
   def invoke(target: Any, args: Object*): R
-  def returnType: Class[R] = classTag[R].runtimeClass.asInstanceOf[Class[R]]
+  def returnType(args: Object*): Class[_] = classTag[R].runtimeClass.asInstanceOf[Class[R]]
```
paw787878, line 82: maybe it is obvious, but maybe it is comment worthy to write that args arg here is always the same as args arg passed to invoke method

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
arkadius: changed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/extension/ExtensionMethods.scala
arkadius: added

paw787878 (approved):
lgtm

PR #7408, opened 2025-01-10
[NU-1777] Nu Runtime app related refactoring

mk-software-pl:
## Describe your changes

- NuRuntimeApp based on IOApp
- ModelClassLoader extends the URLClassLoader
- the `RunnableScenarioInterpreter#run` method returns `IO[Unit]` and doesn't block any thread but completes after the interpreter initiation 
- `ActorSystemBasedExecutionContextWithIORuntime` renamed to `ExecutionContextWithIORuntimeAdapter` and adapted to be an general-purpose util (not the designer/server util only)

## Checklist before merge


utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/util/loader/ModelClassLoader.scala
```diff
@@ -3,12 +3,13 @@ package pl.touk.nussknacker.engine.util.loader
 import com.typesafe.scalalogging.LazyLogging
 
 import java.io.File
-import java.net.{URI, URL, URLClassLoader}
+import java.net.{URI, URL}
 import java.nio.file.Path
+import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
 
-case class ModelClassLoader private (classLoader: ClassLoader, urls: List[URL]) {
+case class ModelClassLoader private (val urls: List[URL], parent: ClassLoader) extends URLClassLoader(urls, parent) {
```
arkadius, line 10: Why did you replace composition with inheritance? ModelClassLoader is not A classloader. To be A classloader it should be useable in other places where other classloaders could be used. For me, it is a value class to explicit show the purpose of the real class loader.

mk-software-pl: > ModelClassLoader is not A classloader

Does it mean the name was wrong? ;)

I don't think I just changed the composition to inheritance. Previously, the ModelClassLoader class was a dummy holder of a class loader (in particular the instance of URLClassLoader). When you had been debugging the code, you saw an instance of URLClassLoader and you guessed what's this (eg. based on URLs in it). Now, it's not a dummy holder. It's still the URL class loader (but not wrapped in the unnecessary package), but without hiding this fact.

The inheritance is 100% justified here.

arkadius: Should we change dummy holders of strings such as `ProcessName` to classes extending `String`/`Charsequence`? I don't see anything wrong in dummy (as you call them) value classes. They narrow the usage of classes and clarify what these classes represent in the real world.

arkadius: But maybe this subject is not worth discussing more, we can follow your approach if you want.

arkadius (approved):

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/util/loader/ModelClassLoader.scala
mk-software-pl: The dummy thing in this approach is the fact that we create the holder, but then in each place we get the held value from it and use it. We didn't take advantage of the type of the holder or the fact that it hides the class loader. 

Plus the things I mentioned earlier - at runtime we couldn't easily see that it was the URLClassLoader that was associated with the model jars.

PR #7434, opened 2025-01-13
[NU-1926] Typing of indexing of map

paw787878:
## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -214,11 +214,21 @@ private[spel] class Typer(
     def typeIndexer(e: Indexer, typingResult: TypingResult): NodeTypingResult = {
       typingResult match {
         case TypedClass(clazz, param :: Nil)
-            if clazz.isAssignableFrom(classOf[java.util.List[_]]) || clazz.isAssignableFrom(classOf[Array[Object]]) =>
+            if classOf[java.util.List[_]].isAssignableFrom(clazz) || classOf[Array[Object]].isAssignableFrom(clazz) =>
```
lukasz-bigorajski: Why have you changes this?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -32,6 +32,9 @@ object AssignabilityDeterminer {
   def isAssignableStrict(from: TypingResult, to: TypingResult): ValidatedNel[String, Unit] =
     isAssignable(from, to, StrictConversionChecker)
 
+  def isAssignableWithNoConversion(from: TypingResult, to: TypingResult): ValidatedNel[String, Unit] =
```
lukasz-bigorajski: IMHO Without is better than with no, what do you think?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -223,6 +226,19 @@ object AssignabilityDeterminer {
 
   }
 
+  private object NoConversionConversionChecker extends ConversionChecker {
```
lukasz-bigorajski: This name is unclear to me, why 'conversion' is repeated? What about `WithoutConversionChecker` is not good enough?

lukasz-bigorajski (changes requested):
Some questions

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
paw787878: Those changes (I changed order of classes in isAssignableFrom in 3 places) are not needed to finish this task, however it looked to me like a mistake so I changed it. I would expect that this code is supposed to check if clazz is list, after my change if condition will evaluate to true also if clazz is ArrayList. Since this code seemed to work on prod probably clazz can never be subclass of List, in that case it can be even changed to clazz == classOf[java.util.List]

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
paw787878: renamed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
paw787878: renamed

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
lukasz-bigorajski: IMHO if we are changing sth it should have equivalent part in tests because I had to ask why you change this part.

paw787878: I reverted this change of order of types in isAssignableFrom. I think that this change was correct, but it was irrelevant to this task and probably it was not really changing any behaviour of nu

lukasz-bigorajski (approved):
LGTM

PR #7443, opened 2025-01-14
[NU-1935] Change record indexer behaviour

paw787878:
## Checklist before merge


docs/Changelog.md
```diff
@@ -48,6 +48,7 @@
 * [#7404](https://github.com/TouK/nussknacker/pull/7404) Fix spel evaluation error when using conversion extensions methods or array.get extension method
 * [#7420](https://github.com/TouK/nussknacker/pull/7420) Add toInteger and toIntegerOrNull conversions. Also add canBeInteger extension
 * [#7438](https://github.com/TouK/nussknacker/pull/7438) Map int32 integer format in OpenAPI schema to the `Integer` type
+* [#7443](https://github.com/TouK/nussknacker/pull/7443) Indexing on record is more similar to indexing on map. Now spel expression "{a: 5, b: 10}[#input.a]" compiles and has type "Integer". Expression "{a: 5}["b"]" still does not compile because it is known at compile time that record does not have property "b".
```
lukasz-bigorajski: Why is it: `[#input.a]` instead of `["a"]`?
Do you mean that you are fixing an indexing using dynamic value like record variable?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -199,11 +200,13 @@ private[spel] class Typer(
             case _                                      => typeFieldNameReferenceOnRecord(indexString, record)
           }
         case indexKey :: Nil if indexKey.canBeConvertedTo(Typed[String]) =>
-          if (dynamicPropertyAccessAllowed) valid(Unknown) else invalid(DynamicPropertyAccessError)
-        case _ :: Nil =>
+          if (dynamicPropertyAccessAllowed) valid(Unknown)
+          else if (record.runtimeObjType.params.size == 2) valid(record.runtimeObjType.params(1))
```
lukasz-bigorajski: What if record value could be nullable?
I don't understand that line: `record.runtimeObjType.params.size == 2)`. When does record runtimeObjtType is not a Map and doesn't have 2 generic params? For me it's always true

lukasz-bigorajski: For me it's the same as:
```
          if (dynamicPropertyAccessAllowed) valid(Unknown)
          else record.runtimeObjType.params match {
            case _ :: value :: Nil => valid(value)
            case _ => valid(Unknown)
          }
```

lukasz-bigorajski (commented):
I left some questions and I wondering who requested the changes?

docs/Changelog.md
paw787878: I used [#input.a] so that value of index is not known at compile time. Yes, my change fixes only dynamic indexing case

lukasz-bigorajski: Can you give a some example of input to show user clear use case of that problem?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
paw787878: I also think that it will always be map with 2 type parameters, but I wanted to make it safe. I changed to match as You requested, I also added check to see if it is map to make it more readable.

I don't understand what You mean by nullable record value. From what I understand spel typing in nu does not have distinction between nullable and not nullable types

docs/Changelog.md
paw787878: I changed changelog description

lukasz-bigorajski (approved):
LGTM

PR #7446, opened 2025-01-14
Fix fragment nodes errors in scenario's tips

Cichan-mcc:
## Describe your changes

Currently when we have a scenario which uses a fragment which has errors we get information about it in the `Tips` section. Unfortunately they are not clickable unlike other node errors. Furthermore the node representing the fragment is not highligted and does not indictate that there is something wrong with it.

output

With these changes error tips regarding fragment nodes are now clickable and open in a new tab problematic fragment node so the user can fix the error. Also, the fragment node will be highlighted now to indicate a problem with it.

output2

## Checklist before merge

### DIFF
--- a/designer/client/src/components/graph/NodeUtils.ts
+++ b/designer/client/src/components/graph/NodeUtils.ts
@@ -213,6 +213,25 @@ class NodeUtils {
             }
         }
     };
+
+    // We can recognize that a node is part of a referenced fragment by making sure that:
+    // 1. The nodeId is not present in the scenarioGraph
+    // 2. There is a fragment node in the scenarioGraph whose name is part of the nodeId
+    isFragmentNodeReference = (nodeId: NodeId, scenarioGraph: ScenarioGraph): boolean => {
+        const isNodePresentInScenarioGraph = !!this.getNodeById(nodeId, scenarioGraph);
+        return !isNodePresentInScenarioGraph && scenarioGraph.nodes.some((n) => nodeId.startsWith(n.id));
+    };
+
+    getDetailsFromFragmentNode = (nodeId: NodeId, scenarioGraph: ScenarioGraph): { fragmentId: string; fragmentNodeId: string } | null => {
+        if (this.isFragmentNodeReference(nodeId, scenarioGraph)) {
+            const fragment = scenarioGraph.nodes.find((n) => nodeId.startsWith(n.id));
+            return {
+                fragmentId: fragment.ref.id,
+                fragmentNodeId: nodeId.replace(`${fragment.id}-`, ""),
+            };
+        }
+        return null;
+    };
 }
 
 //TODO this pattern is not necessary, just export every public function as in actions.js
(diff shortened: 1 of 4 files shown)


designer/client/src/components/tips/error/NodeErrorsLinkSection.tsx
```diff
@@ -22,6 +25,23 @@ export default function NodeErrorsLinkSection(props: NodeErrorsLinkSectionProps)
             <div>
                 <ErrorHeader message={message} />
                 {nodeIds.map((nodeId, index) => {
+                    const isFragmentNodeReference = NodeUtils.isFragmentNodeReference(nodeId, scenario.scenarioGraph);
+                    if (isFragmentNodeReference) {
+                        const { fragmentId, fragmentNodeId } = NodeUtils.getDetailsFromFragmentNode(nodeId, scenario.scenarioGraph);
+                        return (
+                            <React.Fragment key={nodeId}>
+                                <ErrorLinkStyle
+                                    variant={"body2"}
```
lciolecki, line 34: What does mean body2?

designer/client/src/components/graph/NodeUtils.ts
```diff
@@ -213,6 +213,25 @@ class NodeUtils {
...
+    // We can recognize that a node is part of a referenced fragment by making sure that:
+    // 1. The nodeId is not present in the scenarioGraph
+    // 2. There is a fragment node in the scenarioGraph whose name is part of the nodeId
+    isFragmentNodeReference = (nodeId: NodeId, scenarioGraph: ScenarioGraph): boolean => {
+        const isNodePresentInScenarioGraph = !!this.getNodeById(nodeId, scenarioGraph);
+        return !isNodePresentInScenarioGraph && scenarioGraph.nodes.some((n) => nodeId.startsWith(n.id));
+    };
+
+    getDetailsFromFragmentNode = (nodeId: NodeId, scenarioGraph: ScenarioGraph): { fragmentId: string; fragmentNodeId: string } | null => {
+        if (this.isFragmentNodeReference(nodeId, scenarioGraph)) {
```
lciolecki, line 226: Make this isFragmentNodeReference as private and move it below getDetailsFromFragmentNode.

Dzuming (approved):
LGTM

designer/client/src/components/tips/error/NodeErrorsLinkSection.tsx
Cichan-mcc: It applies the theme typography style. It is `body2` to keep it consistent with other `ErrorLinkStyle` components in the `Tips` section as they all use that variant.

designer/client/src/components/graph/NodeUtils.ts
Cichan-mcc: It's not private on purpose 😃 It's also used in `NodeErrorsLinkSection.tsx` and `Graph.tsx` for node highlighting

PR #7447, opened 2025-01-14
[NU-1974] Initialize model classloaders only once

mslabek:
## Describe your changes
Right now model classloaders are initialized each time the model is initialized or reloaded. They are not garbage collected as they are referenced in by something like this `ProtectionDomain[]`  <- `AccessControlContext` <- Netty Thread. Closing the classloaders doesn't help. This results in a metaspace leak.

If we initialize them only once, it stops being a problem at the price of not-reloadability of classpath of model.

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
@@ -5,63 +5,33 @@ import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine._
 import pl.touk.nussknacker.engine.api.process.ProcessingType
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
-import pl.touk.nussknacker.engine.util.loader.{ModelClassLoader, ScalaServiceLoader}
+import pl.touk.nussknacker.engine.util.loader.ScalaServiceLoader
 import pl.touk.nussknacker.ui.configloader.{ProcessingTypeConfigs, ProcessingTypeConfigsLoader}
 import pl.touk.nussknacker.ui.process.processingtype._
 import pl.touk.nussknacker.ui.process.processingtype.loader.ProcessingTypeDataLoader.toValueWithRestriction
 import pl.touk.nussknacker.ui.process.processingtype.provider.ProcessingTypeDataState
 
-import java.nio.file.Path
-import java.util.concurrent.atomic.AtomicReference
-
 class ProcessingTypesConfigBasedProcessingTypeDataLoader(processingTypeConfigsLoader: ProcessingTypeConfigsLoader)
     extends ProcessingTypeDataLoader
     with LazyLogging {
 
-  private object ModelClassloadersHolder {
-
-    private val modelClassloaders: AtomicReference[Option[Map[String, ModelClassLoader]]] =
-      new AtomicReference(None)
-
-    def getOrCreate(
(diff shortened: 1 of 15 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
```diff
@@ -5,16 +5,50 @@ import com.typesafe.scalalogging.LazyLogging
...
+import java.nio.file.Path
+import java.util.concurrent.atomic.AtomicReference
+
 class ProcessingTypesConfigBasedProcessingTypeDataLoader(processingTypeConfigsLoader: ProcessingTypeConfigsLoader)
     extends ProcessingTypeDataLoader
     with LazyLogging {
 
+  private object ModelClassloadersHolder {
+
+    private val modelClassloaders: AtomicReference[Option[Map[String, ModelClassLoader]]] =
```
lciolecki: Why do you need Option here?

mslabek: To have a more explicit uninitialized state - None instead of empty map. Both approaches are imo ok since in this case I think empty map always means uninitialized state, but I'm subjectively leaning towards Option approach because of its explicitness. Do you think simply Map would be better?

designer/server/src/test/scala/pl/touk/nussknacker/test/base/it/NuResourcesTest.scala
```diff
@@ -127,7 +128,8 @@ trait NuResourcesTest
   protected val deploymentManagerProvider: DeploymentManagerProvider =
     new MockManagerProvider(deploymentManager)
 
-  private val modelData = ModelData(processingTypeConfig, modelDependencies)
+  private val modelData =
+    ModelData(processingTypeConfig, modelDependencies, ModelClassLoader.apply(processingTypeConfig.classPath, None))
```
arkadius: Please remove explicit `apply` call

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
arkadius: Why didn't you create this map once before the rest of the initialization code?  Take a look that we already have some initialization steps and each step depends on other steps which is visible in order of initialization - see `NussknackerAppFactory`

mslabek: In `NussknackerAppFactory` we don't have access to `workingDirectoryOpt` but now I see its for test purposes and during execution its hardcoded to None on https://github.com/TouK/nussknacker/blob/aecd8a0a42aa7fa64ef675697208d0ac3c5c543e/designer/server/src/main/scala/pl/touk/nussknacker/ui/server/AkkaHttpBasedRouteProvider.scala#L774

I'll move it to `NussknackerAppFactory` and hardcode the `workingDirectoryOpt` on that level, thanks :+1:

arkadius (approved):
LGTM

designer/server/src/test/scala/pl/touk/nussknacker/test/base/it/NuResourcesTest.scala
mslabek: done :+1:

PR #7453, opened 2025-01-14
Improve scenario loading time

mgoworko:
This PR closed, splitted into a few changes, with different approach to caching

## Describe your changes

Improved scenario loading time. On real-world example, this change managed to reduce loading time by 40-50%. Two examples huge scenarios with complicated models: time reduced from 10 seconds to 5 seconds and from 7 seconds to 3-4 seconds.

Changes:
1. The scenario graph rendering starts as soon as possible. There is an addition API call at the very beginning. The graph is first fetched without validation, which is very quick. And the rendering starts immediately
2. The problematic call to /processDefinitionData/{processingType} is now cached. The response time of that endpoint may have been very long, depending on model, and existing fragments, which were resolved on each call.
3. The test capabilities endpoints no longer trigger compiling the scenario - time reduced from 1-2 seconds to 100-200 millis
4. The "tips" and "creator panel" are no longer waiting for the graph to load. They are shown immediately, and their content is added as soon as is possible.

## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/fragment/FragmentParameterTypingParser.scala
```diff
@@ -8,7 +9,24 @@ import scala.util.Try
 
 class FragmentParameterTypingParser(classLoader: ClassLoader, classDefinitions: Set[ClassDefinition]) {
 
+  private val typingResultCache: Cache[String, Try[TypingResult]] =
```
lciolecki: I wonder, maybe we should move cache and parsing Special object like: ClassTypingResult? And use it everywhere where we do converting class to typing ?

mgoworko: This PR changed a lot since that comment. There is now auto-refreshable, cache on the entire UIDefinitions service.

PR #7502, opened 2025-01-27
[NU-1923] Rename source 'sample-generator' to 'event-generator' and 'period' to 'schedule'

mslabek:
## Checklist before merge


docs/scenarios_authoring/DataSourcesAndSinks.md
```diff
@@ -184,17 +184,25 @@ Holds event in the node until
...
-This source provides functionality of sending a number of given events in a periodic way. It's mainly used for testing.
+Parameters:
 
-This source has the following parameters:
-- period - specifies how often events will be sent
-- count - specifies number of event that will be sent every `period`
-- value - specifies data that event will hold
+- schedule - specifies how often events will be sent
+- count - specifies number of events that will be generated at every run
+- value - specifies data that event will hold. Expressions will be evaluated for every event in the `count`
```
jedrz: > in the `count`

Does this mean that an event is created once for each batch? Perhaps it would be better to remove this fragment?

mslabek: I think this may be not clear enough. What this was meant to express is that each event is evaluated seperately (so for count = 10 it will evaluate the expression 10 times, 1 per 1 count). Previously it worked differently. I'll talk with this paragraph's author.

jedrz: Okay, maybe `specifies data that event will hold. Expressions will be evaluated for every generated/emitted event` is good enough?

mslabek: Right, its ok. Changed it like you proposed just with `Expression` not `Expressions` since its just one expression thats being evaluated

jedrz (approved):

PR #7504, opened 2025-01-27
Return scenario validation error when incompatible changes were introduced for a component

Cichan-mcc:
## Describe your changes

Currently we can introduce incompatible changes to fragment and component parameter definitions which can result in weird errors during validation, displaying weird values and also in some cases passing validation when it shouldn't. The problem happens when we dynamically change editor type and language of a parameter. The simpliest reproduction steps would be:

1. Create a fragment with fixed values which uses a preset.
2. Use that fragment in a scenario and save it.
3. Modify the fragment so now the parameter's input mode is Any Value and save it.
4. Now when we enter the scenario there will be problems with the value, validation and saving the scenario.

Screenshot 2025-01-28 at 10 06 28

The only solution right now is to remove the problematic node from scenario graph and use a new one from creation panel with current definitions.

This situation can be also replicated with regular components by using `AdditionalUIConfigProvider` and reloading the model. 

It happens because of inconsistency between expression's language which is saved in the scenario graph and component's current editor and expression. Fixing that would be nontrivial as it would actually require introducing versioning for fragments and components. Since that would be an overkill right now, I introduced additional validation on parameter level to detect such situations when changes to the component were incompatible. When such issue is detected scenario won't pass validation and informative message will be displayed.

Screenshot 2025-01-28 at 11 54 34

Unfortunately users will still have to drag a new component, copy all the parameters into a new one and then delete the old one. At least now they will be informed about what happened, what to do and won't be able to deploy problematic scenarios.

## Checklist before merge

### DIFF
--- a/common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
+++ b/common-api/src/main/scala/pl/touk/nussknacker/engine/graph/expression/Expression.scala
@@ -4,13 +4,28 @@ import io.circe.{Decoder, Encoder}
 import io.circe.generic.JsonCodec
 import io.circe.syntax.EncoderOps
 import pl.touk.nussknacker.engine.graph.expression.Expression.Language
+import pl.touk.nussknacker.engine.graph.expression.Expression.Language.{
+  DictKeyWithLabel,
+  Spel,
+  SpelTemplate,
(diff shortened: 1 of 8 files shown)


bartektartanus (approved):
LGTM, nice job.

PR #7510, opened 2025-01-29
[NU-1960] API for messages preview in source

pielas:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -349,6 +350,63 @@ class NodesApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoi
...
+    NodesError,
+    String,
+    Any
+  ] = {
+    baseNuApiEndpoint
+      .summary("Fetch latest records for specific node")
+      .tag("Nodes")
+      .post
+      .in(
+        "nodes" / path[ProcessName]("scenarioName") / "fetchLatestRecordsForNode" / path[Int](
```
arkadius: I think, that we are currently following an approach that uses nouns in paths, so to fulfil this approach it should be rather `/nodes/$scenarioName/latestRecords`. I think, that we can skip `Nodes` as it already is a `nodes` endpoint.

Also, lest's think about further extensions of this endpoint. Let's consider situations:
- I want to fetch only one record with given id
- I want to use pagination
- ...

Maybe more extensive would be to use sth like: `/nodes/$scenarioName/recordsPreview?limit=$limit`. Thanks to that we can extend it by adding features such as sort=asc/desc, offset, getting by id etc. The `recordsPreview` part is an example, I'm not convinced that it is a good name for what we are providing. When we talk, for now it is natural for us to emphasize that it is `preview` so It would be easier to find this endpoint but on the other hand this can change in the future. For example amqp call it just a data: https://demo.nussknacker.io/data

WDYT about this?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/test/ModelDataTestInfoProvider.scala
```diff
@@ -110,15 +110,45 @@ class ModelDataTestInfoProvider(modelData: ModelData) extends TestInfoProvider w
...
+    val jobData = JobData(metaData, ProcessVersion.empty)
+    val nodeId  = NodeId(sourceNodeData.id)
+    val testDataGenerator = prepareSourceObj(
+      sourceNodeData
+    )(jobData, nodeId).toList
+      .flatMap(_.cast[TestDataGenerator])
+    val validatedGenerators = NonEmptyList
+      .fromList(testDataGenerator)
+      .map(x => Right(x))
+      .getOrElse(Left("Provided source didn't produce any data"))
```
arkadius: "Provided source didn't produce any data" doesn't really say what happened. I would split this and do a little refactor around this place:
- `prepareSourceObj` is in fact `compileSourceNode` - when it returns `Invalid`, it means that there is some problem in expressions entered in node form or was used a non-existing component type. I would generalize it to "Source node can't be compiled. Problems: ..."
- When `Source.cast[TestDataGenerator]` returns `Invalid` it means that the given source doesn't support records preview

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -349,6 +350,63 @@ class NodesApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoi
...
+                  summary = Some("Simple scenario test data in json stringify form"),
+                  value = "{name: John}"
+                )
+              )
+            )
+        )
+      )
+      .errorOut(
+        oneOf[NodesError](
+          fetchLatestRecordsErrorExample,
```
arkadius: The `Example` is only a part of this error definition. Maybe let's skip it or use some other suffix?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -1472,6 +1542,7 @@ object NodesApiEndpoints {
     sealed trait NodesError
 
     object NodesError {
+      final case class FetchLatestRecordsError(msg: String)             extends NodesError
```
arkadius: :/ It would be nice to have all business cases such as node compilation error, source doesn't support data preview ... etc. here. Can you provide it? At least let's add a TODO for that

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesApiHttpService.scala
```diff
@@ -144,6 +148,39 @@ class NodesApiHttpService(
       }
   }
 
+  expose {
+    nodesApiEndpoints.fetchLatestRecordsForNodeEndpoint
+      .serverSecurityLogic(authorizeKnownUser[NodesError])
+      .serverLogicEitherT { implicit loggedUser =>
+        { case (scenarioName, numberOfRecords, fetchLatestRecordsDto) =>
```
arkadius: I'm guessing that we don't validate if `numberOfRecords` is too big because we assume that `scenarioTestService` already do this because there is a limit for data generation for scenario testing purpose?

It is not visible in the code. If we `scenarioTestService` returned a business errors, we would see it. As an alternative, we can at least add a comment about that assumption.

arkadius (commented):
Nice, a few small comments added and one comment about endpoint naming.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
pielas: it's indeed a good idea to make it extensible. I changed it to just /records

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
pielas: I don't really understand what do you mean. Can you give an example of other suffix that fits? I think that word `Example` fits here because it indicates that it's just an example of `FetchLatestRecordsError` returned by the service. Additionally other example output instances are named using the same convention, with `Example` suffix, like `noScenarioExample`, `malformedTypingResultExample`, `ProcessingTypeExample`

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesApiHttpService.scala
pielas: I added such comment to `NodesApiHttpService` 

              // The service is expected to limit the number of returned records if the response is too large.
              // If it does not, we may need to implement client-side pagination or request size limits.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
arkadius: Tapir call this entities "Outputs". The output contains a status code, information about what encoding will be used (for example plain text/json) and an example of a body. I think that probably someone firstly extracted only a body example and called it an example and then changed it to output and forgot to change the variable name. Take a look for example on `scenarioNotFoundErrorOutput`, `migrateEndpointErrorOutput`. I think that this naming schema is more closer to the model that Tapir provides.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -657,6 +652,78 @@ object NodesApiEndpoints {
...
+            Example.of(
+              summary = Some("Source preview not supported"),
+              value = UnsupportedSourcePreview("sourceId")
+            )
+          )
+      )
+
+    val noDataGeneratedExample: EndpointOutput.OneOfVariant[NoDataGenerated.type] =
+      oneOfVariantFromMatchType(
+        NotFound,
```
arkadius: More than one output variant with the same status code won't work correctly. There is a long discussion in Tapirs repo why it works that way. The unpleasant thing is that Tapir don't event report an Exception in that case. It justs ignore some of variants. To make it work with Tapir, you need to have one output variant with given status code with many examples. Take a look at for example `DeploymentApiEndpoints.runDeploymentEndpoint`

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -587,6 +652,78 @@ object NodesApiEndpoints {
...
+        plainBody[NoDataGenerated.type]
+          .example(
+            Example.of(
+              summary = Some("No test data generated"),
+              value = NoDataGenerated
+            )
+          )
+      )
+
+    val noSourcesWithTestDataGenerationExample: EndpointOutput.OneOfVariant[NoSourcesWithTestDataGeneration.type] =
```
arkadius: This one won't happen because it is a valid error for test generation based on scenario, not a node. Maybe we could another sealed trait level below `ScenarioTestError` - `SourceTestError`. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/NodesApiHttpService.scala
arkadius: I dug a little bit more into the code and found out where this records count is done for test data generation mechanism, take a look at this:
```
      _ <- Either.cond(
        testSampleSize <= testDataSettings.maxSamplesCount,
        (),
        s"Too many samples requested, limit is ${testDataSettings.maxSamplesCount}"
      )
```
in `ScenarioTestService`. At the beginning, I thought that it is somewhere deeper and your endpoint do this check implicitly, but it's not. We have to reuse this code, IMO it is not acceptable that the API accepts any limit. User might blow up our heap memory.

arkadius (commented):
Some comments resolved, a few added. Please take a look at failing tests as well.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
pielas: Ah, I understand now. It indeed doesn't make sense to call them examples. I renamed all ErrorOutputs according to suggested naming convention

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
pielas: I tried to make class hierarchy where ScenarioError extends SourceError, but it didn't really work too well with Eithers. This is why I duplicated classes a little bit and I made it a flat structure for both cases. I don't have any better ideas for now. WDYT?

arkadius (approved):

PR #7519, opened 2025-01-31
Fix different json representation in scheduling mechanism

mgoworko:
## Describe your changes

The scenario graph stored as json in the db tables of the scheduling mechanism is different, than the one stored in process_versions table. It is already pre-resolved, for example fragments are expanded. The previous change in the periodic mechanism did not take that into account and was incorrent. We need to store this pre-resolved json. But there is one optimization possible, and now introduced in this PR in the new tables - cleaning this json when scenario is cancelled.

## Checklist before merge


designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
```diff
@@ -0,0 +1,22 @@
...
+trait V1_062__ScenarioSchedulingTablesModificationsDefinition extends SlickMigration with LazyLogging {
+
+  import profile.api._
+
+  override def migrateActions: DBIOAction[Any, NoStream, Effect.All] = {
+    logger.info("Starting migration V1_062__ScenarioSchedulingTablesModifications")
+    for {
+      _ <- sqlu"""ALTER TABLE "periodic_scenarios" RENAME TO "scheduled_scenarios";"""
+      _ <- sqlu"""ALTER TABLE "scheduled_scenarios" ALTER COLUMN "input_config_during_execution" DROP NOT NULL;"""
+      _ <- sqlu"""ALTER TABLE "scheduled_scenarios" ADD COLUMN "resolved_scenario_json" VARCHAR(10485760);"""
```
piotrp: Why are `input_config_during_execution` and `resolved_scenario_json` optional?

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
```diff
@@ -0,0 +1,22 @@
+package db.migration
+
+import com.typesafe.scalalogging.LazyLogging
+import pl.touk.nussknacker.ui.db.migration.SlickMigration
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+trait V1_062__ScenarioSchedulingTablesModificationsDefinition extends SlickMigration with LazyLogging {
```
piotrp: Why isn't this a common migration in resoures/sb/migration/common/V1_062__ScenarioSchedulingTablesModifications.sql?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/PeriodicProcessesRepository.scala
```diff
@@ -160,10 +159,16 @@ trait PeriodicProcessesRepository {
       periodicProcessId: PeriodicProcessId,
   ): Action[Option[String]]
 
-  def fetchCanonicalProcessWithVersion(
+  def fetchCanonicalProcess(
+      periodicProcessId: PeriodicProcessId,
+      processName: ProcessName,
+      versionId: VersionId,
```
piotrp: `processName` and `versionId` are never used (I don't count `InMemPeriodicProcessesRepository`, as it's used only in tests)

piotrp (changes requested):

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
mgoworko: Both values are now deleted after the schedules scenario is cancelled. (`def markInactive` in the repo - only new, not in the legacy)

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
mgoworko: It is a common migration, just handled by Slick and defined as code.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/repository/PeriodicProcessesRepository.scala
mgoworko: You are right, removed

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
piotrp: In three files, hence my question about your reasons for making additional files for Scala to compile when one resource would have been enough

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
piotrp: Can we add a `CHECK` constraint to ensure that this assumption will be always held? Something like `CHECK scheduled_scenarios_deployment_details ((active AND input_config_during_execution IS NOT NULL AND resolved_scenario_json IS NOT NULL) OR NOT active)`?

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
mgoworko: Done, pure sql script migration and added constraint.

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
mgoworko: Done, pure sql script migration and added constraint.

piotrp: We need that `scheduled_scenarios_` prefix in constraint name - in PostgreSQL names must be unique within parent schema, so we risk name conflicts in the future

piotrp (approved):

designer/server/src/main/scala/db/migration/V1_062__ScenarioSchedulingTablesModificationsDefinition.scala
mgoworko: `scheduled_scenarios_check_config_and_json_present_for_active`

PR #7521, opened 2025-02-03
[Nu-7520] Validate run off schedule comment on FE

### LINKED ISSUE #7520: Validate comment while performing RunOffScheduleButton
## 💥 Proposal

The context of this issue is to validate on FE the comment input for `RunOffSchedule` action. Currently the validation returned by an endpoint: `/api/processManagement/runOffSchedule/{{ processName }}` is completely ignored on the FE and user is misinformed because modal window closes, action is not performed but on UI nothing is being shown. I propose to show the validation result on FE as in the case of `Deploy` or `Cancel` button.

DeamonDev:
## Describe your changes

I propose to throw an exception inside `HttpService#runOffSchedule` method instead of returning value from the catch block. This is then coherent with `deploy/cancel` methods in this service and solves the issue with omitting the validation of comment in the modal window (since it is based on error catching). 

closes #7520 

## Checklist before merge

### DIFF
--- a/designer/client/src/http/HttpService.ts
+++ b/designer/client/src/http/HttpService.ts
@@ -388,7 +388,7 @@ class HttpService {
                     msg: msg,
                 };
                 if (error?.response?.status != 400) return this.#addError(msg, error, false).then(() => result);
-                return result;
+                throw error;
             });
     }
 
--- a/designer/client/src/http/HttpService.ts
+++ b/designer/client/src/http/HttpService.ts
@@ -39,6 +39,7 @@ import { fixAggregateParameters, fixBranchParametersTemplate } from "./parameter
 import { handleAxiosError } from "../devHelpers";
 import { Dimensions, StickyNote } from "../common/StickyNote";
 import { STICKY_NOTE_DEFAULT_COLOR } from "../components/graph/EspNode/stickyNote";
+import { ScenarioActionResult, ScenarioActionResultType } from "../components/toolbars/scenarioActions/buttons/types";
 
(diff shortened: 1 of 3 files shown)


designer/client/src/http/HttpService.ts
```diff
@@ -388,7 +388,7 @@ class HttpService {
                     msg: msg,
                 };
                 if (error?.response?.status != 400) return this.#addError(msg, error, false).then(() => result);
-                return result;
+                throw error;
```
Dzuming: IMO, we could adjust the rest methods related to the deploy dialog to return status, instead of throwing an error. I agree with [it](https://basarat.gitbook.io/typescript/type-system/exceptions#exceptional-cases). 

In a current approach, it looks a little problematic,

```
  const confirmAction = useCallback(async () => {
        try {
            await action(processName, processVersionId, comment);
            props.close();
        } catch (error) {
            setValidationError(error?.response?.data);
        }
    }, [action, comment, processName, props, processVersionId]);
```
We always set a validation error when an error is cached. IIUC, what we want is to display this error only when the backend sends us error response data indicating that something is wrong with a comment field.

So the flow should be:
- Display validation error only in the case where there is a validation error from BE
- In case of other error from BE, display it in the toast message
- In case of an unhandled error, for example, FE request can be rejected on the proxy level, for various reasons, like no connection to BE or request timeout, we should display "Something went wrong on a server-side" (or equivalent of this error)

Dzuming (changes requested):

Dzuming (approved):
LGTM

PR #7524, opened 2025-02-04
Add posibility to choose valid value after editors inconsistencies

Cichan-mcc:
## Describe your changes

In Nussknacker we can have inconsistencies in parameter exression's language due to incompatible changes made to fragments or components which i described in [this PR](https://github.com/TouK/nussknacker/pull/7504)

Currently the only way to "fix" these inconsistencies is to drag a new node from the creator panel, copy all the values and then delete the old inconsistent node. This is quite problemtic for the users. With these changes when user chooses a new valid value the language in the scenario graph will be adjusted based on current editor type (similar detecting based on editor's type is already done in `EditorBasedLanguageDeterminer.scala`). 

ezgif-412a015994e9b

## Checklist before merge


Dzuming (approved):
LGTM

PR #7529, opened 2025-02-06
Add confirmation windows to delete sticky note

philemone:
## Checklist before merge


designer/client/src/components/graph/Graph.tsx
```diff
@@ -395,6 +398,7 @@ export class Graph extends React.Component<Props> {
             this.processGraphPaper.hideTools();
             if (isStickyNoteElement(cellView.model)) {
                 if (!this.props.isPristine) return;
+                if (this.props.selectionState.length > 0) this.props.resetSelection(...[]);
```
Dzuming: Do we need to spread an empty array here?
```suggestion
                if (this.props.selectionState.length > 0) this.props.resetSelection();
```

Dzuming (changes requested):

designer/client/src/components/graph/Graph.tsx
philemone: Yeah, you are right, thanks :)

Dzuming (approved):

PR #7539, opened 2025-02-07
Use programArgsList when deploying to Flink

piotrp:
## Describe your changes

Gets rid of an ugly encoding hack that was needed for very old Flink versions.

## Checklist before merge


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/exception/RestartStrategyFromConfiguration.scala
```diff
@@ -61,7 +61,7 @@ object RestartStrategyFromConfiguration {
     val flinkConfig = new Configuration
     // restart-strategy.fixed-delay.attempts
     val strategy = config.getString(strategyPath)
-    flinkConfig.setString(restartStrategyFlinkConfigPrefix, strategy)
+    flinkConfig.setString(s"$restartStrategyFlinkConfigPrefix.type", strategy)
```
arkadius: Is it some bug fix? Maybe let's mention about that in changelog

piotrp: Kinda, one WARN log less from Flink. Moved to #7542

arkadius (approved):

PR #7553, opened 2025-02-11
[NU-1977] `Key` variable change in window components

ForrestFairy:
## Describe your changes

`key` variable was created in window components by taking values from `groupBy` and parsing them to a list in string. Now it will be passed as a value which was provided, eg.
```
for id = 1
Before -> groupBy: id -> key: "[1]"
After -> groupBy: id -> key: 1
```
For old scenarios users would need to go into aggregation components and edit `groupBy` field for it to take effect. Otherwise it will still create `key` variable as string

## Checklist before merge


docs/Changelog.md
```diff
@@ -92,6 +92,7 @@
   * Kafka source has "offset reset strategy" parameter that controls starting point for reading events.
   * Configuration entry `kafkaEspProperties.forceLatestRead` is replaced with `kafkaEspProperties.defaultOffsetResetStrategy`
 * [#7545](https://github.com/TouK/nussknacker/pull/7545) Added `useMiniClusterForDeployment` option allowing to run Flink scenarios on Flink MiniCluster
+* [#7553](https://github.com/TouK/nussknacker/pull/7553) Key variable created in window component doesn't have to be string
```
raphaelsolarski, line 154: Let's write what it is now, and add a little note to the migration guide.

designer/client/src/components/graph/node-modal/aggregate/useAggParamsSerializer.tsx
```diff
@@ -46,7 +46,7 @@ export function useGroupByParamsSerializer(): [(text: string) => string[], (para
 
         switch (paramName) {
             case "groupBy":
-                return `{ ${content} }.toString`;
+                return `${content}`;
```
raphaelsolarski: IMO we should wrap keys in list if there is more than one key (now the widget works odd if we put multiple key)

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
```diff
@@ -128,16 +143,16 @@ object keyed {
...
 
-    def enrichWithKey[V](ctx: Context, key: String): Context =
+    def enrichWithKey[K, V](ctx: Context, key: K): Context =
       ctx.withVariable(VariableConstants.KeyVariableName, key)
 
     def contextTransformation(ctx: ValidationContext)(
         implicit nodeId: NodeId
     ): ValidatedNel[ProcessCompilationError, ValidationContext] =
-      ctx.withVariableOverriden(VariableConstants.KeyVariableName, Typed[String], None)
+      ctx.withVariableOverriden(VariableConstants.KeyVariableName, Unknown, None)
```
raphaelsolarski: Shouldn't we type the `key` output variable according to type put in `groupBy` parameter? e.g. `String` or `List[String]`.

raphaelsolarski (changes requested):

designer/client/src/components/graph/node-modal/aggregate/useAggParamsSerializer.tsx
raphaelsolarski: Screenshot 2025-02-21 at 14 24 12

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
raphaelsolarski: Screenshot 2025-02-21 at 14 25 54

designer/client/src/components/graph/node-modal/aggregate/useAggParamsSerializer.tsx
```diff
@@ -33,7 +33,19 @@ export function useAggParamsSerializer(): [
 export function useGroupByParamsSerializer(): [(text: string) => string[], (paramName: string, arr: string[]) => string] {
     const parser = useMemo(() => new AggMapLikeParser(), []);
 
-    const deserialize = useCallback((input: string) => parser.parseList(input), [parser]);
+    const deserialize = useCallback(
+        (input: string) => {
+            // if some more advanced transformation is done we should
+            console.log(input);
```
raphaelsolarski: should it stay?

designer/client/src/components/graph/node-modal/aggregate/useAggParamsSerializer.tsx
```diff
@@ -33,7 +33,19 @@ export function useAggParamsSerializer(): [
...
+            if (input.startsWith("{") && input.search("}.to.*") != -1) return [input];
+            // if already a list we only need to add 'toString' so it's nicely displayed in editor
+            // if only one element we need to wrap it in '{}' and add 'toString'
+            const list = input.startsWith("{") && input.endsWith("}") ? input + ".toString" : "{" + input + "}.toString";
+
+            return parser.parseList(list);
```
raphaelsolarski: Can't we try to run parseList first and to return single element list with expression when parseList returns null?
Sniffing the type of expression using regexes and startsWith seems to be a little bit tricky 😄

raphaelsolarski (changes requested):

raphaelsolarski (approved):

arkadius (approved):
I read only last 3 commits, they look ok for me

JulianWielga (approved):

PR #7515, opened 2025-01-30
Scenario visualization loading improvements - part 1 - FE changes

mgoworko:
## Describe your changes

Improved scenario loading time. On real-world example, this change together with other parts of the task managed to reduce loading time by 40-50%.

Changes:

- The scenario graph rendering starts as soon as possible. There is an addition API call at the very beginning. The graph is first fetched without validation, which is very quick. And the rendering starts immediately
- The "tips" and "creator panel" are no longer waiting for the graph to load. They are shown immediately, and their content is added as soon as is possible.

Firefox profiling results on a sample, quite huge scenario:
(profiling started when user sees list of scenarios, then clicks on one of them the chosen scenario is loaded)

- before
before

- after
after

## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [ ] Code is cleaned from temporary changes and commented out lines
- [ ] Parts of the code that are not easy to understand are documented in the code
- [ ] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [ ] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #7554, opened 2025-02-11
Correctly display node details after validation

mgoworko:
## Describe your changes

Fix problem, that was caused by https://github.com/TouK/nussknacker/pull/7515.

In node details, the params with `dictKeyWithLabel` type were not correctly displayed. The scenario graph is now initially fetched without validation. Because of that, the `dictKeyWithLabel` params are missing the `label` field. This change fixes that by fetching labels separately.

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/DictApiEndpoints.scala
```diff
@@ -113,6 +156,12 @@ object DictApiEndpoints {
       )
     }
 
+    implicit val noDictEntryForKeyCodec: Codec[String, NoDictEntryForKey, CodecFormat.TextPlain] = {
+      BaseEndpointDefinitions.toTextPlainCodecSerializationOnly[NoDictEntryForKey](e =>
+        s"No dict item found in dictionary with id: ${e.dictId} for key: ${e.key}"
```
Cichan-mcc: Maybe instead of `item` let's call it `entry`? I think we call it this way everywhere else. WDYT?

Cichan-mcc (approved):
BE looks good to me. Let's wait for some FE review.

Dzuming (approved):
LGTM

PR #7565, opened 2025-02-14
Ordering of map fields

vder:
## Checklist before merge

### DIFF
--- a/engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/JavaCollectionsSerializationTest.scala
+++ b/engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/JavaCollectionsSerializationTest.scala
@@ -14,24 +14,31 @@ import pl.touk.nussknacker.engine.flink.test.FlinkSpec
 import pl.touk.nussknacker.engine.flink.test.ScalatestMiniClusterJobStatusCheckingOps.miniClusterWithServicesToOps
 import pl.touk.nussknacker.engine.flink.util.source.CollectionSource
 import pl.touk.nussknacker.engine.flink.util.transformer.FlinkBaseComponentProvider
+import pl.touk.nussknacker.engine.graph.expression.Expression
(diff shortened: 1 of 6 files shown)


utils/utils/src/main/scala/pl/touk/nussknacker/engine/util/json/ToJsonEncoder.scala
```diff
@@ -66,7 +67,7 @@ case class ToJsonEncoder(
           case a: UUID           => safeString(a.toString)
           case a: DisplayJson    => a.asJson
           case a: scala.collection.Map[_, _] => encodeMap(a.toMap)
-          case a: java.util.Map[_, _]        => encodeMap(a.asScala.toMap)
+          case a: java.util.Map[_, _]        => encode(ListMap(a.asScala.toList: _*))
```
arkadius, line 70: It might by not optimal. You should check what is done in implementation. Maybe there is some better approach when we assume that given map is a `LinkedHashMap`? Can you write a micro benchmark for that and verify the difference in times between old serialization method vs the new one?

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/JavaCollectionsSerializationTest.scala
```diff
@@ -64,6 +73,40 @@ class JavaCollectionsSerializationTest extends AnyFunSuite with FlinkSpec with M
...
+      )
+
+      val linkedHashMap = new util.LinkedHashMap[String, AnyRef]()
+      sampleMap.foreach { case (char, int) => linkedHashMap.put(s"$char", Integer.valueOf(int)) }
+
+      val result = collectingListener.results
+        .nodeResults("end")
+        .map(_.variableTyped[Map[_, _]]("mapVariable"))
+
+      result shouldBe List(Some(linkedHashMap))
```
arkadius, line 105: Does it check the order? I think that the intent would be more clear if you'd compare lists instead

arkadius:
Closing due to inactivity

PR #7578, opened 2025-02-20
[NU-2015] Component labels

ForrestFairy:
## Describe your changes
Added configurable `labels` to components. They can be defined in:

- `ComponentDefinition`
- `componentUiConfig` in model config
- or can be left unset - we will create `label` by converting component `name` to Title Case

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -144,11 +144,30 @@ case class NussknackerVersion(value: Semver)
...
+
+object ComponentDefinition {
+
+  def apply(
+      name: String,
+      component: Component,
+      icon: Option[String] = None,
+      docsUrl: Option[String] = None,
+      designerWideId: Option[DesignerWideComponentId] = None,
+      label: Option[String] = None
```
arkadius, line 166: If we don't want to use some convention of generation of label based on name, let it be mandatory and let we put it after the name for easier recognition how one is related with another

ForrestFairy: Can't we leave it like this? This way:
- if `label` is provided we would use `label`
- if `label` is not provided we would use `name` for now

If we decide to add some more transformations of `name` to create `label` it could be done here as well

arkadius: IMO we shouln't leave it like this because it would propagate a wrong behaviour for further developers of components. They will forgot to put this label and will get ugly label.
We could have a convention that label is name but transformed from kebab-case to Title Case form. With this convention I agree that we can have this field optional following the CoC rule.

Dzuming (approved):
FE LGTM

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/ComponentProvider.scala
```diff
@@ -144,11 +144,34 @@ case class NussknackerVersion(value: Semver)
...
+  ): ComponentDefinition = {
+    label match {
+      case Some(value) => ComponentDefinition(name, component, icon, docsUrl, designerWideId, value)
+      case None => ComponentDefinition(name, component, icon, docsUrl, designerWideId, kebabCaseToTitleCase(name))
+    }
+  }
+
+  def kebabCaseToTitleCase(name: String): String = {
```
arkadius: Use `IdToTitleConverter`. Before it is possible, you have to move it down to `component-api` module which is available for both deployment manager API and for components

arkadius (commented):
Generally looks good, good job. One comment added + can you explicitly define label for `previousValue` available in `FlinkBaseComponentProvider`? I see that it is the only one with camelCase id in our public distribution.

arkadius (approved):

PR #7586, opened 2025-02-24
[NU-1904] Add possiblity to configure deployment parameters for Eager…

lukasz-bigorajski:
## Describe your changes
image

## Checklist before merge


extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessAction.scala
```diff
@@ -109,8 +109,8 @@ object ScenarioActionName {
 }
 
 /**
- * Used to define Source parameters for each action
+ * Used to define Flink deployment parameters for each action
```
gskrobisz: This is not clear for me. I guess action deploy in Flink environment is one of the possible use cases for this configuration.

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/LoggingService.scala
```diff
@@ -31,19 +36,39 @@ object LoggingService extends EagerService {
...
+            case Level.DEBUG => logger.debug(msg)
+            case Level.INFO  => logger.info(msg)
+            case Level.WARN  => logger.warn(msg)
+            case Level.ERROR => logger.error(msg)
+          }
         }
         Future.successful(())
       }
 
+      private def isNotDisabled(componentUseContext: ComponentUseContext) =
```
gskrobisz: Sounds like double negative. Do you mean "isEnabled"?

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/LoggingService.scala
```diff
@@ -31,19 +36,39 @@ object LoggingService extends EagerService {
...
+        componentUseContext
+          .deploymentData()
+          .flatMap(_.get("loggingComponentsDisabled"))
+          .forall(isDisabled => !isDisabled.toBoolean)
+
+      override def actionParametersDefinition: Map[ScenarioActionName, Map[ParameterName, ParameterConfig]] = Map(
+        ScenarioActionName.Deploy -> Map(
+          ParameterName("loggingComponentsDisabled") -> ParameterConfig(
```
gskrobisz: How about inverting logic? When I insert logging component to my scenario diagram I expect that logging is being done. Then, when I invoke "deploy" action, I can decide to supress logging.  

EDIT: I see that default value is "false" (and it IS logging by default), but I am sure that does not help to comprehend what is going on here.

components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenAPIServiceSpec.scala
```diff
@@ -31,9 +31,9 @@ class OpenAPIServiceSpec
     with LazyLogging
     with PatientScalaFutures {
 
-  implicit val componentUseCase: ComponentUseCase = ComponentUseCase.EngineRuntime
-  implicit val metaData: MetaData                 = MetaData("testProc", StreamMetaData())
-  implicit val context: Context                   = Context("testContextId", Map.empty)
+  implicit val componentUseCase: ComponentUseContext = ComponentUseContext.EngineRuntime(None)
```
gskrobisz: Rename componentUseCase to componentUseContext

gskrobisz (changes requested):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
```diff
@@ -0,0 +1,32 @@
...
+  * <li>Validation - used when compiling and validating nodes by Designer. Components should not be invoked in this mode.</li>
+  * <li>ServiceQuery - used when component (Service) is invoked by Designer in ServiceQuery.</li>
+  * <li>TestDataGeneration - used when compiling, but only for purpose of generating test data. Components should not be invoked in this mode.</li>
+  * </ul>
+  */
+sealed trait ComponentUseContext {
+
+  def deploymentData(): Option[NodeDeploymentData] = this match {
+    case EngineRuntime(nodeData) => nodeData
+    case _                       => None
```
arkadius: Let's use explicit cases matching. Thanks to that, we avoid some mistakes during changes around these contexts. Also, why not a normal polymorphism? I mean `def deploymentData`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+}
+
+object ComponentUseContext {
+  case class EngineRuntime(nodeData: Option[NodeDeploymentData]) extends ComponentUseContext
+  case object TestRuntime                                        extends ComponentUseContext
+  case object Validation                                         extends ComponentUseContext
```
arkadius: Can you remove contexts:
- `ServiceQuery` - It was for purpose of external project, I checked their repo and found that they can replace it with `EngineRuntime(None)` without any damage
- `Validation` - Our validation mechanism compiles components, and after that regrets result of compilation so, these components never will be used - we can pass `EngineRuntime(None)` there with some additional note that it is not important
- `TestDataGeneration` - During test data generation things are done very similar to validation. We compile components only to get test data generator, so component is not used for anything else and it doesn't need to know the context.
?

Also, I would rename two other contexts:
- `EngineRuntime` - scenario testing mechanism also run component on some engine so it is confusing, I would rename it to `LiveDeployment`/`RealDeployment` 
- `TestRuntime` - `ScenarioTesting`. Meantime we clarified this name for naming this thing. We have `scenarioTesting` in endpoints and in other places in the code

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
```diff
@@ -27,17 +31,57 @@ abstract class CommonModelDataInfoProvider(modelData: ModelData) {
...
-  protected def collectAllSources(scenario: CanonicalProcess): List[SourceNodeData] = {
-    scenario.collectAllNodes.flatMap(asSource) ++ scenario.collectAllNodes.flatMap(asFragmentInputDefinition)
+  def collectAndCompileAllSourcesAndServices(
+      scenario: CanonicalProcess
+  )(implicit jobData: JobData): Map[NodeComponentInfo, Any] = {
+    compileSourcesAndService(scenario.collectAllNodes) ++
+      compileSourcesAndService(scenario.collectAllNodes.flatMap(asFragmentInputDefinition))
   }
 
+  private def compileSourcesAndService(nodes: List[NodeData])(implicit jobData: JobData): Map[NodeComponentInfo, Any] =
```
arkadius: I think, that we should compile all non-built-in nodes. Consider situations:
- custom node - it is more general case as service. it can be also configurable
- sinks - I could imagine some situation when you want to specify behaviour of sink during deployment
We could introduce a sealed trait for that types of components, I mean `NonBuiltInComponent`. `CustomComponent` would be better but it is already reserved for `CustomNode`.

arkadius (commented):
Looks good, some minor comments added

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/ProcessAction.scala
lukasz-bigorajski: Removed 'deployment' - done

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/LoggingService.scala
lukasz-bigorajski: It was for a purpose, but I refactored it a little

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/LoggingService.scala
lukasz-bigorajski: Done

components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenAPIServiceSpec.scala
lukasz-bigorajski: Done

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
lukasz-bigorajski: Done

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
lukasz-bigorajski: Done - I renamed 'EngineRuntime' to 'LiveRuntime', what do you think?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
arkadius: Scala recommend to use braceless methods (`def deploymentData: Option[NodeDeploymentData`) for methods that have no side effects

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
arkadius: LGTM, nice

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/MetricsProviderForFlink.scala
```diff
@@ -1,18 +1,18 @@
...
 
   def createMetricsProvider(
-      componentUseCase: ComponentUseCase,
+      componentUseContextProvider: ComponentUseContextProvider,
       runtimeContext: RuntimeContext
   ): MetricsProviderForScenario = {
-    componentUseCase match {
-      case ComponentUseCase.TestRuntime => NoOpMetricsProviderForScenario
-      case _                            => new FlinkMetricsProviderForScenario(runtimeContext)
+    componentUseContextProvider match {
```
arkadius: Hmm I see that we use this class not only to provide `ComponentUseContext` but also for some pattern matching for other mechanism. I think that we should revert the name to sth like `RuntimeMode` or sth like that.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
lukasz-bigorajski: Methods without brackets are confusing for me because when invoke it I don't whether it's a variable or a method. This is why I chose method with brackets. So I'm really wondering whether to remove it

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/compiler/MetricsProviderForFlink.scala
lukasz-bigorajski: I like the name

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
lukasz-bigorajski: I fixed others, but I have to find out how to compile all nodes at once

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ComponentUseContext.scala
arkadius: I agree that it is a tricky conceptual decision made by creators of scala, but IMO being consistent with standards is very important value. We could talk about changing all braceless methods to methods with brackets in all scala projects in our ecosystem, but I think that it is not worth it ATM

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
arkadius: You don't need to compile all, built-in components can't be ommitted

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
```diff
@@ -26,18 +31,65 @@ abstract class CommonModelDataInfoProvider(modelData: ModelData) {
...
+
+  def compileSourceNode(
       source: SourceNodeData
   )(implicit jobData: JobData, nodeId: NodeId): ValidatedNel[ProcessCompilationError, Source] = {
     nodeCompiler.compileSource(source).compiledObject
   }
 
-  protected def collectAllSources(scenario: CanonicalProcess): List[SourceNodeData] = {
-    scenario.collectAllNodes.flatMap(asSource) ++ scenario.collectAllNodes.flatMap(asFragmentInputDefinition)
+  def compileScenario(scenario: CanonicalProcess)(implicit jobData: JobData): Map[NodeComponentInfo, Any] = {
```
arkadius: I think that it is important to emphasize somehow (maybe in method name?) that the purpose of this compilation is collection of all custom nodes (not filter or variable). Looking at current name it looks like we compile scenario to some executable unit. Maybe let's rename it to `compileAllCustomNodes`? Also this fallback to `Map.empty` is tricky. Maybe let's move it to place of invocation for clarity?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
```diff
@@ -26,18 +31,65 @@ abstract class CommonModelDataInfoProvider(modelData: ModelData) {
...
+        case source: SourcePart =>
+          acc + (nodeComponentInfo -> source.obj) ++ compilePart(source) ++ extractComponents(source.nextParts)
+        case custom: CustomNodePart =>
+          acc + (nodeComponentInfo -> custom.transformer) ++ compilePart(custom) ++ extractComponents(custom.nextParts)
+        case sink: SinkPart =>
+          acc + (nodeComponentInfo -> sink.obj)
+      }
+    }
+
+  private def compilePart(part: ProcessPart)(implicit jobData: JobData): Map[NodeComponentInfo, Any] =
```
arkadius: The same as above, we need to emphasize that this compilation is for the purpose of collection of all custom nodes. Maybe the same name `compileAllCustomNodes`? Or `compileAllPartCustomNodes`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledNodesCollector.scala
```diff
@@ -0,0 +1,38 @@
+package pl.touk.nussknacker.engine.compiledgraph
+
+import pl.touk.nussknacker.engine.compiledgraph.node._
+
+object CompiledNodesCollector {
+
+  def collectNodes(node: Node): List[Node] = {
```
arkadius: maybe `collectAllNodes` will be a little bit less confusing?

arkadius (commented):
Nice, minor comments added, but in general looks good

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/action/CommonModelDataInfoProvider.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledNodesCollector.scala
lukasz-bigorajski: Done

arkadius (approved):
Good job

gskrobisz (approved):

PR #7588, opened 2025-02-24
[NU-1979] `ProcessingTypeDataProvider`: fix for synchronization didn't work + some simplifications

arkadius:
## Describe your changes

The PR contains two changes:

1. After switching implementation of `ProcessingTypeDataProvider.reloadAll()` to `IO`-based,  the approach with `synchronized` stopped working - access to `stateValue` is not synchronized. The purpose of this change is to repair this situation.

2. Also, in the near future, we want to do significant changes around `ProcessingTypeDataProvider` - we want to reload only model, without deployment manager's. To do that it is necessary to reduce the complexity of this solution. As a step towards that, I changed `ProcessingTypeDataState.getCombined` from a function to `Try`, and replaced `ProcessingTypeDataState.stateIdentityValue` by other, eager, observer-based approach of reloading the data.  

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
@@ -13,6 +13,8 @@ import pl.touk.nussknacker.ui.process.processingtype.ProcessingTypeData.Scheduli
 import pl.touk.nussknacker.ui.process.processingtype.loader.ProcessingTypeDataLoader.toValueWithRestriction
 import pl.touk.nussknacker.ui.process.processingtype.provider.ProcessingTypeDataState
 
+import scala.util.Success
+
 class ProcessingTypesConfigBasedProcessingTypeDataLoader(
     processingTypeConfigsLoader: ProcessingTypeConfigsLoader
 ) extends ProcessingTypeDataLoader
@@ -103,9 +105,10 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
     // to assert the loaded configuration is correct (fail-fast approach).
     val combinedData = CombinedProcessingTypeData.create(processingTypesData)
 
-    ProcessingTypeDataState(
+    new ProcessingTypeDataState(
       processingTypesData.mapValuesNow(toValueWithRestriction),
-      () => combinedData,
+      // We want to fail fast - because of that we don't return Try in CombinedProcessingTypeData.create
+      Success(combinedData),
       // We pass here new Object to enforce update of observers
       new Object
     )
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
@@ -108,9 +108,7 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
     new ProcessingTypeDataState(
       processingTypesData.mapValuesNow(toValueWithRestriction),
       // We want to fail fast - because of that we don't return Try in CombinedProcessingTypeData.create
-      Success(combinedData),
-      // We pass here new Object to enforce update of observers
-      new Object
+      Success(combinedData)
     )
   }
 
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
@@ -13,7 +13,7 @@ import pl.touk.nussknacker.ui.process.processingtype.ProcessingTypeData.Scheduli
 import pl.touk.nussknacker.ui.process.processingtype.loader.ProcessingTypeDataLoader.toValueWithRestriction
 import pl.touk.nussknacker.ui.process.processingtype.provider.ProcessingTypeDataState
 
-import scala.util.Success
+import scala.util.{Success, Try}
 
 class ProcessingTypesConfigBasedProcessingTypeDataLoader(
     processingTypeConfigsLoader: ProcessingTypeConfigsLoader
@@ -29,7 +29,7 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
   ): IO[ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData]] = {
     processingTypeConfigsLoader
       .loadProcessingTypeConfigs()
-      .map(
+      .flatMap(
         createProcessingTypeData(
           _,
           getModelDependencies,
@@ -48,7 +48,7 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
       deploymentManagersClassLoader: DeploymentManagersClassLoader,
       modelClassLoaderProvider: ModelClassLoaderProvider,
       dbRef: Option[DbRef],
-  ): ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData] = {
+  ): IO[ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData]] = {
     // This step with splitting DeploymentManagerProvider loading for all processing types
     // and after that creating ProcessingTypeData is done because of the deduplication of deployments
     // See DeploymentManagerProvider.engineSetupIdentity
@@ -103,13 +103,15 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
 
     // Here all processing types are loaded and we are ready to perform additional configuration validations
     // to assert the loaded configuration is correct (fail-fast approach).
-    val combinedData = CombinedProcessingTypeData.create(processingTypesData)
-
-    new ProcessingTypeDataState(
-      processingTypesData.mapValuesNow(toValueWithRestriction),
-      // We want to fail fast - because of that we don't return Try in CombinedProcessingTypeData.create
-      Success(combinedData)
-    )
+    IO.fromEither(CombinedProcessingTypeData.create(processingTypesData).toEither.map { combinedData =>
+      new ProcessingTypeDataState(
+        processingTypesData.mapValuesNow(toValueWithRestriction),
+        // We return Success instead of passing here Try result of CombinedProcessingTypeData.create because
+        // we want to break reloading logic instead of creating a new state with failing combined date
+        // Try in ProcessingTypeDataState.combinedDataTry is only for initial state purpose when this data are not initialized yet
+        Success(combinedData)
+      )
+    })
   }
 
   private def createDeploymentManagerProvider(
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
@@ -1,5 +1,10 @@
 package pl.touk.nussknacker.ui.process.processingtype.provider
 
+import cats.effect.std.Mutex
+import cats.effect.unsafe.IORuntime
+import cats.effect.{IO, Ref}
+import cats.implicits.toTraverseOps
+import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.api.process.ProcessingType
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 import pl.touk.nussknacker.security.Permission
@@ -7,8 +12,9 @@ import pl.touk.nussknacker.ui.UnauthorizedError
 import pl.touk.nussknacker.ui.process.processingtype.ValueWithRestriction
 import pl.touk.nussknacker.ui.security.api.LoggedUser
 
-import java.util.concurrent.atomic.AtomicReference
-import scala.util.Try
+import java.util.concurrent.CopyOnWriteArrayList
+import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Try}
 
 /**
   * ProcessingType is a context of application. One ProcessingType can't see data from another ProcessingType.
@@ -16,21 +22,29 @@ import scala.util.Try
   *
   * This class is meant to provide access to some scope of data inside context of application to the user.
   * We don't want to pass all ProcessingType's data to every service because it would complicate testing of services
-  * and would broke isolation between areas of application. Due to that, this class is a `Functor`
+  * and would brake isolation between areas of application. Due to that, this class is a `Functor`
   * (to be precise `BiFunctor` but more on that below) which allows to transform the scope of `Data`.
   *
   * Sometimes it is necessary to have access also to combination of data across all ProcessingTypes. Due to that
   * this class is a `BiFunctor` which second value named as `CombinedData`
   *
   * This class caches `Data` and `CombinedData` wrapped in `ProcessingTypeDataState` to avoid computations of
-  * transformations during each lookup to `Data`/`CombinedData`. It behave similar to `Observable` where given
-  * transformed `ProcessingTypeDataProvider` check its parent if `ProcessingTypeDataState.stateIdentity` changed.
+  * transformations during each lookup to `Data`/`CombinedData`.
   *
   * ProcessingType is associated with Category e.g. Fraud Detection, Marketing. Given user has access to certain
-  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if he/she
+  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if their
   * has access to category.
   */
-trait ProcessingTypeDataProvider[Data, CombinedData] {
+abstract class ProcessingTypeDataProvider[Data, CombinedData](initialState: ProcessingTypeDataState[Data, CombinedData])
+    extends LazyLogging {
+
+  // TODO: get rid of unsafeRunSync for Ref and Mutex
+  protected val stateRef: Ref[IO, ProcessingTypeDataState[Data, CombinedData]] =
+    Ref.of[IO, ProcessingTypeDataState[Data, CombinedData]](initialState).unsafeRunSync()(IORuntime.global)
+
+  protected val stateMutex: Mutex[IO] = Mutex[IO].unsafeRunSync()(IORuntime.global)
+
+  private val observers = new CopyOnWriteArrayList[Observer[ProcessingTypeDataState[Data, CombinedData]]]()
 
   // TODO: replace with proper forType handling
   final def forProcessingTypeUnsafe(processingType: ProcessingType)(implicit user: LoggedUser): Data =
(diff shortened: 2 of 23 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProvider.scala
```diff
@@ -18,56 +19,53 @@ import pl.touk.nussknacker.ui.security.api.NussknackerInternalUser
...
-  private var stateValue: ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData] = emptyState
-
-  override private[processingtype] def state
-      : ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData] = {
-    synchronized {
-      stateValue
-    }
-  }
-
-  def reloadAll(): IO[Unit] = synchronized {
```
arkadius, line 37: This `synchronized` stopped working after we changed implementation to `IO`-based

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProvider.scala
```diff
@@ -18,56 +19,53 @@ import pl.touk.nussknacker.ui.security.api.NussknackerInternalUser
...
+            logger.error("Error occurred during reloading state. Rolling back previous state value", ex)
+            setStateValueAndNotifyObservers(beforeReload).map { _ =>
+              throw ex
+            }
+          }
+      } yield ()
     }
   }
 
-  def close(): IO[Unit] = synchronized {
```
arkadius, line 61: The same here

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -1,35 +1,50 @@
...
   * ProcessingType is associated with Category e.g. Fraud Detection, Marketing. Given user has access to certain
-  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if he/she
+  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if their
   * has access to category.
   */
-trait ProcessingTypeDataProvider[+Data, +CombinedData] {
+abstract class ProcessingTypeDataProvider[Data, CombinedData](initialState: ProcessingTypeDataState[Data, CombinedData])
+    extends LazyLogging {
+
+  // TODO: get rid of unsafeRunSync for Ref and Mutex
```
arkadius: I left this TODO because replacing it with a proper `IO`-based solution would complicate `AkkaHttpBasedRouteProvider` where we do a lot of `ProcessingTypeDataProvider.mapValues`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -1,56 +1,75 @@
...
+    extends LazyLogging {
+
+  // We use unsafeRunSync because IO-based solution would complicate a lot our application initialization logic where
+  // we pass to every service, various views for our ProcessingTypeData
+  protected val stateRef: Ref[IO, ProcessingTypeDataState[Data, CombinedData]] =
+    Ref.of[IO, ProcessingTypeDataState[Data, CombinedData]](initialState).unsafeRunSync()(IORuntime.global)
+
+  protected val stateMutex: Mutex[IO] = Mutex[IO].unsafeRunSync()(IORuntime.global)
+
+  private val observers = new CopyOnWriteArrayList[Observer[ProcessingTypeDataState[Data, CombinedData]]]()
```
arkadius: Thanks to this `observers`-based approach, it was possible to remove `ProcessingTypeDataState.stateIdentity`. Now state in `ProcessingTypeData` views, created by `.mapValues`/`.mapCombined` are reloaded eagerly instead of lazy.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -86,112 +88,85 @@ trait ProcessingTypeDataProvider[+Data, +CombinedData] {
...
-        transformState(currentObservedState)
-      } else {
-        currentValue
+  private def state: ProcessingTypeDataState[Data, CombinedData] = {
+    stateMutex.lock
+      .surround {
+        stateRef.get
       }
-    }
+      .unsafeRunSync()(IORuntime.global) // TODO: get rid of unsafeRunSync
```
arkadius: This is tricky after this changes with mutex and ref.  This method is used very often, during almost every request processing.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
```diff
@@ -103,11 +105,10 @@ class ProcessingTypesConfigBasedProcessingTypeDataLoader(
     // to assert the loaded configuration is correct (fail-fast approach).
     val combinedData = CombinedProcessingTypeData.create(processingTypesData)
 
-    ProcessingTypeDataState(
+    new ProcessingTypeDataState(
       processingTypesData.mapValuesNow(toValueWithRestriction),
-      () => combinedData,
-      // We pass here new Object to enforce update of observers
-      new Object
+      // We want to fail fast - because of that we don't return Try in CombinedProcessingTypeData.create
```
coutoPL: I don't get it how this comment refers to the line below. Maybe we should make it more clear?

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.scalatest.funsuite.AnyFunSuiteLike
+import org.scalatest.matchers.should.Matchers
+import pl.touk.nussknacker.test.utils.domain.TestFactory
+import pl.touk.nussknacker.ui.process.processingtype.ValueWithRestriction
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+import cats.effect.unsafe.implicits.global
+
+import scala.util.Success
+
+class ReloadableProcessingTypeDataProviderTest extends AnyFunSuiteLike with Matchers {
```
coutoPL, line 13: Could you please confirm this test detects the issue described in the 1st point of the PR's description?

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
```diff
@@ -0,0 +1,92 @@
...
+    // first load
+    an[Exception] shouldBe thrownBy {
+      reloadableProvider.reloadAll().unsafeRunSync()
+    }
+
+    val allDataAfterFailedLoad = reloadableProvider.all
+    allDataAfterFailedLoad shouldBe empty
+  }
+
+  class AutoClosableWithValue(val value: String) extends AutoCloseable {
```
coutoPL: private

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
```diff
@@ -0,0 +1,92 @@
...
+    @volatile
+    var closeWasInvoked = false
```
coutoPL, line 87: private

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: we don't remove observers. Is it ok?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: I'm not sure what we gain here by applying the lock on the mutex.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -1,56 +1,75 @@
...
-trait ProcessingTypeDataProvider[+Data, +CombinedData] {
+abstract class ProcessingTypeDataProvider[Data, CombinedData](initialState: ProcessingTypeDataState[Data, CombinedData])
+    extends LazyLogging {
+
+  // We use unsafeRunSync because IO-based solution would complicate a lot our application initialization logic where
+  // we pass to every service, various views for our ProcessingTypeData
+  protected val stateRef: Ref[IO, ProcessingTypeDataState[Data, CombinedData]] =
+    Ref.of[IO, ProcessingTypeDataState[Data, CombinedData]](initialState).unsafeRunSync()(IORuntime.global)
+
+  protected val stateMutex: Mutex[IO] = Mutex[IO].unsafeRunSync()(IORuntime.global)
```
coutoPL: I think that exposing the stateRef and stateMutex for the sake of usage by the implementation is not the best solution. It'd be better to give them a method that encapsulates the internals and give them the ability to safely modify the state within the critical section.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -86,112 +88,85 @@ trait ProcessingTypeDataProvider[+Data, +CombinedData] {
...
-) extends ProcessingTypeDataProvider[TT, CC] {
-
-  private val stateValue = new AtomicReference(transformState(observed.state))
+  final def combined: CombinedData = state.combinedDataTry.get
+
+  final protected[provider] def setStateValueAndNotifyObservers(
+      newValue: ProcessingTypeDataState[Data, CombinedData]
+  ): IO[Unit] = {
+    for {
+      _ <- stateRef.set(newValue)
```
coutoPL: An implementation class is able to use this method (setStateValueAndNotifyObservers) outside of the critical section. I'd be great if we could forbid doing that.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
arkadius: No, I didn't do test for 1st point because it is barely possible. Do you have an idea how to detect caching of shared mutated state by threads? It sometimes can be visible only on some computer architectures.
Just for sure, do you know where was the problem in previous solution or my comments around `synchronized` and description didn't help you?  I'm just want to be sure that we are on the same page.
[Edit] LSS It is related with 2nd point

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
arkadius: Oberversa are not created dynamically, the topology of observers is created once `AkkaHttpBasedRouteProvider`, so there is no need to clean it anytime. It is like binding dependencies in IoC, you bind it once for the whole application life.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
arkadius: The problem is that our only two implementations of this class: `ReloadableProcessingTypeDataProvider` and `TransformingProcessingTypeDataProvider` have other areas that they want to wrap with this synchronization block. For   `ReloadableProcessingTypeDataProvider` it is `reloadAll` and for `TransformingProcessingTypeDataProvider`  it is a short `notifyChange` section. We could add something like `def accessStateInCriticalSection(State => IO[T])` method, but it became so geraneric and complex approach that I wondering if copy-pasting both implementations won't be a better choice?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
arkadius: Hmm, take a look at `reloadAll`:
1. lock in
2. before = stateRef.get
3. close(before)
4. newState = ...
5. lock out

If we don't respect lock during read and someone will try to access the state between step 2. and 4., they will access old, closed state. Am I right or I'm missing something?

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
arkadius: This field is accessed by test's body

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypesConfigBasedProcessingTypeDataLoader.scala
arkadius: I rephrased this comment and also added explicit monads to make more explicit what was done here. Take a look now if it helped

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
arkadius: Let's move this discussion to this https://github.com/TouK/nussknacker/pull/7588/files#r1986981795 place, because these things are related.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
coutoPL: yes, I understand the problem with `synchronized`. 
I think we would be able to detect this problem by creating a large number of threads and trying to do the reload for each thread at the same time. 
I guess the thing was overlooked because the tests gave me the false sense of security. From the perspective of maintainability, we should implement the test. 

But at the same time, I'm not sure this is the right time to do it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: Looking from a perspective of `AkkaHttpBasedRouteProvider` class this is ok. But the perspective of the `ProcessingTypeDataProvider` class is different. This class accumulates some state and should clean it up too. Without taking into consideration its current usage (I mean: the class determines what it does, how it does, and how its state is managed, not its clients).

I'd propose to remove the observers in its close method. The behaviour obviously won't change but the solution is cleaner.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: ok, so it's related to this topic:
https://github.com/TouK/nussknacker/pull/7588/files#r1986981795

The problem is that the child class's implementation `reloadAll` method affects the internals of the base class.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: I like the `accessStateInCriticalSection` approach. Why do you think it will become complex? 
I see it like this: you want to do sth with the state, you call the method, do sth with the state and return new state (oh, so it would rather be `accessStateInCriticalSection(State => IO[State])`?).

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
coutoPL: oh, and when we do `unsafeRunSync` (I agree that refactor the app to not have to do that is not so easy right now), it'd be better to do it on a separate IORuntime (not the global one).

arkadius: Ok, I used separate IORuntime

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
arkadius: Sorry my bad, probably joining both issues that this PR resolves was not the best approach and it introduced confusions.  I think that we don't have a time to do intensive tests here. Let's do them when we occurr some problem with this approach. It is a side quest from side quest from the main task that I'm doing, I want to come back on track :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
arkadius: I added such a method, but with `Ref => IO[T]`. We have to be able to do both read and write in this critical section

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
coutoPL: sure

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -86,112 +90,119 @@ trait ProcessingTypeDataProvider[+Data, +CombinedData] {
...
+  }
 
-  override private[processingtype] def state: ProcessingTypeDataState[TT, CC] = {
-    stateValue.updateAndGet { currentValue =>
-      val currentObservedState = observed.state
-      if (currentObservedState.stateIdentity != currentValue.stateIdentity) {
-        transformState(currentObservedState)
-      } else {
-        currentValue
+  def close(): IO[Unit] = {
```
coutoPL: In my opinion, we shouldn't allow this method to be overridden (wrong child implementation may break this base class's internals).

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -1,56 +1,77 @@
...
   * ProcessingType is associated with Category e.g. Fraud Detection, Marketing. Given user has access to certain
-  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if he/she
-  * has access to category.
+  * categories see `LoggedUser.can`. Due to that, during each access to `Data`, user is authorized if they
+  * have access to category.
   */
-trait ProcessingTypeDataProvider[+Data, +CombinedData] {
+abstract class ProcessingTypeDataProvider[Data, CombinedData](
+    initialState: ProcessingTypeDataState[Data, CombinedData]
+)(implicit ioRuntime: IORuntime)
```
coutoPL, line 41: Why don't we create this IORuntime in the constructor of this class? The child class doesn't need it. The runtime is only needed by this class's internals. Moreover, the IORuntime is not tweaked or configured. The lifecycle of this object can be related to the lifecycle of this class and managed from the class's internals.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ProcessingTypeDataProvider.scala
```diff
@@ -86,112 +90,119 @@ trait ProcessingTypeDataProvider[+Data, +CombinedData] {
...
-  private[processingtype] def state: ProcessingTypeDataState[Data, CombinedData]
-
-  final def mapValues[TT](fun: Data => TT): ProcessingTypeDataProvider[TT, CombinedData] =
-    new TransformingProcessingTypeDataProvider[Data, CombinedData, TT, CombinedData](this, _.mapValues(fun))
-
-  final def mapCombined[CC](fun: CombinedData => CC): ProcessingTypeDataProvider[Data, CC] =
-    new TransformingProcessingTypeDataProvider[Data, CombinedData, Data, CC](this, _.mapCombined(fun))
+  final def combined: CombinedData = state.combinedDataTry.get
+
+  final protected[provider] def setStateValueAndNotifyObservers(
```
coutoPL: This method still allows access to the state outside of the critical section. 
There is the assumption that it is always called within the critical section.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProvider.scala
```diff
@@ -18,56 +20,49 @@ import pl.touk.nussknacker.ui.security.api.NussknackerInternalUser
  * Another thing that needs careful consideration is handling exception during ProcessingTypeData creation/closing - probably during
  * close we want to catch exception and try to proceed, but during creation it can be a bit tricky...
  */
-class ReloadableProcessingTypeDataProvider(
-    loadMethod: IO[ProcessingTypeDataState[ProcessingTypeData, CombinedProcessingTypeData]]
-) extends ProcessingTypeDataProvider[ProcessingTypeData, CombinedProcessingTypeData]
+class ReloadableProcessingTypeDataProvider[Data <: AutoCloseable, CombinedData](
```
coutoPL: let's make the constructor private

coutoPL (approved):
LGTM

PR #7591, opened 2025-02-25
Improve scenario testing error messages

jedrz:
## Describe your changes

- Errors returned during scenario testing, if possible point to `testDataSettings` configuration which can be helpful in fixing the problem
- `testDataSettings.testDataMaxLength` is validated also for received test data
- Common scenario testing errors message are placed in single object

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiHttpService.scala
```diff
@@ -135,31 +157,51 @@ object TestingApiHttpService {
 
   object TestingError {
 
-    final case class TestDataGenerationError(msg: String)             extends TestingError
-    final case class NoScenario(scenarioName: ProcessName)            extends TestingError
-    final case class NoProcessingType(processingType: ProcessingType) extends TestingError
-    final case object NoPermission                                    extends TestingError with CustomAuthorizationError
-    final case class MalformedTypingResult(msg: String)               extends TestingError
+    final case class TestDataGenerationError(msg: String)  extends TestingError
```
jedrz, line 160: Wouldn't it be simpler to have a generic error in the endpoint layer with a message? Domain/service errors would be mapped directly to the message, rather than to an endpoint error and then mapping the endpoint error back to the message. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,41 @@
...
+    val trimmedRawTestRecord = rawTestRecord.take(300)
+    if (trimmedRawTestRecord.length < rawTestRecord.length) {
+      s"Could not parse record (shows fragment): '$trimmedRawTestRecord'"
+    } else {
+      s"Could not parse record: '$rawTestRecord'"
+    }
+  }
+
+  def missingSourceForRecord(sourceId: String, recordIndex: Int) =
+    s"Record ${recordIndex + 1} - scenario does not have source id: '$sourceId'"
```
arkadius: It took me a while to understand what this message mean. Let's rephrase it a bit. Some suggestions:
- "Source with id [$sourceId] doesn't exist for record ${recordIndex + 1}"
- "Problem in record  ${recordIndex + 1} detected: Source [$sourceId] doesn't exist"

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  val noDataGenerated = "No data was generated"
+
+  val noInputRecords = "No input records found"
+
+  def requestedTooManySamplesToGenerate(maxSamples: Int) =
+    s"Too many samples requested, limit is $maxSamples. Please configure 'testDataSettings.maxSamplesCount' to increase the limit"
+
+  def tooManyCharactersGenerated(length: Int, limit: Int) =
+    s"$length characters were generated, limit is $limit. Please configure 'testDataSettings.testDataMaxLength' to increase the limit"
```
arkadius: I missed the most important information - "Too many characters generated". Let's start the sentence with this. The same in two messages below. 
Also, the same as with `noDataGenerated`, the obvious question is why you, machine, generated too many characters? Did you have some internal problem and generate them in some infinite loop? ;)  Better message would be for example:
"Too many characters were found in the sample of data when trying to provide $maxSample sample"

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,41 @@
+package pl.touk.nussknacker.ui.api
+
+object TestingApiErrorMessages {
+
+  val noSourcesWithTestDataGeneration = "No sources with test data generation available"
+
+  val noDataGenerated = "No data was generated"
```
arkadius: This message is funny (I know that it is not yours). Take a look:
- We ask test generator to generate data
- Test generators answers: no data was generated

The obvious question is, why? your are a machine, generate it for me! :)

The root problem is that we don't really generate data, but we provide some data samples. In all situations sample of live data, but it is another story.
To make it clear, we can:
1. Start to call this operation "fetch sample of live data" instead of generate data. Then the message would be "No sample live data found"
2. Call this operation "provide sample data" without specifying that it is fetched and it is live. In this case the message would be "Cannot provide sample data. Possible cause: no live sample data available"

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  val noInputRecords = "No input records found"
```
arkadius: Can you group messages by two categories:
1) Messages related by providing sample data by us 
2) Messages related with data passed by user
?
It took me a while to understand what is the difference between noInputRecords vs noDataGenerated, tooManyCharactersGenerated vs tooManyCharactersReceived

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def requestedTooManySamplesToGenerate(maxSamples: Int) =
+    s"Too many samples requested, limit is $maxSamples. Please configure 'testDataSettings.maxSamplesCount' to increase the limit"
+
+  def tooManyCharactersGenerated(length: Int, limit: Int) =
+    s"$length characters were generated, limit is $limit. Please configure 'testDataSettings.testDataMaxLength' to increase the limit"
+
+  def tooManyCharactersReceived(length: Int, limit: Int) =
+    s"Received $length characters, limit is $limit. Please configure 'testDataSettings.testDataMaxLength' to increase the limit"
```
arkadius: We should talk to user from their perspective, not us. So it should be rather passed instead of received, but personally I would skip this part and focus on where these characters were I mean:
"Test data/sample has too many characters ..."

jedrz: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
jedrz: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
jedrz: Changed to second suggestion.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
jedrz: Finally I changed messages to "Problem in sample ${recordIndex + 1} detected: ..." to use only 'sample' word instead of using 'sample' and 'record'  interchangeably.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+
+    def missingSource(sourceId: String): String =
+      messageForRecord(s"source with id '$sourceId' doesn't exist in the scenario")
+
+    def multipleSourcesRequired: String =
+      messageForRecord("scenario has multiple sources, but got sample with unspecified source id")
+
+    private def messageForRecord(message: String) =
```
arkadius: messageForSample

arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
jedrz: Done

PR #7592, opened 2025-02-25
[NU-2048] Fix for invalid classloader used for resources in K8sDeploymentManager

arkadius:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
@@ -13,6 +13,8 @@ import pl.touk.nussknacker.engine.definition.component.{
 }
 import pl.touk.nussknacker.engine.definition.component.Components.ComponentDefinitionExtractionMode
 import pl.touk.nussknacker.engine.deployment.EngineSetupName
+import pl.touk.nussknacker.engine.util.ThreadUtils
+import pl.touk.nussknacker.engine.util.loader.DeploymentManagersClassLoader
 import pl.touk.nussknacker.restmodel.scenariodetails.ScenarioParameters
 import pl.touk.nussknacker.ui.db.DbRef
 import pl.touk.nussknacker.ui.process.periodic.{PeriodicDeploymentManagerDecorator, SchedulingConfig}
@@ -58,6 +60,7 @@ object ProcessingTypeData {
       deploymentManagerProvider: DeploymentManagerProvider,
       schedulingForProcessingType: SchedulingForProcessingType,
       deploymentManagerDependencies: DeploymentManagerDependencies,
+      deploymentManagersClassLoader: DeploymentManagersClassLoader,
       engineSetupName: EngineSetupName,
       deploymentConfig: Config,
       category: String,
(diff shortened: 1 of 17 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
```diff
@@ -97,20 +101,24 @@ object ProcessingTypeData {
...
 
     val validDeploymentManager = for {
-      deploymentManager <- deploymentManagerProvider.createDeploymentManager(
-        modelData,
-        deploymentManagerDependencies,
-        deploymentConfig,
-        scenarioStateCacheTTL
-      )
+      deploymentManager <-
+        ThreadUtils.withThisAsContextClassLoader(deploymentManagersClassLoader) {
```
coutoPL: as we agreed, let's move this code to some deployment manager creator decorator close to the service load code

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/DeploymentManagerProviderCorrectClassloaderHandler.scala
```diff
@@ -0,0 +1,51 @@
...
+import pl.touk.nussknacker.engine.api.deployment.DeploymentManager
+import pl.touk.nussknacker.engine.deployment.EngineSetupName
+import pl.touk.nussknacker.engine.util.ThreadUtils
+import pl.touk.nussknacker.engine.util.loader.DeploymentManagersClassLoader
+
+import scala.concurrent.duration.FiniteDuration
+
+// We wrap created DeploymentManagerProvider with this handler to make sure that DM creation logic will use
+// correct context classloader
+class DeploymentManagerProviderCorrectClassloaderHandler(
```
coutoPL: we can add `private [loader]` to hint that it should not be used in other places.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/DeploymentManagerProviderCorrectClassloaderHandler.scala
```diff
@@ -0,0 +1,51 @@
...
+      dependencies: DeploymentManagerDependencies,
+      deploymentConfig: Config,
+      scenarioStateCacheTTL: Option[FiniteDuration]
+  ): ValidatedNel[String, DeploymentManager] = {
+    ThreadUtils.withThisAsContextClassLoader(deploymentManagersClassLoader) {
+      delegate.createDeploymentManager(modelData, dependencies, deploymentConfig, scenarioStateCacheTTL)
+    }
+  }
+
+  override def metaDataInitializer(config: Config): MetaDataInitializer = delegate.metaDataInitializer(config)
```
coutoPL: are we sure the other methods don't need the changed context class loader?
I'm not sure, but it seems we cannot exactly predict what the user (who implemented the deployment manager) can do, so wouldn't be safe to change the context class loaded in all these places?

arkadius: No, we aren't. I agree that wrapping all method invocations will be safer - I did it. We still can shoot in our feet in some cases - for example when we use lazy vals. I added a comment describing limitations of this solution.

coutoPL (approved):
LGTM

PR #7607, opened 2025-02-28
[NU-2029] explicit schema name in Slick tables and Slick plain sqls

mk-software-pl:
## Describe your changes

In this PR:
1. All SQL statements should use qualified schema name (the slick generated ones and the slick plain too)
2. new SBT task `checkSqlSchemasInStagedScalaFiles` (based on scalafix) that checks if schema was used in the Slick tables and plain SQL queries

## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/db/migration/SlickMigration.scala
```diff
@@ -29,11 +30,29 @@ trait SlickMigration extends BaseJavaMigration {
...
+      case other =>
+        throw new IllegalArgumentException(s"NuJdbcProfile doesn't support ${other.getClass} underlying profile")
+    }
+  }
+
+}
+
+object SlickMigration {
+
+  private val configuredAppSchema = new AtomicReference[String](DbRef.defaultSchemaName)
```
jedrz: Maybe it would be safer to start with uninitialized schema?

designer/server/src/test/scala/pl/touk/nussknacker/test/base/db/DbTesting.scala
```diff
@@ -70,7 +71,7 @@ trait DbTesting extends BeforeAndAfterEach with BeforeAndAfterAll {
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    DatabaseInitializer.initDatabase("db", testDbConfig)
+//    DatabaseInitializer.initDatabase("db", testDbConfig)
```
jedrz: Should it be commented out?

mk-software-pl: I wanted to test if we need it. Because db init is a part of DbRef#create. We run the create above. 
Tests work, so we don't need it. 

So, I'm removing the beforeAll() implementation.

jedrz (approved):

PR #7620, opened 2025-03-07
Allow unlimited number of sticky notes - maxNotesCount is now optional

philemone:
## Describe your changes
`maxNotesCount` type changed to `Optional`. if no value is passed then there is no limit on sticky notes count in scenario.

## Checklist before merge


docs/MigrationGuide.md
```diff
@@ -6,6 +6,8 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 ### Configuration changes
 
+* [#7620](https://github.com/TouK/nussknacker/pull/7620) maxNotesCount in stickyNotesSettings is now optional - if not set the amount of sticky notes in scenario is not limited
```
mslabek: Sticky notes will be introduced in unreleased yet 1.19 so I don't think we need a migration guide entry for a change thats under the same version. Maybe just add the info about it being optional under the `[#7181](https://github.com/TouK/nussknacker/pull/7181) Added designer configuration: stickyNotesSettings` entry?

mslabek (approved):

docs/MigrationGuide.md
philemone: Right, we were using sticky note from staging for so long I forgot it was not released yet :)

PR #7621, opened 2025-03-07
[NU-2015] Fix for component labels with prefix

ForrestFairy:
## Describe your changes
`ComponentProvider` can add prefix to component names. This is added after the creation of `ComponentDefinition` so `label` was created before this addition. 

## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/ComponentDefinitionWithImplementation.scala
```diff
@@ -33,7 +33,7 @@ trait ComponentDefinitionWithImplementation extends ObjectOperatingOnTypes {
 
   protected def uiDefinition: ComponentUiDefinition
 
-  final def label: String = {
+  final def labelOrDefault: String = {
```
arkadius: 1. I think that following the other methods approach, we can still name it `label` as it was before
2. During `ComponentUiDefinition` creation is the place where we should do this fallback. Take a look that in that place everything is well known, determined based on config and other things. BTW you won't need to do any hacking with fragments than

arkadius: This `Ui` in `ComponentUiDefinition` means that we create an object that represents the situation that will be visible in UI, whereas `ComponentDefinition` represents properties of component declared by developer of component extension.

arkadius (approved):
nice

PR #7626, opened 2025-03-10
[NU-2071] Add test capabilities for event generator source

mgoworko:
## Describe your changes

Added test capabilities (tests from file and ad-hoc) for event generator source:

## Checklist before merge


engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -90,6 +104,37 @@ class EventGeneratorSourceFactory(customTimestampAssigner: TimestampWatermarkHan
       override def timestampAssigner: Option[TimestampWatermarkHandler[AnyRef]] = Some(customTimestampAssigner)
 
       override val returnType: typing.TypingResult = value.returnType
+
+      // Test data generation returns a single batch of samples, as specified in the Event Generator source
+      override def generateTestData(size: Int): TestData = TestData(
+        List.fill(size)(TestRecord(value.returnType.asJson))
```
arkadius: You didn't generate values but json representation of type inferred based on values :D Did you use LLM?
`LazyParameter` is a kind of function evaluating expression + inferred nussknacker type based on the expression. Normally, you should be able to invoke `.evaluate(Context("dumb"))` on it. It should evaluate expression with empty set of variables, but ... `LazyParameter`'s created for flink purpose have to be serializable. Because of that it is not so easy to evaluate this variable. See `EvaluableLazyParameterCreator` and `DefaultToEvaluateFunctionConverter` 
You could use `value.returnType.valueOpt` instead, but it will be available only if value for expression can be statically determined. If someone would use some random number generator etc., it won't work

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -90,6 +104,37 @@ class EventGeneratorSourceFactory(customTimestampAssigner: TimestampWatermarkHan
...
+      }
+
+      // No parameters are required for ad-hoc tests. Ad-hoc test generates a single batch of samples,
+      // with count and value as specified in the Event Generator source
+      override def testParametersDefinition: List[Parameter] = List.empty
+
+      override def parametersToTestData(params: Map[ParameterName, AnyRef]): AnyRef =
+        value.returnType.valueOpt match {
+          case Some(value) =>
+            List.fill(count)(value.asInstanceOf[AnyRef])
```
arkadius: Hmm ... LLM? this `value` should be renamed to `valueParamterExpressionWithInferredType` and `LazyParamter` should be probable renamed to `EvaluableExpressionWithInferredType`. Looking from this perspective, there is no sense in passing an expression as a sample of test data.

arkadius (commented):
Some test?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
mgoworko: `You could use value.returnType.valueOpt` - that is exactly what I did. The solution works for statically determinable expressions. I did not test it on dynamic expressions, will have to work on it.

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
mgoworko: As in previous comment - it worked only for statically determinable expressions, so the solution was wrong and I will have to change it.

build.sbt
```diff
@@ -655,6 +655,7 @@ lazy val flinkDevModel = (project in flink("management/dev-model"))
   .dependsOn(
     extensionsApi,
     commonComponents,
+    flinkBaseUnboundedComponents,
```
mgoworko: This dependency is needed to use Event Generator in designer tests

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
mgoworko: Now using ValueDecoder and ValueEncoder (different than what we discussed, the Any encoder/decoder was not working properly types like dates etc.

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
mgoworko: Now encoding value, with tests for various expressions in `value`

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Activities should display activities #0.png
mgoworko, line 1: Adding new available components from the flinkBaseUnboundedComponents module to the tests caused the snapshots to change

build.sbt
arkadius: It is not a good way to go, components modules are added dynamically, not statically, you should change the classpath inside the config to make them work instead of adding dependency

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
arkadius: `ValueEncoder` and `ValueDecoder` was designed to handle encoding of values inferred by our typing system (this "value" in name is not a coincidence). I don't know what assumptions were during designing this encoding mechanism, but it doesn't look like a good choice for encoding of a generic variable passing through the Nussknacker scenario. Let's dig into examples where `ToJsonEncoder`/`FromJsonDecoder` behave wrongly and discuss if we have to either:
1. change its behaviour
2. extract some another encoder designed for our purpose
3. make `ValueEncoder`/`ValueDecoder` more general purpose - rename, describe assumptions and consider if other places where  `ToJsonEncoder`/`FromJsonDecoder` should be replaced by this classes usages

build.sbt
mgoworko: Fixed, without dependency

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
mgoworko: Separate PR with proposition of encoder/decoder refactors https://github.com/TouK/nussknacker/pull/7653/checks

designer/server/src/test/resources/config/business-cases/simple-streaming-use-case-designer.conf
```diff
@@ -8,105 +8,17 @@ authentication:  {
 baseModelConfig {
   classPath: [
     "engine/flink/management/dev-model/target/scala-"${scala.major.version}"/devModel.jar",
+    "engine/flink/components/base-unbounded/target/scala-"${scala.major.version}"/flinkBaseUnbounded.jar",
```
arkadius, line 12: Can you replace `devModel.jar` with:
```
"defaultModel/target/scala-"${scala.major.version}"/defaultModel.jar"
```
?

mgoworko: Done
- It changed the available SPEL utils, I changed the helper used in tests
- I had to add additionally flinkBase.jar to have "dead-end" sink

arkadius: Nice, thanks, one step closer to untangle this mess around dev model

arkadius (approved):

PR #7650, opened 2025-03-13
[Nu-2015] Fragments label fix

ForrestFairy:
## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/defaultconfig/DefaultComponentConfigDeterminer.scala
```diff
@@ -80,7 +80,7 @@ object DefaultComponentConfigDeterminer {
       icon = icon.getOrElse(DefaultsComponentIcon.FragmentIcon),
       docsUrl = docsUrl,
       designerWideId = designerWideId,
-      label = IdToTitleConverter.toTitle(name)
+      label = name
```
arkadius, line 84: Please leave a comment, that in fragments (the same as scenario), the name is an id as well as label, so othere devs won't introduce this regression by mistake. Also would be nice to add a regression test in `DefinitionResourcesSpec`

arkadius (approved):

PR #7658, opened 2025-03-13
[NU-1979] DeploymentManager excluded from processing type data reloading

arkadius:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/SchedulingDependencies.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+final class SchedulingDependencies(
+    val dbRef: DbRef,
+    val actionService: ProcessingTypeActionService,
+    val fetchingProcessRepository: FetchingProcessRepository[Future],
+    val configsFromProvider: Map[DesignerWideComponentId, ComponentAdditionalConfig]
+)
+
+trait ProcessingTypeActionService {
```
arkadius, line 20: It is moved from deployment-manager-api here, because it is related with scheduling mechanism only

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProviderTest.scala
```diff
@@ -29,21 +30,12 @@ class ReloadableProcessingTypeDataProviderTest extends AnyFunSuiteLike with Matc
...
-    val allDataBeforeLoad = reloadableProvider.all
-    allDataBeforeLoad shouldBe empty
-    valueHolder.closeWasInvoked shouldBe false
-    an[Exception] shouldBe thrownBy {
-      reloadableProvider.combined
-    }
-
-    // first load
-    reloadableProvider.reloadAll.unsafeRunSync()
+    // data are loaded at the beginning
```
arkadius, line 35: Now, model data are loaded also in constructor, so this test has been changed a bit

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkScheduledExecutionPerformer.scala
```diff
@@ -63,7 +63,9 @@ class FlinkScheduledExecutionPerformer(
   }
 
   override def provideInputConfigDuringExecutionJson(): Future[InputConfigDuringExecution] =
-    Future.successful(inputConfigDuringExecution)
+    Future {
+      getCurrentInputConfigDuringExecution()
```
arkadius, line 67: During reading this config is used a strange model extension which can be reloaded so we have to access it lazily.

engine/lite/k8sDeploymentManager/src/main/scala/pl/touk/nussknacker/k8s/manager/K8sDeploymentManager.scala
```diff
@@ -92,8 +91,8 @@ class K8sDeploymentManager(
   private val ingressPreparerOpt          = config.ingress.map(new IngressPreparer(_, config.nussknackerInstanceName))
 
   // runtime config is combined from scenarioType.modelConfig and deploymentConfig.configExecutionOverrides
-  private val serializedRuntimeConfig = {
-    val inputConfig     = modelData.inputConfigDuringExecution
+  private def serializedRuntimeConfig() = {
+    val inputConfig     = modelDataProvider.getCurrentModelData().inputConfigDuringExecution
```
arkadius, line 95: The same story here

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/ModelData.scala
```diff
@@ -228,9 +238,8 @@ trait ModelData extends BaseModelData with AutoCloseable {
   // See parameters of implementing functions
   def extractModelDefinitionFun: ExtractDefinitionFun
 
-  final def modelDefinition: ModelDefinition = withThisAsContextClassLoader {
+  final def modelDefinition: ModelDefinition =
```
arkadius, line 241: This `withThisAsContextClassLoader` was some mistake, because `modelDefinitionWithClasses` is already created with  `withThisAsContextClassLoader`

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerDependencies.scala
```diff
@@ -2,20 +2,17 @@ package pl.touk.nussknacker.engine
...
-import pl.touk.nussknacker.engine.api.component.{ComponentAdditionalConfig, DesignerWideComponentId}
-import pl.touk.nussknacker.engine.api.deployment.{ProcessingTypeActionService, ProcessingTypeDeployedScenariosProvider}
+import pl.touk.nussknacker.engine.api.deployment.ProcessingTypeDeployedScenariosProvider
 import sttp.client3.SttpBackend
 
 import scala.concurrent.{ExecutionContext, Future}
 
 case class DeploymentManagerDependencies(
     deployedScenariosProvider: ProcessingTypeDeployedScenariosProvider,
-    actionService: ProcessingTypeActionService,
```
arkadius, line 13: Removed dependencies were only for scheduling mechanism purpose - I moved them there

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProvider.scala
```diff
@@ -11,19 +11,16 @@ import scala.util.control.NonFatal
...
  * to be able to wait for all operations to complete
- *
- * Another thing that needs careful consideration is handling exception during ProcessingTypeData creation/closing - probably during
- * close we want to catch exception and try to proceed, but during creation it can be a bit tricky...
  */
 class ReloadableProcessingTypeDataProvider[Data <: AutoCloseable, CombinedData] private (
     loadMethod: IO[ProcessingTypeDataState[Data, CombinedData]]
 )(implicit ioRuntime: IORuntime)
-    extends ProcessingTypeDataProvider[Data, CombinedData](ProcessingTypeDataState.uninitialized)
+    extends ProcessingTypeDataProvider[Data, CombinedData](loadMethod.unsafeRunSync())
```
arkadius: After splitting deployment managers and models, we don't need to postpone model reload

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/DeploymentData.scala
```diff
@@ -2,29 +2,23 @@ package pl.touk.nussknacker.ui.process.processingtype
...
+    val additionalValidators: List[CustomProcessValidator],
+    val engineSetupName: EngineSetupName
 ) {
 
   def validDeploymentManagerOrStub: DeploymentManager =
     validDeploymentManager.valueOr(_ => InvalidDeploymentManagerStub)
 
   def engineSetupErrors: List[String] = validDeploymentManager.swap.map(_.toList).valueOr(_ => List.empty)
 
-  def close(): Unit = {
```
arkadius, line 24: DeploymentManager.close is now handled by `Resource` in `DomainServices`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
```diff
@@ -40,11 +34,6 @@ final case class ProcessingTypeData private (
       deploymentData.engineSetupErrors
     )
 
-  override def close(): Unit = {
```
arkadius, line 43: Now, `ModelData.close()` is used directly during reload

designer/listener-api/src/main/scala/pl/touk/nussknacker/ui/listener/ProcessChangeListener.scala
```diff
@@ -5,15 +5,3 @@ import scala.concurrent.ExecutionContext
 trait ProcessChangeListener {
   def handle(event: ProcessChangeEvent)(implicit ec: ExecutionContext, user: User): Unit
 }
-
-object ProcessChangeListener {
-
-  def noop: ProcessChangeListener = new ProcessChangeListener {
```
arkadius, line 11: wasn't used anywhere

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/DomainServices.scala
```diff
@@ -90,66 +96,63 @@ object DomainServices {
...
+            additionalUIConfigProvider,
+            _
+          )
+        )
       )
-      _ <- DeploymentsStatusesSynchronizationScheduler.resource(
-        actorSystem,
-        deploymentsStatusesSynchronizer,
-        alreadyLoadedConfig.deploymentsStatusesSynchronizationConfig
+      // ActionService initialization
```
arkadius, line 149: I changed the order a bit to make more clear:
- when ActionService is ready to use
- what services are based on `ProcessingTypeData`

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerDependencies.scala
```diff
@@ -2,20 +2,17 @@ package pl.touk.nussknacker.engine
...
     executionContext: ExecutionContext,
```
coutoPL: If we're changing this class, maybe we should use `ExecutionContextWithIORuntime` instead of separate ExecutionContext and IORuntime?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerDependencies.scala
```diff
@@ -2,20 +2,17 @@ package pl.touk.nussknacker.engine
...
 import cats.effect.unsafe.IORuntime
 import org.apache.pekko.actor.ActorSystem
-import pl.touk.nussknacker.engine.api.component.{ComponentAdditionalConfig, DesignerWideComponentId}
-import pl.touk.nussknacker.engine.api.deployment.{ProcessingTypeActionService, ProcessingTypeDeployedScenariosProvider}
+import pl.touk.nussknacker.engine.api.deployment.ProcessingTypeDeployedScenariosProvider
 import sttp.client3.SttpBackend
 
 import scala.concurrent.{ExecutionContext, Future}
 
 case class DeploymentManagerDependencies(
```
coutoPL: If we're changing this class, maybe we can get rid of `case`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
```diff
@@ -201,12 +201,13 @@ class AppApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoint
...
-  lazy val processingTypeDataReloadEndpoint: SecuredEndpoint[Unit, Unit, Unit, Any] =
+  lazy val modelDataReloadEndpoint: SecuredEndpoint[Unit, Unit, Unit, Any] =
     baseNuApiEndpoint
-      .summary("Processing type data reload service")
+      .summary("Model data reload service")
       .tag("App")
       .withSecurity(auth)
       .post
+      // TODO: rename: to modeldata / reload
       .in("app" / "processingtype" / "reload")
```
coutoPL: maybe we we can add a new `app/modeldata/reload` endpoint and mark this one as deprecated?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/DesignerModelData.scala
```diff
@@ -1,12 +1,14 @@
 package pl.touk.nussknacker.ui.process.processingtype
 
 import pl.touk.nussknacker.engine.ModelData
-import pl.touk.nussknacker.engine.api.component.{ComponentId, ProcessingMode}
+import pl.touk.nussknacker.engine.api.component.{ComponentId, ProcessingMode, ScenarioPropertyConfig}
 import pl.touk.nussknacker.engine.definition.component.ComponentStaticDefinition
 import pl.touk.nussknacker.ui.process.processingtype.DesignerModelData.DynamicComponentsStaticDefinitions
 
 final case class DesignerModelData(
```
coutoPL: it seems we should remove the `case` here too

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/DeploymentManagersLoader.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  def load(
+      processingTypeConfigs: ProcessingTypeConfigs,
+      deploymentManagersClassLoader: DeploymentManagersClassLoader,
+      modelClassLoaderProvider: ModelClassLoaderProvider,
+      modelDataProviders: ProcessingTypeDataProvider[ModelData, _],
+      getDeploymentManagerDependencies: ProcessingType => DeploymentManagerDependencies,
+      schedulingDepsProvider: Option[ProcessingType => SchedulingDependencies],
+  ): Resource[IO, Map[ProcessingType, ValueWithRestriction[DeploymentData]]] = {
+    processingTypeConfigs.configByProcessingType.toList
```
coutoPL, line 32: Let's refactor it a little bit for the sake of readability.
It seems that (at least) we can create some well-named private methods for these two `map`s:

Screenshot 2025-03-19 at 11 34 27

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ModelDataLoader.scala
```diff
@@ -0,0 +1,50 @@
...
+import pl.touk.nussknacker.ui.process.processingtype.loader.ProcessingTypeDataStateFactory.ModelDataWithProcessingTypeDataInput
+import pl.touk.nussknacker.ui.process.processingtype.provider.ProcessingTypeDataState
+
+object ModelDataLoader extends LazyLogging {
+
+  def load(
+      processingTypeConfigs: ProcessingTypeConfigs,
+      getModelDependencies: ProcessingType => ModelDependencies,
+      modelClassLoaderProvider: ModelClassLoaderProvider,
+  ): ProcessingTypeDataState[ModelDataWithProcessingTypeDataInput, _] = {
```
coutoPL, line 21: the same comment here

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypeDataStateFactory.scala
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  private implicit class ValueWithRestrictionOps[T](valueWithRestriction: ValueWithRestriction[T]) {
+    // This is a hack, we should split deployment data and model data
+    def bypassRestriction: T =
+      valueWithRestriction.valueWithAllowedAccess(Permission.Read)(NussknackerInternalUser.instance).get
+  }
+
+  final case class ModelDataWithProcessingTypeDataInput(
```
coutoPL: Let's get rid of the `case`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypeDataStateFactory.scala
```diff
@@ -0,0 +1,76 @@
...
+      modelData: ModelData,
+      category: String,
+      componentDefinitionExtractionMode: ComponentDefinitionExtractionMode
+  ) extends AutoCloseable {
+
+    override def close(): Unit = {
```
coutoPL, line 70: It seems it's easy to move the closing part to the  ModelDataLoader class and return Resource from its `load` method

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/provider/ReloadableProcessingTypeDataProvider.scala
coutoPL: we can avoid calling the `unsafeRunSync` by returning IO by the `apply` method (it would be good to consider renaming it)

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/TestProcessingTypeDataProviderFactory.scala
```diff
@@ -34,7 +49,44 @@ object TestProcessingTypeDataProviderFactory {
...
+  ): ProcessingTypeDataProvider[ProcessingTypeData, CombinedProcessingTypeData] = {
+    val finalProcessingTypeData =
+      ModelDataLoader
+        .load(
+          processingTypeConfigs,
+          _ => modelDependencies,
+          modelClassLoaderProvider
+        )
+        .transform { case (modelDataWithInputs, _) =>
+          val deploymentData = DeploymentManagersLoader
```
coutoPL: 1. I know it's test code, but maybe we can move the `DeploymentManagersLoader.load(...)` and `ProcessingTypeDataStateFactory.create` to private methods for the sake of readability? 

2. let's avoid using _1 by doing it like that: 
```
val (deploymentData, _) = ...
```

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/ProcessStateDefinitionServiceSpec.scala
```diff
@@ -184,22 +186,25 @@ class ProcessStateDefinitionServiceSpec extends AnyFunSuite with Matchers {
       stateDefinitionManager: ProcessStateDefinitionManager,
       category: String
   ): ProcessingTypeData = {
+    val mockManager = MockDeploymentManager.create(
+      customProcessStateDefinitionManager = Some(stateDefinitionManager)
+    )
+    val deploymentData = new DeploymentData(
+      DeploymentManagerType("mock"),
```
coutoPL, line 193: let's use named params

docs-internal/api/nu-designer-openapi.yaml
```diff
@@ -345,12 +345,12 @@ paths:
                         process-version: '0.1'
                         engine-version: '0.2'
                         generation-time: '2023-09-25T09:26:30.402299'
-  /api/app/healthCheck/process/deployment:
```
coutoPL, line 348: does it mean there was some problem with open API generation?

extensions-api/src/main/scala/pl/touk/nussknacker/engine/BaseModelData.scala
```diff
@@ -4,24 +4,28 @@ import com.typesafe.config.Config
...
-
   def modelConfig: Config
 
-  def modelClassLoader: URLClassLoader
+}
 
-  final def modelClassLoaderUrls: List[URL] = modelClassLoader.getURLs.toList
+trait BaseModelDataProvider {
+  // modelClassLoader is set once and won't be changed
+  val modelClassLoader: URLClassLoader
```
coutoPL: can we use `ModelClassLoader` class instead?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypeDataStateFactory.scala
arkadius: It not so easy. AutoClosable is required by `ReloadableProcessingTypeDataProvider` here: `Data <: AutoCloseable`. We could remove this restriction and pass `loadMethod: Resource[]` instead, but TBH I don't know how to handle it in reload. Do you have some idea how to do it?

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/DeploymentManagerDependencies.scala
arkadius: `DeploymentManagerDependencies` are exposed in deployment-manager-api. I'm not sure if it is a good idea to move `ExecutionContextWithIORuntime` to api modules. Especially, if I do this, still we need a constructor variant with both parameters because this class is quite often used in tests where people checking if their DeploymentManager works with this approach:
```
new DeploymentManagerDependencies(
(...)
      ExecutionContext.global,
      IORuntime.global,
```
which in tests, IMO is a pretty good approach. WDYT?

docs-internal/api/nu-designer-openapi.yaml
arkadius: I don't think so. Only the order has been changed. Or maybe you see some other problem?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
```diff
@@ -220,6 +222,26 @@ class AppApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoint
         )
       )
 
+  lazy val modelReloadEndpoint: SecuredEndpoint[Unit, Unit, Unit, Any] =
+    baseNuApiEndpoint
+      .summary("Model reload service")
+      .tag("App")
+      .withSecurity(auth)
+      .post
+      // TODO: rename: to modeldata / reload
```
coutoPL: why don't we do that now?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
```diff
@@ -5,32 +5,66 @@ import com.typesafe.scalalogging.LazyLogging
...
 object PeriodicDeploymentManagerDecorator extends LazyLogging {
 
+  def decorateIfSchedulingSupported(
+      underlying: DeploymentManager,
+      deploymentConfig: Config,
+      dependencies: DeploymentManagerDependencies,
+      schedulingDepsOpt: Option[SchedulingDependencies]
+  ): (DeploymentManager, Map[String, ScenarioPropertyConfig]) = {
+    val schedulingForProcessingType =
+      if (deploymentConfig.hasPath("scheduling") &&
```
coutoPL: let's move this `if` to a well-named, private method

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
```diff
@@ -5,32 +5,66 @@ import com.typesafe.scalalogging.LazyLogging
...
+            val decoratedManager = PeriodicDeploymentManagerDecorator.decorate(
+              underlying = underlying,
+              schedulingSupported = supported,
+              deploymentConfig = deploymentConfig,
+              dependencies = dependencies,
+              schedulingDeps = schedulingDeps,
+            )
+            (decoratedManager, PeriodicDeploymentManagerDecorator.additionalScenarioProperties)
+          case NoSchedulingSupport =>
+            throw new IllegalStateException(
```
coutoPL, line 42: I wonder if we should throw the IllegalStateException here. Maybe we can return the underlying instance with the empty Map

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/TestProcessingTypeDataProviderFactory.scala
```diff
@@ -34,7 +42,77 @@ object TestProcessingTypeDataProviderFactory {
...
+      processingTypeConfigs: ProcessingTypeConfigs,
+      modelClassLoaderProvider: ModelClassLoaderProvider,
+      modelDependencies: ModelDependencies,
+      deploymentManagersClassLoader: DeploymentManagersClassLoader,
+      deploymentManagerDependencies: DeploymentManagerDependencies,
+  ): ProcessingTypeDataProvider[ProcessingTypeData, CombinedProcessingTypeData] = {
+    val finalProcessingTypeData =
+      loadModelData(processingTypeConfigs, modelClassLoaderProvider, modelDependencies)
+        .transform { case (modelDataWithInputs, _) =>
+          val (deploymentData, _) = loadDeploymentManagers(
```
coutoPL: We ignore the closing handler here, which may lead to some leaks in tests.

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/classloader/DeploymentManagersClassLoaderFactory.scala
```diff
@@ -1,14 +1,12 @@
...
 import cats.effect.{IO, Resource}
 import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.util.UrlUtils.ExpandFiles
 
-import java.net.URL
 import java.nio.file.{Files, Path}
-import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
 
-object DeploymentManagersClassLoader extends LazyLogging {
+object DeploymentManagersClassLoaderFactory extends LazyLogging {
```
coutoPL, line 9: TBH, I'd instead create the DeploymentManagersClassLoader compation object extension method instead of the factory (without interface). But I see that we use both approaches, so I'm not going to insist on this.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/loader/ProcessingTypeDataStateFactory.scala
coutoPL: ok, let's leave it as it's

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
arkadius: wops, I forgot to clean it up

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
arkadius: Returning underlying instance is not a good way to go. This exception will be thrown when an admin configures scheduling and DeploymentManager doesn't support it.

designer/server/src/test/scala/pl/touk/nussknacker/test/utils/domain/TestProcessingTypeDataProviderFactory.scala
arkadius: I added more monads in tests ;-) , take a look now

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/classloader/DeploymentManagersClassLoaderFactory.scala
arkadius: This factories exsists, because:
1. In API should be the smallest possible set of classes and methods in order to avoid incompatible changes in the future
2.  `DeploymentManagersClassLoaderFactory` uses cats effects and some internal utils such as `UrlUtils`. If I left this as companion object it would cause more dependencies in api modules

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicDeploymentManagerDecorator.scala
coutoPL: I guess it could be handled better (at the type level, or simply earlier), but it seems it does protect us against the misconfiguration you mention, so it's fair enough.

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/classloader/DeploymentManagersClassLoaderFactory.scala
coutoPL: I agree with the above. I rather meant sth like this:
```
implicit DeploymentManagersClassLoaderExt(val companionObject: DeploymentManagersClassLoader.type) {
   def create(...) 
}
```
Then you can call it like that:
```
DeploymentManagersClassLoader.create(...) 
```
And put in a different module.

coutoPL (approved):
LGTM

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/classloader/DeploymentManagersClassLoaderFactory.scala
arkadius: I see, it is an interesting approach but by looking into the code, we would rather use separate factory/loader classes for such purposes, so let's keep it consistent

PR #7663, opened 2025-03-14
[NU-2071] Event generator dry run tests

mgoworko:
## Describe your changes

There is a problem with "ad-hoc" tests as implemented for the Event Generator in the previous change. This change aims to fix that.

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/Source.scala
```diff
@@ -49,6 +49,26 @@ trait TestWithParametersSupport[+T] { self: Source =>
   def parametersToTestData(params: Map[ParameterName, AnyRef]): T
 }
 
+/**
+ * Optional functionality which should provide field definitions based on input schema
+ * Based on those fields UI creates a window allowing user to test scenario based on schema.
+ */
+// This trait extends TestWithParametersSupport in order to preserve binary compatibility.
```
arkadius: Ignore binary compatibility and change it in this change. We can discuss why this is better approach, but I'd rather prefer to do this offline, because it is a bit longer discussion

mgoworko: I managed to create EvaluableLazyParameterCreator's (prepare them for resolving by converting to EvaluableLazyParameter's) in the `scenario-compiler`, in a way that is acceptable for Flink serialization. 

That way, during tests we will have param's that are resolvable and we can generate ad-hoc value based on source value.

arkadius (approved):
Nice, please add PR number next to changelog entry with test capabilities for Event Generator

PR #7690, opened 2025-03-18
[NU-2077] Make dry run schemaless JSON validation possible

Cichan-mcc:
## Describe your changes

Currently when we try to run ad hoc tests with schemaless JSON kafka topic source we aren't presented with a Json editor and the underlying language for the data is SpEL. Because of that the validation doesn't work properly treating JSONs as SpEL.

Screenshot 2025-03-24 at 08 48 08

In these changes I switched the underlying language to a new JSON one and use `JsonParameterEditor` for the ad hoc tests.

Screenshot 2025-03-24 at 08 38 46

Screenshot 2025-03-24 at 08 39 58

In another PR I'll change the `Unknown` type to a `Json` one and also rename the `Value` to `Input` in ad hoc tests as that name would better suit it.

## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
@@ -38,6 +38,7 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
   private def defaultStringExpression(editor: Option[ParameterEditor]): Expression =
     EditorBasedLanguageDeterminer.determineLanguageOf(editor) match {
       case Language.Spel => Expression.spel("''")
+      case Language.Json => Expression.spel("{}")
       case language @ (Language.SpelTemplate | Language.DictKeyWithLabel | Language.TabularDataDefinition) =>
         Expression(language, "")
     }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
@@ -38,7 +38,7 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
   private def defaultStringExpression(editor: Option[ParameterEditor]): Expression =
     EditorBasedLanguageDeterminer.determineLanguageOf(editor) match {
       case Language.Spel => Expression.spel("''")
-      case Language.Json => Expression.spel("{}")
+      case Language.Json => Expression.json("{}")
       case language @ (Language.SpelTemplate | Language.DictKeyWithLabel | Language.TabularDataDefinition) =>
         Expression(language, "")
     }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
@@ -56,6 +56,6 @@ object JsonParser extends ExpressionParser {
 
 case object JsonExpressionTypingInfo extends ExpressionTypingInfo {
 
-  // TODO: Right now we just use Unknown but we should create appropriate typing in the future
+  // TODO: Right now we just use Unknown but we should create appropriate typing for Jsons in the future
   override def typingResult: TypingResult = Unknown
 }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
@@ -1,0 +1,61 @@
+package pl.touk.nussknacker.engine.language.json
+
+import cats.data.{NonEmptyList, Validated, ValidatedNel}
+import cats.data.Validated.{invalidNel, Valid}
+import io.circe.{parser, Json}
+import pl.touk.nussknacker.engine.api.Context
+import pl.touk.nussknacker.engine.api.context.ValidationContext
+import pl.touk.nussknacker.engine.api.expression.ExpressionTypingInfo
+import pl.touk.nussknacker.engine.api.generics.ExpressionParseError
+import pl.touk.nussknacker.engine.api.typed.typing.{TypingResult, Unknown}
+import pl.touk.nussknacker.engine.expression.parse.{CompiledExpression, ExpressionParser, TypedExpression}
+import pl.touk.nussknacker.engine.graph.expression.Expression
+import pl.touk.nussknacker.engine.graph.expression.Expression.Language
+import pl.touk.nussknacker.engine.spel.SpelExpressionParseError.JsonParsingError
+
+object JsonParser extends ExpressionParser {
+
+  override def languageId: Language = Expression.Language.Json
+
+  override def parse(
+      jsonString: String,
+      ctx: ValidationContext,
+      expectedType: TypingResult
+  ): ValidatedNel[ExpressionParseError, TypedExpression] = {
+    parseJson(jsonString).map { json =>
+      TypedExpression(
+        CompiledJsonExpression(jsonString, json),
+        JsonExpressionTypingInfo
+      )
+    }
+  }
+
+  override def parseWithoutContextValidation(
+      jsonString: String,
+      expectedType: TypingResult
+  ): ValidatedNel[ExpressionParseError, CompiledExpression] = {
+    parseJson(jsonString).map(json => CompiledJsonExpression(jsonString, json))
+  }
+
+  private def parseJson(jsonString: String): Validated[NonEmptyList[JsonParsingError], Json] =
+    parser.parse(jsonString) match {
+      case Left(error) => invalidNel(JsonParsingError(error.message))
+      case Right(json) => Valid(json)
+    }
+
+  case class CompiledJsonExpression(originalJsonString: String, json: Json) extends CompiledExpression {
+
+    override def language: Language = languageId
+
+    override def original: String = originalJsonString
+
+    override def evaluate[T](ctx: Context, globals: Map[String, Any]): T = json.asInstanceOf[T]
+  }
+
+}
+
+case object JsonExpressionTypingInfo extends ExpressionTypingInfo {
+
+  // TODO: Right now we just use Unknown but we should create appropriate typing in the future
+  override def typingResult: TypingResult = Unknown
+}
--- a/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
+++ b/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
@@ -30,6 +30,7 @@ class JsonParserTest extends AnyFunSuite with Matchers {
                        |]""".stripMargin
 
     val resultType = parser.parse(validJson, ValidationContext.empty, Unknown).validValue.typingInfo.typingResult
+    // TODO: Right now we just use Unknown but we should create appropriate typing for Jsons in the future
     resultType shouldBe Unknown
   }
 
--- a/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
+++ b/scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
@@ -1,0 +1,50 @@
+package pl.touk.nussknacker.engine.language.json
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import pl.touk.nussknacker.engine.api.context.ValidationContext
+import pl.touk.nussknacker.engine.api.typed.typing.Unknown
+import pl.touk.nussknacker.engine.spel.SpelExpressionParseError.JsonParsingError
+import pl.touk.nussknacker.test.ValidatedValuesDetailedMessage.convertValidatedToValuable
+
+class JsonParserTest extends AnyFunSuite with Matchers {
+
+  private val parser = JsonParser
+
+  test("should parse JSON expression") {
+    val validJson = """|[
+                       |  {
+                       |    "products": [
+                       |      {"id": 1, "name": "Laptop", "price": 1200.00},
+                       |      {"id": 2, "name": "Smartphone", "price": 800.50},
+                       |      {"id": 3, "name": "Tablet", "price": 450.75}
+                       |    ]
+                       |  },
+                       |  {
+                       |    "someObject": {
+                       |      "someString": "some string value",
+                       |      "someBoolean": true,
+                       |      "someNumber": 21.37
+                       |    }
+                       |  }
+                       |]""".stripMargin
+
+    val resultType = parser.parse(validJson, ValidationContext.empty, Unknown).validValue.typingInfo.typingResult
+    resultType shouldBe Unknown
+  }
+
+  test("should return error when JSON cannot be parsed") {
+    // Missing comma after Laptop object entry
+    val invalidJson = """|{
+                         |  "products": [
+                         |    {"id": 1, "name": "Laptop", "price": 1200.00}
+                         |    {"id": 2, "name": "Smartphone", "price": 800.50},
+                         |    {"id": 3, "name": "Tablet", "price": 450.75}
+                         |  ]
+                         |}""".stripMargin
+    val parsingErrors = parser.parse(invalidJson, ValidationContext.empty, Unknown).invalidValue
+    parsingErrors.size shouldBe 1
+    parsingErrors.head shouldBe JsonParsingError("expected ] or , got '{\"id\":...' (line 4, column 5)")
+  }
+
+}
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
@@ -2,10 +2,10 @@ package pl.touk.nussknacker.engine.json
 
 import cats.data.{NonEmptyList, Validated, ValidatedNel}
 import cats.data.Validated.Valid
-import org.everit.json.schema.{EmptySchema, ObjectSchema, Schema}
+import org.everit.json.schema.{ObjectSchema, Schema}
 import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError
-import pl.touk.nussknacker.engine.api.definition.{JsonParameterEditor, Parameter, ParameterEditor}
+import pl.touk.nussknacker.engine.api.definition.Parameter
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.api.validation.ValidationMode
 import pl.touk.nussknacker.engine.graph.expression.Expression
@@ -38,22 +38,6 @@ object JsonSchemaBasedParameter {
       defaultValue = None,
       isRequired = None
     )
-
-  def emptySchemaBasedParameter(
-      defaultParamName: FieldName,
-      validationMode: ValidationMode
-  ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] = {
-    val emptyJsonSchema = EmptySchema.INSTANCE
-    val typing          = SwaggerBasedJsonSchemaTypeDefinitionExtractor.swaggerType(emptyJsonSchema, None).typingResult
-    val parameter = Parameter(defaultParamName, typing)
-      .copy(isLazyParameter = true, editors = List(JsonParameterEditor))
-    Valid(
-      SingleSchemaBasedParameter(
-        parameter,
-        new JsonSchemaOutputValidator(validationMode).validate(_, emptyJsonSchema, None)
-      )
-    )
-  }
 
   private case class ParameterRetriever(
       rootSchema: Schema,
--- a/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
+++ b/utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
@@ -2,10 +2,10 @@ package pl.touk.nussknacker.engine.json
 
 import cats.data.{NonEmptyList, Validated, ValidatedNel}
 import cats.data.Validated.Valid
-import org.everit.json.schema.{ObjectSchema, Schema}
+import org.everit.json.schema.{EmptySchema, ObjectSchema, Schema}
 import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError
-import pl.touk.nussknacker.engine.api.definition.Parameter
+import pl.touk.nussknacker.engine.api.definition.{JsonParameterEditor, Parameter, ParameterEditor}
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.api.validation.ValidationMode
 import pl.touk.nussknacker.engine.graph.expression.Expression
@@ -39,6 +39,16 @@ object JsonSchemaBasedParameter {
       isRequired = None
     )
 
+  def withEmptySchemaSupport(schema: Schema, defaultParamName: FieldName, validationMode: ValidationMode)(
+      implicit nodeId: NodeId
+  ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] =
+    ParameterRetriever(schema, defaultParamName, validationMode).toSchemaBasedParameterWithEmptySchemaSupport(
+      schema,
+      paramName = None,
+      defaultValue = None,
+      isRequired = None
+    )
+
   private case class ParameterRetriever(
       rootSchema: Schema,
       defaultParamName: FieldName,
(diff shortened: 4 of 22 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
```diff
@@ -0,0 +1,60 @@
...
+object JsonParser extends ExpressionParser {
+
+  override def languageId: Language = Expression.Language.Json
+
+  override def parse(
+      jsonString: String,
+      ctx: ValidationContext,
+      expectedType: TypingResult
+  ): ValidatedNel[ExpressionParseError, TypedExpression] = {
+    parseJson(jsonString).map { json =>
```
Cichan-mcc, line 25: Empty string is not a valid JSON and thus when a user enters the ad hoc modal they immediately get a validation error. Should we provide some default value for the JSON ad hoc test? For example an empty object "{}" so they aren't welcomed with an error?
Screenshot 2025-03-24 at 08 40 11

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
```diff
@@ -0,0 +1,60 @@
...
+      case Right(json) => Valid(json)
+    }
+
+  case class CompiledJsonExpression(originalJsonString: String, json: Json) extends CompiledExpression {
+
+    override def language: Language = languageId
+
+    override def original: String = originalJsonString
+
+    override def evaluate[T](ctx: Context, globals: Map[String, Any]): T = json.asInstanceOf[T]
```
Cichan-mcc, line 52: This looks very wrong but i didn't understand what should we evaluate from received json data...
Any tips how JSON evaluation should look like? Should we compare the `expectedType` with the actual JSON data type?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
arkadius: I think that we can have a `{}` as a generic default for jsons. In the ad-hoc it will be enough for now. In the future we have plans to create even smarter defaults, but IMO it is not a scope of this PR.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  override def parseWithoutContextValidation(
+      jsonString: String,
+      expectedType: TypingResult
+  ): ValidatedNel[ExpressionParseError, CompiledExpression] = {
+    parseJson(jsonString).map(json => CompiledJsonExpression(jsonString, json))
+  }
+
+  private def parseJson(jsonString: String): Validated[NonEmptyList[JsonParsingError], Json] =
```
arkadius, line 40: I thought that we would return Nussknacker's `Any`, I mean that we would do `ToJsonEncoder.encode` on parsed json. How does it work now? We expect somewhere that a value will be of type `Json`?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
arkadius: Json parser has a statically known value (it doesn't change if some variables change), so it is ok to have such a static logic.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -58,6 +58,17 @@ object JsonSchemaBasedParameter {
             paramName,
             isRequired = None
           ) // ObjectSchema doesn't use property required
+        case _: EmptySchema =>
```
arkadius: I assume that this logic is used during dry-run and it is why you have to to add this case? It is a hacky assumption. I see that this logic is used also for logic determining parameters on sink. The user will be surprised if we change the presented parameters in sink when they use the `{}` schema

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
Cichan-mcc: I added empty object `{}` default value for Json editors in default value determiners. Now default value applies properly
Screenshot 2025-03-26 at 09 43 50

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
Cichan-mcc: Ok, I left it this way. Thank you for clarification!

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
Cichan-mcc: After offline conversation we will leave it this way as it works properly. I only added a comment regarding adding typing instead of `Unknown` in the future.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: Hmmm, yeah i didn't notice that is would also change the sink editor.
Schemaless Jsons are actually implemented in Nu as Json empty schema. Not to interfer with sink editor I added `EmptySchemaJsonSupport` only for the reason of determinig parameters for dry run tests. This way sink's editor is the way it was before and JsonEditor is only used for dry run:
Screenshot 2025-03-26 at 09 50 44
Screenshot 2025-03-26 at 09 50 58

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
```diff
@@ -38,6 +38,7 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
   private def defaultStringExpression(editor: Option[ParameterEditor]): Expression =
     EditorBasedLanguageDeterminer.determineLanguageOf(editor) match {
       case Language.Spel => Expression.spel("''")
+      case Language.Json => Expression.spel("{}")
```
arkadius: hmm, why it is in `defaultStringExpression` and why do you use spel language instead of json language? :)

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
```diff
@@ -248,8 +248,11 @@ class UniversalKafkaSourceFactory(
         NonEmptyList.one(CustomNodeError(nodeId.id, "Cannot generate test parameters: no runtime schema found", None))
       )
       .andThen { runtimeSchema =>
-        val universalSchemaSupport: UniversalSchemaSupport =
-          schemaSupportDispatcher.forSchemaType(runtimeSchema.schema.schemaType())
+        val universalSchemaSupport = runtimeSchema.schema match {
+          // For ad hoc tests we want to present the user with json editor when schemaless topic with Json content type was selected
+          case ContentTypesSchemas.schemaForJson => EmptySchemaJsonSupport
```
arkadius: Let's look a momement at this code from perspective of what user see (DDD). User see that they choose topic without schema. They have to pick content type for this topic. So from this perspective I expect to see in the code:
```
case TopicWithoutSchema(contentType = Json) => NoSchemaJsonSupport or SchemalessJsonSupport
```
First part of this pattern was hacked - we have to fix eventually to avoid some bugs around this subject (we already have some because of this hack). But let's avoid going deeper into this hack and fix at least second part of pattern matching.

LSS
Please rename `EmptySchemaJsonSupport` to something more closer to domain: `NoSchemaJsonSupport` or `SchemalessJsonSupport`

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -148,17 +175,33 @@ object JsonSchemaSupport extends ParsedSchemaSupport[OpenAPIJsonSchema] {
...
     }
   }
 
-  override def formValueEncoder(schema: ParsedSchema, mode: ValidationMode): Any => AnyRef = {
-    val encoder   = new ToJsonSchemaBasedEncoder(mode)
-    val rawSchema = schema.cast().rawSchema()
-    (value: Any) => encoder.encodeOrError(value, rawSchema)
+}
+
+object EmptySchemaJsonSupport extends JsonSchemaSupport {
```
arkadius: Can you use composition instead of inheritance? I mean le't extract common methods to sth like `JsonSupportUtils` and use them here. Some method arguments doesn't fit this support e.g. schema (for schemaless support there is no schema), raw mode (for schemaless support this raw mode switch has no sense), so eventually we will need to refactor this. Composition will help in this.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -65,18 +75,45 @@ object JsonSchemaBasedParameter {
...
+        schema: Schema,
+        paramName: Option[ParameterName],
+        defaultValue: Option[Expression],
+        isRequired: Option[Boolean]
+    ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] = {
+      schema match {
+        case _: EmptySchema =>
+          // In a case of an empty JSON schema we want to present the user with Json editor and not standard SpEL one
+          Valid(
+            createJsonSinkSingleValueParameter(
```
arkadius: Let's copy this method instead of hacking the existing one. The new method won't have schema, defaultValue, isRequired and editor. Let's method name include "json editor" to explicitly show what's going on. If you already extracted a new "support" for schemaless case, I assume that  we don't even need to go through `withEmptySchemaSupport` -> `toSchemaBasedParameterWithEmptySchemaSupport` and do pattern matching on empty schema.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
```diff
@@ -0,0 +1,50 @@
...
+                       |    "someObject": {
+                       |      "someString": "some string value",
+                       |      "someBoolean": true,
+                       |      "someNumber": 21.37
+                       |    }
+                       |  }
+                       |]""".stripMargin
+
+    val resultType = parser.parse(validJson, ValidationContext.empty, Unknown).validValue.typingInfo.typingResult
+    resultType shouldBe Unknown
```
arkadius, line 34: Let's add a TODO that we should type this and Unknown type is a temporary type also here. As a new developer you would be confused if you add typing and figure out that someone expected Unknown type

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
Cichan-mcc: Oh, snap! Copy/paste of the line above got me... I changed it to json with `{}` for consistency with other determiners

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
Cichan-mcc: I refactored the code a bit. I renamed `EmptySchemaJsonSupport` to `NoSchemaJsonSupport` and removed inheritance from it. Now it just wraps logic around creating appropriate parameter with `JsonSchemaBasedParameter`

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
Cichan-mcc: I removed inheritance from `NoSchemaJsonSupport` and refactored the code not to use parameters and signatures that are irrelevant in the case of schemaless json topics. No composition with JsonSchemaSupport was actually necesseary this way

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: I created a new method `getJsonEditorInputParameter` which gets us what we need here. It doesn't need these parameters. And also you're correct that we don't need to do pattern matching here after changes.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonParserTest.scala
Cichan-mcc: Done, I added the comment here as well

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
arkadius: Please confirm that you've tested in manually e2e, because with the previous approach I assume that it didn't work?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
arkadius: It is better to use `ParsedSchemaSupport` interface. The logic behind `ParsedSchemaSupport` was to dispatch all logic that is different depending on what schema is associated with topic. Now we by-pass this dispatching logic. Following the original idea, it will be easy to show differences between topic with json schema and topic with json schema - it will be in one place.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -39,6 +39,11 @@ object JsonSchemaBasedParameter {
       isRequired = None
     )
 
+  def emptySchemaBasedParameter(defaultParamName: FieldName, validationMode: ValidationMode)(
+      implicit nodeId: NodeId
+  ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] =
+    ParameterRetriever(EmptySchema.INSTANCE, defaultParamName, validationMode).getJsonEditorInputParameter
```
arkadius: `ParameterRetriever` was designed to provide automatically generated parameters based on schema. Currently, you pass fake empty schema for the situation when there is no schema to provide only one parameter that is quite obvious :-). It looks crazy :-D

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
Cichan-mcc: I tested that before and after the changes. It worked in both cases although the `spel` one was obviously wrong. It worked because `TypeRelatedParameterValueDeterminer` is lower in hierarchy of `DefaultValueDeterminerChain` than other correct determinetrs and it wasn't used because of that.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
Cichan-mcc: Ok, `NoSchemaJsonSupport` implements `ParsedSchemaSupport` now. It uses proxy pattern and only `extractParameter` method is changed to return the Json Editor.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
Cichan-mcc: As i mentioned in the comment above `NoSchemaJsonSupport` implements `ParsedSchemaSupport` now and uses composition.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: Hmmm, I didn't really look at `ParameterRetriever` from the domain perspective... It looked like it had the logic that i needed and `JsonSchemaBasedParameter` seemed like just a facade. So I tried to have the logic encapsulated in `ParameterRetriever` :D 

But I see that now, I moved the logic out of `ParameterRetriever` into `JsonSchemaBasedParameter`

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -39,6 +39,22 @@ object JsonSchemaBasedParameter {
       isRequired = None
     )
 
+  def emptySchemaBasedParameter(
```
arkadius: It is not schema based parameter (we use it only for case when there is no schema) so let's remove it from `JsonSchemaBasedParameter``

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -39,6 +39,22 @@ object JsonSchemaBasedParameter {
...
+      defaultParamName: FieldName,
+      validationMode: ValidationMode
+  ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] = {
+    val emptyJsonSchema = EmptySchema.INSTANCE
+    val typing          = SwaggerBasedJsonSchemaTypeDefinitionExtractor.swaggerType(emptyJsonSchema, None).typingResult
```
arkadius: We don't need to compute type for empty schema, we know that it will be `Unknown`

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -39,6 +39,22 @@ object JsonSchemaBasedParameter {
       isRequired = None
     )
 
+  def emptySchemaBasedParameter(
+      defaultParamName: FieldName,
```
arkadius: we don't need to make these things configurable - there are the same for the only infocation that we make, YAGNI, KISS

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
```diff
@@ -39,6 +39,22 @@ object JsonSchemaBasedParameter {
...
+      validationMode: ValidationMode
+  ): ValidatedNel[ProcessCompilationError, SchemaBasedParameter] = {
+    val emptyJsonSchema = EmptySchema.INSTANCE
+    val typing          = SwaggerBasedJsonSchemaTypeDefinitionExtractor.swaggerType(emptyJsonSchema, None).typingResult
+    val parameter = Parameter(defaultParamName, typing)
+      .copy(isLazyParameter = true, editors = List(JsonParameterEditor))
+    Valid(
```
arkadius: If it is always valid, why to add the noice in returned type? KISS

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -162,3 +162,31 @@ object JsonSchemaSupport extends ParsedSchemaSupport[OpenAPIJsonSchema] {
...
+
+  override def typeDefinition(schema: ParsedSchema): TypingResult = jsonSupport.typeDefinition(schema)
+
+  override def formValueEncoder(schema: ParsedSchema, mode: ValidationMode): Any => AnyRef =
+    jsonSupport.formValueEncoder(schema, mode)
+
+  override def recordFormatterSupport(schemaRegistryClient: SchemaRegistryClient): RecordFormatterSupport =
+    jsonSupport.recordFormatterSupport(schemaRegistryClient)
+
+  override def extractParameter(
```
arkadius: Won't it be used also for sink, for topics without schema? The method (especially `rawMode` which is visible on the interface only in sink context) says that it will be used also there.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: Technically empty json schema is used (`{}`) so that's why i called it like this and placed it in `JsonSchemaBasedParameter` but from the domain perspective there shouldn't be one. I moved the logic outside of `JsonSchemaBasedParameter` directly into `NoSchemaJsonSupport`

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: Oh, right. I removed that line and changed it to Unknown.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: I removed that method from `JsonSchemaBasedParameter` so parameters are no longer a thing here.

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
Cichan-mcc: True, I tried to apply to the type of `NoSchemaJsonSupport.extractParameter` which implements a method from a trait. Since I moved the Json editor implementation I also changed the return type there to `Valid`

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
Cichan-mcc: It would if `NoSchemaJsonSupport` was used for sink creation but it's not. 
`UniversalSchemaSupportDispatcher` is used for accessing `UniversalSchemaSupport` which `NoSchemaJsonSupport` implements now. `UniversalSchemaSupportDispatcher.forParsedSchema` is only used in the context of ad hoc/dry run tests that's why `NoSchemaJsonSupport` will be used only there.

I don't really like the unused/misleading parameters that are present in the `extractParameter` but we have to do it by applying to the `UniversalSchemaSupport` trait now

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonSchemaBasedParameter.scala
arkadius: Yep, this empty schema is a hack that should be cleaned up eventually. If we modeled the domain correctly, there would be `None` somewhere. Let's do everything to make this clean-up as easy as possible :+1:

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
arkadius: Take a look at this comment: https://github.com/TouK/nussknacker/pull/7690#discussion_r2014657282 . Let's follow this original idea and evolve it instead of hacking it.
LSS extract dedicated method for two cases: for ad-hoc parameters and for sink parameters

Cichan-mcc: I split the method into `extractParameterForSink` and `extractParameterForTests`. The base implementations delegate to the `protected extractParameter`. `NoSchemaJsonSupport` overrides `extractParameterForTests` with its Json editor implementation.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaSupport.scala
```diff
@@ -41,25 +48,39 @@ trait UniversalSchemaSupport {
...
+  protected def extractParameter(
       schema: ParsedSchema,
       rawMode: Boolean,
       validationMode: ValidationMode,
       rawParameter: Parameter,
       restrictedParamNames: Set[ParameterName]
   )(implicit nodeId: NodeId): ValidatedNel[ProcessCompilationError, SchemaBasedParameter]
 
-  final def extractParameters(
+  def extractParameterForSink(
```
arkadius, line 50: 1. Please move `extractParameters` before these two methods. It is a good approach to keep method used by other methods below these methods. Thanks to that you reduce jumping over the page.
2. Please make it private in `JsonSchemaSupport` and in `AvroSchemaSupport` and in `NoSchemaJsonSupport` override `extractParameterForSink = jsonSupport.extractParameterForSink()` - it will be easier to understand the relation

arkadius (approved):
Good job

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/UniversalSchemaSupport.scala
Cichan-mcc: I made `extractParameter` `private` in both `JsonSchemaSupport` and `AvroSchemaSupport`. And delegated `extractParameterForSink` to  `JsonSchemaSupport` in `NoSchemaJsonSupport`.

PR #7700, opened 2025-03-20
Default value for event generator

ForrestFairy:
## Describe your changes
With default values, `event generator` won't turn red immediately after placing it in scenario.
image

Designer for cypress tests used some dummy helpers which were different from the ones that users can see. Now we will use the same helpers in tests and in normal usage.

## Checklist before merge


designer/client/cypress/e2e/fragment.cy.ts
```diff
@@ -238,16 +238,16 @@ describe("Fragment", () => {
 
         cy.get("[model-id=sendSms]").should("be.visible").trigger("dblclick");
         cy.intercept("POST", "/api/nodes/*/validation", (request) => {
-            if (request.body.nodeData.ref?.parameters[0]?.expression.expression == "#fragmentResult.") {
```
Dzuming: i have a feeling, that this test was about verification, whether suggestion works after the typing `.` value

Dzuming (approved):
e2e tests LGTM

PR #7711, opened 2025-03-21
[NU-2082] Schema derived from JSON example for Kafka topics without schema

mateuszkp96:
## Describe your changes
Ability to derive schema from data sample for schemaless Kafka topics for Kafka Sources

image

image

image

## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
```diff
@@ -54,8 +55,10 @@ object JsonParser extends ExpressionParser {
 
 }
 
-case object JsonExpressionTypingInfo extends ExpressionTypingInfo {
+case class JsonExpressionTypingInfo(json: Json) extends ExpressionTypingInfo {
+
+  override val typingResult: TypingResult = {
+    Typed.fromInstanceWithNullsAsUnknowns(FromJsonSimpleDecoder.jsonToAny(json))
```
arkadius: The another difference from standard `Typed.fromInstance` invocation that we should  apply here, is that we don't want to have inferred values in types. It is because the user provides not a value that they wish to type but an example, so the type should be wider. You can achieve that using `.withoutValue`. I'm thinking that we don't even need `fromInstanceWithNullsAsUnknowns` method. The `.withoutValue` should generalize `TypedNull` to `Unknown` (Any)

arkadius: Ahh I see that you did it, but in the place where it should be done. Only part with `fromInstanceWithNullsAsUnknowns` is relevant

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
```diff
@@ -281,6 +279,74 @@ class UniversalKafkaSourceFactory(
...
+  private def runtimeDataForJsonSchema = {
+    RuntimeSchemaData[ParsedSchema](
+      new NkSerializableParsedSchema[ParsedSchema](ContentTypesSchemas.schemaForJson),
+      Some(SchemaId.fromString(ContentTypes.JSON.toString))
+    )
+  }
+
+  private def typingResultWithoutValues(typingResult: TypingResult) = {
+    typingResult.withoutValue match {
+      case obj: TypedObjectTypingResult if obj.fields.isEmpty => Unknown
```
arkadius: Let's add a comment that it is a kind of heuristic - we assume that user passes `{}` when the are not sure what will be in fields.
Please extract this method to some new class e.g. `TypingResultFromJsonTypeDeterminer`

mateuszkp96: Done. I added TypingResultFromJsonTypeDeterminer class in the `component-utils` module

arkadius: I'd rather thinking about some closer place to this class (e.g. in the same module, next to this class). I want to keep this logic easy to reuse but not too make it too much generic yet because it can be not easy to figure out that it is a type base on example json. BTW proposed by name doesn't include `Sample` which is very important in this context. Maybe let's call it `TypingResultFromJsonSampleTypeDeterminer`.

utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/typing/TypingResultFromJsonTypeDeterminer.scala
```diff
@@ -0,0 +1,16 @@
+package pl.touk.nussknacker.engine.util.typing
+
+import pl.touk.nussknacker.engine.api.typed.typing.{TypedObjectTypingResult, TypingResult, Unknown}
+
+object TypingResultFromJsonTypeDeterminer {
+
+  def apply(typingResult: TypingResult): TypingResult = {
+    typingResult.withoutValue match {
+      // This is some kind of heuristic; when the user passes '{}', we are not sure what will be in the fields
+      // empty record does not provide any value for the user; he cannot access any fields. Unknown, on the other hand, allows a dynamic navigation
```
arkadius: replace `he` with `he/she` or `they`

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/encode/AvroSchemaOutputValidator.scala
```diff
@@ -69,6 +69,8 @@ class AvroSchemaOutputValidator(validationMode: ValidationMode) extends LazyLogg
         invalid(typingResult, schema, path)
       case (TypedNull, _) if schema.isNullable =>
         valid
+      case (Unknown, Type.NULL) =>
```
arkadius: I guess that we should also check `if schema.isNullable` as we do above

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -158,7 +158,7 @@ object typing {
   //   We can avoid this case by changing this folding logic - see the comment there
   case object TypedNull extends TypingResult {
 
-    override val withoutValue: TypingResult = this
+    override val withoutValue: TypingResult = Unknown
```
arkadius, line 162: Maybe let's add a comment, that `Unknown` is our `Any` and null value can be a subclass of any class so we return it.

arkadius: Maybe also let's add a notice in migration guide that we changed `TypedNull.wihoutValue` behaviour. As we can see in the core code, someone could base on the current behaviour.

arkadius (approved):
Small nipticks added, you can merge after fixing them

PR #7732, opened 2025-03-23
[Nu-7702] Warn user when scenario version is outdated before save or deploy

### LINKED ISSUE #7702: Warn user when scenario version is outdated before save or deploy
## 💥 Proposal

Sometimes it may happen that user tries to

- save scenario version after someone saved theirs version in another machine or browser tab
- deploy of currently seen version while newer deployment was committed in the meantime

In both of these situations we should at least inform the user that current version is not being the latest one. Admittedly the scenario activities
are updated without need of page reload and are properly updated on their panel - but they might be unnoticed by the user.

DeamonDev:
## Describe your changes

I implemented mechanism to warn user when tries to save version of scenario which is not the recent one (for example because someone else modified it in the meantime). 

image

The analogous warning is added for deployment button

image

## Checklist before merge


Dzuming (approved):
LGTM

PR #7768, opened 2025-03-26
[NU-2004] Passing StreamExecutionEnvironment to node compilation

arkadius:
## Checklist before merge


build.sbt
```diff
@@ -824,6 +827,19 @@ lazy val flinkExecutor = (project in flink("executor"))
     flinkTestUtils % Test,
   )
 
+lazy val scenarioCompilerFlinkDeps = (project in flink("scenario-compiler-deps"))
```
arkadius, line 830: A new module was needed for one class that have to be used on both Flink DM side and Flink executor side

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/definition/TablesDefinitionDiscoveryTest.scala
```diff
@@ -1,39 +1,89 @@
...
     with LoneElement
     with ValidatedValuesDetailedMessage
     with TableDrivenPropertyChecks
-    with PatientScalaFutures {
+    with PatientScalaFutures
+    with ModelClassLoaderSimulationSuite {
+
+  override protected type FixtureParam = TablesDefinitionDiscovery
+
+  private val miniClusterWithServices =
```
arkadius, line 42: I rewrote tests using implicit minicluster usage to explicit minicluster usage in order to:
- expose that we use our minicluster wrapper with the same extensions as in normal flink distribution (csv format, json format, filesystem conector, planner, etc)
- better clean-up resources (streamexecutionenv is now closed after each test)

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/FlinkDeploymentManager.scala
```diff
@@ -369,6 +372,18 @@ class FlinkDeploymentManager(
...
   }
 
+  override def scenarioCompilationDependenciesResource: Resource[SyncIO, EngineScenarioCompilationDependencies] = {
+    miniClusterWithServicesOpt
+      .map(
+        _.createAttachedStreamExecutionEnvironment[SyncIO]
+          .map(new FlinkScenarioCompilationDependencies(_))
+      )
+      .getOrElse {
+        // TODO: always create share minicluster
```
arkadius: I'll fix it in the next PR

mslabek (approved):
Please check if we can reduce usages of minicluster in attached mode. And maybe add rename to `withContextClassLoader` . LGTM

arkadius:
> Please check if we can reduce usages of minicluster in attached mode. And maybe add rename to `withContextClassLoader` . LGTM

Ok, done. It works with detached mode

PR #7779, opened 2025-03-26
[NU-2076] Single button for tests with parameters (adhoc) and with generated data

mgoworko:
## Describe your changes

New button and new modal, that handles both tests with parameters and with generated data.
- old buttons not removed from code, still can be configured and used
- but in the default config example the old buttons are replaced with the new one

Screenshots
--------------------
1. When test not available for sources - button inactive

--------------------
2. When only tests with generated data are available

--------------------
3. When only tests with parameters (adhoc) are available

--------------------
4. When both types of tests are available

--------------------
5. Recently added adhoc test animation still works for the new button, for both types of tests
[recording](https://github.com/user-attachments/assets/c3d17d61-0034-4002-a050-f6abe8e92b18)

## Checklist before merge


Dzuming:
@mgoworko 

1. Is it possible to somehow inform a user why the live samples / Form option is disabled and what to do to enable it?
2. After clicking `learn more` the documentation page looks off-sync and not strictly related to the context where the user currently is. Are you planning to update the docs?

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+                value={{
+                    value: adhocTestingCurrentValue,
+                    setValue: setAdhocTestingCurrentValue,
+                    parameters,
+                    variableTypes,
+                    errors: adhocTestingErrors,
+                }}
+            >
+                <NodeTable>
+                    <FormLabel required>{t("addProcessForm.label.testMode", "Data used in sources")}</FormLabel>
```
Dzuming:
```suggestion
                    <FormLabel required>{t("addProcessForm.label.testMode", "Data used in scenario sources")}</FormLabel>
```

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    handleSetTouched: (touched: TouchedValue) => void;
+    touched: TouchedValue;
+    testingData: WindowType<WindowKind, TestingData>;
+    closeDialog: () => void;
+}
+
+const DryRunTestingIcon = loadable(() => import("../../../assets/img/icons/test-dry-run.svg"));
+const GenerateAndTestIcon = loadable(() => import("../../../assets/img/icons/test-using-generated-data.svg"));
+
+export function TestingForm({ value, touched, onChange, handleSetTouched, testingData, closeDialog }: TestingFormProps): JSX.Element {
```
Dzuming: IMO, this component makes way too much, we should separate it into a few components:

1. Scenario, which will have a logic to select a specific form, and after clicking into specific "sources" button will render a specific form
2. Testing Form, which will contain all ad-hoc related logic (buttons included here)
3. TestngFromSingleSource, rendered by the testing form
4. tetsingFromMultipleSourceWarning, rendered by the testing form
5. Live samples, which will contain testing from generated file logic (buttons included here)


Thanks to it, we will be able to hide logic related only to specific variants like useAdhocTesting* hooks, and we can remove redundant [enabled](https://github.com/TouK/nussknacker/pull/7779/files) flag from useAdhocTestingParametersValidation

Dzuming (changes requested):

designer/client/src/components/modals/Testing/TestingForm.tsx
mgoworko: Ad 1. Done, added info in tooltip, as much as is currently available - 2 cases 
-    `Scenario testing is not supported for scenario in current state` - for example for archived, when the test button is disabled because of backed config (but FE does not have information about detailed reason)
- `Scenario testing is not supported for currently configured sources` - when the scenario sources do not support testing (or there are currently no sources)

Ad 2. I will update docs after this PR is merged

Ad. this comment: suggestion is applied

designer/client/src/components/modals/Testing/TestingForm.tsx
mgoworko: Separated into 
- TestingDialog - base for dialog that is opened 
- TestingForm - base content of the dialog, handles test type buttons
- TestVariantForm - chooses correct variant form
- and 3 variants:
  - TestWithParametersSingleSourceForm
  - TestWithParametersMultipleSourcesForm
  - TestWithGeneratedDataForm

Flag removed

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+const DryRunTestingIcon = loadable(() => import("../../../assets/img/icons/test-dry-run.svg"));
+const GenerateAndTestIcon = loadable(() => import("../../../assets/img/icons/test-using-generated-data.svg"));
+
+export function TestingForm({ testingData, closeDialog }: TestingFormProps): JSX.Element {
+    const { t } = useTranslation();
+
+    const testCapabilities = useSelector(getTestCapabilities);
+
+    const availabilityMap: Record<TestType, boolean> = {
```
Dzuming: can you memoize value with useMemo?

designer/client/src/components/modals/Testing/useValidatedLocalStorage.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+export function useValidatedLocalStorage<T>(key: string, initialValue: T, validate: (value: any) => boolean) {
```
Dzuming: Why do we need this custom hook? Looks like we can achieve the same with https://rooks.vercel.app/docs/useLocalstorageState (we already use it in a few places)

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+            handleSetTouched({ ...touched, [field]: true });
+        },
+        [handleSetTouched, onChange, touched, testType],
+    );
+    const onBlurChange = useCallback(
+        (field: keyof TouchedValue, next: boolean) => handleSetTouched({ ...touched, [field]: next }),
+        [handleSetTouched, touched],
+    );
+
+    const testWithParametersIsAvailable = testCapabilities.canTestWithForm;
```
Dzuming: Can we use these testWithParametersIsAvailable and testWithGeneratedDataIsAvailable values in availabilityMap?

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <div className={cx(css({ paddingTop: 10, paddingBottom: 20 }))}>
+            <NodeTable>
+                <FormLabel required>{t("addProcessForm.label.testMode", "Data used in scenario sources")}</FormLabel>
+                <span className={nodeValue}>
+                    <FormGroup
+                        row
+                        sx={(theme) => ({ flexWrap: "flex", alignItems: "center", gap: theme.spacing(1.5) })}
+                        onChange={(event) => {
+                            const target = event.target as HTMLInputElement;
+                            onFieldChange("testType", target.checked ? target.value : target.value);
```
Dzuming: do we need this ternary?

designer/client/src/components/modals/Testing/TestingForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+                                href="https://nussknacker.io/documentation/docs/scenarios_authoring/TestingAndDebugging/"
+                                target="_blank"
+                                rel="noopener noreferrer"
+                            >
+                                to learn more.
+                            </Link>
+                        </Trans>
+                    </Typography>
+                </span>
+                <TestVariantForm testType={testType.testType} testingData={testingData} closeDialog={closeDialog}></TestVariantForm>
```
Dzuming: this testType.testType sounds strange, maybe we can find a better naming?

designer/client/src/components/modals/Testing/TestVariantForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+        ) : (
+            <></>
+        );
+
+    const testWithGeneratedDataElement =
+        testType === TestType.withGeneratedData ? <TestWithGeneratedDataForm closeDialog={closeDialog}></TestWithGeneratedDataForm> : <></>;
+
+    return (
+        <div className={cx(css({ paddingTop: 10, paddingBottom: 20 }))}>
+            {testWithParametersElementWhenSingleSource}
```
Dzuming: Is it possible to render all these components at once? IMO, now. Maybe to increase readability and add some error handling when none of the states available, we should change it to something like?

```js
export function TestVariantForm({ testType, testingData, closeDialog }: TestVariantFormProps): JSX.Element {
const testParameters = useSelector(getTestParameters);
const sourcesFound = testParameters.length;
const thereAreMultipleSources = sourcesFound > 1;

    const testWithParametersElementWhenSingleSource = useMemo(() => TestType.withParameters && !thereAreMultipleSources, [thereAreMultipleSources]);
    const testWithParametersElementWhenMultipleSources = useMemo(() => testType === TestType.withParameters && thereAreMultipleSources, [testType, thereAreMultipleSources]);
    const testWithGeneratedDataElement = useMemo(() => testType === TestType.withGeneratedData, [testType]);

    if (testWithParametersElementWhenSingleSource) {
        return (
            <TestWithParametersSingleSourceForm testingData={testingData} closeDialog={closeDialog}></TestWithParametersSingleSourceForm>
        );
    }

    if (testWithParametersElementWhenMultipleSources) {
        return <TestWithParametersMultipleSourcesForm numberOfSources={sourcesFound}></TestWithParametersMultipleSourcesForm>;
    }


    if (testWithGeneratedDataElement) {
        return <TestWithGeneratedDataForm closeDialog={closeDialog}></TestWithGeneratedDataForm>;
    }

    // Let's handle this case somehow, maybe we need to display ErrorBoundary component via showBoundary metho here?
    return null;
}
```

and move 

```js
<div className={cx(css({ paddingTop: 10, paddingBottom: 20 }))}>
```

to the parent component?

designer/client/src/components/modals/Testing/TestWithParametersMultipleSourcesForm.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function TestWithParametersMultipleSourcesForm({ numberOfSources }: TestWithParametersMultipleSourcesFormProps): JSX.Element {
+    return (
+        <Box
+            sx={{ display: "flex", flexWrap: "wrap", justifyContent: "center", gap: 2, width: "100%" }}
+            style={{ marginBottom: "12px", marginTop: "12px" }}
+        >
+            <Typography component="span" variant={"subtitle1"} noWrap={false} align={"center"}>
+                {`Test with form is supported only for scenario with single source. Your scenario has ${numberOfSources} sources.`}
```
Dzuming: Let's move it to i18n, and let's add https://www.i18next.com/translation-function/plurals

designer/client/src/components/toolbars/test/buttons/ScenarioTestButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+            meta: {
+                viewParams: { Icon: TestingIcon, docs, markdownContent },
+            },
+        });
+    }, [docs, markdownContent, open, t]);
+
+    const isLoading = useSelector(getTestResultsLoading);
+
+    const tooltip: string = disabled
+        ? "Scenario testing is not supported for scenario in current state"
```
Dzuming: let's add i18n here

designer/client/src/components/toolbars/test/buttons/ScenarioTestButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+        : !atLeastOneTypeOfTestIsAvailable
+        ? "Scenario testing is not supported for currently configured sources"
```
Dzuming: let's add i18n here

Dzuming (changes requested):

designer/client/src/components/modals/Testing/useValidatedLocalStorage.ts
mgoworko: We need it, because it additionally validates the content stored in the local store.
- each scenario has different format of test parameters
- we cannot use parameters from another scenario
- so we can use parameters from local storage only when their format is correct and matches the scenario parameters definition

designer/client/src/components/modals/Testing/TestingForm.tsx
mgoworko: removed

designer/client/src/components/modals/Testing/TestingForm.tsx
mgoworko: changed to formValue.testType

designer/client/src/components/modals/Testing/TestVariantForm.tsx
mgoworko: I refactored it to use switch to handle all cases. I think that missing implementation for test type would be a bug in code, and not a runtime error to show to the user.

Dzuming: Yes, that's what I mean here, we need to handle unhandled errors here, and ErrorBoundary is the standard way to do it in our case

mgoworko: Done, error using ErrorBoundary

designer/client/src/components/modals/Testing/useValidatedLocalStorage.ts
mgoworko: As discussed, refactored to use redux state. It was used for test type.

designer/client/src/components/modals/Testing/TestWithParametersMultipleSourcesForm.tsx
Dzuming: defaultValue_plural is missing here

Dzuming (changes requested):

designer/client/src/components/modals/Testing/TestWithParametersMultipleSourcesForm.tsx
Dzuming: There is no option to have singular form here

designer/client/src/components/modals/Testing/TestingForm.tsx
mgoworko: Ad1: added tooltips for not available test types, with basic information, that the scenario sources do not support the test type.

Dzuming (approved):
LGTM

PR #7788, opened 2025-03-27
Safe list navigation

ZbyszekMM:
## Describe your changes

The hidden gem of safe list navigation. Only docs.


philemone (approved):
LGTM :+1:

PR #7801, opened 2025-03-28
[NU-2065] Wrong event time of events coming from time windows

paw787878:
## Checklist before merge


engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/ExtendedWindowOperator.scala
```diff
@@ -113,6 +138,14 @@ private class ValueEmittingWindowFunction(
   ): Unit = {
     elements.forEach { element =>
       val ctx = Option(elementHolder.get()).getOrElse(api.Context(contextIdGenerator.nextContextId()))
+
+      out match {
+        // it should always be an instance of this class
+        case timedOut: TimestampedCollector[_] =>
+          Option(overriddenResultEventTimeHolder.get()).foreach(timestamp => timedOut.setAbsoluteTimestamp(timestamp))
+        case _ =>
```
raphaelsolarski: maybe we should throw something here like `IllegalStateException("Should not happen")`?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/ExtendedWindowOperator.scala
```diff
@@ -17,53 +18,74 @@ import pl.touk.nussknacker.engine.api.ValueWithContext
...
-}
+import pl.touk.nussknacker.engine.flink.util.transformer.aggregate.ExtendedWindowOperator.{Input, elementHolder, stateDescriptorName, overriddenResultEventTimeHolder}
 import pl.touk.nussknacker.engine.flink.util.transformer.aggregate.transformers.AggregatorTypeInformations
-import pl.touk.nussknacker.engine.flink.util.transformer.aggregate.triggers.FireOnEachEvent
+import pl.touk.nussknacker.engine.flink.util.transformer.aggregate.triggers.{ClosingEndEventTrigger, FireOnEachEvent}
 
 import java.lang
 
-object OnEventTriggerWindowOperator {
+object ExtendedWindowOperator {
```
raphaelsolarski, line 28: I wonder if we could find some less generic name for the class, but I have no idea 😄

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/TransformersTest.scala
```diff
@@ -743,6 +760,13 @@ class TransformersTest extends AnyFunSuite with FlinkSpec with Matchers with Ins
 
       val nodeResults = collectingListener.endVariablesForKey(id)
       nodeResults.flatMap(_.variableTyped[TestRecordHours]("input")) shouldBe Nil
+
+      val timeMultiplier = 3600 * 1000
+
+      nodeResults.map(
+        _.variableTyped[Long](CustomTimestampExtractingTransformation.timestampVariableName).get
+          // session timeout is 2
+      ) shouldBe List((3 + 2) * timeMultiplier - 1, 6 * timeMultiplier, (6 + 2) * timeMultiplier - 1)
```
raphaelsolarski: maybe we could use some helper function and comments?
```scala
nodeResults.map(
  _.variableTyped[Long](CustomTimestampExtractingTransformation.timestampVariableName).get
) shouldBe List(
  hoursToMillis(5) - 1, // 3h last event time + (2h - 1ms) timeout
  hoursToMillis(6),     // 6h event time from event witch evaluated stop condition to true
  hoursToMillis(8) - 1  // 6h last event time + (2h - 1ms) timeout
)
```
---
```scala
case class TestRecordHours(id: String, timeHours: Int, eId: Int, str: String) extends TestRecord {
  override def timestamp: Long = hoursToMillis(timeHours)
}

object TestRecordHours {
  def hoursToMillis(hours: Int): Long = hours * 3600L * 1000
}
```

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/ExtendedWindowOperator.scala
```diff
@@ -17,53 +18,74 @@ import pl.touk.nussknacker.engine.api.ValueWithContext
...
   type Input[A] = ValueWithContext[StringKeyedValue[A]]
 
   // We use ThreadLocal to pass context from WindowOperator.processElement to ProcessWindowFunction
   // without modifying too much Flink code. This assumes that window is triggered only on event
   val elementHolder = new ThreadLocal[api.Context]
+  // This ThreadLocal is also used to pass information between two distant places in flink code without
+  // modifying it
+  val overriddenResultEventTimeHolder = new ThreadLocal[Long]
```
raphaelsolarski: mby we can try to gather the timestamp using process function before the window and pass it inside event type? Using the thread local to pass timestamp gathered in trigger to internal window function is imho quite tricky.

raphaelsolarski (changes requested):

raphaelsolarski (approved):

raphaelsolarski (changes requested):
Let's wait for a while before merging - I'm not sure whether the change will work with event matching multiple windows at once (e.g. sliding windows) due to calling trigger for multiple windows in Flink `WindowOperator.processElement` implementation.

raphaelsolarski (approved):
epic. Good job

PR #7824, opened 2025-03-31
[NU-2116] Add Json type for schemaless topics with json content type

Cichan-mcc:
## Describe your changes

Currently when we choose to use Kafka topic without schema and Json content type the type of `#input` is `Unknown`:
Screenshot 2025-04-02 at 15 52 49

With these changes new `Json` type is introduced. For now it is very similar to the `Unkown` type as we want to have the possibility to dynamically access attributes inside. The `Json` type is used when user provides ambiguous data sample (`{}`, `[]`, `null`) which previously meant that `Unknown` type would be used.

Screenshot 2025-04-02 at 15 01 18

In expression fields we have the same suggestions as for the `Unknown` type and we can cast it to an actual type.
Screenshot 2025-04-02 at 15 01 39
Screenshot 2025-04-02 at 15 01 48

In ad hoc/dry run test the parameter was changed to `Input` from `Value` as that's what we pass there with Json data.
Screenshot 2025-04-02 at 15 01 04

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -176,6 +176,14 @@ object typing {
     override val display = "Unknown"
   }
 
+  case object TypedJson extends TypingResult {
```
arkadius: I think it is not a way to go to add everwhere, where we already have `Unknown` the same handling logic as for `TypedJson`. This solution have two disadvantages:
- Our typing system is a part of the component-api, developer of a component will need to think what to do with this new type
- They can handle these cases wrong and introduce costly bugs

IMO we should rather consider adding some optional information about how `Unknown` will be presented and try to keep it as transparent as possible. We already have similar approach  done with arrays. They are presented as lists. See `TypedClass.display`

Tricks that would help us to make this as transparent as possible:
- We can have `object Unkown extends Unkown(defaultPresentationLogic)`   
- Instead of having `case object`, we can manually add `equals`, `hashCode` and extractor (`Unknown.unapply`)

BTW thanks this approach we could change `defaultPresentationLogic` to `Any`. Thanks to that, there is a chance that this change will be quite compatible.

Cichan-mcc: I modified the structure a bit. I removed additional `TypedJson` type and made Json a display strategy of `Unknown` type. I also introduced `object Unknown extends Unknown(UnknownDisplayStrategy)` which mimics how `Unknown` was treated previously which helps to make the change more transparent. I just had to adjust a few pattern matches as now `Unknown` can represent two display strategies.

I tried to manually implement own extractor and other methods but i don't know whether we could make it more transparent than what we have now.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
```diff
@@ -61,6 +61,14 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
       .map(_.getOrElse(Map.empty))
   } yield Typed.record(fields, valueClass, additional)
 
+  private def unknown(obj: HCursor): Decoder.Result[TypingResult] = {
+    obj
+      .downField("display")
+      .as[String]
+      .map(value => if (value == JsonDisplayStrategy.display) Right(TypedJson) else Right(Unknown))
```
arkadius: Maybe we could have just a decoder for display strategy?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -37,8 +37,8 @@ object AssignabilityDeterminer {
 
   private def isAssignable(from: TypingResult, to: TypingResult, conversionChecker: ConversionChecker) = {
     (from, to) match {
-      case (_, Unknown)       => ().validNel
-      case (Unknown, _)       => ().validNel
+      case (_, Unknown(_))    => ().validNel
```
arkadius, line 40: Can we create an extractor similar to these created for case object? Thanks to that we won't need to add `(_)` in pattern matching

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -167,15 +167,28 @@ object typing {
...
 
+  // Unknown is representation of TypedUnion of all possible types
+  case class Unknown(displayStrategy: DisplayStrategy) extends TypingResult {
+    override def withoutValue: TypingResult = this
+    override val valueOpt: None.type        = None
+    override val display: String            = displayStrategy.display
+  }
+
+  object Unknown   extends Unknown(UnknownDisplayStrategy)
+  object TypedJson extends Unknown(JsonDisplayStrategy)
```
arkadius: Almost every type can be created using `Typed.xyz` syntax. I think that we can follow the same approach with json - we can extract `val Typed.json = Unknown(JsonDisplayStrategy)` method. Evenually, we could do the same with `Unknown` object but for sake of backward compatibility it can be done later.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/compile/NodeDataValidatorSpec.scala
```diff
@@ -1190,6 +1190,21 @@ class NodeDataValidatorSpec extends AnyFunSuite with Matchers with Inside with T
     }
   }
 
+  test("should properly validate Json type parameters") {
+    val ctx = ValidationContext(Map("input" -> TypedJson))
+    val expressionsWithExpectedTypes = List(
+      ("#input", TypedJson),
+      ("#input[0]['products'][1]['id']", TypedJson),
+      ("#input[0]['products'][1]['id'].toInteger()", Typed.typedClass[Int])
```
arkadius: We need to handle also:
- `#input.toMap`
- `#input.toList`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
Cichan-mcc: I introduced `val json: Unknown = new Unknown(JsonDisplayStrategy)` and updated places which used `TypedJson`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
Cichan-mcc: I introduced a Decoder for `Unknown` class since that's what we actually needed based on `display` attribute. I had to use `fold` instead of `orElse` for Scala 2.12 compatibility

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
```diff
@@ -42,6 +42,16 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
     case e                     => Left(s"$e is not SingleTypingResult")
   }
 
+  private implicit val displayUnknownDecoder: Decoder[Unknown] = Decoder.instance { cursor =>
+    cursor
+      .as[String]
+      .map {
+        case JsonDisplayStrategy.display => Typed.json
```
arkadius: We can wrap both display strategies in enumatum and do here valueOf getOrElse throw exception. It will be harder to introduce some bug that we add/change strategies and something stopped working

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -167,15 +167,27 @@ object typing {
...
 
-  // Unknown is representation of TypedUnion of all possible types
-  case object Unknown extends TypingResult {
-    override def withoutValue: Unknown.type = Unknown
+  sealed trait DisplayStrategy {
+    val display: String
+  }
 
-    override val valueOpt: None.type = None
+  case object UnknownDisplayStrategy extends DisplayStrategy {
```
arkadius: Maybe let's call it `DefaultDisplayStrategy`. Let it be easy to change the implementation of this strategy.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
arkadius: Let's leave as it is + add migration guide

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
Cichan-mcc: I changed the name to `DefaultDisplayStrategy`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
Cichan-mcc: I wrapped the `DisplayStrategy` in Enum and made sure that exception is thrown if `display` does not correspond to defined display strategies. 
I couldn't use valueOf since the Enums names' are different and I needed to make sure that `Unknown` is returned in case there is no `display` attribute in Json so the change is compatible.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
Cichan-mcc: I added a mention in migration guide on how to match against `Unknown` now

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/compile/NodeDataValidatorSpec.scala
Cichan-mcc: I'm working on this to add correct typing for these cases. 
Right now `#input.toMap` and `#input.toList` return `Map[Json, Json]` and `List[Json]` accordingly. 
I need to change the first one into `Map[String, Json]` and I also need to make sure that calling `get()` on both of them return `Json` as right now `Unknown` is returned.

This is quite tricky as from what I saw we get the methods and the types from `ClassDefinition` by using sth like this:
`lazy val unknown: Option[ClassDefinition] = get(classOf[java.lang.Object])`
And since Json is just a display type of `Unknown` the class definition is the same for both display types and we lose information about it and in result `DefaultDisplayStrategy` is used.

And also after changing these collections to work on `Json` types i get some random failures in Scala 2.12 tests...

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
```diff
@@ -61,6 +75,12 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
       .map(_.getOrElse(Map.empty))
   } yield Typed.record(fields, valueClass, additional)
 
+  private def unknown(obj: HCursor): Decoder.Result[TypingResult] = {
+    val displayAttribute = obj.downField("display")
+    if (displayAttribute.succeeded) displayAttribute.as[Unknown]
```
arkadius: It looks weird that you are decoding field `display` to `Unknown` type. I understand decoding `display` to `DisplayStrategy` but not to type. Let's change `as[Unkown]` to `as[DisplayStrategy]`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
```diff
@@ -42,6 +43,19 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
     case e                     => Left(s"$e is not SingleTypingResult")
   }
 
+  private implicit val displayUnknownDecoder: Decoder[Unknown] = Decoder.instance { cursor =>
+    cursor
```
arkadius: You can shorten it to `Decoder.decodeString.map { ...` - take a look how we do this in other places in this class.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
```diff
@@ -42,6 +43,19 @@ class TypingResultDecoder(loadClass: String => Class[_]) {
...
+      .as[String]
+      .map { displayStr =>
+        DisplayStrategy.valueOfDisplay(displayStr) match {
+          case Some(JsonDisplayStrategy)    => Typed.json
+          case Some(DefaultDisplayStrategy) => Unknown
+          case None =>
+            throw new IllegalStateException(s"Unrecognized display value: '$displayStr'")
```
arkadius: `IllegalStateException` should be reserved for cases when you put your application into a state that shouldn't happen. It is mostly used when you can't (or don't want to invest in it) express something in types and you have to handle something at runtime. The situation that you are handling here is a normal case - some client sent you unexpected data in a request. Circe prepared an error class for that. It is `DecodingFailure`. Also, it is more idiomatic to use `emap` or `.or(Decoder.failedWithMessage`.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
Cichan-mcc: I changed it to `as[DisplayStrategy]`. 
Previously I decided to go with `as[Unknown]` as that's the type that we actually want to be returned from the function and I took a little shortcut 😄

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
Cichan-mcc: I changed it to `Decoder.decodeString.emap`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/TypingResultDecoder.scala
Cichan-mcc: True, it should be a `DecodingFailure` instead of an exception. I used `emap` here with proper failure message now.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/compile/NodeDataValidatorSpec.scala
Cichan-mcc: I modified the code a bit and now `#input.toMap` returns `Map[String, Json]` and `#input.toList` returns `List[Json]`.
I also noticed that suggester suggesting `Undefined` for `get()` method is not specific to my case with json collections. It looks like it behaved like this with all of them.
Screenshot 2025-04-17 at 12 50 34
I inserted type computation for the suggester so the types are suggested properly. Now it works for Json as well as for other types
Screenshot 2025-04-17 at 12 53 35
Screenshot 2025-04-17 at 12 54 23

arkadius (approved):
Nice, let's merge this stage and we will do some end-to-end tests to find out if we didn't miss some place where `Json` should appear

PR #7853, opened 2025-04-01
[Tech] Add example scenarios to e2e tests

mgoworko:
## Describe your changes

Scenarios defined in the nussknacker-scenario-examples-library and deployed as part of the nussknacker-quickstart pipeline detected a bug, that was not detected by the Nu tests in this repo.

So in this PR 2 scenarios are copied to Nu repo - it is only tested, that they are compiled and deployed correctly, without testing their output for now.

## Checklist before merge


e2e-tests/src/test/resources/offer-customer-proposal-based-on-activity-event/OfferCustomerProposalBasedOnActivityEvent.json
```diff
@@ -0,0 +1,446 @@
+{
```
mgoworko, line 1: The entire example scenario directory copied without changes from examples library.

e2e-tests/src/test/resources/rtm-client-near-pos/RTMClientNearPOS.json
```diff
@@ -0,0 +1,660 @@
+{
```
mgoworko, line 1: The entire example scenario directory copied without changes from examples library.

e2e-tests/src/test/resources/batch-nu-designer.override.yml
```diff
@@ -2,13 +2,16 @@ services:
 
   designer:
     environment:
-      CONFIG_FILE: "/opt/nussknacker/conf/application.conf,/opt/nussknacker/conf/application-customizations.conf,/opt/nussknacker/conf/batch-customizations.conf"
+      EXAMPLE_SCENARIOS_LIBRARY_SERVICE_NAME: bootstrap-setup
+      CONFIG_FILE: "/opt/nussknacker/conf/application.conf,/opt/nussknacker/conf/application-customizations.conf,/opt/nussknacker/conf/batch-customizations.conf,/opt/nussknacker/conf/offer-customer-proposal-customizations.conf,/opt/nussknacker/conf/rtm-client-near-pos-customizations.conf"
```
mk-software-pl: it seems we can add `/opt/nussknacker/conf/additional-configuration.conf`. You can check its content. The `offer-customer-proposal-customizations.conf`, `rtm-client-near-pos-customizations.conf` confs should be included in it.

mk-software-pl: oh, and I see some inconsistency here. We add a file related to the bootstrap library, but this `override.yml` file refers to the batch test. I guess the `CONFIG_FILE ` is problematic here. Maybe it means that `e2e-tests/src/test/resources/bootstrap-setup-scenarios.override.yml` and this file should be merged. WDYT?

e2e-tests/src/test/scala/pl/touk/nussknacker/OfferCustomerProposalBasedOnActivityEventSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  ScenarioStatus
+}
+
+class OfferCustomerProposalBasedOnActivityEventSpec
+    extends AnyFreeSpecLike
+    with BaseE2ESpec
+    with Matchers
+    with BasicScenarioInformationFetching {
+
+  // The OfferCustomerProposalBasedOnActivityEvent is a highly dynamic scenario,
```
mk-software-pl: Hmmm. I wonder if we should use generators from the library. Maybe we should prepare a message in the test, put it in the Kafka topic, and read a potential output from the other topic? WDYT?

e2e-tests/src/test/scala/pl/touk/nussknacker/RtmClientNearPosSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class RtmClientNearPosSpec
+    extends AnyFreeSpecLike
+    with BaseE2ESpec
+    with Matchers
+    with VeryPatientScalaFutures
+    with BasicScenarioInformationFetching {
+
+  // The RTMClientNearPOS is a highly dynamic scenario,
```
mk-software-pl, line 20: the same here

examples/installation/flink-with-postgres/Dockerfile
```diff
@@ -0,0 +1,25 @@
+ARG FLINK_VERSION
```
mk-software-pl: maybe instead of having a separate `flink-with-postgres` folder, we could have another `Dockerfile` in the existing `flink` folder? E.g. `FlinkWithPostgresDriverCustomLibraryDockerfile`? 

AFAIS the flink-properties.yml is the same in both cases.

e2e-tests/src/test/resources/batch-nu-designer.override.yml
mgoworko: done, merged with bootstrap-setup-scenarios.override.yml and batch customizations included in the config changes handled by scenario examples

examples/installation/flink-with-postgres/Dockerfile
mgoworko: done

e2e-tests/src/test/scala/pl/touk/nussknacker/RtmClientNearPosSpec.scala
mgoworko: discussed, this scenario is not suitable for this kind of test - too dynamic

e2e-tests/src/test/scala/pl/touk/nussknacker/OfferCustomerProposalBasedOnActivityEventSpec.scala
mgoworko: in progress

e2e-tests/src/test/resources/rtm-client-near-pos/setup/kafka/topics.txt
```diff
@@ -0,0 +1,4 @@
+GeoLocations
```
lciolecki, line 1: topics with small letters and `-` as separator? :)

mk-software-pl: Do you want Michal to change it only here or in other examples too? 
If only in this place, I will say - it's better not to do it.

examples/installation/docker-compose.yml
```diff
@@ -250,3 +250,5 @@ volumes:
     name: nussknacker_designer_data
   nussknacker_flink_data:
     name: nussknacker_flink_data
+  nussknacker_designer_shared_configuration:
```
mk-software-pl: this is not needed in the case of the installation example. Why don't we add it int the bootstrap yaml?

e2e-tests/src/test/resources/rtm-client-near-pos/setup/kafka/topics.txt
lciolecki: I would prefer to use the kebab case convention rather than camelCase. But if that's a problem, just leave it.

examples/installation/docker-compose.yml
mgoworko: done, all references to this module in a single file

e2e-tests/src/test/scala/pl/touk/nussknacker/OfferCustomerProposalBasedOnActivityEventSpec.scala
mgoworko: added test with generating kafka messages and asserting result.But it is still too dynamic for more detailed assertions, for example discount is given randomly to 1/1000 customers and offers depend on the day of week :(

e2e-tests/src/test/resources/rtm-client-near-pos/setup/kafka/topics.txt
mgoworko: The current names follow the convention that is present in all examples - in scenario-examples and in Nu. If we want to change it, we would have to do it everywhere. AFAIK it would break compatiblity, if the scenario was already creaated and persisted in the postgres volume for quickstart.

mk-software-pl (approved):
LGTM

PR #7890, opened 2025-04-04
[NU-2128] Add toolbar configs

mslabek:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioToolbarService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioToolbarService.scala
@@ -35,19 +35,19 @@ object ScenarioToolbarSettings {
       topLeft = scenarioToolbarConfig.topLeft
         .filterNot(tp => verifyCondition(tp.hidden, scenario))
         .map(tp => ToolbarPanel.fromConfig(tp, scenario)),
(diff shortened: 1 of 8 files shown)


arkadius (approved):
LGTM

arkadius (changes requested):
I have a suggestion ...

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/ScenarioToolbarService.scala
```diff
@@ -54,7 +60,9 @@ final case class ScenarioToolbarSettings(
     topLeft: List[ToolbarPanel],
     bottomLeft: List[ToolbarPanel],
     topRight: List[ToolbarPanel],
-    bottomRight: List[ToolbarPanel]
+    bottomRight: List[ToolbarPanel],
```
arkadius: From the first glance I can't see that we have all combinations of positions covered. Also pattern of naming is counter-intuitive. My suggestion for order and naming:
- topLet
- topCenter
- topRight
- bottomLeft
- bottomCenter
- bottomRight

mslabek: Good point, I've changed this both the config case class and the http api case class and on FE.

arkadius (approved):

PR #7914, opened 2025-04-07
Change the order of scenario properties in ProcessingTypeData

gskrobisz:
## Describe your changes

Model config should be able to alter/override fixed scenario parameters configuration provided by deployment manager. E.g. when I want to set additional validators for the parameter "Schedule" (aka "cron")  for PeriodicDeploymentManager.

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
@@ -71,11 +71,11 @@ object ProcessingTypeData {
       processingType: ProcessingType,
       componentDefinitionExtractionMode: ComponentDefinitionExtractionMode
   ) = {
-    val scenarioProperties = modelData.modelConfig
+    val scenarioProperties = deploymentScenarioPropertiesConfig ++ modelData.modelConfig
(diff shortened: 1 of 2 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/processingtype/ProcessingTypeData.scala
```diff
@@ -71,11 +71,11 @@ object ProcessingTypeData {
       processingType: ProcessingType,
       componentDefinitionExtractionMode: ComponentDefinitionExtractionMode
   ) = {
-    val scenarioProperties = modelData.modelConfig
+    val scenarioProperties = deploymentScenarioPropertiesConfig ++ modelData.modelConfig
```
arkadius, line 75: There is some problem with key type in this map. I guess that it is a map of `PropertyName -> ScenarioPropertyConfig` instead of `ProcessingType -> ...`? The same problem is in `fragmentProperties`

gskrobisz: Switched ProcessingType to String (I noticed that we use String as property name and ParameterName as component parameter name). Left "todo" to fix it in next PR.

arkadius (approved):

piotrp (approved):

PR #7937, opened 2025-04-09
[NU-2120] Data sample from Kafka Source in Ad-Hoc test input

mateuszkp96:
## Describe your changes
Data sample from Kafka Source as initial input for Ad-Hoc test

image
image

## Checklist before merge


designer/server/src/test/scala/pl/touk/nussknacker/ui/api/testing/JsonSchemalessAdHocTestsSpec.scala
```diff
@@ -90,6 +96,10 @@ class JsonSchemalessAdHocTestsSpec
       "scenarioTypes.streaming.modelConfig.components.kafka.config.kafkaProperties",
       fromMap(Map("bootstrap.servers" -> "localhost:8070", "schema.registry.url" -> "http://localhost:8069").asJava)
     )
+    .withValue(
+      "scenarioTypes.streaming.modelConfig.components.kafka.config.useDataSampleParamForSchemalessJsonTopicBasedKafkaSource",
```
mgoworko, line 100: Is that config value described anywhere? It seems that it is the only usage. Is it needed?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
```diff
@@ -256,7 +271,14 @@ class UniversalKafkaSourceFactory(
           .extractParameterForTests(parsedSchema)
           .map(_.toParameters)
           .map { params =>
-            KafkaTestParametersInfo(params, prepareTestRecord(runtimeSchema, universalSchemaSupport, topic))
+            val enrichedParams = params.collect {
+              case param if defaultValuesForTestParameters.contains(param.name) =>
+                param.copy(
+                  defaultValue = Some(defaultValuesForTestParameters(param.name))
```
mgoworko, line 277: Maybe just `defaultValuesForTestParameters.get(param.name)` ? It will return Option. Then we wouldn't need collect and pattern matching, just a simple map.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
```diff
@@ -232,21 +246,22 @@ class UniversalKafkaSourceFactory(
...
+      topic: TopicName.ForSource,
+      defaultValuesForTestParameters: Map[ParameterName, Expression]
   )(
       implicit nodeId: NodeId
   ): KafkaTestParametersInfo = {
     Validated
       .fromOption(
         runtimeSchemaOpt,
-        NonEmptyList.one(CustomNodeError(nodeId.id, "Cannot generate test parameters: no runtime schema found", None))
+        NonEmptyList.one(CustomNodeError("Cannot generate test parameters: no runtime schema found", None))
```
mgoworko, line 264: Why is the nodeId removed here? Is it because of how the FE error is formatted? We may need it when there are multiple sources to identify which one has problems.

mgoworko (changes requested):

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/testing/JsonSchemalessAdHocTestsSpec.scala
mateuszkp96: Data sample params for Kafka Sources are disabled by default. This is an internal feature for now and applies only to a specific kind of Kafka Sources (UniversalKafkaSource without schema, with JSON content type). That's why it is not mentioned in the docs. We need to enable this feature to test the default value for ad-hoc tests.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
mateuszkp96: I used the apply method with nodeId as an implicit param (I think it's more typesafe)

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/source/UniversalKafkaSourceFactory.scala
mateuszkp96: I changed `collect` to `map`. Some parameters can already have a default value determined, so I'm not calling the copy method for all parameters to not set a None value for them.

mgoworko (approved):
LGTM

PR #7945, opened 2025-04-09
Support import of sticky notes, dont allow to replace node with stick…

philemone:
…yNote

## Checklist before merge


gskrobisz (approved):
LGTM

designer/client/src/components/graph/Graph.tsx
```diff
@@ -329,7 +330,7 @@ export class Graph extends React.Component<Props> {
         }
 
         const cellBelow = this.getCellBelowCell();
-        if (isModelElement(cellBelow)) {
+        if (isModelElement(cellBelow) && node.type != StickyNoteType) {
```
Dzuming, line 333: Can't we use `isModelOrStickyNote` utility here?

Dzuming (changes requested):

designer/client/src/components/graph/Graph.tsx
philemone: Nope since it is condition little different: "Is model **and isn't** sticky note"

Dzuming: ok, so another question. IIUC `isModelElement` contains `StickyNoteType`, so why do we need `isModelOrStickyNote` utility?

philemone: `isModelElement` does not contain `stickyNoteType` - it check if element is instance of `shapes.devs.Model`
All nodes in graph are constructed as `shapes.devs.Model.define('esp.Model', ...)` but sticky notes are defined as  `dia.Element.define('stickyNote.StickyNoteElement', ...);` 
 
Model is defined as `class Model extends basic.Generic extends dia.Element` so it is not exactly the same although it might seems this way :)

philemone: Also we don't want to treat notes and nodes exactly the same all the time - like in case of replacing nodes or unnecessary edges validation (since noted don't have any)

Dzuming (approved):

PR #7948, opened 2025-04-10
[NU-1948]  Add spel projection and selection expressions on avro GenericRecords

paw787878:
## Checklist before merge


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -710,6 +713,15 @@ private[spel] class Typer(
         if tc.runtimeObjType.canBeConvertedTo(Typed[java.util.Collection[_]]) ||
           tc.runtimeObjType.klass.isArray =>
       valid(tc.runtimeObjType.params.headOption.getOrElse(Unknown))
+    case tc: TypedObjectTypingResult if classOf[GenericRecord].isAssignableFrom(tc.runtimeObjType.klass) =>
```
arkadius, line 716: Use reflection. Thanks to that we won't be statically binded with a specific avro version in api module

PR #7953, opened 2025-04-11
[NU-1948] Add conversions from avro GenericRecord to java Map which fixes using #COLLECTION.merge() on GenericRecord, also add projection and selection on generic record

paw787878:
## Checklist before merge

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -852,8 +852,7 @@ lazy val scenarioCompiler = (project in file("scenario-compiler"))
       Seq(
         "org.typelevel"          %% "cats-effect"                   % catsEffectV,
         "org.scala-lang.modules" %% "scala-java8-compat"            % scalaCompatV,
-        // TODO_PAWEL needs to be removed
-        "org.apache.avro"         % "avro"                          % avroV,
+        "org.apache.avro"         % "avro"                          % avroV          % Test,
         "org.scalacheck"         %% "scalacheck"                    % scalaCheckV    % Test,
         "com.cronutils"           % "cron-utils"                    % cronParserV    % Test,
         "org.scalatestplus"      %% s"scalacheck-$scalaCheckVshort" % scalaTestPlusV % Test
--- a/build.sbt
+++ b/build.sbt
@@ -109,6 +109,8 @@ def defaultMergeStrategy: String => MergeStrategy = {
   case PathList(ps @ _*) if ps.last == "module-info.class"            => MergeStrategy.discard
   // we override Spring's class and we want to keep only our implementation
   case PathList(ps @ _*) if ps.last == "NumberUtils.class"            => MergeStrategy.first
+  case PathList(ps @ _*) if ps.last == "Projection.class"             => MergeStrategy.first
+  case PathList(ps @ _*) if ps.last == "Selection.class"              => MergeStrategy.first
   // merge Netty version information files
   case PathList(ps @ _*) if ps.last == "io.netty.versions.properties" => MergeStrategy.concat
   // due to swagger-parser dependencies having different schema definitions (json-schema-validator and json-schema-core)
@@ -311,6 +313,8 @@ val avroV                 = "1.11.4"
 //we should use max(version used by confluent, version acceptable by flink), https://docs.confluent.io/platform/current/installation/versions-interoperability.html - confluent version reference
 val kafkaV                = "3.8.1"
 // to update we need configurable SpEL length limit from 6.0.9, but 6.x requires JDK 17
+// when updating note that we have copied and modified class org.springframework.expression.spel.ast.Projection
+// and org.springframework.util.NumberUtils and org.springframework.expression.spel.ast.Selection
 val springV               = "5.2.23.RELEASE"
 val scalaTestV            = "3.2.18"
 val scalaCheckV           = "1.17.1"
@@ -848,7 +852,7 @@ lazy val scenarioCompiler = (project in file("scenario-compiler"))
       Seq(
         "org.typelevel"          %% "cats-effect"                   % catsEffectV,
         "org.scala-lang.modules" %% "scala-java8-compat"            % scalaCompatV,
-        "org.apache.avro"         % "avro"                          % avroV          % Test,
+        "org.apache.avro"         % "avro"                          % avroV,
         "org.scalacheck"         %% "scalacheck"                    % scalaCheckV    % Test,
         "com.cronutils"           % "cron-utils"                    % cronParserV    % Test,
         "org.scalatestplus"      %% s"scalacheck-$scalaCheckVshort" % scalaTestPlusV % Test
--- a/build.sbt
+++ b/build.sbt
@@ -852,6 +852,7 @@ lazy val scenarioCompiler = (project in file("scenario-compiler"))
       Seq(
         "org.typelevel"          %% "cats-effect"                   % catsEffectV,
         "org.scala-lang.modules" %% "scala-java8-compat"            % scalaCompatV,
+        // TODO_PAWEL needs to be removed
         "org.apache.avro"         % "avro"                          % avroV,
         "org.scalacheck"         %% "scalacheck"                    % scalaCheckV    % Test,
         "com.cronutils"           % "cron-utils"                    % cronParserV    % Test,
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
@@ -6,7 +6,6 @@ import cats.implicits.{catsSyntaxValidatedId, _}
 import org.apache.commons.lang3.ClassUtils
 import pl.touk.nussknacker.engine.api.typed.ConversionStrategy.NoConversion
 import pl.touk.nussknacker.engine.api.typed.typing._
-import pl.touk.nussknacker.engine.util.AssignabilityUtil
 
 /**
  * This class determine whether we can assign one type to another type - that is if its the same class, a subclass or can be converted to another type.
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
@@ -6,6 +6,7 @@ import cats.implicits.{catsSyntaxValidatedId, _}
 import org.apache.commons.lang3.ClassUtils
 import pl.touk.nussknacker.engine.api.typed.ConversionStrategy.NoConversion
 import pl.touk.nussknacker.engine.api.typed.typing._
+import pl.touk.nussknacker.engine.util.AssignabilityUtil
 
 /**
  * This class determine whether we can assign one type to another type - that is if its the same class, a subclass or can be converted to another type.
@@ -164,6 +165,7 @@ private[typed] object AssignabilityDeterminer {
           )
       }
     }
+
     val givenClass = from.runtimeObjType
 
     val equalClassesOrCanAssign =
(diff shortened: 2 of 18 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -179,18 +180,50 @@ object AssignabilityDeterminer {
...
 
-    val equalClassesOrCanAssign =
-      condNel(
-        givenClass == to,
-        (),
-        f"${givenClass.display} and ${to.display} are not the same"
-      ) orElse
-        isAssignable(givenClass.klass, to.klass)
+    // this check is done because we have GenericRecord to Map conversion in pl.touk.nussknacker.engine.spel.internal.DefaultSpelConversionsProvider
+    if (isGenericRecordToMapConversion(from, to, conversionChecker)) {
```
arkadius: Take a look, that we alread have `TypeConversionHandler` which is designed to keep all conversion related stuff in one place

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
```diff
@@ -179,18 +180,50 @@ object AssignabilityDeterminer {
...
+      canBeSubclass orElse conversionChecker.isConvertable(from, to)
+    }
+  }
+
+  private def isGenericRecordToMapConversion(
+      from: SingleTypingResult,
+      to: TypedClass,
+      conversionChecker: ConversionChecker
+  ): Boolean = {
+    if (!classOf[GenericRecord].isAssignableFrom(from.runtimeObjType.klass) || !classOf[java.util.Map[_, _]]
```
arkadius: 1. We can use `IndexedRecord` which is the highest possible abstraction that fits our needs
2. Use reflection. Thanks to that we won't be statically binded with a specific avro version in api module

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/DefaultSpelConversionsProvider.scala
```diff
@@ -31,6 +34,11 @@ class DefaultSpelConversionsProvider extends SpelConversionsProvider {
     service.addConverter(new ObjectToArrayConverter(service))
     // For purpose of concise usage of numbers in spel templates
     service.addConverter(classOf[Number], classOf[String], (source: Number) => source.toString)
+    service.addConverter(
+      classOf[GenericRecord],
+      classOf[java.util.Map[_, _]],
+      (r: GenericRecord) => r.getSchema.getFields.asScala.map(_.name()).map(n => n -> r.get(n)).toMap.asJava
```
arkadius: Use `Object get(int i);`. It is faster.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/AssignabilityDeterminer.scala
paw787878: Hmm, I need to do assignability check also on the generic type parameters of the map. I want to perform this check with the same conversionChecker. However I have no access to ConversionChecker in TypeConversionHandler

paw787878: I ended up moving it as requested

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -86,6 +96,35 @@ object TypeConversionHandler {
...
+  private def handleIndexedRecordToMapConversion(
+      from: SingleTypingResult,
+      to: TypedClass,
+      strict: Boolean
+  ): Boolean = {
+    if (!AssignabilityUtil.isAssignableToLoadableClass(
+        from.runtimeObjType.klass,
+        "org.apache.avro.generic.IndexedRecord"
+      ) ||
+      !ClassUtils.isAssignable(to.runtimeObjType.klass, classOf[java.util.Map[_, _]])) {
```
arkadius: 1. Use positive logic instead of negative logic. Also, I think that pattern matching would help to understand what's going on. A few lines below you use generic parameters. Thanks to pattern matching it will be easier to understand how checks of this parameters is related to checks of  `runtimeObjType`
2. Use `to.klass` instead of `to.runtimeObjType.klass`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -86,6 +96,35 @@ object TypeConversionHandler {
...
+      false
+    } else {
+      val indexedRecordKeyParam = Typed.genericTypeClass(classOf[String], List())
+      val indexedRecordValueParam = from match {
+        case fromCasted: TypedObjectTypingResult =>
+          superTypeOfTypes(fromCasted.fields.values)
+        case _ => Unknown
+      }
+      val isAssignable =
```
arkadius: Please rebase to my change: #7984 . In this change, conversion checking strategy is passed implicitly, so there is no need to recover it from boolean, which was computed based on strategy which was computed based on boolean ;)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -86,6 +96,35 @@ object TypeConversionHandler {
...
+        if (strict)
+          AssignabilityDeterminer.isAssignableStrict _
+        else
+          AssignabilityDeterminer.isAssignableLoose _
+
+      isAssignable(indexedRecordKeyParam, to.params.headOption.getOrElse(Unknown)).isValid &&
```
arkadius: 1. This `.headOption` / `.drop(1)` is difficult to read, we have pattern matching in scala - take a look at `AssignabilityDeterminer.isSingleAssignableToTypedClass` where we do similar thing
2. For `strict` strategy we shouldn't check it at all. `strict` strategy mean the same as `isAssignable` in java, but witch some tweaks such as unions, unknown etc. In java, `IndexedRecord` can't be converted to `Map`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -86,6 +96,35 @@ object TypeConversionHandler {
...
+      strict: Boolean
+  ): Boolean = {
+    if (!AssignabilityUtil.isAssignableToLoadableClass(
+        from.runtimeObjType.klass,
+        "org.apache.avro.generic.IndexedRecord"
+      ) ||
+      !ClassUtils.isAssignable(to.runtimeObjType.klass, classOf[java.util.Map[_, _]])) {
+      false
+    } else {
+      val indexedRecordKeyParam = Typed.genericTypeClass(classOf[String], List())
```
arkadius: 1. Just `Typed[String]`, `String` is not a generic type with zero list of parameters. We should probably have `Nel` instead of `List` for parameters.
2. Inline this, name of the variable doesn't help. `IndexedRecord` doesn't have key generic parameter

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
```diff
@@ -474,7 +474,7 @@ object typing {
     superTypeOfTypes(list.map(fromInstance))
   }
 
-  private def superTypeOfTypes(list: Iterable[TypingResult]) = {
+  def superTypeOfTypes(list: Iterable[TypingResult]) = {
```
arkadius: `private[typed]`, we want to expose the narrowest set of methods in the api. This is an api exposed for developers of components

build.sbt
```diff
@@ -848,7 +852,7 @@ lazy val scenarioCompiler = (project in file("scenario-compiler"))
       Seq(
         "org.typelevel"          %% "cats-effect"                   % catsEffectV,
         "org.scala-lang.modules" %% "scala-java8-compat"            % scalaCompatV,
-        "org.apache.avro"         % "avro"                          % avroV          % Test,
+        "org.apache.avro"         % "avro"                          % avroV,
```
arkadius: Revert `Test` scope

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/internal/DefaultSpelConversionsProvider.scala
```diff
@@ -31,6 +34,13 @@ class DefaultSpelConversionsProvider extends SpelConversionsProvider {
     service.addConverter(new ObjectToArrayConverter(service))
     // For purpose of concise usage of numbers in spel templates
     service.addConverter(classOf[Number], classOf[String], (source: Number) => source.toString)
+    service.addConverter(
+      classOf[GenericRecord],
+      classOf[java.util.Map[_, _]],
+      (r: GenericRecord) => {
```
arkadius: Unfortunately, we must do it with reflection as well - see the comment in build.sbt.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/typing.scala
paw787878: I ended up moving this method to CommonSupertypeFinder object

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
paw787878: I used more pattern matching. I am not sure if this is how You meant it, I did pattern match separately on "to" and on "from" object as I think it will lead to less code

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
```diff
@@ -71,7 +74,8 @@ private[engine] object TypeConversionHandler {
       implicit conversionStrategy: NonEmptyConversionStrategy
   ): Boolean = {
     handleImplicitConversion(from, to) ||
-    handleNumberConversion(from.runtimeObjType, to)
+    handleNumberConversion(from.runtimeObjType, to) ||
+    handleIndexedRecordToMapConversion(from.runtimeObjType, to)
```
arkadius: IndexRecord to Map conversion is an implicit conversion, which are handled in `handleImplicitConversion`. Maybe we have to work on the method name to make it more understandable? Do you have some idea? Maybe we should directly refer to `SpelConversions` instead of calling them "implicit conversions"?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/TypeConversionHandler.scala
arkadius: I pushed to your branch a commit with pattern matching-based, more idiomatic solution. Take a look at the difference: from the first glance, looking at pattern (case extraction logic) you see which fields will be used in evaluation. Isn't it simpler? After rewriting it to  pattern matching, one test behave differently. Previous behaviour was better or current is?

paw787878: The test that You changed was modeled after how it behaves on prod, "{a: 1}.?[#this.key == 'a'].get('a')" on prod is typed to Unknown. I will check why it changed, also tests for merge, namely "merge should work on avro genericRecord" failed, I will also take a look

arkadius: Ahh, I know what's going on, my solution support conversion to record with underlying Map, but doesn't support conversion to generic Map, with unknown fields. I can prepare a fix for that
[edit] no, it is something else

arkadius: Ok, I got it, I added in one place `.runtimeObjType` by mistake, please pull my change

arkadius (approved):

arkadius (approved):

PR #7964, opened 2025-04-16
[NU-2124] Add Json template parser and editor

lukasz-bigorajski:
## Describe your changes

image

## Checklist before merge

### DIFF
--- a/designer/client/src/brace/mode/jsonTemplate.js
+++ b/designer/client/src/brace/mode/jsonTemplate.js
@@ -65,7 +65,6 @@ ace.define("ace/mode/jsonTemplate", ["require", "exports", "ace/mode/jsonTemplat
     let JsonMode = acequire("ace/mode/json").Mode;
     let JsonTemplateHighlightRules = acequire("ace/mode/jsonTemplate_highlight_rules").JsonTemplateHighlightRules;
 
-    // TODO: lbg disalbe json workes - syntax checking, or add it whether it's easy
     var Mode = function () {
         // initialize all of JSON mode's internals (behaviors, foldingRules, etc.)
         JsonMode.call(this);
--- a/designer/client/src/brace/mode/jsonTemplate.js
+++ b/designer/client/src/brace/mode/jsonTemplate.js
@@ -1,0 +1,77 @@
+ace.define("ace/mode/jsonTemplate_highlight_rules", ["require", "exports", "module", "ace/lib/oop", "ace/mode/json_highlight_rules", "ace/mode/spel_highlight_rules"], function (acequire, exports, module) {
+    "use strict";
+
+    var oop = acequire("../lib/oop");
+    var JsonHighlightRules = acequire("ace/mode/json_highlight_rules").JsonHighlightRules;
+    var SpelHighlightRules = acequire("./spel_highlight_rules").SpelHighlightRules;
+
+    var JsonTemplateHighlightRules = function () {
+        // 1) Inherit all of the built‑in JSON rules
+        JsonHighlightRules.call(this);
+
+        // 2) Tweak the “property” rule so it DOESN’T match keys containing #{…}
+        let startRules = this.$rules.start;
+        let propRule = startRules.find(r => r.token === "variable");
+        if (propRule) {
+          // negative lookahead (?![^"]*#\{) means “no #{ anywhere before the closing quote”
+          propRule.regex = '"(?![^"]*#\\{)(?:\\\\.|[^\\\\"])*"(?=\\s*:)';
+        }
+
+        // 3) Inject template opener into both root AND string states
+        let opener = {
+            token: "meta.block-marker.jsonTemplate",
+            regex: /#\{/,
+            push: "jsonTemplate-start"
+        };
+
+        // 3) …and inject it *both* into the root and into the JSON-string state
+        this.$rules.start.unshift(opener);
+        this.$rules.string.unshift(opener);
+
+        // 4) Existing SpEL embed‑and‑end logic
+        let endRules = [{
+            token: "meta.block-marker.jsonTemplate",
+            regex: /[{}]/,
+            onMatch: function (val, state, stack) {
+                if(state?.startsWith("jsonTemplate-string")) {
+                    return "string";
+                }
+                this.next = "";
+                if (val === "{") {
+                    stack.unshift("jsonTemplate-start", state);
+                    return "paren";
+                }
+                stack.shift();
+                this.next = stack.shift() || "start";
+                return this.next === "jsonTemplate-start" ? "paren" : "meta.block-marker.jsonTemplate";
+
+            }
+        }];
+
+        // 5) Finally wire in the SPEL/template rules
+        this.embedRules(SpelHighlightRules, "jsonTemplate-", endRules);
+        this.normalizeRules();
+    };
+
+    oop.inherits(JsonTemplateHighlightRules, JsonHighlightRules);
+
+    exports.JsonTemplateHighlightRules = JsonTemplateHighlightRules;
+
+});
+
+ace.define("ace/mode/jsonTemplate", ["require", "exports", "ace/mode/jsonTemplate_highlight_rules", "ace/mode/spel", "ace/mode/json"], function (acequire, exports, module) {
+    "use strict";
+    var oop = acequire("../lib/oop");
+    let JsonMode = acequire("ace/mode/json").Mode;
+    let JsonTemplateHighlightRules = acequire("ace/mode/jsonTemplate_highlight_rules").JsonTemplateHighlightRules;
+
+    // TODO: lbg disalbe json workes - syntax checking, or add it whether it's easy
+    var Mode = function () {
+        // initialize all of JSON mode's internals (behaviors, foldingRules, etc.)
+        JsonMode.call(this);
+        // then override just the HighlightRules
+        this.HighlightRules = JsonTemplateHighlightRules;
+    };
+    oop.inherits(Mode, JsonMode);
+    exports.Mode = Mode;
+});
--- a/designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
@@ -1,4 +1,4 @@
-import type { ForwardedRef, ReactNode} from "react";
+import type { ForwardedRef, ReactNode } from "react";
 import React, { forwardRef, useCallback, useMemo } from "react";
 import type ReactAce from "react-ace/lib/ace";
 import { useTranslation } from "react-i18next";
@@ -98,6 +98,10 @@ Use autocompletion to explore available options. To read more see [Documentation
                     )}
                 />
             );
+        }
+
+        if (editorMode === EditorMode.SpELTemplate && !readOnly) {
+            // properties.
         }
 
         return properties;
--- a/designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
@@ -100,8 +100,21 @@ Use autocompletion to explore available options. To read more see [Documentation
             );
         }
 
-        if (editorMode === EditorMode.SpELTemplate && !readOnly) {
-            // properties.
+        if (editorMode === EditorMode.JsonTemplate && !readOnly) {
+            properties.placeholder = placeholder || t("editors.jsonTemplateEditor.placeholder", 'e.g. { "key": "#{ #input.value }" }');
+            properties.InputAdornmentEnd = (
+                <InfoTooltip
+                    text={t(
+                        "editors.jsonTemplateEditor.infoText",
+                        `You are using a json-template-based input, allowing json with embedded expressions. \n 
+Embed expression with \`#{ }\`, e.g., \`{ "name": #{ #input.name } }\`. When accessing variables that support dynamic fields you can use \`#input['dynamicField'].toTargetType\`, e.g. \`#input['accountNo'].toLong\`. \n
+Use autocompletion to explore available options. To read more see [Documentation](https://nussknacker.io/documentation/docs/scenarios_authoring/Spel)`,
+                    )}
+                />
+            );
+        }
+        if (editorMode === EditorMode.JsonTemplate) {
+            properties.useAceWorker = false;
         }
 
         return properties;
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
@@ -23,12 +23,12 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
   ): Option[Expression] = {
     // TODO: use classes instead of class names
     Option(className).collect {
-      case className if TypeValueDeterminer.isIntegerNumberClass(className)      => Expression.spel("0")
-      case className if TypeValueDeterminer.isLikeFloatingPointNumber(className) => Expression.spel("0.0")
-      case className if TypeValueDeterminer.isBoolean(className)                 => Expression.spel("true")
-      case className if TypeValueDeterminer.isString(className)                  => defaultStringExpression(editor)
-      case className if TypeValueDeterminer.isList(className)                    => Expression.spel("{}")
-      case className if TypeValueDeterminer.isMap(className)                     => Expression.spel("{:}")
+      case className if TypeValueDeterminer.isIntegerNumber(className)       => Expression.spel("0")
+      case className if TypeValueDeterminer.isFloatingPointNumber(className) => Expression.spel("0.0")
+      case className if TypeValueDeterminer.isBoolean(className)             => Expression.spel("true")
+      case className if TypeValueDeterminer.isString(className)              => defaultStringExpression(editor)
+      case className if TypeValueDeterminer.isList(className)                => Expression.spel("{}")
+      case className if TypeValueDeterminer.isMap(className)                 => Expression.spel("{:}")
     }
   }
 
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
@@ -23,22 +23,20 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
   ): Option[Expression] = {
     // TODO: use classes instead of class names
     Option(className).map(_.getName).collect {
-      case "long" | "short" | "int" | "java.lang.Number" | "java.lang.Long" | "java.lang.Short" | "java.lang.Integer" |
-          "java.math.BigInteger" =>
-        Expression.spel("0")
-      case "float" | "double" | "java.math.BigDecimal" | "java.lang.Float" | "java.lang.Double" =>
-        Expression.spel("0.0")
-      case "boolean" | "java.lang.Boolean" => Expression.spel("true")
-      case "java.lang.String"              => defaultStringExpression(editor)
-      case "java.util.List"                => Expression.spel("{}")
-      case "java.util.Map"                 => Expression.spel("{:}")
+      case className if TypeValueDeterminer.isLikeIntegerNumber(className)       => Expression.spel("0")
+      case className if TypeValueDeterminer.isLikeFloatingPointNumber(className) => Expression.spel("0.0")
+      case className if TypeValueDeterminer.isBoolean(className)                 => Expression.spel("true")
+      case className if TypeValueDeterminer.isString(className)                  => defaultStringExpression(editor)
+      case className if TypeValueDeterminer.isList(className)                    => Expression.spel("{}")
+      case className if TypeValueDeterminer.isMap(className)                     => Expression.spel("{:}")
     }
   }
 
   private def defaultStringExpression(editor: Option[ParameterEditor]): Expression =
     EditorBasedLanguageDeterminer.determineLanguageOf(editor) match {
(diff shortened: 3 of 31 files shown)


designer/client/src/brace/mode/jsonTemplate.js
```diff
@@ -0,0 +1,77 @@
...
+
+});
+
+ace.define("ace/mode/jsonTemplate", ["require", "exports", "ace/mode/jsonTemplate_highlight_rules", "ace/mode/spel", "ace/mode/json"], function (acequire, exports, module) {
+    "use strict";
+    var oop = acequire("../lib/oop");
+    let JsonMode = acequire("ace/mode/json").Mode;
+    let JsonTemplateHighlightRules = acequire("ace/mode/jsonTemplate_highlight_rules").JsonTemplateHighlightRules;
+
+    // TODO: lbg disalbe json workes - syntax checking, or add it whether it's easy
```
Dzuming: What about this TODO? Are you going to resolve it now?

designer/client/src/brace/mode/jsonTemplate.js
```diff
@@ -0,0 +1,77 @@
+ace.define("ace/mode/jsonTemplate_highlight_rules", ["require", "exports", "module", "ace/lib/oop", "ace/mode/json_highlight_rules", "ace/mode/spel_highlight_rules"], function (acequire, exports, module) {
```
Dzuming, line 1: Shouldn't we validate this case somehow?

image

vs

image

designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
```diff
@@ -100,6 +100,23 @@ Use autocompletion to explore available options. To read more see [Documentation
...
+                        "editors.jsonTemplateEditor.infoText",
+                        `You are using a json-template-based input, allowing json with embedded expressions. \n 
+Embed expression with \`#{ }\`, e.g., \`{ "name": #{ #input.name } }\`. When accessing variables that support dynamic fields you can use \`#input['dynamicField'].toTargetType\`, e.g. \`#input['accountNo'].toLong\`. \n
+Use autocompletion to explore available options. To read more see [Documentation](https://nussknacker.io/documentation/docs/scenarios_authoring/Spel)`,
+                    )}
+                />
+            );
+        }
+        if (editorMode === EditorMode.JsonTemplate) {
+            properties.useAceWorker = false;
```
Dzuming, line 117: Why do we need to disable it? Shouldn't we also disable it in case of JsonEditor?

Dzuming (changes requested):

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
```diff
@@ -23,22 +23,20 @@ protected object TypeRelatedParameterValueDeterminer extends ParameterDefaultVal
...
-      case "long" | "short" | "int" | "java.lang.Number" | "java.lang.Long" | "java.lang.Short" | "java.lang.Integer" |
-          "java.math.BigInteger" =>
-        Expression.spel("0")
-      case "float" | "double" | "java.math.BigDecimal" | "java.lang.Float" | "java.lang.Double" =>
-        Expression.spel("0.0")
-      case "boolean" | "java.lang.Boolean" => Expression.spel("true")
-      case "java.lang.String"              => defaultStringExpression(editor)
-      case "java.util.List"                => Expression.spel("{}")
-      case "java.util.Map"                 => Expression.spel("{:}")
+      case className if TypeValueDeterminer.isLikeIntegerNumber(className)       => Expression.spel("0")
```
arkadius: I think, that we don't need need this "like". Let's rename it to `isDecimalNumberClass`. The same with `isFloatingPointNumberClass`. BTW, can we use `Class[_]` instead of class name? If we have a dedicated value class for that, why not use it?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
```diff
@@ -58,6 +58,7 @@ object JsonParser extends ExpressionParser {
 case class JsonExpressionTypingInfo(json: Json) extends ExpressionTypingInfo {
 
   override val typingResult: TypingResult = {
+    // TODO: For floating point type it is returning Integer and it's ok for now but maybe we should change it in the future.
```
arkadius: Can you explain it? I see in tests that:
```
    FromJsonSimpleDecoder.jsonToAny(
      Json.fromBigDecimal(java.math.BigDecimal.valueOf(Double.MaxValue))
    ) shouldBe java.math.BigDecimal.valueOf(Double.MaxValue)
```
for the `Double` it returns `BigDecimal`. Maybe it works that way for round floating point numbers?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,171 @@
...
+  override def languageId: Expression.Language = Expression.Language.JsonTemplate
+
+  override def parse(
+      original: String,
+      ctx: ValidationContext,
+      expectedType: typing.TypingResult
+  ): ValidatedNel[ExpressionParseError, TypedExpression] =
+    spelTemplateParser
+      .parse(original, ctx, stringTypingResult)
+      .andThen(parsedSpelTemplateExpression =>
```
arkadius: We use 
```
.andThen { parsedSpelTemplateExpression =>
}
```
syntax for lambdas that are not one-linears

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParserTest.scala
```diff
@@ -0,0 +1,206 @@
...
+    val dataSample =
+      s"""{
+         |  "name": "#{#name}",
+         |  "age": #{#age},
+         |  "hasConsent": #{#hasConsent},
+         |  "amount": #{#amount}
+         |}""".stripMargin
+
+    val result = parse[String](dataSample, ctxWithVariables).validValue.expression
+      .evaluate[String](evaluationContext, Map.empty)
```
arkadius: result should be a `java.util.Map[String, _]`

designer/client/src/brace/mode/jsonTemplate.js
lukasz-bigorajski: Fixed

designer/client/src/components/graph/node-modal/editors/expression/SpelEditor.tsx
lukasz-bigorajski: I added explainer comment and also disabled it in case of JsonEditor. With this enabled it want to validate json syntax correctness, but it cannot be done with ace because it requires another dependency

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParserTest.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonParser.scala
lukasz-bigorajski: I removed the comment, the situation occurred for: `100.0` and then it's was treated as integer, so I changed implementation to take default value with number after decimal point present

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/definition/component/parameter/defaults/TypeRelatedParameterValueDeterminer.scala
lukasz-bigorajski: Done

designer/client/src/brace/mode/jsonTemplate.js
lukasz-bigorajski: Removed. It's resolved but I had forgotten to remove it

Dzuming (approved):
FE LGTM

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
arkadius: I can't see the change. This `andThen` still uses square braces.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+          JsonTemplateExpressionTypingInfo(jsonTypeExpression.typingInfo, expectedType),
+        )
+      }
+
+  override def parseWithoutContextValidation(
+      original: String,
+      expectedType: typing.TypingResult
+  ): ValidatedNel[ExpressionParseError, CompiledExpression] = spelTemplateParser
+    .parseWithoutContextValidation(original, stringTypingResult)
+    .andThen(parsedSpelTemplateExpression =>
```
arkadius: The same here

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+      case expression: SpelExpression =>
+        joinLiteralsAndReplaceExpressionWithDefaults(
+          spelExpressionToDefaultJsonValue(ctx),
+          expression.parsedSpringExpression,
+        )
+      case _ => invalidNel(ExpressionCompilationError("Invalid compiled expression type"))
+    }
+
+  private def joinLiteralsAndReplaceExpressionWithDefaults(
+      spelExpressionToDefaultJsonValue: (String, typing.TypingResult) => ValidatedNel[ExpressionParseError, String],
```
arkadius: It is hard to understand what's going on with this function partial application and passing it resureviely. Maybe we could do something to make it easier? A few options that I see:
- extraction of case class for parameters of this function with well-named fields (the easiest)
- extraction of trait with well-named method and parameters with two implemantions - because as I understand it correctly, you have two behaviours for `parseWithoutContextValidation` and `parse` 
- unification of `parseWithoutContextValidation` and `parse` code
- replacing some methods with local methods 
WDYT?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+      expectedType: typing.TypingResult
+  ): ValidatedNel[ExpressionParseError, String] = spelParser
+    .parse(original, ctx, expectedType)
+    .map(_.typingInfo.typingResult)
+    .map(typingResultToDefaultJsonValue)
+
+  private def typingResultToDefaultJsonValue(typingResult: TypingResult): String = typingResult match {
+    case TypedClass(k, _) =>
+      k match {
+        case className if TypeValueDeterminer.isIntegerNumber(className)       => "0"
```
arkadius: It is not a className, it just a class, you can you `klass`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+        case className if TypeValueDeterminer.isFloatingPointNumber(className) => "0.5"
```
arkadius: Magic number. Let's extract a well-named constant for that e.g. `placeholderForFloatingPoint` with the comment that the value is not important, the only requirement that it should fulfil is that it can't be rounded.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+        case className if TypeValueDeterminer.isBoolean(className)             => "true"
+        case className if TypeValueDeterminer.isString(className)              => "unquoted string"
+        // For now, complex types are treated as String
+        case _ => "unquoted string"
+      }
+    // For now, complex types are treated as String
```
arkadius: Let's add: ... In runtime, `.toString` is invoked on these types. Sometimes it can be ok, but in another situation the result of `.toString` can be unexpected for the user.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+        )
+      case _ => invalidNel(ExpressionCompilationError("Invalid compiled expression type"))
+    }
+
+  private def spelExpressionToDefaultJsonValue(
+      original: String,
+      expectedType: typing.TypingResult
+  ): ValidatedNel[ExpressionParseError, String] = spelParser
+    .parseWithoutContextValidation(original, expectedType)
+    .map(_ => "0")
```
arkadius: Magic number. Why "0"? I don't understand it.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+      templateCompiledExpression: CompiledExpression,
+      expectedType: typing.TypingResult
+  ) extends CompiledExpression {
+
+    override def language: Language = languageId
+
+    override def original: String = originalJsonString
+
+    override def evaluate[T](ctx: Context, globals: Map[String, Any]): T =
+      if (expectedType == stringTypingResult) {
```
arkadius: IMO it is not a good behaviour for `String`. This:
```
"foo"
```
is a valid json. Also this:
```
123
```
is a valid json. This is too:
```
null
```
Let's say that someone know that in the passed json template, will be a string. They wanted to get just the `foo` (without quotes). Thanks to current "magic" it will be quoted `"foo"`. If we remove this if, everything should work as expected. For now, I don't see requirement to add such a magic so following YAGNI/KISS rules, let's remove it.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+        templateCompiledExpression.evaluate[T](ctx, globals)
+      } else {
+        val jsonString = templateCompiledExpression.evaluate[String](ctx, globals)
+        parser.parse(jsonString) match {
+          case Left(error)  => throw new IllegalStateException("Parsing JSON failed with error", error)
```
arkadius: `IllegalStateException` means that we didn't expect something. We encounter this situation when, e.g. someone will put something to a placeholder (for example list), which will break the json. We know that it may happen, and we accept it. I think that it will be better to throw our own exception with error as a cause or just error which is also an exception (see `ParsingFailure`)

arkadius (commented):
Nice, some minor comments added, but in general it looks good

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+  ): ValidatedNel[ExpressionParseError, String] = spelParser
+    .parseWithoutContextValidation(original, expectedType)
+    .map(_ => "0")
+
+}
+
+object JsonTemplateParser {
+  private val stringTypingResult = Typed.typedClass[String]
+
+  case class CompiledJsonTemplateExpression(
```
arkadius: It is not a data class, it is not usable for pattern matching, equals and serialization to json ;). Let's switch to normal class.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+    override def original: String = originalJsonString
+
+    override def evaluate[T](ctx: Context, globals: Map[String, Any]): T =
+      if (expectedType == stringTypingResult) {
+        templateCompiledExpression.evaluate[T](ctx, globals)
+      } else {
+        val jsonString = templateCompiledExpression.evaluate[String](ctx, globals)
+        parser.parse(jsonString) match {
+          case Left(error)  => throw new IllegalStateException("Parsing JSON failed with error", error)
+          case Right(value) => FromJsonSimpleDecoder.jsonToAny(value).asInstanceOf[T]
```
arkadius: Have you tried to use `FromJsonTypingResultBasedDecoder` instead of the simple decoder? I think that it would be especially better, when we decide to use this parser in the middle of a scenario - For example for event generator, for the value parameter. With the current approach, `#DATE.now` will be typed as a `String` and in runtime user will got the `String`. With  `FromJsonTypingResultBasedDecoder` it could be `Instant`. My only doubt is that I see that it won't be easy to type such an expression to get this `type` argument. Am I right?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -0,0 +1,181 @@
...
+    override def language: Language = languageId
+
+    override def original: String = originalJsonString
+
+    override def evaluate[T](ctx: Context, globals: Map[String, Any]): T =
+      if (expectedType == stringTypingResult) {
+        templateCompiledExpression.evaluate[T](ctx, globals)
+      } else {
+        val jsonString = templateCompiledExpression.evaluate[String](ctx, globals)
+        parser.parse(jsonString) match {
```
arkadius: Let's use `.map(...).getOrElse(...)`. Currently, the second part of processing (JSON decoding) is a little bit hidden inside pattern matching.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done - extracted trait and case class

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done - it's because integer works well with and without quotes well

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: In json there is not a date data type, so I skipped it. JsonParser will not allow dates without quote so it will be basic String

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
lukasz-bigorajski: Done

arkadius (approved):
Awesome

PR #7969, opened 2025-04-18
Set of small cosmetic docs improvements.

ZbyszekMM:
The changes are tiny and unrelated. 
The biggest - although still small - is a partial rework of Transforming Lists section of SpEL cheat sheet.

### DIFF
--- a/docs/scenarios_authoring/Spel.md
+++ b/docs/scenarios_authoring/Spel.md
@@ -190,23 +190,20 @@ To obtain the last matching element, the syntax is `.$`.
 ### Transforming lists
 
 Special variable `#this` is used to operate on single element of list.
-            
-Examples below assume following structure:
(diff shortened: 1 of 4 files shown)


docs/scenarios_authoring/Spel.md
```diff
@@ -190,23 +190,20 @@ To obtain the last matching element, the syntax is `.$`.
 ### Transforming lists
 
 Special variable `#this` is used to operate on single element of list.
-            
-Examples below assume following structure:
```
ZbyszekMM, line 194: To me this is confusing, if not completely wrong. We do not have an assignment ('=') operator in SpEL; we should not suggest such exists. The references to person1, person2 are incorrect. Finally the toMap() method requires some explanation.

arkadius (approved):

PR #7970, opened 2025-04-18
[NU-2009] feature: active scenarios limit (global and per processing type limits)

mk-software-pl:
## Describe your changes

In this PR:
1. Limits service - checking active scenarios limits per processing type and globally (without division into processing types)
1. Introducing active scenarios limits validation before scenario deployment (for both: the new deployment service and the old deployment service)
2. Some small refactorings

## Checklist before merge


docs/configuration/DesignerConfiguration.md
```diff
@@ -799,6 +800,7 @@ Scenario type configuration consists of parts:
 - `deploymentConfig` - [scenario deployment configuration](./ScenarioDeploymentConfiguration.md)
 - `modelConfig` - [model configuration](./model/ModelConfiguration.md)
 - `category` - category handled by given scenario type
+- `activeScenariosLimit` (optional) - it's the maximum count of active scenarios for the scenario type
```
lciolecki: With all respect, I think we should not inform about that kind of features ;)

coutoPL: There is no doubt that it can be useful for OSS users.
But if you guys want to hide it (for some not very well-specified reason), I can do it.

mk-software-pl: There is no doubt that it can be useful for OSS users.
But if you guys want to hide it (for some not very well-specified reason), I can do it.

lciolecki (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/DeploymentService.scala
```diff
@@ -157,7 +174,30 @@ class DeploymentService(
...
+  ): EitherT[Future, RunDeploymentError, Unit] = {
+    implicit val loggedUser: LoggedUser = deployer
+    activeScenariosLimitProvider.forProcessingType(scenarioMetadata.processingType).flatten match {
+      case Some(ActiveScenariosLimit(activeScenariosLimit)) =>
+        EitherT {
+          scenarioStatusProvider
+            .getActiveScenariosCountFor(scenarioMetadata.processingType)
+            .map { activeScenariosCount =>
+              Either.cond(
+                test = activeScenariosCount < activeScenariosLimit,
```
arkadius: 1. For me, a little bit more readable would be `currentlyActiveScenariosCount + 1 <= activeScenariosLimit` (thanks to that we emphasise that we check the simulated state after deployment
2.  What about `DeploymentUpdateStrategy`? Unfortunatelly we still don't have separated commands for deploy and redeploy, @gskrobisz is working on splitting them. For redeploy we should check `currentlyActiveScenariosCount <= activeScenariosLimit`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
```diff
@@ -78,6 +78,20 @@ class ScenarioStatusProvider(
...
   }
 
+  def getActiveScenariosCountFor(processingType: ProcessingType)(implicit user: LoggedUser): Future[Int] = {
+    implicit val freshnessStatus: DataFreshnessPolicy = DataFreshnessPolicy.Fresh
+    deploymentStatusesProvider
+      .getBulkQueriedDeploymentStatusesForSupportedManagers(processingType :: Nil)
+      .map { statuses =>
+        statuses.getAllDeploymentStatuses
+          .count { status =>
+            status.status == SimpleStateStatus.DuringDeploy ||
```
arkadius: Let's extract a `Set` for that

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
```diff
@@ -78,6 +78,20 @@ class ScenarioStatusProvider(
     )
   }
 
+  def getActiveScenariosCountFor(processingType: ProcessingType)(implicit user: LoggedUser): Future[Int] = {
+    implicit val freshnessStatus: DataFreshnessPolicy = DataFreshnessPolicy.Fresh
+    deploymentStatusesProvider
```
arkadius: There is a little problem with this. The `EngineSideDeploymentStatusesProvider` doesn't take into consideration the little period when deployment is requested but, still not available on the engine side. It is why the name starts from `Engine`. It is why we have this:
```
    } else if (inProgressActionNames.contains(ScenarioActionName.Deploy)) {
      logStatusAndReturn(SimpleStateStatus.DuringDeploy)
```
check before we check status on the engine side in `getScenarioStatus` method.

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/GlobalLimitsConfig.scala
```diff
@@ -0,0 +1,11 @@
+package pl.touk.nussknacker.ui.limits
+
+import pl.touk.nussknacker.ui.limits.GlobalLimitsConfig.ActiveScenariosLimit
+
+final case class GlobalLimitsConfig(activeScenariosLimit: Option[ActiveScenariosLimit])
```
arkadius: WDYT about `maxActiveScenariosCount`? IMO it will be more explicit than "limit". Limit can be either upper or lower.

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/LimitsService.scala
```diff
@@ -0,0 +1,158 @@
...
+      deployingScenario: ProcessName,
+      deploymentUpdateStrategy: DeploymentUpdateStrategy,
+  ): EitherT[IO, ActiveScenariosLimitExceededError, Unit] = {
+    globalLimitsConfig.activeScenariosLimit match {
+      case Some(activeScenariosLimit) =>
+        EitherT {
+          implicit val user: LoggedUser = NussknackerInternalUser.instance
+          val allProcessingTypes        = activeScenariosLimitProvider.all.keys
+          scenarioStatusProvider
+            .getActiveScenariosFor(allProcessingTypes)
```
arkadius: Let's move operations of active scenarios fetching before limit checks to reduce number of calls to flink API

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -151,6 +152,23 @@ class DeploymentService(
     } yield ()
   }
 
+  private def checkActiveScenariosLimits(
+      scenario: ScenarioWithDetailsEntity[CanonicalProcess],
+      deployer: LoggedUser,
+      deploymentUpdateStrategy: DeploymentUpdateStrategy,
+  ) = {
+    implicit val loggedUser: LoggedUser = deployer
+    // note: we use "unsafe" version here (without critical section), because we're already in the actions critical section,
```
arkadius: ... if we assume that nobody will spawn two deployments via both api concurrently :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/LimitsService.scala
```diff
@@ -0,0 +1,158 @@
...
+import pl.touk.nussknacker.ui.process.processingtype.provider.ProcessingTypeDataProvider
+import pl.touk.nussknacker.ui.security.api.{LoggedUser, NussknackerInternalUser}
+
+class LimitsService(
+    globalLimitsConfig: GlobalLimitsConfig,
+    activeScenariosLimitProvider: ProcessingTypeDataProvider[LimitsConfig, _],
+    scenarioStatusProvider: ScenarioStatusProvider
+) extends LazyLogging {
+
+  def checkScenarioLimitsBeforeDeploymentUnsafe(
```
arkadius: Maybe let's call it `...NoCrititicalSection` instead of `...Unsafe`. I thought that it had non-monadic return type :)

arkadius:
Let's change the PR title, to avoid misleading message in the commit - it is not "per processing type" anymore

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/LimitsService.scala
coutoPL: the method is removed

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
coutoPL: I thought that the critical section is used in both deployment services. Tests showed me that it's not. I refactored the code a little bit, and now the "unsafe" method is not needed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/LimitsService.scala
coutoPL: I'm not sure what you ask me to do here.
If you write about moving the call to `getActiveScenariosFor` above these two limit checking methods, notice the params are different in these two invocations. To do that, I'd need to modify the `getActiveScenariosFor` to return not only the scenario names but also processing types related to them. 

I can do that, but I'm not sure if it's worth to do it now.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentStatus.scala
```diff
@@ -9,13 +9,22 @@ import io.circe.generic.extras.semiauto.deriveUnwrappedCodec
 // descriptions depending on DM implementation. It makes implementation of logic based on statuses easier. In case
 // if we have requirement to make it more flexible, we can relax this restriction.
 sealed trait DeploymentStatus extends EnumEntry with UpperSnakecase {
-  def name: DeploymentStatusName = DeploymentStatusName(entryName)
+  def name: DeploymentStatusName
+  def description: Option[String]
```
arkadius: Let's call it `problemDescription` instead of generic description to avoid some not intended usage. I'm surprised that I have to ask, but why in this case it is better to use polymorphism instead of pattern matching as it was before? It looks strange that generic deployment status provides some (optional) information about it children class.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/ScenarioStatusProvider.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import cats.effect.IO
+import pl.touk.nussknacker.engine.api.process.{ProcessingType, ProcessName}
+import pl.touk.nussknacker.ui.process.repository.DBIOActionRunner
+import pl.touk.nussknacker.ui.security.api.LoggedUser
+
+class ScenarioStatusProvider(deploymentRepository: DeploymentRepository, dbioRunner: DBIOActionRunner) {
+
+  def getActiveScenariosFor(
+      processingTypes: Iterable[ProcessingType]
```
arkadius, line 17: `user` is not used. It means that we don't check if user has access to given deployment/processingType. Check `newdeployment.DeploymentService.getDeploymentStatus` where we do similiar thing for one deployment

designer/server/src/main/scala/pl/touk/nussknacker/ui/limits/LimitsService.scala
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  private def checkCurrentlyActiveScenariosCount(
+      currentlyDeployingScenario: ProcessName,
+      currentlyActiveScenarios: Set[ProcessName],
+      maxActiveScenariosCount: Int,
+      deploymentUpdateStrategy: DeploymentUpdateStrategy,
+  ) = {
+    val maxActiveScenariosIncludingCurrentlyDeployingScenario = deploymentUpdateStrategy match {
```
arkadius: Please don't modify both left and right sides of an expression.
This expression is as simple as:
```
val activeScenariosCountAfterDeployment = deploymentUpdateStrategy match {
// case for redeploy => currentlyActiveScenarios.size
// case for normal deploy  => currentlyActiveScenarios.size + 1
}
activeScenariosCountAfterDeployment <= maxActiveScenariosCount
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
```diff
@@ -78,6 +80,36 @@ class ScenarioStatusProvider(
     )
   }
 
+  def getActiveScenariosFor(
```
arkadius, line 86: This approach doesn't use status prefetching approach which give performance benefits for >1 scenario status fetching. Do we care about that? Take a look at:
```
        prefetchedDeploymentStatuses <- DBIO.from(
          deploymentStatusesProvider.getBulkQueriedDeploymentStatusesForSupportedManagers(
            processingTypesWithMoreThanOneScenario
          )
        )
```
in method above

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
```diff
@@ -78,6 +80,36 @@ class ScenarioStatusProvider(
...
+          scenarios <- processRepository.fetchLatestProcessesDetails[Unit](
+            ScenarioQuery.empty
+              .copy(isFragment = Some(false), isArchived = Some(false), processingTypes = Some(processingTypes))
+          )
+          activeScenarios <- DBIOAction
+            .sequence {
+              scenarios.map { process =>
+                for {
+                  inProgressActionNames <- actionRepository.getInProgressActionNames(process.processId)
+                  scenarioStatus <- getScenarioStatusFetchingDeploymentsStatusesFromManager(
```
arkadius: There is another corner case here. At the end of this method is invoked `InconsistentStateDetector`. It can return `ProblemStateStatus.shouldNotBeRunning(true)` which means that scenario s running but based on the local, designer state it is an unexpected situation. Do you think that we should care about? Currently, you count it as inactive scenario but in fact it runs on Flink.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentStatus.scala
coutoPL: I added it here because of the `def name` to follow the convention. But I agree I could improve it because I've already done some refactoring around these things.

Now, there is no def in it (and explicit relation between DeploymentStatus and DeploymentStatusName).

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/ScenarioStatusProvider.scala
coutoPL: Good catch. Because this provider can be used in different places too, I've added a proper check if the user has access to the processing types.

But there is sth more important. We currently define limits in the context of processing types and globally. This is not a user context. So, there should be no user-related permission checks in the LimitsService. So, I've used `NussknackerInternalUser` instead of the currently logged user.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
coutoPL: Yes, but I assumed that users should be able to do sth with the limitation. 
I mean, if there is e.g., a restarting scenario (which exceeds the limit), they can cancel it and run the other scenario. 

In the use case you described, a user has no way to fix the problem. AFAIU, we detect it, but there is no exact cause for it. It may happen because of reasons. So, IMO, we should not "punish" users for that. 

That's why I'd not take it into consideration in case of limits.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/newdeployment/ScenarioStatusProvider.scala
arkadius: Good point with internal user

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/scenariostatus/ScenarioStatusProvider.scala
arkadius: In case of this problem, I think user can still cancel this scenario, the same as with restarting scenario. They also can redeploy it to show to designer that on Flink is the correct state. It is why in `ProblemStateStatus.defaultActions` are both `deploy` and `cancel` actions.

coutoPL: ok, it makes sense if users can cancel the scenario in the problem status. I modified the code to handle this case too.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/simple/SimpleStateStatus.scala
```diff
@@ -15,20 +15,37 @@ object SimpleStateStatus {
...
+  sealed trait ProblemStateStatus extends StateStatus {
     override def name: StatusName = ProblemStateStatus.name
+    def description: String
+    def allowedActions: Set[ScenarioActionName]
+    def tooltip: Option[String]
   }
 
   object ProblemStateStatus {
+
+    final case class ShouldNotBeRunning(
```
arkadius: We need the same for `ProblemStateStatus.multipleJobsRunning`

arkadius (approved):

PR #7982, opened 2025-04-23
[NU-2140] Enricher mocking

raphaelsolarski:
## Checklist before merge

### DIFF
--- a/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
@@ -15,12 +15,11 @@ import type { ExpressionObj } from "./types";
 import { EditorType } from "./types";
 
 const MOCKED_OUTPUT_IN_NODE_FIELD_NAME = "mockedOutput";
-const MOCK_EXPRESSION_PARAMETER_NAME = "mockExpression";
-const EXPRESSION_TEXT_PATH = `${MOCKED_OUTPUT_IN_NODE_FIELD_NAME}.${MOCK_EXPRESSION_PARAMETER_NAME}.expression`;
+const MOCK_EXPRESSION_PARAMETER_NAME = "$mockExpression";
+const EXPRESSION_TEXT_PATH = `${MOCKED_OUTPUT_IN_NODE_FIELD_NAME}.mockExpression.expression`;
 
 const MOCK_EXPRESSION_HINT_TEXT =
-    "If provided the expression will be evaluated in tests instead of invoking real enricher behaviour (e.g. instead of calling a service/database).\n\n" +
-    "Evaluated value will be placed under output variable name during tests.";
+    "If you provide this expression, the real service won't be invoked during tests. Instead, the result of the evaluation will be used.";
 
 const UnknownTypingResult = {
     params: [],
--- a/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
@@ -69,7 +69,7 @@ function MockedOutputField(props: Props): JSX.Element {
     );
 
     const renderMockExpressionParameterLabel = (): JSX.Element => {
-        return <FieldLabel title={MOCK_EXPRESSION_PARAMETER_NAME} label={t("nodes.enricher.mockExpression", "Mocked Output Expression")} />;
+        return <FieldLabel title={MOCK_EXPRESSION_PARAMETER_NAME} label={t("nodes.enricher.mockExpression", "Mock")} />;
     };
 
     const variableTypes = useMemo(() => findAvailableVariables(editedNode.id), [findAvailableVariables, editedNode.id]);
--- a/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
@@ -18,6 +18,10 @@ const MOCKED_OUTPUT_IN_NODE_FIELD_NAME = "mockedOutput";
 const MOCK_EXPRESSION_PARAMETER_NAME = "mockExpression";
 const EXPRESSION_TEXT_PATH = `${MOCKED_OUTPUT_IN_NODE_FIELD_NAME}.${MOCK_EXPRESSION_PARAMETER_NAME}.expression`;
 
+const MOCK_EXPRESSION_HINT_TEXT =
+    "If provided the expression will be evaluated in tests instead of invoking real enricher behaviour (e.g. instead of calling a service/database).\n\n" +
+    "Evaluated value will be placed under output variable name during tests.";
+
 const UnknownTypingResult = {
     params: [],
     type: "Unknown",
@@ -69,7 +73,7 @@ function MockedOutputField(props: Props): JSX.Element {
     );
 
     const renderMockExpressionParameterLabel = (): JSX.Element => {
-        return <FieldLabel title={MOCK_EXPRESSION_PARAMETER_NAME} label={t("nodes.enricher.mockExpression", "Mock")} />;
+        return <FieldLabel label={t("nodes.enricher.mockExpression", "Mock")} hintText={MOCK_EXPRESSION_HINT_TEXT} />;
     };
 
     const variableTypes = useMemo(() => findAvailableVariables(editedNode.id), [findAvailableVariables, editedNode.id]);
--- a/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
@@ -1,11 +1,10 @@
 import { get } from "lodash";
-import type { ReactNode } from "react";
-import { useMemo } from "react";
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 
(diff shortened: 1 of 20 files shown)


raphaelsolarski:
Screenshot 2025-05-07 at 11 13 28
Screenshot 2025-05-07 at 11 13 52
Screenshot 2025-05-07 at 11 14 35

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Dropdown should display menu portal #0.png
arkadius, line 1: Why it is changed if we hide this feature behind feature flag?

designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
```diff
@@ -0,0 +1,101 @@
...
+import type { OnValueChange } from "./Editor";
+import type { ExpressionObj } from "./types";
+import { EditorType } from "./types";
+
+const MOCKED_OUTPUT_IN_NODE_FIELD_NAME = "mockedOutput";
+const MOCK_EXPRESSION_PARAMETER_NAME = "mockExpression";
+const EXPRESSION_TEXT_PATH = `${MOCKED_OUTPUT_IN_NODE_FIELD_NAME}.${MOCK_EXPRESSION_PARAMETER_NAME}.expression`;
+
+const MOCK_EXPRESSION_HINT_TEXT =
+    "If provided the expression will be evaluated in tests instead of invoking real enricher behaviour (e.g. instead of calling a service/database).\n\n" +
```
arkadius: It sounds a little bit complex, maybe simpler:
"If you provide this expression, the real service won't be invoked during tests. Instead, the result of the evaluation will be used."

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -43,6 +48,8 @@ import shapeless.syntax.typeable._
 
 object NodeCompiler {
 
+  val MockExpressionParameterName: ParameterName = ParameterName("mockExpression")
```
arkadius: Maybe let's use `$mockExpression` to avoid some collision

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/InterpreterSpec.scala
```diff
@@ -1019,6 +1022,28 @@ class InterpreterSpec extends AnyFunSuite with Matchers {
     result shouldBe Collections.singletonMap("bool", false)
   }
 
+  test("invokes enricher with mocked output in live mode") {
```
arkadius: Hmm, I have two problems with this:
1. By reading the test case title I don't know what do you expect. The real enriched should be invoked or no?
2. I see that the real enricher wasn't invoked. Instead the mocked value was returned. We wanted to mock enrichers only in tests :) Maybe I misunderstand something?

scenario-api/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
```diff
@@ -208,6 +209,15 @@ object node {
     override def parameters: List[NodeParameter] = service.parameters
   }
 
+  sealed trait EnricherMockedOutput
```
arkadius: When we talked about that I rather thought about something like:
```
EnricherTestingConfiguration(componentBehaviour: EnricherTestingBehaviour) 

sealed trait EnricherTestingBehaviour
object EnricherTestingBehaviour {
  case class ReturnMockedValue(outputValueExpression: Expression) extends EnricherTestingBehaviour
  case object InvokeRealService extends EnricherTestingBehaviour
}
```
and having `InvokeRealService` as a default.

Thanks to `EnricherTestingBehaviour` envelope, we could add something more next to `componentBehaviour`.

But after we talk about further changes around testcases, I think that we should go into totally different, YAGNI approach and not invent this structure at all, because it shows that we have to be prepared for many evolutions of this structure. 

BTW, I now see that this architectural decision about keeping test cases data in scenario graph will cause many scenario migrations which will be shown in the scenario history.

LSS: Let's remove this sealed trait.

designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const renderMockExpressionParameterLabel = (): JSX.Element => {
+        return <FieldLabel label={t("nodes.enricher.mockExpression", "Mock")} hintText={MOCK_EXPRESSION_HINT_TEXT} />;
+    };
+
+    const variableTypes = useMemo(() => findAvailableVariables(editedNode.id), [findAvailableVariables, editedNode.id]);
+    const testResultsState = useTestResults();
+
+    return (
+        <ExpressionTestResults fieldName={MOCK_EXPRESSION_PARAMETER_NAME} resultsToShow={testResultsState.testResultsToShow}>
+            <EditableEditor
```
arkadius: Currently, we are in the middle in the migration to `Json template | Expression` combination of editors in almost all places. So far, we did it conservatively, only in places where at the end, json is processed somehow, but there is the plan to use this approach almost everywhere. Let's use this combination here as well.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
```diff
@@ -381,8 +393,31 @@ class NodeCompiler(
   def compileEnricher(n: Enricher, ctx: ValidationContext, outputVar: OutputVar)(
       implicit nodeId: NodeId,
       scenarioCompilationDependencies: ScenarioCompilationDependencies
-  ): NodeCompilationResult[compiledgraph.service.ServiceRef] = {
-    compileService(n.service, ctx, Some(outputVar))
+  ): NodeCompilationResult[EnricherCompilationResult] = {
+    val serviceCompilationResult = compileService(n.service, ctx, Some(outputVar))
```
arkadius, line 403: I think that we should have somethink like
```
(runtimeMode, mockedOutput) match {
  case (Test, Some(...) => compileMockExpression
  case _ => compileService 
}
```
See next comments

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/InterpreterSpec.scala
raphaelsolarski: 1. I fixed tests descriptions a little bit.
2. The "1111" is returned so it is from real enricher behaviour (real behaviour is returning param1 value as output). Mocked output value is "2222". I added named parameters to make it more visible.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
raphaelsolarski: fixed

designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
raphaelsolarski: fixed

designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
raphaelsolarski: fixed. 
With jsontemplate there were a few issues:
* it didn't allowed null values if expected typing wasn't directly set to `TypedNull` - I fixed it to allow null always despite of typing.
* it jsontempate is not checked against expected typing. It seems that in actual usages there was always `Unknown/Json` type set as expected type (json schemaless kafka source, http enricher body, event generator). I added additional post compile type check for mock expression. We can discuss whether it should be moved to JsonTemplateParser.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeCompiler.scala
raphaelsolarski: As discussed we have to compileService to learn the output typing to check mock typing against. 
Service and mock expression is compiled always and decision whether mock should be used is actually delegated to `Interpreter` invocation.
In future PRs we can change mock related compilation errors to be treated as warns if we want (like warns about disabled filter nodes).

Side note: actually, there is a side effect that open/close methods of service are called even when mock expression is used instead. IMO we should improve it, but we can split it to a separate PR.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/graph/node.scala
raphaelsolarski: fixed, trait removed

designer/client/src/components/graph/node-modal/editors/expression/MockedOutputField.tsx
arkadius: Why not to move this validation to json template parser now? It would be consistent with spel parser.

raphaelsolarski: moved

arkadius (approved):

designer/client/cypress/e2e/__image_snapshots__/electron/Linux/Dropdown should display menu portal #0.png
raphaelsolarski: fixed

PR #8004, opened 2025-04-28
[NU-2149] End scenario without sink

mgoworko:
## Describe your changes

Scenarios no longer have to end with final sink node. Other node may be the final node of the scenario. The change is optional, by default no changes in Nu behavior.

## Checklist before merge


extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ModelSettings.scala
```diff
@@ -0,0 +1,13 @@
+package pl.touk.nussknacker.engine.api.process
+
+/**
+  * General settings for model
+  * @param allowEndingScenarioWithoutSink - indicates, whether it is allowed to have nodes other than sinks as final nodes of the scenario
+  */
+final case class ModelSettings(
```
arkadius: We use `*Config` suffix for such things - see `DesignerConfig`

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ProcessConfigCreator.scala
```diff
@@ -32,4 +34,12 @@ trait ProcessConfigCreator extends Serializable {
   def classExtractionSettings(modelDependencies: ProcessObjectDependencies): ClassExtractionSettings =
     ClassExtractionSettings.Default
 
+  def modelSettings(modelDependencies: ProcessObjectDependencies): ModelSettings = {
```
arkadius: Let's do this differently, let's replace `ProcessObjectDependencies.config` with:
`val modelConfig: ModelConfig`, let's keep `rawConfig: Config` field inside `ModelConfig` with TODO that w should parse config to normal, class fields.

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ProcessConfigCreator.scala
```diff
@@ -32,4 +34,12 @@ trait ProcessConfigCreator extends Serializable {
...
+    Try(
```
arkadius: 1. Catching all exceptions without any logging is never a good option :) Let's use Ficus `.getAs[Boolean]` which returns `Option[Boolean]` - see how it is done in `DesignerConfig`
2. Let's make configuration parsing logic eager; it should be a field of `ProcessObjectDependencies` instead of method. Eventually we should create it during creation of `ProcessingTypeConfig.modelConfig`

scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonize/MaybeArtificial.scala
```diff
@@ -45,15 +45,24 @@ private[engine] object MaybeArtificial {
...
       errors.toList
     )
 
   def artificialSource(errors: ProcessUncanonizationError*): MaybeArtificial[node.SourceNode] =
-    artificialSink(errors: _*).map(
+    missingSinkError(errors: _*).map(
       node.SourceNode(node.Source(generateArtificialName(), SourceRef(artificalSourceSinkRef, Nil)), _)
     )
 
+  def addedArtificialDeadEndSink(previousNodeId: String): MaybeArtificial[node.SubsequentNode] =
```
arkadius: Why do you have to create some artifical sink? For filter's false we never created any artificial dead-end and everything was fine

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/InterpreterSpec.scala
```diff
@@ -196,7 +196,8 @@ class InterpreterSpec extends AnyFunSuite with Matchers {
           ComponentDefinitionExtractionMode.FinalDefinition
         ),
       ModelDefinitionBuilder.emptyExpressionConfig,
-      ClassExtractionSettings.Default
```
arkadius, line 199: Where are tests showing how our runtime behave when there is no sink? Does our scenario builder is ready for such construction of scenario?

mgoworko: There are multiple tests in `ResultCollectingListenerSpec` (in fact, 70% of this PR are tests :) ). They verify, that the scenario without sinks is correctly compiled and executed, and additionally, that the test results are generated.

There is exactly the same difficulty as for testing the more complex scenario structures (splits, choices, fragments etc.). They have to be tested in a specific context - on some engine, with set of components. Otherwise there are too many things that are mocked/artificially created, like in InterpreterSpec.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonize/MaybeArtificial.scala
mgoworko: It was my initial approach - without node, but I abandoned it as too problematic because:
- we want to still require sinks, unless we configure the flag, that disables this requirement
- so sometimes we want to require sinks, and sometimes not; as a result the logic and implementation cannot be fully changed to not require sinks
- there are things in Nu, that depend on the scenario having sink - validations, edge cases in the compiler etc.
- so the simplest, and fully working, solution is to create some noOp sink for execution
- it is invisible and does not perform any operations, it is very similar case to BranchEndDefinition or some other artificial nodes, that are created in the underlying scenario representation for technical reasons

The case with filter "false" branch is different - the "true" branch is required. So it is not the case of ending the scenario.

arkadius: If you want to fade out a feature, the code should make it look as if the feature never existed. If you want to still keep backward compatibility for some period of time, the code keeping this compatibility should be encapsulated and it should be as easy as possible to remove it. It is our case - we want to remove this validation at all eventually. So let's remove the restriction everywhere where it exists, and after that let's add a little class providing backward compatibility, checking what is the type of ending node and returning the previous error.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/InterpreterSpec.scala
arkadius: Hmm interesting, I haven't seen it, now I see it. The approach with the specific engine is fine. The only thing that I'm worry about is that the name of test suggest that you are testing the listener and in fact you wanted to test scenario compilation process. We have `FlinkTestScenarioRunner`, which we use for different scenario combinations tests. Can you reuse it?

mgoworko: I refactored it - there are now 2 separate test files. The new on e contains only tests for scenarios without sinks. There are also assertions of compilation errors (in situations, when ending with sink is required). I use the flinkMiniCluster, but not the  FlinkTestScenarioRunner - it is much more complex (which is not required here) but does not allow the customization (custom config substitutions, multiple custom sources) that is required here.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonize/MaybeArtificial.scala
mgoworko: I created `trait MissingSinkHandler`, with 2 implementations - for the old behavior of returning error, and the new behavior of returning None as the "next" node. It is used as implicit in the ProcessCanonizer

extensions-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ProcessConfigCreator.scala
arkadius: Now it looks nice. Can you use this new class in `ModelData.modelConfig`? This is the first place where model config appear. Thanks to that we will have only one representation of this config.

scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonize/MaybeArtificial.scala
arkadius: Nice

arkadius (approved):
Nice, one minor comment added

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ProcessObjectDependencies.scala
```diff
@@ -1,15 +1,39 @@
...
   }
 
+  def parseModelConfig(modelConfig: Config): ModelConfig = {
+    ModelConfig(
+      allowEndingScenarioWithoutSink = modelConfig.getOrElse[Boolean]("allowEndingScenarioWithoutSink", false),
+      underlyingConfig = modelConfig,
+    )
+  }
+
+  final case class ModelConfig(
```
arkadius: Let's move this class to the same package as `BaseModelData`, to a separate file

components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ProcessObjectDependencies.scala
```diff
@@ -1,15 +1,39 @@
...
+final case class ProcessObjectDependencies private (modelConfig: ModelConfig, namingStrategy: NamingStrategy)
+    extends Serializable {
+  def config: Config = modelConfig.underlyingConfig
+}
 
 object ProcessObjectDependencies {
 
-  def withConfig(modelConfig: Config): ProcessObjectDependencies = {
-    ProcessObjectDependencies(modelConfig, NamingStrategy.fromConfig(modelConfig))
+  def apply(underlyingConfig: Config, namingStrategy: NamingStrategy): ProcessObjectDependencies = {
```
arkadius, line 15: Let's use `ModelConfig` here and in `withConfig` method. Generally, let replace  `Config` with `ModelConfig` everywhere where it is easy. Here it is.

PR #8005, opened 2025-04-28
[NU-1948] Fix projection, selection and #Collection.merge on table api org.apache.flink.types.Row class

paw787878:
## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/FlinkRowToMapConverter.scala
+++ b/scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/FlinkRowToMapConverter.scala
@@ -31,13 +31,11 @@ class FlinkRowToMapConverter(val conversionService: ConversionService) extends C
 //            val fields = r.getFieldNames(true).asScala
 //            fields.map(e => e -> r.getField(e)).toMap.asJava
       val getFieldsMethod = source.getClass.getMethod("getFieldNames", classOf[Boolean])
-      getFieldsMethod.setAccessible(true)
       val fields =
         getFieldsMethod.invoke(source, (true).asInstanceOf[AnyRef]).asInstanceOf[java.util.Set[String]].asScala
       fields
         .map(e => {
           val getFieldMethod = source.getClass.getMethod("getField", classOf[String])
-          getFieldMethod.setAccessible(true)
           e -> getFieldMethod.invoke(source, e)
         })
         .toMap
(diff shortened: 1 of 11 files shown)


scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/FlinkRowToMapConverter.scala
```diff
@@ -0,0 +1,48 @@
...
+  override def convert(source: Any, sourceType: TypeDescriptor, targetType: TypeDescriptor): AnyRef = {
+    if (source == null) {
+      null
+    } else {
+      // if we could import Row class this would look like this:
+//            val r      = source.asInstanceOf[org.apache.flink.types.Row]
+//            val fields = r.getFieldNames(true).asScala
+//            fields.map(e => e -> r.getField(e)).toMap.asJava
+      val getFieldsMethod = source.getClass.getMethod("getFieldNames", classOf[Boolean])
+      getFieldsMethod.setAccessible(true)
```
arkadius: Why is it needed? I see that this method is `public`

scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/FlinkRowToMapConverter.scala
```diff
@@ -0,0 +1,48 @@
...
+      val fields =
+        getFieldsMethod.invoke(source, (true).asInstanceOf[AnyRef]).asInstanceOf[java.util.Set[String]].asScala
+      fields
```
arkadius: Let's do `.toList` before `.map`. AFAIK `.map` rewrite the result to the same structure as it is invoked on, so in this situation it will do:
```
set = new HashSet
set.add(entry1)
set.add(entry2)
```
Adding to a set involves computation for each entry, including hash computation of potentially deep structures in values. We don't need it, because we don't want to have unique set of entries, we only want to have unique set of keys and values for them.

scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/internal/IndexedRecordToMapConverter.scala
```diff
@@ -37,15 +37,17 @@ class IndexedRecordToMapConverter(val conversionService: ConversionService) exte
...
-        val name = invokeReflectiveMethodWithNoParameters(n, "name").asInstanceOf[String]
-        val pos = invokeReflectiveMethodWithNoParameters(n, "pos").asInstanceOf[Int]
-
-        val getMethod = source.getClass.getDeclaredMethod("get", classOf[Int])
-        getMethod.setAccessible(true)
-        val value = getMethod.invoke(source, pos.asInstanceOf[AnyRef])
-        name -> value
-      }).toMap.asJava
+      fields.asScala
+        .map(n => {
```
arkadius: One nitpick: `(foo => {})` in scala can be shortened to `{ foo => }` and we try to keep it used consistently in the code base.

arkadius (approved):
One comment, after resolving it you can merge the change

PR #8019, opened 2025-04-30
[NU-2152] Missing parameters in node are only reported as warnings in logs - no validation error is returned

arkadius:
## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeParametersAdjuster.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeParametersAdjuster.scala
@@ -7,6 +7,7 @@ import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.engine.api.{JobData, NodeId}
 import pl.touk.nussknacker.engine.api.definition.{Parameter => ParameterDefinition}
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
+import pl.touk.nussknacker.engine.compile.CompilationLoggerExtensions.Ops
 import pl.touk.nussknacker.engine.graph.evaluatedparam.{Parameter => NodeParameter}
 import pl.touk.nussknacker.engine.util.Implicits.RichTupleList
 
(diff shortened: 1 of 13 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
```diff
@@ -95,7 +49,7 @@ object Validations extends LazyLogging {
         }
       }
       .sequence
-      .map(_ => parameter)
+      .map(_ => ())
```
mslabek: can also use `.void`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeParametersAdjuster.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  private def addMissingNodeParameters(
+      parameterDefinitions: List[ParameterDefinition],
+      parameters: List[NodeParameter]
+  )(
+      implicit nodeId: NodeId,
+      jobData: JobData
+  ) = {
+    val parameterByName = parameters.map(p => p.name -> p).toMapCheckingDuplicates
+    val (missingParameterNames, adjustedParameters) = parameterDefinitions
```
mslabek: I think we can simplify this by:
1. using simpler Writer or even just `foldLeft` 
2. `traverse` instead of `map ... sequence` 

Like
 ```scala
 val (missingParameterNames, adjustedParameters) =
      parameterDefinitions.traverse { param =>
        parameterByName.get(param.name) match {
          case Some(param) => Writer.value[Set[ParameterName], NodeParameter](param)
          case None =>
            val defaultParam = NodeParameter(param.name, param.finalDefaultValue)
            Writer(Set(param.name), defaultParam)
        }
}.run
```

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/Validations.scala
arkadius: I forgot about that, thx

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeParametersAdjuster.scala
arkadius: I forgot about `Traverse`, thanks. Good tip with `traverse` as well. The only thing that I've not applied is using pattern matching on `Option`. Instead, it is `.map.getOrElse` as it was. Is it fine for you? For me, `.map.getOrElse` approach is a little bit more concise and readable.

mslabek (approved):

PR #8042, opened 2025-05-06
[NU-2011] Merge OpenAPI components into one

lukasz-bigorajski:
## Checklist before merge


components/openapi/src/main/scala/pl/touk/nussknacker/openapi/OpenAPIComponentProvider.scala
```diff
@@ -1,69 +1,29 @@
...
-      } catch {
-        case NonFatal(ex) =>
-          logger.error("OpenAPI service resolution failed. Will be used empty services lists", ex)
-          List.empty
-      }
-    val servicesToUse = services.collect { case Valid(service) =>
-      ConfigFactory.parseString(service.asJson.spaces2).root()
-    }
-    logErrors(services)
-    config.withValue("services", ConfigValueFactory.fromIterable(servicesToUse.asJava))
```
jedrz, line 40: This trick allows OpenAPI services to be passed to a Flink job via configuration during deployment. As a result, the deployed Flink job doesn't need to rediscover the OpenAPI services - their definitions are simply recreated from the provided configuration. Perhaps this behavior should be retained as it was?

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/OpenApiDefinitionDiscovery.scala
```diff
@@ -49,5 +71,20 @@ class SwaggerOpenApiDefinitionDiscovery(implicit val httpBackend: SttpBackend[Fu
...
   }
+}
 
+class CachingOpenApiDefinitionDiscovery(
+    discovery: OpenApiDefinitionDiscovery,
+    openAPIsConfig: OpenAPIServicesConfig,
+) extends OpenApiDefinitionDiscovery {
+  @transient private lazy val servicesCache = new SingleValueCache[List[Validated[ServiceParseError, SwaggerService]]](
+    expireAfterAccess = None,
+    expireAfterWrite = Some(openAPIsConfig.openApiServicesDiscoveryCacheTtl)
```
jedrz, line 82: Do we need to expire fetched OpenAPI services at all? Currently, the only way to fetch them again is by reloading the model. Maybe we should keep this behaviour unchanged?

components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenApiScenarioIntegrationTest.scala
```diff
@@ -125,8 +136,11 @@ class OpenApiScenarioIntegrationTest
     }
   ) { testScenarioRunner =>
     // given
-    val data     = List("10")
-    val scenario = scenarioWithEnricher((SingleBodyParameter.name, "#input".spel))
+    val data = List("10")
+    val scenario = scenarioWithEnricher(
```
jedrz, line 140: Shouldn't the `openAPI` enricher be used in `scenarioWithEnricher` instead of `customer`?

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/OpenApiDefinitionDiscovery.scala
jedrz: On the other hand, the cache TTL can be configured using `openApiServicesDiscoveryCacheTtl`, so I think it can be left as it is 🙂

lukasz-bigorajski: > Do we need to expire fetched OpenAPI services at all? Currently, the only way to fetch them again is by reloading the model. Maybe we should keep this behaviour unchanged?

We agreed to change the behaviour because the problem is when nussknacker is starting and openapi service is currently unavailable.

> I think it can be left as it is
Do you mean to left reading components at nu startup?

Apart I've written upper, other components like ml component read theirs configuration after nu startup, why this component should not be concise with the other?

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/OpenAPIComponentProvider.scala
lukasz-bigorajski: Is not enough to pass chosen swaggerService to implementation method and return enricher in this dynamic component?

components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenApiScenarioIntegrationTest.scala
lukasz-bigorajski: I don't understand you question :/. This test is using given openApi service:
```
// it should "call enricher returning string" in withPrimitiveReturnType -> withPrimitiveReturnType -> new StubService("/customer-primitive-return-swagger.yaml") -> "/customer-primitive-return-swagger.yaml" (yaml below)
openapi: "3.1.0"
info:
  title: Simple API overview
  version: 2.0.0
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /customer:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: string
      responses:
        '200':
          description:
            Returns ComponentsByAccount.
          content:
            application/json:
              schema:
                type: string
```
So this definition allows only one swagger service which I need to pass in order to dynamic component knows which service to choose.

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/discovery/OpenApiDefinitionDiscovery.scala
jedrz: That's good point, new behaviour will be consistent with ML components.

components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenApiScenarioIntegrationTest.scala
jedrz: I see it now. At [191](https://github.com/TouK/nussknacker/pull/8042/files#diff-c5c27ae7dbc523eab6457f53b5c660848db1190f7a1134ca9e85b9afa62a667fR191) line the new Open API component is named `getCustomer`. My suggestion was to rename it to `openAPI` to be consistent with `OpenAPIComponentProvider` but it is up to you.

jedrz (approved):
LGTM

PR #8047, opened 2025-05-07
Live data preview and node transition throughput on Flink minicluster

mgoworko:
## Describe your changes

- added listener, that collects live data processed by the scenario
- the data is stored in new RingBuffer util based on LinkedHashMap with automatic cleaning of old data, samples corresponding to the last N contextIds are stored 
- also added node transition frequency calculation - node transition counts are stored in buckets for each epoch second and calculated in time windows

The data can be fetched using `/api/liveData/{scenario_name}`

At the moment only for Flink minicluster, because it is a part of the same JVM and we just access the data without synchronizing it with external data storage.

## Checklist before merge


designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -89,3 +94,24 @@ trait SchedulingSupported extends SchedulingSupport {
...
+      processIdWithName: ProcessIdWithName,
+  ): Future[Option[LiveDataPreview]]
+
+}
+
+object LiveDataPreviewSupported {
+
+  final case class LiveDataPreview(
+      liveDataSamples: TestResults[Json],
+      nodeTransitionFrequency: Map[NodeTransition, BigDecimal],
```
arkadius: I think that using "throughput" word is more natural in this context. The "frequency" is used when we talk about some cyclic mechanism to show how often it is triggered.

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
```diff
@@ -89,3 +94,24 @@ trait SchedulingSupported extends SchedulingSupport {
...
 
 case object NoSchedulingSupport extends SchedulingSupport
+
+sealed trait LiveDataPreviewSupport
+
+trait LiveDataPreviewSupported extends LiveDataPreviewSupport {
+
+  def getLiveData(
+      processIdWithName: ProcessIdWithName,
+  ): Future[Option[LiveDataPreview]]
```
arkadius: This `Option` is mysterious. It could be interpreted in many ways. Let's use `Either` with explicit error instead. For deployed scenarios, we should always return defined, empty `LiveDataPreview`. The error should be returned when someone ask for scenario that is not deployed or is finished/stopped.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -27,18 +27,23 @@ object Context {
...
 /**
  * Context is container for variables used in expression evaluation
  *
+ * @param initialId     the initial id of the context (id may change during the processing, but the initial id is preserved in this field
  * @param id            correlation id/trace id used for tracing (logs, error presentation) and for tests mechanism, it should be always defined
  * @param variables     variables available in evaluation
  * @param parentContext context used for scopes handling, mainly for fragment invocation purpose
  */
 case class Context(
+    initialId: String,
```
mgoworko: This change will no longer be needed after the ContextId refactor. The initialId is basically ContextId without prefix.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -27,18 +27,23 @@ object Context {
...
+    Context(id, id, variables, parentContext)
+
 }
 
 case class ContextId(value: String)
 
 /**
  * Context is container for variables used in expression evaluation
  *
+ * @param initialId     the initial id of the context (id may change during the processing, but the initial id is preserved in this field
```
arkadius: Can you add one more sentence in the comment, why do we need this field with preserved id? Also, maybe we could replace case class with normal class to make changes of `initialId` statically impossible?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ScenarioTestingApiEndpoints.scala
```diff
@@ -192,11 +185,24 @@ class ScenarioTestingApiEndpoints(auth: EndpointInput[AuthCredentials]) extends
...
+      .withSecurity(auth)
+
+  def scenarioLiveDataEndpoint: SecuredEndpoint[
+    (ProcessName, Option[SkipResultsPerNode], Option[SkipResultsPerTransition]),
+    TestingError,
+    ResultsWithCountsDto,
+    Any
+  ] =
+    baseNuApiEndpoint
+      .summary("Perform test")
```
arkadius: What test this endpoint perform? :P

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ScenarioTestingApiEndpoints.scala
```diff
@@ -192,11 +185,24 @@ class ScenarioTestingApiEndpoints(auth: EndpointInput[AuthCredentials]) extends
...
+      .tag("Testing")
+      .get
+      .in("scenarioTesting" / path[ProcessName]("scenarioName") / "liveData")
```
arkadius: Why endpoint is nested in `scenarioTesing` resource? Let's create a new `liveData` endpoint. I know that under the hood we reuse some features from the scenario testing mechanism, but we don't need to expose it in the interface.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
arkadius: So let's add a TODO/FIXME

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerHolder.scala
```diff
@@ -0,0 +1,47 @@
...
+    // so when we start a new job, we discard all old data
+    cleanResults(processName)
+    new LiveDataCollectingListener(processName, maxNumberOfSamples, frequencyWindowInSeconds)
+  }
+
+  def getLiveDataPreview(processName: ProcessName): Option[LiveDataPreview] = {
+    Option(listenerStorages.getIfPresent(processName.value)).map(_.getLiveDataPreview)
+  }
+
+  private[livedata] def cleanResults(processName: ProcessName): Unit = {
```
arkadius: I assume that you don't clean results after the job finish, because you want to make it still available for the user for some period of time? Maybe let's write a comment about that because from the first glance it looks like some memory leak.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerStorage.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  private def compute(key: K, remappingFunction: (K, Option[V]) => V): Unit = synchronized {
+    val currentValue = Option(results.get(key))
+    val newValue     = remappingFunction(key, currentValue)
+    val isNewKey     = !results.containsKey(key)
+    if (isNewKey) orderOfResults.add(key)
+    results.put(key, newValue)
+    while (orderOfResults.size() > maxNumberOfSamples) {
```
arkadius: `size` has O(n) complexity. Have you considered using the ready solution such as `EvictingQueue` from guave? we do synchronisation, so I can't see too much added value from using thread-safe structures. Keeping two information in two data structures: `results` and `orderOfResults` is hard to analyze in code, e.g. as a reviewer I can't guratanee that it is correctly cleaned.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerStorage.scala
```diff
@@ -0,0 +1,94 @@
...
+    maxNumberOfSamples: Int,
+    frequencyWindowInSeconds: Int,
+) {
+
+  private type K = String
+  private type V = TestResults[Json]
+
+  private val results                             = new ConcurrentHashMap[K, V]()
+  private val orderOfResults                      = new ConcurrentLinkedQueue[K]()
+  private val transitionsByEpochSecond            = new ConcurrentHashMap[Long, ConcurrentLinkedQueue[NodeTransition]]()
```
arkadius: Maybe we could just use a `MultiMap` to make it simpler? Or a Map `NodeTransition` ->`InstantRateMeter`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
mgoworko: I removed the `initialId` field. Instead of it, the initial context id is now recognized by regex. There is a todo added about it. After the ContextId refactor the regex won't be necessary, because we will be able co compare parts of the ContextId case class instead.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
mgoworko: I removed this field, details in the other comment.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerStorage.scala
mgoworko: Unfortunately there is no ready-to-use util to do exactly what is needed here. I asked LLM's about it too :) . (The main difficulty is that we have to drop the transitions, that are older than the treshold)

I created an util `class SlidingWindowCounter`. It wraps the logic around it and I think it is quite simple and easy to understand. I also added unit tests of that util.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerStorage.scala
mgoworko: Unfortunately guava `EvictingQueue` won't work here. It does not support modifying elements by key. But my solution was not good either.

I changed the solution. I created `RingBuffer` util, with unit tests. It is based on the LinkedHashMap, with added synchronization.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/livedata/LiveDataCollectingListenerHolder.scala
mgoworko: done, comment is now above this method and describes when we want to clean the data

designer/deployment-manager-api/src/main/scala/pl/touk/nussknacker/engine/api/deployment/DeploymentManager.scala
mgoworko: Done, either with error in Left. The error only says that the information is not available for the scenario. In that place we do not have detailed information as to why.

build.sbt
```diff
@@ -2023,6 +2024,20 @@ lazy val deploymentManagerApi = (project in file("designer/deployment-manager-ap
   )
   .dependsOn(extensionsApi, testUtils % Test)
 
+lazy val liveDataCollector = (project in file("designer/live-data-collector"))
```
mgoworko, line 2027: I created a new module, reusable by other DMs, instead of adding it in the Flink engine module.

docs/Changelog.md
```diff
@@ -177,6 +177,17 @@ description: Stay informed with detailed changelogs covering new features, impro
 * [#8042](https://github.com/TouK/nussknacker/pull/8042) Merge OpenAPI components into one with multiple services.
 * [7684](https://github.com/TouK/nussknacker/pull/7684) Add action redeploy. Till now action deploy was responsible for both deploy and redeploy operations. 
   Now they are separated in scenario workflow and UI. Kafka source has different deployment parameters for deploy and redeploy actions.
+* [#8047](https://github.com/TouK/nussknacker/pull/8047) Added functionality of collecting live data samples and node transition frequencies
```
arkadius: frequency left here in the PR title

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ScenarioTestingApiEndpoints.scala
```diff
@@ -189,6 +190,23 @@ class ScenarioTestingApiEndpoints(auth: EndpointInput[AuthCredentials]) extends
       .errorOut(testingErrorOutput)
       .withSecurity(auth)
 
+  def scenarioLiveDataEndpoint: SecuredEndpoint[
```
arkadius: Let's do one more step towards splitting live data preview feature and scenario testing features and extract separate classes for both `Endpoint` and `HttpService`. The structure of this classes should be the same as structure of endpoints.

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    buckets.asScala.values
+      .flatMap(_.asScala)
+      .toList
+      .groupBy(identity)
+      .view
+      .map { case (key, value) => (key, value.size) }
+      .toMap
+      .map { case (transition, count) =>
+        transition -> BigDecimal(count)./(samplingInterval).setScale(4, BigDecimal.RoundingMode.HALF_UP)
```
arkadius: 1. `./` looks strange. Let's use:
```
(BigDecimal(count) / samplingInterval).setScale(...)
```
instead
2. Maybe `half_even` instead?

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
```diff
@@ -0,0 +1,60 @@
...
+        .keySet()
+        .asScala
+        .filter(_ < cutoff(now))
+        .foreach(buckets.remove)
+    }
+  }
+
+  private def cutoff(now: Long): Long = now - windowSizeSeconds + 1
+
+  private def now(): Long = Instant.now(clock).getEpochSecond
```
arkadius: clock has `.instant()`

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
```diff
@@ -0,0 +1,60 @@
...
+      .map { case (transition, count) =>
+        transition -> BigDecimal(count)./(samplingInterval).setScale(4, BigDecimal.RoundingMode.HALF_UP)
+      }
+  }
+
+  private val lastCleaned = new AtomicLong(0)
+
+  private def cleanOldBuckets(now: Long): Unit = synchronized {
+    // Clean old buckets at most once per second, not on each call
+    if (lastCleaned.getAndSet(now) != now) {
```
arkadius, line 47: For `windowSizeSeconds` around hours or day it will slow down events processing due to O(n) complexity. For similar problem (sliding window aggregations we used `SortedMap` which has O(log n) complexity, see: `AggregatorFunction` and `FlinkRangeMap`. WDYT about this approach? Should we apply it here as well?

mgoworko: I changed `ConcurrentSkipListMap` to `ConcurrentSkipListMap`, which according to Java docs provides log(n) time cost for the get, put and remove.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ScenarioTestingApiEndpoints.scala
mgoworko: Done, scenarioTesting and liveData APIs are splitted

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
arkadius: The comment is attached to `cleanOldBuckets` where you filter buckets by date. Take a look at `SortedMap.rangeTo` characteristics.

mgoworko: Ok, so:
- I already changed the map implementation to ConcurrentSkipListMap
- now I also changed the cleaning implementation to use `.headMap(cutoff)`, which has `O(log n)` complexity
- this cleaning is invoked at most once per second - there is a condition that prevents more than one cleanup per second, irrespective of number of events per second `if (lastCleaned.getAndSet(now) != now)`

arkadius (approved):

PR #8077, opened 2025-05-12
[NU-2161] Added additional information in Parameter definition: changesCanReloadParameters

arkadius:
…changes in parameter value can reload parameters definitions (changesCanReloadParameters)

## Checklist before merge

### DIFF
--- a/common-api/src/main/scala/pl/touk/nussknacker/engine/util/Implicits.scala
+++ b/common-api/src/main/scala/pl/touk/nussknacker/engine/util/Implicits.scala
@@ -88,6 +88,13 @@ object Implicits {
       case _   => list
     }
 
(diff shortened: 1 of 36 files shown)


jedrz (commented):
Shouldn't the `topic` parameter in `KafkaUniversalComponentTransformer` be extended with `changesCanReloadParameters = true`?

arkadius:
> Shouldn't the `topic` parameter in `KafkaUniversalComponentTransformer` be extended with `changesCanReloadParameters = true`?

It is not needed because the mechanism automatically set `changesCanReloadParameters = true` for every last parameter in `NextParameters.parameters` list. So it will be set for both topic and schema version

jedrz (approved):

PR #8103, opened 2025-05-16
fix cancel problems when cancelling scenario in restarting state

Diamekod0221:
## Describe your changes
Currently if a periodic (others unchecked) scenario is in state restarting, clicking cancel by the user makes the action succeed in nussknacker. But on flink nothing happens. 
This is caised by a missing state check in `PeriodicProcessService.synchronizeDeploymentsStates` - it silently succeeds on restarting, without actually doing anything. That passess an empty list to the cancel command which succeeds. 

I fix the issue by adding an extra case to the `synchronizeDeploymentStates` method.

 Perhaps some other guard could be investigated to avoid silently failing on other statuses or after logic reuse. 

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessService.scala
@@ -296,7 +296,7 @@ class PeriodicProcessService(
         case (_, _, deployment, Some(status))
             if SimpleStateStatus.DefaultFollowingDeployStatuses.contains(status.status) =>
           deployment.id
-        // Without restarting, we silently succeed the restarting scenario cancel without sending anything to flink
+        // Without considering the `restarting` scenario status, we silently succeed cancelling of the scenario in Restarting status, without sending anything to Flink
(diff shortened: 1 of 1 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessService.scala
```diff
@@ -296,6 +296,9 @@ class PeriodicProcessService(
         case (_, _, deployment, Some(status))
             if SimpleStateStatus.DefaultFollowingDeployStatuses.contains(status.status) =>
           deployment.id
+        // Without restarting, we silently succeed the restarting scenario cancel without sending anything to flink
```
mgoworko: Maybe the comment could be a little bit more clear about the "restarting". Current version can be confusing, because "restarting" can be interpreted as "triggering/performing scenario restart", but it means "scenario in Restarting status" in this context.

mgoworko (approved):
One minor comment

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicProcessService.scala
Diamekod0221: changed

PR #8119, opened 2025-05-20
ContextId refactor

mgoworko:
## Describe your changes

Changes:
- separated parts of the ContextId structure - it is no longer a simple string, but instead case class that contains a few fields representing scenario id, source node id, index etc.
- it does not fundamentally change the ContextId representation - after serialization to string is looks the same
- with one significant modification - the Split node now adds suffix to context ids of each of its outgoing branches; Split essentially clones the event and sends copy to each of its branches, those outgoing contexts need to be identifiable
- the new ContextId structure is returned in test results data structures 

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -1,35 +1,39 @@
...
-  def apply(id: String, variables: Map[String, Any]): Context =
+  def apply(id: ContextId, variables: Map[String, Any]): Context =
     Context(id, variables, None)
 
+  def dummy: Context = Context(ContextId.dummy)
+
 }
 
-case class ContextId(value: String)
+case class ContextId(
```
mgoworko: This refactor separates constituent parts of the ContextId, but essentially it does not change this structure. Should we go in any particular direction? J suggested adding some displayableName.

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/RowConversions.scala
```diff
@@ -65,4 +67,37 @@ object RowConversions {
 
   }
 
+  private implicit def contextIdCodec: Codec[ContextId] =
```
mgoworko, line 73: I use the circe codec here, and keep the contextId in the single String field, as before. The reason is that we may need to further modify ContextId representation, without making multiple (very) problematic changes in the Flink serialization. But at the same time I am not sure if we are supposed to use circe in this context - whether it will have impact on performance.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
arkadius: I think that we shouldn't strictly assume restriction that `ContextId` should be exactly the same as before change. Instead, we should think what parts where necessary to provide the current identity behaviour and extract them. We should avoid magical parts:
- `prefix`: did you find some places where it was something different than scenarioName? If, so, maybe we have two separate parts: scenarioName + optional sth else
- `nodeId`: let's make it even more explicit. `nodeId` was added to easily provide identity between context generated by different nodes. Let's call it e.g. `originatingNodeId`
- `suffix`: I assume that it was used in places where new context was generated and neither taskId nor node id was sufficient? can you give some examples?
- `index` was used in for-each or other contexts?

With regards to `displayableName`, IMO we should easily change presentation so I think that we shouldn't provide this and make FE decide how to display these information.

mgoworko: I think it is now done:

case class ContextId(
    scenarioId: String,
    originatingNodeId: String,
    taskId: Long, // identifies Flink runtime task Id
    index: Long,  // next consecutive Long, from counter
    transformations: java.util.List[ContextIdTransformation] = List.empty.asJava,
) {

ContextIdTransformation's contain information about how the initial context was modified - transformed into new context, because of splits/merges. I did not introduce displayableName. There is still legacy string representation of the ContextId returned in the API.

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/ResultCollectingListenerSpec.scala
```diff
@@ -135,35 +139,91 @@ class ResultCollectingListenerSpec
       testResults => {
         transitionVariables(testResults, "bv1", Some("end1")).size shouldBe 4
         transitionVariables(testResults, "bv2", Some("end2")).size shouldBe 4
-        transitionVariables(testResults, "start-foo", Some("split")) shouldBe Set(
-          Map("input" -> 10),
-          Map("input" -> 20),
-          Map("input" -> 30),
-          Map("input" -> 40),
+        transitionVariablesByContextId(testResults, "start-foo", Some("split")) shouldBe Map(
+          ContextId("sample-split", "start-foo", 0, 0) -> Map("input" -> 10),
```
arkadius: Let's call parameters by name

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionTransformersTestModeSpec.scala
```diff
@@ -98,10 +101,46 @@ class UnionTransformersTestModeSpec
       contextIds should have size (data.size * 2)
       contextIds should contain theSameElementsAs contextIds.toSet
       contextIds should contain only (
-        s"$scenarioName-$sourceId-$firstSubtaskIndex-0-$leftBranchId-$leftBranchId",
-        s"$scenarioName-$sourceId-$firstSubtaskIndex-1-$leftBranchId-$leftBranchId",
-        s"$scenarioName-$sourceId-$firstSubtaskIndex-0-$rightBranchId-$rightBranchId",
-        s"$scenarioName-$sourceId-$firstSubtaskIndex-1-$rightBranchId-$rightBranchId",
+        ContextId(
```
arkadius, line 104: Call by name

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
```diff
@@ -42,20 +54,67 @@ object ResultsWithCountsDtoCodecs {
...
+        deriveEncoder[ContextIdDto].contramap(ContextIdDto.from)
+      Encoder.instance { value =>
+        underlying(value).deepMerge(Json.obj("cid" -> contextIdEncoder(f(value))))
+      }
+    }
+
+  }
+
+  final case class ContextIdDto(
+      sid: String,
```
arkadius: I guess that scenarioId is important in some places where `ContextId` is used - e.g. when someone want to log this information in their own `ProcessListener` by checking `Context.id`, but ... in this use case it is usesless - we know what is the `scenarioId` because we run tests for given scenario. Let's remove this field from dto.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
arkadius: Nice, a few notices:
- `scenarioId` can be misleading because we have something like `processId` which is a numeric id of row in the DB. Let's use `scenarioName` instead 
- `tranformations`/`ContextIdTransformation` sounds strange. I don't have a perfect name for that but maybe we should use just a `path` instead? From my perspecive, we have some parent context and we emit some child context based on it. I know that we already have `parentContext` which is in other dimension but I feel that it would be still better than `transformations`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -1,35 +1,43 @@
...
+  def dummy: Context = Context(ContextId.dummy)
+
+}
+
+case class ContextId(
+    scenarioId: String,
+    originatingNodeId: String,
+    taskId: Long,
+    index: Long,
+    transformations: java.util.List[ContextIdTransformation] = List.empty.asJava,
```
arkadius: Let's use scala `List` instead

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -1,35 +1,43 @@
...
+) {
+
+  lazy val legacySerialized: String = List(
```
arkadius: `legacyString` or `legacyEncoded` or `inLegacySingleStringFormat`? `serialized` sounds like we would use it in some binary format

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataCollectingListener.scala
```diff
@@ -120,6 +120,12 @@ class LiveDataCollectingListener private[livedata] (
     }
   }
 
+  private def ignoringDummyContextId(context: Context)(f: => Unit): Unit =
```
arkadius: What the heck? :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
```diff
@@ -22,10 +26,18 @@ object ResultsWithCountsDtoCodecs {
 
   private implicit val testResultsEncoder: Encoder[TestResultsDto] = new Encoder[TestResultsDto]() {
 
-    implicit val nodeResult: Encoder[ResultContext[Json]]                              = deriveConfiguredEncoder
-    implicit val expressionInvocationResult: Encoder[ExpressionInvocationResult[Json]] = deriveConfiguredEncoder
-    implicit val externalInvocationResult: Encoder[ExternalInvocationResult[Json]]     = deriveConfiguredEncoder
-    implicit val nodeTransitionResult: Encoder[NodeTransitionResult]                   = deriveConfiguredEncoder
+    implicit val contextId: Encoder[ContextId] = Encoder.encodeString.contramap(_.legacySerialized)
```
arkadius: We should keep legacy solution as easy as possible to remove. Currently, our base context id encoding strategy, encode context id in legacy format and if you want to use a new approach, you have to use specific `encoderWithDetailedContextId`. Let's invert this relation. For places where we need a legacy contextid format, let's use some explicit legacy encoder, and the default encoder let's encode it to parts. After all, removal of legacy solution should need a few lines marked with TODO to be removed.

engine/lite/components-api/src/main/scala/pl/touk/nussknacker/engine/lite/api/runtimecontext/LiteEngineRuntimeContextPreparer.scala
```diff
@@ -27,7 +27,7 @@ case class LiteEngineRuntimeContext(jobData: JobData, metricsProvider: MetricsPr
     with AutoCloseable {
 
   override def contextIdGenerator(nodeId: String): ContextIdGenerator =
-    IncContextIdGenerator.withProcessIdNodeIdPrefix(jobData, nodeId)
+    IncContextIdGenerator.withProcessIdNodeIdPrefix(jobData, nodeId, taskId = 0)
```
arkadius, line 31: Let's add a notice that for sake of code simplicity, we hardcore this `taskId` instead of passing the real one. In future it might change.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -1,35 +1,43 @@
...
-  }
-
-  def apply(id: String): Context = Context(id, Map.empty, None)
+  def apply(id: ContextId): Context = Context(id, Map.empty, None)
 
-  def apply(id: String, variables: Map[String, Any]): Context =
+  def apply(id: ContextId, variables: Map[String, Any]): Context =
     Context(id, variables, None)
 
+  def dummy: Context = Context(ContextId.dummy)
```
arkadius, line 13: Let's write a comment that it is for tests purpose. After we change `EventGeneratorSourceFactory` it will be true. Won't be?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
mgoworko: We discussed it. We decided to use java List for now. I added a very detailed comment. I also hid the java list from the interface (private constructor, private val and public constructor and getter using Scala List)

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataCollectingListener.scala
mgoworko: I refactored the EventGenerator source (reverse the order of evaluating input and creating Context), so it is not longer needed.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
mgoworko: Ok, scenarioId removed from dto

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/utils/RowConversions.scala
mgoworko: We decided to not worry about compatibility issues and just use the circe Codec without fallback.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
mgoworko: Inverted, now the legacy value is added by "decorating" the encoder.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
```diff
@@ -42,20 +66,55 @@ object ResultsWithCountsDtoCodecs {
...
+      }
+    }
+
   }
 
+  final case class ContextIdDto(
+      nid: String,
+      tid: Long,
+      idx: Long,
+      p: List[ContextIdTransformationDto]
```
arkadius: Let's call it `path`, it uses only 3 characters more, but helps a lot

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
```diff
@@ -42,20 +66,55 @@ object ResultsWithCountsDtoCodecs {
...
+        nid = id.originatingNodeId,
+        tid = id.taskId,
+        idx = id.index,
+        p = id.path.map(t => ContextIdTransformationDto(t.nodeId, t.value)),
+      )
+    }
+
+  }
+
+  final case class ContextIdTransformationDto(n: String, t: String)
```
arkadius: transfomation => path

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/UnionWithMemoTransformer.scala
```diff
@@ -88,7 +88,7 @@ class UnionWithMemoTransformer(
               val flatMapTypeInfo = typeInfoDetector.forValueWithContext(finalCtx, KeyedValueType.info(valueTypeInfo))
 
               stream
-                .map(ctx => ctx.appendIdSuffix(branchId))
+                .map(ctx => ctx.withContextIdTransformation(context.nodeId, branchId))
```
arkadius: path

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/ResultsWithCountsDtoCodecs.scala
```diff
@@ -22,10 +24,32 @@ object ResultsWithCountsDtoCodecs {
 
   private implicit val testResultsEncoder: Encoder[TestResultsDto] = new Encoder[TestResultsDto]() {
 
-    implicit val nodeResult: Encoder[ResultContext[Json]]                              = deriveConfiguredEncoder
-    implicit val expressionInvocationResult: Encoder[ExpressionInvocationResult[Json]] = deriveConfiguredEncoder
-    implicit val externalInvocationResult: Encoder[ExternalInvocationResult[Json]]     = deriveConfiguredEncoder
-    implicit val nodeTransitionResult: Encoder[NodeTransitionResult]                   = deriveConfiguredEncoder
+    implicit val contextIdTransformationDtoEncoder: Encoder[ContextIdTransformationDto] =
```
arkadius: path

arkadius (approved):

PR #8123, opened 2025-05-21
[NU-2174] Enable to save scenario within deploy action

lukasz-bigorajski:
## Checklist before merge


designer/client/src/components/modals/DeployProcessDialog.tsx
```diff
@@ -39,9 +41,16 @@ export function DeployProcessDialog(props: WindowContentProps<WindowKind, Toggle
     const [validationError, setValidationError] = useState("");
     const featureSettings = useSelector(getFeatureSettings);
     const deploymentCommentSettings = featureSettings.deploymentCommentSettings;
+    const unsavedScenarioSource: ScenarioSource = {
+        type: ScenarioSourceType.FROM_GRAPH,
+        scenarioGraph: useSelector(getGraph)?.scenario?.scenarioGraph,
+        scenarioLabels: useSelector(getScenarioLabels),
+    };
+    const savedScenarioSource: ScenarioSource = { type: ScenarioSourceType.LATEST_VERSION };
+    const scenarioSource: ScenarioSource = useSelector(isSaveDisabled) ? savedScenarioSource : unsavedScenarioSource;
```
arkadius: The logic determining if save is allowed is asynchronous. I'm afraid that, we can have some race-conditions. WDYT about always sending scenarioGraph and in the BE side checking if scenarioGraph is the same as the latest version? We could also send a scenario version which was the "starting point" for the given scenario graph. Thanks to that a user could switch between versions and deploy them without creation of new versions.

designer/client/src/http/HttpService.ts
```diff
@@ -103,6 +103,17 @@ export type SourceWithParametersTest = {
 
 export type NodesDeploymentData = Record<NodeId, Record<string, string>>;
 
+export type ScenarioSource = {
```
arkadius: scenario "source" sound like source used in the scenario. Let's call it `scenarioGraphSource` or sth other, not confusing.

designer/client/src/reducers/userSettings.ts
```diff
@@ -39,6 +40,7 @@ const getDefaultUserSettings = (): UserSettings => ({
     "debug.forceDisableModals": getInitialUserFlag("debug.forceDisableModals"),
     "debug.userSettingsVisible": getInitialUserFlag("debug.userSettingsVisible", isDev),
     "editor.jsonTemplate.showLines": getInitialUserFlag("editor.jsonTemplate.showLines", true),
+    "toolbar.enableDeployWithoutSave": getInitialUserFlag("toolbar.enableDeployWithoutSave", false),
```
arkadius: currently, we alsow allow deploy without save if nothing has changed. To be more precise, it should be called `allowDeployOfUnsvedChanges` or similar. Other more, positivie-logic approach could be `autoSaveDuringDeploy`

designer/client/src/components/modals/DeployProcessDialog.tsx
```diff
@@ -39,9 +41,16 @@ export function DeployProcessDialog(props: WindowContentProps<WindowKind, Toggle
     const [validationError, setValidationError] = useState("");
     const featureSettings = useSelector(getFeatureSettings);
     const deploymentCommentSettings = featureSettings.deploymentCommentSettings;
+    const unsavedScenarioSource: ScenarioSource = {
```
Dzuming: This approach is not optimally implemented from a performance point of view because it's not memoized, and the value will be recalculated on every rerender. Also,as I see, we use unsavedScenarioSource in two places, WDYT about moving it to the selector i and use it in both contexts?

```
const getUnsavedScenarioSource = createSelector(
    [getGraph, getScenarioLabels],
    (graph, labels) => ({
        type: ScenarioSourceType.FROM_GRAPH,
        scenarioGraph: graph?.scenario?.scenarioGraph,
        scenarioLabels: labels,
    })
);
```

and in a component

```
const unsavedScenarioSource = useSelector(getUnsavedScenarioSource);
```

designer/client/src/components/modals/DeployProcessDialog.tsx
```diff
@@ -39,9 +41,16 @@ export function DeployProcessDialog(props: WindowContentProps<WindowKind, Toggle
...
+        type: ScenarioSourceType.FROM_GRAPH,
+        scenarioGraph: useSelector(getGraph)?.scenario?.scenarioGraph,
+        scenarioLabels: useSelector(getScenarioLabels),
+    };
+    const savedScenarioSource: ScenarioSource = { type: ScenarioSourceType.LATEST_VERSION };
```
Dzuming: It's a good rule to keep elements not dependent on the state or props outside of the React component to avoid variable recreation.

WDYT about moving it outside of the component and import in both cases?

designer/client/src/reducers/selectors/graph.ts
```diff
@@ -10,6 +10,8 @@ import type { ProcessCounts, TestData } from "../graph";
 import type { RootState } from "../index";
 import { getProcessState } from "./scenarioState";
 
+export const getUserSettings = (state: RootState) => state.userSettings;
```
Dzuming: We already have this selector https://github.com/TouK/nussknacker/blob/e925abd9c6d727e69cde063c25085f60c0b10dc0/designer/client/src/reducers/selectors/userSettings.ts#L10-L9

Dzuming (changes requested):

designer/client/src/reducers/selectors/graph.ts
lukasz-bigorajski: Done

designer/client/src/components/modals/DeployProcessDialog.tsx
lukasz-bigorajski: Done

designer/client/src/components/modals/DeployProcessDialog.tsx
lukasz-bigorajski: Done

designer/client/src/reducers/userSettings.ts
lukasz-bigorajski: Done

designer/client/src/http/HttpService.ts
lukasz-bigorajski: Done

designer/client/src/components/modals/DeployProcessDialog.tsx
lukasz-bigorajski: Done

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
```diff
@@ -75,8 +74,17 @@ export default function DeployButton(props: ToolbarButtonProps) {
...
-            HttpService.deploy(name, comment, nodesDeploymentData).finally(() => dispatch(loadProcessState(name, versionId))),
+        async (
+            name: ProcessName,
+            versionId: ProcessVersionId,
+            comment: string,
+            nodesDeploymentData?: NodesDeploymentData,
+            scenarioGraphSource?: ScenarioGraphSource,
+        ) => {
+            const result = await HttpService.deploy(name, comment, nodesDeploymentData, scenarioGraphSource);
+            dispatch(fetchProcessToDisplay(name, result.deployedScenarioVersionId ?? versionId));
```
Dzuming: What is a case when the frontend won't receive `result.deployedScenarioVersionId`?

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
```diff
@@ -87,10 +95,17 @@ export default function DeployButton(props: ToolbarButtonProps) {
                     await confirm({
                         text: t(
                             "panels.actions.confirm-unsafe-deployment.message",
-                            `There is newer version #{{latestVersion}} created by {{modifyBy}} available. Scenario will be deployed using the newest version.
-                         You're currently checked out on version #{{localVersion}}. 
-                         Are you sure you want to perform this action?`,
-                            { latestVersion: res.data.latestVersion, modifyBy: res.data.modifiedBy, localVersion: res.data.localVersion },
+                            `There is newer version #{{latestVersion}} created by {{modifyBy}} available. Scenario will be deployed using version #{{versionToDeploy}}.
+                             You're currently checked out on version #{{localVersion}}.
```
Dzuming: Should we display ` You're currently checked out on version #{{localVersion}}.` when we have the latest version opened locally?

Dzuming (changes requested):

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
lukasz-bigorajski: The FE will only receive `result.deployedScenarioVersionId` in deploy/redeploy action, other action would require API changes

Dzuming: Ok, so if I understand correctly, in case of a response from HttpService.deploy, `result.deployedScenarioVersionId` will always be present?

lukasz-bigorajski: > in case of a response from HttpService.deploy, result.deployedScenarioVersionId will always be present?

Yes. It's the correct assumption.

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
lukasz-bigorajski: I refactored it a little bit. To be precise this dialog is only displayed if scenario is not the latest.

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
```diff
@@ -82,7 +83,12 @@ export default function DeployButton(props: ToolbarButtonProps) {
             scenarioGraphSource?: ScenarioGraphSource,
         ) => {
             const result = await HttpService.deploy(name, comment, nodesDeploymentData, scenarioGraphSource);
-            dispatch(fetchProcessToDisplay(name, result.deployedScenarioVersionId ?? versionId));
+            if (result.scenarioActionResultType === ScenarioActionResultType.Success) {
+                const deployedScenarioVersionId = (result as DeployedScenarioActionResult).deployedScenarioVersionId;
```
Dzuming: To perform correctly, type narrowing and remove as DeployedScenarioActionResult type assertion, you can change the types to Discriminated Unions

```
export enum ScenarioActionResultType {
    Success = "SUCCESS",
    ValidationError = "VALIDATION_ERROR",
    UnhandledError = "UNHANDLED_ERROR",
}

interface ScenarioActionMessage {
    msg: string;
}

export type ScenarioActionValidationError = ScenarioActionMessage & {
    scenarioActionResultType: ScenarioActionResultType.ValidationError;
};

export type ScenarioActionUnhandledError = ScenarioActionMessage & {
    scenarioActionResultType: ScenarioActionResultType.UnhandledError;
};

export type DeployedScenarioActionResult = ScenarioActionMessage & {
    scenarioActionResultType: ScenarioActionResultType.Success;
    deployedScenarioVersionId: number;
};

export type ScenarioActionResult = ScenarioActionValidationError | ScenarioActionUnhandledError | DeployedScenarioActionResult;
```

Dzuming (changes requested):

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
lukasz-bigorajski: Done

Dzuming (approved):
LGTM

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -104,14 +118,13 @@ class ActionService(
     }
   }
 
-  class ActionProcessor[LatestScenarioDetailsShape: ScenarioShapeFetchStrategy](
-      extractVersionOnWhichActionIsDoneFromLatestScenarioDetails: ScenarioWithDetailsEntity[
-        LatestScenarioDetailsShape
-      ] => Option[VersionId]
+  class ActionProcessor[ScenarioDetailsShape: ScenarioShapeFetchStrategy](
+      extractVersionOnWhichActionIsDone: ScenarioWithDetailsEntity[ScenarioDetailsShape] => Option[VersionId],
+      scenarioGraphSource: ScenarioGraphSource = LatestVersion,
```
arkadius: Let's remove the default value for this parameter. The default value has sense when it is used very very often and is obvious. In any other situation, it can cause a potential bug because developer forgets to pass it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+        processId: ProcessIdWithName,
+        scenarioGraph: ScenarioGraph,
+        scenarioLabels: Option[List[ProcessingType]],
+        scenario: ScenarioWithDetailsEntity[CanonicalProcess]
+    )(implicit user: LoggedUser) = {
+      val parameters = scenarioParametersServiceProvider.combined
+        .getParametersWithReadPermissionUnsafe(scenario.processingType)
+      val scenarioDetails = ScenarioWithDetailsConversions
+        .fromEntityIgnoringGraphAndValidationResult(scenario, parameters)
+      val updateCommand = UpdateScenarioCommand(scenarioGraph, None, scenarioLabels)
```
arkadius: Let's call `None` by name

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+      processService
+        .updateProcessDBIO(processId, scenarioDetails, updateCommand)
+        .flatMap(response =>
+          response.processResponse match {
+            case Some(value) =>
+              logger.debug(
+                s"Scenario: ${value.processName.value} updated with automatic update." +
+                  s" New version is: ${value.versionId.value}"
+              )
+              processRepository.fetchProcessDetailsForId[ScenarioDetailsShape](processId.id, value.versionId)
```
arkadius: Can we avoid fetching details that were already updated by us? Can't we just do some `scenario.copy(...)` ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+            case None =>
+              logger.info(
+                "Scenario is not updated (latest version graph is the same as provided graph) - " +
+                  "deploying latest version"
```
arkadius: Let's add an information about `baseScenarioVersionId`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+    )(implicit user: LoggedUser) = {
+      val parameters = scenarioParametersServiceProvider.combined
+        .getParametersWithReadPermissionUnsafe(scenario.processingType)
+      val scenarioDetails = ScenarioWithDetailsConversions
+        .fromEntityIgnoringGraphAndValidationResult(scenario, parameters)
+      val updateCommand = UpdateScenarioCommand(scenarioGraph, None, scenarioLabels)
+      processService
+        .updateProcessDBIO(processId, scenarioDetails, updateCommand)
+        .flatMap(response =>
+          response.processResponse match {
```
arkadius, line 229: 1. It looks strange that in our domain service (`ActionService`) we process some DTO. I see that it is a deeper problem, because `ProcessService` already is not a real domain service and it returns DTOs in other methods. Do you see some easy way to fix it, or is it too much work to tidy it up?
2. Also, this is very tricky that some `processResponse: None` means that nothing changed in the version. In scala, I would expect:
- ADT: `NoVersionChanges` | `NewVerssionWasAdded(versionId)`
- Nested class with a field called `newVersionInformation: Option[VersionId]`
-  Single `Option[VersionId]` as a return type with some comment

Why this method returns scenario id and scenario name that were passed?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+        case LatestVersion =>
+          processRepository.fetchLatestProcessDetailsForProcessId[ScenarioDetailsShape](processId.id)
+        case FromGraph(scenarioGraph, scenarioLabels, baseScenarioVersionId) =>
+          processRepository
+            .fetchProcessDetailsForId[CanonicalProcess](processId.id, baseScenarioVersionId)
+            .flatMap {
+              case Some(scenario) if CanonicalProcessConverter.toScenarioGraph(scenario.json) == scenarioGraph =>
+                val scenarioWithTargetShape =
+                  scenario.mapScenario(ScenarioShapeFetchStrategy.convertToTargetShape[ScenarioDetailsShape])
+                DBIOAction.successful(Some(scenarioWithTargetShape))
```
arkadius: Let's log this situation as well.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
```diff
@@ -181,9 +190,62 @@ class ActionService(
...
+      val scenarioDetails = ScenarioWithDetailsConversions
+        .fromEntityIgnoringGraphAndValidationResult(scenario, parameters)
+      val updateCommand = UpdateScenarioCommand(scenarioGraph, None, scenarioLabels)
+      processService
+        .updateProcessDBIO(processId, scenarioDetails, updateCommand)
+        .flatMap(response =>
+          response.processResponse match {
+            case Some(value) =>
+              logger.debug(
+                s"Scenario: ${value.processName.value} updated with automatic update." +
```
arkadius: Let's add an information about `baseScenarioVersionId`

arkadius (commented):
Nice, some minor comments added.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: Done

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/ActionService.scala
lukasz-bigorajski: I don't know the logic behind the case so I leave a comment to refactor this and I will refactor it the next commit

arkadius (approved):

PR #8158, opened 2025-05-26
[NU-2216] Spel parser: returning text range for error

arkadius:
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
```diff
@@ -24,6 +24,10 @@ object ExpressionParseError {
       columnDefinitions: List[ColumnDefinition]
   ) extends ErrorDetails
 
+  @JsonCodec final case class TextErrorDetails(start: TextLocation, end: TextLocation) extends ErrorDetails
+
+  @JsonCodec final case class TextLocation(column: Int, row: Int)
```
arkadius: This format is convenient to use on the FE side

scenario-compiler/src/main/java/pl/touk/nussknacker/engine/spel/parser/NuSpelExpressionParser.java
```diff
@@ -0,0 +1,303 @@
...
+ * /**
+ * An expression parser that understands templates. It can be subclassed by expression
+ * parsers that do not offer first class support for templating.
+ *
+ * @author Keith Donald
+ * @author Juergen Hoeller
+ * @author Andy Clement
+ * @since 3.0
+ */
+public class NuSpelExpressionParser {
```
arkadius, line 46: This class was already in the repo (`NuTemplateAwareExpressionParser`). The suggestion mechanism used it. I had to make it used also during expression parsing in order to provide information about location of an expression inside the template which is used in `PositionRange.shiftRight`. Because of that I remodeled a bit the returned types of methods.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpression.scala
```diff
@@ -329,24 +342,4 @@ object SpelExpressionParser extends LazyLogging {
     )
   }
 
-  private class AdjustingTemplateExpressionSpringSpelExpressionParser(configuration: SpelParserConfiguration)
```
arkadius, line 332: This now happens in the `NuSpelExpressionParser`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/Typer.scala
```diff
@@ -129,24 +153,35 @@ private[spel] class Typer(
       current: TypingContext
   ): NodeTypingResult = {
 
-    def toNodeResult(typ: TypingResult) = current.toResult(TypedNode(node, TypingResultWithContext(typ)))
+    implicit class TypingResultOps(typ: TypingResult) {
```
arkadius, line 156: There are a lot of changes in this file, but basically:
- I changed prefix notation to postfix notation for some methods to reduce the level of braces nesting (I had to add one more level (`withPositionRange`).
- I made some private class methods local for `typeNode`.

Thanks to these changes, it was easy to add information about the position range to every error.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
```diff
@@ -2,13 +2,15 @@ package pl.touk.nussknacker.engine.spel
...
-import pl.touk.nussknacker.engine.api.parameter.ParameterName
+import pl.touk.nussknacker.engine.api.generics.ExpressionParseError.TextErrorDetails
 import pl.touk.nussknacker.engine.api.typed.typing.{TypedDict, TypingResult}
+import pl.touk.nussknacker.engine.expression.PositionRange
+import pl.touk.nussknacker.engine.spel.SpelExpressionParseError.SpelExpressionTypingParseError
+import pl.touk.nussknacker.engine.spel.SpelExpressionTypingError.SpelExpressionTypingError
 
-object SpelExpressionParseError {
-  trait SelectionProjectionError extends ExpressionParseError
+object SpelExpressionTypingError {
```
arkadius, line 12: All errors returned in validation are not `ExpressionParseError` because we want to enrich error details (position range) in one place

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
mgoworko: nitpick: maybe the name could be something like `CoordinatesInText` or `TextCoordinates`. At the moment the name suggest, that it is the location, where some text starts.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionSubstitutionsCollector.scala
```diff
@@ -10,4 +11,22 @@ trait ExpressionSubstitutionsCollector {
...
 case class ExpressionSubstitution(position: PositionRange, replacement: String)
 
-case class PositionRange(start: Int, end: Int)
+case class PositionRange(start: Int, end: Int) {
+
+  def shiftRight(delta: Int): PositionRange = PositionRange(start + delta, end + delta)
+
+  def toErrorDetails(expression: String): TextErrorDetails = {
+    def computeTextLocation(text: String): TextLocation =
+      TextLocation(
```
mgoworko: I think that there might be some issue with column/row calculations for multi-line string.

For multiline string:
```
abc
def
ghi
```
with  start=6,. end=7, the result is start=(column=2, row=1 - correct), end=(column=3, row=1 - incorrect, there is no column with index=3)

I think that this issue is related to more than one newline in the string, that is processes by `computeTextLocation`

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
```diff
@@ -233,43 +257,46 @@ object SpelExpressionParseError {
...
-  case class ExpressionTypeError(expected: TypingResult, found: TypingResult) extends ExpressionParseError {
+  case class ExpressionTypeError(expected: TypingResult, found: TypingResult) extends SpelExpressionTypingError {
     override def message: String = s"Bad expression type, expected: ${expected.display}, found: ${found.display}"
   }
 
-  case class ExpressionCompilationError(message: String) extends ExpressionParseError
+}
 
-  case class KeyWithLabelExpressionParsingError(keyWithLabel: String, message: String) extends ExpressionParseError {
+trait SpelExpressionParseError extends ExpressionParseError
```
mgoworko: sealed trait?

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
```diff
@@ -1527,8 +1538,12 @@ class SpelExpressionSpec extends AnyFunSuite with Matchers with ValidatedValuesD
   }
 
   test("should not validate division by zero") {
-    parse[Any]("1 / 0").invalidValue shouldBe NonEmptyList.one(DivisionByZeroError("(1 / 0)"))
-    parse[Any]("1 % 0").invalidValue shouldBe NonEmptyList.one(ModuloZeroError("(1 % 0)"))
+    parse[Any]("1 / 0").invalidValue.toList should matchPattern {
+      case SpelExpressionTypingParseError(DivisionByZeroError("1 / 0"), _) :: Nil =>
```
mgoworko, line 1542: I wonder if this test shows, that there are 2 separate categories of errors:
 - syntax errors (invalid syntax, missing method etc.) 
 - execution errors - when the syntax is correct, but performing the operation causes failure

I think that `SpelExpressionTypingParseError` could be renamed to sth like `SpelExpressionTypingErrorWithDetails` (because it is not necessarily parsing error)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/ast/SpelAst.scala
```diff
@@ -1,6 +1,7 @@
 package pl.touk.nussknacker.engine.spel.ast
 
 import org.springframework.expression.spel.SpelNode
+import pl.touk.nussknacker.engine.api.generics.ExpressionParseError.TextErrorDetails
```
mgoworko: unused import

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -12,17 +12,18 @@ import pl.touk.nussknacker.engine.api.context.ValidationContext
 import pl.touk.nussknacker.engine.api.exception.NonTransientException
 import pl.touk.nussknacker.engine.api.expression.ExpressionTypingInfo
 import pl.touk.nussknacker.engine.api.generics.ExpressionParseError
+import pl.touk.nussknacker.engine.api.generics.ExpressionParseError.TextErrorDetails
```
mgoworko: unused import?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/language/json/JsonTemplateParser.scala
```diff
@@ -12,17 +12,18 @@ import pl.touk.nussknacker.engine.api.context.ValidationContext
...
 import pl.touk.nussknacker.engine.api.json.decoders.FromJsonTypingResultBasedDecoder
 import pl.touk.nussknacker.engine.api.typed.typing
 import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypedClass, TypingResult}
 import pl.touk.nussknacker.engine.definition.component.parameter.defaults.TypeValueDeterminer
+import pl.touk.nussknacker.engine.expression.PositionRange
```
mgoworko: unused import?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
```diff
@@ -233,43 +257,46 @@ object SpelExpressionParseError {
...
 
-    def toProcessCompilationError(
-        nodeId: String,
-        paramName: ParameterName
-    ): ProcessCompilationError.KeyWithLabelExpressionParsingError =
-      ProcessCompilationError.KeyWithLabelExpressionParsingError(keyWithLabel, message, paramName, nodeId)
+object SpelExpressionParseError {
 
-  }
+  case class SpelExpressionUnderlyingParserError(message: String, override val details: Option[TextErrorDetails])
```
mgoworko: Do we need this case class? I noticed it is used only once, in `SpelExpression.baseParse`. 

Maybe the underlying spring's class `ParseException` could be wrapped in some new implementation of `SpelExpressionTypingError`. Then we could just use `SpelExpressionTypingParseError`. And when the exception is something different than `ParseException` maybe we could just throw an exception (because IMO we should not wrap other exceptions as `SpelExpressionUnderlyingParserError`)

mgoworko (changes requested):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/generics/ExpressionParseError.scala
arkadius: Renamed to `TextCoordinates`, also I did some other renames:
- `PositionRange` -> `IndexBasedTextRange` 
- `...WithPosition` -> `...WithTextRange`
- `TextErrorDetails` -> `CoordinatesBasedTextRange`

, to make easier (I hope) to understand what's going on with conversion between index-based positions and coordinates. Take a look

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionSubstitutionsCollector.scala
arkadius: It gives correct coordinates. I added a unit test that shows what happens under the hood during the conversion to coordinates. Take a look

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
arkadius: It doesn't show execution errors. We have built-in mechanism which in typical programming languages is called "static code analysis". In our code it is done in `Typer` (typing process). During this process we recognize what is the `TypedObjectWithValue`. As a "side effect" we do this analysis. Unfortunatelly it is not easy to use this kind of names in our code, because static code analysis is combined with typing. 

By this class name I wanted to show that this class translates typing errors to `ExpressionParseError`. `ExpressionParseError` is errors that can be returned by any of our `ExpressionParser`. I see that these names became long but on the other hand we are consistent in naming, when we use names such as `SpelExpressionTypingErrorWithDetails` we will loose it. What do you thing, this  consistent naming is less important than shorter names?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
arkadius: We have no control over what Exception spel throwns. Runtime exceptions are not visible in the interfaces so we can't assume that the only possible exception that can be thrown is `ParseException`. I don't want to block user when we bump spel and some new type of Exception will be thrown.

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelExpressionSpec.scala
mgoworko: Ok, so after this explanation my comment doesn't make sense anymore, you can resolve it. 

I thought, that errors like `DivisionByZeroError` and other similar cases could be considered as some separate category. But if we look at it as you wrote - `static code analysis` - this is no longer applicable. So the name suggestion is no longer applicable too (I have nothing against reasonably long names; the comment was about the meaning of the name, but is no longer valid).

mgoworko (approved):
LGTM

PR #8165, opened 2025-05-29
Live data changes - apply sample limit per node transition and added total count

mgoworko:
## Describe your changes

Small changes:
- unique ContextId after split 
- current time returned in the liveData API endpoint response

Important changes:
- live data stored in separate ring buffers per transition
- cumulative counter per transition since the scenario is deployed
- added invocation results and exceptions per node id in the liveData API endpoint response

The changes required breaking changes in the liveData API endpoint response

## Checklist before merge


designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/RingBufferWithTotalCount.scala
```diff
@@ -0,0 +1,29 @@
...
+    override protected def removeEldestEntry(eldest: JMap.Entry[Long, T]): Boolean = size() > maxSize
+  }
+
+  private val counter = new AtomicLong(0)
+
+  def values: List[T] = underlying.synchronized {
+    underlying.values().asScala.toList
+  }
+
+  def totalCount: Long = {
```
arkadius: If we have synchronized in 2 of 3 methods, maybe here we should also add `synchronized` and replace counter with just `var: Long`. I don't think that we gain a lot with this `AtomicLong` and I'm worry about nesting two kinds of synchronization. Maybe not this time, but in many similar cases it causes a deadlock

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
```diff
@@ -25,7 +25,7 @@ private[livedata] class SlidingWindowCounter[T](
 
     // We want to calculate correct throughput just after the scenario is started
     val windowStart      = Math.max(counterCreatedAt.getEpochSecond, currentEpochSecond - windowSizeSeconds)
-    val windowsEnd       = currentEpochSecond
+    val windowsEnd       = currentEpochSecond - 1
```
arkadius: After this change, `windowsEnd` can be before `windowStart`. Is it an expected behaviour?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioLiveData/Dtos.scala
```diff
@@ -1,105 +1,120 @@
...
-  implicit def nodeCountSchema: Schema[NodeCount]                                         = Schema.anyObject
-  implicit def nodeTransitionThroughputDtoSchema: Schema[NodeTransitionThroughputDto]     = Schema.derived
-  implicit def resultsWithCountsSchema: Schema[LiveDataDto]                               = Schema.derived
-  implicit def typingResultDecoder: Decoder[TypingResult] = Decoder.decodeJson.map(_ => typing.Unknown)
-  implicit def scenarioGraphSchema: Schema[ScenarioGraph] = Schema.anyObject
+  implicit def exceptionResultDtoSchema: Schema[ExceptionResultDto]                     = Schema.derived
+  implicit def invocationResultDtoSchema: Schema[InvocationResultDto]                   = Schema.derived
+  implicit def liveDataSampleDtoSchema: Schema[LiveDataSampleDto]                       = Schema.derived
+  implicit def liveDataForNodeTransitionDtoSchema: Schema[LiveDataForNodeTransitionDto] = Schema.derived
+  implicit def resultsWithCountsSchema: Schema[LiveDataDto]                             = Schema.derived
```
arkadius: liveDataDtoSchema

designer/live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/SlidingWindowCounter.scala
mgoworko: The situation with negative interval may happen only when we fetch the live data during the second, when the counter is created. I modified implementation to handle it, `val throughput = if (samplingInterval <= 0) BigDecimal(0) else BigDecimal(count) / samplingInterval`. The details are below:

I made the small change (I'm talking about the version that you reviewed, not changes after review), to not present data from the current second, that is in progress. Instead, the data from N _previous_ seconds is taken into account. I did it, because after adding totalCount I observed strange behavior with EventGenerator and totalCount. The result depended on precise timing inside the current second. Now the current second is not taken into account until it is complete. I forgot to mention that in the description, sorry. It is now tested and checked over using LLM (I asked it to analyze all possible edge cases)

arkadius: Can you add a small comment next to this `samplingInterval <= 0` that this is to handle this edge case with handling "throughput just after the scenario is started"? I assume that in other cases `samplingInterval == windowSizeSeconds - 1`? Is it ok, that during computing this value we always divice by size -1 ? Looks like result can be higher than the real result. Or I'm missing sth?

mgoworko: No, the interval is as configured - let's consider the most common case 
- let's say the time window is 60 seconds and the scenario has been running for more than 60 seconds
- so the calculation simplify, and we have
```
    // currentEpochSecond = 1748953079

    // windowStart = 1748953079 - 60 = 1748953019
    val windowStart = currentEpochSecond - windowSizeSeconds  
   
    // windowEnd = 1748953078
    val windowEnd        = currentEpochSecond - 1
    
    // samplingInterval = 1748953078 - 1748953019 + 1 = 60
    val samplingInterval = windowEnd - windowStart + 1
```
- so the sampling interval is exactly 60 seconds - from second 1748953019 to 1748953078 (inclusive)
- that is because before the calculations we clean the buckets older than 1748953019 (non-inclusive)

I added 2 comments about those 2 things

arkadius: Ah, yeah I missed the `+1`. `currentEpochSecond - 1 - (currentEpochSecond - windowSizeSeconds) + 1 =  windowSizeSeconds`

arkadius (approved):

PR #8174, opened 2025-06-02
Nu.2174.mark.automatically.saved.scenarios

lukasz-bigorajski:
## Describe your changes
image

## Checklist before merge


arkadius:
image
IMO it would be better to use some simple icon instead of `(A)`. From my experience, in this activity panel we already 
have too much ascii art, we have a date without label, than some pipe. It looks very "robotic". Also, this (i) symbol next to the dot showing the current version can be confusing. We can add a tooltip for this icon. What do you think?

lukasz-bigorajski:
> image IMO it would be better to use some simple icon instead of `(A)`. From my experience, in this activity panel we already have too much ascii art, we have a date without label, than some pipe. It looks very "robotic". Also, this (i) symbol next to the dot showing the current version can be confusing. We can add a tooltip for this icon. What do you think?

still it look like this:
image

So I set down with Dawid to change only text like to 'autosaved`

image

lukasz-bigorajski:
I eventually choose sth like this:
image

I was also trying with other backgrounds, but I think it looks bad:
image
image

arkadius:
> I eventually choose sth like this

Nice the first option indeed looks the best.

arkadius (approved):

Dzuming (approved):
LGTM

PR #8182, opened 2025-06-04
[NU-2232] Handle chat stream errors in assistant

jedrz:
* `initializeChatStream` returns a generator of chat stream events instead of a promise with mutable response parts array and stream state.
* An error that occurs in the chat stream is displayed in the assistant window, and the last question can be retried.

Screenshot 2025-06-04 at 11-42-25 Nussknacker


Dzuming (approved):
LGTM

PR #8190, opened 2025-06-04
[NU-2154] Restore legacy counts for live data

mgoworko:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
```diff
@@ -64,10 +69,55 @@ class ScenarioLiveDataApiHttpService(
...
+      t.destinationNodeId match {
+        case Some(dest) => List(t.sourceNodeId, dest)
+        case None       => List(t.sourceNodeId)
+      }
+    }.toSet
+
+    val incomingCounts =
+      liveData.nodeTransitions.toList
+        .flatMap { case (transition, data) => transition.destinationNodeId.map(_ -> data.totalCount) }
+        .groupBy(_._1)
```
arkadius: We have a shortcut for that - `toGroupedMap.mapValuesNow(_.sum)`

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
```diff
@@ -64,10 +69,55 @@ class ScenarioLiveDataApiHttpService(
...
-        } yield LiveDataDto.from(liveData)
+        } yield LiveDataDto.from(liveData, computeCounts(scenarioWithDetails, liveData))
       }
   }
 
+  private def computeCounts(scenarioWithDetails: ScenarioWithDetails, liveData: LiveData)(
+      implicit loggedUser: LoggedUser
+  ): Map[String, NodeCount] = {
+    val uniqueNodeIds: Set[String] = liveData.nodeTransitions.keys.flatMap { t =>
+      t.destinationNodeId match {
```
arkadius: You can use `t.sourceNodeId :: t.destinationNodeId.toList` instead

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
```diff
@@ -64,10 +69,55 @@ class ScenarioLiveDataApiHttpService(
...
+        .map { case (nodeId, counts) => nodeId -> counts.map(_._2).sum }
+
+    val outgoingCounts =
+      liveData.nodeTransitions.toList
+        .map { case (transition, data) => transition.sourceNodeId -> data.totalCount }
+        .groupBy(_._1)
+        .map { case (nodeId, counts) => nodeId -> counts.map(_._2).sum }
+
+    // We calculate counts based on transitions incoming to node, but if there are none (for sources) we use outgoing counts
+    val nodeCounts = uniqueNodeIds.map { nodeId =>
```
arkadius: `.map` on `Set` creates a new `Set` (computes all hashcodes), which is expensive, and we don't need to pay for this. To avoid this, you can use `.toList`. Also, please move `uniqueNodeIds` closer to this line to emphasize that this structure is needed for this purpose only.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
```diff
@@ -64,10 +69,55 @@ class ScenarioLiveDataApiHttpService(
...
+      nodeId -> incomingCounts
+        .get(nodeId)
+        .orElse(outgoingCounts.get(nodeId))
+        .getOrElse(throw new IllegalStateException(s"Could not calculate counts for node $nodeId"))
+    }
+
+    val counts = nodeCounts.map { case (nodeId, count) =>
```
arkadius: Use can use `mapValuesNow` and shorten it a bit. BTW I see that after all we pass this map as a fuction so maybe instead of creating a `Map` we can just make a local function that would compute it ad-hoc. WDYT?

mgoworko: Refactored as a function

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
mgoworko: I changed it to use .distinct on list

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioLiveData/Dtos.scala
```diff
@@ -48,6 +50,7 @@ object Dtos {
         exceptionsByNodeId = liveData.exceptions.map { case (nodeId, results) =>
           nodeId.id -> results.map(ExceptionResultDto.from)
         },
+        counts = counts,
```
arkadius: In `ResultsWithCountsDto` we have a nested structure - `results` and `counts` next to it. I guess that our FE would prefer to have exactly the same structure.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioLiveDataApiHttpService.scala
arkadius: I mean sth like this:
```
    def getCount(nodeId: String): Option[RawCount] = {
      Some(
        RawCount(
          incomingCounts
            .get(nodeId)
            .orElse(outgoingCounts.get(nodeId))
            .getOrElse(0),
          liveData.exceptions.getOrElse(NodeId(nodeId), List.empty).size.toLong
        )
      )
    }
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioLiveData/Dtos.scala
mgoworko: Done. But the content of the "results" field is a bit different anyway, because we have slightly different format of nodeTransitions, with total counts and throughput. 

Test results and live data are a bit different cases, and I feel like I'm going in circles - first trying to refactor live data and separate it from tests, and now trying to make them as similar as possible.

arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioLiveData/Dtos.scala
arkadius: Similarity of result type is useful for handling the result on FE side where this result can be passed through a few rendering layers of react components. On the other hand, similarity of endpoints doesn't help in the same way but causes problems in understanding the domain entities exposed via REST api. The result structure is a little bit different, but in type script there are concept like intersection types which are very useful for such a case. You can declare some type with some extension of this type and in one place pass the base type and in other this type with extension. For the same result, on FE it is very useful to return an additional field called "disrimiantor" which helps in routing the result type to the correct sub component, take a look at example: https://medium.com/@km87/typescript-discriminated-unions-d3b54bf14399 we could add it and help FE even more. We could also add an a trait that both DTOs extends. This trait would help us keep both structures synchronized.

mgoworko: Now the `/liveData` endpoint returns exactly the same JSON body as test results. 
I did not introduce discriminator, because:
- the FE knows which endpoint it calls
- the JSON response is the same, with some optional fields (total count and throughput), and FE knows which endpoint it called, so IMO there is no point returning the same `type` field each time

arkadius (approved):

PR #8192, opened 2025-06-05
Rename parameters in the filter and previous-value components

jedrz:
* filter: Expression -> Condition
* previous-value: groupBy -> Key, value -> Value 

Screenshot 2025-06-05 at 09-51-39 Nussknacker
Screenshot 2025-06-05 at 11-17-03 Nussknacker


engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/PreviousValueTransformer.scala
```diff
@@ -18,15 +18,15 @@ case object PreviousValueTransformer extends CustomStreamTransformer with Explic
 
   @MethodToInvoke(returnType = classOf[Value])
   def execute(
-      @ParamName("groupBy") groupBy: LazyParameter[CharSequence],
+      @ParamName("key") key: LazyParameter[CharSequence],
```
arkadius: Let's use `Title case`  for `Key` and `Value`. We use this convention everywhere else.

arkadius (approved):

PR #8213, opened 2025-06-09
[NU-2217] JSON parameter template instead of dynamic form for Kafka Sink

mateuszkp96:
## Describe your changes
image

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -9,6 +9,7 @@ import pl.touk.nussknacker.engine.api.namespaces.NamingStrategy
 
 final case class ModelConfig(
     allowEndingScenarioWithoutSink: Boolean,
+    enableSingleParameterWithTemplateInsteadOfDynamicForm: Boolean,
```
arkadius: As we discussed on the slack, I personally don't like booleans in config because:
- You have to decide which is the default, negative value, and which is the opt-in value, this strategy may change over the time, if you decide that it should be opposite
- You can't easily add the third behaviour; eventually you have to break compatibility by adding another one property or support >1 configuration schemas 
- Names of properties quickly become complex and long

I propose `jsonLikeValuesEnteringMode: DynamicForms | SingleJsonTemplateParameter`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/decoders/FromJsonSimpleDecoder.scala
```diff
@@ -1,25 +1,43 @@
...
-        // We prefer java big decimal over float/double
-        jsonNumber.toBigDecimal.map(_.bigDecimal)
-        getOrElse (throw new IllegalArgumentException(s"Not supported json number: $jsonNumber")),
+    jsonNumber = toNumber,
     jsonString = identity[String],
     jsonArray = _.map(jsonToAny).asJava,
     jsonObject = _.toMap.mapValuesNow(jsonToAny).asJava
   )
 
+  def jsonToAnyWithOrderKeeping(json: Json): Any = json.fold(
```
arkadius: Why not to always keep order? In `jsonToAny` as well?

engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/kafkaschemaless/BaseKafkaJsonSchemalessItSpec.scala
```diff
@@ -51,15 +51,13 @@ abstract class BaseKafkaJsonSchemalessItSpec extends FlinkWithKafkaSuite {
         .emptySink(
           "end",
           "kafka",
-          KafkaUniversalComponentTransformer.sinkKeyParamName.value       -> "".spel,
-          KafkaUniversalComponentTransformer.sinkRawEditorParamName.value -> "true".spel,
```
arkadius, line 55: Is the test for old behaviour, where you have to provide this parameter value, still available somewhere?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonTemplateFromJsonSchemaDeterminer.scala
```diff
@@ -0,0 +1,92 @@
...
+            .to(LazyList)
+            .flatMap(schema => defaultJsonFor(schema))
+            .headOption
+        case _ => None
+      }
+    case refSchema: ReferenceSchema             => defaultJsonFor(refSchema.getReferredSchema)
+    case _: StringSchema                        => Some(Json.fromString(""))
+    case s: NumberSchema if s.requiresInteger() => Some(Json.fromInt(0))
+    case _: NumberSchema                        => Some(Json.fromDoubleOrNull(0.0))
+    case _: BooleanSchema                       => Some(Json.fromBoolean(true))
```
arkadius: Is `false` not a more natural default value for a boolean?

utils/json-utils/src/main/scala/pl/touk/nussknacker/engine/json/JsonTemplateFromJsonSchemaDeterminer.scala
```diff
@@ -0,0 +1,92 @@
...
+      val defaultValue = schema.getDefaultValue
+      Some(jsonToCirce(defaultValue))
+    case other =>
+      defaultValueBasedOnSchema(other)
+  }
+
+  private def defaultValueBasedOnSchema(schema: Schema): Option[Json] = schema match {
+    case _: EmptySchema => Some(Json.obj())
+    case objSchema: ObjectSchema =>
+      val props    = ListMap(objSchema.getPropertySchemas.asScala.toList.sortBy(_._1): _*)
```
arkadius, line 43: Hmm, or maybe let's keep the order as in schema? Sometimes the order have some purpose, e.g. `name`, `surename`, etc.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/encode/AvroSchemaOutputValidator.scala
```diff
@@ -112,7 +119,7 @@ class AvroSchemaOutputValidator(validationMode: ValidationMode) extends LazyLogg
       schema: Schema,
       path: Option[String]
   ): Validated[NonEmptyList[OutputValidatorError], Unit] = {
-    val schemaFields = schema.getFields.asScala.map(field => field.name() -> field).toMap
+    val schemaFields = ListMap(schema.getFields.asScala.toList.map(field => field.name() -> field).sortBy(_._1): _*)
```
arkadius: The same here

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schema/AvroSchemaBasedParameter.scala
```diff
@@ -67,7 +66,7 @@ object AvroSchemaBasedParameter {
           )
         )
       } else {
-        val listOfValidatedParams = recordFields.map { recordField =>
+        val listOfValidatedParams = recordFields.sortBy(_.name()).map { recordField =>
```
arkadius: Why did you change it?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schema/AvroSchemaBasedParameter.scala
```diff
@@ -58,7 +58,6 @@ object AvroSchemaBasedParameter {
         Invalid(
           NonEmptyList.one(
             CustomNodeError(
-              nodeId.id,
```
arkadius, line 61: Why is it removed? Based on this id, errors are dispatched on the FE. Is it now set somewhere else?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/confluent/client/MockSchemaRegistryClientBuilder.scala
```diff
@@ -45,16 +56,25 @@ class MockConfluentSchemaRegistryClientBuilder {
...
 private[client] object RegistryItem {
 
   // Default value for autoincrement mock id
   private val AutoIncId = -1
 
   def apply(topic: String, schema: String, version: Int, isKey: Boolean): RegistryItem =
-    new RegistryItem(topic, AvroUtils.parseSchema(schema), version, isKey, AutoIncId)
+    new RegistryItem(topic, toParsedSchema(AvroUtils.parseSchema(schema), version), version, isKey, AutoIncId)
+
+  def apply(topic: String, schema: Schema, version: Int, isKey: Boolean): RegistryItem = {
```
arkadius: It looks like, `RegistryItem.apply` is used only once. I think that someone, before you, overengineered this API, preparing it for many usages in many different cases. Let's simplify this class by removing all apply methods and discuss the API which is more commolny used (`MockConfluentSchemaRegistryClientBuilder.register`)

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/confluent/client/MockSchemaRegistryClientBuilder.scala
```diff
@@ -31,10 +33,19 @@ class MockConfluentSchemaRegistryClientBuilder {
     this
   }
 
+  def register(
```
arkadius, line 29: Let's remove, unused variant with schema as `String`. After your changes, someone could think that the `String` schema is a json schema.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/universal/ParsedSchemaSupport.scala
```diff
@@ -110,53 +111,45 @@ class AvroSchemaSupport(kafkaConfig: KafkaConfig) extends ParsedSchemaSupport[Av
...
+        rawParameter.copy(
+          defaultValue = Some(defaultJsonTemplateFor(schema)),
+          editors = List(JsonTemplateParameterEditor, SpelParameterEditor)
+        ),
+        new AvroSchemaOutputValidator(validationMode).validate(_, schema.cast().rawSchema())
+      )
+    )
+  }
+
+  override def extractParametersForSink(schema: ParsedSchema, restrictedParamNames: Set[ParameterName])(
```
arkadius: Let's rename it to `extractDynamicParametersForSink` to keep this naming consistent with configuration options

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/sink/UniversalKafkaSinkFactory.scala
```diff
@@ -97,146 +103,118 @@ class UniversalKafkaSinkFactory(
...
+        ) =>
+      val validationAgainstSchemaErrors = state.validateParams(Map(sinkValueParamName -> value))
+      FinalResults(
+        context,
+        validationAgainstSchemaErrors,
+        Some(state)
+      )
+  }
+
+  protected def handleTopicWithSchemaWithJsonTemplateEditor(
```
arkadius: Are you still getting what's going on in this class? Because I'm not :) What does `WithJsonTemplateEditor` mean in this context? Does this method handle configuration option when there is a single parameter for value? I see that this pattern matching is enabled even if this option is not turned on. WDYT about having two `PartialFunction` for both options and using one of them depending on configuration option? Will it decompose the problem? Or maybe at least, we could extract some parts of this pattern matching to dedicated classes that would name the scope of this pattern patching that they solve?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schema/AvroSchemaBasedParameter.scala
mateuszkp96: I used the apply method with the NodeId type passed by an implicit parameter

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schema/AvroSchemaBasedParameter.scala
mateuszkp96: I changed it to enforce deterministic ordering, including field ordering, but since we agreed that the field order in the Avro schema should be preserved, this change is no longer necessary

engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/kafkaschemaless/BaseKafkaJsonSchemalessItSpec.scala
mateuszkp96: If you mean a test for schemaless topics, I don't think it's necessary. I removed `sinkRawEditorParamName` for schemaless topics, so the old parameter will be skipped.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/sink/UniversalKafkaSinkFactory.scala
mateuszkp96: I simplified it a little bit. Please take a look

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/typed/AvroSchemaTypeDefinitionExtractor.scala
```diff
@@ -33,6 +33,7 @@ class AvroSchemaTypeDefinitionExtractor(recordUnderlyingType: TypedClass) {
       case Schema.Type.RECORD => {
         val fields = schema.getFields.asScala
           .map(field => field.name() -> typeDefinition(field.schema()))
+          .sortBy(_._1)
```
arkadius: As we discussed, for Avro, we should keep the original order of fields

engine/flink/schemed-kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/schemedkafka/schema/TestSchema.scala
```diff
@@ -1,13 +1,16 @@
...
 import org.apache.avro.generic.{GenericData, GenericRecord}
 import pl.touk.nussknacker.engine.api.validation.ValidationMode
 import pl.touk.nussknacker.engine.schemedkafka.AvroUtils
 import pl.touk.nussknacker.engine.schemedkafka.encode.ToAvroSchemaBasedEncoder
+import pl.touk.nussknacker.engine.schemedkafka.schemaregistry.confluent.ConfluentUtils
 
 trait TestSchema {
-  lazy val schema: Schema = AvroUtils.parseSchema(stringSchema)
+  lazy val schema: Schema              = AvroUtils.parseSchema(stringSchema)
+  lazy val confluentSchema: AvroSchema = ConfluentUtils.convertToAvroSchema(schema)
```
arkadius: Having two schemas next to each other is confusing. I thought about an approach, that we keep in samples this most commonly know type for schemas, which is easiest to build using DSL and to parse - `org.apache.avro.Schema`, and we pass it to `MockConfluentSchemaRegistryClientBuilder.register` where we convert it to `ParsedSchema` just for `MockSchemaRegistryClient.register` purpose.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/sink/UniversalKafkaSinkFactory.scala
```diff
@@ -340,13 +374,24 @@ class UniversalKafkaSinkFactory(
     )
 
     val clientId = s"${TypedNodeDependency[MetaData].extract(dependencies).name}-${preparedTopic.prepared}"
-    val validationMode = if (params.extractUnsafe[Boolean](sinkRawEditorParamName)) {
-      validationModeParamDeclaration.extractValue(params) match {
-        case Some(validationModeString) => extractValidationMode(validationModeString)
-        case None                       => ValidationMode.strict
+    val validationMode = {
+      if (params.isPresent(sinkRawEditorParamName)) {
```
arkadius: It looks like we can use just `extract` without `Unsafe` and avoid direct `isPresent` calling

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/sink/UniversalKafkaSinkFactory.scala
```diff
@@ -340,13 +374,24 @@ class UniversalKafkaSinkFactory(
...
+            case Some(validationModeString) => extractValidationMode(validationModeString)
+            case None                       => ValidationMode.strict
+          }
+        } else {
+          ValidationMode.strict
+        }
+      } else if (params.isPresent(sinkValidationModeParamName)) {
+        validationModeParamDeclaration.extractValue(params) match {
+          case Some(validationModeString) => extractValidationMode(validationModeString)
+          case None                       => ValidationMode.strict
```
arkadius: This is a dead code. If you checked `isPresent`, `extractValue` will always return true.

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/sink/UniversalKafkaSinkFactory.scala
```diff
@@ -340,13 +374,24 @@ class UniversalKafkaSinkFactory(
...
-    val validationMode = if (params.extractUnsafe[Boolean](sinkRawEditorParamName)) {
-      validationModeParamDeclaration.extractValue(params) match {
-        case Some(validationModeString) => extractValidationMode(validationModeString)
-        case None                       => ValidationMode.strict
+    val validationMode = {
+      if (params.isPresent(sinkRawEditorParamName)) {
+        if (params.extractUnsafe[Boolean](sinkRawEditorParamName)) {
+          validationModeParamDeclaration.extractValue(params) match {
+            case Some(validationModeString) => extractValidationMode(validationModeString)
+            case None                       => ValidationMode.strict
```
arkadius: I thought that for raw editor, validation mode is alywas avialable. Isn't it? It is for schemaless case? If so, lets' write a comment that this value is not important

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/testing/SchemalessKafkaJsonTypeTests.scala
```diff
@@ -268,8 +268,8 @@ object SchemalessKafkaJsonTypeTests {
         .emptySink(
           "end",
           "kafka",
-          "Key"                   -> "".spel,
-          "Raw editor"            -> "true".spel,
+          "Key" -> "".spel,
+//          "Raw editor"            -> "true".spel,
```
arkadius: ?

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/confluent/client/MockSchemaRegistryClientBuilder.scala
```diff
@@ -11,50 +14,41 @@ class MockConfluentSchemaRegistryClientBuilder {
...
-    registry.append(RegistryItem(topic, schema, version, isKey))
-    this
+    val avroSchema = ConfluentUtils.convertToAvroSchema(schema)
+    register(topic, avroSchema, version, isKey)
   }
 
   def register(
       topic: String,
-      schema: String,
+      schema: ParsedSchema,
```
arkadius: Let's use Everit schema instead of Confluent's `ParsedSchema` for consistency with the approach that we follow in the `Avro` case and to avoid spreading internal `Confluent` interfaces usage in our code base.  It can be parsed using `JsonSchemaBuilder.parseSchema()`. Take a look at other places where it was used.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
```diff
@@ -1,22 +1,38 @@
...
+import pl.touk.nussknacker.engine.api.Params.ParamExtractionResult
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.util.Implicits.RichScalaMap
 
 final class Params private (val nameToRawValueMap: Map[ParameterName, Any]) extends Serializable {
 
-  def isPresent(name: ParameterName): Boolean = nameToRawValueMap.contains(name)
+  def extractParam[T](name: ParameterName): ParamExtractionResult[T] = {
+    extractValue(name) match {
+      case ParamExtractionResult.Value(value)     => ParamExtractionResult.Value(cast(value))
```
arkadius: Let's make `ParamExtractionResult` a `Functor` - let's add the `map` method

components-api/src/main/scala/pl/touk/nussknacker/engine/api/Params.scala
```diff
@@ -1,22 +1,38 @@
...
+      case ParamExtractionResult.ParamValueIsNone => ParamExtractionResult.ParamValueIsNone
+      case ParamExtractionResult.MissingParam     => ParamExtractionResult.MissingParam
+    }
+  }
 
-  def extract[T](name: ParameterName): Option[T] = {
-    extractValue(name).map(cast[T])
+  def extractRequiredParam[T](name: ParameterName): Option[T] = {
```
arkadius: 1. I have some doubts regarding the `required` word. `Required` is very close to `mandatory`, and we use the `mandatory` word (as an opposition to `optional` word) in the places where a user has to provide some non-empty value - take a look at `MandatoryParameterValidator`. Here is another situation, we as a developer of a component, declared the parameter and we expect the scenario compiler to provide the corresponding value to us. WDYT about using phrases such as `extractDeclaredParam` / `extractDeclaredParamUsafe`?
2. Let's add the same word to `extractOrEvaluateLazyParam` and `extractOrEvaluateLazyParamUnsafe`

arkadius (commented):
After the ADT extraction it looks much better. Some minor comments left

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/testing/SchemalessKafkaJsonTypeTests.scala
mateuszkp96: sorry, I added these changes by accident

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/confluent/client/MockSchemaRegistryClientBuilder.scala
```diff
@@ -44,6 +43,16 @@ class MockConfluentSchemaRegistryClientBuilder {
     client
   }
 
+  private def register(
```
arkadius: This method does one line and extraction of it took 9. Worth it? Let's inline
```
registry.append(RegistryItem(...))
```
in places where it was invoked

arkadius (approved):
One last comment left, please merge after resolving it

PR #8231, opened 2025-06-12
[NU-2154] Standalone Flink live data

mgoworko:
## Describe your changes

This change introduces basic mechanism of synchronizing live data from standalone Flink jobs using the Designer db. This mechanism may be suitable for demonstrational purposes or low/medium traffic environments. It may be a base for future implementation of betten synchronization methods, using for example Redis.

## Checklist before merge


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkScenarioJob.scala
```diff
@@ -40,7 +43,26 @@ class FlinkScenarioJob(modelData: ModelData) {
       env: StreamExecutionEnvironment,
       processListeners: List[ProcessListener],
   ): JobExecutionResult = {
-    val compilerFactory         = new FlinkProcessCompilerDataFactory(modelData, deploymentData, processListeners)
+    val liveDataCollectingListener =
```
mgoworko: It should be a safe change, because nothing changes in Flink jobs when db uploading of results is disabled.

e2e-tests/src/test/scala/pl/touk/nussknacker/DetectLargeTransactionSpec.scala
```diff
@@ -25,6 +37,477 @@ class DetectLargeTransactionSpec extends AnyFreeSpecLike with BaseE2ESpec with M
     eventually {
       val processedTransactions = client.readAllMessagesFromKafka("ProcessedTransactions")
       processedTransactions should equal(largeAmountTransactions)
+      given()
```
mgoworko, line 40: Added assertion of live data in the E2E test using standalone Flink.

build.sbt
```diff
@@ -2039,9 +2040,8 @@ lazy val liveDataCollector = (project in file("designer/live-data-collector"))
     ),
   )
   .dependsOn(
-    deploymentManagerApi % Provided,
```
arkadius, line 2007: Let's keep this dependency explicit even if it is transitively available via `liveDataCollector`. It should be easy to remove this feature if we decide to replace this approach with collector with something else. BTW I missed one thing in the previous PR. The `collector` module shouldn't be nested inside `designer`. We have the strategy for modules organization:
- nested inside the `designer` module - things that are not operational, only needed during designing the scenario
- nested inside the `engine` module - things that are used during runtime
- on the root level - things that are shared between the designer and operational part (scenario runtime).

Folllowing this approach, the `collector` should be on the root level, the same as `scenario-compiler`

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -39,8 +40,17 @@ object ModelConfig {
     final case class Enabled(
         maxNumberOfRecords: Int,
         throughputTimeWindowInSeconds: Int,
+        dbUploader: Option[DbUploader],
     ) extends LiveDataPreviewMode
 
+    final case class DbUploader(
```
arkadius: `upload` has a very strong connotation with files. We have a very similar mechanism that have a ready set of words for us - the metrics. In the metrics domain, metrics are registered in the registry and then they are reported to some service responsible for exposing them. Based on that we can call it the same manner. Also, when you pick the noun that will be exposed in the configuration, always take a look from the perspective, how you would descibe it in the documenation (even if you don't produce it yet) or during some casual talk with a user. You would rather use sentences such as "To configure the reporting mechanism/storage for live data, you should add ... configuration entry". The noun that was exposed is dedicated for "normal" users (administrators) instead of developers. It describe the mechanism instead of some syntetics role/class in the code. LSS - let's think about the domain model that we want to expose and based on that, let's pick names instead of thinking of implemenatation and then exposing the implementation details in the configuration.

 Take a look how the best do this: https://github.com/apache/flink/blob/master/docs/content/docs/deployment/config.md You can find nouns such as "metrics", "execution", "logging", "checkpointing"

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+          dbSchema = dbUploader.dbSchema,
+        )
+      )
+      .addSink(new EmptySink[String])
+  }
+
+  private class OneShotSource extends SourceFunction[String] {
+
+    override def run(ctx: SourceFunction.SourceContext[String]): Unit = {
+      // emit once and sleep forever
```
arkadius: 1. Why sleep forever?  BTW, it won't sleep forever but only until java process will be interrupted. Is it ok?
2. Let's use a new, `Source` api instead of a legacy one.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+import scala.util.{Failure, Success, Try}
+
+object PeriodicLiveDataUploader {
+
+  def register(
+      env: StreamExecutionEnvironment,
+      processIdWithName: ProcessIdWithName,
+      dbUploader: DbUploader,
+  ): Unit = {
+    env
```
arkadius: Can you write a few words of comment why you implemented it as a flink data stream processing pipeline? e.g. why is it not a `Thread`?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+      dbPassword: String,
+      dbSchema: String,
+  ) extends KeyedProcessFunction[String, String, String] {
+
+    private val logger = LoggerFactory.getLogger(getClass)
+
+    @transient private var connection: Connection = _
+
+    override def open(openContext: OpenContext): Unit = {
+      Class.forName("org.postgresql.Driver")
```
arkadius: It looks like a mistake (dead code).

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+          dbUrl = dbUploader.dbUrl,
+          dbUser = dbUploader.dbUser,
+          dbPassword = dbUploader.dbPassword,
+          dbSchema = dbUploader.dbSchema,
+        )
+      )
+      .addSink(new EmptySink[String])
+  }
+
+  private class OneShotSource extends SourceFunction[String] {
```
arkadius: Maybe let's emphasise the fact that this source lasts forever. One-shot sounds like it does something once and finishes its job.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    override def onTimer(
+        timestamp: Long,
+        ctx: KeyedProcessFunction[String, String, String]#OnTimerContext,
+        out: Collector[String]
+    ): Unit = {
+      val liveDataOpt = LiveDataCollectingListenerHolder.getLiveDataPreview(processIdWithName.name)
+      Try {
+        if (connection == null) connection = DriverManager.getConnection(dbUrl, dbUser, dbPassword)
```
arkadius: Why it doesn't happen in `open()`? It is better to fail job during startup instead of after a while.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+        connection.setSchema(dbSchema)
+        val insertStatement = connection.prepareStatement(
```
arkadius: The purpose of prepared statements is to create them once, and reuse many times with different data.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+    override def open(openContext: OpenContext): Unit = {
+      Class.forName("org.postgresql.Driver")
+    }
+
+    override def processElement(
+        value: String,
+        ctx: KeyedProcessFunction[String, String, String]#Context,
+        out: Collector[String]
+    ): Unit = {
+      val ts = ctx.timerService.currentProcessingTime + 5000
```
arkadius: magic number

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+            insertStatement.setNull(3, java.sql.Types.VARCHAR)
+        }
+        insertStatement.setLong(4, Instant.now.getEpochSecond)
+        insertStatement.executeUpdate()
+      } match {
+        case Success(_) =>
+          logger.debug("Uploaded scenario live data")
+        case Failure(exception) =>
+          connection = null
+          logger.error("Could not update scenario live data", exception)
```
arkadius: Can you extend this message with what we are planning to do with this problem? I see that you would:
- recreate the connection without closing the previous one. 
- skip failing entry

Is it an intended behaviour?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+          Option(exception.getCause)
+            .foreach(cause => logger.error("Could not update scenario live data with cause", cause))
+      }
+
+      ctx.timerService.registerProcessingTimeTimer(
+        timestamp + intervalSeconds * 1000
```
arkadius: magin number

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,132 @@
...
+    ): Unit = {
+      val liveDataOpt = LiveDataCollectingListenerHolder.getLiveDataPreview(processIdWithName.name)
+      Try {
+        if (connection == null) connection = DriverManager.getConnection(dbUrl, dbUser, dbPassword)
+        connection.setSchema(dbSchema)
+        val insertStatement = connection.prepareStatement(
+          """
+            |INSERT INTO live_data (scenario_id, collector_id, live_data, updated_at)
+            |VALUES (?, ?, ?, ?)
+            |ON CONFLICT (scenario_id, collector_id) DO UPDATE
```
arkadius: Will it work with hsqldb?

examples/installation/docker-compose.yml
```diff
@@ -72,6 +72,8 @@ services:
   postgres:
     image: postgres:13
     restart: unless-stopped
+    ports:
```
arkadius: This is our official, installation setup. Are you sure that you want to expose this port for every, production-ready installation?

nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -28,6 +28,22 @@ scenarioTypes {
           componentId: sink-kafka
         }
       }
+      allowEndingScenarioWithoutSink: true
```
arkadius: This is our official, minimal configuration. We don't plan to enable it for everyone

nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -28,6 +28,22 @@ scenarioTypes {
...
+      liveDataPreview {
```
arkadius: The same here.

nussknacker-dist/src/universal/conf/application.conf
```diff
@@ -28,6 +28,22 @@ scenarioTypes {
...
+        enabled: true
+        maxNumberOfSamples: 20
```
arkadius: We can hardcode this values as a default values for `LiveDataPreviewMode.Enabled`. Thanks to that, during configuration, only enabling will be necessary.

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/LiveDataEntityFactory.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  import profile.apiWithEnforcedSchema._
+
+  val flinkLiveDataTable: LTableQuery[LiveDataEntityFactory#LiveDataEntity] = LTableQuery(
+    new LiveDataEntity(_)
+  )
+
+  class LiveDataEntity(tag: Tag) extends TableWithSchema[LiveDataEntityData](tag, "live_data") {
+
+    def scenarioId: Rep[ProcessId] = column[ProcessId]("scenario_id")
```
arkadius, line 18: For batch processing, we should have a (collectorId, activityId) as a primary key. The `activityId` for the deployment is a Flink's jobid. Let's leave a comment that currently, this mechanism is limited to the streaming scenarios.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+      .groupBy(_._1)
+      .mapValuesNow { entries =>
+        val allValues = entries.flatMap(_._2)
+        allValues
+          .sortBy(getTimestamp)
+          .takeRight(maxNumberOfSamples)
+      }
+  }
+
+  private def removeOldEntries(
```
arkadius: 1. What will happen if someone redeploy the scenario during `uploadIntervalInSeconds`? I guess that you will aggregate results from the last job with the new one? Is it ok?
2. For mini cluster, we keep results after scenario is stopped. For the standalone Flink, not?

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkMiniClusterScenarioJobRunner.scala
```diff
@@ -79,11 +60,8 @@ class FlinkMiniClusterScenarioJobRunner(
 
   override def liveDataPreviewSupport: LiveDataPreviewSupport = {
```
arkadius: Why `liveDataPreviewSupport` is a part of `ScenarioJobRunner` if it is not related with it? I don't see any occurrencces of this method usage from the job runner.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkScenarioJob.scala
```diff
@@ -40,7 +43,26 @@ class FlinkScenarioJob(modelData: ModelData) {
...
+      modelData.modelConfig.liveDataPreviewMode match {
+        case LiveDataPreviewMode.Disabled =>
+          None
+        case LiveDataPreviewMode.Enabled(maxNumberOfSamples, throughputTimeWindowInSeconds, dbUploaderOpt) =>
+          val processIdWithName = ProcessIdWithName(processVersion.processId, processVersion.processName)
+          dbUploaderOpt.foreach(PeriodicLiveDataUploader.register(env, processIdWithName, _))
+          Some(
+            LiveDataCollectingListenerHolder.createListenerFor(
```
arkadius: When do you close the allocated resources for jobs that have been run on a remote flink cluster?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+      processIdWithName: ProcessIdWithName,
+      uploadIntervalInSeconds: Long,
+  ) = {
+    run(
+      flinkLiveDataTable
+        .filter(_.scenarioId === processIdWithName.id)
+        .filter(
+          _.updatedAt < Instant.now.getEpochSecond - uploadIntervalInSeconds - 5
+        ) // Drop data older than interval + 5 seconds
```
arkadius: Why drop? why older than interval + 5 seconds? why not 0 seconds or 1 minute? (magic number)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+      case None => CollectedLiveData.empty
+    }
+  }
+
+  def aggregate(
+      data: List[Map[NodeTransition, LiveDataForNodeTransition]],
+      maxNumberOfSamples: Int,
+  ): Map[NodeTransition, LiveDataForNodeTransition] = {
+    data.flatten
+      .groupBy(_._1)
```
arkadius: You can use `toGroupedMap`. Thanks to that we avoid this meaningless `_1`, `_2`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
arkadius: WDYT about an approach, that we collect live data for (jobid, collectorid) and during reading we take the last deployment and return result for jobid = deploymentid? Thanks to that we avoid weak assumptions, magic numbers etc.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkMiniClusterScenarioJobRunner.scala
mgoworko: Is is a method defined in `trait FlinkScenarioJobRunner`. This trait has 2 implementations, for minicluster and remote Flink. Those 2 implementations handle the different ways of running Flink scenarios in those 2 cases. 

I think it is an appropriate place for the definition, how live data is handled in the Flink scenario. It is used by FlinkDeploymentManager (which is common for minicluster and standalone, uses the FlinkScenarioJobRunner)

build.sbt
mgoworko: - I removed this dependency. The `liveDataCollector` no longer depends on `deploymentManagerApi`. In this PR I moved the LiveData data structures from `deploymentManagerApi` module to the `liveDataCollector`, so this dependency is no longer needed.

- liveDataCollector module moved to root level

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
mgoworko: I changed it to 

```
    sealed trait LiveDataStorage

    object LiveDataStorage {

      object DesignerJvm extends LiveDataStorage

      final case class DesignerDb(
          uploadIntervalInSeconds: Int,
          url: String,
          user: String,
          password: String,
          schema: String,
      ) extends LiveDataStorage

    }
 ```

And 
```
      liveDataPreview {              
        enabled: true               
        maxNumberOfSamples: 20      
        throughputTimeWindowInSeconds: 60
        storage {  // This section must be configured for standalone FLink
          type: "DESIGNER_DB" // The only storage type available for now
           // Small `uploadIntervalInSeconds` values make the live data preview look more smooth and similar to the Flink MiniCluster version, bigger values are easier on the db
          uploadIntervalInSeconds: 3 
           // Designer db url, the same as the one provided for the Nussknacker app
          url: "jdbc:postgresql://localhost:5432/test?loggerLevel=OFF"  
          user: "NU"       // db user
          password: "NU"   // db password
          schema: "public" // db schema must be always provided, even if it is "public"
        }
```

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 7 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: Added comment, that it is needed to load the driver (without it, the runner screams about missing jdbc driver). Also modified to work for hsqldb.

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 5 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: I refactored it to use the new Flink API for source and sink. I also use the approach with starting a Thread (instead of Flink timers), so the "sleep forever" is no longer needed.

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 8 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: Refactored to use the new API and name changed to `EmitOnceSource`. The new source emits single element and no longer contains "sleep forever" logic, so the name is now correct IMO. Only the next element in the Flink pipeline starts the thread, that works as long as the flink process.

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 9 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: It now happens in `open()`. Refactored to start a thread from `open` method, instead of using Flink timers. So the execution is always on the same thread, so we can reuse the connection.

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 10 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: Now created once (and recreated, if there is some kind of db problem and previous connection is closed)

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 6 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: Added comment in code. It is done that way in order to be run alongside the scenario - and its lifecycle managed by Flink together with the scenario

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 11 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: removed, source refactored

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 12 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: - extended error message
- the failed entry is skipped by design - we do not want to interrupt the scenario (live data is a cherry on top, it is not crucial and should not interrupt the core execution of scenario
- I refactored the Flink pipeline used for uploading - connection is now closed and recreated on next try

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 14 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: Refactored to work with HSQLDB (separated statements), tested locally

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala (hunk 13 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala]: It is now `Thread.sleep(intervalSeconds * 1000)` inside the thread.

File: examples/installation/docker-compose.yml (hunk 15 above)
Comment (mgoworko) [reply examples/installation/docker-compose.yml]: removed

File: nussknacker-dist/src/universal/conf/application.conf (hunk 17 above)
Comment (mgoworko) [reply nussknacker-dist/src/universal/conf/application.conf]: Removed, added as an additional config file for E2E tests

File: nussknacker-dist/src/universal/conf/application.conf (hunk 16 above)
Comment (mgoworko) [reply nussknacker-dist/src/universal/conf/application.conf]: removed

File: designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala (hunk 23 above)
Comment (mgoworko) [reply designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala]: I changed it. Now the live data for a scenario is cleaned in the db when scenario is deployed. So it works the same as for MiniCluster and JVM version.

File: designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala (hunk 20 above)
Comment (mgoworko) [reply designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala]: I think that working on jobId would require analysis and perhaps some refactors. It would require some additional requests to the DB and/or flink and/or modifying data structures to make the jobId available in all required places. One of the options would be to use the "lastDeployAction", which is deprecated and scheduled to be removed. So I would stay with processId for now.

I refactored it to clean the live data in the db when the scenario is deployed, described in other comment.

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkScenarioJob.scala (hunk 22 above)
Comment (mgoworko) [reply engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/FlinkScenarioJob.scala]: The uploader db connections are closed. The LiveData storage on the JVM is cleaned the same way as in the designed (caffeine cache with auto cleaning, cleaning on new deployment of the same scenario).

File: engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/EmitOnceSource.scala [hunk 25]
```diff
@@ -0,0 +1,70 @@
+package pl.touk.nussknacker.engine.process.livedata
+
+import org.apache.flink.api.connector.source._
+import org.apache.flink.core.io.{InputStatus, SimpleVersionedSerializer}
+
+import java.util.Collections
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.atomic.AtomicBoolean
+
+class EmitOnceSource extends Source[String, DummySplit, Void] {
```
arkadius: @raphaelsolarski can you take a look at this part of the change? You better know Flink Source API

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
arkadius: It won't fail during job startup, because connection creation is done in the `Thread`.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,159 @@
...
+  // Flink runs it alongside the scenario. It is started with the scenario and stopped when scenario is stopped.
+  def register(
+      env: StreamExecutionEnvironment,
+      processIdWithName: ProcessIdWithName,
+      storage: DesignerDb,
+  ): Unit = {
+    env
+      .fromSource(new EmitOnceSource, WatermarkStrategy.noWatermarks(), "live-data-uploader")
+      .keyBy((_: String) => "live-data")
+      .process(
```
arkadius: Why source emitting one event -> processor processing it in the loop instead of
source emitting an event every 1 second -> processor processing it
? It would be more straightforward

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,159 @@
...
+  override def close(): Unit = {
+    running = false
+    if (updaterThread != null) {
+      updaterThread.interrupt()
+      updaterThread.join()
+    }
+    if (connection != null) connection.close()
+  }
+
+  private def prepareConnection(): Unit = {
```
arkadius: It doesn't prepare connection, it does more

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
arkadius: I can't see in the log message that attempt of upload will be skipped, connection will be retried and another attempt will be done in 1 second. Showing only an error might be worrying and force person looking at logs to take a look at code

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,159 @@
...
+      storage: DesignerDb,
+  ): Unit = {
+    env
+      .fromSource(new EmitOnceSource, WatermarkStrategy.noWatermarks(), "live-data-uploader")
+      .keyBy((_: String) => "live-data")
+      .process(
+        new PeriodicLiveDataUploader(
+          processIdWithName = processIdWithName,
+          intervalSeconds = storage.uploadIntervalInSeconds,
+          dbUrl = storage.url,
```
arkadius: It would be nice to use the configuration from `DesignerConfig.db` as initial values for that. I guess that it won't be easy to achieve, looking at the current state of the code around this, so let's add a TODO at least

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/LiveDataEntityFactory.scala
arkadius: What about this?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
arkadius: It is only one additional db query execution needed to take the last deployment. I thought that it is natural to use `ScenarioActivityRepository` for that. I saw a comment next to `ScenarioActionRepository` left by you. I'm a little bit confused. It looks like you introduced some domain model changes around actions/activites/deployments and now, it is too complicated for you to use these changed concepts? :) 

I think that it would cost us more effort to migrate this solution to be deployment-aware in the future instead of doing this cleaner now.

Consider that our bussiness has a new requirement: I want to show the last live data snapshot for the deployment which was assigned to the scenario version that I've just switched to. Consider that a new developer has to extend your solution to make it possible. It won't be easy now. It means that we are introducing an architectural debt. The solution fit only current requirements and is not extensible.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkMiniClusterScenarioJobRunner.scala
arkadius: `ScenarioJobRunner` (as the name suggests) is responsible for running the job. We don't want to tightly connect job running with live data mechanism. It should be easy to remove live data mechanism without breaking a core feature - scenario running mechnism.

I see that you need some `LiveDataCollector` that can be passed by the runner to the flink job, where the job can register it in the streaming environment and `FlinkDeploymentManager` can use it to provide that data for the UI.
 
You've mentioned that you have two implementations depending on the runner.  I see something different - you have some validation logic which checks which implementation can be used in which context. It is a kind of factory logic that should be added in the place where other factory logic happens - in the `FlinkDeploymentManagerProvider.createDeploymentManager`

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -39,18 +40,52 @@ object ModelConfig {
...
+            case "DESIGNER_DB" =>
+              LiveDataStorage.DesignerDb(
+                uploadIntervalInSeconds = config.getInt("liveDataPreview.storage.uploadIntervalInSeconds"),
+                url = config.getString("liveDataPreview.storage.url"),
+                user = config.getString("liveDataPreview.storage.user"),
+                password = config.getString("liveDataPreview.storage.password"),
+                schema = config.getString("liveDataPreview.storage.schema"),
+              )
+            case other =>
+              throw new IllegalStateException(s"Unknown live data storage type [$other]")
```
arkadius: Let's use some other exception type. We rather use this exceptions in situation when we know that is some kind of bug introduced by us (a kind of assertion). In this situation, user just used wrong/unsupported value - it sounds like a normal, very often situation.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
```diff
@@ -0,0 +1,149 @@
...
+        liveData     <- collectedLiveDataDecoder.decodeJson(liveDataJson).left.map(_.message)
+      } yield liveData
+    }).map(
+      _.reverse.foldLeft(Right(Nil): Either[String, List[CollectedLiveData]]) {
+        case (Right(acc), Right(result)) => Right(result :: acc)
+        case (Left(err), _)              => Left(err)
+        case (_, Left(err))              => Left(err)
+      }
+    ).map(_.map {
+      aggregate(_, maxNumberOfSamples)
```
arkadius: Can we more explicitly show in the code, what part of information we have to aggregate to make it full?
I mean it might be many different things: some partial data for given periods, etc. 
Without this, at first glance I'm not sure if there is no such bug as:
- double-counting the same thing twice
- occasionally skipping some part of the information

 I think about:
- wrapping `collectedLiveData: List[CollectedLiveData]` with `case class LiveDataCollectedBySingleCollector()` or just renaming this parameter to `liveDataCollectedByCollectors` 
- logging in debug level, information from how many collectors we aggregated

BTW. What about the situation, when one collector will be hard-terminated and replaced by another collector on the different task manager?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,159 @@
...
+    // Looks like unused code, but needed to load the driver
+    if (dbUrl.startsWith("jdbc:postgresql:"))
+      Class.forName("org.postgresql.Driver")
+    else if (dbUrl.startsWith("jdbc:hsqldb:"))
+      Class.forName("org.hsqldb.jdbc.JDBCDriver")
+
+    running = true
+    updaterThread = new Thread(() => {
+      while (running) {
+        Try {
```
arkadius: Let's extract the whole body of while to a separate method

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,159 @@
...
+          if (connection == null) prepareConnection()
+
+          statement.setLong(1, processIdWithName.id.value)
```
arkadius: Let's extract statement execution into a separate method. Is it ok for you if another developer, instead of creating a repository, would inline the code talking with the infrastructure into the body of a REST endpoint implementation?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: I've already tried two different approaches a few weeks ago. At first I tries passing the config values in code, and the using references in config files. It is impossible in current architecture without some huge refactors, added todo

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: `PeriodicLiveDataUploader` refactored and split into 7 well-named private methods, which are then composed in the thread definition

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: Refactored as:

```
      while (running) {
        Try(uploadLiveData()).recover(handleLiveDataUploadFailure(_))
        sleepUntilNextUploadTimeUnlessInterrupted()
      }
```

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
mgoworko: => `IllegalArgumentException`

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: Entire class refactored and split into multiple private methods.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: I tried to do it the way You described. It caused one significant problem - executions were performed by different threads, and the db connection and statement are not thread-safe. On cloud Flink there were problems especially with setting schema and using it in a prepared statement.

So:
- Starting as Flink source provides Flink lifecycle control - it starts with scenario and ends when it ends. I tried to start it as a thread outside of Flink scenario, but I was not successful, that standalone thread was almost immediately interrupted on Flink and did not continue to run alongside scenario.
- Running in a loop in a single thread gives full control over execution and eliminates problems with thread-safety.

In one of the previous versions Flink timers were already used. But it required creating connection and statement on each upload. I can refactor it to use Flink timers again, if you have any suggestion how better handle multithreading in that case.

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkMiniClusterScenarioJobRunner.scala
mgoworko: done

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
mgoworko: => `s"Could not upload scenario live data to the db. This upload is skipped and will be retried on next scheduled upload time in $intervalSeconds s. If uploading continues to fail, then please check the detailed reason of failure in logs below and look for configuration issues."`

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/runner/PeriodicLiveDataUploader.scala
mgoworko: I added preparing initial connection and statement before the thread starts (it is then recreated as part of error recovery, if upload fails)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
mgoworko: Ok. I managed to add columns "deployment_id" (which is an if of the deployment Activity) and "external_deployment_id" (which is Flink job id in this context).

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
mgoworko: The repository logic is refactored with for comprehension and better names. I moved the aggregation logic to the CollectedLiveData companion object. With changed some names to better describe what is happening there. We take `maxNumberOfSamples` latest samples, and the overall timestamp is the timestamp of latest collected live data (I changed that, was oldest, but that would cause problems related to the BTW comment)

About BTW:
- the data will be cleaned from the db on next deployment
- until then, it will be present in the db, but aggregation will take latest samples, so after there is enough newer samples, it will ignore older ones

For now, all live data is cleaned for the scenario on deploy. But there is now information about deployment ids (both Nu and Flink job id), so in the future it will be possible to have live data per deployed version (but not done in this PR, I don't want it to get any bigger)

Added log about aggregating

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
arkadius: external_deployment_id is always the same as deployment_id, It is some legacy which was added because we didn't know that we can set job id. Basically it is a thing that we should remove eventually

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,189 @@
...
+  ): Unit = ()
+
+  override def close(): Unit = {
+    running = false
+    if (updaterThread != null) {
+      updaterThread.interrupt()
+      updaterThread.join()
+    }
+    if (connection != null) connection.close()
+    if (statement != null) statement.close()
```
arkadius: It would be better to close those things in the reversed order: close statement and then close connection

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkScenarioJobRunner.scala
```diff
@@ -12,6 +12,4 @@ trait FlinkScenarioJobRunner {
       savepointPathOpt: Option[String]
   ): Future[Option[JobID]]
 
-  def liveDataPreviewSupport: LiveDataPreviewSupport
```
arkadius, line 15: You left unused import

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
arkadius: I thought that flink's operators are runned in a single thread. I feal that we are raping Flink ;-) but also I don't want to invest too much effort into investigation how to do this in the more idiomatic way.  
Please add this explanation as a comment next to `PeriodicLiveDataUploader.register`, maybe we will get back to these considerations someday.

arkadius (approved):

engine/flink/management/src/main/scala/pl/touk/nussknacker/engine/management/jobrunner/FlinkScenarioJobRunner.scala
mgoworko: removed

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: done

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
mgoworko: Added comment 

```
// This class runs the live data uploader as Flink function. But once started, it performs uploads in a loop in a single thread.
// It proved difficult to use Flink timers or event generator source with intervals. Db connection and statement are not thread-safe.
// It would need more thorough investigation to do in a more Flink-style way, for now:
//  - Starting as Flink source provides Flink lifecycle control. It starts with scenario and ends when it ends. Standalone thread outside the Flink scenario would be immediately interrupted on Flink.
//  - Running in a loop in a single thread gives full control over execution and eliminates problems with thread-safety.
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/db/entity/LiveDataEntityFactory.scala
mgoworko: We have primary key 

`ADD CONSTRAINT pk_scenario_activity_collector_ids PRIMARY KEY ("scenario_id", "deployment_id", "external_deployment_id", "collector_id")`

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,194 @@
...
+  // Uploading live data to external storage (in order to make it available for the Designer) is done as a Flink pipeline.
+  // Flink runs it alongside the scenario. It is started with the scenario and stopped when scenario is stopped.
+  def register(
+      env: StreamExecutionEnvironment,
+      processIdWithName: ProcessIdWithName,
+      deploymentData: DeploymentData,
+      storage: DesignerDb,
+  ): Unit = {
+    env
+      .fromSource(new EmitOnceSource, WatermarkStrategy.noWatermarks(), "live-data-uploader")
```
raphaelsolarski: 1. mby we can do whole logic inside source reader ended with some no-op processor?
2. we don't need to schedule additional daemon thread for periodic operations - we have dedicated reader/operator chain thread is for our usage there or we can use some Scheduled
3. I'm not sure whether the solution will work when deployed to multiple taskmanager flink - now we have to be sure that we have exactly one `PeriodicLiveDataUploader` that gathers samples from `LiveDataCollectingListenerHolder`

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
```diff
@@ -0,0 +1,194 @@
...
+  @transient private var statement: PreparedStatement = _
+
+  private lazy val jobId: String = getRuntimeContext.getJobInfo.getJobId.toHexString
+
+  override def open(openContext: OpenContext): Unit = {
+    running = true
+    loadJdbcDriver()
+    prepareConnection()
+    prepareStatement()
+    updaterThread = new Thread(() => {
```
raphaelsolarski: side note: if we have to use thread directly for some reason imo it should be marked as `daemon` thread.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
arkadius: Ad 3. We aggregate data across different TaskManagers - we have a dedicated column "collector_id". It is a little bit tricky because we don't do it per slot but for a virtual entity (collector) which basically is 1-to-1 to TM, because we use statics for the communication. I mentioned @mgoworko about potential problems in this comment: https://github.com/TouK/nussknacker/pull/8231#discussion_r2177261612 but it looks like we eventually work correclty so this trick is IMO acceptable
Ad 1&2 it sounds more flink-idiomatic :heart:

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/livedata/LiveDataRepository.scala
mgoworko: external_deployment_id column removed

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/livedata/PeriodicLiveDataUploader.scala
arkadius: Ad 3. We discussed it with @raphaelsolarski offline and now I'm getting where is the problem. Indeed, we don't have guaranty that all data will be uploaded, because we can't assure that this uploading job will be spawned on the same TMs where the job processing data is spawned.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,87 @@
...
+          })
+        )
+      case None =>
+        logger.error(
+          "Live data uploader cannot be started, because the DeploymentId is not defined or has non-UUID underlying value"
+        )
+    }
+  }
+
+  def startPeriodicLiveDataUploader(
```
mgoworko: - when some event is captured by LiveDataCollecting listener (something happens in the scenario and it is recorded by LiveData mechanism), then it is checked whether or not the db uploading is already started and working
- if in that situation db uploading is not working, then it is started, as before with configurable interval, but using `ScheduledExecutorService`
- inactivity (lack of events collected by LiveDataCollectingListener) causes the db uploading to stop (with configurable inactivity timeout)

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+      .newBuilder()
+      .expireAfterAccess(java.time.Duration.ofHours(1))
+      .build[
+        CacheKey,
+        String
+      ]() // Should be [(ProcessIdWithName, DeploymentId), Unit], but Scala 2.12 does not accept tuples and Unit
+
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  def getExistingOrStartLiveDataUploader(
```
arkadius: get method returning Unit? Interesting... :)

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+          })
+        )
+      case None =>
+        logger.error(
+          "Live data uploader cannot be started, because the DeploymentId is not defined or has non-UUID underlying value"
+        )
+    }
+  }
+
+  def startPeriodicLiveDataUploader(
```
arkadius: It can be private

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+import java.time.Instant
+import java.util.concurrent.{Executors, ScheduledExecutorService, ScheduledFuture, TimeUnit}
+import scala.compat.java8.FunctionConverters.asJavaFunction
+
+private[livedata] object LiveDataUploaderHolder {
+
+  private val activeUploaders =
+    Caffeine
+      .newBuilder()
+      .expireAfterAccess(java.time.Duration.ofHours(1))
```
arkadius: Why do we need this cache with expiration time if you have the uploader stopping logic based on `lastUpdatedAt`. How this hardcoded time is related to `uploaderInactivityTimeoutInSeconds`?

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+      .build[
+        CacheKey,
+        String
+      ]() // Should be [(ProcessIdWithName, DeploymentId), Unit], but Scala 2.12 does not accept tuples and Unit
```
arkadius: You can declare `Unit.type` and return `Unit`.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    def uploadLiveData(): Unit = {
+      LiveDataCollectingListenerHolder
+        .storageOpt(processIdWithName.name)
+        .foreach { storage =>
+          val shouldStop =
+            storage.getLastUpdatedAt < Instant.now.getEpochSecond - config.uploaderInactivityTimeoutInSeconds
+          if (shouldStop) {
+            logger.info(s"Stopping live data uploader for scenario $processIdWithName because of inactivity")
+            scheduledTask.foreach(_.cancel(true))
```
arkadius: I'd feel better if these three things would be closed in:
```
try { first.close() } finally { second.close() ...
```
manner. I don't know all situations when they can throw an Exception but leaving unclosed resources in such rare situation sounds problematic

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+            scheduler.shutdownNow()
+            activeUploaders.invalidate(CacheKey(processIdWithName, deploymentId))
+            uploader.close()
```
arkadius: We can create one uploader and one scheduler and reuse it between different deployments on the same task manager. With the current solution, you create many db connections, many threa pools and many threads, we can avoid that.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  def startPeriodicLiveDataUploader(
+      processIdWithName: ProcessIdWithName,
+      deploymentId: DeploymentId,
+      config: LiveDataUploaderConfig,
+  ): Unit = {
+    logger.info(s"Starting live data uploader for scenario $processIdWithName with interval ${config.intervalSeconds}s")
+    val uploader                                  = new LiveDataUploader(config)
+    val scheduler: ScheduledExecutorService       = Executors.newSingleThreadScheduledExecutor()
```
arkadius: Let's name this executor in order to have info in thead dump/logs what is the purpose of the thread

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+    var scheduledTask: Option[ScheduledFuture[_]] = None
+
+    def uploadLiveData(): Unit = {
+      LiveDataCollectingListenerHolder
```
arkadius: Let's add a debug log to monitor if uploading thread is running or not

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+              processIdWithName = processIdWithName,
+              deploymentId = deploymentId,
+              collectedLiveData = storage.getLiveData
+            )
+          }
+        }
+    }
+
+    scheduledTask = Some(
+      scheduler.scheduleAtFixedRate(() => uploadLiveData(), 0, config.intervalSeconds, TimeUnit.SECONDS)
```
arkadius, line 90: Let's add an info log that we spawn a thread - similar to the log during closing to easy match both entries

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
```diff
@@ -0,0 +1,86 @@
...
+    var scheduledTask: Option[ScheduledFuture[_]] = None
+
+    def uploadLiveData(): Unit = {
+      LiveDataCollectingListenerHolder
+        .storageOpt(processIdWithName.name)
+        .foreach { storage =>
+          val shouldStop =
+            storage.getLastUpdatedAt < Instant.now.getEpochSecond - config.uploaderInactivityTimeoutInSeconds
+          if (shouldStop) {
+            logger.info(s"Stopping live data uploader for scenario $processIdWithName because of inactivity")
```
arkadius: Let's add `deploymentId` to the log because you create uploader for each deployment, not for scenario

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
raphaelsolarski: can we do it (sharing uploader among jobs on the same taskmanager) without placing the uploader code in flink `lib` dir (to use AppClassloader)?

arkadius: I don't know. You tell me :) I wonder how `SharedKafkaProducer` works if it is not loaded with `AppClassloader`. The entire solution is based on a shared static context - take a look at that we have a map in `LiveDataCollectingListenerHolder.activeUploaders` instead of just one uploader. I guess that all jobs that are run on the same TM reuse this class, and it is the reason why there is a map instead of a single uploader. Am I right?

.run/NussknackerApp.run.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="NussknackerApp" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="corretto-11" />
+    <option name="ALTERNATIVE_JRE_PATH" value="corretto-17" />
```
lciolecki, line 3: Are you sure about this? I suposse that has been change in rsr PR according Java 17?

build.sbt
```diff
@@ -1995,7 +1996,7 @@ lazy val deploymentManagerApi = (project in file("designer/deployment-manager-ap
   )
   .dependsOn(extensionsApi, testUtils % Test)
 
-lazy val liveDataCollector = (project in file("designer/live-data-collector"))
+lazy val liveDataCollector = (project in file("live-data-collector"))
```
lciolecki, line 1999: Why it is no more a part of designer?

arkadius: https://github.com/TouK/nussknacker/pull/8231#discussion_r2151857598

.run/NussknackerApp.run.xml
arkadius: Currently, our main supported java version is java 17.

mgoworko: It just wasn't changed in the run configurations in the repo, my IDE changed it when I executed NussknackerApp

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
mgoworko: It seems that classloaders are isolated between jobs. On standalone Flink the cache will only contain single active uploader.

In the team that handles the production deployment we started to brainstorm what needs to be done in order to run it for 100s of scenarios. In this basic solution for cloud/demo I think it may stay this way (separate uploader per scenario) and be later improved when clients need it.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
mgoworko: I changed the value to be the handle for stopping the uploader. Uploaders are stopped in shutdown hook.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
mgoworko: Removed expiration time, it does not apply here

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
mgoworko: Changed to `ensureLiveDataUploaderIsActive`

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
arkadius: We discussed it with @raphaelsolarski offline and this static context is not shared between jobs, because the same classes created by different jobs have different memory area.
Based on this, in order to reuse db connection and threads, we'd need to extract a separate lib and add it to flink libs. I think that for now we can ignore this problem, and leave a TODO as a potential direction of changes.
It also implies that we don't need to have a Map scenarioId -> something because we can be sure that this holder is used only by one job, for one scenario. WDYTAT @mgoworko ?

mgoworko: I added todo comment. 

I this we should still have some kind of map (at the moment it is caffeine cache). My argument is, that from the purely JVM point of view, this code should work for multiple scenarios at the same time. It's the Flink design details, that prevent it from working that way. But this project module is not only Flink-specific. It might be reused for other DM's, for example RR (we might need to do that soon if client requests it). Changing the type from caffeine cache (or any kind of map) to single value represented by Option/Atomic/sth similar, will make it not work correctly in all other non-Flink contexts.

designer/server/src/main/resources/db/migration/postgres/V1_066__CreateLiveDataTable.sql
```diff
@@ -0,0 +1,11 @@
+CREATE UNLOGGED TABLE "live_data"
```
mgoworko, line 1: @piotrp I applied the suggestion of making the Postgres table UNLOGGED. I think we can afford that. It is not critical, it does not have to be restored in case of failures and it could benefit from performance optimizations - perfect case for UNLOGGED.

live-data-collector/src/main/scala/pl/touk/nussknacker/engine/livedata/LiveDataUploaderHolder.scala
arkadius: I don't like when we break the YAGNI rule. It caused a lot of problems in the past, but let it be ;)

arkadius (approved):
Good job

PR #8228, opened 2025-06-11
Fix sorting by status in component usages view

Cichan-mcc:
## Describe your changes

Currently when we are in the component usages view sorting by scenario status does not work properly.

bad

The column contains custom objects with icons and it lacked a comparator. After adding that we sort the column by the status name.

ok

## Checklist before merge
- [ ] Related issue ID is placed at the beginning of PR title in \[brackets\] (can be GH issue or Nu Jira issue)
- [ ] Code is cleaned from temporary changes and commented out lines
- [ ] Parts of the code that are not easy to understand are documented in the code
- [ ] Changes are covered by automated tests
- [ ] Showcase in dev-application.conf added to demonstrate the feature
- [ ] Documentation added or updated
- [ ] Added entry in _Changelog.md_ describing the change from the perspective of a public distribution user
- [ ] Added _MigrationGuide.md_ entry in the appropriate subcategory if introducing a breaking change
- [ ] Verify that PR will be squashed during merge

PR #8232, opened 2025-06-12
Fix component usages status sorting with fragments

Cichan-mcc:
## Describe your changes

In [this PR](https://github.com/TouK/nussknacker/pull/8228) i introduced a possibility to sort senarios by status in component usages. Unfortunately it didn't work when a component was used in a fragment as they don't have a state and compared values could be undefined resulting in an error:

bad

Now fragments without the state will be listed as the last ones in order:

ok

## Checklist before merge


Dzuming (approved):
LGTM

PR #8242, opened 2025-06-16
[NU-2169] Toolbar buttons - disable and show tooltips based on user roles

mgoworko:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/config/scenariotoolbar/ScenarioToolbarsConfig.scala
```diff
@@ -78,6 +79,8 @@ object ToolbarConditionType extends Enumeration {
 final case class ToolbarCondition(
     fragment: Option[Boolean],
     archived: Option[Boolean],
+    always: Option[Boolean],
```
arkadius, line 88: What does it mean? I thought that lack of conditions means that it is always visible / enabled

mgoworko: There is now the `oneOfUserRoles: Option[Set[String]]` condition. But I kept the `always` condition. I makes sense, when it is used for settings like "hidden"/"disabled". When we tried to disable the test button on demo some time ago, we found out that we can't. There was no possibility to set something like hidden: "always". We could disable the button for fragments and for archived scenarios - but not "always".

So you are right that the default behavior is to always show enabled button. But the problem was, that before this change we could not create the "always disabled button". I think that we will use the  `oneOfUserRoles` setting in our demo use cases. But it was still a missing feature, so I think it should be added.

arkadius: ahh I'm getting it now. Let's write a comment about that. I guess that this option make sense only for disabling, because for hidding you can just remove the button.

mgoworko: Comment added. I also tested it with Dawid's recent FE fix. After those tests I made a compatible change in the API (new field `titleOverride`), in order to use the tooltip from toolbar config for deployment related buttons (those buttons before my change had their own custom tooltip contents and ignored the value from BE).

arkadius (approved):

designer/client/src/components/toolbars/scenarioActions/buttons/DeployButton.tsx
```diff
@@ -54,21 +54,22 @@ export default function DeployButton(props: ToolbarButtonProps) {
...
     const { t } = useTranslation();
-    const deployToolTip = !capabilities.deploy
-        ? t("panels.actions.deploy.tooltips.forbidden", "Deploy forbidden for current scenario.")
-        : hasErrors
-        ? t("panels.actions.deploy.tooltips.error", "Cannot deploy due to errors. Please look at the left panel for more details.")
-        : !saveDisabled
-        ? t("panels.actions.deploy.tooltips.unsaved", "You have unsaved changes.")
-        : null;
+    const deployToolTip =
+        titleOverride ?? !capabilities.deploy
```
arkadius: Shouldn't it be called tooltipOverride? I thought that the title is this short text label below the icon.

mgoworko: In this API and in buttons-related FE components it is called `title`. The short text below the icon is `name`. I think that in the current situation we should stay with `title`. Changing it IMO would be possible if we refactor it on FE side too.

arkadius: Ok, I got it

arkadius (approved):

PR #8271, opened 2025-06-26
[NU-2268] Pass default params for quick deploy/redeploy

lukasz-bigorajski:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
@@ -3,7 +3,11 @@ package pl.touk.nussknacker.ui.process.deployment
 import cats.data.Validated
 import cats.effect.IO
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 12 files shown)


arkadius (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -225,6 +235,44 @@ class DeploymentService(
...
   }
 
+  private def prepareNodesDeploymentData(
+      command: CommonDeploymentCommand,
+      processDetails: ScenarioWithDetailsEntity[CanonicalProcess],
+      canonicalProcess: CanonicalProcess,
+      actionName: ScenarioActionName
+  )(implicit user: LoggedUser): Future[NodesDeploymentData] = command.nodesDeploymentData match {
+    case Some(data) => Future.successful(data)
+    case None =>
```
arkadius, line 245: I thought about always merging defaults into existing parameters (we do this that way for node parameters), but we can try with this solution as well

PR #8297, opened 2025-07-01
[NU-2277] Fix for: Scenario using fragment deployment ended up with NoSuchElementException

arkadius:
## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
@@ -3,10 +3,12 @@ package pl.touk.nussknacker.ui.process.deployment
 import cats.data.Validated
 import cats.effect.IO
 import com.typesafe.scalalogging.LazyLogging
+import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.component.{
   ComponentAdditionalConfig,
   DesignerWideComponentId,
-  NodesDeploymentData
+  NodesDeploymentData,
+  StaticParameterConfig
 }
 import pl.touk.nussknacker.engine.api.deployment._
 import pl.touk.nussknacker.engine.api.process._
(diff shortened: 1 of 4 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentService.scala
```diff
@@ -238,39 +239,38 @@ class DeploymentService(
...
       actionName: ScenarioActionName
-  )(implicit user: LoggedUser): Future[NodesDeploymentData] = command.nodesDeploymentData match {
-    case Some(data) => Future.successful(data)
-    case None =>
+  )(implicit user: LoggedUser): NodesDeploymentData = {
+    command.nodesDeploymentData.getOrElse {
       processingTypeToActionInfoService
         .forProcessingTypeUnsafe(processDetails.processingType)
-        .getActionParameters(
-          canonicalProcess.toScenarioGraph,
```
arkadius, line 249: Here was the problem. Given `canonicalProcess` was a scenario with resolved fragments. This scenario form wasn't meant to be converted to the scenario graph representation (with nodes and edges). Only unresolved scenario was.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceSpec.scala
```diff
@@ -845,6 +852,33 @@ class DeploymentServiceSpec
     }
   }
 
+  "should allow to deploy scenario using fragment with unspecified nodesDeploymentData" in {
```
mgoworko, line 855: nitpick: I'd suggest more descriptive test name - describing situation, when scenario ends with fragment, which ends with a sink.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/TestDeploymentServiceFactory.scala
```diff
@@ -129,24 +130,34 @@ class TestDeploymentServiceFactory(dbRef: DbRef) {
         dbioRunner
       )
 
-    val validator = ProcessTestData.testProcessValidator(validator = ProcessValidator.default(modelData))
+    val fragmentResolver = new FragmentResolver(newFragmentRepository(dbRef))
     val deploymentService = new DeploymentService(
```
mgoworko, line 134: Looking at those changes I think that the way we create the DeploymentService in the tests should be more closely associated with how we create it in the production code. Maybe by some create/factory methods, that we could also use in tests. But that's something for another time, not in this hotfix PR.

mgoworko (approved):
LGTM with 2 small comments

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/DeploymentServiceSpec.scala
arkadius: Actually, I figured out that this requirement that the scenario ends with a fragment, which ends with a sink wasn't necessary to reproduce the bug. Any scenario using a fragment caused this bug, so this test case represents the minimal use case reproducing the bug.

designer/server/src/test/scala/pl/touk/nussknacker/ui/process/deployment/TestDeploymentServiceFactory.scala
arkadius: You are right, I did some changes in this direction in previous changes and in this change as well. There is a lot of work to do in this area.

PR #8301, opened 2025-07-03
[NU-2263] Fix broken collection helper suggestions

mateuszkp96:
## Describe your changes
  SpelExpressionSuggester calls `pl.touk.nussknacker.engine.definition.clazz.MethodDefinition#computeResultType` with the type taken from the method definition.
  For the COLLECTION helper, arguments often have a type java.util.Collection instead of java.util.List
## Checklist before merge


utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -74,7 +74,7 @@ trait CollectionUtils extends HideToString {
   @Documentation(description =
     "Returns a slice of the list starting with start index (inclusive) and ending at stop index (exclusive)"
   )
-  @GenericType(typingFunction = classOf[ListTyping])
+  @GenericType(typingFunction = classOf[ListOrCollectionTyping])
```
arkadius: 1. Why `ListTyping` here was causing an error?  Because `List` is too specific and given argument is a `Collection`? I wonder if we could make sth to detect such problems earlier.  Maybe we could create a test that would walk through all helpers in the default model and ask for suggestions for `#${helper}.`? Or we could have some similar sanity check during application startup?
2. Why List Or Collection? Can't we just check in this typing function that given "container" is a subclass of `Collection`?
3. Do we still use old typing functions somewhere or we can remove them?

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -490,12 +498,80 @@ object CollectionUtils {
...
+    private val fClass: Class[F[_]]                             = classTag.runtimeClass.asInstanceOf[Class[F[_]]]
+    private val collectionClass: Class[java.util.Collection[_]] = classOf[java.util.Collection[_]]
+
+    override def computeResultType(
+        arguments: List[typing.TypingResult]
+    ): ValidatedNel[GenericFunctionTypingError, typing.TypingResult] = arguments match {
+      case (f @ TypedClass(`fClass`, element :: Nil)) :: _ =>
+        f.copy(params = element.withoutValue :: Nil).validNel
+      case (f @ TypedClass(`collectionClass`, element :: Nil)) :: _ =>
+        f.copy(klass = fClass, params = element.withoutValue :: Nil).validNel
```
arkadius: This copy doesn't leave anything unchanged. I think that creating a new generic type from scratch would better show the intention.

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -490,12 +498,80 @@ object CollectionUtils {
 
   }
 
+  class JavaCollectionTyping[F[_] <: java.util.Collection[_]](implicit classTag: ClassTag[F[_]])
```
arkadius: Let's call `F` a `ResultCollectionType` or sth similar to emphasize that whatever collection type is passed, we return this collection type

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
mateuszkp96: 1. SpelExpressionSuggester for a few months calls `pl.touk.nussknacker.engine.definition.clazz.MethodDefinition#computeResultType` with the type taken from the method definition.
For the COLLECTION helper, arguments often have a type `java.util.Collection` instead of `java.util.List`, and that was a source of errors. 
I added a new test for all helpers in `NodesApiHttpServiceBusinessSpec`. In tests, the DevModel is used instead of DefaultModel (helpers of DefaultModel are defined in code and are hard to access, and DevModel seems to have all DefaultModel helpers).

2. I fixed it according to your suggestion.

3. I didn't find any usages of old typing functions. It seems we can remove them.

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/NodesApiHttpServiceBusinessSpec.scala
```diff
@@ -905,6 +908,235 @@ class NodesApiHttpServiceBusinessSpec
...
+          )
+          .post(s"$nuDesignerHttpAddress/api/parameters/streaming/suggestions")
+          .Then()
+      }
+
+      requestSuggestions("#")
+        .statusCode(200)
+        .body(
+          "methodName",
+          Matchers.contains(
```
arkadius: Let's replace this matcher with checking that the list of suggestions is not empty. We already have tests that check what methods are exposed by classes (`*ClassDiscoveryTest`), I don't want to make changes of the helpers API more "concretened".

utils/default-helpers/src/main/scala/pl/touk/nussknacker/engine/util/functions/collection.scala
```diff
@@ -490,13 +491,56 @@ object CollectionUtils {
 
   }
 
+  class JavaCollectionTyping[ResultCollectionType[_]](implicit classTag: ClassTag[ResultCollectionType[_]])
```
arkadius, line 494: `ResultCollectionType[_]: ClassTag` syntax didn't work?

mateuszkp96: This syntax didn't work here, I get error `type ResultCollectionType takes type parameters`

arkadius (approved):

PR #8308, opened 2025-07-04
ToJsonEncoder refactor

arkadius:
- extracted default encoder
- extracted loose encoder which so far was used locally in a few places
- local usages of ToJsonEncoder replaced with one of variants above
- ToJsonEncoderWithFallback become an implementation detail, everywhere we use ToJsonEncoder instead
- encode -> encodeUnsafe

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
```diff
@@ -0,0 +1,72 @@
...
+import cats.data.ValidatedNel
+import io.circe.Json
+import io.circe.Json._
+import pl.touk.nussknacker.engine.util.json.ToJsonEncoderCustomisation
+
+import java.util.ServiceLoader
+import scala.jdk.CollectionConverters._
+import scala.util.Try
+
+object ToJsonEncoder {
```
arkadius, line 13: Moved to components-api to allow usage in Typed value determining. Unfortunatelly, body also is changed - because of that git recognize it as a new file

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/FromInstanceTypeDeterminer.scala
```diff
@@ -35,7 +35,7 @@ object FromInstanceTypeDeterminer {
       case other =>
         Typed(other.getClass) match {
           case typedClass: TypedClass =>
-            ToJsonEncoderWithFallback.encodeValue(other) match {
+            ToJsonEncoder.default.encode(other) match {
```
arkadius, line 38: Here is a functional change - the result will also be valid for `DisplayJson` and custom encoders will be used. IMO it will only be better with that approach

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/testmode/TestInterpreterRunner.scala
```diff
@@ -12,7 +12,9 @@ object TestInterpreterRunner {
    * and in this case there is loaders conflict
    */
   private[engine] def testResultsVariableEncoder: Any => io.circe.Json = {
-    lazy val encoder = ToJsonEncoder(failOnUnknown = false, Thread.currentThread().getContextClassLoader)
+    // Variables keep some intermediate state of a scenario. We can't be sure that there are only types that are supported
+    // for json encoding. It is better to print not-well formatted .toString result than fail in this case
+    lazy val variablesEncoder = new ToJsonEncoder(failOnUnknown = false, Thread.currentThread().getContextClassLoader)
```
mgoworko: Perhaps there can be a method ToJsonEncoder.looseEncoder(classLoader) for cases like that, where custom loader is used?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/typed/FromInstanceTypeDeterminer.scala
mgoworko: I remember, that I didn't do that change some time ago because of some failing typing tests after making that modification. But there were a lot of refactors since that time, if tests pass it's great change :+1:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
```diff
@@ -0,0 +1,72 @@
...
+  val looseEncoder = new ToJsonEncoder(failOnUnknown = false, classOf[ToJsonEncoder].getClassLoader)
+
+}
+
+object StrictToJsonEncoder extends ToJsonEncoder(failOnUnknown = true, classOf[ToJsonEncoder].getClassLoader) {
+  // Unhides encode method, which is hidden for loose encoder
+  override def encode(obj: Any): ValidatedNel[String, Json] = super.encode(obj)
+}
+
+class ToJsonEncoder(failOnUnknown: Boolean, classLoader: ClassLoader) {
```
mgoworko: Maybe the class can have private constructor and only `default`(`strict`)/`loose` variants in the ToJsonEncoder? It would require adding methods accepting custom classLoader's

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
```diff
@@ -0,0 +1,72 @@
...
+  val default: StrictToJsonEncoder.type = StrictToJsonEncoder
+
+  // For this encoder, encodeUnsafe() doesn't throw exceptions for unknown values - as a fallback .toString method is used
+  // encode() method is hidden because it always returns Valid
+  val looseEncoder = new ToJsonEncoder(failOnUnknown = false, classOf[ToJsonEncoder].getClassLoader)
+
+}
+
+object StrictToJsonEncoder extends ToJsonEncoder(failOnUnknown = true, classOf[ToJsonEncoder].getClassLoader) {
+  // Unhides encode method, which is hidden for loose encoder
```
mgoworko: It is not really important nitpick, but maybe for better readability underlying method could be changed to `protected def doEncode`, and this method would just invoke it, without unhiding/using `super.` etc..

mgoworko (commented):
Some small refactors suggested, can be merged after applying or rejecting them

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/testmode/TestInterpreterRunner.scala
arkadius: I applied simillar approach, take a look

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
arkadius: It can't be done by methods because this class is used by an external project. I could make it sealed trait and leave only `StrictToJsonEncoder` and `LooseToJsonEncoder` but I have to be sure that this project doesn't have some dynamic logic based on `failOnUnknown`. Can you check it for me? You know which project ;)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
arkadius: Good point changed

components-api/src/main/scala/pl/touk/nussknacker/engine/api/json/encoders/ToJsonEncoder.scala
mgoworko: It is used like that:

  private val encoder = ToJsonEncoder(failOnUnknown = false, loader, highPriority)
  private def highPriority: PartialFunction[Any, Json] = ...

arkadius: Thanks. I made `ToJsonEncoder` a sealed trait, `LooseToJsonEncoder` will still be available for extending. After doing this I changed the boolean-based logic to method-based (IMO the intention is more clear now).  I also added a MigrationGuide entry

mgoworko (approved):
LGTM

PR #8349, opened 2025-07-17
Ability to configure the default editors for the String type parameters

mateuszkp96:
## Checklist before merge


lciolecki (approved):
Good feature

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -74,6 +77,23 @@ object ModelConfig {
...
+
+  object EditorConfig {
+
+    val default: EditorConfig = EditorConfig(
+      editorsForStringType = NonEmptyList.of(SpelTemplateParameterEditor, SpelParameterEditor)
+    )
+
+  }
+
+  sealed trait DefaultEditorForStringType
```
arkadius: Where it is used?

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -74,6 +77,23 @@ object ModelConfig {
     case object SingleJsonTemplateParameter extends JsonLikeValuesEnteringMode
   }
 
+  final case class EditorConfig(editorsForStringType: NonEmptyList[ParameterEditor])
```
arkadius: Let's generalise it to `GlobalParamersConfig` or similar. Thanks to that it will be easier to configure default for other parameter definition properties such as validator, hint text etc.

components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
mateuszkp96: Sorry, leftover after a draft version

arkadius (approved):
LGTM

PR #8371, opened 2025-07-21
[NU-2189] Make ContextInitializer be responsible only for variables from Raw preparing, instead of complete Context

arkadius:
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/process/ContextInitializer.scala
```diff
@@ -15,11 +13,9 @@ import pl.touk.nussknacker.engine.api.typed.typing.TypingResult
 trait ContextInitializer[Raw] extends Serializable {
 
   /**
-    * Initializes `Context` with raw event value.
-    *
-    * @param contextIdGenerator - context id generator
+    * convert a raw event to initial Context variables in runtime
```
mslabek: How about
```suggestion
    * Converts a raw input event (for example a Kafka message) into initial context variables in runtime.
```

mslabek (approved):

PR #8449, opened 2025-08-04
Update Jackson to 2.18.4

piotrp:
## Describe your changes

TODO: solve Jackson incompatibility by removing changes from #7066. To sum up Slack discussion:
> If it's unused let's remove it

## Checklist before merge


utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/ContentTypes.scala
```diff
@@ -9,8 +9,6 @@ object ContentTypes extends Enumeration {
 }
 
 object ContentTypesSchemas {
-  val schemaForJson: OpenAPIJsonSchema = OpenAPIJsonSchema("{}")
-  // note: "" deserializes to com.fasterxml.jackson.databind.node.MissingNode,
-  // which causes NPE when using Jackson 2.18.3+, because from this version on it decodes to null
-  val schemaForPlain: OpenAPIJsonSchema = OpenAPIJsonSchema("")
+  val schemaForJson: OpenAPIJsonSchema  = OpenAPIJsonSchema("{}")
```
lciolecki, line 12: Hmm, do we even need these two schemas?

build.sbt
```diff
@@ -287,7 +287,7 @@ val circeV                  = "0.14.14"
 val circeGenericExtrasV     = "0.14.4"
 val circeYamlV              = "0.15.2"     // 0.15.3 drops Scala 2.12
 val jwtCirceV               = "11.0.2"
-val jacksonV                = "2.18.2"     // 2.18.3+ breaks ContentTypesSchemas.schemaForPlain
```
lciolecki: Hmm there is 2.19.2 version and 2.18.4.1 (I don't know what it means..),  https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core. Why not to upgrade to newest version?

piotrp: * 2.18.4.1 - that would require a few changes in build.sbt, because that version exists only for jackson-core
* 2.19 - I want to update one major version at a time, in case I break something that isn't covered by tests

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/ContentTypes.scala
piotrp: Unfortunately, yes:
```scala
class UniversalSchemaSupportDispatcher private (kafkaConfig: KafkaConfig) {
...
  def forParsedSchema(parsedSchema: ParsedSchema): UniversalSchemaSupport = parsedSchema match {
    // For ad hoc tests we want to present the user with json editor when topic has no schema and content type Json was selected
    case ContentTypesSchemas.schemaForJson => NoSchemaJsonSupport
    case _                                 => forSchemaType(parsedSchema.schemaType())
  }
```

utils/schemed-kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/schemedkafka/schemaregistry/ContentTypes.scala
```diff
@@ -9,8 +9,6 @@ object ContentTypes extends Enumeration {
...
+  val schemaForPlain: OpenAPIJsonSchema = OpenAPIJsonSchema("true")
```
lciolecki, line 13: Are you sure about this? :)

piotrp: Yes, it's that or some heavier refactoring. `true` after parsing results in the same schema: `{}`. An alternative would be to use `{} ` (`{}` with some whitespace around) with custom `equals` method for `OpenAPIJsonSchema`.

I mention custom `equals` because `OpenAPIJsonSchema` is a fake case class - default Scala's `equals` for case classes doesn't work there, it's implemented like this:
```scala
  test("boxes are equal") {
    class BoxParent {
      override def equals(obj: Any): Boolean = true
      override def hashCode(): Int = 0
    }

    case class Box(v: String) extends BoxParent

    // this equality is true in Scala, but not in real world
    Box("socks") shouldBe Box("gold")
  }
```

piotrp:
Updated in internal 2.3 branch, will be part of a future GitHub release

PR #8467, opened 2025-08-06
Fix periodic-dev processing mode config in dev configuration

mgoworko:
## Describe your changes

The periodic-dev processing mode stopped working in local test environment and on our development environment, fixed. It was configured to use minicluster, but scheduling mechanism on Flink does not support minicluster (it is not needed outside of tests and not trivial to add).

## Checklist before merge

### DIFF
--- a/nussknacker-dist/src/universal/conf/dev-application.conf
+++ b/nussknacker-dist/src/universal/conf/dev-application.conf
@@ -263,12 +263,6 @@ scenarioTypes {
       restUrl: "http://jobmanager:8081"
       restUrl: ${?FLINK_REST_URL}
       useMiniClusterForDeployment: false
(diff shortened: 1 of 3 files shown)


nussknacker-dist/src/universal/conf/dev-application.conf
```diff
@@ -263,12 +263,6 @@ scenarioTypes {
       restUrl: "http://jobmanager:8081"
       restUrl: ${?FLINK_REST_URL}
       useMiniClusterForDeployment: false
-      useMiniClusterForDeployment: ${?FLINK_USE_MINICLUSTER_FOR_DEPLOY}
```
mgoworko, line 266: In `periodic-dev` we should always use standalone Flink

piotrp (approved):

PR #8480, opened 2025-08-11
Expose current url location as query param in tab link

mgoworko:
## Checklist before merge

### DIFF
--- a/designer/client/src/components/MenuBar/TabElement.tsx
+++ b/designer/client/src/components/MenuBar/TabElement.tsx
@@ -5,17 +5,25 @@ import { NavLink } from "react-router-dom";
 import type { DynamicTabData } from "../../containers/DynamicTab";
 
 function UnstyledTabElement({ tab, ...props }: { tab: DynamicTabData; className?: string }): JSX.Element {
-    const { id, type, url, title } = tab;
+    const { id, type, url, title, currentLocationInQuery } = tab;
+    const fullUrl =
+        !currentLocationInQuery || !currentLocationInQuery.enabled
+            ? url
+            : (() => {
+                  const enrichedUrl = new URL(url);
+                  enrichedUrl.searchParams.set(currentLocationInQuery.parameterName, window.location.href);
+                  return enrichedUrl.toString();
+              })();
     switch (type) {
         case "Local":
             return (
-                <NavLink to={url} {...props}>
+                <NavLink to={fullUrl} {...props}>
                     {title}
                 </NavLink>
             );
         case "Url":
             return (
-                <a href={url} target={"_blank"} rel="noreferrer" {...props}>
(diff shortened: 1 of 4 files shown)


designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
```diff
@@ -130,6 +131,11 @@ object TopTabType extends Enumeration {
     parameterName: Option[String] = Some("auth_token")
 )
 
+@JsonCodec final case class CurrentLocationInQueryTabSettings(
```
lciolecki, line 134: I don't like defaults at case classes. I would prefer move it to companion object as default instance?

lciolecki (commented):
@Dzuming can you take a glance at FE part?

designer/client/src/components/MenuBar/TabElement.tsx
```diff
@@ -1,21 +1,32 @@
 import { styled } from "@mui/material";
 import React from "react";
-import { NavLink } from "react-router-dom";
+import { NavLink, useLocation } from "react-router-dom";
 
 import type { DynamicTabData } from "../../containers/DynamicTab";
 
 function UnstyledTabElement({ tab, ...props }: { tab: DynamicTabData; className?: string }): JSX.Element {
-    const { id, type, url, title } = tab;
+    const { id, type, url, title, currentLocationInQuery } = tab;
```
Dzuming: Maybe we can move it outside of the React component since it's not related to the react state or props?

```
export function getTabNavigationUrl(tab: DynamicTabData): string {
    const { url, currentLocationInQuery } = tab;

    const fullOriginalPath = window.location.pathname + window.location.search + window.location.hash;
    const fullOriginalUrl = window.location.origin + fullOriginalPath;

    if (!currentLocationInQuery || !currentLocationInQuery.enabled) {
        return url;
    }

    const enrichedUrl = new URL(url);
    enrichedUrl.searchParams.set(currentLocationInQuery.parameterName, fullOriginalUrl);
    return enrichedUrl.toString();
}
```

and use it like

```
const navigationUrl = getTabNavigationUrl(tab);
```

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/SettingsResources.scala
mgoworko: Defaults are not needed. I just copied them from preexisting `AccessTokenInQueryTabSettings`. Removed.

designer/client/src/components/MenuBar/TabElement.tsx
mgoworko: Ok, done. But I had to use `useLocation()` - in order to change the url without reloading GUI.

Dzuming (approved):

lciolecki (approved):

PR #8540, opened 2025-08-27
[NU-2190] Kafka source: Allow to define custom watermark startegy

arkadius:
## Describe your changes

https://github.com/user-attachments/assets/7e202aa0-8937-43a4-b029-1254a2cf662e

## Checklist before merge


engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/watermarkstrategy/FlinkWatermarkStrategyRuntimeHandler.scala
```diff
@@ -0,0 +1,85 @@
...
+import pl.touk.nussknacker.engine.api.typed.typing.Typed
+import pl.touk.nussknacker.engine.flink.api.process.FlinkLazyParameterFunctionHelper
+import pl.touk.nussknacker.engine.flink.api.typeinformation.TypeInformationDetection
+import pl.touk.nussknacker.engine.util.watermarkstrategy.WatermarkStrategyOptions
+
+import java.time.{Duration, Instant}
+
+object FlinkWatermarkStrategyRuntimeHandler {
+
+  // TODO: rename this hidden variable to human readable name and make it available for user during typing as well
```
arkadius: It is something that we should discuss this. IMO, exposing this variable to the user can open some extra features for the user. For example: they can do some transformation based on this time, and easily switch it to another field without breaking the whole transformation logic. We can also add a timestamp parameter to kafka sink and give the user the ability to change the default behaviour.  We could do this in some next change

engine/lite/components/kafka/src/main/scala/pl/touk/nussknacker/engine/lite/components/LiteKafkaSourceImplFactory.scala
```diff
@@ -30,7 +32,9 @@ class LiteKafkaSourceImplFactory[K, V] extends KafkaSourceImplFactory[K, V] {
       formatter: UniversalToJsonFormatter[K, V],
       contextInitializer: ContextInitializer[ConsumerRecord[K, V]],
       testParametersInfo: KafkaTestParametersInfo,
-      namingStrategy: NamingStrategy
+      namingStrategy: NamingStrategy,
+      // This options are not handled TODO we should either not present these parameters or add the support for them
```
arkadius, line 36: I think that we should postpone this for a later time. Lite streaming isn't top-important right now.

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/watermarkstrategy/FlinkWatermarkStrategyRuntimeHandler.scala
lciolecki: I'm not sure, but we have something similar in `#metaInput` or `#meta` variable? I remember we have something like this with kafka key, timestamp.

lciolecki (approved):
Nice!

PR #8565, opened 2025-09-04
[NU-2190] Event generator: component rewritten to DynamicComponent

arkadius:
...  for further changes around TimestampStrategy

## Checklist before merge

### DIFF
--- a/engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
+++ b/engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
@@ -14,8 +14,13 @@ import pl.touk.nussknacker.engine.api._
 import pl.touk.nussknacker.engine.api.VariableConstants.InputVariableName
 import pl.touk.nussknacker.engine.api.component.UnboundedStreamComponent
 import pl.touk.nussknacker.engine.api.context.{ProcessCompilationError, ValidationContext}
-import pl.touk.nussknacker.engine.api.definition.Parameter
-import pl.touk.nussknacker.engine.api.editor.{Editor, EditorType}
+import pl.touk.nussknacker.engine.api.context.transformation.{
+  DefinedLazyParameter,
+  NodeDependencyValue,
+  SingleInputDynamicComponent
+}
+import pl.touk.nussknacker.engine.api.definition._
+import pl.touk.nussknacker.engine.api.definition.ParameterCategory.Advanced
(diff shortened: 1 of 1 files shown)


engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -53,36 +57,84 @@ object EventGeneratorSourceFactory
...
-  def create(
-      @ParamName("schedule")
-      @Editor(
-        `type` = EditorType.DURATION_EDITOR,
-        timeRangeComponents = Array(ChronoUnit.DAYS, ChronoUnit.HOURS, ChronoUnit.MINUTES, ChronoUnit.SECONDS)
+  override type State = Nothing
+
+  override def nodeDependencies: List[NodeDependency] = List.empty
+
+  private lazy val scheduleParameterName = ParameterName("schedule")
```
lciolecki, line 67: Can we keep one standard, and parameters name start with upper case ?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -105,7 +157,7 @@ class EventGeneratorSourceFactory(customTimestampAssigner: TimestampWatermarkHan
           flinkNodeContext: FlinkCustomNodeContext
       ): DataStream[Context] = {
         // Without this local variable, flatMap function is not serializable
-        val count = Option(nullableCount).map(_.toInt).getOrElse(1)
+        val localCount = count
```
lciolecki, line 159: Do we need this?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
arkadius: Yes, we can, but ... It shouldn't be in this PR - it is only a rewite, without modification. We should review all components and align names

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
arkadius: Take a look at the comment above this line

lciolecki (approved):

PR #8574, opened 2025-09-05
Adapt dictionary values when migrating scenario

mgoworko:
## Describe your changes

When we have dictionary value of a parameter on a source environment, and the target environment accepts spel values or fixed list of values, we should be able to adapt the ductionary value to raw spel value or value from fixed list (when it matches one of allowed items.

This PR introduces adaptatation logic, which is applied on target environment and adapts (when possible) dictionary values to value types accepted on target env - instead of returning errors.

## Checklist before merge


jedrz (approved):
LGTM

PR #8576, opened 2025-09-05
Using NodeId in most places of an API instead of switching between String and NodeId and vice versa

arkadius:
## Describe your changes

After this change, the last missing place where we have `String` -> `NodeId` conversion is the scenario api.

## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
@@ -1,5 +1,7 @@
 package pl.touk.nussknacker.engine.api
 
+import pl.touk.nussknacker.engine.api.process.ProcessName
+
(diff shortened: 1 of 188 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/Context.scala
```diff
@@ -15,8 +17,8 @@ object Context {
 }
 
 final case class ContextId private (
-    scenarioName: String,
-    originatingNodeId: String,
+    scenarioName: ProcessName,
```
mslabek, line 20: scenarioName and originatingNodeId are used in `legacyString` on line 42, 43. Theres no compile error since its still a string, but it will be ProcessName(name) instead of just name

mslabek (approved):

PR #8617, opened 2025-09-15
Run off schedule button fixes

mgoworko:
## Describe your changes

There was a bug in scenario state handling on FE side. The state was refreshed, but the new state was merged with old state, instead of replacing it. As a result, maps and arrays inside the state representation were not correctly updated. (For example, the "Run off schedule" button was not disabled when scenario state changed - it was not removed from the list of allowed actions, because the new list of allowed actions was merged with older one instead of replacing it, same with custom tooltips).

Additionally, updated error messages and tooltips for "run now" button.

## Checklist before merge

### DIFF
--- a/designer/client/src/reducers/scenarioState.ts
+++ b/designer/client/src/reducers/scenarioState.ts
@@ -1,5 +1,4 @@
 import { produce } from "immer";
-import { merge } from "lodash";
 
 import type { Reducer } from "../actions/reduxTypes";
 import type { ProcessStateType } from "../components/Process/types";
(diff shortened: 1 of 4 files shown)


designer/client/src/reducers/scenarioState.ts
```diff
@@ -11,8 +10,7 @@ export const reducer: Reducer<ProcessStateType> = produce((draft, action) => {
             // Since scenario endpoint doesn't return null attributes the state will be undefined for fragments.
             // Redux and Immer does not allow to return undefined values so in that case we return null explicitly.
             if (action.scenario.state) {
-                merge(draft, action.scenario.state);
```
Dzuming, line 14: Why don't we want to merge the state here?

mgoworko: That is because we need to replace the state stored on FE with the new state returned from backend, and not use some merged value of previous and current state. 

Merging state caused problems - because the sets and maps in the state data structure were concatenated, instead of replaced. As a result, when some action became unavailable (in this case "Run now"), it was not removed from the list of available actions. Same with custom tooltips - they were never cleared, even when the backend ceased to return them.

Dzuming (approved):

PR #8621, opened 2025-09-17
Fix for query parameters decoding, fix for node details window opening with encoded node ID

mateuszkp96:
## Checklist before merge


designer/client/src/containers/hooks/useSearchQuery.tsx
```diff
@@ -1,30 +1,69 @@
...
 const DEFAULT_ARRAY_FORMAT: ParseOptions["arrayFormat"] = "comma";
 
 function parseQuery(searchString = window.location.search): QueryRecord {
-    return parse(searchString, {
+    const parsedRawQuery = parse(searchString, {
         arrayFormat: DEFAULT_ARRAY_FORMAT,
         parseNumbers: true,
         parseBooleans: true,
-        decode: true,
+        decode: false,
```
Dzuming, line 20: Do you know what the problem is with the query-string library and automatic decode?

Dzuming (changes requested):

designer/client/src/containers/hooks/useSearchQuery.tsx
mateuszkp96: This library does not handle query params arrays well with a comma delimiter. It appears that during the encoding of query parameters containing a comma, the comma is not encoded, resulting in the query parameter value being split into an array of values instead of a single value. That's why I decided to encode and decode them manually. 

You can test invalid behaviour by adding a comma to the name of the scenario node. Then try to open and close the node details window. The query string will contain the nodeId param with a value that is a part of the node name.

Dzuming: 1. Could you provide a few unit tests for this behaviour?
2. Could you add a comment, why we need to parse it manually?

Dzuming (changes requested):

designer/client/src/containers/hooks/useSearchQuery.tsx
mateuszkp96: Done

Dzuming (approved):
LGTM

PR #8623, opened 2025-09-17
[NU-2190] Scenario testing / common test data format: respecting watermark strategy configured by the user + Testkit: ability to run scenario with input test records

arkadius:
## Checklist before merge


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/scenariotesting/DataRecordsSource.scala
```diff
@@ -0,0 +1,128 @@
...
+              .forMonotonousTimestamps[ContextWithEventTime]
+              .withTimestampAssigner(EventTimeTimestampAssigner)
+          )
+      )
+      .map((ctxWithEventTime: ContextWithEventTime) => ctxWithEventTime.context, flinkNodeContext.contextTypeInfo)
+  }
+
+  // env.fromData uses DataGeneratorSource which even for an empty list try to process one element
+  // which ends up with NoSuchElementException in FromElementsGeneratorFunction.tryDeserialize
+  private def fromDataProperHandlingEmptyList(env: StreamExecutionEnvironment): DataStreamSource[DataRecord] = {
```
mslabek, line 71: What are cases where we want to build DataRecordsSource with empty list of records? I assume we can't just treat this as illeagal state. Is it for situations where you test a scenario with multiple sources but not all sources had records specified for them and we need to stub them?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/scenariotesting/TestFlinkProcessCompilerDataFactory.scala
```diff
@@ -194,16 +189,14 @@ class TestFlinkProcessCompilerDataFactory(
...
+        outputValidationContext: ValidationContext,
         typingResult: TypingResult,
-        compilationDependencies: NodeCompilationDependencies,
-        invocationContext: Option[ComponentImplementationSpecificInvocationContext]
+        compilationDependencies: NodeCompilationDependencies
     ): Any = {
-      val outputValidationContext =
-        determineOutputValidationContext(originalSource, compilationDependencies, invocationContext)
-
       val commonFormatRecordsNelOpt =
```
mslabek: this nel is build just to match on if its empty. This can be simplified to an `if` statement with `.isEmpty` check

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/scenariotesting/DataRecordsSource.scala
arkadius: Yes, this is one of those situations. I also use it during the scenario state verification process, which basically behaves similarly to scenario testing, but with empty input test records and uses dumped scenario state as an initial scenario state.

mslabek (approved):

PR #8626, opened 2025-09-18
[NU-2190] LiveDataProvider: returing upstreamTimestamp

arkadius:
## Describe your changes

Changes:
- `LiveDataProvider` returns  `upstreamTimestamp` instead of some not-well-defined `timestamp`
- Input test data for "common format" also use `upstreamTimestamp` + type was changed to `Instant`
- `table` source uses `rowtime` column in expression for `Event time`

## Checklist before merge

### DIFF
--- a/engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/CustomWatermarkStrategySpec.scala
+++ b/engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/CustomWatermarkStrategySpec.scala
@@ -5,7 +5,6 @@ import io.circe.Json
 import org.scalatest.{LoneElement, OptionValues}
 import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.component.ComponentDefinition
-import pl.touk.nussknacker.engine.api.livedata.DataRecord
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.api.process.TopicName.ForSource
 import pl.touk.nussknacker.engine.build.ScenarioBuilder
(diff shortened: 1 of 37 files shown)


engine/flink/tests/src/test/scala/pl/touk/nussknacker/defaultmodel/CustomWatermarkStrategySpec.scala
```diff
@@ -127,24 +126,83 @@ class CustomWatermarkStrategySpec extends FlinkWithKafkaSuite with OptionValues
...
+
+    testResults.successes shouldBe List(
+      givenData,
+      givenData * 2,
+      givenData
+    )
+  }
+
+  test(
+    "should user upstream timestamp if event timestamp is configured as null expression by a user during scenario testing"
```
mslabek:
```suggestion
    "should use upstream timestamp if event timestamp is configured as null expression by a user during scenario testing"
```

utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/watermarkstrategy/WatermarkStrategyValidationHandler.scala
```diff
@@ -57,7 +57,7 @@ trait WatermarkStrategyValidationHandler { self: SingleInputDynamicComponent[_]
           outputValidationContext.localVariables.mapValuesNow(AdditionalVariableProvidedInRuntime(_)),
         defaultValue = Some(eventTimeDefaultValueExpression),
         hintText = Some(
-          s"An expression that determines the Event Time to be used in stateful stream processing. " +
+          s"An expression that determines the Event Time to be used in stateful stream processing. If this field won't be specified, the upstream source event time will be used.  " +
```
mslabek:
```suggestion
          s"An expression that determines the Event Time to be used in stateful stream processing. If this field is empty, the upstream source event time will be used.  " +
```

mslabek: I think from user perspective its better to say that field `is empty` than field `is specified`. And I think present tense in first part of conditional is correct (its first or even zero conditional)

mslabek (approved):

utils/components-utils/src/main/scala/pl/touk/nussknacker/engine/util/watermarkstrategy/WatermarkStrategyValidationHandler.scala
arkadius: You are right, my mistake, Thank you for catching it out!

PR #8644, opened 2025-09-25
fix: make defaultValue in input data records optional

Dzuming:
## Checklist before merge

### DIFF
--- a/designer/client/src/components/modals/TestingDataRecords/Dialog.tsx
+++ b/designer/client/src/components/modals/TestingDataRecords/Dialog.tsx
@@ -61,7 +61,7 @@ function Dialog(props: WindowContentProps<WindowKind, TestingData>): ReactElemen
                 ? {
                       sourceId: defaultParameter.sourceId,
                       timestamp: undefined,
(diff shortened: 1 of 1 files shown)


designer/client/src/components/modals/TestingDataRecords/Dialog.tsx
```diff
@@ -61,7 +61,7 @@ function Dialog(props: WindowContentProps<WindowKind, TestingData>): ReactElemen
                 ? {
                       sourceId: defaultParameter.sourceId,
                       timestamp: undefined,
-                      variables: defaultParameter.parameters[0].defaultValue.expression,
+                      variables: defaultParameter?.parameters?.[0]?.defaultValue?.expression ?? "",
```
arkadius: I see inconsitency with the first `?`. 2 lines above you call `.sourceId` directly, so either we have bug in `.sourceId` invocation or this `?` is redundant.

Dzuming: Good point, optional parameter here, `defaultParameter?` is redundant. Done ✔️

arkadius (approved):

PR #8663, opened 2025-10-03
[Nu-8656] Abstract FlinkEngineRuntimeContext

### LINKED ISSUE #8656: Adapt exception handler preparation in FlinkCustomNodeContext for Flink 2.0
## 💥 Proposal

In the current implementation of `FlinkCustomNodeContext` we have `exceptionHandlerPreparer: RuntimeContext => ExceptionHandler`. With new Flink API there are changes (for example for sinks) which cause that context is not being retrieved via `RichFunction#getRuntimeContext` method and context is being passed down explicitly via constructor injection.

We should change the signature of this function to `(RuntimeContext | WriterInitContext | ... ) => ExceptionHandler` (this ADT will probably expand while adapting other components to use new API).

We also should add new analogous trait for `WithExceptionHandler` since it uses a self-type requirement: `self: RichFunction` and new `sink2.Sink` does not conform this requirement.

DeamonDev:
## Describe your changes

In this PR I add wrapper for `RuntimeContext` to support `WriterInitContext` from new Flink's API. See #8656 for context.

Thanks for @piotrp for suggesting me the shape of `SinkWriterWithExceptionHandler` and general idea of wrapping this context.

## Checklist before merge


engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/FlinkEngineContext.scala
```diff
@@ -0,0 +1,34 @@
...
+final case class WriterInitCtx(ctx: WriterInitContext) extends FlinkEngineContext
+
+object FlinkEngineContextOps {
+
+  implicit class RichExceptionHandler(val f: FlinkEngineContext => ExceptionHandler) extends AnyVal {
+    def narrowToRuntimeCtx: RuntimeContext => ExceptionHandler =
+      rc => f(RuntimeCtx(rc))
+  }
+
+  implicit class RichFlinkEngineContext(val ctx: FlinkEngineContext) extends AnyVal {
```
arkadius: TBH I would prefer to replace this extension class by abstract methods in the FlinkEngineContext. We are in language which mix functional and object-oriented programming so why not to leverege the potential of polymorphism for this purpose? The approach with extension classes (Ops-classes) is good if:
1. you have type classes - not our case
2. you have no control over the class that you want to enrich with some methods - in our case, we have a control over `FlinkEngineContext`

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
```diff
@@ -124,19 +126,31 @@ class FlinkProcessRegistrar(
...
+      val runtimeCtxExceptionHandlerPreparer = (runtimeContext: RuntimeContext) =>
+        compilerDataForProcessPart(None)(runtimeContext.getUserCodeClassLoader).prepareExceptionHandler(
+          RuntimeCtx(runtimeContext)
+        )
+      val exceptionHandlerPreparer = (flinkEngineContext: FlinkEngineContext) =>
+        flinkEngineContext match {
+          case RuntimeCtx(runtimeContext) =>
+            runtimeCtxExceptionHandlerPreparer(runtimeContext)
+          case WriterInitCtx(writerInitContext) =>
+            compilerDataForProcessPart(None)(writerInitContext.getUserCodeClassLoader.asClassLoader())
```
arkadius: You extracted a function for `RuntimeContext` and for `WriterInitCtx` this logic is inlined. It doesn't help with the analysis of the differences between the two cases.. Let's unify it somehow. Also, I see that `compilerDataForProcessPart(None)(runtimeContext.getUserCodeClassLoader)` can be deduplicated. Am I right?

engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/exception/SinkWriterWithExceptionHandler.scala
```diff
@@ -0,0 +1,17 @@
+package pl.touk.nussknacker.engine.flink.api.exception
+
+import org.apache.flink.api.connector.sink2.SinkWriter
+
+trait SinkWriterWithExceptionHandler[T] {
```
arkadius, line 5: Where is the usage of this class?

DeamonDev: See #8656 (last section)

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/FlinkProcessRegistrar.scala
DeamonDev: Yeah, I unified them. Thanks

engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/FlinkEngineContext.scala
DeamonDev: Good idea

engine/flink/components-api/src/main/scala/pl/touk/nussknacker/engine/flink/api/exception/SinkWriterWithExceptionHandler.scala
arkadius: Can you provide an example usage based on some very simple Sink using sink v2 API and add a test showing that the exception handling logic really works as we expected? For example this sink can use `LazyParameter` and we can provide an expression dividing by 0 / accessing non-existing object in array etc.

arkadius: The sink can save a result to `TestResultsHolder`. We can pass one invalid event, which should be handled by exception handler and then we can pass a valid element. Without exception handling logic, job would restart after invalid element. Thanks to exception handler, invalid element will be skipped. WDYT?

DeamonDev: Sure :bulb:

engine/flink/executor/src/test/scala/pl/touk/nussknacker/engine/process/exception/SinkWriterWithExceptionHandlerTest.scala
```diff
@@ -0,0 +1,95 @@
...
+  private lazy val flinkMiniClusterWithServices = FlinkMiniClusterFactory.createUnitTestsMiniClusterWithServices()
+
+  override protected def afterAll(): Unit = {
+    super.afterAll()
+    flinkMiniClusterWithServices.close()
+  }
+
+  before {
+    flinkMiniClusterWithServices.withDetachedStreamExecutionEnvironment(env => {
+      env.setParallelism(1)
```
arkadius: Let's use our scenarios api instead of direclty calling flink api.
Take a look at usages of `TestScenarioRunner.flinkBased`. It is not too much work to use it and thanks to that tests will be more functional - showing how people can use this feature instead of showing some technical perspective

DeamonDev: As I understand then I would need to add this source via `.withExtraComponents` and register sink via `ComponentDefinition`. Then I would probably (correct me if I'm wrong) define sink by extending `SinkFactory`. Then  I would need to implement `BasicFlinkSink` but it currently does not support new API. So I propose leave the test as it is or get rid on this new trait at all (thanks to new Flink context lifecycle we dont need to even have this trait since runtime context is initialized and passed down via constructor to sink)

arkadius: You are almost 100% right. The only difference is that you don't need to implement `BasicFlinkSink`. Instead, you need to implement `FlinkSink`. It operates on `DataStream` so you have full access to Flink's API. This discussion shows us that this step is very important because providing new API have no value if you can't show how to use it to other devs. For the `SinkFactory` implementation, I recommend using the `@MethodToInvoke` approach. It is the quickest way to achieve our goal.

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/table/sink/SinkWriterWithExceptionHandlerTest.scala
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  test("new source handle exceptions properly when using SinkWriterWithExceptionHandler trait") {
+    val scenario = ScenarioBuilder
+      .streaming("test")
+      .source("source", "source")
+      .emptySink("sinkV2", "sinkV2", "value" -> "#input".spel)
+
+    val result = runner.runWithData(scenario, List(1, 2, 3, 0, 4))
+    result.validValue.errors shouldBe empty
```
arkadius: Why are errors empty? I expected that there would be one error with dividing by zero problem

DeamonDev: I think these errors are collected exceptions being thrown during job execution, so I think this convicts the handler works properly

arkadius: Nope, this `errors` field contains errors handled by the exception handler mechanism. Unhandled exceptions would cause the job switch to the FAILURE state or be restarted. This test showed us that we have a bug either in our solution or in test runner.

DeamonDev: okay I see, that's probably because I created ad hoc `ExceptionHandler`. I will try to use`TestFlinkExceptionHandler`

arkadius (commented):
:+1: It looks good. Let's wait until all the paperwork is completed and then we can merge it.

arkadius (approved):

PR #8683, opened 2025-10-14
Fix/parallel unions in scenario

philemone:
Fix union invocation order when scenario have unions in parallel.

Invocation of unions (which are treated by NU as sources) was based on the order provided by the compiler (`ProcessCompiler.compileSources`).
This worked correctly for scenarios where unions were executed in series. However, when unions run in parallel, sorting alone is not enough.

In the worst case, a scenario could run without any visible errors, but internally call the wrong union definition — resulting in computations being performed using expressions from another parallel union. :fire: 
In the best case, we received an error like: `key 'expressionName' not found`


engine/lite/request-response/runtime/src/test/scala/pl/touk/nussknacker/engine/requestresponse/test/RequestResponseTestMainSpec.scala
```diff
@@ -207,6 +217,56 @@ class RequestResponseTestMainSpec extends AnyFunSuite with Matchers with BeforeA
...
+        ContextIdPathPart(NodeId(unionName), varValue),
+        ContextIdPathPart(NodeId("unionBeforeEnd"), unionName)
+      )
+      ContextId(ProcessName("unionTest"), NodeId("start"), 0, 0, contextIdPath)
+    }
+
+    // Scenario go to 'left' nodes, where we have split and later union -> hence two variables
+    val forLeftUnion = runTest(process, ScenarioTestData(List(createTestRecord("left", ""))))
+
+    forLeftUnion.nodeResults.get(NodeId("unionRight")) should not be defined
```
raphaelsolarski: `leftUnionResults`/`rightUnionResults`?

engine/lite/request-response/runtime/src/test/scala/pl/touk/nussknacker/engine/requestresponse/test/RequestResponseTestMainSpec.scala
```diff
@@ -207,6 +217,56 @@ class RequestResponseTestMainSpec extends AnyFunSuite with Matchers with BeforeA
     endNodeIdInvocationResult.value shouldBe variable(List("aa", "bb"))
   }
 
+  test("scenario with multiple unions should work properly") {
+    val process = fromJsonUnsafe(Source.fromResource("scenarios/scenario-with-multiple-unions.json").mkString)
```
raphaelsolarski: it is hard to see how the scenario looks like just by looking on the file (ofc it can be imported into NU).
I wonder whether it'll be more readable if we could build it using `ScenarioBuilder` (it can be exported back using `.asJson` if somebody would like to import it into NU).

WDYT?

Corresponding scenario build expression is something like that:
```scala
val process = ScenarioBuilder
  .requestResponse("unionTest", "unionTest")
  .additionalFields(properties = Map(
    "inputSchema" -> """{"type": "object", "properties": {"field1": {"type": "string"}, "field2": {"type": "string"}}}""",
    "outputSchema" -> "{}"
  ))
  .sources(
    GraphBuilder
      .source("start", "request")
      .filter("filter", "#input.field1 == 'right'".spel,
        nextFalse = {
          GraphBuilder.buildSimpleVariable("variable 1", "var1", "1".spel)
            .split("Split",
              GraphBuilder.buildSimpleVariable("leftVariableB", "leftVariableB", "'leftVariableB'".spel)
                .branchEnd("leftVariableB", "unionLeft"),
              GraphBuilder.buildSimpleVariable("leftVariableA", "leftVariableA", "'leftVariableA'".spel)
                .branchEnd("leftVariableA", "unionLeft"),
            )
        })
      .filter("filter 1", "true".spel, nextFalse = {
        GraphBuilder.buildSimpleVariable("rightVariableA", "rightVariableA", "'rightVariableA'".spel)
          .branchEnd("rightVariableA", "unionRight")
      })
      .buildSimpleVariable("rightVariableB", "rightVariableB", "'rightVariableB'".spel)
      .branchEnd("rightVariableB", "unionRight"),
    GraphBuilder
      .join("unionLeft", "union", Some("unionLeft"),
        branchParams = List(
          "leftVariableA" -> List("Output expression" -> "#leftVariableA".spel),
          "leftVariableB" -> List("Output expression" -> "#leftVariableB".spel)
        )
      )
      .branchEnd("unionLeft", "unionBeforeEnd"),
    GraphBuilder
      .join("unionRight", "union", Some("unionRight"),
        branchParams = List(
          "rightVariableA" -> List("Output expression" -> "#rightVariableA".spel),
          "rightVariableB" -> List("Output expression" -> "#rightVariableB".spel)
        )
      )
      .branchEnd("unionRight", "unionBeforeEnd"),
    GraphBuilder
      .join("unionBeforeEnd", "union", Some("unionBeforeEnd"),
        branchParams = List(
          "unionRight" -> List("Output expression" -> "#unionRight".spel),
          "unionLeft" -> List("Output expression" -> "#unionLeft".spel)
        )
      )
      .emptySink("Response", "response", "Raw editor" -> "false".spel, "Value" -> "#unionBeforeEnd".spel),
  )
```

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
```diff
@@ -168,7 +168,11 @@ object ScenarioInterpreterFactory {
       capabilityTransformer: CapabilityTransformer[F],
   )(implicit ec: ExecutionContext, shape: InterpreterShape[F]) {
     // we collect errors and also typing results of sinks
-    type CompilationResult[K] = ValidatedNel[ProcessCompilationError, WithSinkTypes[K]]
+    private type CompilationResult[K] = ValidatedNel[ProcessCompilationError, WithSinkTypes[K]]
+
+    private type PartsComputedToPassAndRun = (List[JoinResult], List[PartResult], List[PartResult])
```
raphaelsolarski: maybe lets create some `case class` for `PartsComputedToPassAndRun` and `JoinCompilationResult`? For me 's hard to distinguish what is a difference between 2nd and 3rd part in PartsComputedToPassAndRun tuple.

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
```diff
@@ -362,50 +367,69 @@ object ScenarioInterpreterFactory {
...
+                  val branchesToInvoke = results.map { p => (BranchId(p.reference.branchId), p.context) }
+                  joinPartToInvoke(JoinDataBatch(branchesToInvoke)).map(_ |+| resultsToPassThrough)
+              }
             }
           )
         }
       }
     }
 
+    private def computePartsToRunAndPass(parts: List[PartResult], nodeId: NodeId): PartsComputedToPassAndRun = {
```
raphaelsolarski: wdyt about such renames?
`computePartsToRunAndPass` -> `partitionPartResultsAccordingToCurrentNode`
`nodeId` -> `currentNodeId`
`parts` -> `partResults`

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
```diff
@@ -362,50 +367,69 @@ object ScenarioInterpreterFactory {
 
     // First we compute scenario parts compiled so far. Then we search for JoinResults and invoke joinPart
     // We know that we'll find all results pointing to join, because we sorted the parts
+    // Where join are placed in parallel sorting can be not enough - see `computePartsToRunAndPass`
```
raphaelsolarski: Will such a comment be true? (I am writing as I understand it, so correct me if I am wrong)

> due to `PartSort.sort` done in `ProcessCompilerBase.compileSources` we have guarantee that `join part` will be put in order **after** parts that that references to them (in branches). Still, because of possibility to create nested/parallel unions, we have to invoke currently iterated join **only** for  `part results` that **refer to current join**, and **pass other results further** (they should be catch by joins which were put later due to topological sort).

raphaelsolarski (commented):
A lot of good work. I left a few comments.

engine/lite/request-response/runtime/src/test/scala/pl/touk/nussknacker/engine/requestresponse/test/RequestResponseTestMainSpec.scala
philemone: Much better - replaced :+1:

engine/lite/request-response/runtime/src/test/scala/pl/touk/nussknacker/engine/requestresponse/test/RequestResponseTestMainSpec.scala
philemone: That is the comment I like -> not only proposition but ready part to use :heart:

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
philemone: Sure, I've replaced type with case class :+1:

engine/lite/runtime/src/main/scala/pl/touk/nussknacker/engine/lite/ScenarioInterpreterFactory.scala
philemone: This shows that you really understand it :+1: I've used your comment since it might be much better for another person in this section.

raphaelsolarski (approved):

raphaelsolarski:
@philemone I added fixup to tests 2.12 step -> https://github.com/TouK/nussknacker/pull/8707
 I hope it will help.

PR #8687, opened 2025-10-17
Handle null value for key parameter in aggregators

mateuszkp96:
## Describe your changes

The key parameter passed to Flink's data-stream keyBy function requires a non-null key. In some places (KeyValueMappers), the null key was replaced with an empty string, which caused the data stream elements with null keys to be added to the global aggregate.  
When the null key was not replaced and was passed to the data-stream keyBy function, it caused the scenario to crash.
In this PR, the behavior changed to filter out elements with a null key, and for such an element, a dedicated NotTransientException is thrown.

## Checklist before merge

### DIFF
--- a/engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
+++ b/engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
@@ -94,6 +94,10 @@ object keyed {
         options: KeyOptions
     ): Either[InterpretationError.NullKeyError, T] = {
       Either.cond(
+        // Validate the key selector before using it in a keyed transformation.
+        // Flink's keyed stateful transformations require a non-null key selector.
+        // Passing a null key selector would cause the scenario to crash.
+        // This check can be bypassed if `allowNullableKeys` is explicitly enabled.
         options.allowNullableKeys || key != null,
         key,
         InterpretationError.NullKeyError(keyParameterName)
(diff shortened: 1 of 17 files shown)


engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/sampleTransformers.scala
```diff
@@ -60,6 +61,7 @@ object sampleTransformers {
       val windowDuration = Duration(length.toMillis, TimeUnit.MILLISECONDS)
       transformers.slidingTransformer(
         groupBy,
+        ParameterName("groupBy"),
```
lciolecki: Can we move this name to const?

engine/flink/components/base/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/DelayTransformer.scala
```diff
@@ -32,7 +33,7 @@ class DelayTransformer extends CustomStreamTransformer with ExplicitUidInOperato
         Option(key)
           .map { _ =>
             stream
-              .groupBy(key)(ctx)
+              .groupBy(key, ParameterName("key"))(ctx)
```
lciolecki: Again, let's move param name to const.

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/keyed.scala
```diff
@@ -43,99 +45,163 @@ object keyed {
...
+      final case class NullKeyError(parameterName: ParameterName) extends InterpretationError
+    }
+
+    def checkKey[T](
+        key: T,
+        keyParameterName: ParameterName,
+        options: KeyOptions
+    ): Either[InterpretationError.NullKeyError, T] = {
+      Either.cond(
+        options.allowNullableKeys || key != null,
```
arkadius, line 101: Let's add a comment, what happens when `null` appears in this place during keyed stateful transformation in Flink

arkadius (approved):
Nice, one small comment added

PR #8691, opened 2025-10-23
[NU-2382] Allow to enforce changesCanReloadParameters is false, even for the last parameter "in batch"

arkadius:
## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
@@ -93,7 +93,7 @@ object Parameter {
       hintText = None,
       labelOpt = None,
       category = ParameterCategory.Standard,
-      changesCanReloadParameters = false,
+      changesCanReloadParameters = None,
       nonImportantForExecution = false,
     )
 
@@ -118,7 +118,7 @@ object Parameter {
       hintText = None,
       labelOpt = None,
       category = ParameterCategory.Standard,
-      changesCanReloadParameters = false,
+      changesCanReloadParameters = None,
       nonImportantForExecution = false,
     )
 
@@ -152,7 +152,7 @@ case class Parameter(
     hintText: Option[String],
     labelOpt: Option[String],
     category: ParameterCategory,
-    changesCanReloadParameters: Boolean,
+    changesCanReloadParameters: Option[Boolean],
     nonImportantForExecution: Boolean
 ) extends NodeDependency {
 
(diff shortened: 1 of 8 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
```diff
@@ -152,7 +152,7 @@ case class Parameter(
     hintText: Option[String],
     labelOpt: Option[String],
     category: ParameterCategory,
-    changesCanReloadParameters: Boolean,
+    changesCanReloadParameters: Option[Boolean],
```
raphaelsolarski, line 155: WDYT about using some sealed trait instead of `Option[Boolean]` on `Parameter` level?
```scala
sealed trait OtherParametersReloadPolicy
case object CanReloadOtherParams extends OtherParametersReloadPolicy
case object CanNotReloadOtherParams extends OtherParametersReloadPolicy
case object ForcedToCanNotReloadOtherParams extends OtherParametersReloadPolicy
```
?

Now, to understand what specific `Option[Boolean]` states mean somebody have to read `TransformationStepsProcessor.processRemainingTransformationSteps` implementation.

Ideally, we could have separate a `ExternalParameter` exposed to our `ComponentProvider` api and which would be used only to express "Component Developer" intent and a separated `InternalParameter` used e.g. by `DynamicNodeValidator` and other designer's internal classes.

raphaelsolarski (approved):

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/NodeDependency.scala
arkadius: I don't understand what you mean by `ForcedToCanNotReloadOtherParams`. Is it a value for `None` or for `Some(false)`?

The `Option`-based approach is consistent with other "features" of parameters.
`None` means "determine the value internally by the compiler"
See:
- `defaultValue` - `None` means that the expression's default value will be determined based on either parameter's type, editor or optionality
-  `label` - `None` means that the label will be the same as `name`

In this case, `None` means that the `compiler` infers if parameter changes can reload other parameters.

> Ideally, we could have separate a ExternalParameter exposed to our ComponentProvider api and which would be used only to express "Component Developer" intent and a separated InternalParameter used e.g. by DynamicNodeValidator and other designer's internal classes.

See a comment next to the `Parameter` class starting with "This class is currently a part of the component API but also used as our domain model class...", We know about this debt.

PR #8702, opened 2025-10-29
[NU-2359] Live data retention time

mgoworko:
## Describe your changes

This PR introduces retention time for live data samples. Samples older than configured treshold will not be returned in the API and shown in GUI.

## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/ModelConfig.scala
```diff
@@ -45,6 +45,7 @@ object ModelConfig {
 
     final case class Enabled(
         maxNumberOfRecords: Int,
+        retentionTimeInMinutes: Int,
```
arkadius: 1. Hocon supports the `Duration`. We don't have to force the user to do the math.
2. Let's write a short notice that this retention is applied to the data preview, but counts are still computed cumulatively from the beginning

mgoworko: 1. Done, also for the older setting `windowSizeSeconds` renamed to `windowSize`, with backwards compatibility
2. Done, in changelog

arkadius (approved):

arkadius (approved):

PR #8707, opened 2025-10-31
fixup: e2e tests fixup (docker-compose postgres volume + freeing up disk space)

raphaelsolarski:
Context
* Postgres>=18 requires mounting volume to `/var/lib/postgresql` without `data` dir. The docker compose is also used in e2e tests, so this led to scala 2.12 test step crashing.
* Our e2e tests are likely running out of disk space, which is further manifested by truncated logs because we print logs for all containers in case of failed e2e environment start.

Changes:
* docker-compose files which uses postgres now have mounted only `/var/lib/postgresql` directory
* I've added the removal of unused resources before testing, just like we do for releases. This adds an additional ~6-minute build overhead, but for now, it's better than randomly failing the pipeline.


philemone (approved):

arkadius (approved):

philemone:
Good job :+1:  if it is possible we can merge it so I can rebase and merge my changes as well :)

PR #8713, opened 2025-11-06
PoC: Add traceId at Context

lciolecki:
## Checklist before merge

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/ContextTypeHelpers.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/ContextTypeHelpers.scala
@@ -15,7 +15,13 @@ object ContextTypeHelpers {
     ConcreteCaseClassTypeInfo(
       ("id", contextIdInfo),
       ("variables", variables),
-      ("parentContext", parentCtx)
+      ("parentContext", parentCtx),
+      (
+        "traceId",
+        new OptionTypeInfo[String, Option[String]](
+          TypeInformation.of(classOf[String])
+        )
+      )
     )
 
   def infoFromVariablesAndParentOption(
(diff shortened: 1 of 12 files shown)


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/typeinformation/internal/ContextTypeHelpers.scala
```diff
@@ -15,7 +15,13 @@ object ContextTypeHelpers {
     ConcreteCaseClassTypeInfo(
       ("id", contextIdInfo),
       ("variables", variables),
-      ("parentContext", parentCtx)
+      ("parentContext", parentCtx),
+      (
+        "traceId",
+        new OptionTypeInfo[String, Option[String]](
+          TypeInformation.of(classOf[String])
+        )
```
philemone, line 23: It would be nice to move this to some private def / val, simillar to `contextIdInfo`

engine/lite/components-api/src/main/scala/pl/touk/nussknacker/engine/lite/api/utils/sources/BaseLiteSource.scala
```diff
@@ -32,8 +32,14 @@ trait BaseLiteSource[T] extends LiteSource[T] with Lifecycle {
...
-        .fromEither(Try(transform(record)).toEither)
-        .map(contextVariables => context.withVariables(contextVariables.variables))
+        .fromEither(Try((transform(record), traceId(record))).toEither)
+        .map { case (contextVariables, traceId) =>
+          val contextWithVariables = context.withVariables(contextVariables.variables)
+
+          traceId
+            .map(contextWithVariables.withTraceId)
+            .getOrElse(contextWithVariables)
+        }
```
philemone, line 42: I might be just me, but it is confusing to mix this method `def traceId(record: T)` with variable `traceId`. maybe we should call this method `extractTraceId` since, I guess, it is what it does? WDYT?

philemone (commented):
Looks cool, but only if it will be applied to `staging` branch as well :slightly_smiling_face:

engine/lite/components-api/src/main/scala/pl/touk/nussknacker/engine/lite/api/utils/sources/BaseLiteSource.scala
lciolecki: Yea yea, it's just a PoC.. I though about the same name ;)

PR #8714, opened 2025-11-06
[NU-2393] Add MultiSelect Editor

mslabek:
## Checklist before merge


components-api/src/main/java/pl/touk/nussknacker/engine/api/editor/Editor.java
```diff
@@ -19,6 +19,14 @@
      */
     LabeledExpression[] possibleValues() default {};
 
+    /**
+     * This field should be used only with {@link EditorType#MULTI_SELECT_EDITOR}
+     */
+    // TODO: renames: - rename
```
raphaelsolarski: are we doing something about it?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -265,6 +265,63 @@ case object JsonValidator extends ParameterValidator {
 
 }
 
+case class MultiSelectFixedValuesValidator(possibleSelectValues: List[MultiSelectFixedValue])
+    extends ParameterValidator {
+
+  val possibleValues: List[String] = possibleSelectValues.map(_.value)
+
+  override def isValid(paramName: ParameterName, expression: Expression, value: Option[Any], label: Option[String])(
```
raphaelsolarski, line 275: maybe we can extract a few methods from here, wdyt?

```scala
  override def isValid(paramName: ParameterName, expression: Expression, value: Option[Any], label: Option[String])(
    implicit nodeId: NodeId
    import scala.jdk.CollectionConverters._
    value match {
      case Some(elements: java.util.List[_]) =>
        ensureOnlyStringElements(paramName, elements.asScala.toList)
          .andThen(stringElements => ensureOnlyPossibleElements(paramName, stringElements))
      case Some(other) => invalid(invalidFormatError(s"Expected a List, got value: $other", paramName, nodeId))
      case None =>
        invalid(
          invalidFormatError(s"Unexpected value from parsing expression: ${expression.expression}", paramName, nodeId)
        )
    }
  }

  private def ensureOnlyStringElements(paramName: ParameterName,
                                       elements: List[_])(
                                        implicit nodeId: NodeId
                                      ): Validated[PartSubGraphCompilationError, List[String]] = {
    import scala.collection.compat._
    val (stringsElements, nonStringElements): (List[String], List[Any]) = elements.partitionMap {
      case s: String => Left(s)
      case other => Right(other)
    }
    nonStringElements.headOption
      .map(nonStringValue => invalid(
        invalidFormatError(
          s"Expected only String type elements in JSON List, got: ${nonStringValue.toString}",
          paramName,
          nodeId
        )
      ))
      .getOrElse(valid(stringsElements))
  }

  private def ensureOnlyPossibleElements(parameterName: ParameterName, elements: List[String])(
    implicit nodeId: NodeId
    elements.find(!possibleValues.contains(_))
      .map(unallowedValue => invalid(unallowedValueError(parameterName, nodeId, unallowedValue)))
      .getOrElse(valid())
  }

  private def unallowedValueError(parameterName: ParameterName, nodeId: NodeId, unallowedValue: String) = {
    MultiSelectUnallowedValue(
      unallowedValue,
      possibleSelectValues,
      parameterName,
      nodeId
    )
  }

```

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -265,6 +265,63 @@ case object JsonValidator extends ParameterValidator {
 
 }
 
+case class MultiSelectFixedValuesValidator(possibleSelectValues: List[MultiSelectFixedValue])
+    extends ParameterValidator {
+
+  val possibleValues: List[String] = possibleSelectValues.map(_.value)
```
raphaelsolarski: private?

raphaelsolarski (approved):

components-api/src/main/java/pl/touk/nussknacker/engine/api/editor/Editor.java
mslabek: Not sure about this, removed it

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
mslabek: I've made some changes the logic, but according to your proposed style. Thanks :+1:

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
mslabek: Done :+1:

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/service/MultiSelectEditorService.scala
```diff
@@ -0,0 +1,29 @@
...
+      @ParamName("multiSelectParam")
+      @Editor(
+        `type` = EditorType.MULTI_SELECT_EDITOR,
+        possibleMultiSelectValues = Array(
+          new MultiSelectLabeledValue(value = "option1", label = "option1"),
+          new MultiSelectLabeledValue(value = "option2", label = "option2")
+        )
+      )
+      @Editor(`type` = EditorType.JSON_EDITOR)
+      multiSelect: Any,
```
raphaelsolarski: Correct me if I'm wrong, but I think `io.circe.Json` can be used here instead of `any`

engine/flink/tests/src/test/scala/pl/touk/nussknacker/devmodel/MultiSelectEditorTest.scala
```diff
@@ -0,0 +1,188 @@
...
+      @Editor(
+        `type` = EditorType.MULTI_SELECT_EDITOR,
+        // Only String values may be configured using annotation based component API
+        possibleMultiSelectValues = Array(
+          new MultiSelectLabeledValue(value = "option1", label = "option1"),
+          new MultiSelectLabeledValue(value = "option2", label = "option2")
+        )
+      )
+      @Editor(`type` = EditorType.JSON_EDITOR)
+      multiSelect: Any,
```
raphaelsolarski: same like here -> https://github.com/TouK/nussknacker/pull/8714/files#r2554971616

components-api/src/main/java/pl/touk/nussknacker/engine/api/editor/MultiSelectLabeledValue.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MultiSelectLabeledValue {
+
+    String value();
```
raphaelsolarski, line 26: 1. I'm not sure whether value as not escaped json string is a valid option for now - but maybe it's more of discussion about whether annotation-based api should allow more complex types than strings values now.
2. Whatever we choose, maybe let's add some java doc/comment here what we expect in the value and possibly what are the limitations at the moment.

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/service/MultiSelectEditorService.scala
mslabek: Good catch, changed to circe.Json :+1:

engine/flink/tests/src/test/scala/pl/touk/nussknacker/devmodel/MultiSelectEditorTest.scala
mslabek: changed it :+1:

components-api/src/main/java/pl/touk/nussknacker/engine/api/editor/MultiSelectLabeledValue.java
mslabek: This is not the first occasion where annotation based API is less powerful than `Parameter` from `DynamicComponent`. For now, I'd leave it as is, since escaping the value could be annoying for strings. When someone needs it, we can change this to handle complex types, but I don't expect it. I've added a javadoc with references to example components.

PR #8779, opened 2025-12-09
Fix OpenAPI enricher config parsing

mslabek:
## Describe your changes
Previously there was a bug where `secret` / `secrets` fields weren't parsed at all. Tests were using different parsing method, so this behaviour wasn't visible in tests.

## Checklist before merge


philemone (approved):
Awesome! Can we also add some note in `OpenAPI.md` about the way we read this config? I mean sth like
> When no `security` is defined in the config the first secret matching expected security schema type will be used'

WDYT? does it make sense?

mslabek:
> Awesome! Can we also add some note in `OpenAPI.md` about the way we read this config? I mean sth like
> 
> > When no `security` is defined in the config the first secret matching expected security schema type will be used'
> 
> WDYT? does it make sense?

I've added a `Configuring authentication` doc section that should clarify this. Its hard to describe, so I used an example. I've also added a test that checks if `security` takes precedence over `secret` if both match, since I wanted to add it to the doc

PR #8782, opened 2025-12-10
Fix live data counts in node directly after fragment

mgoworko:
## Describe your changes

The live data counts for the node that immediately follows the fragment were doubled. It was because there are two separate transitions between nodes:
- the transition from the fragment start (which represents the entire fragment in the scenario graph displayed on FE) to the node after fragment
- the transition from the fragment output to the node after fragment

Live data mechanism needs both transitions to properly display the live data flows and samples. But it interfered with counts, the first of those transitions must be ignored when counts are calculated

## Checklist before merge


jedrz (approved):
LGTM :+1:

PR #8790, opened 2025-12-11
[NU-2414] Flink: Usage of Flink's SourceFunction replaced by Flink's Source + some deprecation clean-ups

arkadius:
## Checklist before merge


engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/BlockingQueueSource.scala
```diff
@@ -38,44 +38,17 @@ class BlockingQueueSource[T](returnType: TypingResult, extractTimestampFun: T =>
...
   ): DataStream[Context] = {
     env
-      .addSource(flinkSourceFunction, TypeInformationDetection.instance.forType[T](returnType))
-      .name(s"${flinkNodeContext.metaData.name}-${flinkNodeContext.nodeId}-source")
+      .fromSource(
+        new BlockingQueueFlinkSource[T](id, extractTimestampFun, maxOutOfOrderness),
+        WatermarkStrategy.noWatermarks(),
+        s"${flinkNodeContext.metaData.name}-${flinkNodeContext.nodeId}-source",
+        TypeInformationDetection.instance.forType[T](returnType)
+      )
```
piotrp, line 51: Shouldn't we also assign a UID, for the sake of consistency, and for people who may be copying this code to make their own source?

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/CollectionSource.scala
```diff
@@ -48,37 +47,38 @@ case class CollectionSource[T](
...
-    boundedness match {
-      case Boundedness.BOUNDED =>
-        env.fromData(recordsListDependingOnBoundedness(list).asJava, typeInformation)
-      case Boundedness.CONTINUOUS_UNBOUNDED =>
-        env.addSource(new FromElementsFunction[T](recordsListDependingOnBoundedness(list).asJava), typeInformation)
+  ): Source[T, _, _] = {
+    val typeInformation  = TypeInformationDetection.instance.forType[T](returnType)
+    val listWithoutNulls = handleNullsInData(list)
+    // DataGeneratorSource which an empty list try to process one element
+    // which ends up with NoSuchElementException in FromElementsGeneratorFunction.tryDeserialize
```
piotrp: Maybe:
```scala
    // DataGeneratorSource will throw NoSuchElementException from FromElementsGeneratorFunction.tryDeserialize
    // when created with an empty element list
```
?

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/CollectionSource.scala
```diff
@@ -48,37 +47,38 @@ case class CollectionSource[T](
...
-      //       FromElementsFunction or checking if input list has nulls and fast failing instead
-      records.filterNot(_ == null)
+  // FromElementsGeneratorFunction doesn't accept nulls in the passed collection - see constructor
+  // so we filter them out
+  // TODO: we probably should resolve this problem otherwise either by delivering our own GeneratorFunction
+  //       or checking this collection on early stage, but before we do that we should check every places
+  //       where this source is used
+  private def handleNullsInData(list: List[T]) = {
+    if (list.contains(null)) {
+      logger.warn("Collection passed to CollectionSource contains null. They will be filtered out")
```
piotrp: Will our tests pass if we throw an exception here?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -177,16 +177,17 @@ object EventGeneratorSourceFactory
...
+            WatermarkStrategy.noWatermarks(),
+            flinkNodeContext.nodeId.id,
+          )
           .setUidAndNameToNodeId(flinkNodeContext.nodeId)
           .flatMap(
-            (_: Unit, out: Collector[Boolean]) => {
-              // This 'true' is a dummy value. It has to exist for each event, but its value is completely ignored.
-              // The type is not important too, it just has to be serializable by Flink.
-              // For each of those dummy values, a new Context will be created.
-              (1 to localCount).foreach(_ => out.collect(true))
```
piotrp, line 187: Maybe let's keep this comment, but with "This 'true'" changed to "This 'null'"?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -250,6 +251,22 @@ object EventGeneratorSourceFactory
     }
   }
 
+  private def flinkSource(schedule: Duration) = {
+    new DataGeneratorSource[Void](
+      (_: java.lang.Long) => null,
+      Long.MaxValue,
+      new RateLimiterStrategy {
```
piotrp: Would `RateLimiterStrategy.perSecond(1.second / schedule)` work here? If I'm reading the source right the only difference is that `perSecond` takes parallelism into account, and our implementation has always assumed parallelism=1.

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/transformers.scala
```diff
@@ -114,7 +111,8 @@ object transformers {
             new UnwrappingAggregateFunction[AnyRef](aggregator, aggregateBy.returnType, identity)
           val offsetMillis = windowOffset.getOrElse(Duration.Zero).toMillis
           val windowDefinition =
-            TumblingEventTimeWindows.of(Time.milliseconds(windowLength.toMillis), Time.milliseconds(offsetMillis))
+            TumblingEventTimeWindows
```
piotrp, line 115: I believe our types here are wrong. Windows have finite durations, so `windowLength` should be a `FiniteDuration`, and that would allow us to use `toJava` from `import scala.compat.java8.DurationConverters._`. 

With this import you could also handle `java.time.Duration.ofMillis` (`Time.milliseconds`) without writing out full package name.

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/sink/TableSink.scala
```diff
@@ -70,7 +69,9 @@ class TableSink(
       Flink docs show something like this when integrating table api with inserts into dataStream. For details read:
       https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/data_stream_api/.
      */
-    dataStream.addSink(new DiscardingSink())
+    // TODO Variant with SinkV2 causes java.math.BigDecimal cannot be cast to class java.lang.Integer in tests
```
piotrp, line 72: Will you try to fix it here or in next PR?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/scenariotesting/DataRecordsSource.scala
```diff
@@ -62,20 +71,16 @@ class DataRecordsSource(
       .map((ctxWithEventTime: ContextWithEventTime) => ctxWithEventTime.context, flinkNodeContext.contextTypeInfo)
   }
 
-  // env.fromData uses DataGeneratorSource which even for an empty list try to process one element
-  // which ends up with NoSuchElementException in FromElementsGeneratorFunction.tryDeserialize
-  private def fromDataProperHandlingEmptyList(env: StreamExecutionEnvironment): DataStreamSource[DataRecord] = {
+  private def flinkSource(executionConfig: ExecutionConfig): Source[DataRecord, _, _] = {
+    // DataGeneratorSource which an empty list try to process one element
+    // which ends up with NoSuchElementException in FromElementsGeneratorFunction.tryDeserialize
```
piotrp: Same as previous instance of this comment:
```scala
    // DataGeneratorSource will throw NoSuchElementException from FromElementsGeneratorFunction.tryDeserialize
    // when created with an empty element list
```
?

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/StaticSource.scala
```diff
@@ -1,47 +1,57 @@
...
     }
   }
 
-  sealed trait Timer
+  sealed trait Command
+
+  case class EmitWatermark(time: Long) extends Command
 
-  case class Watermark(time: Long) extends Timer
+  case class CollectData(time: Long, value: String) extends Command
```
piotrp: This makes sense when you look at implementation, but wouldn't `EmitData` match public API better? Data is "collected" by source's ReaderOutput, but the source "emits" this data.

Example of current usage, taken from tests:
```
    StaticSource.add(CollectData(1000, "1"))
    StaticSource.add(EmitWatermark(3000))
    StaticSource.add(CollectData(2000, "2"))
    StaticSource.add(EmitWatermark(8000))
```

And while we're here: maybe let's use the full name for timestamp, i.e. `time` -> `timestamp`?

engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
```diff
@@ -162,11 +146,12 @@ class FlinkKafkaSource[K, V](
...
-  @nowarn("cat=deprecation")
-  protected def flinkSourceFunction(
-      consumerGroupId: String,
+  private def flinkSource(
       flinkNodeContext: FlinkCustomNodeContext
-  ): SourceFunction[ConsumerRecord[K, V]] = {
+  ): Source[ConsumerRecord[Array[Byte], Array[Byte]], _, _] = {
+    val consumerGroupId = prepareConsumerGroupId(flinkNodeContext)
+
+    // TODO setStartingOffsets(OffsetsInitializer) instead of using KafkaUtils
```
piotrp: This PR / next / we're keeping this TODO?

Would this work?
```scala
    val offsetsInitializer = offsetResetStrategy match {
      // not sure about this case at the moment - force EARLIEST or use NONE (consumer default)?
      case OffsetResetStrategy.None => OffsetsInitializer.committedOffsets(FlinkOffsetResetStrategy.EARLIEST)
      case OffsetResetStrategy.ToEarliest => OffsetsInitializer.earliest()
      case OffsetResetStrategy.ToLatest => OffsetsInitializer.latest()
    }
```

engine/flink/management/dev-model/src/main/scala/pl/touk/nussknacker/engine/management/sample/source/NoEndingSource.scala
```diff
@@ -46,32 +52,30 @@ class NoEndingSource(val implementWatermarkStrategyForTest: Boolean)
...
+
+        override def pollNext(output: ReaderOutput[String]): InputStatus = {
           if (afterFirstRun.getAndSet(true)) {
-            ctx.collect("TestInput" + r.nextInt(10))
+            output.collect("TestInput" + r.nextInt(10))
           } else {
-            ctx.collect("TestInput1")
+            output.collect("TestInput1")
           }
           Thread.sleep(2000)
```
piotrp: Let's use `RateLimiter` instead?

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/BlockingQueueSource.scala
arkadius: Good point, changed

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
arkadius: Wow, I didn't expect that it could handle fractional numbers as well. Changed

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/transformers.scala
arkadius: Changed here and in other places

engine/flink/components/table/src/main/scala/pl/touk/nussknacker/engine/flink/table/sink/TableSink.scala
arkadius: In the next PR, it looks like some deeper problem that I need to investigate

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/util/source/StaticSource.scala
arkadius: Good point, changed

engine/flink/kafka-components-utils/src/main/scala/pl/touk/nussknacker/engine/kafka/source/flink/FlinkKafkaSource.scala
arkadius: I resolved this TODO in this change. Let's check if all tests pass.  `FlinkOffsetResetStrategy.NONE` won't work for us. It doesn't mean "use auto.offset.reset from kafka client properties" or "use some default" but instead, it explicitly means "fail if there is no offset". For backward compatibility, we have to check auto.offset.reset from kafka client properties.
[edit]
All tests passed

piotrp (approved):

PR #8802, opened 2025-12-16
[NU-2432] SQL enrichers: sql metadata uses connection pooling

arkadius:
## Checklist before merge

### DIFF
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
@@ -19,7 +19,9 @@ object MetaDataProviderFactory {
           // We limit the pool to 1 connection instead of using original dbPoolConfig.maxTotal because we
           // have two separate pools (one for metadata and one for runtime) and we want to reduce the risk
           // that we exceed the limit of connections
-          maxTotal = 1
+          maxTotal = 1,
+          // We don't want to monitor this pool because it is not used operationally
+          registerMbeans = false
         ),
         poolName = dbEnricherName.value + "-metadata"
       )
--- a/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
+++ b/components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
@@ -1,24 +1,16 @@
 package pl.touk.nussknacker.sql.db.schema
 
 import pl.touk.nussknacker.engine.util.ThreadUtils
+import pl.touk.nussknacker.sql.DbEnricherName
 import pl.touk.nussknacker.sql.db.ignite.IgniteMetaDataProvider
 import pl.touk.nussknacker.sql.db.pool.{DBPoolConfig, HikariDataSourceFactory}
-import pl.touk.nussknacker.sql.db.schema.MetaDataProviderFactory.igniteDriverPrefix
 
 import java.sql.Connection
-import java.util.Properties
 
 object MetaDataProviderFactory {
   private val igniteDriverPrefix = "org.apache.ignite.IgniteJdbc"
-}
 
-class MetaDataProviderFactory {
-
-  def create(dbPoolConfig: DBPoolConfig): JdbcMetaDataProvider = {
-    val props = new Properties()
-    dbPoolConfig.dataSourceProperties.foreach { case (k, v) =>
-      props.put(k, v)
-    }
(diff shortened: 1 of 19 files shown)


components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
```diff
@@ -21,12 +20,15 @@ class MetaDataProviderFactory {
       props.put(k, v)
     }
     val ds = ThreadUtils.withContextClassLoader(getClass.getClassLoader) {
-      new DriverDataSource(
-        dbPoolConfig.url,
-        dbPoolConfig.driverClassName,
-        props,
-        dbPoolConfig.username,
-        dbPoolConfig.password
+      HikariDataSourceFactory(
```
piotrp, line 15: After your changes the `props` variable is no longer used, you can remove it

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
```diff
@@ -21,12 +20,15 @@ class MetaDataProviderFactory {
...
+        dbPoolConfig.copy(
```
piotrp, line 16: We are using this pool config potentially twice (here + for data access), so we should take care of generating a unique `poolName`, otherwise metrics will be broken.

Now that I look further, we should add `poolName` and `registerMbeans` properties to our `DBPoolConfig` and `HikariDataSourceFactory` classes.

components/sql/src/test/scala/pl/touk/nussknacker/sql/service/DatabaseConnectionPoolingTest.scala
```diff
@@ -0,0 +1,159 @@
...
+    )
+  )
+
+  override def prepareHsqlDDLs: List[String] = List(
+    "CREATE TABLE people (id INT, name VARCHAR(40));",
+    "INSERT INTO people (id, name) VALUES (1, 'John');",
+  )
+
+  test("DatabaseQueryEnricher should use connection pooling during scenario validation and run") {
+    ConnectionCountingDriver.clearInvocationsCount()
```
piotrp, line 66: This should be in `beforeEach`

piotrp (changes requested):

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
arkadius: I agree, these pool names were named `HikariPool-1` and `HikariPool-2` (depending on which was created first) - now they are explicitly named. 

I didn't add `poolName` to `DBPoolConfig` because it is a DTO for configuration. Instead, I used the enricher name + suffix. Do you think it is ok?

Why do you think that we should add `registerMbeans` to the configuration? You want to not expose mbeans for the pool created for metadata purposes? I wonder if it should be in the configuration or in `HikariDataSourceFactory.apply` parameters.

piotrp: I want to expose `registerMbeans` because it's disabled by default: https://github.com/brettwooldridge/HikariCP/wiki/MBean-(JMX)-Monitoring-and-Management (or at least it is according to docs)

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
```diff
@@ -1,32 +1,29 @@
...
+          // We always keep one idle connection to decrease the time needed to compile a node
+          initialSize = 1,
+          // We limit the pool to 1 connection instead of using original dbPoolConfig.maxTotal because we
+          // have two separate pools (one for metadata and one for runtime) and we want to reduce the risk
+          // that we exceed the limit of connections
+          maxTotal = 1,
+          // We don't want to monitor this pool because it is not used operationally
```
piotrp: But we may be interested in its activity on UI, and this monitoring is basically free?

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
arkadius: Ahh, now I'm getting it!

components/sql/src/main/scala/pl/touk/nussknacker/sql/db/schema/MetaDataProviderFactory.scala
arkadius: Sure, I removed this. Now the configuration option is also applied to this pool

PR #8803, opened 2025-12-16
Drop cross-compile, use only Scala 2.13

mgoworko:
## Checklist before merge

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,16 +14,14 @@ import scala.util.Try
 import scala.xml.Elem
 import scala.xml.transform.{RewriteRule, RuleTransformer}
 
-val scala212 = "2.12.20"
 val scala213 = "2.13.16"
 
 lazy val defaultScalaV = sys.env.get("NUSSKNACKER_SCALA_VERSION") match {
   case None | Some("2.13") => scala213
-  case Some("2.12")        => scala212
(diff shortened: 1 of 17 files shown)


build.sbt
```diff
@@ -14,16 +14,14 @@ import scala.util.Try
 import scala.xml.Elem
 import scala.xml.transform.{RewriteRule, RuleTransformer}
 
-val scala212 = "2.12.20"
```
arkadius, line 17: Adding an initial support for scala 3.x instead of removing support for scala 2.12 would be much harder?

mgoworko: Adding Scala 3 cross-compile to the code is not something extra hard, but it is also not as simple as adding version number to build.sbt, it would take a few days to complete
  - there are some libraries, that do not support Scala 3, or require higher version to support Scala 3 (for example kind-projector, circe-generic-extras - does not exist for Scala 3, needs to be replaced, etc)
  - there are some syntactic changes that need to be done in order to cross compile
  - we would have to verify, if it works correctly with Flink, or maybe dedicated adaptations are neededfor Scala 3
  
Overall - it is not a rocket science, but it should be a separate task, with analysis first (and verifying Flink compat). I think right now it would be best to just use single version - Scala 2.13

arkadius (approved):

PR #8806, opened 2025-12-17
[NU-2399] Running test cases

raphaelsolarski:
#### As part of this PR is implemented:
* Running a TestCase on a scenario (endpoint).
  * The scenario must already exist, but you can run a test case on a draft — it is passed as a graph.
* Using the input data defined in the test case (we enforce a common format for them).
* Replacing mocks in the provided scenario based on the mocks defined in the test case.
* Validating assertion expressions defined in the test case, based on the typing derived from the compiled scenario.
* Passing the scenario with substituted mocks to the DM for execution.
* Verifying per-node assertions based on TestResults (which has been extended to include a non-serialized version of originalNodeResults).
* A basic assertEquals assertion that performs light type alignment (e.g., normalizes array vs. list to a common type), but only on the first level (no nested structures yet — TBD), and returns differences as SpEL literals (`SpelValuePrettyPrinter.prettyPrintValue`).

#### Side notes
* Thanks to preparing test data/mocks and verifying assertions using TestResults, executing a test case is independent of the DMs (the existing logic is sufficient), which is IMO a significant advantage (no lite/flink/RR specific logic).
* This may be insufficient if the DM runs tests on remote environments (e.g., a Flink-based DM executing tests on a remote cluster).
* In that case, we would need proper serialization of TestResults.originalNodeResults or move assertion verification from ScenarioTestService to the execution environment.
* TestCase has the same limitations as the existing tests — it runs during an HTTP request and must finish before that request’s read timeout triggers.
An alternative approach would be submitting tests and polling (or using WebSockets) for results.
* Missing: persistence & validation flow.

### DIFF
--- a/scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/CanonicalProcess.scala
+++ b/scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/CanonicalProcess.scala
@@ -83,7 +83,7 @@ case class CanonicalProcess(
     // DON'T use these fields, rely on allStartNodes or mapAllNodes instead.
     nodes: List[CanonicalNode],
     additionalBranches: List[List[CanonicalNode]] = List.empty,
-    stickyNotes: List[StickyNote] = List.empty,
+    stickyNotes: List[StickyNote] = List.empty
 ) extends CanonicalTreeNode {
 
   import CanonicalProcess._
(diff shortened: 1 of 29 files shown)


scenario-api/src/main/scala/pl/touk/nussknacker/engine/canonicalgraph/CanonicalProcess.scala
```diff
@@ -78,12 +78,12 @@ object CanonicalProcess {
 
 //in fact with branches/join this form is not canonical anymore - graph can be represented in more than way
 case class CanonicalProcess(
-    metaData: MetaData,
-    // separation of nodes and additionalBranches is just for compatibility of stored json
-    // DON'T use these fields, rely on allStartNodes or mapAllNodes instead.
-    nodes: List[CanonicalNode],
-    additionalBranches: List[List[CanonicalNode]] = List.empty,
-    stickyNotes: List[StickyNote] = List.empty,
+                             metaData: MetaData,
```
jedrz: Could you reformat the changes? It looks like the indentation doesn't match the configured scalafmt.

jedrz (approved):

PR #8840, opened 2025-12-23
[NU-2399] Save a single test case along with the scenario graph

jedrz:
## Checklist before merge

### DIFF
--- a/designer/client/cypress/fixtures/fragmentWithInput.json
+++ b/designer/client/cypress/fixtures/fragmentWithInput.json
@@ -193,5 +193,5 @@
     ],
     "additionalBranches": [],
     "stickyNotes": [],
-    "testCase": null
+    "testCases": null
 }
--- a/designer/client/cypress/fixtures/fragmentWithInput.json
+++ b/designer/client/cypress/fixtures/fragmentWithInput.json
@@ -192,5 +192,6 @@
         }
     ],
     "additionalBranches": [],
-    "stickyNotes": []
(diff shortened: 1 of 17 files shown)


designer/client/cypress/fixtures/fragmentWithInput.json
```diff
@@ -192,5 +192,6 @@
         }
     ],
     "additionalBranches": [],
-    "stickyNotes": []
+    "stickyNotes": [],
+    "testCase": null
```
Dzuming: `testCase` or `testCases`? In the future, for enterprise users, it will be a list of test cases

jedrz: `testCase`. However, as you pointed out, I will make it a wrapper to make it easier to add a list variant, with test cases available only for enterprise users.

jedrz: In the end, I renamed the field to `testCases` (plural). And the single test case is encoded as:
```
"testCases": {
  "value": {
    "id": "62bed328-d1f0-404e-8f1c-b275db4cb08b",
    "name": "test case name",
  }
}
```

It should be possible in the future to extend this structure to support multiple test cases:
```
"testCases": {
  "list": [
     {
      "id": "62bed328-d1f0-404e-8f1c-b275db4cb08b",
      "name": "test case name",
     }
  ]
}
```

philemone (approved):
LGTM :+1:

PR #8893, opened 2026-01-20
UI tweaks & fixes bundled together

JulianWielga:
override List[Record{name: String, value: String}] editor for some fields

https://github.com/user-attachments/assets/04e1aa3e-398b-498d-90b5-0214b51b9592


Dzuming (approved):
LGTM

PR #8896, opened 2026-01-21
[NU-2399] Define assertions without #TESTS helper

jedrz:
## Describe your changes

Additional small improvements:
- Add `notEquals` assertion.
- Pretty print string as SpEL expression. 

## Checklist before merge

### DIFF
--- a/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/AssertionsCompiler.scala
+++ b/designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/AssertionsCompiler.scala
@@ -5,6 +5,7 @@ import cats.data.Validated.{Invalid, Valid}
 import cats.syntax.all._
 import pl.touk.nussknacker.engine.api._
 import pl.touk.nussknacker.engine.api.context.{ProcessCompilationError, ValidationContext}
+import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.api.typed.typing.{Typed, TypingResult, Unknown}
 import pl.touk.nussknacker.engine.compile.ExpressionCompiler
 import pl.touk.nussknacker.engine.test.testcase.{Assertion, TestCase}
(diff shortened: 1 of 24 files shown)


scenario-api/src/main/scala/pl/touk/nussknacker/engine/test/testcase/TestCase.scala
```diff
@@ -34,4 +34,51 @@ object TestCases {
...
+
+    val name: String = this match {
+      case AssertionOperator.Equals    => "equals"
+      case AssertionOperator.NotEquals => "notEquals"
+    }
+
+  }
+
+  object AssertionOperator {
+    case object Equals    extends AssertionOperator
```
mslabek, line 62: Some boilerplate could be avoided by using enumeratum, like here https://github.com/TouK/nussknacker/blob/6a1189766c9f5d1e63352c8054de9475993d7e76/designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/scenarioTesting/Dtos.scala#L116 . That would also be more type safe since now if developer adds a new AssertionOperator, he could forget to add it in `assertionOperatorDecoder`.

Same with `PredicateAssertionCompilationError.Field`.

I prefer the enumeratum approach, but imo either one is fine, up to you.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/AssertionsCompiler.scala
```diff
@@ -86,71 +82,109 @@ class AssertionsCompiler(
       )
   }
 
-  private def compileAssertionExpression(nodeId: NodeId, context: ValidationContext, assertion: Assertion) = {
+  private def compileAssertion(
+      nodeId: NodeId,
```
mslabek: This class could be more readable if we passed the NodeId parameter implicitly. Almost all methods require it and we pass the NodeId in multiple places like this in this codebase.

mslabek (approved):

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/AssertionsCompiler.scala
jedrz: Good catch, thanks!

scenario-api/src/main/scala/pl/touk/nussknacker/engine/test/testcase/TestCase.scala
jedrz: Thanks, switched to `Enumeratum`. I needed to add this dependency to the `scenario-api` module.

PR #8897, opened 2026-01-21
Fix: Passing TraceId after collect in RequestResponse

lciolecki:
## Checklist before merge

### DIFF
--- a/engine/lite/components/request-response/src/main/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/CollectTransformer.scala
+++ b/engine/lite/components/request-response/src/main/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/CollectTransformer.scala
@@ -10,6 +10,7 @@ import pl.touk.nussknacker.engine.api.typed.typing.Typed
 import pl.touk.nussknacker.engine.lite.api.commonTypes.{DataBatch, ResultType}
 import pl.touk.nussknacker.engine.lite.api.customComponentTypes.{CustomComponentContext, LiteCustomComponent}
 
+import scala.collection.immutable.::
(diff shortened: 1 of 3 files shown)


engine/lite/components/request-response/src/main/scala/pl/touk/nussknacker/engine/lite/components/requestresponse/CollectTransformer.scala
```diff
@@ -49,9 +51,19 @@ class CollectTransformer(outputVariable: String, inputExpression: LazyParameter[
     lazy val contextIdGenerator = runtimeContext.contextIdGenerator(context.nodeId)
     (inputCtx: DataBatch) =>
       val outputList = inputCtx.map(inputExpression.evaluate).asJava
+
+      // FIXME: This is an ugly workaround for the fact that we can't pass traceId from org Context to collect.
+      val tracesId = inputCtx.value.flatMap(_.traceId)
```
piotrp: Let's do this without additional HashSet+List allocations and O(n) `size` calls?
```scala
      // FIXME: This is an ugly workaround for the fact that we can't pass traceId from org Context to collect.
      val ctxTracesId = inputCtx.value.collectFirst(_.traceId) match {
        case Some(traceIdOpt) if inputCtx.value.forall(_.traceId == traceIdOpt) => traceIdOpt
        // this should never happen
        case Some(_) => logger.warn(s"Inconsistent traceId in context: ${inputCtx.value.flatMap(_.traceId)}"); None
        case _ => None
      }
```
I threw in an additional logging so that we may know if our invariant ever stops being one

piotrp (changes requested):

piotrp (approved):

PR #8904, opened 2026-01-23
Properly handle InterruptedException in SyncInterpretationFunction

mateuszkp96:
## Describe your changes

This change introduces a new extension method `unsafeRunTimedAttempt` for IO, based on Cats Effect’s IO.unsafeRunTimed, but returning structured information about the reason for termination (timeout or interruption).

Previously, both timeout and thread interruption resulted in a TimeoutException with the message "Interpreter is running too long" in SyncInterpretationFunction, making it impossible to distinguish the real cause of termination.

With this change:

* SyncInterpretationFunction returns proper ExceptionInfo only for timeout error, interruption error is not propagated to the exceptionHandler
* Additional node details are now passed to NuExceptionInfo to improve error diagnostics.

## Checklist before merge

### DIFF
--- a/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
+++ b/engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
@@ -1,6 +1,9 @@
 package pl.touk.nussknacker.engine.process.registrar
 
+import cats.data.Validated
+import cats.data.Validated.{Invalid, Valid}
 import cats.effect.IO
+import com.typesafe.scalalogging.LazyLogging
 import org.apache.flink.api.common.functions.RichFlatMapFunction
 import org.apache.flink.util.Collector
 import pl.touk.nussknacker.engine.InterpretationResult
@@ -30,7 +33,8 @@ private[registrar] class SyncInterpretationFunction(
     nodeComponentInfo: NodeComponentInfo,
     useIOMonad: Boolean
 ) extends RichFlatMapFunction[Context, InterpretationResult]
-    with ProcessPartFunction {
+    with ProcessPartFunction
+    with LazyLogging {
 
   private lazy val compiledNode =
     compilerData.compileSubPart(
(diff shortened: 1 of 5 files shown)


engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
```diff
@@ -61,9 +64,12 @@ private[registrar] class SyncInterpretationFunction(
...
       compilerData.interpreter
         .interpret[IO](compiledNode, compilerData.jobData, input, serviceExecutionContext)
-        .unsafeRunTimed(compilerData.processTimeout) match {
-        case Some(result) => result
-        case None =>
+        .unsafeRunTimedAttempt(compilerData.processTimeout) match {
+        case Right(result) =>
+          result
+        case Left(Error.Interrupted) =>
+          throw new RuntimeException(s"Interpreter action was interrupted")
```
lciolecki: But we still throw here an exception?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/util/IoUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+    private def unsafeRunTimed(limit: FiniteDuration)(implicit runtime: unsafe.IORuntime): Either[Error, A] = {
+      val queue = new ArrayBlockingQueue[Either[Throwable, A]](1)
+
+      action.unsafeRunAsync { r =>
+        queue.offer(r)
+        ()
+      }
+
+      try {
+        val result = blocking(queue.poll(limit.toNanos, TimeUnit.NANOSECONDS))
```
lciolecki, line 37: result is Either[Throwable, A], are you sure about comparing it to null?

lciolecki (approved):
Nice one!

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/util/IoUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+  sealed trait Error
+
+  object Error {
+    case object Timeout     extends Error
+    case object Interrupted extends Error
+  }
+
+  implicit class SyncIoOps[A](val action: IO[A]) extends AnyVal {
+
+    def unsafeRunTimedAttempt(limit: FiniteDuration)(implicit runtime: unsafe.IORuntime): Either[Error, A] = {
```
lciolecki, line 20: Do we need 2 methods? This one just forwards calling.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
piotrp: Can we throw an `InterruptedException` here?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/util/IoUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+        queue.offer(r)
+        ()
+      }
+
+      try {
+        val result = blocking(queue.poll(limit.toNanos, TimeUnit.NANOSECONDS))
+        if (result eq null) Left(Error.Timeout) else result.fold(throw _, Right(_))
+      } catch {
+        case _: InterruptedException =>
+          Left(Error.Interrupted)
```
piotrp, line 41: Can we add a test that verifies that our wrapper actually returns this `Error.Interrupted`?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/util/IoUtils.scala
mateuszkp96: It's code taken from `cats.effect`. Internally, `queue.poll` returns null in case of timeout.

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/registrar/SyncInterpretationFunction.scala
```diff
@@ -81,4 +94,8 @@ private[registrar] class SyncInterpretationFunction(
     }
   }
 
+  private def exceptionInfo(error: Throwable, input: Context): NuExceptionInfo = {
```
lciolecki: You can pass here List(Right( ... ) ) as well. Then invoking will be shorter and you will get rid of boilerplate?

engine/flink/executor/src/main/scala/pl/touk/nussknacker/engine/process/util/IoUtils.scala
mateuszkp96: I wanted to stay as close as possible to the method definition from cats-effect IO. When we update the library, we should update this code snippet.

piotrp (approved):

PR #8941, opened 2026-01-29
[NU-2399] Endpoint to generate enricher mock expression

jedrz:
## Checklist before merge


designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/SpelExpressionGenerator.scala
```diff
@@ -0,0 +1,115 @@
...
+    case _: Unknown =>
+      Some("null")
+
+    case TypedNull =>
+      Some("null")
+
+    case TypedTaggedValue(underlying, _) =>
+      generateForTypingResult(underlying, indentLevel)
+
+    case TypedObjectWithValue(underlying, _) =>
```
mslabek: For this case know the value - we could just render the value instead of generating it. But then you'd have to worry about things like escaping characters for values, so its up to you.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/SpelExpressionGenerator.scala
```diff
@@ -0,0 +1,115 @@
...
+      generateForTypingResult(elementType, indentLevel) match {
+        case Some(elementExpr) => Some(s"{$elementExpr}")
+        case None              => Some("{}")
+      }
+
+    case TypedClass(JavaEnumConstants(firstEnumConstant :: _), _) =>
+      Some(s"T(${typedClass.klass.getName}).${firstEnumConstant.name()}")
+
+    case TypedClass(clazz, _) if clazz == InstantClass =>
+      Some("T(java.time.Instant).parse('1900-01-01T00:00:00Z')")
```
mslabek: For dates, UUID, or even numbers we could generate values for implicit conversions. So like String "1900-01-01T00:00:00" for LocalDateTime or String "00000000-0000-0000-0000-000000000000" for UUID. We wouldn't have to use scary T(something.something.Class) just simple types to be implicitly converted. But its more 'magical' and could lead to some misunderstandings since its not out best explained feature. But imo in this case it fits perfectly. WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/SpelExpressionGenerator.scala
```diff
@@ -0,0 +1,115 @@
...
+    case _ =>
+      None
+  }
+
+  private def generateForRecord(fields: ListMap[String, TypingResult], indentLevel: Int): Option[String] = {
+    val fieldExpressions = fields.mapValuesNow(generateForTypingResult(_, indentLevel + 1).getOrElse("null"))
+    val recordExpression = if (fieldExpressions.isEmpty) {
+      "{:}"
+    } else if (fieldExpressions.size == 1) {
+      fieldExpressions.map { case (key, value) => s"$key: $value" }.mkString("{", ", ", "}")
```
mslabek: keys also have to be treated as strings since not all valid keys can be without parentheses. For example with like: 'ab-cd' - it would result in error. Also these keys could contain " or '

jedrz: Good catch, thx :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/SpelExpressionGenerator.scala
jedrz: Done.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/test/testcase/SpelExpressionGenerator.scala
jedrz: Done.

mslabek (approved):

PR #8945, opened 2026-01-30
Move E2E tests to separate job

piotrp:
## Describe your changes

Moves E2E tests to a separate job:
* allows _Tests_ to pass, or randomly fail, a bit sooner
* removes background downloads of multiple Docker images - less random IO and CPU usage spikes should increase chance of some time/IO-sensitive tests to pass

And most importantly - _Tests_ job passes, proving to me that my recent changes didn't break staging...

## Checklist before merge


.github/workflows/pr.yml
```diff
@@ -177,48 +177,41 @@ jobs:
...
 
+  e2eTests:
+    name: E2ETests
+    runs-on: ubuntu-latest
+    needs: [ build, setup ]
+    strategy:
+      matrix:
+        scalaVersion: [2.13]
+    env:
+      NUSSKNACKER_SCALA_VERSION: ${{ matrix.scalaVersion }}
```
lciolecki, line 200: Hmm, we have got rid of scala 2.12, so we don't need to run matrix test right now?

.github/workflows/pr.yml
```diff
@@ -177,48 +177,41 @@ jobs:
...
-      - name: Test Report
+      - name: Upload test report
         if: success() || failure()
         uses: actions/upload-artifact@v6
         with:
           name: BackendTests-results-${{ matrix.scalaVersion }}
           path: '**/test-reports/*.xml'
 
+  e2eTests:
+    name: E2ETests
```
lciolecki: Let's be consistent - Backend e2e tests?

.github/workflows/pr.yml
piotrp: I kept it because it doesn't hurt us, and with our fully custom Scala serialization it is possible to update to Scala 3 in the future

.github/workflows/pr.yml
piotrp: Done

lciolecki (approved):

PR #8974, opened 2026-02-16
Add control over strict type matching for fragments

mateuszkp96:
## Checklist before merge


designer/client/src/components/graph/node-modal/fragment-input-definition/settings/variants/fields/ValidationFields.tsx
```diff
@@ -88,6 +90,27 @@ export default function ValidationFields({
                     />
                 </div>
             </FormControl>
+            <FormControl>
```
jedrz, line 93: WDYT about adding a setting flag and a global default so that the `strictTypeChecking` checkbox could be shown only when the setting flag is on ?

PR #8987, opened 2026-02-19
[NU-2399] Add optional assertion description

jedrz:
## Checklist before merge

### DIFF
--- a/scenario-api/src/main/scala/pl/touk/nussknacker/engine/test/testcase/TestCase.scala
+++ b/scenario-api/src/main/scala/pl/touk/nussknacker/engine/test/testcase/TestCase.scala
@@ -38,13 +38,16 @@ sealed trait Assertion
 
 object Assertion {
 
-  @JsonCodec final case class ExpressionAssertion(expression: Expression, description: Option[String]) extends Assertion
+  @JsonCodec final case class ExpressionAssertion(
+      expression: Expression,
+      description: Option[String] = None,
+  ) extends Assertion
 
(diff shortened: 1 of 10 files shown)


scenario-api/src/main/scala/pl/touk/nussknacker/engine/test/testcase/TestCase.scala
```diff
@@ -38,10 +38,14 @@ sealed trait Assertion
...
-
-  @JsonCodec final case class PredicateAssertion(operator: AssertionOperator, expected: Expression, actual: Expression)
-      extends Assertion
+  @JsonCodec final case class ExpressionAssertion(expression: Expression, description: Option[String]) extends Assertion
+
+  @JsonCodec final case class PredicateAssertion(
+      operator: AssertionOperator,
+      expected: Expression,
+      actual: Expression,
+      description: Option[String]
```
piotrp: You use a lot of `description = None`, maybe we should add this as a default here, or add a secondary constructor without this argument.? Then tests would be simpler.

jedrz: Good point, especially since we aren't doing much with the description on the backend anyway.

piotrp: I'm giving 👍 , but check diff again - there are some places where you are introducing only formatting changes

piotrp (approved):

PR #8997, opened 2026-02-23
[NU-2468] Introduce NodeId + NodeName for all nodes

philemone:
## Checklist before merge


components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -493,7 +493,7 @@ object ProcessCompilationError {
   final case class NodeIdValidationError(errorType: IdErrorType, override val nodeId: NodeId)
       extends IdError
       with InASingleNode {
-    override val id: String = nodeId.id
+    override val id: String = nodeId.value
```
lciolecki, line 534: Why it's a String?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/NodeId.scala
```diff
@@ -2,16 +2,21 @@ package pl.touk.nussknacker.engine.api
 
 import io.circe.{Decoder, Encoder, KeyDecoder, KeyEncoder}
 
-case class NodeId(id: String) {
-  override def toString: String = id
+import java.util.UUID
+
+// Holds a UUID-format string generated once at node creation
+final case class NodeId(value: String) {
```
lciolecki, line 8: Why it's still String? Why don't make it just UUID?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/PdfExporter.scala
```diff
@@ -312,8 +312,8 @@ object PdfExporter extends LazyLogging {
           <table-row>
               <table-cell border="1pt solid black" padding-left="1pt" font-weight="bold">
                 <block>
-                  <basic-link internal-destination={node.id}>
-                    {node.id}
+                  <basic-link internal-destination={node.name.value}>
```
lciolecki: Shouldn't be the id here?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/NodeId.scala
philemone: Since UUID would lock as for specific type (we introduced value class for now) also it is much smoother to make such a big change for now, we can discuss it if you want though,.

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+
+object NodeIdToUuidMigration extends ProcessMigration {
+
+  override def description: String = "Migrate node IDs to node name, introduce nodeId as static uuid"
+
+  override def migrateProcess(canonicalProcess: CanonicalProcess, category: String): CanonicalProcess = {
+    // BranchEndData nodes have synthetic IDs ($edge-...), not real node IDs — skip them
+    val idMapping: Map[String, String] = canonicalProcess.collectAllNodes
+      .filterNot(_.isInstanceOf[BranchEndData])
+      .map(n => n.id.value -> UUID.randomUUID().toString)
```
piotrp: > Migrate node IDs to node name, introduce nodeId as static uuid

So, do we assign random UUIDs or copy nodeName to nodeId? JSON decoding uses node names, so maybe we should be consistent here, and just copy name to id?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -196,12 +205,50 @@ class UIProcessValidator(
...
+        case Some(node) => Some(nodeId -> node.name.value)
+        case None =>
+          canonicalNodesById.keys
+            .find(outerNodeId => nodeId.startsWith(outerNodeId + "-"))
+            .flatMap { outerNodeId =>
+              val outerNode   = canonicalNodesById(outerNodeId)
+              val innerNodeId = nodeId.drop(outerNodeId.length + 1)
+              outerNode match {
+                case fi: FragmentInput =>
+                  fragmentsByName.get(fi.ref.id).flatMap { fragmentCanonical =>
```
mslabek: Nested flatMaps - maybe some for-comprehension or extract some of it to seperate methods?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -196,12 +205,50 @@ class UIProcessValidator(
...
+                    fragmentCanonical.collectAllNodes
+                      .find(_.id.value == innerNodeId)
+                      .map(innerNode => nodeId -> s"${outerNode.name.value} - ${innerNode.name.value}")
+                  }
+                case _ => None
```
mslabek: What does this case mean? Illegal state?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -196,12 +205,50 @@ class UIProcessValidator(
...
+  )(implicit loggedUser: LoggedUser): Map[String, String] = {
+    val referencedNodeIds =
+      (result.errors.invalidNodes.keys ++ result.warnings.invalidNodes.keys).map(_.value).toSet
+
+    if (referencedNodeIds.isEmpty) return Map.empty
+
+    val canonicalNodesById = canonical.collectAllNodes.map(n => n.id.value -> n).toMap
+
+    lazy val fragmentsByName =
+      fragmentResolver.fetchFragmentsSync(processingType).map(f => f.name.value -> f).toMap
```
mslabek: This db call can be removed - in `validateCanonicalProcess` theres a `resolveFragments` call which fetches fragments, just doesn't return them. We could make `resolveFragments` return them and pass it here.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -196,12 +205,50 @@ class UIProcessValidator(
...
+          finalResult.withNodeNames(resolveNodeNamesForResult(finalResult, canonical))
         }
       }
       .unsafeRunSync()
   }
 
+  private def resolveNodeNamesForResult(
+      result: ValidationResult,
+      canonical: CanonicalProcess
+  )(implicit loggedUser: LoggedUser): Map[String, String] = {
```
mslabek: Maybe return `Map[NodeId, NodeName]`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -196,12 +205,50 @@ class UIProcessValidator(
...
+
+    lazy val fragmentsByName =
+      fragmentResolver.fetchFragmentsSync(processingType).map(f => f.name.value -> f).toMap
+
+    referencedNodeIds.flatMap { nodeId =>
+      canonicalNodesById.get(nodeId) match {
+        case Some(node) => Some(nodeId -> node.name.value)
+        case None =>
+          canonicalNodesById.keys
+            .find(outerNodeId => nodeId.startsWith(outerNodeId + "-"))
```
mslabek: if we allow non-UUID ids, using `startsWith` could result in collisions. I can write some test to reproduce it on Monday. If I'm not mistaken, there would be a couple of ways to solve it, we should talk it over.

designer/client/src/types/validation.ts
```diff
@@ -13,6 +13,7 @@ export type ValidationResult = {
     nodeResults: NodeResults;
     errors: ValidationErrors;
     warnings?: Pick<ValidationErrors, "invalidNodes">;
+    nodeNames?: Record<string, string>;
```
mslabek: Why nullable? Isn't it always present? Also same for NodeType.name. If we assume that `name` is mandatory, the FE code can be greatly simplified.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/ValidationResults.scala
```diff
@@ -23,7 +23,8 @@ object ValidationResults {
   @JsonCodec final case class ValidationResult(
       errors: ValidationErrors,
       warnings: ValidationWarnings,
-      nodeResults: Map[String, NodeTypingData]
+      nodeResults: Map[String, NodeTypingData],
+      nodeNames: Map[String, String] = Map.empty
```
mslabek, line 27: May be a stupid question, but why can't FE build a mapping between id and names? Is it something about nodes from fragments?

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
```diff
@@ -92,6 +92,17 @@ object CirceUtil {
 
   }
 
+  // For backward compatibility: wraps a decoder so that when the "name" field is missing in JSON,
```
mslabek: If we migrate persisted scenarios by setting random UUIDs as ids, maybe we should also set it in the decoder? Or the other way around - migrate the persisted scenarios by setting id = name for backwards compatibility?

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+import java.util.UUID
+
+object NodeIdToUuidMigration extends ProcessMigration {
+
+  override def description: String = "Migrate node IDs to node name, introduce nodeId as static uuid"
+
+  override def migrateProcess(canonicalProcess: CanonicalProcess, category: String): CanonicalProcess = {
+    // BranchEndData nodes have synthetic IDs ($edge-...), not real node IDs — skip them
+    val idMapping: Map[String, String] = canonicalProcess.collectAllNodes
+      .filterNot(_.isInstanceOf[BranchEndData])
```
mslabek: Maybe not skip it, but for this BranchEndData case instead of generating UUID, just leave the old id explicitly? Then the map would contain all nodes and could be used safely directly without the need for  `mapId: String => String`

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    def mapId(oldId: String): String = idMapping.getOrElse(oldId, oldId)
+
+    canonicalProcess.mapAllNodes(rewriteNodes(_, mapId))
+  }
+
+  private def rewriteNodes(nodes: List[CanonicalNode], mapId: String => String): List[CanonicalNode] =
+    nodes.map(rewriteNode(_, mapId))
+
+  private def rewriteNode(node: CanonicalNode, mapId: String => String): CanonicalNode = node match {
```
mslabek: The `asInstanceOf` can be removed if the rewrite would be inlined for these cases. Like: 
```scala
  private def rewriteNode(node: CanonicalNode, mapId: String => String): CanonicalNode = {
    def id(n: NodeData): NodeId = NodeId(mapId(n.id.value))

    node match {
      case canonicalnode.FlatNode(data) =>
        val newData = data match {
          case n: Source                   => n.copy(id = id(n))
          case n: Join                     => n.copy(id = id(n), branchParameters = n.branchParameters.map(bp => bp.copy(branchId = mapId(bp.branchId))))
          case n: Filter                   => n.copy(id = id(n))
          case n: Switch                   => n.copy(id = id(n))
          case n: VariableBuilder          => n.copy(id = id(n))
          case n: Variable                 => n.copy(id = id(n))
          case n: Split                    => n.copy(id = id(n))
          case n: Enricher                 => n.copy(id = id(n))
          case n: CustomNode               => n.copy(id = id(n))
          case n: Processor                => n.copy(id = id(n))
          case n: Sink                     => n.copy(id = id(n))
          case n: FragmentInput            => n.copy(id = id(n))
          case n: FragmentUsageOutput      => n.copy(id = id(n))
          case n: FragmentInputDefinition  => n.copy(id = id(n))
          case n: FragmentOutputDefinition => n.copy(id = id(n))
          case n: BranchEndData            => BranchEndData(BranchEndDefinition(mapId(n.definition.id), mapId(n.definition.joinId)))
        }
        canonicalnode.FlatNode(newData)
      case canonicalnode.FilterNode(data, nextFalse) =>
        canonicalnode.FilterNode(data.copy(id = id(data)), rewriteNodes(nextFalse, mapId))
      case canonicalnode.SwitchNode(data, nexts, defaultNext) =>
        canonicalnode.SwitchNode(
          data.copy(id = id(data)),
          nexts.map(c => c.copy(nodes = rewriteNodes(c.nodes, mapId))),
          rewriteNodes(defaultNext, mapId)
        )
      case canonicalnode.SplitNode(data, nexts) =>
        canonicalnode.SplitNode(data.copy(id = id(data)), nexts.map(rewriteNodes(_, mapId)))
      case canonicalnode.Fragment(data, outputs) =>
        canonicalnode.Fragment(
          data.copy(id = id(data)),
          outputs.map { case (k, v) => k -> rewriteNodes(v, mapId) }
        )
    }
  }
```
but not necessary, up to you.

designer/client/src/components/graph/node-modal/nodeIdFieldHelpers.tsx
```diff
@@ -16,8 +15,9 @@ function isEditingNodeId(node: EditedNode | NodeType): node is EditedNode {
 
 export function applyIdFromFakeName(node: EditedNode): NodeType {
     if (!isEditingNodeId(node)) return node;
-    const { [FAKE_NAME_PROP_NAME]: name, ...rest } = node;
-    return { ...rest, [PROP_NAME]: name ?? node[PROP_NAME] };
+    const { [FAKE_NAME_PROP_NAME]: editedName, ...rest } = node;
```
mslabek, line 18: I think this whole name-id-juggling mess was necessary because name was id. Maybe this can be removed now?

designer/client/src/http/HttpService/types.ts
```diff
@@ -94,7 +94,9 @@ export type DeployResponse = {
 
 export type NodeUsageData = {
     fragmentNodeId?: string;
+    fragmentNodeName?: string;
```
mslabek, line 97: Is `fragmentNodeName` and `nodeName` correct here? On BE the `FragmentUsageData` and  `ScenarioUsageData` wchich are only subtypes of `NodeUsageData` don't have these fields

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeValidationExceptionHandler.scala
```diff
@@ -35,7 +35,7 @@ object NodeValidationExceptionHandler extends LazyLogging {
         Validated.invalidNel(CustomNodeError(exc.message, exc.paramName))
       case NonFatal(e) =>
         logger.error(
-          s"Exception during validation handling of node '${nodeId.id}' in scenario ${metaData.name.value}",
+          s"Exception during validation handling of node '${nodeId.value}' in scenario ${metaData.name.value}",
           e
         )
         Validated.invalidNel(CannotCreateObjectError(e, nodeId))
```
mslabek: For user-facing errors like this, the message should contain node name instead of node id

Image

designer/client/src/components/tips/Warnings.tsx
```diff
@@ -49,19 +49,22 @@ const Warnings = ({ warnings, showDetails, scenarioGraph }: WarningsProps) => {
...
+                                return (
+                                    <Typography
+                                        variant={"body2"}
+                                        fontWeight={"bold"}
+                                        component={LinkStyled}
+                                        key={uuid4()}
+                                        to={""}
+                                        onClick={(event) => showDetails(event, node)}
+                                    >
+                                        {node?.name ?? warning.key}
```
mslabek, line 63: Here, if we stay with nullable `name`, maybe instead of `.key` we could get the node name by looking up `validationResult.nodeNames` ? But I'm strongly for making `.name` mandatory so that wouldn't be an issue then.

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -228,7 +275,7 @@ class UIProcessValidator(
 
   private def validateNodesId(scenarioGraph: ScenarioGraph): ValidationResult = {
     val nodeIdErrors = scenarioGraph.nodes
-      .map(n => IdValidator.validateNodeId(NodeId(n.id)))
+      .map(n => IdValidator.validateNodeId(n.id))
```
mslabek: If we were to limit ids to UUIDs by type constraints or some simple UUID validation, we could get rid of this validator. 

Also FE still has some leftover mandatory node name validation which highlights the input field red. I think its tied to the field name in the form.

Image

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
mslabek, line 144: We should have some validation for node names too. I'd argue for validating against 2 things:

1. Duplicate names - because generally user should ever see only node name. In some contexts he wont be able to differentiate between 2 nodes with the same name. For example in contexts like PDF export now or compare versions.

2. Blank names - these are problematic on FE and not permitting them is not a big limitation of users freedom

docs/Changelog.md
```diff
@@ -15,6 +15,10 @@ description: Stay informed with detailed changelogs covering new features, impro
 
 ### 1.19.0 (Not released yet)
 
+* [#8997](https://github.com/TouK/nussknacker/pull/8997) Introduce Node IDs as a stable UUIDs
```
mslabek, line 19: Lets write here or in migration guide what are breaking changes implied by generating random ids. Things like:
1. Metrics for nodes
2. Consumer groups for Kafka sources
3. URLs for nodes
4. Flink operator uids (so the state is also incompatible)

Also in migration guide it would be good to at least generally note the changes in -api modules and in our HTTP interface types.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ScenarioGraphComparator.scala
```diff
@@ -11,18 +11,18 @@ object ScenarioGraphComparator {
       currentGraph.nodes.map(node => node.id -> node).toMap,
       otherGraph.nodes.map(node => node.id -> node).toMap
     )(
-      notPresentInOther = current => NodeNotPresentInOther(current.id, current),
-      notPresentInCurrent = other => NodeNotPresentInCurrent(other.id, other),
-      different = (current, other) => NodeDifferent(current.id, current, other)
+      notPresentInOther = current => NodeNotPresentInOther(current.id.value, current),
+      notPresentInCurrent = other => NodeNotPresentInCurrent(other.id.value, other),
+      different = (current, other) => NodeDifferent(current.id.value, current, other)
```
mslabek: Here we should use node name instead of id. This gets displayed on FE directly. Or FE should somehow map ids to names. Same for edges

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
```diff
@@ -174,7 +175,13 @@ class NodesApiEndpoints(auth: EndpointInput[AuthCredentials]) extends BaseEndpoi
             List(
               Example.of(
                 value = NodeValidationRequestDto(
-                  Filter("id", Expression(Language.Spel, "#longValue > 1"), isDisabled = None, additionalFields = None),
+                  Filter(
+                    NodeId("id"),
+                    NodeName("id"),
```
mslabek: Ideally, examples of node id for openapi should be UUIDs and names something more like name, not "id"

mslabek (commented):
Generally, theres a couple of key decisions here:
- FE compatibility with nodes without name - I think non nullable node name would greatly reduce complexity - I don't see relevant drawbacks (https://github.com/TouK/nussknacker/pull/8997#discussion_r2866851263, https://github.com/TouK/nussknacker/pull/8997#discussion_r2866668471)

- Backwards compatibility of scenario JSON - we either break compatibility by generating fresh IDs like with the current db migration or we leave the id alone and just set name to current id. Either has its problems   (https://github.com/TouK/nussknacker/pull/8997#discussion_r2866704736, https://github.com/TouK/nussknacker/pull/8997#discussion_r2866646452, https://github.com/TouK/nussknacker/pull/8997#discussion_r2866906534)

- How to approach node ids in logs / error messages (https://github.com/TouK/nussknacker/pull/8997#discussion_r2866836811)

- Validating the node name, especially against duplicates - https://github.com/TouK/nussknacker/pull/8997#discussion_r2866877717 

It would probably be most effective to discuss these personally.


Some node ids are still visible from UI - from what I noticed:

1. in input/output data records display

Image

2. in components usages 

Pasted image 20260227213200

3. in node search

Pasted image 20260227213735

4. in testing window

Pasted image 20260227213039

5. in compare versions (https://github.com/TouK/nussknacker/pull/8997#discussion_r2866911921)

Pasted image 20260227212620

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/ValidationResults.scala
piotrp: Having names in validation results is useful e.g. in cross-environment migration tests, on in scripted validation via API, so that results may be be interpreted by a human

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
piotrp: If we need UUIDs we can generate 128 bit hash of old name, flip some bits to have this value in proper UUID format, and use that. We can even make sure that it will be in a format that will never conflict with our generated IDs. Then the migrated value would be deterministic.

That deterministic ID generation of UUID from name would be useful in tests (if we decide that all IDs must be in UUID format), that way we could have stable identifiers for test nodes.

designer/client/src/components/graph/EspNode/element.ts
```diff
@@ -82,7 +82,9 @@ export const updateNodeCounts =
     (node: shapes.devs.Model): void => {
         if (userSettings["debug.dontRenderCountsOnNodes"]) return;
         const shortCounts = userSettings["node.shortCounts"];
-        const count = processCounts[node.id];
+        const nodeData = node.get("nodeData") as NodeType | undefined;
+        const lookupKey = nodeData?.name || node.id;
```
piotrp: Shouldn't we use only `id`, because scenario structure is already migrated?

Maybe we should we have a user preference that would allow to fetch counts by `name`, until user is sure that all counts are in the new format, with both `id` + `name` available?

designer/client/src/components/toolbars/search/utils.ts
```diff
@@ -17,7 +17,7 @@ type FilterSelector = { name: string; selector: Selector }[];
 const fieldsSelectors: FilterSelector = [
```
piotrp, line 17: We should be able to search node by `id` - exact match searching would be enough, just so that we can e.g. quickly find a node that we noticed in logs.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
```diff
@@ -189,7 +189,7 @@ object TestingApiErrorMessages {
       assertion: Assertion,
       nodeId: NodeId
   ) =
-    s"Assertion compilation error. Node: ${nodeId.id}. ${prettyPrintAssertion(assertion)}. Errors: ${errors.toList.mkString(", ")}"
+    s"Assertion compilation error. Node: ${nodeId.value}. ${prettyPrintAssertion(assertion)}. Errors: ${errors.toList.mkString(", ")}"
```
piotrp: We should use name here, or name + id. If possible, users shouldn't see UUIDs in error messages.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/PdfExporter.scala
```diff
@@ -244,8 +244,8 @@ object PdfExporter extends LazyLogging {
       NodeSeq.Empty
     } else {
       <block margin-bottom="25pt" margin-top="5pt">
-        <block font-size="13pt" font-weight="bold" text-align="left" id={node.id}>
-          {node.getClass.getSimpleName} {node.id}
+        <block font-size="13pt" font-weight="bold" text-align="left" id={node.name.value}>
```
piotrp: `id.value`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/PdfExporter.scala
piotrp: Setting as unresolved, I have the same question :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
```diff
@@ -214,7 +261,7 @@ class UIProcessValidator(
     }
     val disabledNodesWarnings =
       disabledNodes
-        .map(node => (NodeId(node.id), List(PrettyValidationErrors.formatErrorMessage(DisabledNode(NodeId(node.id))))))
+        .map(node => (node.id, List(PrettyValidationErrors.formatErrorMessage(DisabledNode(node.id)))))
```
piotrp: Make sure that all error messages passed to `formatErrorMessage` have name accessible, and use it when generating error message. This call will result in UI warning about saying "Node 543913b0-458c-468e-9925-821df1a063f2 is disabled".

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
piotrp: Even if we migrate all identifiers to UUIDs it's still helpful to know that no incorrect ids are being processed - let's keep it

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/util/exception/ExceptionRateMeter.scala
```diff
@@ -22,7 +22,7 @@ class ExceptionRateMeter(metricsProvider: MetricsProviderForScenario) {
   private def getMeterForNode(nodeId: NodeId): RateMeter = nodeErrorsMeterMap.getOrElseUpdate(
     nodeId,
     metricsProvider.instantRateMeterWithCount(
-      MetricIdentifier(NonEmptyList.of("error", "instantRateByNode"), Map(nodeIdTag -> nodeId.id))
+      MetricIdentifier(NonEmptyList.of("error", "instantRateByNode"), Map(nodeIdTag -> nodeId.value))
```
piotrp: I believe we should still have some kind of name validation (so that our IDs can be e.g. correct InfluxDB tags), and all metrics should log both node ID and name, or our dashboards will be useless. This applies to all instances where we use `nodeIdTag` - we should add `nodeNameTag` and always set both.

On that note, we will need to update our dashboards to use `name`, not `id`, when displaying values.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/PdfExporter.scala
philemone: Yup, should be like this:
``` 
<basic-link internal-destination={node.id.value}>
 {node.name.value}
```
updated 👍

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
philemone: Thanks, I've changed migration and this decoder to use deterministic UUID based on id field (current name). This way decoder and migration works the same way and are predictable 👍

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
philemone: Thanks, as I've wrote to @mslabek - I've changed it to `nameUUIDFromBytes` to generate deterministic uuid based on id. We have guarantee that names are unique in a scenario so there should be no conflicts and we can update tests to static uuid rather than regex pattern

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
philemone: That looks neat and have `BranchEndData` handled in same place as the rest - I will use it, thanks :)

designer/client/src/components/graph/EspNode/element.ts
philemone: We should, I would love to anyway -> sometimes I've left since I had problems on FE and it helped me investigate -> I will delete remaining fallback as this in all places. Thanks for pointing this one out.

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/ValidationResults.scala
philemone: You (both of you) are correct
(we are unable to resolve `{fragmentId}-{fragmentNodeId}` on FE without it)

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/NodesApiEndpoints.scala
philemone: True, it was lazy of me 😅 Fixed

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
philemone: Changed 👍

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Validation for name exists - just need to rename it (it was `validateNodesId` before, since name was equal to id before this MR) -> that is a good observation though, I will rename it and update, thanks :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Unfortunately it could be little misleading since this is the sketchy place where, for validation errors, NU connects nodeName from fragments and its nodes. Those could be placed in values classes `NodeId` / `NodeName` but it is not always NodeName and NodeId, but can be concated  version of those

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Thanks, I saw that on FE and I am in the middle of creating a patch :)

docs/Changelog.md
philemone: Sure, I will add it as the last touch to this MR 🙂

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Changed to for comprehensions 👍

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Kind-of, it is case which should not happen (sth is using `{uuid}-{uuid}` as nodeId but it is reserved for nodes in fragment (it is same policy as it was before changes)  -> I've added log to have some feedback

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: We can go back to it, but user never sets uuid himself - lets put a pin in this for now

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: I've changed it a little bit so we won't call it so often - one call and then pass it around - WDYT?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: I've added fix, I will check it 3 more times since with this scale of small changes I can miss sth again 🙃

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/NodeValidationExceptionHandler.scala
philemone: Fixed

utils/utils-internal/src/main/scala/pl/touk/nussknacker/engine/util/exception/ExceptionRateMeter.scala
philemone: Thanks - I've updated metrics with new tag nodeNameTag :+1:

designer/client/src/components/tips/Warnings.tsx
philemone: You are right, but it is not because name was (I've change it to be mandatory as you suggested) optional. It is because node is fetched from `getNodeById` which gets it from `nodesFromScenarioGraph` by find -> so it can be undefined -> I've added it since it was insecure before.

designer/client/src/components/graph/node-modal/nodeIdFieldHelpers.tsx
philemone: I would keep this flow for now, because it still acts as temporary modal edit state and is used in apply/validation. We can remove it safely in a separate refactor after we confirm this MR works as should?

designer/client/src/http/HttpService/types.ts
philemone: `NodesUsagesData` is a union for scenario/fragment usage data. in `ScenarioUsageData` you have `nodeName` and `nodeId` and `FragmentUsageData` has `fragment*` fields, that is why those are optional 
```
  object NodeUsageData {

    final case class FragmentUsageData(
        fragmentNodeId: String,
        fragmentNodeName: String,
        nodeId: NodeId,
        nodeName: NodeName
    ) extends NodeUsageData

    final case class ScenarioUsageData(nodeId: NodeId, nodeName: NodeName) extends NodeUsageData
  }
```

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
mslabek: Nice 👍 Can you update the comment? Something like
`For backward compatibility: wraps a decoder so that when the "name" field is missing in JSON, "name" is filled from the "id" value, and "id" is replaced with a deterministic UUID generated from that name.
`

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
mslabek: 1. The validations from NameValidator work but they dont do duplicate validation. You can check this by a test in `ProcessValidatorSpec` like 
```scala
    val sameName = NodeName("same")
    val process = CanonicalProcess(
      MetaData("process1", StreamMetaData()),
      List(
        FlatNode(Source(NodeId("1"), sameName, SourceRef("source", Nil))),
        FlatNode(Sink(NodeId("2"), sameName, SinkRef("sink", Nil))),
      )
    )
    validate(process, baseDefinition).result.isInvalid // returns true
```

2. Do we want some validation for node id? Duplicates for ids are validated in `uiValidation` but do we want to validate further?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
mslabek: Nice! Also `implicit loggedUser: LoggedUser` can now be removed from `resolveNodeNamesForResult`

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -69,13 +71,19 @@ object ProcessCompilationError {
 
   final case class DuplicatedNodeIds(nodeIds: Set[NodeId]) extends ProcessCompilationError with ScenarioGraphLevelError
```
piotrp: Shouldn't this also include names, so that we can use it in error message? So eg. `Set[(NodeId, NodeName)]`?

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
```diff
@@ -490,10 +507,10 @@ object ProcessCompilationError {
     override val id: String = name.value
   }
 
-  final case class NodeIdValidationError(errorType: IdErrorType, override val nodeId: NodeId)
+  final case class NodeNameValidationError(errorType: IdErrorType, nodeName: NodeName, override val nodeId: NodeId)
```
piotrp: Our argument order is always id -> name, let's change this to `(errorType: IdErrorType, override val nodeId: NodeId, nodeName: NodeName)`

defaultModel/src/test/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigrationSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+import java.nio.charset.StandardCharsets
+import java.util.UUID
+import scala.io.{Source => IoSource}
+
+class NodeIdToUuidMigrationSpec extends AnyFreeSpecLike with Matchers {
+
+  private def uuidOf(s: String) = UUID.nameUUIDFromBytes(s.getBytes(StandardCharsets.UTF_8)).toString
+
+  private def loadAndMigrate(resourcePath: String): CanonicalProcess = {
+    val json     = IoSource.fromResource(resourcePath).mkString
```
piotrp: Always wrap `Source` with `Using.resource` or this will hold an open file descriptor until JVM terminates.
```scala
    val json     = Using.resource(IoSource.fromResource(resourcePath))(_.mkString)
```

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
```diff
@@ -164,17 +165,19 @@ object PrettyValidationErrors {
...
-      case UnsupportedPart(id)         => node("UnsupportedPart", s"Type of node $id is unsupported right now")
-      case UnknownFragment(id, nodeId) => node("Unknown fragment", s"Node $nodeId uses fragment $id which is missing")
-      case InvalidFragment(id, nodeId) => node("Invalid fragment", s"Node $nodeId uses fragment $id which is invalid")
+      case MissingPart(_, nodeName)     => node("MissingPart", s"Node $nodeName has missing part")
+      case UnsupportedPart(_, nodeName) => node("UnsupportedPart", s"Type of node $nodeName is unsupported right now")
+      case UnknownFragment(id, _, nodeName) =>
+        node("Unknown fragment", s"Node $nodeName uses fragment $id which is missing")
+      case InvalidFragment(id, _, nodeName) =>
+        node("Invalid fragment", s"Node $nodeName uses fragment $id which is invalid")
       case FatalUnknownError(message, maybeNodeId) =>
```
piotrp: We have access to node name in at least one place where this is used (`contextAfterMethodBasedCreatedComponentExecutor`)

piotrp (changes requested):

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
```diff
@@ -6,6 +6,7 @@ import io.circe.generic.extras.Configuration
 
 import java.net.{URI, URL}
```
piotrp, line 7: We format some error messages in `ValidationErrorOps` class, but I don't see it changed in this PR

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
piotrp: id -> name (with optional id in detailed `description`?)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/NodeComponentInfo.scala
```diff
@@ -1,18 +1,23 @@
...
+      componentName: String
+  ): NodeComponentInfo = {
+    NodeComponentInfo(nodeId, nodeName, Some(ComponentId(componentType, componentName)))
   }
 
 }
 
 // componentId is not used anywhere in this repo - it is only for external project's purpose for now
 // TODO: Remove Option from componentId - every Node used in runtime is related with some Component. After doing it, we should
 //       report this id in KafkaExceptionInfo - see KafkaJsonExceptionSerializationSchema
```
piotrp, line 22: We should add nodeName to `KafkaExceptionInfo` - these are error messages available to users

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/metrics/InvocationMetrics.scala
```diff
@@ -42,7 +42,7 @@ class InvocationMetrics(context: EngineRuntimeContext) {
   private def markErrorTimer(startTime: Long, nodeId: Option[NodeId] = None): Unit = {
     val id = nodeId.getOrElse(NodeId("unknown"))
     nodeErrorTimers
-      .getOrElseUpdate(id, espTimer(Map(nodeIdTag -> id.id), NonEmptyList.of("invocation", "failure")))
+      .getOrElseUpdate(id, espTimer(Map(nodeIdTag -> id.value), NonEmptyList.of("invocation", "failure")))
```
piotrp: \+ `nodeNameTag`

defaultModel/src/main/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigration.scala
philemone: Changed :+1:

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
```diff
@@ -76,7 +76,7 @@ trait ProcessValidator extends LazyLogging {
 
     try {
       CompilationResult.map4(
-        CompilationResult(IdValidator.validate(process, isFragment)),
+        CompilationResult(NameValidator.validate(process, isFragment)),
```
piotrp, line 79: Should we add a `UniqueIdValidator`, to verify that all of our ids are truly unique?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeDataValidator.scala
```diff
@@ -70,11 +70,11 @@ class NodeDataValidator(modelData: ModelData) {
           )
         case a: FragmentInput =>
           validateFragment(validationContext, outgoingEdges, a, fragmentResolver)
-        case Split(_, _) | FragmentUsageOutput(_, _, _, _, _) | BranchEndData(_) =>
+        case Split(_, _, _) | FragmentUsageOutput(_, _, _, _, _, _) | BranchEndData(_) =>
           ValidationNotPerformed
       }
 
-      val nodeIdErrors = IdValidator.validateNodeId(NodeId(nodeData.id)) match {
+      val nodeIdErrors = NameValidator.validateNodeName(nodeData) match {
```
piotrp: `nodeIdErrors` -> `nodeNameErrors`

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
```diff
@@ -180,7 +180,7 @@ object EventGeneratorSourceFactory
           .fromSource(
             flinkSource(schedule),
             WatermarkStrategy.noWatermarks(),
-            flinkNodeContext.nodeId.id,
+            flinkNodeContext.nodeId.value,
           )
           .setUidAndNameToNodeId(flinkNodeContext.nodeId)
```
piotrp: You won't like this, but we need to always set `id` and `name` separately, because names are visible to users in JobManager's UI. So, we need to:
* add a required `name` argument to `setUidAndName` (both declarations - `DataStream` and `DataStreamSink`)
* remove `setUidAndNameToNodeId` -> migrate uses to `setUidAndName` (both declarations)

piotrp (changes requested):

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
philemone: I found it! `s"\n  $nodeId: ${nodeErrors.map(_.message)` , thanks - added nodeName for that

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
philemone: Done, thanks for ready comment to use :)

components-api/src/main/scala/pl/touk/nussknacker/engine/api/component/NodeComponentInfo.scala
philemone: Done 👍

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/TestingApiErrorMessages.scala
philemone: 🤷  I've recovered it, thanks :)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/ProcessCompiler.scala
philemone: I thought that we don't since it is editable but I've looked at the name validator and those are so simillar that I've added one with similar rules - just to be safe :)

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compile/nodecompilation/NodeDataValidator.scala
philemone: done 👍

engine/lite/request-response/runtime/src/main/scala/pl/touk/nussknacker/engine/requestresponse/metrics/InvocationMetrics.scala
philemone: 🕵️  Thanks, missed that

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/UIProcessValidator.scala
philemone: Thanks for noticing that - I had some mix-up there with validation - should work now (i've added id and node validations + unique node name as suggested)

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
philemone: Done (but I've left nodeId in details, not description)

designer/restmodel/src/main/scala/pl/touk/nussknacker/restmodel/validation/PrettyValidationErrors.scala
philemone: Done (by extending error definition)

defaultModel/src/test/scala/pl/touk/nussknacker/defaultmodel/migrations/NodeIdToUuidMigrationSpec.scala
philemone: 🤦

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/EventGeneratorSourceFactory.scala
philemone: I love it! :D I've added required name to `setUidAndName` and removed old declaration. Thanks -> those things are hard to come back to and we should stick to new model where we have both id and name.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
philemone: Updated 👍

designer/client/src/types/validation.ts
philemone: It is not anymore :)

designer/client/src/components/toolbars/search/utils.ts
philemone: I've prepared small change to filter by id as well, but I will upload it when cypress passes :D

components-api/src/main/scala/pl/touk/nussknacker/engine/api/context/ProcessCompilationError.scala
philemone: Updated, it is better with name :+1:

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
piotrp: Can we add name handling to `BadRequestTestingError.SourcesCompilationError`? It's the only remaining case when `toHumanReadableMessage` doesn't use node names.

engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaExceptionConsumerSerializationSpec.scala
```diff
@@ -33,7 +33,7 @@ class KafkaExceptionConsumerSerializationSpec extends AnyFunSuite with Matchers
   private val context = Context(ContextId.dummy, variables, None, Some(TraceId.generate()))
 
   private val exception = NuExceptionInfo(
-    Some(NodeComponentInfo(NodeId("nodeId"), ComponentType.Service, "componentName")),
+    Some(NodeComponentInfo(NodeId("nodeId"), NodeName("nodeId"), ComponentType.Service, "componentName")),
```
piotrp: Let's use `NodeName("nodeName")` here, to ensure that our tested code doesn't mix NodeId.value and NodeName.value

docs/Changelog.md
piotrp: 5. Changes in error messages on Kafka topic - new `nodeName` field and changed contents of existing `nodeId` field

engine/flink/kafka-components-utils/src/test/scala/pl/touk/nussknacker/engine/kafka/exception/KafkaExceptionConsumerSerializationSpec.scala
philemone: Done :+1:

common-api/src/main/scala/pl/touk/nussknacker/engine/api/CirceUtil.scala
philemone: Sure, added :+1:

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ScenarioTestingApiHttpService.scala
```diff
@@ -348,31 +355,37 @@ class ScenarioTestingApiHttpService(
...
 
-  private def toDto(error: ParametersDefinitionError): TestingError = {
+  private def toDto(error: ParametersDefinitionError, scenarioGraph: ScenarioGraph): TestingError = {
+    val nodeNamesById = extractNodeNamesById(scenarioGraph)
     error match {
       case ParametersDefinitionError.SourcesCompilationError(nodesWithErrors) =>
         SourcesCompilationError(
-          ValidationErrors.initial(invalidNodes = collectInvalidNodes(nodesWithErrors))
+          ValidationErrors.initial(invalidNodes = collectInvalidNodes(nodesWithErrors)),
+          nodeNamesById = nodeNamesById
```
piotrp: We should filter this to contain only required ids, so that we won't risking returning a API response with 100 id->name entries and a single error (the same goes for the second `toDto` method in this file)

philemone: Added with test for it :+1:

docs/Changelog.md
philemone: updated :+1:

designer/client/src/components/toolbars/search/utils.ts
philemone: Ok, added :)

mslabek (approved):

docs/Changelog.md
```diff
@@ -16,6 +16,10 @@ description: Stay informed with detailed changelogs covering new features, impro
 ### 1.19.0 (Not released yet)
 
 * [#8988](https://github.com/TouK/nussknacker/pull/8988) Fixed fragment test data validation for required or restricted input parameters
+* [#8997](https://github.com/TouK/nussknacker/pull/8997) Introduce Node IDs as a stable UUIDs
+  * Each scenario node now has two distinct fields: a stable `id` (UUID, never changes) and a user-editable `name` (human-readable label).
+  * Existing scenarios are migrated, the old human-readable ID becomes the node `name`, and a new UUID is generated as `id`.
+  * Introduce `nodeNames` field to `ValidationResult` response to easily translate uuid to user-friendly names. Fragment inner nodes are covered using the `"fragmentName - innerNodeName"` format.
```
piotrp, line 22: You also need to:
* adjust telegraf.conf files in example/ directory (remove "nodaName_" form measurement names, same way it is done for "nodeId_")
* adjust three Grafana dashboards example/ directory - we need to add "nodeName" to all GROUP BY expressions and replace "nodeId" with "nodeName" in aliases (`tag_nodeId`)

"installation" and "dev" directories use the same files

piotrp (changes requested):

docs/Changelog.md
philemone: Updated, also updated some other docs and few usages of nodeId's I've missed earlier so 3 commits in total

piotrp (approved):

PR #9086, opened 2026-03-18
[NU-2477] don't add empty value to mock on testing tab open

Dzuming:
## Checklist before merge

### DIFF
--- a/designer/client/src/components/graph/node-modal/editors/expression/MockExpressionField.tsx
+++ b/designer/client/src/components/graph/node-modal/editors/expression/MockExpressionField.tsx
@@ -55,8 +55,6 @@ function MockExpressionField(props: Props): React.JSX.Element {
 
     const editMock: OnValueChange = useCallback(
         (expression) => {
-            if (expression.expression === "") return;
-
             dispatch(setTestCaseMock(editedNode.id, expression));
         },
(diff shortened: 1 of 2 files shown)


designer/client/src/components/graph/node-modal/editors/expression/MockExpressionField.tsx
```diff
@@ -55,6 +55,8 @@ function MockExpressionField(props: Props): React.JSX.Element {
 
     const editMock: OnValueChange = useCallback(
         (expression) => {
+            if (expression.expression === "") return;
```
philemone: What happens if someone enters a value in the mock field, saves it, and then clears it? The early return will fire here, so dispatch won’t run 🤔  doesn’t that break clearing the mock after it has been set?

Dzuming: Right i changed an approach and now to also handle this case, so when there is an empty (or not defined) expression, it will be removed from the redux state (or not added at)

philemone: Thanks, looks good :)

philemone (approved):

PR #9091, opened 2026-03-19
Refresh test capabilities after node changes in testing tab

jedrz:
## Describe your changes

* New endpoints to fetch and validate a single source instead of a full scenario.
* Fixes fetching capabilities for a stale scenario with an old version of the node.
* Fixes fetching capabilities for  RR source by removing artificial parameters with input/output schema before sending a request.
* The table editor for input records is aware of only a single source, as we do not allow editing records from multiple sources simultaneously. 
* Next PR: remove unused `scenarioTestValidationEndpoint`.

Before:

[before.webm](https://github.com/user-attachments/assets/d99dbe61-55ad-43f9-a820-1788df621a01)

After:

[after.webm](https://github.com/user-attachments/assets/03650732-03f7-4753-b740-6ca784e3e5ce)

Fixes bad expression type after changing node:

[fixes bad expression type for changed record.webm](https://github.com/user-attachments/assets/8bbb38c0-b886-4a6a-9d80-8cfc55dd3677)

## Checklist before merge


designer/client/src/components/graph/node-modal/node/NodeDetails.tsx
```diff
@@ -88,9 +89,18 @@ function NodeDetails(props: NodeDetailsProps): React.JSX.Element {
...
-    const buttonFromToolbar = useGetButtonFromToolbar(CustomButtonTypes.scenarioTest);
 
     const { node, editedNode, onChange, outputEdges, performNodeEdit, editState } = useNodeState(data.meta);
+
+    const _savedScenarioGraph = useAppSelector(getScenarioGraph);
+    const _scenarioGraphWithEditedNode = useMemo<ScenarioGraph>(() => {
+        const nodeInGraph = _savedScenarioGraph.nodes.find((n) => n.id === editedNode.id);
+        return nodeInGraph === editedNode
+            ? _savedScenarioGraph
+            : { ..._savedScenarioGraph, nodes: _savedScenarioGraph.nodes.map((n) => (n.id === editedNode.id ? editedNode : n)) };
```
jedrz: Maybe there is a better way? Node changes are not being propagated to the scenario graph.

Dzuming: If I understand correctly, the current problem is:

Someone edits the value field and adds/removes a new field
1. The user switches to the testing tab and adds a new row to the test data
2. The new field is not added as default data, because the value field is still in the 3. local node details state — it hasn't been propagated to the scenario graph state yet (that only happens when the user clicks "Apply"). Because of this, capabilities are also not refreshed.

To resolve this issue, instead of relying on the (potentially stale) Redux state, we should:

1. On "Add record" click, make a separate call to test/capabilities with the current local node state (including unsaved changes)
2. Use the response data to populate the new row's default values

We will be symmetrical with `Append from live data` button, where, we send a data from locally edited node in the request

jedrz: Finally, I switched to a node-level approach instead of a scenario-level one, which contained possibly stale data. Thanks to that, there is now a single request to fetch capabilities for a source to determine the default record.

designer/client/src/actions/nk/process.ts
```diff
@@ -70,7 +70,7 @@ export function loadProcessState(processName: ProcessName, processVersionId: num
 
 const getTestCapabilities = memoizeByArgsWithTTL((processName: ProcessName, scenarioGraph: ScenarioGraph) => {
     return HttpService.getTestCapabilities(processName, scenarioGraph);
-}, 1000);
+});
```
Dzuming: Do we need this change?

designer/client/src/components/graph/node-modal/node/NodeContent/TestingContentElements/InputDataRecords.tsx
```diff
@@ -1,30 +1,43 @@
...
     const [isExpanded, setIsExpanded] = useState(true);
     const maxTestingRecords = useAppSelector(getMaxTestingRecords);
-    const testingDataRecordsForSource = useAppSelector((state) => getInputDataRecordsForSingleSource(state, sourceId));
+    const testingDataRecordsForSource = useAppSelector((state) => getInputDataRecordsForSingleSource(state, node.id));
+    const scenarioProperties = useAppSelector(getProcessProperties);
+    const scenarioName = useAppSelector(getProcessName);
+
+    const [sourceParameters, setSourceParameters] = useState<TestFormParameters | undefined>(undefined);
+
+    useEffect(() => {
```
Dzuming: We could use usePromise from rooks here, to automatically handle async operation, with pending promise cleaning in case of state changes, out of the box

```tsx

const { data: sourceParameters } = usePromise(
    async () => {
        const { data } = await HttpService.getSourceTestCapabilities(scenarioName, scenarioProperties, node);
        const capabilities = data.testWithParameters;
        return capabilities.status === TestCapabilityStatus.AVAILABLE ? capabilities.sourceParameters : null;
    },
    [scenarioName, scenarioProperties, node],
);

```

designer/client/src/components/modals/TestingDataRecords/Table.tsx
```diff
@@ -131,42 +125,31 @@ export const Table: React.FC<TableProps> = ({
...
+        [theme, onValidateVariables],
     );
 
-    const defaultVariablesBySourceId = useMemo(() => buildDefaultVariablesMap(sourceParameters), [sourceParameters]);
-
-    const sourceNameById = useMemo(
-        () => sourceParameters.reduce<Record<string, string>>((acc, sp) => ({ ...acc, [sp.sourceId]: sp.sourceName }), {}),
-        [sourceParameters],
-    );
+    const defaultVariables = useMemo(() => buildDefaultVariables(sourceParameters?.parameters), [sourceParameters]);
```
Dzuming: Do we need it? it looks like a duplication of the logic from InputDataRecords

designer/client/src/components/modals/TestingDataRecords/utils.ts
```diff
@@ -36,27 +37,17 @@ export function safeParseExpression<T = unknown>(expr: string): T {
...
-            defaultsBySourceId[sourceParameter.sourceId] = JSON.stringify(values, null, 2);
-        } catch {
-            // If stringify fails for any reason, fall back to empty string
-            defaultsBySourceId[sourceParameter.sourceId] = "";
-        }
+export function buildDefaultVariables(parameters?: UIParameter[]): string {
+    if (!parameters) return "";
+    let values: unknown = {};
+    for (const param of parameters ?? []) {
+        values = safeParseExpression(param?.defaultValue?.expression ?? "");
```
Dzuming: It worked this way previously, but this implementation means, that in case of multiple parameters, values will be always override.  What is the case, where we will have multiple parameters here?

designer/client/src/components/modals/TestingDataRecords/utils.ts
```diff
@@ -36,27 +37,17 @@ export function safeParseExpression<T = unknown>(expr: string): T {
...
-
-        try {
-            defaultsBySourceId[sourceParameter.sourceId] = JSON.stringify(values, null, 2);
-        } catch {
-            // If stringify fails for any reason, fall back to empty string
-            defaultsBySourceId[sourceParameter.sourceId] = "";
-        }
+export function buildDefaultVariables(parameters?: UIParameter[]): string {
+    if (!parameters) return "";
+    let values: unknown = {};
```
Dzuming: In the current implementation this type should be 

```ts
let values: Record<string, unknown> | undefined = {};
```
because `safeParseExpression` can return an undefined

which means, that `buildDefaultVariables` can return string | undefined

maybe we should at least return something like

```ts
        return values ? JSON.stringify(values, null, 2) : '';

```
?

designer/client/src/components/modals/TestingDataRecords/Table.tsx
```diff
@@ -354,13 +335,13 @@ export const Table: React.FC<TableProps> = ({
...
                     onCellsEdited={onCellEdited}
-                    onRowAppended={recordsToAddLimitExceeded || noSourcesDefined ? undefined : onCellAdded}
+                    onRowAppended={recordsToAddLimitExceeded ? undefined : onCellAdded}
                     rowMarkers="both"
                     rows={data.length}
                     smoothScrollX
                     smoothScrollY
                     theme={tableTheme}
-                    width={TABLE_WIDTH}
+                    width="100%"
```
Dzuming: Why const is removed here? it's consistent with:

```ts
const TRAILING_ROW_HINT = "Add record";
const COLUMN_SOURCE_ID = "sourceId";
const COLUMN_VARIABLES_ID = "variables";
const COLUMN_SOURCE_TITLE = "Source";
const COLUMN_VARIABLES_TITLE = "Input variables";
const COLUMN_SOURCE_WIDTH = 150;
const COLUMN_VARIABLES_WIDTH = 300;
```

designer/client/src/http/HttpService/HttpService.ts
```diff
@@ -750,6 +751,17 @@ export class HttpService {
         return promise;
     }
 
+    getSourceTestCapabilities(processName: string, scenarioProperties: PropertiesType, nodeData: NodeType) {
+        const promise = api.post<SourceTestCapabilities>(`/scenarioTesting/${encodeURIComponent(processName)}/sourceCapabilities`, {
+            scenarioProperties,
+            nodeData,
+        });
+        promise.catch((error) =>
+            this.#addError(i18next.t("notification.error.failedToGetCapabilities", "Failed to get capabilities"), error, true),
```
Dzuming: "Failed to get source test capabilities"?

Dzuming (changes requested):

designer/client/src/actions/nk/process.ts
jedrz: Yes, I think it can be restored to the original version.

designer/client/src/components/modals/TestingDataRecords/utils.ts
jedrz: I reworked the function slightly and added a comment that the parameters should contain a single element.

Dzuming (approved):
LGTM

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/testing/WithAdHocTestsLogic.scala
```diff
@@ -95,10 +152,24 @@ trait WithAdHocTestsLogic {
...
+    exampleScenario.nodes
+      .collectFirst { case FlatNode(s: SourceNodeData) => s }
+      .getOrElse(fail("No source node found in exampleScenario"))
+
   protected def validParameters: TestSourceParameters
 
   protected def expectedTestParametersJson: String
 
+  protected def expectedTestSourceParametersJson: String =
+    io.circe.parser
```
philemone: We have it imported in the scope, if you want it can be just `parser`

philemone (approved):
Nice

PR #9095, opened 2026-03-19
Always close `SttpBackend` instances in OpenAPI services

piotrp:
## Describe your changes

Always close `SttpBackend` instances in OpenAPI services

- we could leak a backend in service discovery - deployment created a new backend with its dedicated thread pool, this change also uses backend configuration from service config instead of hardcoded one
- Lite/RR never closed their `AsyncHttpClient` - `FixedAsyncHttpClientBackendProvider.close` was not implemented
- `InfluxDbHttpReporter` didn't close its backend - it didn't have to, as `close` in `TryHttpURLConnectionBackend` is a noop, but let's use a correct lifecycle pattern anyway

## Checklist before merge

### DIFF
--- a/components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenAPIServiceSpec.scala
+++ b/components/openapi/src/it/scala/pl/touk/nussknacker/openapi/functional/OpenAPIServiceSpec.scala
@@ -41,7 +41,7 @@ class OpenAPIServiceSpec
   def withFixture(test: OneArgTest): Outcome = {
     val definition = ResourceLoader.load("/customer-swagger.json")
 
-    val client = new DefaultAsyncHttpClient()
+    val clientProvider = new FixedAsyncHttpClientBackendProvider(new DefaultAsyncHttpClient())
     try {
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -845,8 +845,8 @@ lazy val benchmarks = (project in file("benchmarks"))
     name                                 := "nussknacker-benchmarks",
     libraryDependencies ++= {
       Seq(
-        "pl.touk"         %% "flink-scala"                    % flinkScalaV exclude ("com.esotericsoftware", "kryo-shaded"),
```
lciolecki, line 848: I'm not sure these should be part of this change? :)

piotrp: Cleanup after previous MR, this commit will be merged in #9094

lciolecki (approved):

lciolecki:
It's not only in OpenApi? Maybe let's leave: Always close SttpBackend instance ?

piotrp:
It is only OpenAPI, some classes may look generic but are used only there. And `InfluxDbHttpSender` change is purely cosmetic, that `close` does nothing there.

PR #9155, opened 2026-03-27
Fix NPE in comparison operators with literal null

jedrz:
## Describe your changes

Before:

before

After:

after

## Checklist before merge

### DIFF
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
@@ -234,6 +234,10 @@ object SpelExpressionTypingError {
       override def message: String = s"Bad '$operator' operator construction"
     }
 
+    case class OperatorNullOperandError(operator: String) extends OperatorError {
+      override def message: String = s"Operator '$operator' cannot be used with a null operand"
+    }
+
     case class DivisionByZeroError private (expression: String) extends OperatorError {
       override def message: String = s"Division by zero: $expression"
     }
--- a/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelTyper.scala
+++ b/scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelTyper.scala
@@ -210,12 +210,14 @@ private[spel] class SpelTyper(
       val resultType       = Typed.fromDetailedType[R]
       withTypedChildren {
         case castExpectedType(left) :: castExpectedType(right) :: Nil =>
-          val typeFromOp = for {
-            leftValue  <- left.valueOpt
-            rightValue <- right.valueOpt
-            res = op(leftValue, rightValue)
-          } yield Typed.fromInstance(res)
-          typeFromOp.getOrElse(resultType).validTypingResult
+          (left.valueOpt, right.valueOpt) match {
+            case (Some(leftValue), Some(rightValue)) if leftValue == null || rightValue == null =>
+              OperatorNullOperandError(operatorName).invalidTypingResult(fallbackType = resultType)
+            case _ =>
+              operationOnTypesValue[A, A, R](left.typingResult, right.typingResult, resultType)((a, b) =>
+                Valid(op(a, b))
+              )
+          }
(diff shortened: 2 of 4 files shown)


scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
```diff
@@ -234,6 +234,10 @@ object SpelExpressionTypingError {
       override def message: String = s"Bad '$operator' operator construction"
     }
 
+    case class OperatorNullOperandError(operator: String) extends OperatorError {
+      override def message: String = s"Operator '$operator' cannot be used with a null operand"
```
lciolecki, line 238: Dot at the end? :)

jedrz: I think it looks better in the UI without the ending dot :)

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelTyperSpec.scala
```diff
@@ -212,6 +218,22 @@ class SpelTyperSpec extends AnyFunSuite with Matchers with ValidatedValuesDetail
       TypedObjectWithValue(Typed.typedClass[String], longString)
   }
 
+  test("comparison operators with null literal operand should produce a compilation error") {
```
piotrp, line 234: These are forbidden cases, but do we have a test verifying that `expression == null`/`expression != null` can pass?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelExpressionTypingError.scala
lciolecki: OK, let's leave it. At the moment, we're not presenting errors consistently.

lciolecki (approved):

scenario-compiler/src/test/scala/pl/touk/nussknacker/engine/spel/SpelTyperSpec.scala
jedrz: Good point, added a new test for equality operators with null.

piotrp (approved):

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/spel/SpelTyper.scala
```diff
@@ -210,12 +210,14 @@ private[spel] class SpelTyper(
...
         case castExpectedType(left) :: castExpectedType(right) :: Nil =>
-          val typeFromOp = for {
-            leftValue  <- left.valueOpt
-            rightValue <- right.valueOpt
-            res = op(leftValue, rightValue)
-          } yield Typed.fromInstance(res)
-          typeFromOp.getOrElse(resultType).validTypingResult
+          (left.valueOpt, right.valueOpt) match {
+            case (Some(leftValue), Some(rightValue)) if leftValue == null || rightValue == null =>
+              OperatorNullOperandError(operatorName).invalidTypingResult(fallbackType = resultType)
```
mateuszkp96, line 215: This case validates nulls only when two arguments are literals, but it still allows expressions like
`null <= 'abc'.length`

mateuszkp96: You can handle this case before casting:
```
      withTypedChildren {
        case TypedNull :: _ :: Nil | _ :: TypedNull :: Nil =>
          OperatorNullOperandError(operatorName).invalidTypingResult(fallbackType = resultType)
        case castExpectedType(left) :: castExpectedType(right) :: Nil =>
```

jedrz: Sorry, I clicked 'merge' before seeing your comments. Here is follow up: https://github.com/TouK/nussknacker/pull/9168.
Regarding `null <= 'abc.length'`, I think it works because the pattern matching uses an OR condition, which only requires one side to be a literal null.

PR #9161, opened 2026-03-28
feat: expression builder in Actual assertion field

qoomak:
## Summary
- Adds SpEL expression builder dialog to the **Actual** field in assertion rows, gated by `node.showFieldExpressionBuilder` user setting
- Builds context data from up to 20 incoming test records (`#records`) and passes it to the builder
- Adds `variableTypes` and `contextData` override props to `SpelExpressionBuilderComponent`; skips expensive `getAvailableContexts` call when `contextData` is already provided

## Test plan
- [ ] Enable `node.showFieldExpressionBuilder` in `__settings`
- [ ] Run a test on a scenario with incoming records
- [ ] Open a node with assertions — wand icon should appear next to Actual field
- [ ] Click wand icon — expression builder should open quickly even with many records
- [ ] Insert expression — it should be inserted into the Actual field
- [ ] Verify builder is absent when setting is disabled


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9202, opened 2026-04-01
[NU-2470] Validate test case name length

jedrz:
## Checklist before merge

### DIFF
--- a/designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ValidationResourcesSpec.scala
+++ b/designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ValidationResourcesSpec.scala
@@ -512,7 +512,7 @@ class ValidationResourcesSpec
       validation.errors.globalErrors should have size 3
       validation.errors.globalErrors.map(_.error.typ) should contain only (
         "DuplicateTestCaseNames",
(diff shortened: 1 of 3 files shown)


designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ValidationResourcesSpec.scala
```diff
@@ -493,6 +494,13 @@ class ValidationResourcesSpec
                 mocks = Map.empty,
                 assertions = Map.empty
               ),
+              TestCase(
+                id = UUID.randomUUID(),
+                name = "too long name " * 10,
```
philemone: Can we also here use `TestCaseValidator.MaxTestCaseNameLength` to make sure that it is too long? `* 10` is ok, but the case with MaxTestCaseNameLength is IMHO better

philemone (approved):

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/ValidationResourcesSpec.scala
jedrz: Thanks, fixed

PR #9232, opened 2026-04-07
improve: do not expand target field definition when opening DataMapper

qoomak:
## Summary
- When clicking the magic wand on a Kafka sink field, the DataMapper now scrolls to the target field without expanding its detail panel
- Previously, opening the DataMapper would expand (select) the focused field automatically

## Test plan
- [ ] Open a Kafka sink node with multiple fields
- [ ] Click the magic wand icon next to a field that is not visible in the initial scroll position
- [ ] DataMapper should open, scroll to the field, but NOT expand it
- [ ] Clicking the field header should still expand it as usual


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9237, opened 2026-04-07
improve: expand dotted param names and Record types into nested DataMapper fields

qoomak:
## Summary

When a Kafka sink has parameters with dotted names (e.g. `combinedState.key`, `combinedState.tool_record`), DataMapper now groups them into a nested hierarchy:

- Parameters sharing a common prefix (e.g. `combinedState.*`) are grouped under a single `combinedState` parent node
- Parameters with a Record type are expanded into individual sub-fields using `typ.fields`
- On insert, the nested tree is flattened back to individual param expressions via `flattenToParams`

Depends on #9232.

## Test plan
- [ ] Open DataMapper on a Kafka sink with dotted params
- [ ] Verify a `combinedState` parent node appears with `key`, `tool_record`, ... as children
- [ ] Verify `tool_record` is pre-expanded with its sub-fields (toolId, ts, wear)
- [ ] Click Insert — each param should receive its individual expression

PR #9238, opened 2026-04-07
improve: DataMapper schema mode — hide field controls for schema-driven fields

qoomak:
## Summary

When DataMapper is opened with `hideFieldControls=true` (Kafka sink with a topic schema):
- **Name** and **Type** fields are hidden — fixed by schema
- **Build Record** button hidden (also restricted to `Map`-typed fields in free mode)
- **Add field** button inside nested records hidden
- Record fields no longer open an empty expanded panel when clicked
- Expression chip shows full expression (up to `maxWidth: 300`) with tooltip instead of truncating at 30 chars

Depends on #9237.

## Test plan
- [ ] Open DataMapper on a Kafka sink with a topic schema
- [ ] Click a leaf field (e.g. `x: Double`) — only value + default value, no Name/Type/Build Record
- [ ] Click a record field (e.g. `axis_snapshot`) — no empty panel
- [ ] Expression chip shows full SpEL with hover tooltip
- [ ] In free mode, Name/Type/Build Record/Add field controls are still present

### DIFF
--- a/designer/client/src/components/dataMapper/FieldRow.tsx
+++ b/designer/client/src/components/dataMapper/FieldRow.tsx
@@ -217,18 +217,21 @@ export function FieldRow({
                     />
                 )}
                 {!field.isRecord && field.expression && (
-                    <Chip
-                        label={field.expression.length > 30 ? field.expression.slice(0, 30) + "…" : field.expression}
-                        size="small"
-                        sx={{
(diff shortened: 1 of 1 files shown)


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9239, opened 2026-04-07
fix: DataMapper automap handles default values and Record-typed primitive context variables

qoomak:
## Summary

Two related fixes to the DataMapper automap button:

**1. Overwrite trivial/default expressions**
`autoMap` previously skipped any field with a non-empty `expression`. Fields initialized from schema defaults (`''`, `0`, `null`, `false`) were silently skipped. Now automap overwrites trivial literals and only preserves real expressions.

**2. Traverse Record-typed variables even when context value is primitive/null**
`enrichedContext` built from test-record `pretty` values sometimes contains primitive values (string, number, null) for variables whose typing declares them as Records. The traversal was skipping these, so `pathMap` never got entries for their sub-fields. Now when the typing has `fields`, the variable is replaced by `typingResultToSample` to enable traversal.

Depends on #9238.

## Test plan
- [ ] Open DataMapper on a Kafka sink with schema params (e.g. `combinedState.key`, `modelOutput.value`)
- [ ] Click Automap — all fields including top-level ones (`key`, `value`) should be filled
- [ ] Fields that already have a real expression (e.g. `#combinedState?.key`) should NOT be overwritten
- [ ] Fields with default values (`''`, `0`, `null`) should be overwritten by automap

### DIFF
--- a/designer/client/src/components/dataMapper/useDataMapper.ts
+++ b/designer/client/src/components/dataMapper/useDataMapper.ts
@@ -171,8 +171,16 @@ export function useDataMapper({
         for (const [key, typingResult] of Object.entries(variableTypes)) {
             const existing = merged[key];
             const elem0 = Array.isArray(existing) ? existing[0] : undefined;
+            const typingHasFields =
+                typingResult !== null &&
+                typeof typingResult === "object" &&
+                "fields" in typingResult &&
+                typingResult.fields !== null &&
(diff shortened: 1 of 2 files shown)


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9241, opened 2026-04-07
fix: nested SpEL record when switching Kafka sink raw editor mode

qoomak:
## Summary
- Fixes switching between structured form and raw editor mode in Kafka sink node
- **Dynamic → raw editor**: groups dotted param names (`combinedState.key`, `combinedState.tool_record`) into proper nested SpEL structure instead of flat dotted keys — flat keys are invalid SpEL and caused type mismatch / validation errors
- **Raw editor → dynamic**: handles both old flat format and new nested `Value` expression when distributing back to individual params; normalizes indentation of embedded SpEL records

## Test plan
- [ ] Open a Kafka Avro sink with dotted params (e.g. `combinedState.key`, `combinedState.tool_record`)
- [ ] Toggle **Raw editor** ON — Value expression should be nested: `{ combinedState: { key: ..., tool_record: {...} }, modelOutput: { value: ... } }`
- [ ] Toggle **Raw editor** back OFF — individual params should be restored with correct expressions and clean 2-space indentation
- [ ] Verify no validation errors ("only string keys supported") on the Value expression in raw editor mode

### DIFF
--- a/designer/client/src/components/graph/node-modal/ParametersUtils.ts
+++ b/designer/client/src/components/graph/node-modal/ParametersUtils.ts
@@ -50,8 +50,14 @@ function splitTopLevelParts(inner: string): string[] {
     return parts.filter(Boolean);
 }
 
-/** Parse a top-level SpEL record `{ k1: expr1, k2: expr2 }` into a name→expression map. */
-function parseSpelRecord(expression: string): Record<string, string> | null {
+/**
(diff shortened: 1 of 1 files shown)


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9244, opened 2026-04-07
feat: expand Record-typed DnD into nested isRecord fields in DataMapper

qoomak:
## Summary
- When dragging a Record-typed variable onto the DataMapper drop zone, recursively builds a nested `isRecord` FieldDef tree with children instead of a flat leaf with `toString()` expression
- When dragging a Record onto an **existing field**, replaces it with the nested isRecord structure (preserving the field's ID and name)
- Adds `addChildFromDrop` callback: dragging onto the nested children area (or **Add field** zone of a record) adds a child field, also with recursive Record expansion
- Extracts `buildFieldFromTyping` as a shared module-level helper used by all three drop handlers

## Test plan
- [ ] In DataMapper (free-form mode), drag a Record-typed variable from the context tree onto the drop zone — should create a nested isRecord field with all children pre-filled
- [ ] Drag a Record-typed variable onto an existing empty leaf field — should convert it to an isRecord structure
- [ ] Drag a non-Record variable onto an existing field — should behave as before (set expression)
- [ ] Drag any variable onto the nested children area of a record field — should add a new child (recursively expanding if it's a Record)
- [ ] Verify nested isRecord correctly generates SpEL output

### DIFF
--- a/designer/client/src/components/dataMapper/DataMapper.tsx
+++ b/designer/client/src/components/dataMapper/DataMapper.tsx
@@ -323,6 +323,7 @@ export function DataMapper({
                                         onDragOverId={dm.setDragOverId}
                                         onChange={dm.updateField}
                                         onAddChild={dm.addChildField}
+                                        onAddChildFromDrop={dm.addChildFromDrop}
                                         onRemove={dm.removeField}
                                         onMove={dm.moveField}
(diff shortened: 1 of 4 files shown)


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9249, opened 2026-04-08
fix: expand dotted keys in fieldsFromSample for From Schema in DataMapper

qoomak:
## Summary
- Kafka topic schema responses use flat dotted notation (`"combinedState.key": null`) rather than nested objects
- `fieldsFromSample` now pre-processes input with `expandDottedKeys`, converting flat dotted keys into a nested structure before building `FieldDef`s
- **From Schema** now produces nested `isRecord` fields (e.g. `combinedState` with children `key`, `tool_record`, …) instead of flat leaf fields with dotted names
- Mixed schemas (some dotted leaf keys + some nested object siblings under the same prefix) are correctly merged under the same parent record

## Test plan
- [ ] Open DataMapper on a Kafka sink without a schema (free-form mode)
- [ ] Click **From Schema**, select a topic that has dotted field names in its schema
- [ ] Fields should appear as nested isRecord tree, not flat dotted names
- [ ] Verify SpEL output is properly nested (`{combinedState: {key: ..., tool_record: {...}}}`)

### DIFF
--- a/designer/client/src/components/dataMapper/dataMapperUtils.test.ts
+++ b/designer/client/src/components/dataMapper/dataMapperUtils.test.ts
@@ -337,4 +337,32 @@ describe("fieldsFromSample", () => {
         expect(result[0].children[0]).toMatchObject({ name: "city", type: "String" });
         expect(result[0].children[1]).toMatchObject({ name: "zip", type: "String" });
     });
+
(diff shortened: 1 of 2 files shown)


qoomak:
Zastąpiony przez #9261

qoomak:
Superseded by #9261

PR #9257, opened 2026-04-08
fix: guard against null testCapabilities in InputDataRecords

qoomak:
## Summary

- `testCapabilities` is initialized as `null` in the graph reducer and populated asynchronously after capabilities are fetched
- Accessing `.testWithParameters` before capabilities load causes a crash when the node modal opens during the loading window
- Added optional chaining (`?.`) on `testCapabilities` and `testCapabilitiesParameters` accesses

## Test plan

- [ ] Open a node modal immediately after page load (before capabilities are fetched) — should no longer crash
- [ ] Verify test data records tab works normally when capabilities are available


Dzuming (approved):
LGTM

PR #9258, opened 2026-04-08
fix: remove assertions section from source nodes in testing tab

qoomak:
## Summary

- Source nodes have their own test data records view in the Testing tab
- Assertions section was incorrectly shown for Source nodes alongside the test data records
- Removed the `<Assertions>` component from the Source node render path

## Test plan

- [ ] Open a scenario with a Source node, go to Testing tab — verify only test data records are shown (no Assertions section)
- [ ] Open a non-Source node (e.g. Enricher, Filter) — verify Assertions section still appears normally


Dzuming (approved):
LGTM

PR #9296, opened 2026-04-14
Docs - instance (environment) variables

ZbyszekMM:
just docs


docs/scenariosAuthoring/introduction.mdx
```diff
@@ -136,6 +136,10 @@ The `#meta` variable carries meta information about the scenario under execution
 The `#inputMeta` predefined variable is specific to streaming processing mode only; check [here](/docs/scenariosAuthoring/streamingProcessingMode.md#predefined-variables) for more information. 
 
 
+### Environment variables
```
ZbyszekMM: @mslabek Just wondering - would not 'instance vars' be a better name than 'environemnt variables'? What constitutes an environment here? Are not these variables instance-wide?

mslabek (approved):

PR #9309, opened 2026-04-16
Accept _unsafe_-prefixed forward-compat fields on scenario graph DTOs

JulianWielga:
External tooling that wants to attach extension data to a scenario graph today has no safe place to put it: any field Circe doesn't recognize on `UserDefinedAdditionalNodeFields` (node-level `additionalFields`) or `ProcessAdditionalFields` (scenario-level `additionalFields`) is dropped on decode, so the data is gone the next time the user saves. An earlier attempt parked a nested `_unsafeExtra: Map` inside `LayoutData`, but that mixed unrelated concerns into a layout type and put extensions in a hard-to-discover spot.

This PR introduces a prefix-based escape hatch on both `additionalFields` DTOs: any top-level JSON property whose name starts with `_unsafe_` is preserved verbatim through decode → encode. The prefix is the contract — it says "I know this field is not schema-validated, I accept responsibility for it" — which keeps the escape
  hatch cheap without polluting the real schema. A request body like `{ "description": "x", "_unsafe_foo": "1", "_unsafe_bar": 2 }` round-trips unchanged.

The mechanics: a hand-written Circe codec on each case class spreads/collects the prefixed fields via a `Map[String, Json]` container. The prefix and the extract/merge helpers live in one place (`CirceUtil.UnsafePrefixedFields`) so neither the convention nor the magic string is duplicated. A tiny `omitSchemaField` helper strips the synthetic container from the Tapir-derived schema, so the generated
  OpenAPI stays byte-for-byte identical to before and API consumers see no cosmetic churn. TS types mirror the shape via a shared `UnsafeOptions = Record<\`_unsafe_${string}\`, unknown>` intersected onto both DTOs.

Also bundled: a workaround for sbt-git not supporting git worktrees. In a worktree `.git` is a file, not a directory, and sbt-git's embedded JGit chokes on that layout when reading `gitCurrentBranch` / `gitHeadCommit`. Detecting `.git` as a file and flipping `useConsoleForROGit` to `true` delegates those reads to the git CLI, which handles worktrees correctly; normal checkouts keep the JGit default.

## Test plan

- [ ] Circe round-trip: POST a scenario with `_unsafe_*` entries at both node and scenario `additionalFields`, GET it back, confirm they are unchanged.
- [ ] Unknown non-prefixed fields are still dropped (no accidental passthrough).
- [ ] `sbt generateDesignerOpenApi` produces no diff on `docs-internal/api/nu-designer-openapi.yaml`.
- [ ] `NuDesignerApiAvailableToExposeYamlSpec` passes (examples still match schema).
- [ ] Frontend typecheck green — `additionalFields._unsafe_xxx` is typed as `unknown` everywhere it is consumed.
- [ ] `sbt compile` from a git worktree no longer logs JGit `"Not a directory"` errors.


build.sbt
```diff
@@ -54,6 +55,15 @@ crossScalaVersions := Nil
 
 ThisBuild / isSnapshot := version(_ contains "-SNAPSHOT").value
 
+// Workaround for sbt-git not supporting git worktrees (see https://github.com/sbt/sbt-git/issues/#worktrees).
```
piotrp: Does this link enable some filtering in Safari? The `#worktreees` part does nothing for me.

With a working link - do we need this explanation below?

mslabek (approved):

PR #9314, opened 2026-04-17
Rename handler error -> Error strategy, move to advanced parameters

philemone:
## Checklist before merge

### DIFF
--- a/components/openapi/src/main/scala/pl/touk/nussknacker/openapi/enrichers/OpenAPIEnricherFactory.scala
+++ b/components/openapi/src/main/scala/pl/touk/nussknacker/openapi/enrichers/OpenAPIEnricherFactory.scala
@@ -16,17 +16,23 @@ import pl.touk.nussknacker.engine.api.definition.{
   FixedValuesParameterEditor,
   NodeDependency,
   OutputVariableNameDependency,
-  Parameter
+  Parameter,
+  ParameterCategory
 }
 import pl.touk.nussknacker.engine.api.parameter.ParameterName
 import pl.touk.nussknacker.engine.api.runtimecontext.EngineRuntimeContext
 import pl.touk.nussknacker.engine.api.typed.typing.Typed
+import pl.touk.nussknacker.engine.graph.expression.Expression
 import pl.touk.nussknacker.engine.util.service.{EagerServiceWithErrorSupport, TimeMeasuringService}
-import pl.touk.nussknacker.engine.util.service.EagerServiceWithErrorSupport.{handleErrorsParam, isHandleErrorsEnabled}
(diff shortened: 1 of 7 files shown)


components/openapi/src/main/scala/pl/touk/nussknacker/openapi/enrichers/OpenAPIEnricherFactory.scala
```diff
@@ -180,6 +186,20 @@ object OpenAPIEnricherFactory {
     fixedValuesWithNullParameter(ServiceParamName, values)
   }
 
+  private val errorStrategyParam: Parameter = Parameter[Boolean](HandleErrorsParamName).copy(
+    defaultValue = Some(Expression.spel("false")),
+    editors = List(
+      FixedValuesParameterEditor(
+        List(
+          FixedExpressionValue("false", "Fail on error"),
```
jedrz, line 194: Wouldn't it be better to allow the same three options as in HTTP enricher: 'Accept Only 2xx', 'Accept 2xx and 404' and 'Accept All'? For 'Accept All', `ServiceResponseWithError` should be returned.

philemone: I knew this would come up 🙂

They look similar, but they are not equal. In HTTP Enricher, Error Strategy is status-code based (2xx / 2xx+404 / all). In OpenAPI, some failures (especially OAuth/token-flow errors from polyvariant-sttp) happen before we get a business response, and those errors may not have an HTTP status code.

So a 1:1 status-based strategy from HTTP Enricher would not cover all OpenAPI/OAuth cases cleanly.
That’s why our OpenAPI strategy focuses on handling mode (fail vs return wrapped error), and we expose statusCode only when it is available.

jedrz (approved):

components/openapi/src/main/scala/pl/touk/nussknacker/openapi/enrichers/OpenAPIEnricherFactory.scala
```diff
@@ -180,6 +186,20 @@ object OpenAPIEnricherFactory {
     fixedValuesWithNullParameter(ServiceParamName, values)
   }
 
+  private val errorStrategyParam: Parameter = Parameter[Boolean](HandleErrorsParamName).copy(
```
lciolecki, line 189: @philemone HandleErrorsParamName -> ErrorStrategyParamName? :)

philemone: I merged it 4 hours ago :sweat_smile:  but I will keep that in mind for next MR, thanks :)

PR #9353, opened 2026-05-12
Improvement: introduce MapState in aggregations and joins

lciolecki:
## Motivation

The existing sliding window aggregators store all time buckets in a single ValueState[SortedMap[Long, Aggregate]]. This means every read/write operation serializes and deserializes the entire map, which becomes increasingly expensive as the number of buckets grows.

The new MapState-based implementation stores each time bucket as an independent entry in Flink's MapState[Long, Aggregate]. With RocksDB state backend, this allows per-bucket access without deserializing the full state, significantly reducing serialization overhead for large windows.

A separate ValueState[List[Long]] maintains the list of bucket timestamps, avoiding expensive MapState key iteration (prefix scan) on RocksDB.


engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunctionWithMapState.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  private def handleElementLeftSlide(
+      timestamp: Long,
+      ctx: FlinkOnTimerCtx,
+      out: Collector[ValueWithContext[AnyRef]]
+  ): Unit = {
+    // Check if entries exist in [maxBucketTs - windowLength + 1, timestamp - windowLength].
+    // This range represents entries that were visible when the last event was emitted
+    // but are now outside the window boundary.
+    val maxBucketTsOpt = bucketState.keys().asScala.map(_.longValue()).maxOption
```
piotrp: You are reading keys from disk twice (three times if you include read from `computeFinalValue`).

`WithMapState` is supposed to optimize for state with a lot of value buckets, so maybe we should read keys once? They can be also passed to `computeFinalValue`, so that it won't need to read them for the third time.

You can prefilter state entries here and skip state read in `computeFinalValue` entirely - reading entire entries from `RocksDBMapState` has the same cost as reading only keys - the only optimization is that keys and values are lazily deserialized.

Maybe we should store keys in a separate ValueState? That would allow us to actually get rid of entire state reads on each item add/removal.

engine/flink/components-utils/src/main/scala/pl/touk/nussknacker/engine/flink/api/state/EvictableState.scala
```diff
@@ -120,15 +120,13 @@ trait LatelyEvictableStateFunctionMixin[StateType] extends RichFunction with Sta
...
 
   override protected def doMoveEvictionTime(time: Long, timeService: TimerService): Unit = {
-    val latestEvictionTimeValue = latestEvictionTimeForKey.value()
-    val maxEvictionTime = if (latestEvictionTimeValue == null || time > latestEvictionTimeValue) {
-      time
-    } else {
-      latestEvictionTimeValue.longValue()
-    }
-    if (latestEvictionTimeValue == null) {
+    val latestEvictionTime = latestEvictionTimeForKey.value()
```
piotrp: Let's undo this change

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Manages bucket data for MapState-based sliding window aggregators.
+ *
+ * Maintains a separate ValueState with bucket timestamps (keys) to avoid
+ * expensive prefix scans over RocksDB MapState for key-only operations
+ * like range checks, max computation, and eviction decisions.
+ */
+private class BucketsState(
+    private val mapState: MapState[java.lang.Long, AnyRef],
+    private val keysState: ValueState[java.util.List[java.lang.Long]]
```
DeamonDev: why not `ListState` here?

lciolecki: For 1 day there will be approximately 1440 elements. We probably will not have time window bigger  then 10 / 30days? So the number of elements isn't demanding. ValueState is enough IMO. Ofc we can discuss about it with @piotrp :)

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
```diff
@@ -0,0 +1,225 @@
...
+    timeHistogram.update(System.nanoTime() - start)
+    out.collect(ValueWithContext(finalVal, KeyEnricher.enrichWithKey(value.context, value.value)))
+  }
+
+  protected def computeFinalValue(timestamp: Long): AnyRef = {
+    val rangeStart = timestamp - timeWindowLengthMillis + 1
+    var count      = 0
+
+    val foldedState = bucketsState.keys.asScala
+      .map(_.longValue())
```
piotrp: Do we need this `.map` and `.toList`?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
```diff
@@ -0,0 +1,225 @@
...
+      .foldLeft(aggregator.createAccumulator()) { (acc, key) =>
+        count += 1
+        aggregator.merge(acc, bucketsState.get(key))
+      }
+
+    retrievedBucketsHistogram.update(count)
+    aggregator.alignToExpectedType(aggregator.getResult(foldedState), outputType)
+  }
+
+  protected def doMoveEvictionTime(time: Long, timeService: TimerService): Unit = {
```
piotrp: Why did you remove comment abount intentional code duplication?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunctionWithMapState.scala
```diff
@@ -0,0 +1,103 @@
...
+    allKeys.maxOption.foreach { maxBucketTs =>
+      val leavingRangeStart = maxBucketTs - timeWindowLengthMillis + 1
+      val leavingRangeEnd   = timestamp - timeWindowLengthMillis
+
+      val hasLeavingEntries =
+        leavingRangeEnd >= leavingRangeStart &&
+          allKeys.exists(k => k >= leavingRangeStart && k <= leavingRangeEnd)
+
+      if (hasLeavingEntries) {
+        val foldRangeStart = timestamp - timeWindowLengthMillis + 1
```
piotrp: This is exactly the same code as in `computeFinalValue`, let's change it or add an overloaded `def computeFinalValue(timestamp: Long, bucketKeys: util.List[Long])` and use it in both places.

piotrp (changes requested):

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
piotrp: We read this list or replace it entirely, so the end result would be the same. So it doesn't matter which type we use.

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
lciolecki: Reverted

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunctionWithMapState.scala
lciolecki: Good catch

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunctionWithMapState.scala
```diff
@@ -178,48 +184,50 @@ trait AggregatorFunctionWithMapStateMixin extends AggregatorFunctionBase {
...
+   * Maintains a separate ValueState with bucket timestamps (keys) to avoid
+   * expensive prefix scans over RocksDB MapState for key-only operations
+   * like range checks, max computation, and eviction decisions.
+   */
+  class BucketsState(
+      private val mapState: MapState[java.lang.Long, AnyRef],
+      private val keysState: ValueState[java.util.List[java.lang.Long]]
+  ) {
+
+    def keys: java.util.List[java.lang.Long] = keysState.value() match {
```
piotrp: Let's add a comment saying that result is always sorted?

piotrp (approved):

engine/flink/components/base-tests/src/test/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/TransformersTest.scala
```diff
@@ -62,10 +62,11 @@ class TransformersTest extends AnyFunSuite with FlinkSpec with Matchers with Ins
...
       ConfigFactory.empty(),
-      ComponentDefinition("start", sourceComponent) :: FlinkBaseUnboundedComponentProvider.create(
-        DocsConfig.Default,
-        aggregateWindowsConfig
-      ) ::: FlinkBaseComponentProvider.Components
+      ComponentDefinition("start", sourceComponent)
+        :: FlinkBaseUnboundedComponentProvider.create(
+          DocsConfig.Default,
+          aggregateWindowsConfig
+        ) ::: FlinkBaseComponentProvider.Components
```
piotrp: Do we still need changes in this class?
* this is a formatting-only change
* extracted aggregate test methods are always used with "aggregate-sliding"

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -141,71 +155,131 @@ trait AggregatorFunctionMixin[MapT[K, V]] extends RichFunction { self: StateHold
...
+   * Maintains a separate ValueState with bucket timestamps (keys) to avoid
+   * expensive prefix scans over RocksDB MapState for key-only operations
+   * like range checks, max computation, and eviction decisions.
+   */
+  class BucketsState(
+      private val mapState: MapState[java.lang.Long, AnyRef],
+      private val keysState: ValueState[java.util.List[java.lang.Long]]
+  ) {
+
+    // Keys are always ordered. We order them before saving them to the state
```
piotrp: This should be a Scaladoc comment, not a line comment

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -141,71 +155,131 @@ trait AggregatorFunctionMixin[MapT[K, V]] extends RichFunction { self: StateHold
...
+    def get(key: java.lang.Long): AnyRef =
+      mapState.get(key)
+
+    def put(key: java.lang.Long, value: AnyRef): Unit = {
+      mapState.put(key, value)
+      val currentKeys = keys
+      if (!currentKeys.contains(key)) {
+        currentKeys.add(key)
+        currentKeys.sort(Comparator.naturalOrder())
+        // We keep sorted list of keys
```
piotrp: This should be before the line where we sort this list?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitExtraWindowWhenNoDataTumblingAggregatorFunction.scala
```diff
@@ -74,39 +68,25 @@ class EmitExtraWindowWhenNoDataTumblingAggregatorFunction[MapT[K, V]](
...
       )
     )
 
-    val previousTimestampStateAndRest = stateForTimestampToReadUntilEnd(currentStateValue, previousTimestamp)
-    if (previousTimestampStateAndRest.toScalaMapRO.isEmpty) {
-      evictStates()
-    } else {
+    val rangeStart  = previousTimestamp - timeWindowLengthMillis + 1
+    val hasMoreData = bucketsState.keys.stream().anyMatch(k => k >= rangeStart)
+    if (hasMoreData) {
```
piotrp, line 87: `addElementToState` no longer trims state, and here you handle only entire state clearing - so you just introduced an unbounded state growth ;)

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunction.scala
```diff
@@ -67,33 +57,34 @@ class EmitWhenEventLeftAggregatorFunction[MapT[K, V]](
...
-    val currentStateValue = readStateOrInitial()
-    handleElementLeftSlide(currentStateValue, timestamp, ctx, out)
-    super.onTimer(timestamp, ctx, out)
+    handleElementLeftSlide(timestamp, ctx, out)
+    handleOnTimer(timestamp, ctx.timerService())
   }
 
-  protected def handleElementLeftSlide(
-      currentStateValue: MapT[Long, aggregator.Aggregate],
+  private def handleElementLeftSlide(
```
piotrp: `handleElementsLeavingWindow`?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunction.scala
```diff
@@ -67,33 +57,34 @@ class EmitWhenEventLeftAggregatorFunction[MapT[K, V]](
...
-      val finalVal = computeFinalValue(currentStateValue, timestamp)
-      out.collect(
-        ValueWithContext(
-          finalVal,
-          KeyEnricher.enrichWithKey(NkContext(contextIdGenerator.nextContextId()), ctx.getCurrentKey)
+    val allKeys = bucketsState.keys
+
+    if (!allKeys.isEmpty) {
+      val maxBucketTs       = allKeys.last
+      val leavingRangeStart = timestampToReadUntilEnd(maxBucketTs)
```
piotrp: Can we replace all `timestampToReadUntilEnd` occurrences with `previousWindowEnd(...) + 1` (which will allow us to remove `timestampToReadUntilEnd`)?

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/EmitWhenEventLeftAggregatorFunction.scala
lciolecki: I will not argue with this ;)

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -141,71 +157,170 @@ trait AggregatorFunctionMixin[MapT[K, V]] extends RichFunction { self: StateHold
...
 
   private def computeTimestampToStore(timestamp: Long): Long = {
     (timestamp / minimalResolutionMs) * minimalResolutionMs
   }
 
-  protected def stateForTimestampToSave[T](stateValue: MapT[Long, T], timestamp: Long): MapT[Long, T] = {
-    stateValue.from(
-      timestamp - timeWindowLengthMillis + 1 - allowedOutOfOrderMs
-    ) // we must have exclusive range like Flink's sliding/tumbling have
+  // +1 gives an exclusive lower bound, matching Flink's half-open window range semantics [start, end):
```
piotrp: * where dou you have that "+1"? (you will have if we remove `timestampToReadUntilEnd`)
* this should be placed before code that does this, not as entire function comment

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -141,71 +157,170 @@ trait AggregatorFunctionMixin[MapT[K, V]] extends RichFunction { self: StateHold
...
-    stateForTimestampToReadUntilEnd(stateValue, timestamp).toRO(
-      timestamp
-    ) // we must have exclusive range like Flink's sliding/tumbling have
+  // +1 gives an exclusive lower bound, matching Flink's half-open window range semantics [start, end):
+  // the bucket at (timestamp - timeWindowLengthMillis) falls in the preceding window and must not be
+  // included in the current one.
+  protected def timestampToReadUntilEnd(timestamp: Long): Long =
+    timestamp - timeWindowLengthMillis + 1
+
+  // Returns the end of the window preceding the current timer, i.e. timestamp shifted back by one full
```
piotrp: This should be a Scaladoc comment

engine/flink/components/base-unbounded/src/main/scala/pl/touk/nussknacker/engine/flink/util/transformer/aggregate/AggregatorFunction.scala
```diff
@@ -141,71 +157,170 @@ trait AggregatorFunctionMixin[MapT[K, V]] extends RichFunction { self: StateHold
...
-  override protected def stateDescriptor: ValueStateDescriptor[MapT[Long, AnyRef]] =
-    new ValueStateDescriptor[MapT[Long, AnyRef]](
-      "state",
-      rangeMap.typeInformation[Long, AnyRef](TypeInformation.of(classOf[Long]), aggregateTypeInformation)
-    )
+    def keysInRange(from: lang.Long, to: lang.Long): View[lang.Long] = {
+      val buf     = list.asScala
+      val fromIdx = buf.search(from)
+      val toIdx   = buf.search(to, fromIdx.insertionPoint, buf.length)
+      val endIdx = toIdx match {
```
piotrp: `untilIdx`? It's an exclusive upper bound, i.e. index after the last element that we need

piotrp (approved):

PR #9364, opened 2026-05-19
Runtime parameter validation (ESP)

mgoworko:
## Describe your changes
* Added possibility of runtime parameter validation.
  * New `RuntimeParameterValidator` trait for validators that execute during scenario runtime.
  * Custom validators can now implement `CustomRuntimeParameterValidator` to validate a parameter value at runtime.
  * A failed runtime validation throws `ParameterValidationAtRuntimeException`.
  * Existing built-in validators (`@NotBlank`, `@NotNull`, etc.) remain compile-time only — their behaviour is unchanged.

PR on ESP branch, will cherry pick to staging when accepted

## Checklist before merge

### DIFF
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -477,16 +477,17 @@ object CustomParameterValidatorByClassLoader {
   def apply(clazz: Class[_ <: CustomParameterValidator]): CustomParameterValidatorByClassLoader =
     new CustomParameterValidatorByClassLoader(clazz.getName)
 
-  private val cache: TrieMap[(String, ClassLoader), CustomParameterValidator] = TrieMap.empty
+  private val cache: TrieMap[String, CustomParameterValidator] = TrieMap.empty
 
-  private def getOrLoad(className: String): CustomParameterValidator = {
-    val classLoader = Thread.currentThread().getContextClassLoader
-    cache.getOrElseUpdate((className, classLoader), load(className, classLoader))
-  }
+  private def getOrLoad(className: String): CustomParameterValidator =
+    cache.getOrElseUpdate(className, load(className))
 
-  private def load(className: String, classLoader: ClassLoader): CustomParameterValidator =
+  private def load(className: String): CustomParameterValidator =
     try {
-      Class.forName(className, true, classLoader).getDeclaredConstructor().newInstance() match {
+      Class
+        .forName(className, true, Thread.currentThread().getContextClassLoader)
+        .getDeclaredConstructor()
+        .newInstance() match {
         case v: CustomParameterValidator => v
         case _ => throw new RuntimeException(s"Class $className does not extend CustomParameterValidator")
       }
--- a/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
+++ b/components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
@@ -2,12 +2,13 @@ package pl.touk.nussknacker.engine.api.definition
 
 import cats.data.Validated
 import cats.data.Validated.{invalid, valid}
-import io.circe.generic.extras.ConfiguredJsonCodec
+import io.circe._
 import io.circe.parser._
-import pl.touk.nussknacker.engine.api.CirceUtil._
+import io.circe.syntax.EncoderOps
 import pl.touk.nussknacker.engine.api.NodeId
 import pl.touk.nussknacker.engine.api.context.PartSubGraphCompilationError
 import pl.touk.nussknacker.engine.api.context.ProcessCompilationError._
+import pl.touk.nussknacker.engine.api.definition.CustomParameterValidatorLoader.WithUnderlyingCustomParameterValidator
 import pl.touk.nussknacker.engine.api.parameter.{ParameterName, ParameterValueCompileTimeValidation}
 import pl.touk.nussknacker.engine.graph.expression.Expression
 import pl.touk.nussknacker.engine.graph.expression.Expression.Language
@@ -17,11 +18,23 @@ import java.util.regex.Pattern
 import scala.collection.concurrent.TrieMap
 import scala.util.Try
 
-trait Validator {
+sealed trait Validator
+
+trait CompileTimeValidator extends Validator {
 
   def isValid(paramName: ParameterName, expression: Expression, value: Option[Any], label: Option[String])(
       implicit nodeId: NodeId
   ): Validated[PartSubGraphCompilationError, Unit]
+
+}
+
+final case class ParameterRuntimeValidationError(input: String, message: String)
+
+trait RuntimeValidator extends Validator {
+
+  def isValid(paramName: ParameterName, expression: Expression, value: Any)(
+      implicit nodeId: NodeId
+  ): Validated[ParameterRuntimeValidationError, Unit]
 
 }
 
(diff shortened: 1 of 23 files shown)


components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
```diff
@@ -322,3 +466,33 @@ object CustomParameterValidatorDelegate {
...
+object CustomParameterValidatorByClassLoader {
+
+  def apply(clazz: Class[_ <: CustomParameterValidator]): CustomParameterValidatorByClassLoader =
+    new CustomParameterValidatorByClassLoader(clazz.getName)
+
+  private val cache: TrieMap[(String, ClassLoader), CustomParameterValidator] = TrieMap.empty
+
+  private def getOrLoad(className: String): CustomParameterValidator = {
+    val classLoader = Thread.currentThread().getContextClassLoader
+    cache.getOrElseUpdate((className, classLoader), load(className, classLoader))
```
piotrp: Do we need to segregate cache by ClassLoader? If we really do - please add an explanation

docs/MigrationGuide.md
```diff
@@ -245,6 +254,14 @@ To see the biggest differences please consult the [changelog](Changelog.md).
...
 * [#8209](https://github.com/TouK/nussknacker/pull/8209) Nussknacker now requires flink to be run with replaced `lib/flink-scala_2.12-x.x.x.jar` by `pl.touk:flink-scala` lib for the same scala version as used Nussknacker distribution. We provide prebuild flink docker images on [Docker Hub](https://hub.docker.com/r/touk/flink)    
 * [#8478](https://github.com/TouK/nussknacker/pull/8478) The behavior of `enum` to json encoding has been changed - now it uses `.name()` instead of `.toString()` 
+* Runtime parameter validation — custom validators and `Validator` API changes:
+  * `Validator.isValid` return type changed from `Validated[PartSubGraphCompilationError, Unit]` to `Validated[ParameterValidationError, Unit]`.
+    Custom validators implementing `CustomParameterValidator` must update their `isValid` signature accordingly.
+  * Validators annotated on parameters are now always evaluated at runtime (not only at compile time).
+    The previously opt-in `globalParameters.enableRuntimeParameterValidation` configuration key has been removed.
```
piotrp: Where did you find this `enableRuntimeParameterValidation`?

docs/MigrationGuide.md
```diff
@@ -245,6 +254,14 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 * [#7137](https://github.com/TouK/nussknacker/pull/7137)[#8317](https://github.com/TouK/nussknacker/pull/8317) Updated Flink 1.19.2 -> 1.20.2.
 * [#8209](https://github.com/TouK/nussknacker/pull/8209) Nussknacker now requires flink to be run with replaced `lib/flink-scala_2.12-x.x.x.jar` by `pl.touk:flink-scala` lib for the same scala version as used Nussknacker distribution. We provide prebuild flink docker images on [Docker Hub](https://hub.docker.com/r/touk/flink)    
 * [#8478](https://github.com/TouK/nussknacker/pull/8478) The behavior of `enum` to json encoding has been changed - now it uses `.name()` instead of `.toString()` 
+* Runtime parameter validation — custom validators and `Validator` API changes:
+  * `Validator.isValid` return type changed from `Validated[PartSubGraphCompilationError, Unit]` to `Validated[ParameterValidationError, Unit]`.
```
piotrp: This concerns only the new `RuntimeTimeValidator`s, old `CompileTimeValidator`s still returns `ParameterValidationError`. You removed `isValid` from the `Validator` trait

docs/MigrationGuide.md
```diff
@@ -245,6 +254,14 @@ To see the biggest differences please consult the [changelog](Changelog.md).
...
+    Custom validators implementing `CustomParameterValidator` must update their `isValid` signature accordingly.
+  * Validators annotated on parameters are now always evaluated at runtime (not only at compile time).
+    The previously opt-in `globalParameters.enableRuntimeParameterValidation` configuration key has been removed.
+  * New `@CustomValidator(classOf[MyValidator])` annotation in `pl.touk.nussknacker.engine.api.validation` allows
+    attaching a `CustomParameterValidator` implementation (identified by class) directly to a parameter.
+    The corresponding sealed subtype `CustomParameterValidatorByClass` was added to `ParameterValidator`.
```
piotrp: There is no `CustomParameterValidatorByClass` class in this PR

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionEvaluator.scala
```diff
@@ -64,18 +71,38 @@ class ExpressionEvaluator(
...
-          Option(evaluatedValue)
-        else if (param.shouldBeWrappedWithJavaOptional)
-          Optional.ofNullable(evaluatedValue)
-        else
-          evaluatedValue
+    val valueWithModifiedContext =
+      try {
+        evaluate[AnyRef](param.expression, param.name.value, nodeId.id, ctx)
+      } catch {
+        case NonFatal(ex) => throw CustomNodeValidationException(ex.getMessage, Some(param.name), ex)
```
piotrp, line 78: What if the expression evaluation itself throws e.g. a NPE or some other custom exception?

The same goes for exceptions from listeners, all of them suddenly become validation exceptions.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/exception/ParameterValidationAtRuntimeException.scala
```diff
@@ -0,0 +1,4 @@
+package pl.touk.nussknacker.engine.api.exception
+
+class ParameterValidationAtRuntimeException(input: String, message: String, cause: Throwable = null)
```
piotrp: You also have `ParameterRuntimeValidationError`, so maybe `ParameterRuntimeValidationException`?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionEvaluator.scala
```diff
@@ -64,18 +71,38 @@ class ExpressionEvaluator(
...
+
+  // Validation runs on the raw value before Option/Optional wrapping, so validators receive null
+  // for parameters where the expression evaluated to null (e.g. before wrapping to None/Optional.empty).
+  private def validateParameterAtRuntime(param: CompiledParameter, rawValue: AnyRef)(
+      implicit nodeId: NodeId
+  ): Unit = {
+    param.runtimeValidators.foreach { validator =>
+      validator.isValid(param.name, param.expressionForValidation, rawValue) match {
+        case Invalid(ParameterRuntimeValidationError(input, message)) =>
+          throw new ParameterValidationAtRuntimeException(input, message)
```
piotrp: We have access to parameter name, shouldn't we add it as exception field, like `CustomNodeValidationException` does?

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledParameter.scala
```diff
@@ -27,8 +36,20 @@ final case class CompiledParameter(
     override val expression: CompiledExpression,
     override val shouldBeWrappedWithScalaOption: Boolean,
     override val shouldBeWrappedWithJavaOptional: Boolean,
-    typingInfo: ExpressionTypingInfo
-) extends BaseCompiledParameter
+    typingInfo: ExpressionTypingInfo,
+    validators: List[Validator] = Nil,
+) extends BaseCompiledParameter {
+
+  lazy val expressionForValidation: Expression =
```
piotrp: Why do you need this copy?

This doesn't need to be `lazy`, you are just copying value references between case classes.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledParameter.scala
```diff
@@ -27,8 +36,20 @@ final case class CompiledParameter(
...
+    Expression(expression.language, expression.original)
+
+  private lazy val resolvedValidators: List[Validator] =
+    ParameterValidator.resolveLoaders(validators)
```
piotrp: You don't need this `lazy` synchronization chain, that line can be moved to a inline `val` in `runtimeValidators` initialization expression.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledParameter.scala
```diff
@@ -1,22 +1,31 @@
...
+import pl.touk.nussknacker.engine.graph.expression.Expression
 
 object CompiledParameter {
 
   def apply(
       typedExpression: TypedExpression,
-      parameterDefinition: Parameter
+      parameterDefinition: Parameter,
+  ): CompiledParameter =
+    apply(typedExpression, parameterDefinition, Nil)
```
piotrp: Why two constructors when we can just add default value to the `validators` parameter (`validators: List[Validator] = Nil`)?

designer/server/src/main/scala/pl/touk/nussknacker/ui/validation/ScenarioPropertiesValidator.scala
```diff
@@ -1,12 +1,12 @@
 package pl.touk.nussknacker.ui.validation
 
-import cats.data.Validated
+import cats.data.{NonEmptyList, Validated}
```
piotrp: nit: unused

docs/MigrationGuide.md
mgoworko: Migration guide and changelog updated, I did not commit changes in those files before, sorry about that.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/compiledgraph/CompiledParameter.scala
mgoworko: Yes, can be non-lazy. It is a performance optimization. When we call the validator in runtime, it requires instance of Expression. This optimization reduces number of class allocations.

scenario-compiler/src/main/scala/pl/touk/nussknacker/engine/expression/ExpressionEvaluator.scala
mgoworko: This behavior did not change, and I don't want to change it in this PR. Please look at the original code:
```
    try {
      val valueWithModifiedContext = evaluate[AnyRef](param.expression, param.name.value, nodeId.id, ctx)
      <something>
    } catch {
      case NonFatal(ex) => throw CustomNodeValidationException(ex.getMessage, Some(param.name), ex)
    }
```

In my change I excluded executing the validator from the try/catch block, so that exceptions from validator are thrown (there is test for that). But the evaluation itself was already inside the try/catch block before my changes, and as I said, in this PR I would not want to change it.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/definition/ParameterValidator.scala
mgoworko: It seems that we don't, removed.

components-api/src/main/scala/pl/touk/nussknacker/engine/api/exception/ParameterRuntimeValidationException.scala
```diff
@@ -0,0 +1,8 @@
+package pl.touk.nussknacker.engine.api.exception
+
+case class ParameterRuntimeValidationException(
+    paramName: pl.touk.nussknacker.engine.api.parameter.ParameterName,
```
piotrp: Why do you use fully qualified class name here?

piotrp (approved):

PR #9399, opened 2026-06-30
Scenario compare improvements

mgoworko:
## Describe your changes
Screenshot from 2026-07-02 19-50-15
Screenshot from 2026-07-02 19-50-18
Screenshot from 2026-07-02 19-50-58
Screenshot from 2026-07-02 19-51-04
Screenshot from 2026-07-02 19-51-10
Screenshot from 2026-07-02 19-51-23


designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -27,18 +31,78 @@ import type { ProcessVersionType } from "../Process/types";
 import { PropertiesForm } from "../properties";
 import { CompareContainer, CompareModal, VersionHeader } from "./Styled";
 
+type Environment = "local" | "remote";
```
lciolecki, line 32: I wonder, shouldn't these come from BE?

designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -27,18 +31,78 @@ import type { ProcessVersionType } from "../Process/types";
...
+                        padding: "8px 12px",
+                        cursor: "pointer",
+                        textAlign: "center",
+                        fontSize: "0.85em",
+                        opacity: 0.7,
+                        borderTop: "1px solid rgba(128,128,128,0.2)",
+                        "&:hover": { opacity: 1, background: "rgba(128,128,128,0.1)" },
+                    })}
+                >
+                    Load older versions…
```
lciolecki: Missing translations? Check also other textes?

designer/client/test/CompareVersionDialog-test.tsx
```diff
@@ -144,7 +169,7 @@ describe("CompareVersionsDialog", () => {
 
         fireEvent.keyDown(screen.getByText("Select..."), DOWN_ARROW);
 
-        const historyItemText = "34 - created by admin 2024-05-31|00:00";
+        const historyItemText = "34 - 2024-05-31|00:00 admin";
```
lciolecki: Why did  you change this text format?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -224,6 +225,38 @@ class ProcessesResources(
               .map(processToolbarService.getScenarioToolbarSettings(_).asJson)
           }
         }
+      } ~ path("processes" / ProcessNameSegment / VersionIdSegment / "versions-with-differences") {
+        (processName, currentVersionId) =>
+          (get & processId(processName) & parameters(Symbol("offset").as[Int].withDefault(0))) {
+            (processId, offset) =>
```
lciolecki: Waht is offset? Shouldn't we pass here information about env?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
```diff
@@ -117,6 +118,43 @@ class RemoteEnvironmentResources(
             }
           }
         } ~
+        path(ProcessNameSegment / VersionIdSegment / "versions-with-differences") {
+          (processName, currentLocalVersionId) =>
```
lciolecki, line 130: I don't like business logic in Resources / Controller. Can we extract some service for it?

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ScenarioGraphComparator.scala
```diff
@@ -59,8 +64,40 @@ object ScenarioGraphComparator {
...
+      case StickyNoteDifferent(id, _, _)      => s"Note '$id' modified"
+      case StickyNotePresentInOther(id, _)    => s"Note '$id' added"
+      case StickyNotePresentInCurrent(id, _)  => s"Note '$id' removed"
+      case PropertiesDifferent(_, _)          => "Properties modified"
+    }.toList
+
+  private def isLayoutOnlyNodeDiff(current: NodeData, other: NodeData): Boolean =
+    nodeJsonWithoutLayout(current) == nodeJsonWithoutLayout(other)
+
+  private def nodeJsonWithoutLayout(node: NodeData): Json = {
```
lciolecki: Maybe better is create some DTO without those fields then just dropping them?

examples/dev/local-testing.docker-compose.yml
```diff
@@ -64,7 +64,7 @@ services:
   ### KAFKA-related services:
 
   kafka:
-    image: bitnami/kafka:3.7.0
+    image: bitnamilegacy/kafka:3.7.0
```
lciolecki, line 67: Iv'e already fixed these and RunEnv.. So, you can rebase :)

lciolecki (changes requested):

designer/client/test/CompareVersionDialog-test.tsx
mgoworko: restored

designer/client/src/components/modals/CompareVersionsDialog.tsx
mgoworko: It is an internal mode. The FE label shows the local/remote and the environment name.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
mgoworko: I introduced pagination - 10 versions are fetched immediately, and then more can be loaded. The offset is a parameter of pagination.

designer/server/src/main/scala/pl/touk/nussknacker/ui/util/ScenarioGraphComparator.scala
mgoworko: Refactored a bit and added comment. It is used only to compare scenarios inside private method.

mgoworko: refactored even more, using exhaustive matching instead of operations on json fields

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
mgoworko: done

designer/client/src/components/graph/node-modal/fragment-input-definition/TypeSelect.tsx
```diff
@@ -63,16 +66,17 @@ export function TypeSelect({
 
     return (
         <NodeValue marked={isMarked} onKeyDown={preventEsc} sx={{ width: "100%" }}>
-            <CreatableSelect
+            <CreatableSelect<Option>
```
lciolecki, line 78: Option, here? Without ending tag?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
lciolecki: Why not move all these expressions to VersionsWithDifferencesService.compute?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/RemoteEnvironmentResources.scala
```diff
@@ -117,12 +120,48 @@ class RemoteEnvironmentResources(
...
+            }
+        } ~
         path(ProcessNameSegment / "versions") { processName =>
           (get & processId(processName)) { processId =>
             complete {
               remoteEnvironment.processVersions(processId.name)
             }
           }
+        } ~
+        path(ProcessNameSegment / "activity" / "activities") { processName =>
```
lciolecki: path(ProcessNameSegment / "activities")  ?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/HttpRemoteEnvironment.scala
```diff
@@ -69,6 +72,65 @@ class HttpRemoteEnvironment(
       result.fold(_ => List(), _.historyUnsafe)
     }
 
+  // Implementations should not fail the returned Future - errors resolve to an empty map, since a partial
+  // or incompatible response should just show fewer versions rather than break the comparison feature.
+  override def scenarioGraphsForVersions(
```
lciolecki: Do we need these (ai) comments ?:) If yes, then use scaladoc.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
```diff
@@ -0,0 +1,66 @@
...
+import pl.touk.nussknacker.ui.util.ScenarioGraphComparator
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// Computes one page of "versions with meaningful differences from the current graph", shared by the local
+// (ProcessesResources) and remote (RemoteEnvironmentResources) versions-with-differences routes, so the actual
+// diffing/pagination/filtering logic doesn't live in - and isn't duplicated across - the REST resource classes.
+object VersionsWithDifferencesService {
+
+  val PageSize: Int = 10
```
lciolecki: Why not send it from FE?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/HttpRemoteEnvironment.scala
```diff
@@ -69,6 +72,65 @@ class HttpRemoteEnvironment(
...
+      .recover { case NonFatal(ex) =>
+        logger.warn(s"Failed to fetch activities from remote environment for scenario ${processName.value}", ex)
+        List.empty
+      }
+
+  // A 404 most likely means the remote environment is running an older Nussknacker version that doesn't
+  // expose this endpoint yet - expected and not actionable, so it's only logged at debug. Any other status
+  // (5xx, auth failures, ...) is a real problem on a reachable, up-to-date remote and is logged at warn,
+  // same severity as the NonFatal-exception branch below, so it isn't silently invisible in the logs.
+  private def logFetchError(what: String, processName: ProcessName, error: NuDesignerError): Unit = error match {
```
lciolecki, line 179: Error but we log debug?

mgoworko: It is an error of fetch operation, but not treated as critical in this context. Changed to warn

designer/client/src/components/graph/node-modal/fragment-input-definition/TypeSelect.tsx
mgoworko: it is generic type parameter

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/HttpRemoteEnvironment.scala
mgoworko: removed most of them

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
mgoworko: already done after previous iteration

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -242,6 +266,28 @@ class ProcessesResources(
...
           }
+      } ~ path("processes" / ProcessNameSegment / "versions-with-differences") { processName =>
+        (post & processId(processName) & versionsToCompare) { (processId, limit) =>
+          // both must precede `entity`, which streams and parses the body
+          canRead(processId) {
+            withSizeLimit(VersionsWithDifferencesService.MaxSuppliedGraphBytes) {
+              entity(as[ScenarioGraph]) { suppliedGraph =>
+                // The byte limit still admits a graph with far more elements than any real scenario, and
+                // each element costs a diff entry per version compared.
+                if (suppliedGraph.nodes.size + suppliedGraph.edges.size > VersionsWithDifferencesService.MaxSuppliedGraphElements) {
```
lciolecki: I think this logic should be moved to service as well?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
```diff
@@ -22,7 +26,18 @@ trait RemoteEnvironment {
       remoteProcessVersion: Option[VersionId]
   ): Future[Either[NuDesignerError, Map[String, Difference]]]
 
-  def processVersions(processName: ProcessName): Future[List[ScenarioVersion]]
+  def processVersions(processName: ProcessName): Future[RemoteScenarioVersions]
+
+  // The remote does the comparing, so this sends one graph and receives a summary rather than pulling its
```
lciolecki: Do we need all these comments? :)

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
```diff
@@ -0,0 +1,228 @@
...
+    implicit val encoder: Encoder[VersionsWithDifferences] =
+      deriveEncoder[VersionsWithDifferences].mapJson(_.deepDropNullValues)
+
+    // Spelled out rather than derived: a derived decoder does not apply case-class defaults, so an
+    // answer that omits `versionComments` would fail to decode instead of meaning "none".
+    implicit val decoder: Decoder[VersionsWithDifferences] = Decoder.instance { cursor =>
+      for {
+        versions          <- cursor.get[List[VersionWithDifference]]("versions")
+        oldestCompared    <- cursor.get[Option[VersionId]]("oldestComparedVersionId")
+        comments          <- cursor.getOrElse[Map[Long, String]]("versionComments")(Map.empty)
```
lciolecki: So maybe versionComments should by Option? And why don't use there @JsonCodec with decode only?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
```diff
@@ -0,0 +1,228 @@
...
+      )
+      commentsFuture = remoteEnvironment.activities(processIdWithName.name).map(versionComments)
+      remoteResult <- remoteResultFuture
+      comments     <- commentsFuture
+    } yield remoteResult
+      .map(_.copy(versionComments = comments))
+      .getOrElse(VersionsWithDifferences(Nil, remoteUnavailable = Some(true)))
+  }
+
+  // a comment belongs to the version whose id its activity carries; the newest one describes it
```
lciolecki: Like other places - do we need comment here?

docs/MigrationGuide.md
```diff
@@ -73,6 +73,13 @@ To see the biggest differences please consult the [changelog](Changelog.md).
...
...ersions-with-differences`, which takes a scenario graph in the request body and answers with the subset of that environment's own versions that differ from it. The comparison deliberately runs on the environment that already stores those versions, so one graph is sent and a summary comes back, rather than full scenario graphs being pulled across the network to be compared locally. One request answers for as many of the scenario's most recent versions as the `limit` query parameter asks for (default 50, maximum 500) - the answering environment applies that limit before loading any graph, so it never walks its whole history, and reports `oldestComparedVersionId` when it stopped short. Version comments are returned separately, in `versionComments`, covering every version rather than only the compared ones - they come from the activity list and cost no scenario graph. The endpoint requires the `Read` permission on the scenario, and rejects a request body larger than 32 MB. Comparing scenario versions against a remote environment therefore requires that environment to run Nussknacker 1.19.0 or newer; against an older one the designer reports that the differences could not be determined.
```
lciolecki: Docs and code disagree on the body cap. The guide says 32 MB, `withSizeLimit` gets `MaxSuppliedGraphBytes`, which is 8.

designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -25,19 +28,157 @@ import { PathsToMarkProvider } from "../graph/node-modal/PathsToMark";
...
+        versions.map(({ versionId, changedElements, differencesUnknown, totalChangedElements }) => [
+            versionId,
+            { changedElements, differencesUnknown, totalChangedElements },
+        ]),
+    );
+
+const describeVersionDiff = (diffInfo: VersionDiffInfo | undefined, t: TFunction): string => {
+    if (!diffInfo) return "";
+    if (diffInfo.differencesUnknown) {
+        return t("dialog.compareVersions.unknownDifferences", "Unable to determine differences with the remote environment");
```
lciolecki: This blames the remote environment, but it fires locally too.

`describeVersion` sets `differencesUnknown = true` whenever `fetchGraphs` came back without a graph, regardless of who asked. Locally that happens when `ProcessMarshaller.fromJson` can't decode a stored version - the repo skips it and logs a warning. Rare, sure. But then someone comparing two local versions gets told about a remote environment that isn't in the picture at all.

Reword it without naming a source, e.g. "Could not determine the differences for this version". The contract is fine as is, keeping the version in the list beats dropping it.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -44,6 +45,22 @@ class ProcessesResources(
...
+      case limit if VersionsWithDifferencesService.isValidLimit(limit) => provide(limit)
+      case _ =>
+        Directive[Tuple1[Int]] { _ =>
+          complete(
+            StatusCodes.BadRequest,
+            s"limit must be between ${VersionsWithDifferencesService.MinVersionsCompared} " +
+              s"and ${VersionsWithDifferencesService.MaxVersionsCompared}"
+          )
+        }
+    }
```
lciolecki: `versionsToCompare` is copy-pasted between this and `RemoteEnvironmentResources.scala:51-63`, comment included.

Both copies already reach into `VersionsWithDifferencesService` for `DefaultVersionsCompared`, `isValidLimit` and the two bounds, so the directive may as well live in its companion. As it stands, changing the bounds means remembering two files and nothing breaks if you forget one.

`new VersionsWithDifferencesService(processService)` in each class body is fine by me, it's stateless and its only dependency is already injected. Just the directive.

designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -25,19 +28,157 @@ import { PathsToMarkProvider } from "../graph/node-modal/PathsToMark";
...
+
+type VersionDiffInfo = {
+    changedElements: string[];
+    differencesUnknown: boolean;
+    totalChangedElements?: number;
+};
+
+const MAX_DESCRIBED_CHANGES = 20;
+
+type VersionDiffsResponse = {
```
lciolecki: `VersionDiffsResponse` is the same shape as `VersionsWithDifferencesResponse`, which `HttpService.ts:195` already exports. It'll drift the first time the response grows a field. Drop it and import the real one.

Also three imports from the same module in a row up at lines 15-17. First two merge; the third has to stay separate because of `consistent-type-imports`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
```diff
@@ -0,0 +1,228 @@
...
+  // means loading and diffing its stored graph, so the whole history is not something to walk by default.
+  val DefaultVersionsCompared = 50
+
+  val MinVersionsCompared = 1
+
+  val MaxVersionsCompared = 500
+
+  val MaxGraphsPerFetch = 25
+
+  val MaxChangedElementsPerVersion = 50
```
lciolecki, line 37: Two limits truncating the same list, neither aware of the other.

Server cuts at `MaxChangedElementsPerVersion = 50` and sets `totalChangedElements`. The client then cuts the same list again at `MAX_DESCRIBED_CHANGES = 20` (`CompareVersionsDialog.tsx:42,64`). So 30 descriptions per version get built and shipped for nothing, roughly 15k strings on the floor at `limit=500`. The bigger problem is that the visible cutoff now lives in two places.

I wouldn't fix it by lowering the server to 20 though. `MaxChangedElementsPerVersion` is part of a wire contract with a remote environment that may be on a different Nussknacker version, and pinning it to what today's dialog renders is the wrong coupling.

What I'd do instead: delete `MAX_DESCRIBED_CHANGES`, render what the server sent, take the "…and N more" count from `totalChangedElements`. Truncation then happens once, server side, and `MaxChangedElementsPerVersion` can be tuned to a sane tooltip length (~25) for its own reasons. The opposite fix works for me too, as long as there's one limit instead of two.

lciolecki (commented):
Went through the version-comparison rework. Five new comments below, alongside the four I had drafted earlier. Nothing here blocks a merge except the docs/code mismatch on the request body cap, and that one only because it is a published cross-environment contract.

Things I went looking for and found fine: `skipValidateAndResolve=true` only reaches `compare` and not `migrate`; the `headers = Nil` probes are still authenticated, since `request()` attaches basic auth itself; the sticky-note field names in the new `isLayoutChangeOnly` branch match the ADT; and newest-comment-wins comes out the same on the Scala and the TS side. Test coverage is solid, especially `ScenarioGraphComparatorSpec` and `VersionsWithDifferencesServiceSpec`.

One thing I deliberately left alone: the comment density in the new code. You already asked about it twice in your own drafts here, so it is yours to call.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
mgoworko: Moved. The element-count check is now `VersionsWithDifferencesService.suppliedGraphTooLargeError`,                                                                                                                        
  and `computeAgainstSuppliedGraph` returns `Future[Either[String, VersionsWithDifferences]]` with the                                                                                                                      
  route mapping `Left` to a 400.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/RemoteEnvironment.scala
mgoworko: Pruned across the PR — dropped the ones restating the code, kept the few describing non-trivial behavior.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
mgoworko: It's `Option[Map[Long, String]]` now and the hand-written decoder is gone — `deriveDecoder` handles  the missing field.                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                             `@JsonCodec` would not be enough here, the encoder still needs `mapJson(_.deepDropNullValues)`

designer/client/src/components/modals/CompareVersionsDialog.tsx
mgoworko: Reworded to "Could not determine the differences for this version", kept the version in the list, as suggested.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
mgoworko: deduplicated

designer/client/src/components/modals/CompareVersionsDialog.tsx
mgoworko: Dropped `VersionDiffsResponse` and imported `VersionsWithDifferencesResponse`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/VersionsWithDifferencesService.scala
mgoworko: Down to one. Removed `MAX_DESCRIBED_CHANGES` and the client-side slice — the dialog renders what the server sent and takes the "…and N more" count from `totalChangedElements`.

docs/MigrationGuide.md
mgoworko: Removed size limit altogether

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/ProcessesResources.scala
```diff
@@ -142,16 +152,17 @@ class ProcessesResources(
...
+          } ~ (get & skipValidateAndResolveParameter & skipNodeResultsParameter & skipScenarioGraphParameter) {
             // FIXME: The `skipValidateAndResolve` flag has a non-trivial side effect.
             //        Besides skipping validation (that is the intended and obvious result) it causes the `dictKeyWithLabel` expressions to miss the label field.
             //        It happens, because in the current implementation we need the full compilation and type resolving in order to obtain the dict expression label.
-            (skipValidateAndResolve, skipNodeResults) =>
+            (skipValidateAndResolve, skipNodeResults, skipScenarioGraph) =>
               complete {
                 processService.getLatestProcessWithDetails(
                   processId,
                   GetScenarioWithDetailsOptions(
```
lciolecki, line 164: WDY about create companion object with skipScenarioGraph param and move this if logic inside?

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/migrate/HttpRemoteEnvironment.scala
```diff
@@ -61,14 +64,119 @@ class HttpRemoteEnvironment(
...
+              s"comparison endpoint - it likely runs a Nussknacker version without it"
+          )
+          None
+        case false => Some(VersionsWithDifferences(Nil))
+      }
+    case _ => Future.successful(None)
+  }
+
+  // HEAD rather than GET: a GET would transfer the whole graph just to discard it
+  private def scenarioExistsOnRemote(processName: ProcessName): Future[Boolean] =
```
lciolecki: The probe fails into the wrong answer.

`scenarioExistsOnRemote` is what separates "the remote doesn't have this scenario" from "the remote is too old to have the endpoint". `invokeForSuccess(...).map(_.isRight)` gives `false` for a 405 from a proxy that won't pass HEAD through, and for a 401 when our credentials are wrong. The `recover` on the next line adds every transport failure to that, with the exception bound to `_` so there's no log line either.

`emptyIfScenarioAbsent` then reads `false` as "not there" and answers `Some(VersionsWithDifferences(Nil))`. That's a success with no differences, so the dialog tells the user "No other version differs from this one" about an environment we never reached. Which is the thing the scaladoc above says this method exists to prevent.

A failed probe doesn't establish that the scenario is absent. I'd make it return `Option[Boolean]` and treat `None` the same as the `true` branch, so it lands on `None` at the call site and the user is told we couldn't compare. And log the exception.

Side note on line 133: the reason isn't right. With `transparent-head-requests` the remote still runs the whole GET route and builds the response, it only drops the body. `GET /processes/{name}?skipScenarioGraph=true` is the parameter you added in this PR, `invokeForSuccess` throws the entity away anyway, and it stops us depending on how the remote handles HEAD.

designer/server/src/main/resources/defaultDesignerConfig.conf
```diff
@@ -44,6 +44,13 @@ environment: "default"
 
 pekko {
   http {
+    client {
+      parsing {
+        # responses from a remote environment can exceed pekko's 8MB default, and exceeding it surfaces as
+        # missing data rather than as an error
+        max-content-length = 100000000 #100MB
```
lciolecki: This is global, not scoped to the remote environment.

`HttpRemoteEnvironment` uses `Http()`, the ActorSystem-wide extension, so the raised parsing limit hits every pekko-http client in the designer. Deployment managers talking to Flink or K8s, schema registry clients, whatever an extension brings in. Each in-flight response can now buffer 100MB in heap and concurrent ones stack.

I'd scope it: `ConnectionPoolSettings` carrying the bigger `ParserSettings`, passed to `http.singleRequest(request, settings = ...)` in `HttpRemoteEnvironment.request`. Same effect, and the reason ends up next to the code that needs it, which is what the comment here is already reaching for.

If it stays global it needs a Changelog entry, because it moves the memory envelope of the whole process.

designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -226,30 +444,112 @@ const VersionsForm = ({ predefinedOtherVersion }: Props) => {
...
+                    filterOption={matchVersionOption}
                     value={versionOptions.find((option) => option.value === state.otherVersion)}
                     options={versionOptions}
                     fieldErrors={[]}
+                    selectComponents={VERSION_MENU_COMPONENTS}
+                    isLoading={isLoadingVersions}
+                    isValidNewOption={noNewVersionOptions}
+                    noOptionsMessage={noVersionsMessage}
                 />
+                {state.environment === "remote" && (remoteDiffsState.unavailable || state.remoteVersionsFailed) && (
```
lciolecki: `remoteDiffsState.error` is missing from this condition.

`unavailable` means the server answered and told us it couldn't reach the remote. `error`, set on line 85, means our own request failed. Different things, and only the first one gets a message.

In the `error` case the toast is everything the user gets. Once it fades, `showUnfilteredVersions` is true, so they're looking at a full version picker with every description blank and nothing saying the comparison didn't run. Open the dialog a few seconds earlier and there's no error at all.

Local is worse. The whole block sits behind `state.environment === "remote"`, so a failed local comparison has no inline message whatsoever.

`showUnfilteredVersions` on line 214 is derived from exactly these flags, so I'd render the helper text whenever it's true, worded as "couldn't compute the differences, so all versions are listed".

designer/client/src/components/modals/CompareVersionsDialog.tsx
```diff
@@ -226,30 +444,112 @@ const VersionsForm = ({ predefinedOtherVersion }: Props) => {
...
+                        onChange={handleEnvironmentChange}
+                        value={environmentOptions.find((o) => o.value === state.environment)}
+                        options={environmentOptions}
+                        fieldErrors={[]}
+                        isValidNewOption={noNewVersionOptions}
+                    />
+                </FormControl>
+            )}
             <FormControl>
                 <FormLabel>Version to compare</FormLabel>
```
lciolecki: This is the "check also other textes" half of my earlier comment.

Everything the PR adds goes through `t()`, but this label stayed literal while the `aria-label` five lines down already uses `t("dialog.compareVersions.versionToCompare", "Version to compare")`. Screen reader gets the translated name, the visible label doesn't. Same split at 544/547 for "Difference to pick". The file-wide `eslint-disable i18next/no-literal-string` on line 1 is why nothing complained.

```suggestion
                <FormLabel>{t("dialog.compareVersions.versionToCompare", "Version to compare")}</FormLabel>
```

`:274`, `:286`, `:329`, `:333`, `:351`, `:370`, `:375` and `:439` are still literal too. Narrowing that disable would catch the next one.

lciolecki (commented):
Went through head `fa014ef878`. Of my earlier comments, 15 are actually addressed in the code and not only in the replies, including three you never answered. Two are half done: the translations (comment below), and the in-code comments, where there are still a lot of paragraphs restating the test name, and `VersionsWithDifferencesService.scala:120-121` repeats the scaladoc eleven lines above it.

Two replies have drifted from the code, both harmlessly. `suppliedGraphTooLargeError` doesn't exist and `computeAgainstSuppliedGraph` doesn't return an `Either`, because the later "removed size limit altogether" undid that. And the `@JsonCodec` question was about the `decodeOnly = true` variant, which would generate exactly the `deriveDecoder` that is now written by hand. Same outcome either way.

Only the first comment blocks. The rest can go in after the merge if you prefer.

Checked and fine: the exhaustive match on `NodeData` under `-Xfatal-warnings`, the remote failure matrix in `HttpRemoteEnvironmentSpec`, `generationRef` in `useVersionDiffs`, the `limit` bounds, and that truncation really did end up happening in one place.

lciolecki (approved):

PR #9416, opened 2026-07-20
High Availability mode for Designer

mateuszkp96:
## Checklist before merge


designer/server/src/test/scala/pl/touk/nussknacker/ui/ha/SlickDistributedLockSpec.scala
```diff
@@ -1,26 +1,26 @@
...
-  override def cleanDB(): Try[Unit] = {
-    val superResult = super.cleanDB()
-    val locksResult = Using(testDbRef.db.createSession()) { session =>
-      session.prepareStatement(s"""DELETE FROM "${getSchemaName()}"."distributed_locks"""").execute()
-    }.map(_ => ())
-    superResult.flatMap(_ => locksResult)
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    Using(testDbRef.db.createSession()) { session =>
+      session.prepareStatement(s"""DELETE FROM "${getSchemaName()}".distributed_locks""").execute()
```
piotrp: Can we use `TRUNCATE` / `TRUNCATE TABLE` here? It deletes entire data set, without iterating row-by-row.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/HaMode.scala
```diff
@@ -0,0 +1,56 @@
...
+import pl.touk.nussknacker.engine.util.config.CustomFicusInstances._
+
+import java.net.InetAddress
+import scala.concurrent.duration._
+
+sealed trait HaMode
+
+object HaMode {
+
+  case object Disabled extends HaMode
```
piotrp: How about adding `instanceId` field to disabled HA configuration? Then it would never be optional, so our API endpoint would have a simpler schema, and JSON encoder for `LeaderResponseDto` would also be a bit simpler.

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
```diff
@@ -269,6 +288,17 @@ object AppApiEndpoints {
       def apply() = new HealthCheckProcessSuccessResponseDto(status = Status.Ok, message = None, processes = None)
     }
 
+    @derive(decoder, schema)
+    final case class LeaderResponseDto private (
+        isLeader: Boolean,
+        instanceId: Option[String],
```
piotrp: We should add a marker (Boolean?) telling us whether HA mode is enabled

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
```diff
@@ -269,6 +288,17 @@ object AppApiEndpoints {
       def apply() = new HealthCheckProcessSuccessResponseDto(status = Status.Ok, message = None, processes = None)
     }
 
+    @derive(decoder, schema)
```
piotrp: Is decoder needed here?

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/AppApiHttpService.scala
```diff
@@ -32,7 +29,8 @@ class AppApiHttpService(
     modelInfos: ProcessingTypeDataProvider[ModelInfo, _],
     categories: ProcessingTypeDataProvider[String, _],
     processService: ProcessService,
-    shouldExposeConfig: Boolean
+    shouldExposeConfig: Boolean,
+    leadership: Leadership
```
piotrp: nit: maybe we should add it before `shouldExposeConfig`?
(feel free to close this thread)

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import pl.touk.nussknacker.engine.api.db.{DbRef, NuPostgresProfile}
+import slick.jdbc.JdbcBackend
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.FiniteDuration
+
+trait DistributedLock {
+
+  // Owner-bypass: succeeds if the lock is expired OR already held by this instance
```
piotrp: This deserves to be a ScalaDoc comment

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,97 @@
...
+    run(releaseLock(name))
+
+  // toSeconds truncates; JDBC setQueryTimeout(0) means "no timeout", so round up to at least 1 s
+  private val lockQueryTimeoutSeconds = Math.max(1, lockQueryTimeout.toSeconds.toInt)
+
+  private def acquireOrRenewLock(name: String, durationMillis: Long): DBIO[Boolean] = {
+    sqlu"""INSERT INTO "#${profile.schemaName}"."distributed_locks" (name, lock_until, locked_at, locked_by)
+             VALUES (
+               $name,
+               CURRENT_TIMESTAMP + CAST($durationMillis || ' milliseconds' AS INTERVAL),
```
piotrp: I believe this can be done with a bit less parsing in the database: would `INTERVAL ${s"$durationMillis ms"}` work?

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,97 @@
...
+               CURRENT_TIMESTAMP,
+               $instanceId
+             )
+             ON CONFLICT (name) DO UPDATE SET
+               lock_until = EXCLUDED.lock_until,
+               locked_at  = CURRENT_TIMESTAMP,
+               locked_by  = EXCLUDED.locked_by
+             WHERE "distributed_locks"."lock_until" < CURRENT_TIMESTAMP
+                OR "distributed_locks"."locked_by" = $instanceId"""
```
piotrp: You don't need to add `"distributed_locks".` - there is only a single table in scope.

We can also remove quotes around column names (and table name on the first line). And schema name, this is supposed to be already correctly escaped (https://scala-slick.org/doc/3.5.0/sql.html#splicing-literal-values)?

The same applies to the `releaseLock` method below.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+import scala.util.{Failure, Success}
+
+trait Leadership {
+  def isLeader(): Boolean
+  def instanceId: Option[String]
+
+  /** Registers a callback invoked once each time this node acquires leadership (false→true transition).
+    * If already a leader at registration time, the callback fires immediately.
+    */
+  def onLeadershipAcquired(callback: () => Unit): Unit
```
piotrp: It may be useful to add a `onLeadershipLost` method - this can happen when we fail to prolong leadership lock

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+    if (alreadyLeader) Future { callback() }
+  }
+
+  private def scheduleHeartbeat(): Unit = {
+    val tick = actorSystem.scheduler.scheduleOnce(config.leaderHeartbeatInterval) {
+      if (!shuttingDown.get()) {
+        val acquire = lock.acquireOrRenew()
+        inflightAcquire.set(acquire)
+        acquire.onComplete { result =>
+          inflightAcquire.set(Future.successful(false))
```
piotrp: Why this `set`? At this point `inflightAcquire` already holds a completed `Future`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+    scheduledHeartbeat.set(tick)
+  }
+
+  def stop(): Future[Unit] = {
+    shuttingDown.set(true)
+    scheduledHeartbeat.get().cancel()
+    // Wait for any in-flight acquireOrRenew before releasing: if release() runs first, the owner-bypass
+    // in acquireOrRenew (WHERE locked_by = instanceId) would re-acquire the lock after it was released,
+    // extending the lease by leaderLeaseDuration and leaving other nodes without a leader for that window.
+    inflightAcquire.get().recover { case _ => false }.flatMap(_ => lock.release())
```
piotrp: It may be useful to have this line behind some `releaseOnStop` flag, defaulting to true. Then we could allow for a somewhat deterministic leader assignment on application restart, because a restarted instance could take over its previous lock (provided that heartbeat window is long enough).

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+    leadershipCallbacks.add(callback)
+    if (alreadyLeader) Future { callback() }
+  }
+
+  private def scheduleHeartbeat(): Unit = {
+    val tick = actorSystem.scheduler.scheduleOnce(config.leaderHeartbeatInterval) {
+      if (!shuttingDown.get()) {
+        val acquire = lock.acquireOrRenew()
+        inflightAcquire.set(acquire)
+        acquire.onComplete { result =>
```
piotrp: Two things:
1. What if this takes longer than `leaderLeaseDuration`, e.g. due to thread starvation or hung socket read? Then two instances may consider themselves to be the lelader.
2. A related issue - when acquire succeeds we can be sure that the lock is valid only until `{acquireOrRenew start time} + {leaderLeaseDuration}`.

Maybe we should remember when this call started, and use this in `isLeader` to determine whether `isCurrentLeader` can be trusted when it's set to `true`?

docs/MigrationGuide.md
```diff
@@ -9,6 +9,33 @@ To see the biggest differences please consult the [changelog](Changelog.md).
 
 ## In version 1.19.0 (Not released yet)
 
+### High Availability (HA) mode
```
piotrp: This is a new optional feature so we don't need to mention it here as it doesn't influence existing Nussknacker deployments. But - we need to add it to documentation.

We should also mention that this feature requires PostgreSQL database.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  // toSeconds truncates; JDBC setQueryTimeout(0) means "no timeout", so round up to at least 1 s
+  private val lockQueryTimeoutSeconds = Math.max(1, lockQueryTimeout.toSeconds.toInt)
+
+  private def acquireOrRenewLock(name: String, durationMillis: Long): DBIO[Boolean] = {
+    sqlu"""INSERT INTO "#${profile.schemaName}"."distributed_locks" (name, lock_until, locked_at, locked_by)
+             VALUES (
+               $name,
+               CURRENT_TIMESTAMP + CAST($durationMillis || ' milliseconds' AS INTERVAL),
+               CURRENT_TIMESTAMP,
```
piotrp, line 68: The correct function to use in this query is `LOCALTIMESTAMP`, which returns a `TIMESTAMP` result, consistent with column definition. `CURRENT_TIMESTAMP` returns `TIMESTAMPTZ` (`TIMESTAMP WITH TIME ZONE`).

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+              }
+          }
+          if (!shuttingDown.get()) scheduleHeartbeat()
+        }
+      }
+    }
+    scheduledHeartbeat.set(tick)
+  }
+
+  def stop(): Future[Unit] = {
```
piotrp: This releases the lock, but if `isCurrentLeader` was set to `true` it's never switched to `false`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+  override def instanceId: Option[String] = Some(config.instanceId)
+
+  override def onLeadershipAcquired(callback: () => Unit): Unit = {
+    val alreadyLeader = isCurrentLeader.get()
+    leadershipCallbacks.add(callback)
+    if (alreadyLeader) Future { callback() }
+  }
+
+  private def scheduleHeartbeat(): Unit = {
+    val tick = actorSystem.scheduler.scheduleOnce(config.leaderHeartbeatInterval) {
```
piotrp: Isn't the current scheduling basically a `scheduleWithFixedDelay` in disguise?

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerAppFactory.scala
```diff
@@ -40,10 +40,15 @@ class NussknackerAppFactory(
...
-        // We don't wait for recovery because it may take a while, and we don't want health check to detect that we have problem with application starting
-        recoverNotRunningDeploymentsThatShouldBeRunning(domainServices.reconciler)(
-          infrastructureServices.executionContextWithIORuntime
-        )
+        // Recovery is triggered each time this node acquires leadership rather than once at startup.
+        // A node that starts as a non-leader (another node holds the lease) would otherwise never
+        // run recovery even after winning leadership following a failover.
+        // The callback is dispatched asynchronously inside LeadershipService so it never blocks the heartbeat.
+        domainServices.leadership.onLeadershipAcquired { () =>
+          recoverNotRunningDeploymentsThatShouldBeRunning(domainServices.reconciler)(
```
piotrp: You should guard against concurrent executions, because this can happen:
* leadership is acquired - `onLeadershipAcquired` is run
* a single heartbeat fails - leadership is renounced
* leadership is acquired again - `onLeadershipAcquired` is run again, so if `recoverNotRunningDeploymentsThatShouldBeRunning` was already running we now have two concurrent executions

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+    extends Leadership
+    with LazyLogging {
+
+  private val isCurrentLeader     = new AtomicBoolean(initiallyLeader)
+  private val shuttingDown        = new AtomicBoolean(false)
+  private val scheduledHeartbeat  = new AtomicReference[Cancellable](Cancellable.alreadyCancelled)
+  private val inflightAcquire     = new AtomicReference[Future[Boolean]](Future.successful(false))
+  private val leadershipCallbacks = new CopyOnWriteArrayList[() => Unit]
+
+  scheduleHeartbeat()
```
piotrp: This makes it obligatory to call `stop` if this `LeadershipService` instance is no longer used. Shouldn't this class be a `AutoCloseable` ti indicate this somehow?

designer/server/src/main/resources/db/migration/postgres/V1_072__DistributedLocksTable.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE "distributed_locks"
+(
+    "name"       VARCHAR   NOT NULL PRIMARY KEY,
+    "lock_until" TIMESTAMP NOT NULL,
+    "locked_at"  TIMESTAMP NOT NULL,
+    "locked_by"  VARCHAR   NOT NULL
+);
```
piotrp: nit: both new .sql files should have a trailing newline, as our EditorConfig dictates

designer/server/src/test/scala/pl/touk/nussknacker/ui/api/AppApiHttpServiceHaSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+          |  instanceId: "test-instance"
+          |  leaderLeaseDuration: 30s
+          |  leaderHeartbeatInterval: 2s
+          |  lockQueryTimeout: 1s
+          |}""".stripMargin
+      )
+      .withFallback(super.designerRawConfig)
+
+  "The app leader endpoint should" - {
+    "return isLeader=true with instanceId when HA is enabled and this node wins leadership" in {
```
piotrp, line 51: Tests should clean up after themselves

designer/server/src/test/scala/pl/touk/nussknacker/test/base/it/NuItTestWithPostgres.scala
```diff
@@ -0,0 +1,48 @@
...
+import pl.touk.nussknacker.ui.factory.NussknackerAppFactory
+
+trait NuItTestWithPostgres
+    extends WithPostgresDbTesting
+    with DefaultUniquePortProvider
+    with WithClock
+    with BeforeAndAfterAll {
+  this: Suite with WithDesignerConfig =>
+
+  private val port = nextPort()
```
piotrp: `nextPort` should be called in `beforeAll`, just before the server is started, to reduce the chance of port collisions

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,112 @@
...
+          }
+          if (!shuttingDown.get()) scheduleHeartbeat()
+        }
+      }
+    }
+    scheduledHeartbeat.set(tick)
+  }
+
+  def stop(): Future[Unit] = {
+    shuttingDown.set(true)
```
piotrp: There is still a possible race here:
* `scheduleHeartbeat` enters `lock.acquireOrRenew()` because `shuttingDown == false`
* `stop` sets `shuttingDown = true`
* `stop` calls and completes `lock.release()` because `inflightAcquire` contained a finished `Future` from the previous heartbeat run
* `scheduleHeartbeat`'s `lock.acquireOrRenew()` renews the lock

IMO `shuttingDown` and `inflightAcquire` should be represented by a single state atomic, e.g. `Idle` -> `Renewing(Future)` -> `Stopped`, which can be switched between states by CAS operation.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/HaMode.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  final case class Enabled(
+      instanceId: String,
+      leaderLeaseDuration: FiniteDuration,
+      leaderHeartbeatInterval: FiniteDuration,
+      periodicLockDuration: FiniteDuration,
```
piotrp, line 26: Can we validate it against `SchedulingConfig.deployInterval` and `SchedulingConfig.rescheduleCheckInterval`?

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
piotrp: My bad about the alias, you need `INSERT ... distributed_locks AS dl` and then you can use shorter syntax: `WHERE dl.lock_until ...`

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
mateuszkp96: Decoder is needed because Tapir's jsonBody[T] requires both Encoder and Decoder regardless of whether the endpoint is input or output

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/LeaderLock.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  private val name = "designer-leader"
+
+  def acquireOrRenew(): IO[LeaderLockResult] =
+    for {
+      // validUntil is measured before the DB call — conservative by design: the actual DB-side
+      // lock_until is slightly later (LOCALTIMESTAMP + duration), but accounting for call duration
+      // as "lost time" protects isLeader() from returning true after the lock has expired.
+      validUntil <- IO(clock.instant().plusMillis(duration.toMillis))
+      result <- IO.fromFuture(IO(underlying.acquireOrRenew(name, duration))).map {
```
mateuszkp96: `validUntil` could come directly from the DB via `RETURNING lock_until` instead of being pre-calculated — but only if the column is `TIMESTAMP WITH TIME ZONE`. With `TIMESTAMP WITHOUT TIME ZONE`,               `java.sql.Timestamp.toInstant()` uses the JVM timezone for conversion, which gives the wrong result when the JVM and DB session timezones differ. I can fix it in the next commits.

piotrp: You should be able to use `RETURNING EXTRACT(EPOCH FROM lock_until) AS lock_until_epoch`  and construct your `Instant` from that

designer/server/src/main/scala/pl/touk/nussknacker/ui/api/description/AppApiEndpoints.scala
```diff
@@ -288,17 +287,13 @@ object AppApiEndpoints {
       def apply() = new HealthCheckProcessSuccessResponseDto(status = Status.Ok, message = None, processes = None)
     }
 
-    @derive(decoder, schema)
+    @derive(encoder, decoder, schema)
     final case class LeaderResponseDto private (
+        instanceId: String,
+        isHaEnabled: Boolean,
```
piotrp: Should we drop the `is` prefix here?

designer/server/src/main/scala/pl/touk/nussknacker/ui/factory/NussknackerAppFactory.scala
```diff
@@ -38,34 +36,32 @@ class NussknackerAppFactory(
...
-  private def recoverNotRunningDeploymentsThatShouldBeRunning(
-      reconciler: ScenarioDeploymentReconciler
-  )(implicit ec: ExecutionContext): Unit = {
-    reconciler.recoverNotRunningDeploymentsThatShouldBeRunning(_.recoverJobsOnStart).onComplete {
-      case Success(_) =>
-      case Failure(
-            exception
-          ) => // It is done just in case, it rather shouldn't happen because we have exception handling for each recovered deployment
-        logger.error("Error while deployments recovery", exception)
+  // Recovery is triggered each time this node acquires leadership rather than once at startup.
```
piotrp: A class-level method in production code deserves a Scaladoc comment, not inline comments

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/HaMode.scala
```diff
@@ -25,30 +31,48 @@ object HaMode {
...
     }
 
+  private def defaultInstanceId: String =
+    Try(InetAddress.getLocalHost.getHostName).getOrElse(
+      throw new IllegalArgumentException(
+        "Cannot determine instanceId automatically (getLocalHost failed). Set ha.instanceId explicitly in the config."
+      )
+    )
+
+  private final case class LeaderEnabledConfig(
```
piotrp: We can add these defaults to `LeaderConfig` and remove this class, and simplify field assignment in `fromConfig`

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
+
+  // Called after the heartbeat fiber is cancelled — see Resource.make in startHeartbeat
+  private def stop(): IO[Unit] =
+    for {
+      wasLeader <- IO(state.getAndSet(LeaderLockResult.NotAcquired).isAcquired)
+      released  <- if (wasLeader && leaderConfig.releaseOnStop) releaseLock else IO.pure(false)
+      _         <- if (released) notifyLeadershipLostAndWait() else IO.unit
+    } yield ()
+
+  private def releaseLock: IO[Boolean] =
```
piotrp: nit: _Mutator-like named method is parameterless_

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
-  override def isLeader(): Boolean        = isCurrentLeader.get()
-  override def instanceId: Option[String] = Some(config.instanceId)
+  override val isHaEnabled: Boolean = true
 
-  override def onLeadershipAcquired(callback: () => Unit): Unit = {
-    val alreadyLeader = isCurrentLeader.get()
-    leadershipCallbacks.add(callback)
-    if (alreadyLeader) Future { callback() }
-  }
+  override def onLeadershipAcquired(callback: () => IO[Unit]): IO[Unit] =
```
piotrp: We can move implementation of both methods to the `Leadership` trait. The only difference will be that `SingleNodeLeadership` will keep a list of `onLeadershipLost` callbacks, but that shouldn't be a problem.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
   def isLeader(): Boolean
-  def instanceId: Option[String]
+  def instanceId: String
+  def isHaEnabled: Boolean
 
   /** Registers a callback invoked once each time this node acquires leadership (false→true transition).
-    * If already a leader at registration time, the callback fires immediately.
+    * If the node is already a leader when [[startHeartbeat]] is called, the callback fires then.
+    * If the previous invocation is still in-progress when leadership is re-acquired, the callback is skipped.
+    * Callbacks must be registered before [[startHeartbeat]] is called.
```
piotrp, line 27: Let's enforce this in both methods, to prevent interface misuse. It's fine if this check is simple and not fully thread safe.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
     */
-  def onLeadershipAcquired(callback: () => Unit): Unit
+  def onLeadershipAcquired(callback: () => IO[Unit]): IO[Unit]
```
piotrp: `() => IO[Unit]` - isn't `IO` already lazy? If we really need to construct a new `IO` on each invocation there should be an explanation on `CallbackState` class. Or am I misunderstanding how IO works?

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
+              (if (initialResult.isAcquired) handleLeadershipAcquired() else IO.unit) >>
+              heartbeatLoop.start
           }
-          if (!shuttingDown.get()) scheduleHeartbeat()
-        }
+      )(fiber => fiber.cancel >> stop())
+      .void
+
+  // poll(sleep) is cancellable; heartbeat is not — fiber.cancel waits for in-flight acquireOrRenew
+  private def heartbeatLoop: IO[Unit] =
```
piotrp: Shouldn't this be a composition of a cancelable sleep and an uncancelable `heartbeat` call?

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
-                logger.warn("Leader heartbeat failed — relinquishing leadership", ex)
-              }
+  override def onLeadershipLost(callback: () => IO[Unit]): IO[Unit] =
+    IO(leadershipLostCallbacks.add(new CallbackState(callback, "lost"))).void
+
+  override def startHeartbeat(): Resource[IO, Unit] =
+    Resource
+      .make(
+        lock
+          .acquireOrRenew()
```
piotrp: We can reuse `tryAcquireLock` here if we add error message as its argument

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -1,112 +1,190 @@
...
+  private def releaseLock: IO[Boolean] =
+    lock
+      .release()
+      .handleError { ex =>
+        logger.warn("Failed to release leader lock on stop — lease will expire naturally", ex)
+        false
+      }
+
+  private def notifyLeadershipLostAndWait(): IO[Unit] =
+    leadershipLostCallbacks.asScala.toList.traverse_(_.fireAndWait())
```
piotrp: We should add a INFO log, same as we do in `handleLeadershipLost`

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/PeriodicLock.scala
```diff
@@ -1,26 +1,32 @@
...
 
 import cats.effect.{IO, Resource}
+import com.typesafe.scalalogging.LazyLogging
 import pl.touk.nussknacker.ui.ha.{DistributedLock, HaMode, NoOpDistributedLock}
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
-final class PeriodicLock(underlying: DistributedLock, duration: FiniteDuration) {
+final class PeriodicLock(underlying: DistributedLock, val lockDuration: Option[FiniteDuration]) {
```
piotrp: `PeriodicDeploymentLock` (with `name = "periodic-process-deployment"`)? Currently the name suggests that it's some kind of generic recurring lock.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  private def heartbeat: IO[Unit] =
+    for {
+      result    <- tryAcquireLock
+      wasLeader <- IO(state.getAndSet(result).isAcquired)
+      _         <- if (result.isAcquired && !wasLeader) handleLeadershipAcquired() else IO.unit
+      _         <- if (!result.isAcquired && wasLeader) handleLeadershipLost() else IO.unit
+    } yield ()
+
+  private def tryAcquireLock: IO[LeaderLockResult] =
```
piotrp: Can we run a async timer that will log a warning when `acquireOrRenew` this takes longer than `lockQueryTimeout`, or when it's stuck?

designer/server/src/main/resources/db/migration/postgres/V1_072__DistributedLocksTable.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE "distributed_locks"
+(
+    "name"       VARCHAR   NOT NULL PRIMARY KEY,
+    "lock_until" TIMESTAMP NOT NULL,
```
piotrp: I thought some more on that and came to the conclusion that we should switch to `TIMESTAMP WITH TIME ZONE` (`TIMESTAMPTZ`) here and use `CURRENT_TIMESTAMP` in queries... current version generally works, but we will have issues during DST switches, and we can't tell everybody to use UTC on the server.

This will be inconsistent in how Nussknacker handles timestamps in other columns, but gives us the least tricky code.

An easier alternative (but more tricky to use) is to replace `LOCALTIMESTAMP`, and currently incorrect `CURRENT_TIMESTAMP` in new tests, with `now() AT TIME ZONE 'UTC'`.

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/periodic/DeploymentActor.scala
```diff
@@ -2,79 +2,122 @@ package pl.touk.nussknacker.ui.process.periodic
...
+  override def receive: Receive = inState(handleIdle)
+
+  private def inState(f: Msg => Unit): Receive = { case msg: Msg => f(msg) }
+
+  private def handleIdle(msg: Msg): Unit = msg match {
     case CheckToBeDeployed =>
       logger.trace("Checking scenarios to be deployed")
-      findToBeDeployed.onComplete {
-        case Success(runDetailsSeq) =>
+      findToBeDeployed
```
piotrp: We should acquire lock before this call, it's expensive - all pending scenarios are checked on Flink. Then we should `acquireOrRenew` before deployment attempt just in case this took a long time to execute.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -51,27 +53,30 @@ class SlickDistributedLock(
...
+    sql"""INSERT INTO #${profile.schemaName}.distributed_locks AS dl (name, lock_until, locked_at, locked_by)
+           VALUES (
+             $name,
+             CURRENT_TIMESTAMP + INTERVAL '#${durationMillis} milliseconds',
+             CURRENT_TIMESTAMP,
+             $instanceId
+           )
+           ON CONFLICT (name) DO UPDATE SET
+             lock_until = EXCLUDED.lock_until,
+             locked_at  = CURRENT_TIMESTAMP,
```
piotrp: For consistency: `locked_at = EXCLUDED.locked_at`?

scalafix-rules/src/test/scala/utils/SchemaInSqlDetectorSpec.scala
```diff
@@ -48,6 +48,28 @@ class SchemaInSqlDetectorSpec extends AnyFunSuite {
       """CREATE OR REPLACE FUNCTION "${profile.schemaName}".generate_random_uuid() RETURNS UUID AS 'BEGIN RETURN uuid_in(overlay(overlay(md5(random()::text || '':'' || random()::text) placing ''4'' from 13) placing to_hex(floor(random() * (11 - 8 + 1) + 8)::int)::text from 17)::cstring);END' LANGUAGE plpgsql;""",
       false
     ),
+    (
+      """INSERT INTO #arg.distributed_locks AS dl (name, lock_until, locked_at, locked_by)
```
piotrp: We can simplify this a bit and introduce some fake table name. And we should introduce cases that are correctly detected as invalid.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,103 @@
...
+           ON CONFLICT (name) DO UPDATE SET
+             lock_until = EXCLUDED.lock_until,
+             locked_at  = EXCLUDED.locked_at,
+             locked_by  = EXCLUDED.locked_by
+           WHERE dl.lock_until <= CURRENT_TIMESTAMP
+              OR dl.locked_by = EXCLUDED.locked_by
+           RETURNING lock_until"""
+      .as[java.sql.Timestamp]
+      .withStatementParameters(statementInit = _.setQueryTimeout(lockQueryTimeoutSeconds))
+  }.map(_.headOption.map(_.toInstant))
```
piotrp: We should ignore database's `lock_until` here and return `Instant.now() + durationMillis` (or in a more correct form - `clock.instant() + durationMillis`), so that local time comparisons will always use local clock. Otherwise we may be in a situation where we have two leaders. After this change the returned lock validity period will be guaranteed to be no longer than database-side lock.

piotrp (changes requested):

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,217 @@
...
+    clock: Clock,
+) extends Leadership {
+
+  private val state: AtomicReference[LeaderLockResult] = new AtomicReference(LeaderLockResult.NotAcquired)
+
+  override val haEnabled: Boolean = true
+
+  override def isLeader(): Boolean = {
+    state.get() match {
+      case LeaderLockResult.Acquired(validUntil) => validUntil.isAfter(clock.instant())
```
piotrp: This may be dangerous when we start using `onLeadershipLost`, currently the following is possible (`A` and `B` are Nussknacker instances):
* `A` acquires leadership lock - starts some background work in `onLeadershipAcquired`
* `A` fails to renew lock in a way that makes `heartbeatLoop` hang, so `updateLeadershipState` is never called
* `B` acquires leadership lock - starts some background work in `onLeadershipAcquired`

I think the easiest solution is to add comments to:
* `onLeadershipAcquired` - mention that callbacks should always use `isLeader` when performing work (which should be the case anyway)
* `onLeadershipLost` - mention that it is not guaranteed to be called when leadership was lost due to heartbeat timeout

These callbacks are advisory anyway, so let's make this documented.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,217 @@
...
+      wasLeader <- IO(state.getAndSet(result).isAcquired)
+      _         <- if (result.isAcquired && !wasLeader) notifyLeadershipAcquired(supervisor) else IO.unit
+      _         <- if (!result.isAcquired && wasLeader) notifyLeadershipLost(supervisor) else IO.unit
+    } yield ()
+  }
+
+  private def stop(): IO[Unit] = {
+    for {
+      wasLeader <- IO(state.getAndSet(LeaderLockResult.NotAcquired).isAcquired)
+      released  <- if (wasLeader && leaderConfig.releaseOnStop) releaseLock() else IO.pure(false)
```
piotrp: We can also add a check for non-expired leader - releasing an expired leader is a no-op and performs useless database roundtrip.

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/Leadership.scala
```diff
@@ -0,0 +1,217 @@
...
+      _         <- if (released) notifyLeadershipLostAndWait() else IO.unit
```
piotrp: If `wasLeader && !leaderConfig.releaseOnStop` then we won't log the "Leadership lost on shutdown" message

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/reconciliation/FinishedDeploymentsStatusesSynchronizationScheduler.scala
```diff
@@ -2,35 +2,37 @@ package pl.touk.nussknacker.ui.process.deployment.reconciliation
...
     Resource.make(IO {
       actorSystem.scheduler.scheduleAtFixedRate(0 seconds, config.delayBetweenSynchronizations) { () =>
-        Try(
-          Await.result(reconciler.synchronizeEngineFinishedDeploymentsLocalStatuses(), config.synchronizationTimeout)
-        ).failed.foreach { ex =>
-          logger.error(
-            s"Error during finished deployments statuses synchronization. Synchronization will be retried in ${config.delayBetweenSynchronizations}",
-            ex
-          )
+        if (leadership.isLeader()) {
```
piotrp, line 26: Let's add a comment at both `isLeader` use sites saying that the fenced operation is fast and safe to execute even if leadership is lost mid-execution

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/HaMode.scala
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      val instanceId =
+        if (config.hasPath("ha.instanceId")) config.getString("ha.instanceId")
+        else defaultInstanceId
+      Disabled(instanceId)
+    }
+
+  private def defaultInstanceId: String =
+    Try(InetAddress.getLocalHost.getHostName).getOrElse(
+      throw new IllegalArgumentException(
```
piotrp: We should pass the original exception to `IllegalArgumentException`

designer/server/src/main/scala/pl/touk/nussknacker/ui/ha/DistributedLock.scala
```diff
@@ -0,0 +1,103 @@
...
+    run(acquireOrRenewLock(name, duration.toMillis))
+
+  override def release(name: String): Future[Boolean] =
+    run(releaseLock(name))
+
+  // toSeconds truncates; JDBC setQueryTimeout(0) means "no timeout", so round up to at least 1 s
+  private val lockQueryTimeoutSeconds = Math.max(1, lockQueryTimeout.toSeconds.toInt)
+
+  private def acquireOrRenewLock(name: String, durationMillis: Long): DBIO[Option[Instant]] = {
+    sql"""INSERT INTO #${profile.schemaName}.distributed_locks AS dl (name, lock_until, locked_at, locked_by)
```
piotrp: Let's restore quotes around schema name in both places in this class (`"#${profile.schemaName}".distributed_locks`), that way it will be consistent with existing migration code. It;s still incorrect, but better than nothing. Sorry for that...

designer/server/src/main/scala/pl/touk/nussknacker/ui/process/deployment/reconciliation/ScenarioDeploymentReconciler.scala
```diff
@@ -73,40 +73,59 @@ class ScenarioDeploymentReconciler(
...
-          results              <- resultsFuture
-          resultForOneScenario <- recoverScenarioJob(processingType, deployCommand)
-        } yield results :+ resultForOneScenario
-    }
+      runDeploymentCommandsByProcessingType: List[(ProcessingType, DMRunDeploymentCommand)],
+      isLeader: () => Boolean,
+  )(implicit user: LoggedUser): Future[List[Try[Unit]]] = {
+    def loop(
+        remaining: List[(ProcessingType, DMRunDeploymentCommand)],
+        results: List[Try[Unit]],
```
piotrp: nit: can be a `Vector`, then `results :+ result` would be `O(1)`, not `O(n)`

docs/Changelog.md
```diff
@@ -15,6 +15,39 @@ description: Stay informed with detailed changelogs covering new features, impro
 
 ### 1.19.0 (Not released yet)
 
+* [#9416](https://github.com/TouK/nussknacker/pull/9416) Feature: High Availability (HA) mode for the Designer.
```
piotrp, line 18: HA mode's configuration should be described in documentation, along with new API endpoints

piotrp (approved):

